PR #177, opened 2014-09-29
added packages 

JosephineYu:
gfanInterface2, PolyhedralObjects, Polyhedra2, PolymakeInterface

### DIFF
--- a/M2/Macaulay2/packages/Polyhedra2.m2
+++ b/M2/Macaulay2/packages/Polyhedra2.m2
@@ -1,0 +1,1311 @@
+--*- coding: utf-8 -*-
+---------------------------------------------------------------------------
+--
+-- PURPOSE: Computations with convex polyhedra 
+-- PROGRAMMER : Nathan Ilten, Josephine Yu, Qingchun Ren 
+-- UPDATE HISTORY : August 2012 
+---------------------------------------------------------------------------
+newPackage("Polyhedra2",
+    Headline => "A package for computations with convex polyhedra",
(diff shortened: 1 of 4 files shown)


DanGrayson:
Josephine, will `gfanInterface2` eventually replay `gfanInterface`?  We probably don't want two such packages.  Are the others intended as replacements, too?

JosephineYu:
Hi Dan (cc: Diane)

Yes, gfanInterface2 should replace gfanInterface.  I don't know if anybody
(or any packages) still use gfanInterface.  Diane and her students have
been working on gfanInterface2 recently, so this version should be replaced
with what Diane has.

Josephine

On Wed, Jul 27, 2016 at 8:23 AM, Daniel R. Grayson <notifications@github.com

lkastner:
Some of these packages are related to refactoring Polyhedra.m2.

- Polyhedra2.m2 seems to be a replacement for Polyhedra.m2 with lazy evaluation and optional polymake interface. Since it does not use CacheTables and it has no tests, I ignored it for now. However, when I am finished with Polyhedra.m2 I would like to talk about refactoring this package or maybe even incorporating it. Please see my M2 fork for the refactoring progress.
- PolyhedralObjects.m2 contains the basic definitions for polyhedral objects, used in Polyhedra2.m2 (and also in Polyhedra.m2). We had the discussion whether it makes sense to have these objects without their polyhedral functionality and I share the opinion that it does not. Is there any other reason to have this separate?

Issue #208, opened 2014-12-31
pairs of list


zteitler:
Something like this would add a bit of orthogonality:

    i13 : pairs {a,b,c,d}
    stdio:13:1:(3): error: expected a hash table or a raw polynomial

    i12 : pairs List := x -> apply(#x, i -> (i,x#i));

from: bugs/dan/0-pairs-of-list

zteitler:
Ah, I made a mistake in this. Bad commit, sorry. I'm fixing it right now---it is taking a little bit of time to check things carefully (I don't want to make *another* mistake).

(The documentation key (pairs, BasicList) didn't match the contents of the documentation node, and also there needs to be a method installed. Those changes should have happened at the same time as putting in the documentation key. I didn't check before committing. Well, they are fixed on my computer & I'm building M2 right now. If it builds and the documentation is okay then you will see one more commit here.)

DanGrayson:
I'm already fixing it, Zach...

DanGrayson:
See
https://github.com/Macaulay2/M2/commit/a1ff04ac15b99e9790df0016fd5f97f294959c11
for my
changes, which partially fixed things.

zteitler:
Okay!

Issue #201, opened 2014-12-03
assert doc


DanGrayson:
The documentation for "assert" should link to the documentation for "generateAssertions".

DanGrayson:
Thanks, Zach!

Issue #173, opened 2014-09-18
The documentation should mention the difference between frac QQ[x] and frac (QQ[x])


tom111:
frac QQ [x] yields (somewhat surprisingly) a polynomial ring since the parser evaluates this as (frac QQ)[x]. The "Caveat" section in the documentation of 'frac' may be a good place to put this.

PR #209, opened 2014-12-31
setRandomSeed doc, assert doc, frac doc, pairs of list

zteitler:
1. documentation for setRandomSeed() revised for clarity
2. #201: improve assert doc, link to "generateAssertions"
3. #173: documentation (frac QQ)[x] vs frac (QQ[x])
4. #208: pairs of list

The first 3 are just documentation improvements; of course proof-read them, but hopefully they are uncontroversial. The 4th involves changes in the "d" directory: adding functionality to the "pairs" CompiledFunction. I didn't touch the existing functionality (for dictionaries and hash tables).

### DIFF
--- a/M2/Macaulay2/packages/Macaulay2Doc/doc4.m2
+++ b/M2/Macaulay2/packages/Macaulay2Doc/doc4.m2
@@ -121,7 +121,7 @@ document {
      }
 
 document {
-     Key => {pairs,(pairs, HashTable),},
+     Key => {pairs,(pairs, HashTable)},
      Headline => "list the pairs in a hash table",
      TT "pairs x", " -- makes a list of all key/value pairs ", TT "(k,v)", " in
      a hash table ", TT "x", ".",
@@ -133,7 +133,7 @@ document {
      }
 
 document {
-     Key => {(pairs, Sequence),(pairs, List),},
+     Key => {(pairs, Sequence),(pairs, List)},
      Headline => "list the pairs in a sequence or list",
      Usage => "pairs L",
      Inputs => { "L" => Sequence },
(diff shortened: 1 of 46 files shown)


zteitler:
Good catches. Thanks Dan. They are all addressed now.

M2/Macaulay2/packages/Macaulay2Doc/doc4.m2
```diff
@@ -133,6 +133,19 @@ document {
      }
 
 document {
+     Key => {(pairs, Sequence),(pairs, List)},
```
DanGrayson: The key should be simply (pairs,BasicList), because "List" in our "d"
code corresponds to "BasicList" at top level, and List and Sequence
or both types of BasicList.

M2/Macaulay2/packages/Macaulay2Doc/doc5.m2
```diff
@@ -616,11 +620,17 @@ document {
      Key => 1 : setRandomSeed,
      Usage => "setRandomSeed()",
      Consequences => {
-	  {"Initializes the random number generator to a fixed state, identical to the
-	       initial state in version 1.2 and earlier of Macaulay2.  After version 1.2,
-	       the random number seed is initially set to a number that depends on the current date, 
-	       time (in seconds), and process id, except for the running of examples and tests
-	       in packages, where it is always initialized to 0."}
+	  {"Initializes the random number generator to a certain fixed state
+	       determined at the beginning of the current Macaulay2 session. In
```
DanGrayson: This gives the wrong impression about what "setRandomSeed()" does.  It is provided for backward
compatibility to set the seed as it was initially set in versions 1.2 and earlier.  It is not what is used
to determine the seed at the beginning of a Macaulay2 session.

DanGrayson:
I've merged in the changes to MultiplierIdeals.  One suggestion is to make separate branches
on your fork for separate changes and pull requests.  I hope the changes to MultiplierIdeals
were intentional and ready.

PR #215, opened 2015-04-02
Revised BoijSoederberg.m2

b-stone:
In this revision of the BoijSoederberg.m2 package we have added more options to the decomposition of a Betti diagram. In particular, the user now has the option to view a decomposition in which the Betti diagrams have entries from the Herzog-Kuhl equations that appear in the original Boij-Soederberg paper, entries corresponding to the graded Betti numbers of realization modules as constructed by Eisenbud-Schreyer, entries that are the least positive integral values as in the original package.  Further, we have updated the method names to conform to M2 style guides. Other additional features can be found in the documentation.


DanGrayson:
Could you remove "Revised by" from the new author names?  You could clarify
the situation in a comment.

b-stone:
I have moved the "Revised by" to a comment in the package. How does this look?

DanGrayson:
Looks good.

PR #217, opened 2015-04-04
New package BinomialEdgeIdeals

windisch:
Hi, this package provides methods for computations with (parity) binomial edge ideals and their permanental versions over arbitrary fields.

### DIFF
--- a/M2/Macaulay2/packages/=distributed-packages
+++ b/M2/Macaulay2/packages/=distributed-packages
@@ -108,3 +108,4 @@ MinimalPrimes
 Bertini
 CodepthThree
 Permanents
+BinomialEdgeIdeals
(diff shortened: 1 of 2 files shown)


DanGrayson:
I tried installing it with "installPackage", and the examples don't run, because Graphs isn't loaded.  That should be fixed, either by loading Graphs in the example code, or by moving Graphs to the PackageExports option.  Then I'll take another look.

windisch:
Thanks. This should be fixed now.

DanGrayson:
I just noticed that you export the variable "s".  That shouldn't be done, because the single letter variables belong to the user as variables for rings.

windisch:
Thanks. This is fixed in #280.

PR #233, opened 2015-05-05
New package: FiniteFittingIdeals

GStahl:
This package gives a fast algorithm for computing Fitting ideals of modules with constant Hilbert polynomial, together with methods for applying this to computations of Quot schemes.


DanGrayson:
Thank you for the submission!  It will appear in the 1.8 binary release.

PR #235, opened 2015-05-10
Add new package

eisenbud:
Finished for now editing CompleteIntersectionResolutions. If Mike puts in my code for creating elements of Hom from homomorphisms, then that should be subtracted.

### DIFF
--- a/M2/BUILD/david/Makefile
+++ b/M2/BUILD/david/Makefile
@@ -1,6 +1,7 @@
 all:
 	$(MAKE) -C build.tmp all install
 configure:
+	cd ../..; make -f Makefile
 	mkdir -p build.tmp 
 	cd build.tmp; ../../../configure --enable-download --prefix=$(HOME)/Macaulay2-built CPPFLAGS="-I/usr/local/include -I/usr/local/opt/readline/include" LDFLAGS="-L/usr/local/lib -L/usr/local/opt/readline/lib" CC="gcc-4.9 -m64" CXX="g++-4.9 -m64" --build=x86_64-apple-darwin
 
--- a/M2/Macaulay2/packages/CompleteIntersectionResolutions.m2
+++ b/M2/Macaulay2/packages/CompleteIntersectionResolutions.m2
@@ -1,13 +1,13 @@
 newPackage(
               "CompleteIntersectionResolutions",
-              Version => "0.8", 
-              Date => "April 2015",
+              Version => "0.9", 
+              Date => "May 10, 2015",
               Authors => {{Name => "David Eisenbud", 
                         Email => "de@msri.org", 
                         HomePage => "http://www.msri.org/~de"}},
               Headline => "Analyzing Resolutions over a Complete Intersection",
 	      PackageExports => {"BGG"},
-              DebuggingMode => true --should be false when submitted
+              DebuggingMode => false --should be false when submitted
               )
 	  export{
          --some utilities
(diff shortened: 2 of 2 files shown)


DanGrayson:
David, could you give me a sentence or two for the "changes, 1.8"
documentation node, if the
changes are substantial enough to advertise?

DanGrayson:
I had to disable one of your tests, because it didn't work.  (I also had to fix 4 others so they would work.)  Take a look at the one I disabled and see how it can be made to work.

eisenbud:
Humph!  — your program! They *all* worked on my machine when I sent it off!

Anyway, thanks for fixing those, I’ll take a look at the last.

David
--
David Eisenbud
Professor of Mathematics, University of California, Berkeley
www.msri.org/~de

DanGrayson:
It's impossible that they worked on your machine.  I bet you ran them  manually, line by line, rather than running them with ```  check "CompleteIntersectionResolutions" ```

DanGrayson:
... and the only reason they worked is because of your init file, which probably defines ```kk```

eisenbud:
I ran them with “check” — and it took me quite a few fixes. But you’re right about “kk”.

David
--
David Eisenbud
Professor of Mathematics, University of California, Berkeley
www.msri.org/~de

DanGrayson:
Strange - if you run with "check", then the test M2 session would not load your init file, since it's run with the "-q" option, unless you specify ```UserMode=>true```.

eisenbud:
Dear Dan,

That is odd. The other test that failed was also because of a special definition of Matrix^Array in my init file — so somehow that “check" mechanism doesn’t seem to be working.

Anyhow, I fixed it, ran “check”, all seems to work fine.

I **think** I successfully made a “pull” request. 

I must say I find the git lingo, such as “checkout” for changing from one branch to another, unintuitive, though I think I’m beginning to use it more correctly.

David
--
David Eisenbud
Professor of Mathematics, University of California, Berkeley
www.msri.org/~de

eisenbud:
Possible text for the “Changes” file:

The package “CompleteIntersectionResolutions” implements a number of old and new ideas about minimal resolutions over a complete intersection developed by Eisenbud-Peeva, Avramov-Jorgensen, Eisenbud-Peeva-Schreyer, and Burke-Eisenbud-Schreyer. Let S = k[x_1..x_n] be a polynomial ring, ff a codimension c regular sequence of homogeneous forms of the same degree, and R = S/(ff). It contains:

1) routines to compute the structure of “higher matrix factorization” on a “high” R-syzygy M — one for which the modules Ext_R^even(M,k) and Ext_R^odd(M,k) have negative regularity over the ring of CI operators. There are also routines to extract various information from the higher matrix factorization.

2) routines that implement the reconstruction algorithm of Avramov and Jorgensen that constructs modules M having (certain kinds of) specified Ext-modules.

3) routines to test of a conjecture of Eisenbud about the vanishing of certain local cohomology of Ext-modules, implementing the map from a module to its saturation.

4) routines to compute the higher homotopies for ff on an S-free resolution of an S-module M annihilated by ff, and understanding the structure of module over an exterior algebra, determined by the ff-homotopies on a resolution of M, on Tor^S(M,N) and Ext_S(M,N), when M and N are S-modules annihilated by ff. These routines led to conjectures that were later proven, and will appear in a work-in-progress of Eisenbud, Peeva and Schreyer.

5) routines to compute Hom in the stable category of Cohen-Macaulay R-modules, and test for stable triviality. This is used in understanding possible obstructions to commutativity of CI-operators. 


David Eisenbud
Professor of Mathematics, University of California, Berkeley
www.msri.org/~de

DanGrayson:
Thanks, I've used it.

DanGrayson:
I take it back : the ```-q``` option is used during ```check``` only if it was used on the command line that you used to start M2.  Hmm....

PR #236, opened 2015-05-10
fixed the disabled test 

eisenbud:
(problem came from using a routine defined in my init file, Matrix^Array.)


DanGrayson:
Thanks!

PR #259, opened 2015-05-28
New routine in Graphs.m2

windisch:
Hi, this routine computes the expansion of a graph. However, the implemented algorithm is probably not the most efficient. It would be great if this could appear in version 1.8. Thanks!


DanGrayson:
Thanks.  How about adding a test or two, to make sure it works in the future?

windisch:
Done. I just have noticed that these tests were the first tests at all in this package. I will tackle this in the near future.

PR #267, opened 2015-05-30
LieTypes and ConformalBlocks packages

dswinarski:
Hi Mike and Dan,

Here are my LieTypes and ConformalBlocks packages.  These are the most recently revised versions that I sent to JSAG.  They work in Macaulay2-1.7.


DanGrayson:
Notice what I changed in commit 5dce4af14d3939da80696586e76494142e88f975.

Thanks!

Issue #185, opened 2014-10-29
make ceiling and floor methods


kschwede:
Currently, the ceiling and floor functions apply to numbers.  However, there are some instances where it would be natural (from a language point of view) to make them into methods that can be overloaded to allow other inputs (for instance divisors).

Dan suggested I create this as an issue and possibly assign it to myself.

d-torrance:
This was fixed way back in #276

PR #276, opened 2015-06-01
Make ceiling and floor methods #185

zteitler:
Making ```ceiling``` a method was easy as it was already defined at top level anyway. Making ```floor``` a method involved "hiding" the lower-level ```floor``` by renaming it ```floor0```, and defining a "new" top-level ```floor```. Addresses Karl's request in issue #185.

Is this a reasonable approach for replacing a CompiledFunction with a MethodFunction? There are about 200 instances of CompiledFunction and 175 of FunctionClosure. I assume that only a fraction of them are of interest for "upgrading" to MethodFunction. Even though this procedure is not difficult I wouldn't want to go through 375 of them. But a small number, say the very most interesting ones, should be pretty easy I think.


M2/Macaulay2/d/actors3.d
```diff
@@ -1023,23 +1023,23 @@ agm(e:Expr):Expr := (
...
+	  toExpr(floor0(x.v.re))
 	  )
-     is x:QQcell do toExpr(floor(x.v))				    -- # typical value: floor, QQ, ZZ
-     is ZZcell do e				    -- # typical value: floor, ZZ, ZZ
+     is x:QQcell do toExpr(floor0(x.v))
+     is ZZcell do e
      else buildErrorPacket("expected an integral, rational, or real number")
      );
-setupfun("floor",floor);
+setupfun("floor0",floor0);
```
DanGrayson: I think this should be the single line of code in *.d files that changes, and the only change should be
to the first floor on the line.  That will minimize the number of changes, and it's okay, because the the identifiers
in the D language are not visible to the interpreter.

M2/Macaulay2/packages/Macaulay2Doc/functions/floor-doc.m2
```diff
@@ -2,14 +2,12 @@
 --- author(s): MES
 --- notes: just added example
 
-undocumented{(floor, ZZ)}
-
 document { 
-     Key => {floor,(floor,RR),(floor, QQ),(floor, CC)},
+     Key => {floor,(floor,Number)},
```
DanGrayson: Good change.  I suggest having separate nodes for the simple keys "floor" and "ceiling".  Here is the way
the node for "prune" looks:

```
document { 
     Key => prune,
     Headline => "prune, e.g., compute a minimal presentation",
     SeeAlso => {minimalPresentation}
     }
```

The idea is that if other methods are going to be added, this simple node can be an umbrella node for
all of them.

DanGrayson:
This is a good method for making a compiled function into a method function.  Some compiled functions are implicitly method functions, e.g., "plus", so this work would not be needed for them.  What they do is to check for internal cases quickly (numbers, strings, etc.), and if the types aren't appropriate, then do a lookup for a method, as on this line: https://github.com/Macaulay2/M2/blob/8075fcf2544f1942c6cf7d75844c43f7536ceef7/M2/Macaulay2/d/actors.d#L80

zteitler:
Oh, that is much better than what I did. How about if I change

https://github.com/Macaulay2/M2/blob/8075fcf2544f1942c6cf7d75844c43f7536ceef7/M2/Macaulay2/d/actors3.d#L1033

to

```else unarymethod(e,floor)```

I will have a chance in a few hours to try it. If it works I'll update the pull request tonight or tomorrow.

DanGrayson:
No, I like your original approach better, because it's more transparent to the user.

M2/Macaulay2/d/actors3.d
zteitler: To be precise, d/basictests/L02.m2 also has changes. This is the only line change outside of d/basictests/.

DanGrayson:
See also commit 16dcbd7db4e302a627565e3f8b0161ed42d2d6f8, for a needed change to the list of typical values.

PR #277, opened 2015-06-01
Bracket power of an ideal.

moorewf:
This pull request includes an implementation of the bracket power of an ideal.

Either a singleton [k] or a tuple [k_1,\dots,k_n] is allowed, where n is the number of variables.  See the documentation node for an example.


DanGrayson:
Shouldn't operations on ideals be well defined, in the sense that two
ideals that are equal
but have different generators yield the same result?

moorewf:
The way I have implemented it, the bracket power of an ideal I is the image of I under the ring map that raises the generators of the ring to the given exponents, rather than the one obtained by raising the generators of the ideal to the prescribed powers.  

In the case that the ring is characteristic p and the exponent is a singleton power of p, or in the case of a monomial ideal with a singleton power, these are the same of course.

DanGrayson:
Thanks, I should have read it more carefully.  Of course, if the ring is a
quotient ring, then the
ring homomorphism may not be well defined -- perhaps add a remark about
that.

It seems to me that if gb I has already been computed, then gb I^[n] can be
derived from it.
It might be nice (later) to implement that, too.  (It occurs to me also
that we might want
a way of recording the possibility of doing that, so it gets done quickly
later on if it turns out
to be needed, rather than doing it prospectively.)

moorewf:
I mentioned that at the end of the Ideal ^ Array documentation, but I wasn't sure whether to put it in as a caveat or in the 'body' of the doc node.

DanGrayson:
I see, thank you.

PR #280, opened 2015-06-03
Method "s" in BinomialEdgeIdeals

windisch:
Hi, this corrects the name of the method "s" so that it does not interfere with ring variables any more (see #217).

### DIFF
--- a/M2/Macaulay2/packages/BinomialEdgeIdeals.m2
+++ b/M2/Macaulay2/packages/BinomialEdgeIdeals.m2
@@ -14,25 +14,24 @@ newPackage("BinomialEdgeIdeals",
 export {
 
    --methods
-   "binomialEdgeIdeal",
-   "parityBinomialEdgeIdeal",
-   "disconnectors",
-   "isDisconnector",
-   "isEffective",
-   "s",
+   binomialEdgeIdeal,
+   parityBinomialEdgeIdeal,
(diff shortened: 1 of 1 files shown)


mikestillman:
Hi, I see that the function "sFunc" is exported, but doesn't have documentation, could you please add a doc node?

Also, what is it doing?  (I looked at the source code, but it wasn't immediate to me).  Could it be renamed to indicate what it is actually computing?

-- mike

DanGrayson:
Perhaps we can find a better name than sFunc...

DanGrayson:
We have ZZ, QQ, RR, HH, etc., so ss is a possibility.  But still, something descriptive would be good.

windisch:
The name `s` was taken from our paper parity binomial edge ideals, but I agree, that is not really descriptive. I renamed it to `weightedConnectedComponents`, which is quite lengthy but I hope descriptive enough. I also added a few lines of documentation and additional tests. I hope everything is alright now.

windisch:
Since this function computes an invariant of a graph and has (a priori) nothing to do with binomial edge ideals, maybe the `Graphs` package would be a better place to put it. What do you think?

DanGrayson:
That's a good name.

We're getting close to 1.8, but you can always move it to another package later on.  The authors of ```Graphs``` might have an opinion about the move.

Issue #319, opened 2015-07-21
Arithmetic with infinite numbers and rationals/reals


dibernst:
Everything is fine when I do something like

`infinity - 1`

but when I try 

`infinity - 1/2`

it tells me that there is no method for applying the operator "-" to an infinite number and a rational.

zteitler:
Why not RR and Constant too. And CC for + and -, maybe.

zteitler:
Close?

PR #322, opened 2015-07-23
add methods for arithmetic operations with infinite numbers #319

zteitler:
issue #319


DanGrayson:
More care is needed, because elements of RR and CC can be infinite or indeterminate.  Try something like 1./0. - infinity, which should result in something indeterminate, for example.

DanGrayson:
The function ```isFinite``` can be used to determine whether a real or complex number
is finite.  Real numbers can be infinite in the positive direction or infinite in the negative
direction, but complex infinite numbers come in only one flavor.

zteitler:
Does this seem okay now?

DanGrayson:
x//y is used for division with remainder, so why add a method for that?

```
InfiniteNumber // InfiniteNumber := (x,y) -> indeterminate
```

DanGrayson:
I wonder about your code for infinity^infinity, in light of this bug:
```
i21 : ii/0. > 0

o21 = true
```
Perhaps you could fix that, too?

zteitler:
What should "ii/0. > 0" output?

DanGrayson:
The operator ">" is defined in terms of the operator "?", so a good fix would be to make

    i8 : ii/0. ? 0

    o8 = >

return the symbol ```incomparable```.  That's mentioned in the documentation of "?".

zteitler:
Note that ii/0. is a complex number (ie a CC rather than an InfiniteNumber). It seems to me that the comparison is carried out in actors3.d. That code, at line 310 of actors3.d, uses "flagged()" to detect incomparability. I can't tell why it's allowing this comparison to go through. At the moment it looks like fixing this will require changing actors3.d's compare (lines 212-369) to explicitly check "isfinite" in a number of places.

Let me confirm that you want infinity as a complex number to be incomparable to any other number (except maybe itself?), while keeping finite complex numbers comparable. What about "ii/0. ? 1/0." (I am guessing it should be incomparable)?

DanGrayson:
Yes, we think of there being just one infinite complex number, of
indeterminate
argument.  So it's neither positive nor negative.

Perhaps flagged() is not being triggered, and the problem is in this
routine from
gmp.d:

export compare(x:CC,y:ZZ):int := (
     r := compare(x.re,y);
     if flagged() || r != 0 then r
     else compare0(x.im,0));

If the real part of the complex number is +infinity, this routine will
indicate that the
complex number is positive.

If that doesn't work out, then
to determine whether flagged() is being triggered, use gdb, the debugger.
There
are configure options for compiling in debug mode.

zteitler:
Let me try to correct my earlier statement. 'flagged()' does not detect incomparability; it detects if a comparison is being made with one or more nans. In particular 'flagged()' does not care about the presence of infinity.

We want infinity = 1/0. (real number) to remain comparable to real numbers. So, no change to compare0 or compare(x:RR,y:RR). The only time infinity is not comparable is when it comes from a complex number. One place to detect this is in gmp.d's compare(x:CC,y:CC), compare(x:CC,y:RR), etc. (every "compare" with a CC). Another candidate is actors3.d's compare. I don't know which would be better; I have tried with gmp.d but if it raises problems this can be changed.

You may be aware that Macaulay2 allows such nonsense as 1/0. + 5*ii, 1/0.-1/0. + ii/0., and so on --- the real part and imaginary part can each be infinity, negative infinity, nan, or finite numbers, separately from each other. So I am changing gmp.d's "toCC" to check finiteness and nan-ness (? what is the right word) in the following way. In toCC(x:RR,y:RR) if either x or y is nan, then return the complex number (nan,nan); if either x or y is infinite (positive or negative infinity) then return the complex number (infinity,infinity). (In nan versus infinity, I think nan should win, but it can be changed if desired.)

Next: currently (in trunk) (1/0.) ? (1/0.-1/0.) is "incomparable" (good) but infinity ? (1/0.-1/0.) is ">". Ugh. I just noticed this; I think I will be able to fix it (make it "incomparable").

Finally I am going to write a block of assertions to be added to packages/Macaulay2Doc/tests.

If it all works expect a push tonight or tomorrow.

DanGrayson:
That sounds good!

zteitler:
It should be okay now, or at least ready for review.

zteitler:
What would be a reasonable response to `infinity ? 1/0.` ? Should it be `=`? Right now (in trunk) it is `>`, i.e., `infinity > 1/0.` is true. Mathematically I would think it should be `=` but I don't know if there is some subtlety about floating point real numbers versus mathematical real numbers, perhaps this comparison is desirable?

DanGrayson:
I think equality would be desirable here.  Certainly to return > is not
right.

M2/Macaulay2/d/gmp.d
```diff
@@ -1185,30 +1188,58 @@ export (x:CC) === (y:QQ) : bool := x.re === y && x.im === 0;
...
+     if ( isinf(x.re) || isinf(x.im) ) then (
+       setflag0();
+       return 0;
+     );
      r := compare(x.re,y);
      if flagged() || r != 0 then r
      else compare0(x.im,0));
 export compare(x:QQ,y:CC):int := (
+     if ( isinf(y.re) || isinf(y.im) ) then (
+       setflag0();
```
DanGrayson: Why do you set the erange flag here, Zach?  Is there some other place in the code which checks whether the
flag has been set?

The erange flag is something internal to mpfr, so setting it at all seems unnatural.

zteitler: Yes, there is another place in the code that checks whether the flag has been set. Specifically, in actors3.d, the definition of `compare` on line 212 has "`if flagged() then incomparableE else`" in several branches.

I'm afraid I don't have a very clear memory of the specific purpose of setting this flag. I think the idea was to catch things like
``1. + 2.*ii ? 2. + (1/0.)*ii``
which should return `incomparable` (the second complex number is complex infinity) but without the flag, would just return `<` based on the comparison of real parts. To be pedantic there are two issues. One is that it needs to check if any imaginary parts are infinite before blindly comparing real parts. I guess it is a little redundant to also check if the real parts are infinite, as any infinities in the real parts would be caught and flagged when the real parts are compared. Ditto, some of the checking in other cases (e.g., line 1198, `compare(x:CC,y:RR)`, it would be sufficient to check if `x.im` is infinite); the extra checks could be stripped out but I think they do no harm. The second issue is that this level of compare can't pass `incomparable` up to the next level except in a GMP flag, as far as I can see. If there is another way, it could be used instead.

DanGrayson:
Thanks!

Issue #247, opened 2015-05-21
separate run script in environment-setting and M2-binary call:


jakobkroeker:
prurpose: set environment without starting M2 (e.g. to check linked libraries with 'ldd')

possible solution:
file 'M2':
```
exec /run/media/jakob/TOEXT4/work/home_fc18/Projects/M2-master/M2/BUILD/dan/builds.tmp/fedora32/usr-dist/i686-Linux-Fedora-18/bin/M2.run "$@"
```
file 'M2.sh'
```
cd /run/media/jakob/TOEXT4/work/home_fc18/Projects/M2-master/M2/BUILD/dan/builds.tmp/fedora32/usr-dist/i686-Linux-Fedora-18/bin/ 
. M2.sh `pwd`
exec bash
```
file 'usr-dist/i686-Linux-Fedora-18/bin/M2.run':
```
#! /bin/sh
# We assume that @ pre_bindir @/.. is ${pre_exec_prefix} (see configure.ac)
. `dirname "$0"`/M2.sh `dirname "$0"`
exec `dirname "$0"`/M2-binary "$@"
# Local Variables:
# compile-command: "make -C $M2BUILDDIR/Macaulay2/bin M2 "
# End:
```
file  'usr-dist/i686-Linux-Fedora-18/bin/M2.sh':
```
#! /bin/sh
# We assume that @ pre_bindir @/.. is ${pre_exec_prefix} (see configure.ac)
case "linux-gnu" in
   cygwin) export PATH=$1/../lib/Macaulay2/i686-Linux-Fedora-18/lib:$PATH ;;
  darwin*) # Mac OS X:
           export DYLD_LIBRARY_PATH=$1/../lib/Macaulay2/i686-Linux-Fedora-18/lib:$DYLD_LIBRARY_PATH ;;
        *) export   LD_LIBRARY_PATH=$1/../lib/Macaulay2/i686-Linux-Fedora-18/lib:$LD_LIBRARY_PATH ;;
esac
```

DanGrayson:
Do you have a proposal?  I'm not sure what it is.

jakobkroeker:
I would like to have a possibility to set the environment variables (LD_LIBRARY_PATH) without doing this by hand. Therefore I separated the M2 start script into parts, as above.

DanGrayson:
[...]

Issue #234, opened 2015-05-06
M2 --script


DanGrayson:
The "--script" command line option no longer works with M2 scripts.

DanGrayson:
The documentation for "execve" says that script interpreter specified on the first line
of the script cannot itself be a script.  So that's the problem, and it seems there is no
fix.

DanGrayson:
One fix would be to return to using a binary file for M2.  This would work provided,
during an installation script, we edit the "rpath" so the dynamic libraries can be found.
Under Mac OS X that can be done with ```install_name_tool ```.  Under linux that
can be done with ```chrpath``` or ```patchelf```.  The ubuntu package containing ```chrpath```
has the name ```chrpath```.

See also the comments at http://stackoverflow.com/questions/13769141/can-i-change-rpath-in-an-already-compiled-binary .

d-torrance:
> **Written by Claude** (Claude Opus 5, via Claude Code), not by @d-torrance, whose account posted it -- please weigh it accordingly.

Retested on both platforms in August 2026. The diagnosis in this thread still holds on
macOS, no longer holds on Linux, the macOS failure is worse than recorded here — and there
is a workaround that works on both today. Options for a real fix at the end.

### The platform split

`M2` is still a `/bin/sh` wrapper on both platforms — `/usr/bin/M2` on Debian,
`/opt/homebrew/bin/M2` on macOS — so everything turns on whether the kernel will resolve an
interpreter that is itself a script.

**macOS 26 (Tahoe), Apple M5 Pro, Homebrew.** It will not:

```
$ python3 -c 'import os,sys; os.execv(sys.argv[1],[sys.argv[1]])' hello.m2
OSError: [Errno 8] Exec format error
```

**Linux 6.8.** It will, and it is the kernel rather than a shell fallback — verified through
`os.execv` with no shell in the picture:

```
$ python3 -c 'import os,sys; os.execv(sys.argv[1],[sys.argv[1]])' hello.m2
OK -- shebang ran
```

Linux permits a bounded chain of script interpreters; macOS returns `ENOEXEC` and stops.
Neither is wrong — POSIX does not require the recursion, so Linux's behaviour is the
extension.

### The part not recorded here: on macOS it does not fail cleanly

Run from a shell rather than through `execve`, the shell catches the `ENOEXEC` and falls
back to interpreting the file **as a shell script**:

```
hello.m2: line 2: stderr: command not found
exit status: 127
```

[...]

PR #324, opened 2015-07-24
added M2_setenv which starts a subshell with Macaulay2 environment

jakobkroeker:


DanGrayson:
What's the purpose of this change?

jakobkroeker:
this would also be a workaround for 
https://github.com/Macaulay2/M2/issues/234

DanGrayson:
See my comment on a better way at #247

DanGrayson:
In what sense is this a workaround for #234 ?

DanGrayson:
Discussion seems to have died down.  Closing, but feel free to reopen to revive the discussion.

PR #335, opened 2015-08-28
Document a caveat of isHomogeneous

tom111:
isHomogeneous only looks at the given generators and relations of a
module.  Therefore it can return the value false even for homogeneous
modules.  Extend the documentation to highlight this.

Maybe we have discussed this in person at some point? I somehow remember that the behaviour should not be changed for performance reasons.  In any case, I was recently bitten by it, so it should be documented properly.

### DIFF
--- a/M2/Macaulay2/packages/Macaulay2Doc/functions/isHomogeneous-doc.m2
+++ b/M2/Macaulay2/packages/Macaulay2Doc/functions/isHomogeneous-doc.m2
@@ -73,6 +73,14 @@ document {
 	  "isHomogeneous(a*M)",
      	  "isHomogeneous((a+1)*M)"
 	  },
+     PARA{},
+    "Note that no implicit simplification is done.",
+     EXAMPLE {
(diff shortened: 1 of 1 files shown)


DanGrayson:
Looks good to me.

It probably never happens that a user needs help converting inhomogeneous generators to
homogeneous ones, so avoiding computation is probably wise, even though the value
returned depends on the chosen generators and not on just the ideal.

tom111:
A few more thoughts:

- Hopefully all the existing code using this was written with the convention in mind.
- The function "isBinomial" in the Binomials package computes the reduced Gröbner basis before answering.
- To be completely explicit one could have two functions "isHomogeneous" and "isPresentedHomogeneously" (or some other less awkward name).

DanGrayson:
No need for something like "isPresentedHomogeneously" when "isHomogeneous gens" does the same thing.  Unfortunately, for modules, which are always submodules, there are two matrices that have to be examined, and "presentation" doesn't return the pair of them.

PR #337, opened 2015-09-01
Use pkg-config to detect factory.

d-torrance:
This is motivated by the Debian singular package's choice to rename factory
to singular-factory and put the headers inside a singular/ directory, as
pkg-config gives us the necessary CPPFLAGS to find these headers.

It also has the advantage of simplifying the check for version >= 4.0.0.

### DIFF
--- a/M2/configure.ac
+++ b/M2/configure.ac
@@ -1271,20 +1271,32 @@ AC_DEFINE_UNQUOTED(LAPACK,$val,whether to link with lapack)
 
 BUILD_gftables=yes		dnl it seems that factory now includes gftables, and installs it
 
+AC_MSG_CHECKING([whether factory library is installed])
 SINGULARLIBS="-lfactory  "
+FACTORY_MINVERSION=4.0.0
 if test $BUILD_factory = no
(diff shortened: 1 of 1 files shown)


DanGrayson:
Thank you.

Please add "aclocal" to the Makefile where "autoconf" is run, to get that macro defined.

The test for factory should not be made silent.  Announce the test and the result as was done before, and try to distinguish in the error messages between absence of factory and factory being installed but too old, stating the version number required.

d-torrance:
Thanks for the comments!  I believe this latest commit addresses them.  I ended up dropping the use of the PKG_CHECK_MODULES macro, so we shouldn't need aclocal.

DanGrayson:
Thanks, that looks cleaner.  But BUILD_factory might be "yes", due to configure command line
options, so in that case, do not check for the presence of factory.

d-torrance:
Thanks again for the review!

The check already didn't occur if BUILD_factory was "yes".  However, "checking whether factory installed" was still being printed, so I moved it inside the if statement.

PR #341, opened 2015-09-08
add routines to Graphs.m2

windisch:
Hi, this adds two routines to Graphs.m2 (including a bit of documentation). I just realized that all examples in Graphs.m2 run on error (is this due to the new version of M2?). Does anybody know how to fix that?


DanGrayson:
The examples and tests work for me.

DanGrayson:
I've asked the authors for their opinion.

PR #343, opened 2015-09-23
fix makefiles, now that memtailor uses libtool again

d-torrance:


DanGrayson:
I've reverted this pull request, since M2's mathicgb package couldn't find memtailor now, under
Ubuntu 14.04.

d-torrance:
I see -- when I tested this pull request, I imagine that I had the Debian memtailor package installed and so mathicgb used that.  I'll try troubleshooting it in a clean environment and see if I can fix things.

Thanks!

d-torrance:
I had some time to test this.

I set up a clean 14.04 chroot with only the minimum packages needed installed and tried installing from git (with c39f2d16ca5fc54035a3001f2022e15892100d91 checked out).

Initially, I did have a build error, which looks like what you described:

    /usr/bin/ld: cannot find -lmemtailor
    make[1]: *** [mgb] Error 1
    make[1]: Leaving directory `/root/src/macaulay2/M2/M2/submodules/mathicgb'
    make: *** [build-in-mathicgb] Error 2
    make: Leaving directory `/root/src/macaulay2/M2/M2'

However, a git log in the memtailor directory showed that it was pointing to an earlier commit (7b48b98178b247c948621d7509111aadd8d48dd5)

When checked out a8212c679c59172ea552d18c0a1f1e1a73c0b2c8, it compiled without problems.

DanGrayson:
Oh, right!  So, the fix would be for the M2 pull request to include the update to the new release
of memtailor.  You would see that as a change to M2/submodules/memtailor, and such a change can be
committed and made part of the pull request.

I should have noticed.

d-torrance:
No worries!

What would be the easiest way to proceed?  Should I submit brand new versions of both pull requests?  Or would you rather de-revert them and I submit a pull request with just the change to M2/submodules/memtailor?

DanGrayson:
I prefer you to submit a pair of new pull requests, with the one change to the M2 one.

d-torrance:
Sounds good.  Just submitted the memtailor pull request (https://github.com/Macaulay2/memtailor/pull/3).

PR #345, opened 2015-10-19
Divisor.m2 Pullback.m2 PushForward.m2

kschwede:
This has a new package "Pullback.m2".  Which computes a pullback in the category of rings.  It also has updates for Divisor.m2 and PushForward.m2  (Claudiu Raicu asked me to create the pull request for PushForward.m2)

### DIFF
--- a/M2/Macaulay2/packages/Divisor.m2
+++ b/M2/Macaulay2/packages/Divisor.m2
@@ -1,5 +1,5 @@
 newPackage( "Divisor",
-Version => "0.1o", Date => "June 15th, 2015", Authors => {
+Version => "0.1p", Date => "August 6th, 2015", Authors => {
      {Name => "Karl Schwede",
      Email=> "kschwede@gmail.com",
      HomePage=> "http://www.math.utah.edu/~schwede"
@@ -54,6 +54,7 @@ export{
 	"findElementOfDegree", 
 	"getLinearDiophantineSolution",		--has Unsafe option
 	"canonicalDivisor", --has IsGraded option
+	"ramificationDivisor",
     --tests and related constructions
     	"isWDiv",
 	"isEffective",
@@ -770,6 +771,7 @@ idealToDivisor = method();
 idealToDivisor( Ideal ) := (I1) ->
 (
 	if  (I1 == 0*I1) then (error "idealToDivisor: cannot form divisor from the zero ideal";);
+	--question, is it faster to reflexify the ideal or just check whether the ideal has too high of dimension... if multiprocess gets working, then we may want to check both...
 	I2 := reflexifyIdeal(I1);
 	L2 := {};
 	if ( isSubset(ideal sub(1, ring I2), I2) == false ) then (L2 = minimalPrimes(I2););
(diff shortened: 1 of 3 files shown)


DanGrayson:
Thanks!

I made a minor change to the header of the new package, so you should merge with our master branch now.

kschwede:
Done and thanks!

DanGrayson:
Karl, your changes to PushFoward.m2 broke examples in Schubert2 overnight, so in commit 6ff20cd3a5e0d0a0f4346d47f1f7cc06c836bcf3 I've rolled it back to what we had before.

Could you take a look and submit a new pull request that doesn't do that?

kschwede:
Hi Dan,

Thanks, I see that.  Here's my guess as to what is going on as this is the only change Claudiu made that I am aware of that could cause this.

So back in May Claudiu made the following change because we were getting unexpected behavior out of PushFwd.  

Ok, given a ring map f : R->S, and an S-module M, Claudiu's pushFwd actually returns three things.  

The third of which is "a map that assigns to each element of the target of f its representation as an element of the pushed forward module [S]".

In other words, if you apply (pushFwd(f))#2 to an element s of S, you should get a map

R -> f_* S

which sends 1 to s.  Instead what pushFwd was returning was a map

R -> F

where F was a free module which maps to the generators of f_* S.

My guess would be that Schubert2 was expecting the old behavior.  I have made some changes to a local copy of Schubert2, and I reinstalled that version and the checks passed even with the new PushFwd.  These details are below.   Dan, you are an author of Schubert2, so I'm happy to have a conversation with the author who maintains this part of Schubert2, as well as Claudiu, etc. to figure out what the right behavior of PushFwd should be.  

There are two places you call pushFwd and I modified them as follows

     (BasAModule, bas, iLowerMod2) := pushFwd iupper;     
     -- iLowerMod(element b of B) = one column matrix over A whose product with bas is b
     iLowerMod := zz -> matrix(iLowerMod2(zz));

     (BasAModule, Bbasis, fLowerMod2) := pushFwd f;
     fLowerMod := zz -> matrix(fLowerMod2(zz));

THis fixed the errors with the Schubert2 check for me.

DanGrayson:
If pushFwd changes, then we have to change the code that uses it, so let's
go ahead.

Could you submit an additional pull request that undoes the rollback, fixes
Schubert2,
and modifies the documentation of pushFwd to clarify?

What if we also take advantage of this opportunity to bring PushForward
into line with our
previously existing method "pushForward", as it is confusing to have that
as well as
"pushFwd"?  Is "pushFwd" good enough that it should just overwrite
"pushForward"?
Is it compatible?  If not, can it be made compatible?

Also, "pushForward" puts the ring first, and "pushFwd" puts the module
first.

kschwede:
Hi Dan,

Ok, let me see if I can figure out how to undo the rollback etc. all in a pull request.  :-)

In terms of the other question, for me pushFwd has worked without issue.  It is substantially faster than pushForward and works on more objects (ie, pushForward requires everything in sight to be graded).  I asked Claudiu to comment here though, as he probably knows more if there are any differences.

In terms of compatibility, one could add a command pushForward to Claudiu's package that works the same way as the classical pushForward does.

Best wishes, 

Karl

DanGrayson:
We could also consider making PushForward preloaded, and jettison the old
code for pushForward.

DanGrayson:
Re: "In terms of compatibility, one could add a command pushForward to
Claudiu's package that works the same way as the classical pushForward
does."

That sounds confusing.

kschwede:
Hi Dan,

I guess I wasn't clear.  I was basically just suggesting that pushForward just call pushFwd.

Claudiu is fixing one small bug I pointed out to him (pushing forward things from polynomial rings with empty sets of variables crashes right now), and I'll try to do a pull request for an updated version next week.

DanGrayson:
That still sounds confusing, because of the similarity of the three names.
What about having one
function called pushForward in the package that is upward compatible with
the existing one in the Core,
removing the old one in the Core, and arranging for PushForward to be
preloaded by adding it as the
second from the end in m2/loadsequence?

The existing function has 5 options that it passes onward to gb; something
should be done about that.

DanGrayson:
And there is the issue of the option "NoPrune", which takes true or false.  Better would
be an option "Prune", so users don't have to think about double negatives.  In fact, "Prune"
is already an option to Ext and Tor, but it's ignored!

mikestillman:
I see no problem with replacing the pushForward in m2/matrix3.m2 with this package (i.e. removing it, and making this package pre-loaded).  But "pushForward" does get used for kernel of a ring map, it seems.

DanGrayson:
I don't see it being used for kernels of ring maps ...

mikestillman:
Oops, my mistake.  "pushNonlinear" is being used for kernels of matrices, when the source and target rings are different.  This function is probably not being replaced.

DanGrayson:
Would letting "kernel for ring maps" call the new code in that case be an improvement?

PR #351, opened 2015-11-03
Updated PushForward.m2 with Claudiu's changes.  Note that the order o…

kschwede:
pushFwd(module, f)  has been switched to  (f, module), to match the previous pushForward function.

Updated Pullback.m2 to work with the revised order.

Updated Schubert.m2 to work with the revised pushFwd.


kschwede:
Dan, this has the changes we discussed.  I certainly don't mind if this pushFwd is changed to overwrite the old one.

mikestillman:
I'm going to merge this, then test it after that.

mikestillman:
Karl,

Does this pull request change the m2 pushFwd?  Or do we need to do that separately?

-- mike

kschwede:
It does not change the m2 pushForward.  We need to do that separately.

PR #356, opened 2015-11-19
New package ToricTopology

alvisetrevisan:
Would like to contribute a package for computations in Toric Topology. Co-authored with Alex Suciu at Northeastern.


DanGrayson:
Thank you for the package.  It would be nice if there were some tests so that when we
run ` check "ToricTopology" ` every night we can find out whether anything breaks.

Issue #354, opened 2015-11-17
debug build & flint


bapike:
I pulled a7cdee1fafb57cd3a62fcf94293a066d1971281a and tried to do a build on Debian 8 with these parameters:
```
./configure --enable-download --prefix=/home/bapike/git/M2-test --disable-ntl-wizard --enable-build-libraries="ntl pari" --enable-debug
```

It fails to build because DEBUG is defined via the compiler calls and `flint-2.4.5/fmpz_factor/factor_pp1.c` has:
```
#define DEBUG 0 /* turn on some trace information */
```
Changing this to
```
/* DEBUG can be defined on the command line... */
#ifdef DEBUG
#undef DEBUG
#endif

#define DEBUG 0 /* turn on some trace information */
```
allows the build to finish.  However, about 18 tests fail with the message:
```
M2-binary: ctx_init_conway.c:45: _fq_nmod_ctx_init_conway: Assertion `*cp == -1' failed.
```
Picking one of these, the message is reproducible by just running
```
GF 9
```
in a debug build of Macaulay2.  GDB says:
```
M2-binary: ctx_init_conway.c:45: _fq_nmod_ctx_init_conway: Assertion `*cp == -1' failed.

Program received signal SIGABRT, Aborted.
[Switching to Thread 0x7ffff2e66700 (LWP 17086)]
0x00007ffff40ea107 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
56      ../nptl/sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  0x00007ffff40ea107 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007ffff40eb4e8 in __GI_abort () at abort.c:89
#2  0x00007ffff40e3226 in __assert_fail_base (fmt=0x7ffff4219ce8 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0xb9f658 "*cp == -1", file=file@entry=0xb9f646 "ctx_init_conway.c", line=line@entry=45, 
    function=function@entry=0xb9f750 <__PRETTY_FUNCTION__.8643> "_fq_nmod_ctx_init_conway") at assert.c:92
[...]

Issue #35, opened 2013-05-30
"Duplicate large block deallocation" message when calling numRealTrace from realroots.m2


larstiq:
M2 aborts after outputting the message "Duplicate large block deallocation" when running the following code:

https://gist.github.com/larstiq/5676124  (it seems github does not support file attachements but wants one to post a "gist", hrmph).

It fails on 64bit Debian Wheezy and Ubuntu 12.04 computers.  The computation is
still running on my 32bit laptop but  it may just be slow.

The code depends on http://www.math.tamu.edu/~sottile/research/ps/realroots.m2 as featured in
the chapter "From Enumerative Geometry to Solving Systems of Polynomial Equations" of the 
"Computations in algebraic geometry with Macaulay2" book.

mahrud:
Closing this issue since the example seems to work now and there seem to be related patches merged in 2015. Feel free to reopen if it is still an issue.

PR #357, opened 2015-11-20
partially fix debug build

bapike:
These changes allow a debug build to compile, and fix the 19 errors in Macaulay2Doc examples that were reported in #354.  It successfully runs all examples in packages before MinimalPrimes (it has an issue there).

One of the 19 errors was "sqrt 2p3000000", as mentioned in #35, although the originally reported bug still does not work.

### DIFF
--- a/M2/Macaulay2/d/actors4.d
+++ b/M2/Macaulay2/d/actors4.d
@@ -489,7 +489,7 @@ lenseq(y:Sequence):int := (
 	  is y:SymbolClosure do l = l + length(y.symbol.word.name)
 	  is y:ZZcell do (
 	       if isInt(y) 
-	       then l = l + toInt(y)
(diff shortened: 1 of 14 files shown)


DanGrayson:
Thanks, I'll test it!

bapike:
For a debug build, `make` now works, and I'm testing `make check`.

There are two legitimate bugs here.  Commit 60210bc fixes a problem where, e.g., `concatenate(-2^20,2^20)` will try to overwrite a megabyte of memory with spaces.  Commit 052dda4 makes `substitute(1/101,ZZ/101)` give an error.  I've added tests for these.

The rest of the changes are pretty straightforward.  I got a little carried away and tried to fix a few undistributed packages.

bapike:
`make check` now works in a debug build without optimization, though I had to allow more time for some tests.

DanGrayson:
Thanks!

PR #358, opened 2015-11-30
Normaliz Macaulay2 interface update for 3.0

csoeger:
* nmzMinExecVersion increased to 2.11 because of the use of "inequalities"
* version check use option --version (still compatible with older versions)
* use string input types
* new file types added

### DIFF
--- a/M2/Macaulay2/packages/NormalToricVarieties.m2
+++ b/M2/Macaulay2/packages/NormalToricVarieties.m2
@@ -719,7 +719,8 @@ isAmple ToricDivisor := D -> (
 	if w_(0,0) < 0 then w = (-1)* w;
 	(v // w)_(0,0) > 0))))
 
-hilbertBasis Matrix := Matrix => C -> transpose (normaliz(transpose C,0))#"gen"
+hilbertBasis Matrix := Matrix => C -> (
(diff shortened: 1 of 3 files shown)


csoeger:
Please let me know if you have questions to this pull request or if I can be of any help.

DanGrayson:
Probably we have to update the Makefile for normaliz in the libraries directory to 
compile that version, too, otherwise we'll get errors when trying to run the examples.  Did 
you test it for that?

csoeger:
This is only a patch for the Normaliz.m2. It will work with the supplied version of Normaliz (2.12) and also newer versions (3.0 +). As stated the minimal requirement is Normaliz 2.11.

To update the used version of Normaliz is then a further step. I can also look into that, but I'm never able to find the right place in which it is controlled.

csoeger:
I think the update of Normaliz.m2 is uncritical.
If you want to use Normaliz3.1.0 in the suggested way is up to you. Changing the compile flags there is surely not optimal. You might consider configuring Normaliz via cmake. This is way more flexible and should work on most systems. But it adds the need to have cmake installed.

DanGrayson:
Thanks!

PR #362, opened 2016-01-06
Update RunExternalM2 package to v0.83

### LINKED ISSUE #355: make option to exclude packages
Minor feature request: I have been having trouble with RunExternalM2 (fails `make check` on my laptop). It would be nice to have an option like `EXCLUDEPACKAGES='Foo Bar'` to exclude Foo and Bar from the list of packages to build, check, and install. For example this might be possible via gmake's filter-out command <https://www.gnu.org/software/make/manual/html_node/Text-Functions.html>; something like `PACKAGES = $(filter-out $(EXCLUDEPACKAGES),$(PACKAGES))`. I tried to get this working but I could not figure out what file to put it in (configure.ac? Macaulay2/packages/Makefile.in?). A different or shorter name than EXCLUDEPACKAGES would be fine.

_edit: I shouldn't have said RunExternalM2 was 'broken'._

bapike:
Changes to RunExternalM2:
* remove the `runExternalM2InClone` function and its tests, to fix #355 
* improve some documentation
* improve an error message

I have also:
* changed M2's fork() documentation to be more discouraging of its use


DanGrayson:
Thanks!

PR #365, opened 2016-01-27
update from Anton

antonleykin:


mikestillman:
Dan, I'm going to merge this pull request, and then let the tests check tonight (possibly the Bertini tests will fail again).  This is a merge request in preparation for a branch that Anton and will base off of this.

DanGrayson:
The overnight build failed, so let's roll this pull request back.

mikestillman:
Anton, Did you forget to checkin a file?  Is this something we can fix immediately?

antonleykin:
Hmm... compiles for me. Does the overnight build recompile the d/ directory?

The problem seems to be coming from numericalAlgebraicGeometryTrace
introduced in this commit:
400107a6dd332d3d80c92c915c316a020d1f60e9

DanGrayson:
Yes, it compiles everything.

Anton, the error is with "make check" in the directory "e/unit-tests".  Try
that.

mikestillman:
I’m also recompiling, and I’ll run the tests in a bit.

antonleykin:
Yep, I can reproduce that.
I am not sure what the problem is... Could it be that the object file
defining "_M2_numericalAlgebraicGeometryTrace" is missing?

mikestillman:
It is probably quite simple: the tests do not link with the garbage collector, so we probably need to add an object file that got added.

Once my examples run, I’ll run the tests.

DanGrayson:
The file in question is d/M2.o

mikestillman:
OK, I need to look at it, but the fix should be the same as the way I had to deal with M2_gbTrace.

I’ll (hopefully) make the fix.

mikestillman:
OK, I added a line to e/unit-tests/M2-replacement.c, and the e/unit-tests directory now passes.

I’ll be testing the other tests this morning too.

mikestillman:
OK, all tests now pass (except that one for running externalM2, which always still fails on my mac).

PR #366, opened 2016-01-27
Some build system cleanup

doughdemon:


DanGrayson:
These are probably good changes.  I'll test them.

doughdemon:
thanks!

PR #372, opened 2016-03-11
Fix makefiles, now that memtailor uses libtool again (v2)

d-torrance:
This commit was previously applied, but then reverted because the
corresponding commit in the memtailor submodule was not updated
(see [1]).

Note that we have updated M2/submodules/memtailor to point to [2], one
commit earlier than the current HEAD of the master branch.  It may be
preferable to merge [3] first and use that commit instead.

[1] https://github.com/Macaulay2/M2/pull/343
[2] https://github.com/Macaulay2/memtailor/commit/a8212c6
[3] https://github.com/Macaulay2/memtailor/pull/3

### DIFF
--- a/M2/Macaulay2/bin/Makefile.in
+++ b/M2/Macaulay2/bin/Makefile.in
@@ -180,7 +180,7 @@ else
 M2_LIBRARIES := $(M2_LIBRARIES) -L../d -lM2inits2
 endif
 M2_LIBDEPS += ../d/libM2inits2.a				\
-	@abs_top_builddir@/submodules/memtailor/libmemtailor.a	\
+	@abs_top_builddir@/submodules/memtailor/.libs/libmemtailor.a	\
 	@abs_top_builddir@/submodules/mathic/libmathic.a	\
 	@abs_top_builddir@/submodules/mathicgb/.libs/libmathicgb.a
 ifeq (@CONSTRUCTOR_ORDER_OBJ@,LR)
--- a/M2/include/config.Makefile.in
+++ b/M2/include/config.Makefile.in
@@ -74,13 +74,13 @@ endif
 vpath %.h $(BUILTLIBPATH)/include
 LDFLAGS :=						\
 	-L$(BUILTLIBPATH)/lib				\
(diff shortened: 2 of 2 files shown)


DanGrayson:
Thanks!  I see now there is now way for you to submit the required pair of pull requests
simultaneously.  I'll pull the memtailor one and let you update the M2 one as you suggest.

DanGrayson:
Okay, I did that.  Please update the pointer from M2 to memtailor by committing it to the branch
for this pull request -- no need to withdraw this pull request or anything like that.

d-torrance:
Done -- thanks!

DanGrayson:
Okay, we'll see now if that survives the overnight build of M2.

DanGrayson:
It did not survive the overnight build.  Could you fix memtailor so that after
running configure in the build directory, the command "make distclean" will work?

Here is the error message:
```
configure: error: cannot run /bin/bash /mike-raid/home/dan/src/M2/trunk-git/M2/BUILD/dan/../../submodules/memtailor/build/autotools/config.sub
make[2]: *** [config.status] Error 127
make[2]: Leaving directory `/mike-raid/home/dan/src/M2/trunk-git/M2/BUILD/dan/builds.tmp/Ubuntu-14.04-x86_64.production/submodules/memtailor'
make[1]: *** [distclean-in-memtailor] Error 2
make[1]: Leaving directory `/mike-raid/home/dan/src/M2/trunk-git/M2/BUILD/dan/builds.tmp/Ubuntu-14.04-x86_64.production'
make: *** [distclean2] Error 2
```

d-torrance:
I'm not able to reproduce this.  In a fresh 14.04 chroot, I cloned the M2 repository and ran

    make
    cd BUILD/build
    ../../configure
    make distclean

with no problems.  Am I missing something?

DanGrayson:
I think I got past it -- the problem was that there are various files in the source directory of memtailor, such as Makefile.in, that should have been remade, but weren't.  So I added "git clean" in all the submodules as part of the overnight build script.  I've also added a dependency to GNUmakefile.in
that should help.

PR #376, opened 2016-03-14
Remove dependency on gftables; included in factory since singular 4-0-0.

### LINKED ISSUE #295: gftables
The gftables seem to be included in factory, so we shouldn't need to have a separate library for installing them.  The only problem is getting them installed in the right place at the right time.

```
$ find usr* -name 961
usr-dist/common/share/Macaulay2/Core/factory/gftables/961
usr-host/share/factory/gftables/961
```

### LINKED ISSUE #375: Is gftables necessary if we don't build factory?
Another question related to creating the Debian package:

Currently, the behavior of the M2 build system is to download gftables and install them in ${prefix}/share/Macaulay2/Core/factory/.  In Debian, however, downloading things during build is not allowed and using additional packages as build dependencies is preferred.

I'm trying to write a patch which will work for the Debian package but also keep the current behavior as default.

If I'm understanding the code correctly, it appears that M2 only does this so that the gftables are available to factory.  M2 doesn't use them directly.

However, if factory is already installed prior to building M2, then it seems as if gftables wouldn't be necessary at all, as factory should already know where they are.

I'm thinking that if BUILD_factory is no, then we should automatically set BUILD_gftables to no and skip the "setFactoryGFtableDirectory gfdir" line in factor.m2.

Am I understanding this correctly?  Thanks!

d-torrance:
This closes #295 and closes #375.


d-torrance:
Argh, don't merge this (yet).

It breaks building examples for the RandomObjects package...
    i9 : (random planeCurve)(2,R,Certify=>true)
    can not open GF(q) addition table: gftables/81
    error: can not open GF(q) table
    gfops.cc:132: In function `void gf_get_table(int, int)':
    failed assertion `inputfile'
    Aborted (core dumped)

d-torrance:
Going to have to think about this some more.  Sorry for the noise.  Closing.

DanGrayson:
M2 likes to have all its files located relative to the location of the M2 binary itself.  To that
end, when we install gftables, we put the files in

  `usr-dist/common/share/Macaulay2/Core/factory/gftables/10201`

and when factory installs gftables, it puts it into

  `usr-host/share/factory/gftables/10201`

I have no memory now of how I arranged for the files to be found!  But notice also
we want the files to be found while M2 is running in the build tree, as well as when it gets
installed somewhere or untarred somewhere.

The error message `can not open GF(q) addition table: gftables/81` is inadequate, for
it doesn't tell what directory was examined for the files.  If you're under linux, you can use
`strace` to see.

d-torrance:
As I've been looking through the code, it looks like the current behavior is this:

- Set the location of the gftables in the Layout hash table [1].
- Confirm that this directory works [2].
- Tell factory what the directory is [3].

Of course, [1] could just point instead to $datadir/factory if we build factory ourselves.  But this breaks if factory is already present prior to build, since gftables would never be installed there.

If factory *is* already present, then perhaps we could hardcode the location of gftables in the hash table instead of giving a relative path.  If I'm understanding the factory code [4], factory doesn't know the directory of gftables unless it's explicity given with set_gftable_dir(), so we would still need to do that.

I'll work on this and submit a new pull request.

[1] https://github.com/Macaulay2/M2/blob/496552ddc64381cd160106d5959930b6acc52ef7/M2/Macaulay2/d/startup.m2.in#L287
[2] https://github.com/Macaulay2/M2/blob/496552ddc64381cd160106d5959930b6acc52ef7/M2/Macaulay2/m2/factor.m2#L51
[3] https://github.com/Macaulay2/M2/blob/496552ddc64381cd160106d5959930b6acc52ef7/M2/Macaulay2/d/actors5.d#L2075
[4] https://github.com/Singular/Sources/blob/23812d3dd3a69ed0874b01c271d1908acd7f0f05/factory/gfops.cc#L108

PR #377, opened 2016-03-19
Use gftables from factory if factory is not built.

### LINKED ISSUE #375: Is gftables necessary if we don't build factory?

d-torrance:
The configure script checks in the expected location for the gftables
directory or it may be specified using GFTABLESDIR.  Note that we must
expand the ${datadir} variable recursively in order for AC_CHECK_FILE
to work.  For this, we use the adl_RECURSIVE_EVAL macro from the Autoconf
Archive [1].

This fixes #375.

[1] http://ac-archive.sourceforge.net/adl/relpaths.html

### DIFF
--- a/M2/Macaulay2/d/startup.m2.in
+++ b/M2/Macaulay2/d/startup.m2.in
@@ -284,7 +284,7 @@ if firstTime then (
 	       "programs" => "@programsdir@/",
 	       "program licenses" => "@licensesdir@/",
 	       "package" => "@packagesdir@/PKG/",
-	       "factory gftables" =>  "@packagesdir@/Core/factory/",
+	       "factory gftables" =>  "@gftablesdir@",
 	       "packagedoc" => "@docdir@/PKG/",
 	       "packageimages" => "@docdir@/PKG/images/",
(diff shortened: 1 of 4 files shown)


DanGrayson:
I had to remove the line in GNUmakefile.in that would set
    PKG_CONFIG_PATH, for then when it re-runs the top level configure
    script, the version of factory we built as part of M2 the previous gets
    found!

PR #379, opened 2016-03-21
Check whether factory was built with --enable-streamio.

### LINKED ISSUE #378: Check whether factory was built with --enable-streamio
If we build factory, then --enable-streamio is passed to configure by default [1].  This is needed for the debugging display routines in [2].

But if we don't build factory, then we don't know how it was built and whether we have access to iostream with factory or not.  In fact, it defaults to *not* being enabled [3].

I think it might be possible to use something like AC_COMPILE_IF_ELSE to check and see if this feature is enabled if we don't built factory, and then choose whether or not to include [2].

[1] https://github.com/Macaulay2/M2/blob/0c90490/M2/libraries/factory/Makefile.in#L35
[2] https://github.com/Macaulay2/M2/blob/0c90490/M2/Macaulay2/e/x-factor.cpp#L222
[3] https://github.com/Singular/Sources/blob/485fc18/factory/configure.ac#L84

d-torrance:
Fixes #378.

### DIFF
--- a/M2/Macaulay2/e/x-factor.cpp
+++ b/M2/Macaulay2/e/x-factor.cpp
@@ -221,7 +221,7 @@ static struct enter_factory foo1;
 
     // debugging display routines to be called from gdb
     // needs factory to be configured without option --disable-streamio
-#if 1
+#if FACTORY_STREAMIO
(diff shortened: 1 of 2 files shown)


DanGrayson:
Thanks.

PR #388, opened 2016-03-26
If frobby isn't built, check to see if we can gets its version number.

### LINKED ISSUE #383: Frobby version number
In [1], we get the frobby version number from stdinc.h, which will definitely exist if we've built frobby, but may not if we've installed the frobby shared library beforehand, e.g., by using the Debian package.

I suppose we could use AC_CHECK_HEADERS to see if it's there, and just paste the corresponding lines from stdinc.h into version.dd if it's not.

Ideally, it seems like this should be something that's included in frobby.h instead of stdinc.h  This is even suggested in header file itself [2].

[1] https://github.com/Macaulay2/M2/blob/4af5a550200754e6b44dc4f0e5cca47a49ad97b1/M2/Macaulay2/d/version.dd#L41
[2] http://anonscm.debian.org/cgit/debian-science/packages/frobby.git/tree/src/stdinc.h#n100

d-torrance:
This fixes #383.

### DIFF
--- a/M2/Macaulay2/d/version.dd
+++ b/M2/Macaulay2/d/version.dd
@@ -38,8 +38,12 @@ header "
    extern char timestamp[];
    #undef ASSERT
    #ifdef HAVE_FROBBY
-     #include <stdinc.h>		// to get version number of frobby
-     #define FROBBY_VERSION constants::version
+     #if HAVE_FROBBY_VERSION_NUMBER
+       #include <stdinc.h>		// to get version number of frobby
+       #define FROBBY_VERSION constants::version
+     #else
+       #define FROBBY_VERSION \"\"
+     #endif
    #else
      #define FROBBY_VERSION \"not present\"
    #endif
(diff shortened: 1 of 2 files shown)


DanGrayson:
That looks like it might work, but it's hard to predict, since we haven't encountered any installations
of frobby in the wild that do provide its version number.  Still, it's worth trying.

d-torrance:
Frobby is interesting in that "make install" only installs the binary, not any of the library files.  So anyone that has the frobby library installed will have either installed those files manually or is using some distribution's package.  (The Debian package, for example, only installs frobby.h, not stdinc.h).

I did test this by manually copying stdinc.h to /usr/include alongside the Debian package, and the configure script worked as intended.

DanGrayson:
I doubt anyone making a distribution of the library would let the name "stdinc.h" survive, since
it doesn't contain "frobby".

d-torrance:
Good point.  Should we just automatically set HAVE_FROBBY_VERSION_NUMBER to 0 if we're not building frobby?

DanGrayson:
I think what you have now is fine, but I would suggest "unknown" instead of the empty string,
in that one case.

DanGrayson:
Thanks!

PR #390, opened 2016-03-27
Check if Prem() is public if not building factory.

### LINKED ISSUE #384: "Prem" if we don't build factory
In [1], we patch factory to make "Prem" visible.  However, if we use a pre-built factory, this won't necessarily be true.

Would it be worth submitting this patch upstream to the Singular devs?

[1] https://github.com/Macaulay2/M2/blob/4af5a550200754e6b44dc4f0e5cca47a49ad97b1/M2/libraries/factory/patch-4.0.2

d-torrance:
This fixes #384.


DanGrayson:
Thanks!

PR #394, opened 2016-04-08
Add new package VectorFields

bapike:
It does basic computations with polynomial vector fields (e.g., Lie bracket, computes the module of logarithmic derivations, etc.).  It has full documentation and tests.

I'm not sure if this matters, but there are some naming conflicts with other packages:
1) the HyperplaneArrangements package also defines a `der` method function.  It has a similar purpose to the `der` method function in VectorFields, although the types of the parameters are different
2) the Dmodules package also defines an `isHolonomic` method function.  It also has an implementation that takes an Ideal.

Please let me know if you see any problems or have any suggestions.


DanGrayson:
Thanks!

Issue #350, opened 2015-11-02
Kernel of a homomorphism of modules over different rings


craicu:
There is a problem with computing the kernel of a homomorphism of modules defined over different rings, when the target ring has no variables.

When the target ring was QQ[] I got:

i1 : A = QQ[x]

o1 = A

o1 : PolynomialRing

i2 : B = QQ[]

o2 = B

o2 : PolynomialRing

i3 : f = map(B,A,{0})

o3 = map(B,A,{0})

o3 : RingMap B <--- A

i4 : g = map(B^1,A^1,f,{{1}})

o4 = | 1 |

i5 : ker g
stdio:5:1:(3): error: assertion failed

When the target ring was QQ I got:
i1 : A = QQ[x]

o1 = A

o1 : PolynomialRing

i2 : B = QQ

o2 = QQ

o2 : Ring

i3 : f = map(B,A,{0})

o3 = map(QQ,A,{0})

o3 : RingMap QQ <--- A

i4 : g = map(B^1,A^1,f,{{1}})

o4 = | 1 |

i5 : ker g
stdio:5:1:(3): error: no coefficient ring present

DanGrayson:
For the first one, it seems that when the number of variables in a polynomial ring is 0, it picks a "degreesRing" for it containing one variable of degree 0, which doesn't match the degree 1 of the variable in the degrees ring for QQ[x].  Work around it this way:

```
B = QQ[Heft=>{1}]
```

For the second one, it's a limitation in the code to insist that both rings be polynomial rings or quotients of them.  To work around it, use the previous approach.

DanGrayson:
By the way, to debut such things, the first step is to rerun after setting `errorDepth = 0`, to see what the complaint actually is.

DanGrayson:
We should fix this eventually, both aspects.

PR #412, opened 2016-04-20
Small changes

kschwede:
This pull request fixes a minor issue in Pullback.m2 related to bug #350.  It also improves the documentation in both Pullback.m2 and in Divisor.m2


DanGrayson:
Thanks, I'll take a look at it.  You are in time for 1.9.

PR #413, opened 2016-04-21
Add meta tag to HTML docs to force UTF-8 on offline HTML documentation

bapike:
This change adds `<meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>` to the header of the HTML files containing M2's documentation.  No change should be visible in the online documentation at `www.math.uiuc.edu`, which already sends the HTTP header `Content-Type: text/html; charset=utf-8`.  However, this should fix a problem I encountered when viewing local (offline) documentation using my web browser (Firefox 45 on Debian), namely, characters such as "ö" display as gibberish (a pair of extended ASCII characters) unless I manually switch Firefox's 'text encoding' setting to unicode.

Validating the product produces the same 2 warnings that apply to the current online documentation (see [this](https://validator.w3.org/check?uri=http%3A%2F%2Fwww.math.uiuc.edu%2FMacaulay2%2Fdoc%2FMacaulay2-1.8.2%2Fshare%2Fdoc%2FMacaulay2%2FMacaulay2Doc%2Fhtml%2F_gb.html&charset=%28detect+automatically%29&doctype=Inline&group=0), for example).

Feel free to delay this change until after the release.

### DIFF
--- a/M2/Macaulay2/m2/html.m2
+++ b/M2/Macaulay2/m2/html.m2
@@ -204,6 +204,11 @@ links := tag -> (
      LINK { "href" => toURL doccss, "rel" => "stylesheet", "type" => "text/css" }
      )
 
+-- Also set the character encoding with a meta http-equiv statement. (Sometimes XHTML
+-- is parsed as HTML, and then the HTTP header or a meta tag is used to determine the
+-- character encoding.  Locally-stored documentation does not have an HTTP header.)
+defaultCharSet := () -> META { "http-equiv" => "Content-Type", "content" => "text/html; charset=utf-8" }
+
 BUTTON := (s,alt) -> (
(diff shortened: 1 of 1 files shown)


DanGrayson:
Thanks for the useful patch.  I've fixed the two warnings, too.

DanGrayson:
The new doc for 1.9 is online and the validator approves it totally, thanks!!:

https://validator.w3.org/check?uri=http%3A%2F%2Fwww.math.uiuc.edu%2FMacaulay2%2Fdoc%2FMacaulay2-1.9%2Fshare%2Fdoc%2FMacaulay2%2FMacaulay2Doc%2Fhtml%2F&charset=%28detect+automatically%29&doctype=Inline&group=0

PR #439, opened 2016-05-23
on a fresh repo one needs  --init --recursive

dimpase:
a small fix for INSTALL

### DIFF
--- a/M2/INSTALL
+++ b/M2/INSTALL
@@ -47,6 +47,10 @@ The following commands will download the latest changes to the source code:
     git pull
     git submodule update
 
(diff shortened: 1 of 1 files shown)


DanGrayson:
No, one doesn't, the makefile takes care of populating the submodules that need to be populated.  Letting it do so will save disk space by not populating the unused submodules.

dimpase:
well, at least one should tell the user how to get everything for building without internet access...

DanGrayson:
Yes, that's described in the section of the INSTALL file on building a fat tar file.

PR #443, opened 2016-05-23
Added RationalMaps package.

kschwede:
This is the package we mostly finished at the Utah Macaulay2 workshop.  It computes things like whether a rational map is birational/invertible/an embedding/etc.

### DIFF
--- a/M2/Macaulay2/packages/RationalMaps.m2
+++ b/M2/Macaulay2/packages/RationalMaps.m2
@@ -1,5 +1,5 @@
 newPackage( "RationalMaps",
-Version => "0.1", Date => "May 7th, 2016", Authors => {
+Version => "0.2", Date => "June 10th, 2016", Authors => {
      {Name => "Karl Schwede",
      Email=> "kschwede@gmail.com",
      HomePage=> "http://www.math.utah.edu/~schwede"
@@ -20,7 +20,7 @@ export{
 	"isBirationalMap",
 	"idealOfImageOfMap",
 	"baseLocusOfMap",
-	"dimImage",
+	"dimImage", --probably this shouldn't be exported
 	"isRegularMap",
 	"isEmbedding",
 	"relationType",
@@ -28,10 +28,21 @@ export{
 	"isBirationalOntoImage",
 	"inverseOfMap",
 	"mapOntoImage",
-    "isSameMapToPn", -- Dan: maybe we shouldn't export this.  Karl: I commented it out and made it internal.
+	--"blowUpIdeals", --at some point we should document this and expose it to the user
+	--"nonZeroMinor",-- it is internal because the answer is probobalistic (either it finds one or it doesn't) and it is controlled by MinorsCount option
+        "isSameMap", 
+        "sourceInversionFactor",
+        --"simisAlgebra", --at some point we should document this and expose it to the user
     --**********************************
     --*************OPTIONS**************
     --**********************************
+    "SaturationStrategy", --an option for controlling how inversion of maps is run.
+    "ReesStrategy", --an option for controlling how inversion of maps is run.
+    "SimisStrategy", --an option for controlling how inversion of maps is run.
+    "HybridStrategy", --an option for controlling how inversion of maps is run. (This is the default)
+    "MinorsCount", --an option for how many times we should randomly look for a minor before calling syz in inverseOfMap
+    "HybridLimit", --an option for controlling inversion of maps (whether to do more simis or more rees strategies)
+    "CheckBirational", --an option for inverseOfMap, whether or not to check if something is birational
     "SaturateOutput",  --option to turn off saturation of the output
     "AssumeDominant" --option to assume's that the map is dominant (ie, don't compute the kernel)
 }
@@ -42,8 +53,6 @@ export{
 --************************************************************--
 ----------------------------------------------------------------
 
-
---***Karl:  I dislike how this returns stuff.   The image of a map is not an ideal.  Hence I renamed it
 idealOfImageOfMap = method();
 
 idealOfImageOfMap(Ideal,Ideal,Matrix) := (a,b,f) -> (
@@ -69,7 +78,6 @@ idealOfImageOfMap(Ring,Ring,BasicList) := (R,S,g) -> (
 	);
 
 idealOfImageOfMap(RingMap) := (p) -> (
-        --idealOfImageOfMap(target p, source p, first entries matrix p)
         h := map(target p, ambient source p,p);
         im := ker h;
 	im
@@ -99,7 +107,6 @@ dimImage(Ring,Ring,BasicList) := (R,S,g) -> (
         );
 
 dimImage(RingMap) := (p) -> (
-	--dimImage(target p, source p, first entries matrix p)
 	I := idealOfImageOfMap(p);
 	(dim I) - 1
 );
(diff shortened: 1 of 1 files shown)


giovannistagliano:
Hi, isn't this very similar to Cremona package?

kschwede:
Indeed, it is similar.  After we discovered this, we tried to choose our function names not to conflict with yours.

However, I believe this handles more general cases.  For instance it doesn't require that the source be projective space in order to invert, but perhaps we misunderstood.

giovannistagliano:
The method invertBirMap works also when the source is not a projective space as well as works the method kernelComponent. 
My goal was to give general methods that in special important cases are faster, as show the examples in the documentation of Cremona.
 
I have not tested whether inverseOfMap is faster than invertBirMap or invertBirationalMap (in Parametrization.m2), is it so?

However, I was just curious if you were aware of this.

Bye

kschwede:
We didn't notice that your method called another package to compute inverses in the general case.  We noticed that your documentation said it computes things when the source is projective space. We haven't tried to compare that speed to ours.

giovannistagliano:
If I'm understanding properly, the algorithm implementated in inverseOfMap, in the case when the source is a projective space, is basically the same than that of invertBirMap (i.e. that of Russo and Simis), but it seems slower for me too.
Ex:
load "Cremona.m2"
load "RationalMaps.m2"
ZZ/3331[t_0..t_6];
phi=toMap(minors(3,matrix{{t_0..t_4},{t_1..t_5},{t_2..t_6}}),Dominant =>2);
time invertBirMap(phi);
-- used 0.468727 seconds
time inverseOfMap(phi,AssumeDominant=>true);
-- ????

kschwede:
Indeed, this should be basically the same...

I'm not sure why it's going so much slower...  We'll see if we can figure out what's going on.  Thanks for pointing this out!

HamidHassanzadeh:
Hi All,
First I also apologize for not contacting with Giovanni before doing the pull request. Indeed I have learnt about the package Cremona one day before the Workshop started.  (I always was supposing that the package Parametrization only works for curves and not higher dimensions, and just here I found that the function invertRationalMap is working generally.)
However because of the following reason we decided to write the package:
 The paper of Russo and Simis 2001 provides a methods to compute the inverse map if the ideal of definition of the map is "Linearly Presented". Notice that this is not a trivial assumption and I would say that this condition rarely happens for Binational maps, alias a very interesting example is the "Secant of Rational Normal Scroll" that Giovanni mentioned above 
minors(3,matrix{{t_0..t_4},{t_1..t_5},{t_2..t_6}}.
However, the first non trivial example of Plane Cremona maps is no more linearly presented
e.g. (x^3,x*y*z,y*z^2) I would say any non-saturated Plane Cremona map given by Polynomials of degree bigger than 5 does not have linear syz (more example is in my joint work with Simis, Plane Cremona Maps 2012). 
Anyway, in a later work with Simis and Doria (A characteristic free criterion of birationality, 2012) we could find the general procedure to compute the inverse map without considering the linear syz and it is the algorithm we used here. 
Honestly, the nature of the algorithm is not much different from the original work of Russo and Simis 2001 and even the pioneer ideas of Vasconcelos, moreover it is what that essentially used in the package Parametrization to compute the inverse. 
A reason why inverseOfMap is so slower in the above particular example in comparison with  invertBirMap,  returns to the   way we computed the blowUpIdeals. The codes that we use to compute the blowup ideal is the simplest Elimination techniques however if we had used the saturation of the symmetric algebra, the time used  would have been 0.532393 seconds which is very close to the optimized time  0.468727 seconds in Giovanni's techniques using the linear presentation of the ideal. I hope we will fix this issue soon

giovannistagliano:
Honestly, I cannot understand the meaning of your package.

baseLocusOfMap(phi) is not ideal(matrix(phi)) ?
dimImage(phi) is not dim kernel(phi) -1 ?
idealOfImageOfMap(phi) is not kernel(phi) ?
etc etc..

The purpose of Cremona was to provide faster methods (which are eventually probabilistic), but your methods are quite standard!

Your tests for birationality are much slower than computing the inverse using Cremona:
load "Cremona.m2"
load "RationalMaps.m2"
ZZ/3331[t_0..t_6];
phi=toMap(minors(3,matrix{{t_0..t_4},{t_1..t_5},{t_2..t_6}}),Dominant =>2);
time invertBirMap(phi);
-- used 0.472162 seconds
time isBirationalMap(phi);
-- used ?????????????????

DanGrayson:
This is an interesting discussion.  We can probably accept the pull request for inclusion with Macaulay2 even if there is substantial overlap, I suppose.  It might be good for the newer package to refer to the older one and offer a comparison in its top documentation node.

HamidHassanzadeh:
Hi Giovanni,
No, baseLocusOfMap(phi) is not just ideal(matrix(phi)), because a rational map may have different presentations so that one should take care of all different presentations(for example remove the common divisor if there is any). 
dimImage(phi) is not dim kernel(phi) -1 ? yes it is 
idealOfImageOfMap(phi) is not kernel(phi) ?yes it is
 However these 3 functions  are more internal functions which are used in isEmbedding function. 
About "Cremona.m2" I would say that I like it very much, first isBirational is very practical and fast, degreeOfRationalMap is also very helpful, however when you are working with general points one should be very careful if your field is finite or if the map is not dominant (because the probability of choosing a bad point is not zero any more ) e.g.
loadPackage"Cremona"
T=QQ[u,v,w,t]; map(T,T,matrix{{u*v^2,u*w^2,v^2*w,v*w^2}})
o79 : RingMap T <--- T
i80 : degreeOfRationalMap oo
o80 = 0.!!

About the example you are mentioning  invertBirMap. (it is a linearly presented ideal)
 betti res  minors(3,matrix{{t_0..t_4},{t_1..t_5},{t_2..t_6}})
o17 = total: 1 10 15 6
          0: 1  .  . .
          1: .  .  . .
          2: . 10 15 6
Such a phenomenon is not usual when one is working with binational maps.
I explained in the previous comment that invertBirMap does not work properly if Ideal is not linearly presented, the method that we present in RationalMap is general (there is no assumption on the map)  we also try to improve the speed that will be even better than invertBirationalMap, I hope so
Please let me know if I am not understanding your point of view

HamidHassanzadeh:
Sure Daniel,
We will edit the documentations and add the points you mentioned.

giovannistagliano:
Just an off-topic comment for Hamid.
Your map phi defined by uv^2,uw^2,v^2w,vw^2 is a map from P^3 onto a cubic surface, and thus it has degree 0, as degreeOfRationalMap gave.
Even, with examples so simple, you can use the option MathMode=>true, which makes the method deterministic (it replaces in the algorithm "random point" with "symbolic point").

load "Cremona.m2"
T=QQ[u,v,w,t]; 
phi=map(T,T,matrix{{u*v^2,u*w^2,v^2*w,v*w^2}});
time degreeOfRationalMap(phi)
-- used 0.0120477 seconds
0
time degreeOfRationalMap(phi,MathMode=>true) 
-- used 0.0295496 seconds
0

On the other example: your deterministic isBirationalMap takes twice the time than my deterministic degreeOfRationalMap(,MathMode=>true)==1
load "Cremona.m2"
load "RationalMaps.m2"
R=QQ[x,y,z]; phi=map(R,R,{x^3,x*y*z,y*z^2})
time degreeOfRationalMap(phi,MathMode=>true) == 1
-- used 0.0267972 seconds
true
time isBirationalMap phi
-- used 0.0488219 seconds
true

kschwede:
Dear Giovanni Staliano,

Thanks for pointing this out.  We'll definitely work to increase the speed and hopefully we'll have a new version shortly.

I'm not as expert at this as you and Hamid are, but I just want to confirm what's going on.  In the example above:

R=QQ[x,y,z]; f=map(R,R,{x^3,xyz,y*z^2})

invertBirMap f
o15 = map(R,R,{0, 1, 0})
(and MathMode gives a warning about this if turned on, but still returns the wrong answer)

This doesn't give the correct output presumably because the map doesn't have the maximal linear rank as explained above?  It looks like this might be documented in your package:  "a certain condition on ranks of appropriate matrices be satisfied"?  A random thought on my part, it might be better to call the parametrization version if mathmode gives that warning.

kschwede:
@DanGrayson 
Hi Dan, we will have a new version of this in a couple days I hope which is massively faster.  Is the right way to do this to close this pull request and make a new one?

DanGrayson:
No, the good way is to just update the branch used in the pull request.  We'll get informed and the discussion can continue on this thread.

DanGrayson:
@kschwede -- will your update be ready soon?

HamidHassanzadeh:
Hi Dan,
Yes it is almost ready we just need to review the documentation.

On Thu, Jun 9, 2016 at 12:54 PM, Daniel R. Grayson <notifications@github.com
> wrote:

> @kschwede <https://github.com/kschwede> -- will your update be ready soon?
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
> <https://github.com/Macaulay2/M2/pull/443#issuecomment-224991746>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/ARd_IoCmrkg9EqLpcsmThmnQb09Reklzks5qKGFXgaJpZM4Ik0VG>
> .
>



-- 




*Professor, Federal University of Rio de Janeiro*

kschwede:
I think I just updated the pull request with the updated documentation etc.  Dan, do you see this now points to version 0.2?

DanGrayson:
Yes, thanks.

kschwede:
Ok, you can go ahead and merge then.  Thanks!

HamidHassanzadeh:
Do you why in http://habanero.math.cornell.edu:3690
when I type
loadPackage"RationalMaps"
it returns:
stdio:1:1:(3): error: file not found on path: "RationalMaps.m2"

DanGrayson:
It's a new package, not in version 1.9.  It will be in 1.9.1, which will eventually be released in binary form and then eventually installed on the server.

PR #459, opened 2016-05-26
Fix for PARI's stack overflow.

### LINKED ISSUE #455: Segmentation fault when calling PARI functions with large numbers.
```
 i1 : factor 2^10000000
   ***   the PARI stack overflows !
   current stack size: 1000000 (0.954 Mbytes)
   [hint] you can increase GP stack with allocatemem()
 
 -- SIGSEGV
 -- stack trace, pid 17745:
 level 0 -- return addr: 0x0x10e57279f -- frame: 0x0x700000988dd0
 level 1 -- return addr: 0x0x10fb3e800 -- frame: 0x0x700000988df0
 -- end stack trace
```

gblanco92:
This fixes #455.

### DIFF
--- a/M2/Macaulay2/d/pari-c.c
+++ b/M2/Macaulay2/d/pari-c.c
@@ -93,44 +93,16 @@ static void closepari() {
 
 static GEN toPari(mpz_t x) {
   if (pari_disabled) return NULL;
-
-  /* Start PARI computations. */
   GEN z;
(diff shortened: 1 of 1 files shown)


DanGrayson:
It would be nice to add a test to make sure it stays fixed.

DanGrayson:
Why did I close this?  Reopening.

PR #477, opened 2016-06-02
improvements to toMap and added new method approximateInverseMap

giovannistagliano:


giovannistagliano:
Summary of changes from Cremona 1.0 to Cremona 2.0

Added method:
1) approximateInverseMap. This is a method to compute the base locus of the inverse map of a birational map. In some cases, e.g. for Cremona transformations, it can be used for computing the inverse.

2) isDominant (deterministic and probabilistic version, both very fast)

Added deterministic versions for the methods:
3) projectiveDegrees.  As far as I know, there is no other similar method that works when the source variety is not a projective space (neither probabilistic nor deterministic).

4) isBirational (unfortunately the deterministic version is not fast enough)

Other improvements 

5) invertBirMap now works in any case. I'm keeping to use the algorithm in invertBirationalMap from "Parametrization.m2", but I implemented it by myself and adapting it to my purposes.

6) toMap 

Best wished,
Giovanni

PR #478, opened 2016-06-02
Master

JoseMath:


DanGrayson:
You have committed changes to the versions of two submodules, probably unintentionally.  I'll fix that and bring the submodules up to date.

PR #493, opened 2016-07-07
David dev

eisenbud:


eisenbud:
I added a number of minor fixes of packages I'd written in this cycle.

Issue #516, opened 2016-08-25
Refactoring Polyhedra


lkastner:
This issue is dedicated to discussing various questions that came up while refactoring Polyhedra.m2. The current status can be found at [my fork](https://github.com/lkastner/M2/) and can hopefully be merged soon. The bugs [1](https://groups.google.com/forum/#!msg/macaulay2/Tu8X5Nw0zvA/uR_GwDdCBQAJ;context-place=forum/macaulay2), in minkowskiSum and at various other places have been fixed. Many tests have been and will be added. The following should be discussed:

1. What should the output for a polytope look like? Should it display any of the properties or just the type? Since everything was switched to lazy evaluation, we certainly would not want to compute anything here.
2. What about other packages connected to this refactor, see the following [pull request](https://github.com/Macaulay2/M2/pull/177). Does PolyhedralObjects.m2 make sense without Polyhedra.m2? Is Polyhedra2.m2 still being worked on (this duplicates old code, that has been refactored)? @JosephineYu 
3. As I understand I should not just fix packages that rely on Polyhedra.m2 to adjust them to the new version. What is best practice here?

Additionally I have the following minor questions:

1. There is now an interface to lrs. What is best practice for the tests for this interface? Can I make tests that only run if lrs is installed? This concerns future interfaces as well.
2. Can I make tests that should fail, i.e. check that an error is thrown?
3. Are there any bugs that I am not aware of? Should I write to the google group asking for bugs in Polyhedra?

Can I send a pull request, even though the documentation is not fully updated yet? All the old and new tests work, as well as all the examples from the documentation. Except for a few exceptions, all functions still exhibit the same behaviour.

JosephineYu:
I haven't been working on Polyhedra2.m2.  The original ideas was that gfanInterface2 would depend on PolyhedralObjects.m2 but not on Polyhedra.m2.  I haven't been keeping up with the current status of gfanInterface2,Tropical, etc.  You might want to check with Diane.

DanGrayson:
Re: "What is best practice for the tests for this interface? Can I make tests that only run if lrs is installed? "

For interfaces to proprietary software, such as Maple, we don't run tests.  There are two ways to do that: have no tests; or use the option CacheExampleOutput => true.  The latter will result in no tests being run automatically, and you must run the tests manually and add the output to the repository yourself whenever you think of it.

[...]

PR #523, opened 2016-09-05
First refactored version of Polyhedra.

lkastner:
Please see my latest comment at [issue 516](https://github.com/Macaulay2/M2/issues/516).

Besides Polyhedra.m2 I made changes to
1. LatticePolytopes.m2
2. MonomialAlgebras.m2

I did not make changes to SRdeformations.m2, although I think they might be relatively straight forward.


lkastner:
As @DanGrayson recommended, I will comment here:

1. @eisenbud @jankoboehm : I made changes to MonomialAlgebras.m2 to make it run with the new version of Polyhedra.m2. Please see commit [9797233b765b816a957cc535a1189c70702e369e](https://github.com/lkastner/M2/commit/9797233b765b816a957cc535a1189c70702e369e). Are you ok with these changes? Also I did not find Max Nitsche on github.
2. @alundman @mikestillman : I made changes to LatticePolytopes.m2 to make it run with the new version of Polyhedra.m2. This fixed a few bugs, but there are more. Maybe this is due to the latticePoints method from NormalToricVarieties.m2. @mikestillman : Did we already have a similar issue? How did we resolve it? Also I did not find Gustav Saeden Staahl on github.
3. @jankoboehm SRdeformations.m2 will not work with this version of Polyhedra.m2. Since you also know MonomialAlgebras.m2, the changes here should be relatively straightforward for you.

lkastner:
@kathlenkohn Does gfanInterface2.m2 work with this version? Just to make sure.

eisenbud:
Dear Lars,

Looks fine to me. I don't have Max' address either. Janko might...

David
--
David Eisenbud
Professor of Mathematics,University of California, Berkeley
www.msri.org/~de

jankoboehm:
Hello Lars,

changes to MonomialAlgebras.m2 look fine to me too. I will ask Max but I think he is fine. I will do the changes to SRdeformations.

Best,
Janko

mikestillman:
This compiles and all tests run.  I need to now check to make sure that NormalToricVarieties doesn't have any issue.  Perhaps @ggsmith should look too. There are a couple of undocumented functions which redefine hilbertBasis in a way that was not compatible with new versions in Polyhedra.  This might have been fixed, I just need to check.

mikestillman:
I am getting the following error when building:

make: Entering directory `/Users/mike/src/M2-lkastner/M2/BUILD/mike/builds.tmp/darwin64-gcc4.9/Macaulay2/packages/Polyhedra-temporary'
cd Polyhedra-temporary && /Users/mike/src/M2-lkastner/M2/BUILD/mike/builds.tmp/darwin64-gcc4.9/usr-dist/x86_64-Darwin-MacOS-10.10.5/bin/M2  -q --stop --silent -e errorDepth=3 -e debugLevel=0  -e "installPackage(\"Polyhedra\", RemakeAllDocumentation => false, IgnoreExampleErrors => false, RerunExamples => false, CheckDocumentation => true, AbsoluteLinks => false, UserMode => false, Verbose => false, InstallPrefix => \"/Users/mike/src/M2-lkastner/M2/BUILD/mike/builds.tmp/darwin64-gcc4.9/usr-dist/\", SeparateExec => true, DebuggingMode => true); exit 0"
currentString:1:1:(3):[3]: error: mutable unexported unset symbol(s) in package Polyhedra: 'nFacets'
../../../../../../../Macaulay2/packages/Polyhedra/core/cone/properties.m2:58:14-58:21: here is the first use of 'nFacets'

lkastner:
I'm sorry, I had some issue with nFacets. It should work now.

mikestillman:
I am also removing SRdeformations from the distributed package list (on my copy of this pull request), so I can see if all other packages build correctly.

mikestillman:
ToricVectorBundles doesn’t build (installPackage gives errors).  Possibly the function isPointed now takes optional arguments, or no longer does?  I’ll need to check.

lkastner:
The ''LatticePolytopes.m2'' package will not work. Presumably the issue with latticePoints is the same as with hilbertBasis. It might even be solved by just solving the issue for hilbertBasis. Do you see an easy solution?

mikestillman:
I just disabled ToricVectorBundles, and I’m trying NormalToricVarieties.  There are 5 example errors here.  I’ll look at the latticePoints, hilbertBasis problems here.  Hopefully they are easy to fix.

mikestillman:
Here is one of the errors from NormalToricVarieties:

i1 : PP3 = projectiveSpace 3;

i2 : F = fan PP3

o2 = F

o2 : Fan

i3 : rays F
stdio:3:1:(3): error: No input rays given.

lkastner:
The 'isPointed' function for fans was not fixed yet. I also modified the constructor for a fan in 'NormalToricVarieties.m2', your code works now for me.

Are there any other packages that I've missed? I cannot get 'ToricVectorBundles.m2' to run.

mikestillman:
Here is another issue.  hilbertBasis is (re-)defined in NormalToricVarieties, but now the function takes optional arguments in Polyhedra, whereas it didn't previously.  I think that @ggsmith said that the version in NormalToricVarieties was a stopgap measure until the Polyhedra function improved enough to use.

     ../../../../Macaulay2/packages/Polyhedra/core/cone/methods.m2:26:32-30:12: --source code:
     hilbertBasis Cone := List => o -> (C -> (
           if isPointed C then getProperty(C, computedHilbertBasis)
           else error("Hilbert basis not implemented for non-pointed cones yet.")
        )
     )
     ---------------------------------
     -- code for method: hilbertBasis(Matrix)
     ../../../../Macaulay2/packages/NormalToricVarieties.m2:722:36-723:55: --source code:
     hilbertBasis Matrix := Matrix => C -> (
       transpose (normaliz(transpose C,"integral_closure"))#"gen")

mikestillman:
This function in NormalToricVarieties needs to be rewritten for new Polyhedra interface.  @ggsmith
It should be easy, as the new ‘rays C’ is more what Greg wanted anyway, I think.

-- this function interfaces with the Polyhedra package
normalToricVariety Fan := opts -> FF -> (
  R := rays FF;
  F := sort apply(maxCones FF, C -> (
      Cr := rays C; 
      Cr = set apply(numColumns Cr, i -> Cr_{i}); 
      positions(R,r -> Cr#?r)));
  R = apply(R, r -> flatten entries r);
  return normalToricVariety(R, F,  
    WeilToClass => opts.WeilToClass,
    CoefficientRing => opts.CoefficientRing,
    Variable => opts.Variable))

mikestillman:
There appear to be only two issues with NormalToricVarieties and its interface to Polyhedra: hilbertBasis, and the fact that the interface for maxCones has changed (line 316 in that file).  Both of these are easy to fix.  @ggsmith: should I just do so, and show you the result afterwords for your review?

lkastner:
The optional arguments for hilbertBasis originate from FourTiTwo.m2, which I am using to compute a hilbert Basis.

The rays are probably not easier as before, but the maxCones should be exactly what NormalToricVarieties.m2 wants.

mikestillman:
Lars, @ggsmith did the following for Hilbert basis in NormalToricVarieties:

hilbertBasis(Matrix) := Matrix => opts -> (C) -> (
    transpose (normaliz(transpose C,"integral_closure"))#"gen")

(The cone is generated by the columns of this matrix, whereas the hilbertBasis command in 4ti2 package wants the walls of the cone.

I changed the code in NormalToricVarieties to use your version, via posHull.

Which is better?

mikestillman:
Lars, it looks like there are a number of issues with LatticePolytopes.  One option is for us to review it together, on skype, sometime in next few days. I'll continue to look at it.

lkastner:
I am adding @nilten , since he hopefully knows ToricVectorBundles.m2 better than us. Most minor problems are fixed, but now the following issue came up:
```
i1 : E = tangentBundle hirzebruchFan 3

o1 = {dimension of the variety => 2 }
      number of affine charts => 4
      number of rays => 4
      rank of the vector bundle => 2

o1 : ToricVectorBundleKlyachko

i2 : HH_0^E {matrix{{1},{0}},matrix{{-1},{0}}}
stdio:2:5:(3): error: expected nonempty list
```
I had to make a small change to the cechComplex funtion for the Klyachko case, which is probably triggering this, see my [commit](https://github.com/Macaulay2/M2/pull/523/commits/dd94010607e8fa00150e5c6eb56562884667f084). Basically I am just turning list of indices back into cones. @nilten Can you make any sense of this?

@mikestillman is there already a Cech complex functionality in M2 we could use to get rid of the cechComplex methods in ToricVectorBundles?

lkastner:
We were able to fix all documentation errors by replacing cones as keys in hashtables by pairs of matrices. The problem is now that matrices cannot be sorted canonically. Thus I added the old sorting method from Polyhedra and wrapped it into customConeSort. This is very inelegant. And some tests produce really different output than before. Probably the sorting will still not be the same everywhere, since Polyhedra.m2 will not sort the rays and instead just accept the output from FourierMotzkin, hence changing the ordering of the cones. Thus, if lrs or cdd are used instead of FourierMotzkin.m2, this might break things again. This seems very fragile.

lkastner:
Some final comments on my fixes in ToricVectorBundles.m2:
- It is still not clear why the ordering of cones is important. Maybe a
  different data structure should be used?
- There is a lot of old commented out code. Should this maybe be removed to
  improve readability? The file is under version control after all.
- All the fixes have been done without any understanding how this package
  works. I just adapted every call to Polyhedra.m2 to the new usage and removed the usage of cones as keys in HashTables.
- CacheTables are used in a weird way, more like MutableHashTables. The usage as in NormalToricVarieties would probably enhance readability.

mikestillman:
All examples passed, but 3 tests failed, one in NormalToricVarieties, and 2 in the schenck book (could be the same issue).  I’ll isolate it, and post failing code here (but right now I am busy for the next few hours…)

kathlenkohn:
I'm very busy until November 1, but I'll look into it in November!
Diane had some problems either way with the latest M2 update, so I'll try
this as well.

Am 05.09.2016 23:48 schrieb "lkastner" <notifications@github.com>:

mikestillman:
Lars, I’m going to build it from scratch, test it.  Is it at the point where I should allow the pull request?

lkastner:
I got one test failing in NormalToricVarieties.m2, which I could fix. The examples all worked for me. Did you already fix something?

Which package has the book examples?

mikestillman:
The problems I had were in ‘make check’, where it runs stuff from the book examples, and flags changes.  The examples, I believe, ran to completion.

The ones that had issues were some of: packages/Macaulay2Doc/test/schenck-book-*

lkastner:
I found schenck-book-2.m2 and fixed the usage there. I also added a few extra tests, to make sure that it still gives the right things.
```
make check
```
does not work for me unfortunately. I will make a reinstall tonight.

mikestillman:
try make -k check, 

search for *** through the log.

lkastner:
```
make -k check
```
reports many errors, but they seem not to be related to Polyhedra.m2. Does it work for you? I still have my submodules problem, even though I restarted + -compiled from scratch. I did --recurse-submodules when I cloned.

If it works for you now, it should be ok to accept the pull request.

mikestillman:
make -k check
should report very few (i.e. zero) errors.  What errors is it reporting?

lkastner:
It is just asking for gtest.h over and over, e.g.
```
src/test/Scanner.cpp:6:25: fatal error: gtest/gtest.h: No such file or directory
```
I don't have a gtest.h. I'll try to reconfigure and -build with enabling gtest as a build library.

lkastner:
I configured with
```
FC=gfortran ../M2/configure --enable-download --enable-build-libraries="pari gtest"
```
but I still haven't got gtest.h. The final message I get after make -k check is
```
make[1]: Leaving directory '/home/user/M2/lars/check-configure'
make: Target 'check' not remade because of errors.
make: Leaving directory '/home/user/M2/lars'
```

DanGrayson:
What is the status of this pull request?  Should I test it and install it?  If not, we could close it until it's ready.

DanGrayson:
What's the point of introducing new packages with these names?
```
OldPolyhedra
OldToricVectorBundles
```

PR #524, opened 2016-09-20
SpectralSequences.m2

n-m-g:
We would like to release version 1 of SpectralSequences.m2.  The current version to be released is in: Macaulay2/Workshop-2016-Utah/SpectralSequences


lkastner:
Could you untrack the [.DS_Store](https://en.wikipedia.org/wiki/.DS_Store) files? They are only needed for the Mac filesystem.

n-m-g:
Hi,

Can you please explain to me the easiest way to do that?

Many thanks.

Best,
Nathan

DanGrayson:
Of the seven files affected by this pull request, only one was intended.  The other accidental deletions and additions should be undone with "git" and the results should be pushed to your branch.

Also, the directions in `packages/README` explain that new package, such as yours, have to have their names added to a file where the names of the packages are listed.

DanGrayson:
Dear Nathan,

I think the problem is that you are working from a copy of our repository that you checked out on June 10, 2013, and you haven't updated it since then.  I didn't notice that before.

Is that possible?

If so, it might be too complicated for you to fix these errors on this branch of yours, and the best way might be to abandon the branch in favor of a new one that's in sync with ours.

DanGrayson:
Shall I just grab the file and add it directly?  Then we could close your pull request for now.  The file would come with no history, though.

n-m-g:
Hi Dan,

I think this is exactly the problem.  

Best,
Nathan

n-m-g:
Sure, that’s probably the easiest thing to do.  Many thanks for your help.

DanGrayson:
Okay, I added it.  Thanks for the new package!

It looks like a lot of examples that naturally would have had input expressions
spanning multiple lines were recast to avoid that.  I suggest fixing that.  (Multi-line
input expressions are entered in SimpleDoc by indenting the lines after the first
line of each expression.)

There are misspelled words.  E.g., "relavant".

In your repository you could do something like this:
```
git remote add upstream git@github.com:Macaulay2/M2.git
git checkout -b new-master upstream/master
git push -u origin new-master
```
Then just forget about `master` and work on `new-master`.  It should have your package
there, and all co-authors should now start working on it there.

n-m-g:
Great.  Thanks for this and for the feedback.  I hope people find the package useful.

Issue #528, opened 2016-10-22
outdated instructions for building on Arch


dimpase:
1. AUR package to build M2 is broken (probably it has to be adjusted to use gcc-5).

2. I needed to add `factory` to the proposed in Arch-specific instructoins `--enable-build-libraries` list.

DanGrayson:
What does the phrase "AUR package" refer to?

dimpase:
AUR in `You may prefer to use the Arch package for Macaulay2 at https://aur.archlinux.org/packages/macaulay2/.` (it's a distribution system for optional packages, see https://aur.archlinux.org/)

DanGrayson:
Well, then, this is not an issue for us, is it?  I'll close it, but reopen it if you think it needs attention by some of us.

dimpase:
IMHO an instruction that leads people into wasting time is not a good idea :-)

Regardless, missing `factory` in the list `--enable-build-libraries` is a separate issue, not related to AUR, but about building M2 on Arch linux directly from source.

DanGrayson:
I've just figured out that you are talking about our file `M2/INSTALL`!   I've repaired it in commit 00d4ccbfe552c1a4a2708734c2caae797f65a9c0 .

dimpase:
next time I do a pull request straight away. :-)

DanGrayson:
Sure, it's probably simpler than explaining the problem.

PR #529, opened 2016-10-23
some Arch-related edits

dimpase:
See #528


DanGrayson:
I've made equivalent changes in commit 00d4ccbfe552c1a4a2708734c2caae797f65a9c0.

Issue #533, opened 2016-11-13
yet another factor bug


pzinn:
I know I've already partly rewritten "factor", but for the record this bug was always there, it's just that my successive fixes never quite get to the bottom of things. Here we go:
```
i1 : R1=ZZ/101[x]

o1 = R1

o1 : PolynomialRing

i2 : R2=R1[y]

o2 = R2

o2 : PolynomialRing

i3 : promote(x^2,R2)

      2
o3 = x

o3 : R2

i4 : factor oo
stdio:4:1:(3): error: assertion failed
```
What's happening here? the problem is that sometimes factory returns a constant in front, sometimes it doesn't. To test this we use liftable(facs#0,R) where R is the base ring of the polynomial ring.
Unfortunately in the case of iterated rings this test is not quite right, and it will sometimes return true even if there is no constant in front (as in the case of x^2 -> (x,2)), and then the assert(exps#0==1) figures out there's something fishy.
I could put in yet another fix (like "leadCoeff" which is an iterated leadCoefficient) but I can't help thinking maybe we should rewrite everything, maybe by flattening the ring first...

pzinn:
OK, so it's not so bad, it's fairly straightforward to fix.
if nobody has any thoughts, I'll submit a pull request.

DanGrayson:
Thanks, that will be appreciated!

pzinn:
oh yeah. done.

DanGrayson:
Fixed by #534

PR #534, opened 2016-11-14
fix problem of factor with iterated polynomial rings, cf issue #533

pzinn:
- monomial prefactor correctly computed in the case of ring with inverses
- factory constant correctly identified


DanGrayson:
How about adding a test that fails currently and is fixed by this patch?

pzinn:
done

PR #553, opened 2016-12-28
added K3Carpets and replaced CompleteIntersectionResolutions. Also mo…

eisenbud:
…dified =distributed-packages


DanGrayson:
You probably don't want to change INSTALL.mac

DanGrayson:
@eisenbud : do you have any comments for the "changes 1.9.3" documentation node?

PR #558, opened 2017-01-03
Added the package "Chi2" implementing the constructions in Walker's p…

eisenbud:
…roof of the BEH conjecture

### DIFF
--- a/M2/Macaulay2/packages/=distributed-packages
+++ b/M2/Macaulay2/packages/=distributed-packages
@@ -134,3 +134,4 @@ EquivariantGB
 ExampleSystems
 RationalMaps
 K3Carpets
+Chi2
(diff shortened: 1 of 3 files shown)


DanGrayson:
You forgot to add the package, too.

mikestillman:
The name Chi2 isn't very descriptive.  I wonder if there is a better name for the package?

DanGrayson:
A better name would be good.  How about "BettiSums"?

How about a reference to Walker's work, so we know which Walker it is?

Perhaps mention that sym2 and wedge2 don't work in characteristic 2.  Those functions are rather special, focusing on just the second Adams operation, so it's okay if the package doesn't have a name indicating that it's a generally useful utility.

I can take care of the  conflict, but you could also do that by merging our master branch into your branch.

PR #566, opened 2017-01-10
MonodromySolver and ReactionNetworks

antonleykin:
Added these to =distributed-packages (target=1.9.3)


mikestillman:
After merging: while building M2, I get the error:

make[2]: *** No rule to make target `../../../../../../Macaulay2/m2/../packages/ReactionNetworks/oneSiteModificationA', needed by `../../../../../../Macaulay2/m2/TAGS'.  Stop.

DanGrayson:
fixed.  There was a file with a space in the name, so I renamed it.

antonleykin:
The problem pointed out by Mike still persists for me (on RHEL).

DanGrayson:
I fixed that days ago.  The problem was there was a file with a space in its name.

antonleykin:
False alarm, sorry...
RHEL compilation goes fine.
Now OS X 10.12.2 hiccups:
```checking whether package libxml-2.0 is provided... configure: error: no, package libxml-2.0 not found, install it```

but
```
ls -l /usr/lib/libxml*
-rwxr-xr-x  1 root  wheel  2337232 Dec 10 04:01 /usr/lib/libxml2.2.dylib
lrwxr-xr-x  1 root  wheel       15 Nov  5 14:10 /usr/lib/libxml2.dylib -> libxml2.2.dylib
``` 
so I am baffled.

**Comment by DanGrayson** [comment]
What does the command
```
pkg-config --version libxml-2.0
```

**Comment by antonleykin** [comment]
```
bash-3.2$ pkg-config --version libxml-2.0
pkg-config --version libxml-2.0
0.29
```

**Comment by DanGrayson** [comment]
Show me the output from
```
pkg-config --libs libxml-2.0
pkg-config --cflags libxml-2.0
```
and, in the M2 tree,
```
 git log -1
 git status
```

**Comment by antonleykin** [comment]
bash-3.2$ pkg-config --libs libxml-2.0

pkg-config --libs libxml-2.0

Package libxml-2.0 was not found in the pkg-config search path.
Perhaps you should add the directory containing `libxml-2.0.pc'
to the PKG_CONFIG_PATH environment variable
No package 'libxml-2.0' found

**Comment by antonleykin** [comment]
I solved the problem by putting 
```export PKG_CONFIG_PATH=/usr/lib/pkgconfig:/usr/lib:/usr/local/lib/pkgconfig:/usr/local/lib:/usr/local/opt/libxml2/lib/pkgconfig```
in `.bashrc`. It still does not find whatever is in `/usr/lib` but finds the "brewed" alternative in `/usr/local/opt/libxml2/lib/pkgconfig`

By the way, `--modversion` gets the version of the package, `--version` gets the version of `pkg-config`.

**Comment by DanGrayson** [comment]
Well, that will work, but it's not the way I do it. 

I don't use brew to install libxml2 any longer,
and pkg-config, as provided by brew, knows about the mac os version in /usr/lib and /usr/include,
without me setting PKG_CONFIG_PATH.  I suspect it gets it from this file:
```$ cat /usr/local/Homebrew/Library/Homebrew/os/mac/pkgconfig/10.12/libxml-2.0.pc
prefix=/usr
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include
modules=1

Name: libXML
Version: 2.9.4
Description: libXML library version2.
Requires:
Libs: -L${libdir} -lxml2
Libs.private: -lpthread -lz  -lm
Cflags: -I${includedir}/libxml2
```
It's a mystery to me why your pkg-config doesn't know about it.  I wonder if your homebrew has that file.

antonleykin:
Yep, my `/usr/local/Homebrew/Library/Homebrew/os/mac/pkgconfig/10.12/libxml-2.0.pc` is identical to yours. It is indeed a mystery.

PR #577, opened 2017-01-26
Merged MultipolynomialResultants and ChowForms into Resultants

giovannistagliano:


giovannistagliano:
Hi, I have a question: In the code (line 93), I assumed that if A = K[x_0,...,x_n]/I is a zero-dimensional algebra over a field K, then the command basis(A) always returns a monomial basis for the module A^1. Is this correct? Thanks.

DanGrayson:
I assume so.  @mikestillman ?

DanGrayson:
Hmm, are we still waiting for an answer here from @mikestillman ?

giovannistagliano:
Hi, it was just to be 100% sure. I did several tests with the package and I would say it works well.

Issue #598, opened 2017-05-01
Defect in differentiation of Laurent series.


gblanco92:
The `diff` command is wrong when you differentiate a Laurent series:

`i1 : R = QQ{t, Inverses => true, MonomialOrder => RevLex};`
`i2 : diff(t, t^-1)`
`o2 = 0`

gblanco92:
I propose a fix in #599.

d-torrance:
> **Written by Claude** (Claude Opus 5, via Claude Code), not by @d-torrance, whose account posted it -- please weigh it accordingly.

Closing as fixed. Both the defect reported here and the follow-on question raised in
[#599](https://github.com/Macaulay2/M2/pull/599) have been resolved by the code, even though
that pull request was never merged.

Checked on 1.26.06-40, `R = QQ{t, Inverses => true, MonomialOrder => RevLex}`:

| | 2017 | today |
| --- | --- | --- |
| `diff(t, t^-1)` | `0` | **`-t^(-2)`** |
| `diff(t, 1_R)` | — | `0` |
| `contract(t, t^-1)` | `0` | **`t^(-2)`** |
| `contract(t, 1_R)` | — | **`t^(-1)`** |

### The reported defect

`diff(t, t^-1)` is `-t^(-2)`, which is what the quotient rule gives, and `diff(t, 1_R)` is
`0`. That is the whole subject of this issue and it is right.

### The question left in #599

@gblanco92 closed that pull request after finding the patch incomplete, writing:

> This fix does not work for `contract` and Laurent series. It still outputs
> `contract(t, t^-1)` → `0`. I'm not sure what the expected output of `contract` on a
> Laurent series should be. For instance, should `contract(t, 1_R)` be equal to `0` or
> `t^-1`?

Today M2 gives `contract(t, 1_R) = t^(-1)` — the second of the two branches he named — and
`contract(t, t^-1) = t^(-2)`, consistently with it. So the case that blocked his patch now
behaves, and the convention he was unsure about has been settled in practice.

Worth being precise about *how* it was settled, though: as far as I can tell nobody stated
the choice. The behaviour changed and the question was never answered in the thread, so the
convention is recorded by the implementation rather than by a decision or by the
documentation. The four values above are here so that it is written down somewhere at least
once.

Thank you for the report and for the attempt at a fix — the analysis in #599, including
identifying that `contract` shares the code path and that the `is_invertible` check was the
relevant guard, is why the case is easy to check now.

[...]

PR #599, opened 2017-05-01
Fix for defect in differentiation of Laurent series

gblanco92:
I propose a one-line fix for #598.

### DIFF
--- a/M2/Macaulay2/e/poly.cpp
+++ b/M2/Macaulay2/e/poly.cpp
@@ -1856,7 +1856,7 @@ ring_elem PolyRing::diff_term(const int *m, const int *n,
                                     int use_coeff) const
 {
   int sign = 0;
-  if (!M_->divides(m, n)) return K_->from_long(0);
+  //if (!M_->divides(m, n)) return K_->from_long(0);
   if (is_skew_ && use_coeff)
     {
       exponents EXP1 = ALLOCATE_EXPONENTS(exp_size);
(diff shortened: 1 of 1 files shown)


M2/Macaulay2/e/poly.cpp
```diff
@@ -1856,7 +1856,7 @@ ring_elem PolyRing::diff_term(const int *m, const int *n,
                                     int use_coeff) const
 {
   int sign = 0;
-  if (!M_->divides(m, n)) return K_->from_long(0);
```
gblanco92: This optimization prevents the code below to work for Laurent series.

gblanco92:
This fix does not work for `contract` and Laurent series. It still outputs:

`i1 : R = QQ{t, Inverses => true, MonomialOrder => RevLex};`
`i2 : contract(t, t^-1)`
`o2 = 0`

I'm not sure what the expected output of `contract` on a Laurent series should be. For instance, should `contract(t, 1_R)` be equal to `0` or `t^-1`?

If `contract(t, 1_R) == t^-1`, then changing the fix to:

`if ((is_skew_ || (!use_coeff && !M_->is_invertible(m))) && !M_->divides(m, n))`
`return K_->from_long(0);`

makes both `diff` and `contract` work for Laurent series.

DanGrayson:
This pull request is old.  What is the current thinking about it?  Does it still need further improvement?  Should we pull it now and file a new issue for those further improvements?

PR #614, opened 2017-05-23
Added isVeryAmple method (to check if a divisor is very ample).

kschwede:
Just added a method to the divisor package.


kschwede:
Whoops, it looks like I was dumb and forgot to do some sanity checking (it was crashing if the divisor had an empty complete linear system).  This is fixed now in my fork (and it looks like the github pull request already knows I pushed the change to my fork...)

DanGrayson:
Yes, pull requests track the tip of the branch used.

PR #617, opened 2017-05-28
Master

JoseMath:


DanGrayson:
I've merged this pull request, but I undid the changes to the three submodules.  Let me know if that was wrong.

PR #618, opened 2017-05-29
Numerical packages update

antonleykin:
Updates to MonodromySolver, NumericalAlgebraicGeometry, NAGtypes, etc. for 1.9.3 release.


DanGrayson:
I've merged this, but I'd like to know why MonodromySolver and NumericalSchubertCalculus cache their example output.  If there's no good reason, we should undo that.

DanGrayson:
Oh, I see now that MonodromySolver may run phc, which isn't distributed with Macaulay2.  So perhaps that's the reason for that.  Is it?

antonleykin:
Yes, I think `phc` is the reason `NumericalSchubertCalculus` caches examples too.
By the way, is there a recommended way to run tests (TEST statements) that involve some software that is not in the distribution?

DanGrayson:
Like this:

```
installPackage ("PHCpack", RerunExamples=>true)
```

PR #619, opened 2017-05-30
Fixing some bugs and adding functionality to NormalToricVarieties

ggsmith:
This new version of NormalToricVarieties fixes bugs in computing the cartierDivisorGroup and isProjective.  It also adds a new database of toric divisors corresponding to embeddings into low-dimensional projective spaces.


DanGrayson:
PS: After merging, I found that another package used `wDiv`, whose name is now different.  So I fixed it.  It's a good name change, by the way.

PR #656, opened 2017-07-20
on a fresh repo one needs  --init --recursive

dimpase:
Something I prepared last year, but didn't get around to propose, it seems.


dimpase:
oops, wrong branches.

PR #707, opened 2017-11-12
made updates to tests and documentation in SymbolicPowers

eloisagrifo:
Deleted the test that involved frobeniusPower (which is now no longer exported) and made small documentation updates. Thanks!


DanGrayson:
By the way, if you really do want to test an unexported function, just run `debug SymbolicPowers` first.

eloisagrifo:
Thanks!

PR #710, opened 2017-11-19
Add support for localization of polynomial rings with respect to prime ideals.

mahrud:
This commit introduces changes in the core and in the engine, includes a new package, PruneComplex, and updates the LocalRings package. A new raw engine type LocalRing for localizations of polynomial rings at prime ideals, along with arithmetic operations for such rings, is defined in **e/localring.hpp**. The **m2/localring.m2** file in the core utilizes that type, allowing for making local rings as:
  ```RP = localRing(ZZ/32003[x,y,z,w], ideal"xy-zw")```

The **PruneComplex** package includes various methods for pruning chain complexes over polynomial and local rings. In particular, in the local or graded case, the output is guaranteed to be a minimal free resolution. Algorithms in this package are also implemented using C++ in **e/mutablecomplex.hpp** for speed. This introduces a new type MutableComplex in the engine.

The **LocalRings** package uses the package above to extend the following methods to such local rings: syz, res, trim, mingens, minimalPresentation, symbol//, inducedMap, symbol:, saturate, annihilate. This package now also includes a new method for computing **length** of modules over local rings and the **Hilbert-Samuel function** of modules over local rings, optionally given a parameter ideal. See [arXiv:1710.09830](https://arxiv.org/abs/1710.09830) for reference.

**Note**: Methods isSubset and symbol== are fixed in m2/modules2.m2 and reduce is fixed in m2/matrix.m2. Many other methods that only rely on the methods above, such as map, modulo, subquotient, kernel, cokernel, image, homology, Hom, Ext, Tor, etc. work for local rings automatically.

**Note**: The old LocalRings.m2 package from 2008 is stored in packages/LocalRings/legacy.m2 and is still usable, though it is slower and incompatible with new operations.

If you need specific methods that do not work, please email Mahrud Sayrafi.

### DIFF
--- a/M2/Macaulay2/d/actors4.d
+++ b/M2/Macaulay2/d/actors4.d
@@ -1044,6 +1044,7 @@ tostringfun(e:Expr):Expr := (
      is x:RawFreeModuleCell do toExpr(Ccode(string, "IM2_FreeModule_to_string(",x.p,")" ))
      is x:RawMatrixCell do toExpr(Ccode(string, "IM2_Matrix_to_string(",x.p,")" ))
      is x:RawMutableMatrixCell do toExpr(Ccode(string, "IM2_MutableMatrix_to_string(",x.p,")" ))
+     is x:RawMutableComplexCell do toExpr(Ccode(string, "rawMutableComplexToString(",x.p,")" ))
      -- NAG stuff begin
      is x:RawHomotopyCell do toExpr(Ccode(string, "rawHomotopyToString(",x.p,")" ))
      is x:RawSLEvaluatorCell do toExpr(Ccode(string, "rawSLEvaluatorToString(",x.p,")" ))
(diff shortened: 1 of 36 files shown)


mikestillman:
Mahrud, I think I saw changes to this pull request after you made it. Are you ready for me to review it, and merge it if it is ok?

mahrud:
Yes; I only added a few tests and improved the documentation a little bit.

By the way, should I update the documentation for (length, Module) in the
core? Or should I not touch that for now?

mahrud:
@mikestillman , @DanGrayson , it just occurred to me that some of the 162 commits are merges from upstream/master, some are things that are later taken out of this branch, etc.
If you want, I can make a patch file and apply it to a new clone of the upstream/master branch and then request a pull from that branch. This way there will be a single commit that contains all the changes. Should I do this instead and send a new pull request?

mikestillman:
I just finished testing the pull request (i.e. compiling, running tests), and it worked.  

Would it be fast for you to make those changes?  They are probably not necessary, but could be nice.

DanGrayson:
That would be welcome, but we've never even thought of requesting it of anyone.

Git will do that for you, with the "rebase" subcommand.

There is no need to make a new clone, because you can just start a new branch at the top of our master branch.

mahrud:
It is done. I also compiled and ran checks, only saw some errors when making the ComputationBook folder, but I don't think that's caused by my code.

A few notes: Since the previous LocalRings package was version 1.0, I changed it to 2.0. However, if you feel like 1.1 is more appropriate, feel free to change that. I did not update the documentation for (length, Module) because I'm not sure what would be the best way to fix it. Currently it is in Macaulay2Doc/functions/degree-doc.m2. Lastly, the options for Direction in PruneComplex are still "left", "right", etc., but will probably change later to something less confusing.

PR #713, opened 2017-11-20
RandomMonomialIdeals Release

dwilburne:
The Macaulay2 package RandomMonomialIdeals provides users with a set of tools that allow for the systematic generation and study of random monomial ideals. It also introduces new objects, Sample and Model, to allow for streamlined handling of random objects and their statistics in Macaulay2.


mikestillman:
I needed to change the =distributed file, but otherwise all examples and doc and tests run fine.  I merged the pull request.

Issue #737, opened 2018-02-25
incorrect computation of multidegrees


pzinn:
This is actually a very old issue, first raised (by your humble servant) in 
https://groups.google.com/forum/#!searchin/macaulay2/multidegree%7Csort:date/macaulay2/YYHqFaH8KTg/SMB7c__nAVIJ
when multidegree was implemented.
After that I forgot about it, mostly because my own experimental fork of M2 fixed it years ago, and I only realized now (10 years later) that the issue is still there in vanilla M2.
Anyway, first, the problem:
```
i1 : R=QQ[x,y,Degrees=>{{1,-1},{1,1}}]

o1 = R

o1 : PolynomialRing

i2 : factor multidegree ideal (x,y)

                2
o2 = (T  - 1)(-T )
       1        0

o2 : Expression of class Product
```
Anyone that knows anything about multidegrees knows the result is incorrect: it should be a homogeneous polynomial, and more precisely (T_0-T_1)(T_0+T_1) (product of additive degrees of variables). The reason Macaulay2 gets confused, as explained in the lengthy thread above, is that it does not differentiate multiplicative degrees from additive ones. Namely, the T's above are morally the logarithms of the ones living in the degrees ring. In many circumstances, that confusion is more or less harmless, but here it isn't. To make a long story short, these additive variables should always have degree 1, i.e., their ring is the polynomial ring *with ordinary grading*, which is not always the case with degree rings in Macaulay2 as currently implemented.
I'll propose a minimal fix now.

PR #738, opened 2018-02-25
fix for #737

pzinn:
Multidegree now correctly assigns degree 1 to variables when expanding at lowest order.
A test has been added.


DanGrayson:
Thanks.

Issue #743, opened 2018-02-27
integralClosure produce normalizations with repeated variable names


kschwede:
Consider the following example.

```
C = QQ[B1,B2,B3,B4,B5,B6]
I =  ideal(B4*B5+B1*B6,B1*B4+B2*B4-B3*B6,B1^2+B1*B2+B3*B5,B2*B5^2-B6^2,B1*B2*
       B5+B4*B6,B3*B4^2-B6^2,B3^2*B4-B1*B6-B2*B6,B2*B3*B4-B3^2*B6-B5*B6,B3^3-B1
       *B2-B2^2+B3*B5,B1*B3^2+B1*B5+B2*B5,B1*B2*B3+B3^2*B5+B5^2,B1*B2^2+B2*B3*
       B5+B4^2,B3^2*B5^2+B5^3-B3*B4*B6,B2^3*B4-B2^2*B3*B6-B3^2*B5*B6-B4^3-B5^2*
       B6)
D = C/I

integralClosure D
```
outputs 
```
QQ[w_(0,0), w_(0,0), B1, B2, B3, B4, B5,
       B6]/(B4*B5+B1*B6,B1*B4+B2*B4-B3*B6,B1^2+B1*B2+B3*B5,w_(0,0)*B6-B3*B4,w_(
       0,0)*B5+B1*B3,w_(0,0)*B4-B6,w_(0,0)*B3-B1-B2,w_(0,0)*B2-B3^2-B5,w_(0,0)*
       B1+B5,w_(0,0)^2-B3,w_(0,0)*B6-B2*B5,w_(0,0)*B5-B6,w_(0,0)*B4+B1*B2,w_(0,
       0)*B1+B4,w_(0,0)^2-B2)
```

Note that the variable w_(0,0) appears twice...

However, if I rename the new variables, only the first new variable is renamed.

```
i5 : toString integralClosure(D, Variable=>symbol TTT)

o5 = QQ[TTT_(0,0), w_(0,0), B1, B2, B3, B4, B5,
     B6]/(B4*B5+B1*B6,B1*B4+B2*B4-B3*B6,B1^2+B1*B2+B3*B5,w_(0,0)*B6-B3*B4,w_(0,0
     )*B5+B1*B3,w_(0,0)*B4-B6,w_(0,0)*B3-B1-B2,w_(0,0)*B2-B3^2-B5,w_(0,0)*B1+B5,
[...]

PR #751, opened 2018-03-07
fix for #743

pzinn:
- integralClosure correctly increments the first index of variables
- options (variable name, verbosity) are correctly passed to makeS2
- example given in the issue added as test


DanGrayson:
Thanks!

PR #755, opened 2018-03-10
fix #690

### LINKED ISSUE #690: Unexpected type: Tally + Set

> o1 = false
> o2 = true

pzinn:
- `merge` now assigns common ancestor class to the output.
- `merge` now allows `MutableHashTable` as input, and correctly assigns mutable property to the output if both input hash tables are mutable.
- no test: this is basic functionality, I'd say


pzinn:
As mentioned in the discussion of #690, I'm ready to add the switch of VirtualTally and Tally to this pull.

pzinn:
I need to unblock my whole pipeline of branches, so I'll move over the switch of VirtualTally and Tally now.

PR #759, opened 2018-03-15
Added Seminormalization package.  Added a minor update to Pullback (s…

kschwede:
This adds a new package Seminormalization.m2 that computes seminormalizations of rings.

This also adds a minor update to the Pullback package.  In particular:
If the user provides invalid arguments, the pullback function now specifies more carefully what is wrong.


DanGrayson:
Please add the name of the new package to =distributed-packages, too.

kschwede:
I always forget to do that.  This is now done.

PR #769, opened 2018-04-08
Fix #690

### LINKED ISSUE #690: Unexpected type: Tally + Set

pzinn:
repackaging this pull request since there was a mess with the commit history: 
- `merge` now assigns common ancestor class to the output.
- `merge` now allows `MutableHashTable` as input, and correctly assigns mutable property to the output if both input hash tables are mutable.
- switched `VirtualTally` & `Tally`, so that `Tally` is a descendant of `VirtualTally`, as should be.


pzinn:
damn, I'm cursed. closing while resolving conflict

PR #772, opened 2018-04-12
fix #690 with test added!

### LINKED ISSUE #690: Unexpected type: Tally + Set

pzinn:
- `merge` now assigns common ancestor class to the output (technically, if they have same parent, just like before)
- `merge` now allows `MutableHashTable` as input, and correctly assigns mutable property to the output if both input hash tables are mutable
- switched `VirtualTally` & `Tally`, so that `Tally` is a descendant of `VirtualTally`, as should be
- added some tests that display the new behavior, in an effort to make this PR more palatable


DanGrayson:
Thanks!  I hope all the overnight tests pass.

PR #791, opened 2018-06-01
New points

galettof:
Hello!

I am sending an updated version of the Points package with the work started at the Wisconsin workshop. I created a branch called NewPoints.

Best,
Fred.


galettof:
I apologize for the late reply. In case the information is still useful, this new version of the Points package includes methods to compute Gröbner bases for ideals of:
1) fat points in affine spaces;
2) reduced and fat points in projective spaces;
both using variants of the Buchberger-Möller algorithm already implemented in a previous version for reduced points in affine spaces.

PR #794, opened 2018-06-03
New version of VersalDeformations

nilten:
I've fixed a longstanding bug in VersalDeformations. The lifting algorithm would give an error or incorrect output if the tangent cone for the obstruction equations wasn't generated by the lowest order terms of the obstruction equations. This has been remedied by actually calculating the tangent cone, and making appropriate adjustments to the algorithm.


DanGrayson:
Thanks.  If you don't object, I'll use your comment above in the changes documentation node for version 1.11.1.

nilten:
Thanks -- yes, please use the comment in the changes documentation.

PR #802, opened 2018-06-07
Changed code for eagonNorthcott complex 

emduart2:
Updated the code of the eagonNorthcott complex. tIt now displays the correct degrees of the matrices in the case that the input matrix is homogeneous but whose entries are not necessarily linear forms.


DanGrayson:
Thank you, Eliana!

But, how about installing some tests, so it is ensured to continue to work?

DanGrayson:
If there is a suitable test, you could remove this assertion and save time:
```
assert isHomogeneous(chainComplex d);
```

Issue #782, opened 2018-05-10
toString Vector


pzinn:
I guess I missed that one in my recent Vector PR:
```
i9 : toString vector {1,2,3}
stdio:9:1:(3): error: no method found for applying super to:
                                     3
     argument   :  | 1 | (of class ZZ )
                   | 2 |
                   | 3 |

i10 : code(toString,Vector)

o10 = -- code for method: toString(Vector)
      /usr/share/Macaulay2/Core/modules.m2:78:22-78:73: --source code:
      toString Vector := v -> concatenate ( "vector ", toString entries super v )

i11 : code(net,Vector)

o11 = -- code for method: net(Vector)
      /usr/share/Macaulay2/Core/modules.m2:74:17-74:36: --source code:
      net Vector := v -> net super first v
```
note the missing "first" in toString(Vector). either that or super(Vector) should be defined.

pzinn:
might as well add some more so we get it right this time. here are a few more missing functions:
- `- Vector`
- `Vector ^ List`
will keep updating as I find more

DanGrayson:
Okay, thanks.  Vectors are not often used, in practice, and that can explain why no one has asked for more methods for them.

pzinn:
actually, after fixing this, I'm wondering: is there any reason "super" is called in the first place? `toString Matrix` just calls `toString entries` (w/o any "super"), why can't `toString Vector` call `toString entries` as well? unless I find a counter-example where `entries super` is not the same as `entries`, I'll just remove the `super` altogether.

pzinn:
in any case, a PR will be ready soon.
of course it'll have to wait till #772 is accepted...

PR #826, opened 2018-06-26
#782 fix

pzinn:
Fixes the toString Vector bug mentioned in #782, as well as implements all suggestions in the thread except `ideal Vector`.


DanGrayson:
Thanks!

PR #827, opened 2018-06-28
Find system gftables if they exist.

d-torrance:
d5bf979 introduced a regression in the feature added in 68a00e7.  In
particular, if we don't build factory, then we should use the gftables
already present on the user's system.

So we looked in two paths -- one prepended with `prefixDirectory`, for
M2-built factory, and one not prepended with `prefixDirectory`, for
system factory.

The second option was removed, an so we looked for gftables only in the
build directory.  But this would fail if we were using system factory.

This patch restores both options.

### DIFF
--- a/M2/Macaulay2/m2/factor.m2
+++ b/M2/Macaulay2/m2/factor.m2
@@ -51,11 +51,14 @@ inversePermutation = v -> ( w := new MutableList from #v:null; scan(#v, i -> w#(
 -- We mimic the procedure for finding a finite field addition table used in the routine gf_get_table
 -- for building the file name in "gffilename", in the file BUILD_DIR/libraries/factory/build/factory/gfops.cc .
 -- Reminder: the contents of currentLayout are determined by the file ../d/startup.m2.in .
-gfdirs = {prefixDirectory | currentLayout#"factory gftables"}
+gfdirs = {prefixDirectory | currentLayout#"factory gftables",
+	  currentLayout#"factory gftables"}
 i := position(gfdirs, gfdir -> fileExists(gfdir | "gftables/961")) -- 961==31^2
 if i === null
 then error ("sample Factory finite field addition table file missing, needed for factorization: ",
-    prefixDirectory, currentLayout#"factory gftables")
+    prefixDirectory, currentLayout#"factory gftables",
+    " or ",
+    currentLayout#"factory gftables")
 setFactoryGFtableDirectory gfdirs_i
 
 irreducibleCharacteristicSeries = method()


DanGrayson:
The issue seems moot, since we always build factory, as it depends on mpir, which we patch and build, and so we always provide gftables.    At the very least, there is no way for us to test this proposed patch.

DanGrayson:
It would be great if we could avoid building anything ourselves, and if I recently undid something of yours, it was unintentional.

The assignment
```
gfdirs = {prefixDirectory | currentLayout#"factory gftables",
	  currentLayout#"factory gftables"}
```
can't be correct as it stands.  We think of the paths in `currentLayout` as relative paths, and we expect them always to be paths relative to a "prefix" mentioned in `prefixPath` -- that's what the layout scheme is about.  So if we want to be able to refer to a system file via some absolute path, it has to be done by some mechanism outside the "layout" scheme.  We can come up with ideas for that together.

Recently Mike and I have been discussing with Anton Leykin about how to avoid building anything ourselves.  The main obstacle is that we patch mpir so we can guarantee that it uses libgc for all its memory allocations, by arranging for the function `mp_set_memory_functions` to do nothing.  But the whole thing is confusing me now, as always.  I'll forward you the latest couple of emails between us about it.

In any case, nothing has changed -- we never really had the option to use the system factory before now, and in the future we would like to be able to use it.

@antonleykin @mikestillman

d-torrance:
Thanks for forwarding the email discussion!

After looking at the code a bit more, I don't think absolute paths are even necessary for gftables.  *After* installation, when `prefixDirectory` is  `/usr/`, then we can just set `currentLayout#"factory gftables"` to `share/Macaulay2/Core/factory` if we're building factory ourselves or `share/singular/factory` (or wherever we find it) if we're using system factory.

The only issue is *during* installation, when `prefixDirectory` points to somewhere in the build directory.  So I think it's possible to fix this issue without resorting to absolute paths.  I'll play around with this idea a bit this weekend and hopefully will have a new commit soon.

DanGrayson:
For an installed distribution: one problem I see with that approach is that the layout of the operating system under /usr/ may not be the same as what we have in the Layout table.  Another problem with that approach is that we allow Macaulay2 to be unpacked anywhere, and then it should run, even if it's not under /usr/, and `prefixDirectory` is pointing somewhere random.

DanGrayson:
When depending on libfactory provided by the OS, it should be the job of the configure script to determine the (absolute) path to the gf tables and put it somewhere.  Our distributed Macaulay2 package can expect to find it in the same place, *provided* we also ensure that the package lists the gftables package as a dependency in the *.deb file (or the *.rpm file).

d-torrance:
Closing, as this issue is fixed with https://github.com/Macaulay2/M2/commit/7c1f738

PR #831, opened 2018-07-23
Several changes to SimplicialComplexes and GenericInitialIdeal

LorenzoVenturello:
I made several changes to the two packages:
- The method fVector was returning the wrong answer if the simplicial complex was defined over a multigraded ring. I fixed that and added the option Flag, which computes a finer f-vector if the simplicial complex is (d+1)-colorable, where d is its dimension.
- The same was done for the method hVector. I know it is implemented in the package SimplicialDecomposability, but I though it made sense to add it here as well.
- Added 3 minor functions: star, which computes the star of a simplicial complex w.r.t. a face, skeleton, which computes the j-th skeleton and joinSimplicial, which computes the simplicial join.
- I implemented a method to compute the algebraic shifting of a simplicial complex. Since this goes through the gin the package GenericInitialIdeal is now exported by SimplicialComplexes. Hope this is ok.
- Added the option Multigraded to the method gin, which computes the multigraded generic initial ideal if the starting ideal is defined over a multigraded ring. This carries over to the method algebraicShifting, to compute the multigraded (or colored) algebraic shifting of a complex.
- The documentation has been updated and tests added.

### DIFF
--- a/M2/Macaulay2/packages/GenericInitialIdeal.m2
+++ b/M2/Macaulay2/packages/GenericInitialIdeal.m2
@@ -13,7 +13,7 @@ newPackage(
      )
 --=========================================================================--
      
-export{"gin","lexgin","AttemptCount","Modular" } -- if the new routines which you are adding have new
+export{"gin","lexgin","AttemptCount","Modular","Multigraded"} -- if the new routines which you are adding have new
 -- names, then they need to be exported; otherwise they should not be
 -- exported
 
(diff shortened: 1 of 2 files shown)


DanGrayson:
@ggsmith @mikestillman @aseceleanu : what do you think of the proposed changes to your packages?

mikestillman:
@aseceleanu , are you fine with the changes for GenericInitialideals in this pull request ?

DanGrayson:
Two tests fail:
```
stdio:6:1:(3): error: 2 error(s) occurred running tests for package SimplicialComplexes
```

Can you fix it in the next few days?

DanGrayson:
One of the errors is a simple misprint, but the other is an undefined identifier, `algebraicShifted`.

LorenzoVenturello:
It turned out that also the second one was a misprint. I pushed an updated version and the tests seem to be ok.

PR #838, opened 2018-08-14
New package: toric variety invariants from its polytope

Martin-Helmer:
Given a projective toric variety X_A defined by a full rank integer matrix A with the vector (1,1,...,1) in its row space, the package computes the degree and codimension of the projective dual variety (i.e. the A-discriminant variety), the Euclidean distance degree of X_A, the polar degrees of X_A, and the Chern-Mather class of X_A working from the polytope conv(A), i.e. the convex hull of the columns of A. Note that we do not require that X_A is normal. This implements methods described in the papers:

[1] Martin Helmer and Bernd Sturmfels. "Nearest points on toric varieties." Mathematica Scandinavica 122, no. 2 (2018): 213-238. Arxiv version: https://arxiv.org/abs/1603.06544.
[2] Martin Helmer and Bernt Ivar Utstol Nodland. "Polar degrees and closest points in codimension two." Journal of Algebra and Its Applications (2017): 1950095. Arxiv version: https://arxiv.org/abs/1711.02381. 

This package depends on OldPolyhedra.m2, currently the normalized volume routine in the new Polyhedra.m2 package gives many warnings when used, and since my package must compute many normalized volumes it is practical/desirable to use OldPolyhedra.


DanGrayson:
Thanks, I'll take a look at it.

What are all those warnings about in Polyhedra?  In general, warning messages aren't helpful.

Martin-Helmer:
I think the warnings in Polyhedra are something to do with the 'volume' method in the new package still using behind the scenes methods the package identifies as deprecated. This shows up in any volume computation for example, the warnings are even printed in the online docs page for 'volume' in Polyhedra.m2:

http://www2.macaulay2.com/Macaulay2/doc/Macaulay2-1.12/share/doc/Macaulay2/Polyhedra/html/_volume.html

DanGrayson:
Would it be good enough, from your point of view, if such a warning message appeared only once per session?

mikestillman:
@lkastner, what do you think?

Martin-Helmer:
Yes, for my package it would be totally fine if such a warning message only appeared once per session. It is just the warning appearing (multiple times) each time you call 'volume' that is a problem.

DanGrayson:
I fixed it:
```
commit f7b4155476a29e0455d626a07d31d4057493fae3 (HEAD -> master, upstream/master, upstream/HEAD, dan/master)
Author: Daniel R. Grayson <dan@math.uiuc.edu>
Date:   Tue Aug 14 17:44:05 2018 -0400

    fix package Polyhedra
    
    so deprecation messages appear only once
```

DanGrayson:
Are you now willing to switch from OldPolyhedra to Polyhedra?

Martin-Helmer:
Yes, I can do that. It will require rewriting my package a little since the syntax is a bit different for the faces of a polytope etc, but it shouldn't be too hard to make the switch. I will make the changes and upload a new version this week.

Martin-Helmer:
I have rewrote the package to use 'Polyhedra.m2' . I accidently closed this pull request so I will submit a new pull request for the updated package. 

One other thing of note I discovered in Polyhedra is that the volume function still prints strange warnings in some cases. I was able to make this disappear, but the fix seems rather odd, perhaps @lkastner would want to look into this at some point. An example is below

--the first two loops below produce the warning  "Warning: Property empty already assigned." for each iteration of the loop, while the last example does not:

needsPackage "Polyhedra"
    for i from 1 to 4 do(
	    C=convexHull(matrix{{0,1,1},{0,-2,3}});
	    lV=latticeVolume C;
	    );
    for i from 1 to 4 do(
	    C=convexHull(matrix{{0,1,1},{0,-2,3}});
	    lV=volume C;
	    );
--this does not produce any warnings    
    for i from 1 to 4 do(
	    C=convexHull(matrix{{0,1,1},{0,-2,3}});
	    vertices C;
	    lV=volume C;
	    );

Since I need to create different polytopes by taking convex hulls of points in a loop many times in my package I have added the call to vertices (which I don't need) to suppress the warnings. There is little overhead to doing this, so it is fine from my point of view, but it is a bit strange.

PR #843, opened 2018-08-22
New Package: Containment, algebraic multiplicity of a subvariety and intersection product via Segre

Martin-Helmer:
This is a new package which tests containment of varieties and computes algebraic multiplicity of subvarieties and Fulton-MacPherson intersection products using Segre classes. 

More precisely this package tests containment of (irreducible) varieties and computes Segre classes, algebraic multiplicity, and Fulton-MacPherson intersection products. For ideals defining multi-projective schemes, this package tests if a top-dimensional irreducible component of the scheme associated to an ideal is contained in the scheme associated to another ideal. Specialized methods to test the containment of a variety in the singular locus of another are provided, these methods work without computing the ideal of the singular locus and can provide significant speed-ups relative to the standard methods when the singular locus has a complicated structure. The package works for subschemes of products of projective spaces. The package implements methods described in:

Corey Harris and Martin Helmer. "Segre class computation and practical applications." arXiv preprint arXiv:1806.07408 (2018). Link: https://arxiv.org/abs/1806.07408.

### DIFF
--- a/M2/Macaulay2/packages/SegreClasses.m2
+++ b/M2/Macaulay2/packages/SegreClasses.m2
@@ -1,0 +1,1134 @@
+newPackage( "SegreClasses",
+    Version =>"1.02",
+    Date => "July 13, 2018",
+    Authors => {
+        {Name => "Martin Helmer",
+         Email => "m.helmer@math.ku.dk",
+         HomePage => "http://martin-helmer.com"},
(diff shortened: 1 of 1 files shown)


DanGrayson:
Thanks, we'll take a look at it (after a delay, I expect)!

PR #847, opened 2018-09-04
Updated Divisor.m2 to version 0.3.

kschwede:
This updates Divisor.m2 to the latest version.


kschwede:
Note, I also just added a new package (FThresholds) to this pull request and included a substantial update to another package (TestIdeals)

kschwede:
Whoops, I realized I hadn't correctly added the FThresholds package.  Now it is correctly included.

mikestillman:
@kschwede, I'm getting errors with trying to install Divisor.m2.  It appears that the newPackage at the top of the file is all messed up.  Could you fix it, and I'll try again?

It should also not have Reload set to true, nor DebuggingMode set to true, I think.

mikestillman:
@kschwede, actually, the newPackage header is just badly indented.  But when I try to install the package, I get errors.

kschwede:
@mikestillman 
Sorry about that.  I fixed the DebugginMode and Reload, and also fixed the extra error that seemed to be causing the problem when installing (an extra line break where it shouldn't be I think).  It seems like it installs correctly on my machine and also Habanero (both 1.12).
Thanks!

kschwede:
@mikestillman Hi Mike.  Was there anything else I needed to do?  Hopefully I did that right.

kschwede:
This pull request has just been updated with updates to Seminormalization.m2

PR #851, opened 2018-09-10
added a package

ycid:


giovannistagliano:
Hi! Everything that this package does, it could also be done by the Cremona package. It's right? Is there any overlap? Perhaps, are the methods of this package faster? Thanks.

ycid:
Hi, thanks for the interest. 
Yes, regarding computations with rational maps, everything this package does can also be done with the Cremona package.
Also, the probabilistic method of Cremona is faster.

The main idea of this package is to make the computations via a new algebra that we call the ``saturated special fiber ring''.
The main contribution of this package is being able of computing this algebra, and then using it to study rational maps.
There is also an implementation of a multigraded version of the Jacobian dual criterion (that works without using a Segre embedding).

giovannistagliano:
Thanks for the reply. I will try to take a closer look as soon as I can.

giovannistagliano:
Anyway, it seems to me a bit strange that the Cremona package is not mentioned neither in the documentation nor in the article. Even because I found this file 
http://www.ub.edu/arcades/ycid/degComputationExamples.m2 
where you compare your new method with some of the methods provided by Cremona. So you were aware of the overlap...

ycid:
In the article we acknowledge very graciously the use of Macaulay2!
If we had to acknowledge all the packages we used then the list would have been large (ReesAlgebra, RationalMap, Cremona, IntegralClosure, etc).
I don't see very clear why in a scientific paper a particular Macaulay2 package should be mentioned.

I sincerely apologize for not mentioning the packages RationalMap and Cremona in the documentation of the package. The use of these two packages was very useful. 
This is the first package that I write, and I did it during the Macaulay2 workshop in Leipzig (https://github.com/Macaulay2/Workshop-2018-Leipzig/tree/Tate/MultigradedBirationality).

I knew of the overlap, and at no time did I try to hide it. I even published that public file that you found and which is mentioned in the acknowledgements of the paper.

Also, I think you are missing the point of the file, because we don't ``compare'' with Cremona, we ``certify'' with Cremona (it is written literally in that way in the beginning of the file).
We were not sure that our computations were correct, so we were humbly trying to see if we got the correct results from the package Cremona.

I am grateful to the creators of the packages RationalMap and Cremona, and I sincerely apologize for not mentioning these two packages in the documentation of the package that I wrote.
I just wanted to share a package, and since it is first time that I write a package, I clearly made some mistakes.

giovannistagliano:
The package Cremona is not able to certify anything! You can just compare your computations, and you can do this because someone has done a mathematical and computer work to develop the algorithms and implement them in code. Thus thanks for mentioning the package. Bye.

ycid:
I have improved the documentation and explained the overlap with other packages.

aarondall:
The M2 packages you use should also be cited in your article. People who contribute packages that you find useful deserve to be recognized for their effort. 

Citing Macaulay2 packages in .tex documents is painless using the `cite` method. In any  Macaulay2 session (with, e.g., Cremona loaded) just type

	i2 : cite Cremona
	
	o2 = @misc{CremonaSource,
	       title = {{Cremona: Some computations for rational maps between projective varieties. Version~4.2.1}},
	       author = {Giovanni Staglian{\`o}},
	       howpublished = {Available at \url{https://github.com/Macaulay2/M2/tree/master/M2/Macaulay2/packages}}
	     }

and copy/paste into your bibliography file.

mikestillman:
Hi @yicd, what is the status of this pull request?  Is it ready for me to take a look at it?

ycid:
Hi @mikestillman, from my part is ready. I tried to fix some mistakes that I made. Thanks for considering it.

Issue #764, opened 2018-03-24
probable typo in expressions.m2


pzinn:
expressions.m2, L778:
```
isNumber Holder := i -> isNumber i#1
```
since Holders contain only one item, I'm assuming this should be isNumber i#0?

DanGrayson:
I agree.

PR #855, opened 2018-09-17
the great expression/texMath upgrade

pzinn:
This PR is a significant rewrite of `expression` (mostly in expressions.m2) as well as of `texMath`.
- `expression` now works for pretty much any mathematical object, not just elementary ones like ring elements, effectively breaking down expressions into smaller elementary components. It still fulfills the same role as before -- in particular `value expression x == x` in more cases than before (see tests). This upgrade makes it significantly easier to define alternate output modes.
- Solves issue #764 (but that's just the tip of the iceberg)
- `texMath` expanded to properly treat every `Type`. Note that `texMath` is now MathJax/KaTeX-compatible.
- The output routines of `Expression` types were full of small problems because many were little or never used. In particular, spacing and under/over-parenthesizing of expressions were fixed. (e.g., pre-PR, compare `fact(QQ[x])` with its `expression`, or try `(expression true) and (expression false)`)
- A new file `webapp.m2` to accommodate the use of the currently developed web app
https://github.com/pzinn/InteractiveShell
(a first step towards a fully graphical, web-based interface) building on Mike Stillman et al's current web app. `webapp.m2` relies crucially on the `expression` and `texMath` upgrades (and so does the web app itself). It consists of two parts: first it defines a new method, `texOrHtml`, which can be used independently of the web app, and which outputs HTML with math in (La)TeX format surrounded by \\( \\); secondly, it defines a new output mode, `TopLevelMode=WebApp`, which is used by the web app to communicate with Macaulay2.
- Change to `describe`: now outputs an `Expression` (more precisely, a `Describe`) to make it output mode-agnostic. The net formerly obtained by `describe x` would now be `net describe x`. (in particular, this change is invisible in standard output mode)
- Tiny improvements to `describe` and `toExternalString`.
- Documentation fixes (mostly, taking care of undocumented warnings related to this PR).
- Tests for expressions (4 small examples that currently fail).
- Warning, this PR is large. Despite extensive testing over the past 6 months, it probably still requires careful review.


DanGrayson:
`QQ` no longer prints as ℚ in texmacs mode.  Similarly for ZZ.  I can't figure out why.

DanGrayson:
Some of these need documentation:
```
+	"Describe",
+	"MapExpression",
+	"MatrixDegreeExpression",
+	"SheafExpression",
+	"VectorExpression",
+	"WebApp",
+	"texOrHtml",
```

DanGrayson:
This change is probably not good:
```
@@ -54,7 +55,7 @@ expression Expression := identity
 Expression#operator = ""
 
 expressionValue = method(Dispatch => Thing)
-expressionValue Sequence := x -> apply(x,expressionValue)
+expressionValue BasicList := x -> apply(x,expressionValue)
 expressionValue Thing := identity
 
 -- with the following line we have no way to distinguish between "hold symbol x" and "hold x" when x has a value:
```

If a type is implemented as a type of BasicList, then it's up to the implementer to figure out how to make an expression out of it.  Basic lists are internal computerish things, not mathematical things.

DanGrayson:
This change seems wrong, since &bsol; is a single backslash, and since backslashes have not syntactic meaning in html.

```
 -- html output
 
-htmlLiteral = s -> if not match("<|&|]]>|\42",s) then s else (
+htmlLiteral = s -> if not match("<|&|]]>|\42|\\\\",s) then s else (
      s = replace("&","&amp;",s);			    -- do this one first
      s = replace("<","&lt;",s);
      s = replace("]]>","]]&gt;",s);
      s = replace("\42","&quot;",s);  -- note: \42 is "
+     s = replace("\\\\","&bsol;",s);
      s )
 
```

DanGrayson:
If the purpose of `texOrHtml` is to support mathjax, then the documentation should say that.  Also, there should be a better name, since the "Or" makes it sound ambiguous.

DanGrayson:
Thanks for all the great work!  That's all the comments I have.

pzinn:
Thanks for all the comments. you're absolutely right about everything. I'm especially surprised with this expressionValue BasicList, I tend to be careful with such things. I'll replace it with VisibleList, I think we do need it for all visible lists.
Concerning \\ not being special in HTML, I agree. In fact until last week I had a separate routine that did this extra substitution (called htmlAltLiteral), and then I decided to merge it with htmlLiteral because that extra substitution is harmless (if unnecessary). The reason I need it is of course for texOrHtml, for which \\ is definitely a special character. I'll revert that change, I guess.
I haven't looked at all at this texmacs mode, oops. I hope I haven't broken anything else. I'll look into ZZ/QQ.

pzinn:
- docs fixed (minors improvements to `MapExpression` and `MatrixDegreeExpression` while rewriting their docs).
- `texOrHtml` renamed `htmlWithTex`.
- all other issues mentioned  by Dan fixed.

The only nontrivial one is actually the first one (texmacs).
The point is that mathml.m2 might need some rewriting to take into account the expression update. (the good news being, it'll be easier than before! one (only) needs to define `mathML` of stuff whose `expression` is a `Holder` of themselves, or or `Expression` subtypes, and for all the rest have `mathML x := mathML expression x`, similarly to say `texMath`). 
For now, I have done the minimum to make ZZ / QQ work again. (at least I hope -- I don't use texmacs myself so can't test)

DanGrayson:
ZZ and QQ now work again.  By the way, to test it yourself, just add --texmacs to the command line and enter ZZ as input.  The result includes this:
```
html:<math xmlns="http://www.w3.org/1998/Math/MathML"><mtable columnalign="right center left">
<mtr><mtd><mstyle color="red"><mi>o2</mi></mstyle></mtd>
<mtd><mstyle color="red"><mo>=</mo></mstyle></mtd>
<mtd><mi>ℤ</mi></mtd>
</mtr>
<mtr></mtr>
<mtr><mtd><mstyle color="red"><mi>o2</mi></mstyle></mtd>
<mtd><mstyle color="red"><mo>:</mo></mstyle></mtd>
<mtd><mtext>Ring</mtext></mtd>
</mtr>
</mtable>
</math>
```

PR #858, opened 2018-10-05
Numerical Certification package is now documented

klee669:
I created documentations for NumericalCertification package.

Also, it now supports interval arithmetic certification (Krawczyk method).


mikestillman:
One small typo I saw: in the Headline for the package, the name of the package is mispelled (Numercial).  Do you want to change it, and then I'll accept the pull request?  Or, I can accept it and fix that myself. @klee669, which do you prefer?

mikestillman:
@klee669 , also, the NumericalCertification/ directory is essentially already the Documents directory for the package.  I don't think that you need the Documents subdirectory.  If you really want it, I suppose it is fine, but it seems unnecessary.

klee669:
Hi!

Please accept it and fix the typo. Also, it is okay to delete the NumericalCertification/ directory if it is not necessary.

Thanks!

Issue #695, opened 2017-09-22
"strange result" with substitute no longer strange


foxjwill:
The [documentation for `substitute`](https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.10/share/doc/Macaulay2/Macaulay2Doc/html/_substitute.html) says "If the values of all of the variables are in a different ring, then the result will be in that ring." Then below this there's some output, but this output is the same as the output for "By changing one of the values to a rational number, we insure that the result will be rational."

And while we're here, the correct word to use is "ensure", not "insure".

DanGrayson:
Thanks for both catches!  Here is the relevant bit:
```

      If the values of all of the variables are in a different ring, then the result will be in that ring.

      +----------------------------------------+
      |i10 : use ring F;                       |
      +----------------------------------------+
      |i11 : sub(F, {a=>1, b=>3, c=> 1, d=>13})|
      |                                        |
      |o11 = 88                                |
      |                                        |
      |o11 : QQ                                |
      +----------------------------------------+

      This can have strange results, if the values are all integers, but fractions are present.

      +--------------------------------------------+
      |i12 : sub(1/3*a*b, {a=>1, b=>1, c=>1, d=>1})|
      |                                            |
      |      1                                     |
      |o12 = -                                     |
      |      3                                     |
      |                                            |
      |o12 : QQ                                    |
      +--------------------------------------------+

      By changing one of the values to a rational number, we insure that the result will be rational.

      +------------------------------------------------------+
      |i13 : sub(1/3*a*b, {a=>1_QQ, b=>1, c=>1, d=>1})       |
      |                                                      |
      |      1                                               |
      |o13 = -                                               |
      |      3                                               |
      |                                                      |
      |o13 : QQ                                              |
      +------------------------------------------------------+
```

Part is wrong, too, since we changed the noxious behavior being documented.  We'll have to come up with better examples.  I think what we made it do is to replace the ring containing the values by one minimally large enough to sensibly receive a ring homomorphism from the coefficient ring of the polynomial ring.

d-torrance:
This was fixed way back in #870.

Issue #651, opened 2017-07-14
Documentation for flattenRing doesn't fully describe functionality


kschwede:
Sorry if this isn't the right place (ie, do Documentation Issues go here?), but the function flattenRing asserts:

> write a ring as a (quotient) of a polynomial ring over ZZ or a prime field

However, as far I can tell, it also works for things over QQ, Galois fields, function fields, etc.  The lower part of the documentation even seems to acknowledge this (ie, you can use toField...)

lilysilverstein:
Indeed, several of the methods for flattenRing are currently undocumented. Will be fixed shortly.

lilysilverstein:
Rewrote this documentation to include things like Galois fields and rings declared as fields.

I was extremely confused by the description of the Result option. I changed the page to go into much more detail, with more examples, of the various numbers, types, and sequences that can all be passed to Result. Finally, I added documentation nodes for the options.

Commit is [here](https://github.com/lilysilverstein/M2/commit/39cc70ff91e2e24626e66a1cf7055dc86c242303), and will be bundled into my next documentation pull request.

d-torrance:
> **Written by Claude** (Claude Opus 5, via Claude Code), not by @d-torrance, whose account posted it -- please weigh it accordingly.

Closing as fixed — this was done in 2018, reported here at the time, and just never closed.

@lilysilverstein's second comment on this thread,

> Rewrote this documentation to include things like Galois fields and rings declared as
> fields.

is `39cc70ff91` ("rewrite flattenRing and its options", 2018-07-22), carried by **#870**,
merged 2018-11-14.

### The specific complaint is gone

The headline that prompted the report is no longer there. `flattenRing-doc.m2:15` now reads

> write a ring as a (quotient of a) polynomial ring

with the "over ZZ or a prime field" clause dropped — which is the whole of what was
inaccurate.

All three cases named in the report are now documented with worked examples:

| named in the report | on the page today |
| --- | --- |
| Galois fields | *"`flattenRing` can also be used to represent GaloisFields as quotients of polynomial rings"*, with `K = GF(5^3)` |
| function fields | `K = frac(ZZ[a])`, then `B = K[x,y,z]/(...)`, showing the coefficient ring of the result is `K` |
| `toField` rings | *"Once a ring has been declared to be a field with `toField` then it will be used as the coefficient ring"*, with `A = QQ[a]/(a^2-3)` |

The page also now states the general rule the old headline was a special case of — *"By
[...]

PR #870, opened 2018-10-23
polarize(MonomialIdeal) function, documentation update, and topcom build issue

lilysilverstein:
It was easiest for me to combine these three separate issues into a single pull request, hope that is okay.

1) Adding a function and documentation for polarizing a monomial ideal. This code is slightly modified from the Computations... book. It would be better for the function to be called "polarization" instead of "polarize," but this creates a conflict with the book. Is there a better way to resolve this than the name change?

2) Some recent documentation page rewrites, specifics are in the commit messages. Additionally, these should close issues #695 and #651 

3) Commented out a line from the topcom configure.ac problem which caused me to be unable to build from source. When someone has the software SoPlex installed, as I do, this line checks that SoPlex exists and then demands nonexistent files from an older version of SoPlex. Not finding them causes the entire build to fail. Simply removing the command seems to fix the problem.


DanGrayson:
Edit the file again, in the build directory corresponding to M2/libraries/topcom/ run "make diffs", take the resulting file and put the part of it that corresponds to your change into the file M2/libraries/topcom/patch-0.17.8 .  Test it with "make" after "make clean" in that same build directory.

> Oof, I just realized the Topcom configure file is generated during the build. Force committing the manually edited and usually git-ignored .ac file is surely wrong... What is the right way to address the problem?

DanGrayson:
We'll wait for you to let us know that the pull request is ready to consider once again.

lilysilverstein:
Thanks for your instructions, I changed the patch file as described and it
works on my computer. Please consider the pull request again

mikestillman:
@DanGrayson are you ok with accepting this pull request yet?

DanGrayson:
Mike, I'm busy with other things, so if it looks okay to you after testing, go ahead.

mikestillman:
Ok, compiling it all now!

mikestillman:
Thanks @lilysilverstein , the doc changes are very nice!

I'm going to accept the pull request, but first I'd like your thoughts on the following:

  One thing that might be nice to add to the "llists and sequences" page might be to also mention:
`for a in L list a+1`
`for a in L do print a`
and perhaps things like:
`for a in L list if a === null then continue else a`
Basically, the for loop over a list can be used instead of apply, scan, select, and is sometimes clearer in code. (break also works in here).  "/" and "\" are still very useful, and my favorites for when they apply.

Did you consider this, and decide it shouldn't be included there?  If so, fine, if not, this would be useful.  That is an incredibly nice page by the way, as are the others you have added.

We can either accept the pull request and change that later, or you can change that, and then I'll accept the pull request after, or tell me if shouldn't be there (but I think it would be good to be there, unless you have a good reason!).  Your choice.

mikestillman:
@DanGrayson , can you comment on this aspect of (the first comment on) this pull request:

Adding a function and documentation for polarizing a monomial ideal. This code is slightly modified from the Computations... book. It would be better for the function to be called "polarization" instead of "polarize," but this creates a conflict with the book. Is there a better way to resolve this than the name change?

DanGrayson:
I think we want the book to continue to work, and a name change is a good way to achieve that.

DanGrayson:
No, currently there is no function named "polarize" or "polarization".  One of the chapters of the book has code that defines a function called "polarization".  That code will break if M2 ever defines such
a function, because the assignment to the variable "polarization" will fail.

lilysilverstein:
@mikestillman Your suggestion is great, I simply overlooked it. I say accept the pull request as is (pending consensus on polarize), since I might not have time to work on documentation this week. Actually, the "lists and sequences" overview was part of my last pull request and is already in the current release. The reason it shows up here is that I reorganized some files by moving content to individual files in Macaulay2Doc/functions/. So the for...list description will have to be added after the fact anyway. I will put it on my to do list.

Regarding polarization, Mike has pointed out that a noun is usually preferred over a verb for a function like this, since a verb might misleadingly suggest that the function modifies its input. We could also consider other names for this function, like squarefreePolarization or idealPolarization. I have no strong feelings about the name, but went with polarize for brevity and auto-complete reasons. I'd love to find a better way to resolve the book/core name clash, but I haven't been able to think of one.

mikestillman:
I think polarize is fine.  Perhaps the documentation should also include the string "polarization" as a key pointing to this node.  But I think I'm going to just accept the request as is.

PR #878, opened 2018-11-05
Remove a bashism in configure.ac

kmarius:
The `function` keyword is not defined in POSIX. The build fails on systems where `/bin/sh` is linked to e.g. `/bin/dash` instead of `/bin/bash` or similar.


DanGrayson:
Thanks!

PR #888, opened 2018-11-13
Update SOS

tom111:
Hi. This branch updates the SOS and SemidefiniteProgramming packages to the latest versions and makes the Csdp makefile compatible with compilation on 32 Bit. As it turns out, one can just omit the -DBITS64 directive and then it works on both 64 and 32 Bit.


DanGrayson:
I think you should remove `-march=native -mtune=native -Ofast`, too.  Leave that to the builder of M2.

tom111:
I removed the architecture dependend flags and the duplicate exports.

PR #891, opened 2018-11-19
fix #875

### LINKED ISSUE #875: homogenize incorrectly keeps track of weights
```
i1 : R1=QQ[a,DegreeRank=>0]

o1 = R1

o1 : PolynomialRing

i2 : R2=R1[x,y]

o2 = R2

o2 : PolynomialRing

i3 : homogenize(1+x,y)
stdio:3:1:(3): error: homogenization: improper weight function
```
the answer should be `x+y`, of course. the problem is that the weights given to `rawHomogenize` should be the weights of the raw ring of `R2`, namely `{1,1,0}`, and not those of `R2`, which are only `{1,1}`.

pzinn:
it's all in the title. no new code added. I don't think it's worth a test, though I can add one if necessary.

### DIFF
--- a/M2/Macaulay2/m2/matrix2.m2
+++ b/M2/Macaulay2/m2/matrix2.m2
@@ -419,6 +419,8 @@ homogenize(Matrix, RingElement) := Matrix => (f,n) -> (
      wts := apply(generators(R, CoefficientRing=>ZZ), x -> first degree x);
      homogenize(f,n,wts))
 
+homogenize(RingElement, RingElement) := lookup(homogenize,Matrix,RingElement)
+
 homogenize(Module,RingElement) := Module => (M,z) -> (
      if isFreeModule M then M
      else subquotient(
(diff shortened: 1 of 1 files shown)


M2/Macaulay2/m2/matrix2.m2
```diff
@@ -419,6 +419,8 @@ homogenize(Matrix, RingElement) := Matrix => (f,n) -> (
      wts := apply(generators(R, CoefficientRing=>ZZ), x -> first degree x);
      homogenize(f,n,wts))
 
+homogenize(RingElement, RingElement) := lookup(homogenize,Matrix,RingElement)
```
DanGrayson: This would look better as
```
homogenize(RingElement, RingElement) := 
homogenize(Matrix, RingElement) := Matrix => (f,n) -> (
...
```

pzinn:
thanks, I knew this wasn't the nicest way to do it...

PR #894, opened 2018-11-23
Version 2.0; improved user interface 

samuellundqvist:
The main improvement is a better user interface, for instance allowing the user to perform lie multiplication of elements a and b in the Lie algebra by using the space operator.


pzinn:
you may want to specify in the title that this is a *package* upgrade.

PR #896, opened 2018-11-25
fix #877

### LINKED ISSUE #877: Unexpected behavior of random(List)
The following code applies random permutation to the given list, but not a _uniformly_ random permutation.
  
      random List := opts -> s -> (
           n := #s;
           if n <= 1 then return s;
           s = new MutableList from s;
           for i from 0 to n-1 do (
                j := random n;
                t := s#i ; s#i = s#j ; s#j = t;
                );
           new List from s)

"j := random n" is the culprit, as discussed [here](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle#Implementation_errors), paragraphs 2 & 3. I'm not sure whether this should be changed or mentioned in the documentation. Is there a reason to shuffle in this way?

pzinn:
`random List` now produces a uniformly random permutation.

before:
```
i1 : time tally(apply(1000000,i->random toList(0..2)))

     -- used 10.9433 seconds

o1 = Tally{{0, 1, 2} => 148460}
           {0, 2, 1} => 184949
           {1, 0, 2} => 184837
           {1, 2, 0} => 185435
           {2, 0, 1} => 147764
           {2, 1, 0} => 148555

o1 : Tally
```
after:
```
i1 : time tally(apply(1000000,i->random toList(0..2)))
     -- used 9.15754 seconds

o1 = Tally{{0, 1, 2} => 166470}
           {0, 2, 1} => 167373
           {1, 0, 2} => 167345
           {1, 2, 0} => 165955
           {2, 0, 1} => 166531
           {2, 1, 0} => 166326

o1 : Tally
```


DanGrayson:
Great, thanks!

Issue #902, opened 2018-12-27
hVector duplication


DanGrayson:
Commit 2a3806d7766c6ecc86d579914245124f8bc4c024 defined `hVector` in the package `SimplicialComplexes`, despite it already being defined in the package `SimplicialDecomposability`, thereby breaking `SimplicialDecomposability`.

@LorenzoVenturello -- is there a way to fix this in the next few days for the release of 1.13, or should I roll back the commit referred to above?

DanGrayson:
Meanwhile, I've renamed `hVector` to `hVector'` in `SimplicialComplexes` to avoid the conflict.  There must be a better solution.  See commit 6a207f32b9f39eb4c439c7e416459b68beb6f1e0

LorenzoVenturello:
As I see it there are three possible solutions. I can modify this function and add tests directly in the `SimplicialDecomposability` package and delete the one in `SimplicialComplexes`. I can simply delete `hVector'` from `SimplicialComplexes` (but in this way the current one will still give the wrong answer over multigraded rings). Finally one can even think of merging the two packages, but it is the most complicated solution and I'm not sure if I'm allowed to.

DanGrayson:
The simplest solution would be to have just one package with code for hVector in it, so I like the first solution.  You would also delete hVector from SimplicialComplexes, and any existing tests for hVector in SimplicialDecomposability would be retained and would have to work.

Can you do that?

DanGrayson:
... in the next few days?

LorenzoVenturello:
Yes, it shouldn't take too long. Should I raise another pull request when I'm done?

DanGrayson:
Yes, thanks!

PR #905, opened 2018-12-28
Function hVector moved to SimplicialDecomposability

LorenzoVenturello:
As pointed discussed in #902 I moved the modified hVector function from `SimplicialComplexes` back to the `SimplicialDecomposability` package to avoid conflicts. I moved the tests as well and they appear to be working.


LorenzoVenturello:
@DanGrayson

DanGrayson:
Thanks!

PR #906, opened 2018-12-28
Feature: MultigradedBettiTally type and (multigraded,BettiTally) function

mahrud:
This PR adds a new `MultigradedBettiTally` type that inherits from `BettiTally` and implements `multigradedBetti` which pretty-prints that the multigraded betti tally of an input. The compact format, along with the documentation, was written by Greg Smith.

The definitions are added in `m2/chaincomplexes.m2` immediately after the definition of `betti` for two reasons:
* two required helper functions used by `betti` are also used by `multigradedBetti`;
* `multigradedBetti` can be useful across multiple packages, so it seemed like a reasonable idea to have it exported by default.

The only remaining question is where to add the documentation. I thought about adding it after `packages/Macaulay2Doc/functions/betti-doc.m2` in the old style. Would that be okay, or should I make a new file and use the new style?

ps: the second commit fixes a small error in documentation of the `Points` package.


DanGrayson:
Did you notice that the existing `betti` function already returns Betti tallies that contain all the multigraded information you need?  Here is an example.
```

i1 : R = QQ[x,y,Degrees => {{1,0}, {0,1}}]

o1 = R

o1 : PolynomialRing

i2 : betti res coker vars R

            0 1 2
o2 = total: 1 2 1
         0: 1 2 1

o2 : BettiTally

i3 : peek oo

o3 = BettiTally{(0, {0, 0}, 0) => 1}
                (1, {0, 1}, 1) => 1
                (1, {1, 0}, 1) => 1
                (2, {1, 1}, 2) => 1
```
The second member of each key is the multidegree.

DanGrayson:
What does the display of the new multigraded tallies look like?

mahrud:
Sorry, I should have included at least one examples:
```
i14 : multigradedBetti C

         0         1         2    3
o14 = 0: 1         .         .    .
      2: . a2+2ab+b2         .    .
      3: .         . 2a2b+2ab2    .
      4: .         .         . a2b2

o14 : MultigradedBettiTally
```
A more compact form is also possible: (it doesn't look that different in this case, but in larger cases it's very noticeable)
```
i15 : compactMatrixForm=false

o15 = false

i16 : multigradedBetti C

             0        1        2        3
o16 = 1:{0, 0} 1:{0, 2} 2:{1, 2} 1:{2, 2}
               1:{2, 0} 2:{2, 1}         
               2:{1, 1}                  
                                         

o16 : MultigradedBettiTally
```
Most of the added code is the pretty-printing step and just defining the new function hooks. The internals of the hash table are the same.
```
i17 : peek multigradedBetti C

o17 = MultigradedBettiTally{(0, {0, 0}, 0) => 1}
                            (1, {0, 2}, 2) => 1
                            (1, {1, 1}, 2) => 2
                            (1, {2, 0}, 2) => 1
                            (2, {1, 2}, 3) => 2
                            (2, {2, 1}, 3) => 2
                            (3, {2, 2}, 4) => 1

i18 : betti C

             0 1 2 3
o18 = total: 1 4 4 1
          0: 1 . . .
          1: . 4 4 1

o18 : BettiTally

i19 : peek betti C

o19 = BettiTally{(0, {0, 0}, 0) => 1}
                 (1, {0, 2}, 2) => 1
                 (1, {1, 1}, 2) => 2
                 (1, {2, 0}, 2) => 1
                 (2, {1, 2}, 3) => 2
                 (2, {2, 1}, 3) => 2
                 (3, {2, 2}, 4) => 1
```

DanGrayson:
I ran it and got this example of a new display:
```

i4 : multigradedBetti C

        0   1  2
o4 = 0: 1   .  .
     1: . a+b  .
     2: .   . ab
```

DanGrayson:
It seems that the new type contains exactly the same information as the old type of Betti tally:
```

o4 : MultigradedBettiTally

i5 : peek o3

o5 = BettiTally{(0, {0, 0}, 0) => 1}
                (1, {0, 1}, 1) => 1
                (1, {1, 0}, 1) => 1
                (2, {1, 1}, 2) => 1

i6 : peek o4

o6 = MultigradedBettiTally{(0, 0, {0, 0}) => 1}
                           (1, 1, {0, 1}) => 1
                           (1, 1, {1, 0}) => 1
                           (2, 2, {1, 1}) => 1
```

DanGrayson:
A beneficial change may be to arrange for there to be a function that converts a BettiTally to a MultigradedBettiTally with exactly the same contents, and then `multigradedBetti` could be eliminated in favor of the user applying that function to the result of `betti`.  If that function were called `multigraded`, it wouldn't even be more typing.

DanGrayson:
Thanks.  I think it doesn't much matter where you add the documentation now, or whether it's new or old format.

mahrud:
Sorry for the delay! I missed the notification.

DanGrayson:
Release 1.13 is well under way, but your changes will be in 1.14.

DanGrayson:
This bit causes an error:
```
    EXAMPLE lines ///
      B = new MultigradedBettiTally from {(0, {0, 0}, 0) => 1, (1, {0, 2}, 2) => 1, (1, {1, 1}, 2) => 2,
	  (1, {2, 0}, 2) => 1, (2, {1, 2}, 3) => 2, (2, {2, 1}, 3) => 2, (3, {2, 2}, 4) => 1}
      peek oo
    ///,
```
This you could have tested by building Macaulay2's documentation.

The problem is that `lines` divides the string up into three examples, but it's really two.

The error message is this:
```
currentString:1:1:(3):[2]: error: example results terminate prematurely: MultigradedBettiTally
```

mahrud:
Ahh, thank you! I tried building it, which helped me fix various other errors, but couldn't figure out this one.

mahrud:
@DanGrayson I think this PR is ready. I rebased and tested it again.

M2/Macaulay2/packages/Points.m2
```diff
@@ -737,16 +737,16 @@ doc ///
...
+    n:ZZ
+     number of points
    Outputs
     i:Ideal
      ideal of the random points
    Description
     Text
      The script defines a ring R with r+1 variables, and calls
-     points(R,randomPointsMat(r,n))
+     points(R,randomPointsMat(R, n))
```
DanGrayson: Should the second `R` be `r`?

mahrud: Definitely `R`,  the first input to `randomPointsMat` is a ring:
https://github.com/Macaulay2/M2/blob/73e7016b27dd79d4c115b4821f5a6ab05ca94212/M2/Macaulay2/packages/Points.m2#L582-L587

DanGrayson:
Thanks!

PR #908, opened 2018-12-29
Update Matroids.m2 to 0.9.7

jchen419:


M2/Macaulay2/packages/Matroids.m2
```diff
@@ -285,16 +287,10 @@ flats Matroid := List => M -> (
...
-sort (List, Function) := opts -> (L, f) -> (
-	H := hashTable((a, b) -> (a, b), apply(L, l -> f(l) => l));
-	deepSplice join apply(sort keys H, k -> H#k)
-)
-
 latticeOfFlats = method()
 latticeOfFlats Matroid := Poset => M -> poset(flats M/toList, (a, b) -> isSubset(a, b))
 
-fVector Matroid := HashTable => opts -> M -> hashTable pairs tally(flats M/rank_M)
+fVector Matroid := HashTable => M -> hashTable pairs tally(flats M/rank_M)
```
DanGrayson: `fVector` was recently changed to take options, so this change will break your code.  Are you up to date with the master branch and running M2 as built from source code?

DanGrayson:
I fixed it ...

jchen419:
Hi Dan, thanks for the fix - I had been running the latest release version (1.12), but in the future I will build from source. The issue with options for fVector did seem a bit strange (as I previously had options enabled, but then had to remove them, which now apparently should be reverted...). In any case this update isn't particularly time-sensitive, so it's fine if it doesn't make it in 1.13. Thanks again!

DanGrayson:
It will be in 1.13.

PR #913, opened 2019-01-22
minor updates to NormalToricVarieties

ggsmith:
Fixed issues with expression (NormalToricVariety); adjusted normalToricVariety (Polyhedron) to handle the non-full-dimensional case; added documentation node discussing various ways to construct projective space.


DanGrayson:
Thanks!

PR #920, opened 2019-01-27
fix RingZZ::lower_content()

### LINKED ISSUE #740: negative signs not cancelled in frac ZZ[...]
compare
```
i1 : R=QQ[x]

o1 = R

o1 : PolynomialRing

i2 : -1/-x

     1
o2 = -
     x

o2 : frac(R)
```
with
```
i3 : R=ZZ[x]

o3 = R

o3 : PolynomialRing

i4 : -1/-x

     -1
o4 = --
     -x

o4 : frac(R)
```
why aren't the - signs removed? the code (frac.cpp) claims that it divides by the "content", which is supposed to take care of the sign, but I don't see it happening.

radoslavraynov:
Makes ringZZ::lower_content() parallel the M2::QQ variant. In particular, this fixes #740.


DanGrayson:
Looks good to me.

PR #924, opened 2019-02-02
better degree inference in matrix concat

### LINKED ISSUE #607: concatenation of homogeneous matrices of nonzero degree does not preserve degree
Concatenation is not respecting degrees of matrices.
```
  S = ZZ/101[a..d]
  m = map(S^1, S^1, {{a^2}}, Degree=>2)
  isHomogeneous m
  isHomogeneous(m | m)
  isHomogeneous(m || m)
  degree m
  degree (m|m) -- should be {2}
  assert(degree m == degree (m|m))
  assert(degree m == degree(m||m))
```

### LINKED ISSUE #736: concatenation of homogeneous matrices with degrees
Do it better:
```

i1 : R=QQ[x];

i2 : f = matrix {{x}}; g = x * id_(R^1);

             1       1
o2 : Matrix R  <--- R

             1       1
o3 : Matrix R  <--- R

i4 : isHomogeneous f, isHomogeneous g

o4 = (true, true)

o4 : Sequence

i5 : f|g

o5 = | x x |

             1       2
o5 : Matrix R  <--- R

i6 : isHomogeneous oo

o6 = false
```

radoslavraynov:
This should fix #736 and fix #607.

The interesting part of the first commit (37c8d97) is a bug where `deg` doesn't get computed because of short-circuit evaluation. The second commit (cb4d946) adjusts the twisting of the sources to make the resulting map homogenous of degree `deg mats#0`.

---

For example, after the following snippet is executed,
```
S = ZZ[x,y]
f = matrix{{x}}
g = x*(id_(S^1))
h = map(S^2,S^{1},matrix{{x},{y}},Degree=>{2})
bar = (op1,op2) -> ({op1,op2,op1|op2} / (op->print(degree op, isHomogeneous(op), degrees op));)
```

the output of `bar(h,h)`, shown below, looks better after the change.
```
({2}, true, {{{0}, {0}}, {{-1}}})
({2}, true, {{{0}, {0}}, {{-1}}})
({0}, false, {{{0}, {0}}, {{-1}, {-1}}})     <-- before
({2}, true, {{{0}, {0}}, {{-1}, {-1}}})      <-- after
```

same for the output of `bar(f,g)`
```
({0}, true, {{{0}}, {{1}}})
({1}, true, {{{0}}, {{0}}})
({0}, false, {{{0}}, {{1}, {0}}})            <-- before
({0}, true, {{{0}}, {{1}, {1}}})             <-- after
```

and `bar(g,f)`
```
({1}, true, {{{0}}, {{0}}})
({0}, true, {{{0}}, {{1}}})
({0}, false, {{{0}}, {{0}, {1}}})            <-- before
({1}, true, {{{0}}, {{0}, {0}}})             <-- after
```

Also, the last line in the second commit can be optimized if one can assume the sources are never non-free.


mikestillman:
Radoslav,

This looks good, thanks!.  Perhaps it would be good to place a test for this behavior into a file in Macaulay2/packages/Macaulay2Doc/tests, also including a test indicating that the 2 issues mentioned (#736 , #607 ) are fixed, and including an assertion based test for the 'bar' example you have above?. (This could be one file).  Can you add that?  If not, I can do so.

radoslavraynov:
Definitely! I’ll do that tonight.

radoslavraynov:
I am adding this as a separate unit test under `packages/Macaulay2Doc/test/` although it can probably safely go inside the matrix.m2 one. Let me know if that works.

radoslavraynov:
On the other hand, I noticed that `make check` doesn't work for me&mdash;not even out of `master`. It's probably just my build though. Either way, let me know!

DanGrayson:
> On the other hand, I noticed that `make check` doesn't work for me—not even out of `master`. It's probably just my build though. Either way, let me know!

Show us the error message.

radoslavraynov:
I think my error boils down to the following: building `master` (same as `upstream master`) and running `/.../usr-dist/x86_64-Darwin-MacOS-10.14.1/bin/M2-binary` in packages/Macaulay2Doc/tests, I get
```
Macaulay2, version 1.13.0.1
../../../m2/robust.m2:98:26:(1):[18]: error: no method for adjacent objects:
--            integralClosure (of class Symbol)
--    SPACE   R (of class QuotientRing)
../../../m2/robust.m2:98:26:(1):[18]: --entering debugger (type help to see debugger commands)
../../../m2/robust.m2:98:26-98:85: --source code:
                         error toString stack(line1,commentGuard line2,commentGuard line3))))));

i1 : 
```

Same for the script version. The program runs fine if started from any other dir I have tested. Running with `--no-preload` makes it OK, so somehow when run from that exact dir, `integralClosure` is not defined.

mikestillman:
The way I run the tests is to do ‘make check’ (or ‘make -k check’) in the build tree.  This runs far more tests than are in that directory.  I’m not sure how to just run those tests…

— mike

radoslavraynov:
I get the same error when `make` hits that dir &mdash; 
```
...
gmake -C packages/Macaulay2Doc/test check 
gmake[1]: Entering directory '/Users/Radoslav/Projects/Macaulay2/m2-new/M2/Macaulay2/packages/Macaulay2Doc/test'
testing: file-aluffi.m2
../../../m2/robust.m2:98:26:(1):[18]: error: no method for adjacent objects:
IntegralClosure.m2:5:10:(0):[17]: --back trace--
../../../m2/setup.m2:113:27:(1):[17]: --back trace--
../../../m2/setup.m2:106:18:(1):[16]: --back trace--
../../../m2/setup.m2:126:41:(1):[15]: --back trace--
../../../m2/setup.m2:129:6:(1):[14]: --back trace--
../../../m2/packages.m2:62:6:(1):[13]: --back trace--
../../../m2/methods.m2:119:80:(1):[12]: --back trace--
../../../m2/option.m2:15:8:(1):[11]: --back trace--
../../../m2/packages.m2:87:11:(1):[10]: --back trace--
../../../m2/methods.m2:119:80:(1):[9]: --back trace--
../../../m2/option.m2:15:8:(1):[8]: --back trace--
../../../m2/packages.m2:304:6:(1):[6]: --back trace--
../../../m2/methods.m2:119:80:(1):[5]: --back trace--
../../../m2/option.m2:15:8:(1):[4]: --back trace--
../../../m2/last.m2:44:16:(1):[3]: --back trace--
../../../d/startup.m2.in:74:58:(0):[2]: --back trace--
../../../d/startup.m2.in:74:32:(0):[1]: --back trace--
file-aluffi.errors:0: error output left here for the errors above:
gmake[1]: *** [Makefile.test:67: file-aluffi.out] Error 1
gmake[1]: Leaving directory '/Users/Radoslav/Projects/Macaulay2/m2-new/M2/Macaulay2/packages/Macaulay2Doc/test'
gmake: *** [Makefile:20: check-in-packages/Macaulay2Doc/test] Error 2
```

DanGrayson:
I fixed that problem, see 1a8e3f4cbd8703907487dbcfffb618f0d594baec

radoslavraynov:
Thanks! I moved the twisting of the source higher up, because `concatCols` is pretty low level and I didn't want to mess up engine-interfacing code (e.g. polynomial ring degrees formation).

I ran all tests (`make check` in M2/Macaulay2) but the change causes several packages' tests to fail.

* `HyperplaneArrangements`: very benign. test checks then witnessed behavior, that is, the degrees of the source.
* `MultiplierIdeals`: the assertion on line 1649 seems to take too much time or memory. This can perceivably be caused by the tensoring and can be fixed if I can call shift on the raw free module of the source instead.
* `VectorFields`: much like in the first case, checks expect same degrees of the sources in certain maps.

So, at this point I can just put this off and may be come back around another time? In general, how is this handled? That is, if a package tests or even uses what is witnessed and not what is documented or intended in the language, and that behavior later changes.

DanGrayson:
If the change is a desirable one, we change the bits of code that break when it happens.  Mike, do you agree that we want this change?

DanGrayson:
If you want to volunteer to fix it, I'll assign the task to you, gladly.

Issue #874, opened 2018-11-01
crash in computation of inverse


pzinn:
Macaulay2 fails to compute the inverse of this invertible element (and crashes):
```
i1 : R=(frac(QQ[z]))[q]/ideal(q^10)

o1 = R

o1 : QuotientRing

i2 : (z+q)^(-1)
--internal error: ring element gcd computation failed
/bin/sh: line 1:  3646 Aborted                 (core dumped) M2 --no-readline --print-width 115

Process M2 exited abnormally with code 134

``

**Comment by radoslavraynov** [comment]
Hi, inverse should never work on a non-scalar, right? Instead, you can represent `1_R` in that basis.
```
i1 : R=(frac(QQ[z]))[q]/ideal(q^10); inv=1//matrix{{z+q}}

o3 = {1} | (-1)/z10q9+1/z9q8+(-1)/z8q7+1/z7q6+(-1)/z6q5+1/z5q4+(-1)/z4q3+1/z3q2+(-1)/z2q+1/z |

             1       1
o3 : Matrix R  <--- R

i4 : assert((z+q)*inv_(0,0)===1_R)
```

The internal error and crash are unwanted. The way this is handled gets fixed in [#928](https://github.com/Macaulay2/M2/pull/928) &mdash;
```
i1 : R=(frac(QQ[z]))[q]/ideal(q^10); (z+q)^(-1)
stdio:5:38:(3): error: division is not defined in this ring
```

DanGrayson:
Despite #928 being pulled, this is still a problem.

radoslavraynov:
I think it's be because I reverted 0d26ecb. I'll rework it this weekend.

mikestillman:
I made a change to the C++ code today which might have fixed this, although your solution might be more flexible.

DanGrayson:
Saturday morning I start making the next release, version 1.14.

pzinn:
If nobody objects I will close this issue. AFAICT the original bug is resolved.

DanGrayson:
Okay, closing, thanks!

Issue #657, opened 2017-07-21
exponentiation is too slow


kschwede:
Hi,

I know this has been suggested before, but exponentiation of ring elements (ie, just f^N where f is a polynomial) is much slower than it can be in positive characteristic.  For example in the Fsing/PosChar package we have the following code which is much much faster.  I would suggest that something like this be done for Macaulay2 core since the speed increase is so large.

```
fastExponentiation = method( TypicalValue => RingElement )

fastExponentiation ( ZZ, RingElement ) := RingElement => ( N, f ) ->
(
    if N < 0 then error "fastExponentiation: first argument must be a polynomial over a nonnegative integer.";
    if char ring f == 0 then 
        error "fastExponentiation: second argument must be a polynomial over a field of positive characteristic.";
    p:=char ring f;
    E:=adicExpansion(p,N);
    product(#E, e -> sum( terms f^(E#e), g -> g^(p^e) ) )
)

adicExpansion = method( TypicalValue => List ); 

--Computes the terminating base p expansion of a positive integer.
--Gives expansion in reverse... so from left to right it gives
--the coefficient of 1, then of p, then of p^2, and so on

adicExpansion( ZZ, ZZ ) := List => ( p, N ) ->
(
    if p <= 0 then error "adicExpansion: Expected first argument to be positive";
    if N < 0 then error "adicExpansion: Expected second argument to be nonnegative";
    if N < p then { N } else prepend( N % p, adicExpansion( p, N // p ) ) 
    -- would this be faster if it were tail-recursive? we could do this w/ a helper function.
)

```
For example, 

```
i5 : R = ZZ/3[x,y,z]; f = random(4, R) + random(3, R)
i7 : time fastExponentiation(21, f);
     -- used 0.00179836 seconds

i8 : time (f^21);
     -- used 0.218102 seconds
```

DanGrayson:
Thanks, indeed, this would be good to do.  

This issue is subsumed by #264, so closing.

PR #928, opened 2019-02-11
Fast exponentiation in prime characteristic

### LINKED ISSUE #264: fast powers in char p
should be implemented in the engine.  I.e., do f^p by additivity and do f^n by expanding n in base p.

Hat tip to PosChar authors, see https://github.com/Macaulay2/Workshop-2014-Berkeley/blob/master/F-sing/PosChar.m2

radoslavraynov:
This fixes #264.

I've moved the actual logic from PolyRing::power() into a new PolyRing::power_direct(), making the former a dispatcher based on the characteristic being 0 or not. In particular, the overhead in zero-characteristic is a single function call.

In the way of benchmarking, below is the average runtime of raising to power a given random polynomial `f` using engine and high-level code (for example, as in #657). The `(p,{n1,n2,...},k)` col spec describes a test case, in which `f` is the sum of random forms in `ZZ/p[x,y,z]` of degrees `n1,n2,...` and `k` is the exponent. ~~I am not sure why~~ the high-level code before is sometimes *slightly* faster than the engine code after the change.

```
               (5,{4, 3},10), (17,{5},4), (19,{10},21), (7,{5, 6},31), (11,{3},51)
[engine-before] .0698586,    .000479839, 2.37298,     26.8513,       .347187,
[hi-lvl-before] .000392648,  .00162537,   .00630307,    .379307,     .0127448
[engine-after ] .0000616816, .000492968,  .0092006,     .487706,     .0149149
[hi-lvl-after ] .000453994,  .00159426,   .0100856,     .500135,     .0171947
```

### DIFF
--- a/M2/Macaulay2/e/poly.cpp
+++ b/M2/Macaulay2/e/poly.cpp
@@ -846,6 +846,40 @@ ring_elem PolyRing::power(const ring_elem f0, mpz_t n) const
   return result;
 }
 
+ring_elem PolyRing::power_direct(const ring_elem f, int n) const
+{
+  ring_elem result = from_long(0);
+  Nterm *lead = f;
+  if(!lead) return result;
+  ring_elem tail = lead->next;
+
+  mpz_t binom_c;
+  mpz_init_set_si(binom_c, 1);
+  int *m = M_->make_one();
+  ring_elem c1, c2, c3;
+  ring_elem rest=from_long(1), temp;
+
+  for(int i=0; i<=n; ++i)
+  {
+    c1 = K_->from_int(binom_c);
+    c2 = K_->power(lead->coeff, n-i);
+    c3 = K_->mult(c1, c2);
+    M_->power(lead->monom, n-i, m);
+    temp = mult_by_term(rest, c3, m);
+    add_to(result, temp);
+
+    if(!tail) break;
+
+  // maintenance
+    mpz_mul_si(binom_c, binom_c, n-i);
+    mpz_fdiv_q_ui(binom_c, binom_c, (unsigned)(i+1));
+    temp = mult(rest, tail);
+    rest = temp;
+  }
+
+  return result;
+}
+
 ring_elem PolyRing::power(const ring_elem f0, int n) const
 {
   ring_elem ff;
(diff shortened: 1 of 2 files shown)


DanGrayson:
Thanks!  I'll let Mike take a look at this.

radoslavraynov:
I just noticed that the optimization takes place only if the ring is syntactically `ZZ/p[...]`. Calculations in `ZZ[x,y]/5` or `ZZ/5[x,y,z]/z`, for instance, fall back to Ring::power(const ring_elem, mpz_t), which is implemented in the typical fashion. So I am not sure how useful the optimization would be without rewiring the power() methods in the related classes. Either way, let me know what you think when you get the chance.

DanGrayson:
I think we don't have to worry about rings such as `ZZ[x,y]/5`.  We would also like it to work for rings over fields such as GF 81.

At top level, polynomial rings over finite fields have the characteristic stored under the key `symbol char`, by the way.

radoslavraynov:
I pushed another commit addressing the issues below. I'll take another look over the weekend and then leave it in that state for Mike to review when he gets the chance.

I limited Ring::power() to `int`. This is the behavior for all rings except for the number rings and it resulted in the following inconsistency before.
```
S=ZZ[x]; (x+1)^2147483648;       -- throws stdio:2:15:(3): error: exponent too large
S=ZZ[x,y]/y; (x+1)^2147483648;   -- runs for a very long time
```

PolyRingQuotient::power() now uses PolyRing::power(), so the optimization takes place for `K[...]/I` for `K` being `ZZ/p` or `GF p^k` and `I` being trivial or not. For example, on my build
```
S=ZZ/5[x,y] -- or ZZ/5[x,y,z]/z or (GF 125)[x,y]
f=(x+y)
tot = 0.0;
for i from 1 to 100 do tot = tot + (timing f^5001)#0;
tot / 100
```
consistently gives 0.00001s instead of the 0.03987s before.

radoslavraynov:
Incidentally, this also fixes #874.

radoslavraynov:
I reverted the last commit (0d26ecb70b0fcc1c4bff58e8a2be837276ceae82) because it was too invasive. I will reimplement those changes later.

mikestillman:
This pull request looks good, and really speeds up exponentiation over finite fields.  I will add a test that I wrote for it after I accept the pull request.

PR #929, opened 2019-02-11
adding a top-level printing method

radoslavraynov:
I would like to add a top-level printing method to use in the Jupyter kernel and possibly at other places. For example, it allows for using a simple finite-state machine to semantically identify the output &mdash; as input, STDOUT, value or class. It's modeled after the [WebApp](https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.13/share/doc/Macaulay2/Macaulay2Doc/html/___Web__App.html) mode.


M2/Macaulay2/m2/jupyter.m2
```diff
@@ -0,0 +1,28 @@
...
+ZZ#{Jupyter,InputPrompt} = lineno -> concatenate(concatenate("[INP]",newline,"[INP]"), lastprompt = concatenate(interpreterDepth:"i", toString lineno, " : "));
+ZZ#{Jupyter,InputContinuationPrompt} = lineno -> concatenate("[INP]",#lastprompt);
+setJupyter = () -> ( ZZ#{Standard,InputPrompt} = ZZ#{Jupyter,InputPrompt}; ZZ#{Jupyter,InputPrompt} = x->print("hi") )
+
+Thing#{Jupyter,Print}   = x -> ( << "[VAL]" << endl; Thing#{Standard,Print}(x) )
+Nothing#{Jupyter,Print} = identity
+
+----------------------
+-- why doesn't the following work!?
+-- for cls in {Thing, Nothing, ...} do (cls)#{Jupyter,AfterPrint} = (cls)#{Standard,AfterPrint}
```
DanGrayson: In what sense does this line not "work"?

radoslavraynov: For example, after
```
for cls in {Thing, Nothing, Boolean, ZZ, Matrix} do
    (cls)#{Jupyter,AfterPrint} = x -> ( << "[CLS]" << endl; (cls)#{Standard,AfterPrint} x );
```
and starting the program with `./M2-binary -e "topLevelMode=Jupyter"`, I get
```
[INP]
[INP]i1 : 501
[VAL]

o1 = 501
[CLS]

o1 : Matrixstdio:1:1:(3): error: after print: no method found for applying target to:
     argument   :  501 (of class ZZ)
[INP]
[INP]i2 : 
```

radoslavraynov: On the other hand, I leave out `Matrix` but add `Ideal`, the output becomes
```
[CLS]

o1 : Ideal of ZZ
```
So it seems that somehow only the last assignment gets registered.

DanGrayson:
Your problem is that `cls` is a global variable, whose value will be obtained too late.  Do something like this:
```
for cls in {Thing, Nothing, Boolean, ZZ, Matrix} 
do cls#{Jupyter,AfterPrint} = 
  (cls -> x -> (<< "[CLS]" << endl; cls#{Standard,AfterPrint} x)) cls
```

radoslavraynov:
Thanks, that worked!

DanGrayson:
Good!

PR #930, opened 2019-02-21
Ignore more auto-made files

radoslavraynov:
This is extremely low-impact. There are no gitignore entries for the Makefiles for *cohomcalg*, *csdp* and *topcom*. Also, on some systems one has `.gdbinit.USERNAME` instead of just `.gdbinit`.

Shouldn't the items (at least within each group) be alphabetically sorted?

### DIFF
--- a/M2/.gitignore
+++ b/M2/.gitignore
@@ -47,6 +47,9 @@
 /Macaulay2/c/.gdbinit
 /Macaulay2/m2/.gdbinit
 /libraries/atomic_ops/Makefile
+/libraries/cohomcalg/Makefile
+/libraries/csdp/Makefile
+/libraries/topcom/Makefile
 /libraries/M2/Makefile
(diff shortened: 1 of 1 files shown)


M2/.gitignore
```diff
@@ -44,9 +44,12 @@
 /libraries/missing/src/Makefile
 /libraries/Makefile.library
 /Macaulay2/bin/M2
-/Macaulay2/c/.gdbinit
-/Macaulay2/m2/.gdbinit
+/Macaulay2/c/.gdbinit*
+/Macaulay2/m2/.gdbinit*
```
DanGrayson: We don't want to ignore M2/Macaulay2/m2/.gdbinit.dan and M2/Macaulay2/m2/.gdbinit.mike

radoslavraynov: I see. I reverted that.

PR #947, opened 2019-05-21
Add new package FGLM

dylanpeifer:
This PR adds a new package which implements the FGLM algorithm for computing Groebner bases. Mahrud Sayrafi (@mahrud) and I put this together last semester and plan on continuing work with it in the fall, including moving most of the code into the engine. For now, the interface is settled and the package can already beat `gb` on a variety of standard benchmarking examples, which we felt made it suitable for an initial submission.


DanGrayson:
Thank you for the contribution.

For the future: add an explanation of the cryptic acronym FGLM.  And think of a better name for the package.  Such cryptic acronyms may not help users find what they need.

PR #949, opened 2019-05-23
Fix bug in isVeryAmple, add section in documentation to clarify conventions

lkastner:
The bugfix is in 6f6875e7b64318bb7d9558d585e68e40b3e39e68 in line 45 of `M2/Macaulay2/packages/Polyhedra/extended/polyhedron/properties.m2`, it replaces the old constructor by the new one that should be used now. Otherwise just variables got renamed.

Furthermore:
- The new section of the documentation is in `M2/Macaulay2/packages/Polyhedra/documentation/v_h_representation.m2`
- I added a command `\cdot` to `M2/Macaulay2/m2/texhtml.m2`
- I referenced this section at hopefully useful spots throughout the documentation
- Added references packages `FourTiTwo` and `Topcom` in the introductory section

I did not include the bugfix for `installPackage` in order to avoid conflicts. Please tell me if there is anything to fix.


DanGrayson:
Thanks.

By the way, you have 413 separate test items, which means that M2 has to be started 413 times, and that takes a lot of time.  You could speed up the tests considerably by combining them.

PR #960, opened 2019-06-16
Correction of an error in the package

DiffAlg:
It seems that the line 5 of the package had a wrong definition of the imaginary unit i. Now it is corrected.

### DIFF
--- a/M2/Macaulay2/packages/DiffAlg.m2
+++ b/M2/Macaulay2/packages/DiffAlg.m2
@@ -1,9 +1,13 @@
 -- -*- coding: utf-8 -*-
 
+i := getSymbol "i"
+QQi := QQ(monoid [i])
+F := toField (QQi / (QQi_0^2+1))
+
 newPackage (
 	"DiffAlg",
-	Version => "1.5",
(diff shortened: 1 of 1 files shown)


DanGrayson:
Thanks for the correction.  I wonder why no test failed as a result.

DiffAlg:
Hi Dan.
It seems that the problem it is still there. How can we fix it?!
It is on line 5. Where it says: F := toField (QQi / QQi_0^2)
it should say: F := toField (QQi / QQ(i_0^2+1))

DiffAlg:
Ok, no. Here on github it is everything fine. It should be updated in the web page of Macaulay2.

DanGrayson:
Which web page?

DiffAlg:
http://www2.macaulay2.com/Macaulay2/Packages/

DanGrayson:
I think in commit 81eff478f5715d250645bbf177eb481fb7372377 you made the desired change, and then in commit 6de1e663ce9c7876991cf6b16d5a9219200285d5 you mysteriously undid it.  So make the change again and submit a pull request.

PR #970, opened 2019-07-26
updated Tropical.m2 and gfanInterface.m2 to work with recent polymake…

kathlenkohn:
… versions

Signed-off-by: Kathlen Kohn <kathlen.korn@gmail.com>


DanGrayson:
Thanks.

By the way, I can't install polymake under the latest Mac OS because polymake insists on an older version of perl.  Strange.

PR #988, opened 2019-08-12
Added nanosleep function.

kschwede:
Hi, myself and Boyana Martinova (bmartinova18) have added a nanosleep function.  

It seems to work on our systems.


DanGrayson:
Thank you.  The code looks good and it seems to work.  Could you add some documentation and link to it from appropriate places?

DanGrayson:
Something like this would be a good example for the documentation:
```

i10 : elapsedTime nanosleep 500000000
     -- 0.503543 seconds elapsed

o10 = 0
```

bmartinova18:
We have updated it with documentation. Let us know if this works.

DanGrayson:
Thanks!

PR #1002, opened 2019-09-16
Update DiffAlg.m2

DiffAlg:
Ok, now it should be everything fixed.


DanGrayson:
Hmm?  What are you trying to fix?  All you do here is to remove one blank line.

DiffAlg:
Well, I told you that here on GitHub it's everything ok. But it would be great if the code is updated in the web page of Macaulay2, in the package section.

DanGrayson:
Please look at https://github.com/Macaulay2/M2/blob/master/M2/Macaulay2/packages/DiffAlg.m2 and tell me which lines are wrong.

DiffAlg:
They are all ok.

But here 
https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.14/share/Macaulay2/DiffAlg.m2
it's not ok. It should be updated there.

DanGrayson:
Oh, well, that's not going to get updated, since 1.14 was released months ago.  You can look forward to the code being correct in 1.15, in December.  Perhaps notify users publicly on the google group that they should download a more recent version from somewhere rather than using the one in 1.14.

PR #1003, opened 2019-09-16
fixed IntegralClosure

### LINKED ISSUE #933: Integral closure produces wrong result
I am trying to compute the integral closure of a monomial ideal using the code below.
```
S=QQ[a..f]
I=ideal(a*b*c,a*d*f,c*e*f,b*e*d)
integralClosure(I)
```
Macaulay2 returns ideal () which is incorrect.

eisenbud:
resolves #933


DanGrayson:
Thanks.  I've also added a test to make sure the bug stays fixed.

PR #1029, opened 2019-10-11
add List !

FatemehTarashi:
add List ! function and document


DanGrayson:
Thank you for the contribution, but it represents a departure from standard mathematical notation, so doesn't seem admissible in the core of Macaulay2.  Perhaps this is something individual users can simply put in their personal `init.m2` files, if they wish.

FatemehTarashi:
It was nothing. I saw (missing documentation) on the factorial doc page, so I added it. I don't think you can solve this missing without adding List! function, but do whatever you feel is fit. Thank you for the Macaulay2 project. :)

DanGrayson:
Oh, well that missing documentation is a real issue, I've never noticed that before.  But there is no method for `List !`, so how did that entry get generated?

Thank you for bringing it to our attention!

PR #1034, opened 2019-10-23
Adding the package SpechtModule to Macaulay2.

janino937:
This package implements and algorithm for calculating efficiently secondary invariants of permutation matrices. The algorithm uses representation theory of the symmetric group and for that reason it includes implementations of character tables, Young tableaux, Specht polynomials and Specht modules among other things. All of this functionality is also exported so that it can be used in other types of problems different from the one above.


DanGrayson:
Please add it also to the list of packages.

PR #1038, opened 2019-10-25
Updating various packages

kschwede:
This adds the package:
FastLinAlg
which is an experimental package offering some linear algebra functionality over function fields which sometimes speeds up certain computations.  For instance, when computing whether a ring or variety is R_n (regular in codimension n).

This also updates:
RationalMaps 
Seminormalization
TestIdeals
FThresholds

to the latest versions.

**EDIT:** Perhaps I should also note that this includes updates from the refereeing process on most of these packages.

### DIFF
--- a/M2/Macaulay2/packages/=distributed-packages
+++ b/M2/Macaulay2/packages/=distributed-packages
@@ -135,6 +135,7 @@ Visualize
 EquivariantGB
 ExampleSystems
 RationalMaps
+FastLinAlg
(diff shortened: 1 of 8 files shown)


kschwede:
Hi Mike, Dan, I think I fixed the conflict on my end.

DanGrayson:
We should convince @mikestillman that "FastLinearAlgebra" can go away so you can give your package a better name.

PR #1046, opened 2019-11-13
VirtualResolutions v1.2.0

mahrud:
This PR makes the following changes in VirtualResolutions package:
- rename findGensUpToIrrelevance to idealSheafGens and change the output type
- adds a UseSyzygies strategy to virtualOfPair
- changes the input types of isVirtual
- various improvements to documentation nodes
- fixes a previously commented test after fixing a small indexing issue in TateOnProducts, allowing correct computation of tateResolutions for products of three or more projective spaces.

For full commit history see https://github.com/Macaulay2/Workshop-2018-Madison/commits/Virtual

### DIFF
--- a/M2/Macaulay2/packages/TateOnProducts.m2
+++ b/M2/Macaulay2/packages/TateOnProducts.m2
@@ -672,7 +672,7 @@ tateResolution(Module, List, List) := (M,low, high) ->(
     hi := apply(#regs, i->max(regs_i, high_i+1)); --hi
     N := presentation truncate(hi, M)**S^{hi};-- betti N
     Q := symExt(N,E); --betti Q
-    (res (coker Q,LengthLimit=>(sum hi-sum low)))**E^{hi}[sum hi + 2 - length hi]
+    (res (coker Q,LengthLimit=>(sum hi-sum low)))**E^{hi}[sum hi]
     )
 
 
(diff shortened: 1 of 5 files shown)


M2/Macaulay2/packages/TateOnProducts.m2
```diff
@@ -672,7 +672,7 @@ tateResolution(Module, List, List) := (M,low, high) ->(
     hi := apply(#regs, i->max(regs_i, high_i+1)); --hi
     N := presentation truncate(hi, M)**S^{hi};-- betti N
     Q := symExt(N,E); --betti Q
-    (res (coker Q,LengthLimit=>(sum hi-sum low)))**E^{hi}[sum hi]
+    (res (coker Q,LengthLimit=>(sum hi-sum low)))**E^{hi}[sum hi + 2 - length hi]
```
DanGrayson: @eisenbud @mikestillman -- is this change to your package okay?

mahrud: By the way, I have spoken to both David and Dan Erman about this before. Without it the Tate resolution for modules over products of three or more projective spaces will be incorrect and the [very last test](https://github.com/Macaulay2/M2/pull/1046/files#diff-057d56d65bb5d796a7b3c98fb2f88eddL231-L237) would fail, so if we don't hear back from them the short term solution would be to comment out that test again.

mahrud: @DanGrayson added a new commit to undo this change so the PR can be merged for now. If you hear back in time before the release, please revert the last commit.

PR #1049, opened 2019-11-15
Fixes the expression of local rings and update documentation of (length, Module)

mahrud:
This PR has two small changes that I was waiting to submit with an update to the local rings package, but that hasn't finished yet, so just wanted to submit these two for now:
- adds documentation for (length,Module) of local rings (I did this by moving the documentation node from degree-doc.m2 to LocalRings/doc.m2, but if you think it's better I can move the same thing to degree-doc.m2)
- fixes the expression of local rings. This has been an issue since v1.13, where named local rings were not presented correctly. (for instance compare [v1.12](https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.12/share/doc/Macaulay2/LocalRings/html/index.html) and [v1.13](https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.13/share/doc/Macaulay2/LocalRings/html/index.html))

### DIFF
--- a/M2/Macaulay2/packages/Macaulay2Doc/functions/degree-doc.m2
+++ b/M2/Macaulay2/packages/Macaulay2Doc/functions/degree-doc.m2
@@ -163,20 +163,7 @@ document {
      SeeAlso => {hilbertPolynomial, isHomogeneous}
      }
 
-document { 
-     Key => (length,Module),
-     Usage => "length M",
-     Inputs => {
-	  "M"
-	  },
-     Outputs => {
-	  ZZ => {"the length of ", TT "M"}
-	  },
-     "We assume that ", TT "M", " is a graded module over a singly graded 
-     polynomal ring or a quotient of a polynomial ring, 
-     over a field ", TT "k", ".  In this case, the length is the same as the degree, 
-     see ", TO (degree,Module), "."
-     }
+-- (length,Module) was moved to packages/LocalRings/doc.m2
 
 document { 
      Key => (degree,Ring),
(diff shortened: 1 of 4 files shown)


DanGrayson:
Since there is code in the Core for length(Module), we also need documentation for that code.

mahrud:
The documentation for that code, which is the length of projective modules,
is already included in this documentation node (the first paragraph of the
description). Is there anything else?

DanGrayson:
Since there is code in the Core for length(Module), we also need documentation for that code in the Core.  Otherwise people won't find it.

mahrud:
The documentation node for "length", which is in core, will have a link to
"length (module)", won't it? If not, I can move the code to core.

mahrud:
This should fix both problems, though without any examples.

DanGrayson:
Isn't the documented behavior over local rings dependent on whether the package is loaded?d  If so, it should say so.

mahrud:
I've added a sentence that in the local case "the @TO LocalRings@ package must be loaded first." Should it be in a different format, like a caveat?

DanGrayson:
> I've added a sentence that in the local case "the @to LocalRings@ package must be loaded first." Should it be in a different format, like a caveat?

Maybe you forgot to push it.

M2/Macaulay2/packages/Macaulay2Doc/functions/degree-doc.m2
```diff
@@ -163,20 +163,34 @@ document {
...
+  l: ZZ
+    the length of M
+Description
+  Text
+    If M is a graded module over a singly graded polynomal ring or a quotient of a
+    polynomial ring over a field k then length is the same as the degree.
+
+    If M is over a local ring then length is computed by summing the output of
+    the Hilbert-Samuel function until it vanishes. Note that in this case the @TO LocalRings@
+    package must be loaded first.
```
mahrud: It is in lines 184 and 185.

DanGrayson:
Ah, okay, good.

PR #1055, opened 2019-11-25
Precursor complex to the Total Ext over a complete intersection

eisenbud:


eisenbud:
If Rbar = R/(f1..fc) is a complete intersection, and Mbar is an R-module over finite projective dimension over S, this script realizes the even and odd parts of the Eisenbud-Shamash resolution, lifted to S, as a Z/2-graded finitely generated free module over R[s_1..s_c].  It thus provides a different route to computing the total Ext functor over Rbar, not as efficient as the built-in function for that purpose, but useful for certain functoriality.

PR #1065, opened 2020-01-19
Dockerfiles for M2

mahrud:
This PR will track work on Dockerfiles for Macaulay2 images, continuing from the Google groups [thread](https://groups.google.com/d/msg/macaulay2/tYMtN1hpIXk/caF9eM-hEQAJ).

The first commit adds a minimal Dockerfile which includes M2 v1.15 along with a graphical Emacs user interface. Run `make build` to build the image (<5 minutes) and `make run` to start and instance (<1 second).

We can add several other Dockerfiles here:
- compile from source and run tests for CI
- start a local Habanero instance (see [Dockerfile](https://github.com/Macaulay2/InteractiveShell/blob/master/Dockerfile) from InteractiveShell)
- start a local Jupyter instance with M2 (see @radoslavraynov's [Dockerfile](https://github.com/radoslavraynov/Macaulay2-Jupyter-Kernel/blob/master/docker/Dockerfile))

I'm not sure if `M2/BUILD/docker` is the most easy to find place for these so any suggestions are welcome.

### DIFF
--- a/M2/BUILD/docker/README.md
+++ b/M2/BUILD/docker/README.md
@@ -1,0 +1,28 @@
+# Running Macaulay2 in Docker
+
+Docker is a software for building, shipping, or running applications on different
+platforms. A Dockerfile contains instructions for building an image with Macaulay2.
+
+## Getting Started
+0. Install Docker and start the daemon.
+
+1. Build a Macaulay2-in-a-container:
+```
+make build
+```
+
+2. Run Macaulay2-in-a-container:
+```
+make run
+```
+
+## Accessing files and packages
+
+A local folder named `storage` is shared as the `/home` directory inside the container.
+Furthermore, the current Macaulay2 git repository is available at `/home/M2`.
+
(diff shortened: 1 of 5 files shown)


M2/BUILD/docker/README.md
```diff
@@ -0,0 +1,28 @@
+# Running Macaulay2 in Docker
+
+Docker is a software for building, shipping, or running applications on different
```
DanGrayson: Change "a software" to "software".

mahrud: I'll fix this in the next commit. Just need a few more things before I push.

mahrud:
Small update: to try this out, you can also pull the image from the docker registry rather than building it:
```
docker pull mahrud/macaulay2:v1.15
docker run --volume `pwd`/storage:/home --net=host --env="DISPLAY" mahrud/macaulay2:v1.15
```

PR #1067, opened 2020-01-28
fixes #1061: teaching setup() about zsh

### LINKED ISSUE #1061: teach "setup()" about "zsh"
Mac users are now pressured to switch to "zsh", which uses ".zprofile" instead of ".profile" as its init filename, so "setup()" should take that into account.

mahrud:
macOS 10.15 Catalina switched the default shell to zsh, which for some reason does not know about M2 by default. This became an issue for at least one person at a coding sprint.

I'm not confident that this patch works on the prescribed situation though, as `.zshrc` or `.zprofile` files were not present, only `.zsh_history`. It could be that this machine was an exception, so someone else who has upgraded their Apple laptop to this version should ideally try it first.

It might also be worth using the `SHELL` environment variable instead.

fixes #1061


DanGrayson:
Your suggestion is a good one, see #1068

eisenbud:
It's not hard to switch back to bash as default, which is what I did.

David
--
David Eisenbud
Professor of Mathematics,University of California, Berkeley
www.msri.org/~de

PR #1072, opened 2020-02-10
simplify (show, URL)

mahrud:
This PR simplifies part of `html.m2` that deals with opening a browser. With this change, instead of specifying how to open a few browsers, we look for the `open` on macOS and `[xdg-open](https://www.freedesktop.org/wiki/Software/xdg-utils/)` on Linux. I'm fairly sure all Linux distributions that we support come with `xdg-utils` scripts, but I haven't tested them all, so as a backup we still check if firefox exists.

### DIFF
--- a/M2/Macaulay2/m2/html.m2
+++ b/M2/Macaulay2/m2/html.m2
@@ -1181,6 +1181,7 @@ makePackageIndex List := path -> ( -- TO DO : rewrite this function to use the r
      )
 
 runnable := fn -> (
+     if fn == "" then false;
      if isAbsolutePath fn then fileExists fn
      else 0 < # select(1, apply(separate(":", getenv "PATH"), p -> p|"/"|fn), fileExists)
      )
@@ -1190,8 +1191,9 @@ show URL := x -> (
      url := x#0;
      if runnable "open" then browser := "open" -- Apple varieties
      else if runnable "xdg-open" then browser = "xdg-open" -- most Linux distributions
+     else if runnable getenv "WWWBROWSER" then browser = getenv "WWWBROWSER" -- compatibility
      else if runnable "firefox" then browser = "firefox" -- backup
-     else error "neither open nor xdg-open are found";
+     else error "neither open nor xdg-open are found and WWWBROWSER is not set";
      cmd := { browser, url };
      if fork() == 0 then (
 	  setGroupID(0,0);
--- a/M2/Macaulay2/m2/html.m2
+++ b/M2/Macaulay2/m2/html.m2
@@ -1193,7 +1193,7 @@ show URL := x -> (
      else if runnable "xdg-open" then browser = "xdg-open" -- most Linux distributions
      else if runnable getenv "WWWBROWSER" then browser = getenv "WWWBROWSER" -- compatibility
      else if runnable "firefox" then browser = "firefox" -- backup
-     else error "neither open nor xdg-open are found and WWWBROWSER is not set";
+     else error "neither open nor xdg-open is found and WWWBROWSER is not set";
      cmd := { browser, url };
      if fork() == 0 then (
 	  setGroupID(0,0);
--- a/M2/Macaulay2/m2/html.m2
+++ b/M2/Macaulay2/m2/html.m2
@@ -1181,38 +1181,18 @@ makePackageIndex List := path -> ( -- TO DO : rewrite this function to use the r
      )
 
 runnable := fn -> (
-     if isAbsolutePath fn then (
-	  fileExists fn
-	  )
-     else (
-     	  0 < # select(1,apply(separate(":", getenv "PATH"),p -> p|"/"|fn),fileExists)
-	  )
+     if isAbsolutePath fn then fileExists fn
+     else 0 < # select(1, apply(separate(":", getenv "PATH"), p -> p|"/"|fn), fileExists)
      )
-chk := ret -> if ret != 0 then (
-     if version#"operating system" === "MicrosoftWindows" and ret == 256 then return;     
-     error "external command failed"
-     )
-browserMethods := hashTable {
-     "firefox" => url -> {"firefox", url},
-     "open" => url -> {"open", url},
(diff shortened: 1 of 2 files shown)


DanGrayson:
The environment variable WWWBROWSER is most likely documented and users might be depending on it, so it should still be referred to first.  

Could you also update the documentation to refer to xdg-open (which I wasn't aware of)?  And give a hint about how one might set the browser preference?

mahrud:
I can't find any documentation for `WWWBROWSER` other than two mentions in the google groups a few years ago, but I added info about open/xdg-open. Since it's also not documented, can we deprecate it? We can't anticipate all the browsers someone might want to use in `browserMethods` and if we use `WWWBROWSER` we also have to keep `browserMethods` since just running whatever is in an environment variable isn't safe.

DanGrayson:
We don't have to anticipate all the browsers, because the default is to run the command with one argument consisting of the file name.

I can't find the documentation, either, but I bet people are using it.  It's needed for the case where xdg-open and open are not there, so we can't get rid of it.  

Another point about 'xdg-open' is that the preference probably must be set on the gnome or kde desktop, and not everyone uses a desktop GUI, either.

mahrud:
> We don't have to anticipate all the browsers, because the default is to run the command with one argument consisting of the file name.

I think this unsafe in general because `WWWBROWSER` could set to something dangerous. If we don't mind this, why not let users set wwwbrowser in `init.m2` instead?

> Another point about 'xdg-open' is that the preference probably must be set on the gnome or kde desktop, and not everyone uses a desktop GUI, either.

It's set by browsers when they ask you to make them the default browser. In particular, I tested xdg-open in a headless remote shell and Links 2.12 popper up! So I still think this is a preferred standard over an environment variable, but I added it again.

mahrud:
Just realized there was an extra semicolon in the last commit that made it fail. It's resolved now.

DanGrayson:
> > We don't have to anticipate all the browsers, because the default is to run the command with one argument consisting of the file name.
> 
> I think this unsafe in general because `WWWBROWSER` could set to something dangerous. If we don't mind this, why not let users set wwwbrowser in `init.m2` instead?

Why is "init.m2" any safer?  It could also run something dangerous.

mahrud:
Absolutely. I meant if we don't mind that danger, setting it in init.m2 is easier than setting an environment variable before running M2.

DanGrayson:
I think we don't have a way to set it in `init.m2`.

mahrud:
Is there anything else needed on this PR?

M2/Macaulay2/m2/html.m2
```diff
@@ -1181,38 +1181,20 @@ makePackageIndex List := path -> ( -- TO DO : rewrite this function to use the r
...
-	  if runnable "firefox" then browser = "firefox"
-	  else if runnable "netscape" then browser = "netscape"
-	  else error "no browser found, and none specified in $WWWBROWSER"
-	  );
-     cmd := if browserMethods#?browser then browserMethods#browser url else { browser, url };
+     if runnable "open" then browser := "open" -- Apple varieties
+     else if runnable "xdg-open" then browser = "xdg-open" -- most Linux distributions
+     else if runnable getenv "WWWBROWSER" then browser = getenv "WWWBROWSER" -- compatibility
+     else if runnable "firefox" then browser = "firefox" -- backup
+     else error "neither open nor xdg-open are found and WWWBROWSER is not set";
```
DanGrayson, line 1196: The error message could also mention that firefox was not found.

M2/Macaulay2/packages/Macaulay2Doc/doc1.m2
```diff
@@ -94,9 +94,13 @@ document {
 	  "s" => "a descriptor for a documentation node (see below for examples)"
 	  },
      Consequences => {
-	  {"The given documentation page is displayed in your default web browser.  If
-	  the browser is not running, it is started.  If no argument is given to ", TT "viewHelp", 
-	  "then the top page of your local html documentation is displayed."}},
+	  {"The given documentation page is displayed in your default web browser, as determined
+	    by either ", TT "open", " on macOS or ", TT "xdg-open", " on Linux distributions.
+	    As backup for when neither ", TT "open", " nor ", TT "xdg-open", " are available,
+	    the environmental variable ", TT "WWWBROWSER", " or ", TT "firefox", " are used.
```
DanGrayson, line 100: Change "are" to "is".  On the previous line, also.

DanGrayson:
I've added a couple of comments.  Also, I've noticed that if the browser cannot be found, then
this code does not flag an error in the parent process:
```
     if fork() == 0 then (
	  setGroupID(0,0);
     	  try exec cmd;
     	  stderr << "exec failed: " << toExternalString cmd << endl;
	  exit 1
	  )
```
Can you think of a better way to do it?

DanGrayson:
What about adding "chrome" to the list?  (See #1152 , where I was not remembering this pull request.)

mahrud:
I think we should not add individual browsers and instead use the OS to decide that for us, based on the users preferences. I wanted to remove Firefox as well, but left it because you asked and for backwards compatibility. Even if not for that reason, the binary of chrome is `google-chrome` on one of my machines and `chromium` on another.

mahrud:
> I've added a couple of comments. Also, I've noticed that if the browser cannot be found, then
> this code does not flag an error in the parent process:
> 
> ```
>      if fork() == 0 then (
> 	  setGroupID(0,0);
>      	  try exec cmd;
>      	  stderr << "exec failed: " << toExternalString cmd << endl;
> 	  exit 1
> 	  )
> ```
> 
> Can you think of a better way to do it?

I personally prefer that it didn't flag an error. If you'd like to push a commit to this PR that raises an error or reverts to "help", feel free to. I've allowed commits to this PR from maintainers.

mahrud:
I'm working on a larger refactoring in the Core that would conflict with this if it isn't merged soon. If you think more work is left, I'm inclined to close this and solve the issue separately.

PR #1073, opened 2020-02-10
fix (check, Package)

mahrud:
When Macaulay2 runs tests or creates documentation examples, if the process is interrupted with ctrl-c we want to stop running the rest of the tests or examples. This happened by checking whether the return code is 2. At least on my system, however, `(run, String)` multiplies the exit code by 256, which prevents Macaulay2 from detecting ctrl-c. This PR fixed this behavior.

**Question:** I couldn't find the source of `(run, String)` to verify whether multiplication by 256 happens in Macaulay2 or by some system call on my system. Where is it defined? I found it in `actors3.d` but can't find it in engine.

### DIFF
--- a/M2/Macaulay2/m2/html.m2
+++ b/M2/Macaulay2/m2/html.m2
@@ -509,9 +509,9 @@ runFile := (inf,inputhash,outf,tmpf,desc,pkg,announcechange,usermode,examplefile
      if # findFiles rundir == 1
      then removeDirectory rundir
      else stderr << rundir << ": error: files remain in temporary run directory after program exits abnormally" << endl;
-     r = if r<256 then r else r//256;
-     stderr << "M2: *** Error " << r << endl;
-     if r == 2 then error "interrupted";
+     if r % 256 == 0 then r = r >> 8; -- cancelling (ret) << 8 in bits/waitstatus.h
+     if r % 128 == 2 then error ("interrupted" | if r & 128 =!= 0 then " (core dumped)" else "")
+     else stderr << "M2: *** Error " << r << endl;
      hadExampleError = true;
      numExampleErrors = numExampleErrors + 1;
      return false;
--- a/M2/Macaulay2/m2/html.m2
+++ b/M2/Macaulay2/m2/html.m2
@@ -509,7 +509,8 @@ runFile := (inf,inputhash,outf,tmpf,desc,pkg,announcechange,usermode,examplefile
      if # findFiles rundir == 1
      then removeDirectory rundir
      else stderr << rundir << ": error: files remain in temporary run directory after program exits abnormally" << endl;
-     stderr << "M2: *** Error " << (if r<256 then r else r//256) << endl;
+     r = if r<256 then r else r//256;
+     stderr << "M2: *** Error " << r << endl;
      if r == 2 then error "interrupted";
      hadExampleError = true;
      numExampleErrors = numExampleErrors + 1;


DanGrayson:
Thanks!

Here's what it does under Ubuntu 18.04:
```
i1 : run "sleep 333"

o1 = 33280

i2 : oo/256.

o2 = 130

o2 : RR (of precision 53)

i3 : oo-128

o3 = 2

o3 : RR (of precision 53)
```

Could you add this?:
```
     if r >= 128 then r = r-128;
```
(The high order bit is used for indicating whether core was dumped.)

The logic for all this is in macros in the include file bits/waitstatus.h, and the names of the macros are provided in the man page for the "system" library function.

mahrud:
This is bizarre. Can you tell me what running the following two returns?
```
M2 -q --silent -e "exit 2"; echo $?
M2 -q --silent -e "print run ///bash -c 'exit 2'///; exit 0"
```
and also this one, interrupted with ctrl-c:
```
M2 -q --silent -e "x=run ///sleep 333///; print x; exit 0"
```

The results I tried are:
| Platform                      | results       |
| ----------------------------- | -------------- |
| Mike's mac                  | 2, 512, 2   |
| Mike's ubuntu 18.04    | 2, 512, (didn't test 3rd one) |
| my Fedora 31              | 2, 512, 2   |
| my ubuntu 16.04         | 2, 512, 2   |

One issue with this is that for instance `M2 -q --silent -e "exit run ///bash -c 'exit 2'///"; echo $?` returns 0 despite the error inside because bash mods the return code by 256.

I also tried killing `sleep` with a signal, and in that case if it is run within M2 I get the kill signal but if it's running in bash I get 128+kill signal. How did you get a core dump?

mahrud:
If I understand how things work correctly, `run` from the interface is wired to `run(e:Expr)` defined in `d/system.d` which is wired to `system_run(M2_string command)` defined in `d/scclib.c`. 

I wrote a similar C code to run sleep and tried interrupting it with ctrl-c, I still get 2, so I'm confused how we end up with 512!
```
int main()
{
        char *c = "sleep 100";
        int r = system(c);
        return r;
}    
```

DanGrayson:
The sh command "echo $?" shows you only the low-order 8 bits of the integer return value of the program.  That's how the 512 value of "run" got changed to 0.

Running a command that returns a nonzero value does not result in an M2 error, by design.  That enables the programmer to examine the return code and decide what to do.

There has to be a way to tell whether a process was interrupted, and in case it was, to get the interrupt number, and in case not, to get the return code (the 2 in "exit 2").  And we want M2 programs to be able to tell the difference, so the M2 "run" command, implemented by the C function "system_run", returns the return code that "system()" provided, unmodified.  The difference between 512 and 2 provides that distinction.

mahrud:
I see. So where does the 8 bit shift happen? `system` doesn't do that (the C code in my last comment).

mahrud:
Ah, never mind, I was getting myself confused by all the different combinations. I think the last commit should do the job well.

DanGrayson:
Yes, it's confusing.  And it may not be possible to get it right under all OS's at top level, without access to the macros referred to by the "system()" man page.  We can always add some code
in the d directory for taking care of application of those macros to the return code, if necessary.

Run a test to make sure that a C program that returns with "exit(2)" is not treated as having been interrupted to verify your code.

mahrud:
At least on my machine I don't think that's possible, because ctrl-c and exit(2) behave the same; both result in return code 512.

DanGrayson:
Which OS is that?

mahrud:
This is from a fresh build from master on Fedora 31,
Kernel 5.4.17-200.fc31.x86_64. I tried running:
```
i1 : run "M2 -q --stop --silent -e \"sleep 10; exit 2;\""
^Cinterrupted, stopping
o1 = 512
i2 : run "M2 -q --stop --silent -e \"sleep 10; exit 2;\""
o2 = 512
```

DanGrayson:
Try a test that does not pass the return code through the shell, which changes it:
```
i3 : run "sleep 10000"

o3 = 2

i4 : run "exit 2"

o4 = 512
```

mahrud:
Oh, I see. I get the same result then. How did you get a core dump earlier?
I haven't been able to test that.

DanGrayson:
Well, I didn't actually see a core dump or any indication of it, other than that high-order bit being set.  Try this:
```
As root I run these commands:
```
[root@fedora tmp]# ulimit -c 0
[root@fedora tmp]# ./foo
 *** system() returns b
[root@fedora tmp]# ls -l core*
ls: cannot access 'core*': No such file or directory
[root@fedora tmp]# ulimit -c unlimited
[root@fedora tmp]# ./foo
 *** system() returns 8b
[root@fedora tmp]# ls -l core*
-rw-------. 1 root root 385024 Feb 12 10:37 core.1339
```

Here's what I did in another window to generate interrupts:

```
[root@fedora dan]# ps aux | grep sleep
root      1332  0.0  0.0   5252   692 pts/1    S+   10:36   0:00 /bin/sleep 10000
root      1334  0.0  0.0   6000   816 pts/2    S+   10:36   0:00 grep --color=auto sleep
[root@fedora dan]# kill -SEGV 1332

[root@fedora dan]# ps aux | grep sleep
root      1339  0.0  0.0   5252   692 pts/1    S+   10:37   0:00 /bin/sleep 10000
root      1341  0.0  0.0   6000   892 pts/2    S+   10:37   0:00 grep --color=auto sleep
[root@fedora dan]# kill -SEGV 1339
```

Here is the program foo.c:

```
#include <stdlib.h>
#include <stdio.h>
int main () {
     printf(" *** system() returns %x\n",system("/bin/sleep 10000"));
     return 0;
     }
```

**Comment by DanGrayson** [comment]
PS: you also have to do this, or core dumps don't go to a file:
```
[root@fedora dan]# echo core > /proc/sys/kernel/core_pattern 
```

DanGrayson:
I've changed the interrupt handling in
commit ab6b03b8679d909b975f071c6d6142f09846131a,
commit d02d853c1fad2445fdc13e65ea82989aea0074fd, and
commit 0304da334de3cfb7ee9e224741240a857a03ee5c.  Is this pull request now unneeded?

mahrud:
Probably,but I think now this needs to be fixed:
https://github.com/Macaulay2/M2/blob/d6c70324fb5dc511bfec32cee9a8880a1e48e5e9/M2/Macaulay2/m2/html.m2#L460-L463

DanGrayson:
What's wrong with it?

PR #1074, opened 2020-02-11
Fix the build with -fno-common (default for GCC 10)

jamesjer:
The upcoming GCC 10 release changes the long-standing default of -fcommon to -fno-common (see https://gcc.gnu.org/gcc-10/changes.html), thereby breaking the Macaulay2 build in Fedora Rawhide.  This patch fixes the lone instance of a multiply-defined variable.


DanGrayson:
Thanks!

PR #1078, opened 2020-02-18
Joining tests in groups of at most 10 to avoid restarting M2 too often

lkastner:
As recommended, I joined many of the tests together to have less test items for Polyhedra.


DanGrayson:
That's a lot of tests!

PR #1084, opened 2020-03-28
fixed tests in package SlackIdeals.m2

antostain:
fixed tests in package SlackIdeals.m2

### DIFF
--- a/M2/Macaulay2/packages/SlackIdeals.m2
+++ b/M2/Macaulay2/packages/SlackIdeals.m2
@@ -3134,25 +3134,6 @@ doc ///
 		Object
 ///
 
-end
-
-doc /// Key
-   Headline
-   Usage
-   Inputs
-   Outputs
-   Consequences
-    Item
-   Description
-    Text
-    Code
-    Pre
-    Example
-   Subnodes
-   Caveat
-   SeeAlso
-///
-
 ------ TESTS ------
 
 TEST ///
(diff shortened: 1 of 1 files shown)


DanGrayson:
I tested it with `check  "SlackIdeals"` and got 13 tests failing.  Do they succeed on your end?

antostain:
fixing tests in package SlackIdeals.m2 by adding appropriate rings

antostain:
Dear Prof. Grayson,

I fixed the tests, the problem was the definition of the appropriate ring for matrices and polynomial ring for symbolic matrices and ideals.

 

I noticed a weird behaviour when matrices are concatenated: if a numeric matrix B is obtained as a sequential concatenation of columns and rows (see below), this equals the same matrix A defined by rows. On the other hand this is no longer true if A and B are symbolic matrices. This was the cause of failure of some tests, which I fixed. Is this the expected behaviour? 

 

A = matrix{{1,0},{0,1}}

C = matrix{{0}}||matrix{{1}}

B = matrix{{1},{0}}|C

A == B

o1 = true

 

R = QQ[x,y]

A = matrix{{x,0},{0,y}}

C = matrix{{0}}||matrix{{y}}

B = matrix{{x},{0}}|C

A == B

o2 = false

 

Best,

Antonio

 

 

 

Da: Daniel R. Grayson <notifications@github.com> 
Inviato: 28 March 2020 19:34
A: Macaulay2/M2 <M2@noreply.github.com>
Cc: antostain <macchia.antonello@outlook.com>; Author <author@noreply.github.com>
Oggetto: Re: [Macaulay2/M2] fixed tests in package SlackIdeals.m2 (#1084)

 

I tested it with check "SlackIdeals" and got 13 tests failing. Do they succeed on your end?

DanGrayson:
That's intentional:

      _Matrices
      --------

      Two "matrices" are equal if their entries are equal, the source and target are the same
      (including degrees), and the degree of the matrices are the same.  In this example, m
      and n have different source free modules._

In this case, the matrices' source modules have different degrees for their generators:

```

i11 : degrees source A

o11 = {{1}, {1}}

o11 : List

i12 : degrees source B

o12 = {{1}, {0}}

o12 : List
```

PR #1103, opened 2020-04-22
MonodromySolver: delete duplicated and unnecessary files 

timduff35:
This is the first of planned pull requests for improving the package MonodromySolver. This one is mostly deleting lots of files-- most were inadvertently duplicated. Thanks @mahrud for catching this a few months back.


mahrud:
@DanGrayson this PR looks good. Should I go ahead and merge it or would you prefer to test it also?

DanGrayson:
My general procedure for pull requests to packages is to verify that the requester is an author, the package still works with 'installPackage' and 'check', and the only files modified are within the package.  If you've verified that, go ahead and pull.  Thanks.

(Actually, those tests aren't a guarantee that all is well, because other packages might use the modified package.)

If you assign yourself to such pull requests, it'll prevent duplication of effort, if I happen to try to test it out, too.

mahrud:
Will do. I checked this one in a full build on master.

DanGrayson:
By the way, a good way to get the files for testing a pull request is like this:
```
git fetch upstream pull/1103/merge
git checkout FETCH_HEAD
```

Issue #1058, opened 2019-12-02
a mathicgb test doesn't compile


DanGrayson:
compiling under Ubuntu 19.10 with gcc 9.2.1:

```
[...]

PR #1104, opened 2020-04-23
Check that we're using googletest >= 1.10.0 during build.

d-torrance:
mathicgb's unittests fail using 1.8.1 (Macaulay2/mathicgb#12
and Macaulay2/mathicgb#13).  We adapt the check proposed in
Macaulay2/mathicgb#15 to Macaulay2.

closes: #1058

### DIFF
--- a/M2/INSTALL
+++ b/M2/INSTALL
@@ -83,7 +83,7 @@ various modern operating systems:
 
   Ubuntu and Debian:
     Install packages as root with:
-      apt-get install -y -q sudo autoconf bison curl emacs fflas-ffpack flex g++ gcc gfortran install-info libatomic-ops-dev libboost-dev libc6-dev libcdd-dev libgc-dev libgdbm-dev libgivaro-dev libglpk-dev libgmp3-dev libgtest-dev liblapack-dev liblzma-dev libmpc-dev libmpfr-dev libncurses-dev libncurses5-dev libntl-dev libpari-dev libreadline-dev libtool libxml2-dev libz-dev make openssh-server patch pinentry-curses pkg-config time unzip xbase-clients yasm zlib1g-dev polymake w3c-markup-validator git dpkg-dev gfan
+      apt-get install -y -q sudo autoconf bison cmake curl emacs fflas-ffpack flex g++ gcc gfortran install-info libatomic-ops-dev libboost-dev libc6-dev libcdd-dev libgc-dev libgdbm-dev libgivaro-dev libglpk-dev libgmp3-dev libgtest-dev liblapack-dev liblzma-dev libmpc-dev libmpfr-dev libncurses-dev libncurses5-dev libntl-dev libpari-dev libreadline-dev libtool libxml2-dev libz-dev make openssh-server patch pinentry-curses pkg-config time unzip xbase-clients yasm zlib1g-dev polymake w3c-markup-validator git dpkg-dev gfan
         # note: libz-dev seems to have been replaced by zlib1g-dev
         # note: libncurses-dev seems to have been replaced by libncurses5-dev
         # note: libreadline-gplv2-dev is an older GPL v2 version of libreadline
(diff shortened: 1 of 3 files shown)


DanGrayson:
How can this test for the version number being at least 1.10.0 if "1.10.0" appears only in a message and nowhere else?

d-torrance:
In `AC_COMPILE_IFELSE`, we try to compile a function which uses `TYPED_TEST_SUITE` .  It wasn't introduced until 1.10.0, and so the check will fail with earlier versions.

d-torrance:
No problem!  I pushed a new commit with a detailed comment a few minutes ago.

It looks like M2 itself still builds 1.8.0, and the mathicgb unit tests ran just fine until 1.8.1.  (The [corresponding Debian bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=910237) was filed right after 1.8.1 was uploaded.)  It seems a little strange to require 1.10.0 at this stage and then end up building 1.8.0!  So this probably shouldn't be merged until one of the two happens:

* We build 1.10.0
* This check is rewritten to also allow googletest <= 1.8.0.

I'm happy to work on either one.  Any thoughts?

DanGrayson:
Thanks for volunteering, I've assigned you (for both).

Yes, not only do we build 1.8.0, but it calls itself 1.7.0!  This package is a pain.

I've just uploaded `gtest-1.10.0.tar.gz` to our website's "OtherSourceCode" directory for you.  It seems to be no longer based on "autoconf".

d-torrance:
Thanks for uploading the tarball!  I made a couple changes to the Makefile and everything is building fine and the mathicgb unit tests are passing, at least on my system (Ubuntu 19.10).

They did switch from autotools to cmake, however.  It doesn't look like any of the other libraries use it.  Is that a big issue?

I still need to figure out how to check for googletest <= 1.8.0...

mahrud:
@d-torrance in addition to gtest, flint, gc, mathic, mathicgb, and memtailor all have cmake build tools now, at least on their master branch. If you're interested and have time to try it out, I have a [cmake build guide](https://github.com/mahrud/M2/blob/feature/cmake/M2/INSTALL-CMake.md) for M2 as well. It really makes building M2 easier imho.

d-torrance:
Cool, I'll check it out!

DanGrayson:
I suppose we can assume that cmake is available on all the systems we use, but we should discover how to install it (i.e., what the name of the package it's in is) and add that to the various OS-specific instructions in M2/INSTALL.  I've just checked Arch Linux - the package name is "cmake".

PS: Mahrud is working on a cmake build system for Macaulay2 itself.

d-torrance:
Sounds good!  The package name is also "cmake" on Debian-based distros.

d-torrance:
I just added a commit updating INSTALL with (most) Linux distros.  I wasn't sure what to do with Gentoo, as installing the other build tools (g++, etc.) aren't mentioned anywhere in that section.

DanGrayson:
Is it ready yet for me to test?

d-torrance:
For building with googletest 1.10.0, yes, I think it's ready.

I still need to write a check to see if a user's previously-installed googletest is older, but not 1.8.1.  If they have 1.7.0 or 1.8.0 or something, then it should still be okay to use that.

(Unless https://github.com/d-torrance/mathicgb/commit/05bb211 is merged in, which would break testing with googletest < 1.10.0 completely.)

DanGrayson:
Thanks!

mahrud:
I don't think this works. CMake is not simply a substitute for make, and the build command is not going to be just `cmake`. The `cmake` command generates the build environment (e.g. creating Makefiles, or a number of other alternatives), then the command for the specific build system should be used.

d-torrance:
Well, it "works" in the sense that Macaulay2 still builds.  :)

It may be that we don't even need `cmake`...  I know that memtailor, mathic, and mathicgb just need to know where the googletest source code is -- they don't actually need a *built* copy of it.

mahrud:
Macaulay2 does not need gtest to build, or run, so sure it builds, but tests can't be run because the gtest library is not being built. The headers are being copied, but those aren't enough. Here is the beginning of the output that this Makefile.in produces:
```
make: Entering directory '/home/mahrud/Projects/M2/quickfix/M2/BUILD/mahrud/build/libraries/gtest'
+ set +x
make: Entering directory '/home/mahrud/Projects/M2/quickfix/M2/BUILD/mahrud/build/libraries/gtest/build/googletest-release-1.10.0/googletest'
+ cmake
Usage

  cmake [options] <path-to-source>
  cmake [options] <path-to-existing-build>
  cmake [options] -S <path-to-source> -B <path-to-build>

Specify a source directory to (re-)generate a build system for it in the
current working directory.  Specify an existing build directory to
re-generate its build system.

Run 'cmake --help' for more information.

+ :
+ touch .compiled-1.10.0
```
It's printing the usage for cmake.

mahrud:
I see, they are building gtest internally, so we aren't really building gtest separately. I think the BUILDCMD (as well as cmake in INSTALL) should just be commented out.

d-torrance:
Sounds good.  I'm doing a test build right now without `cmake`, and if everything looks good, I'll push it to the same branch as #1106.

PR #1106, opened 2020-04-25
Confirm that googletest version != 1.8.1

d-torrance:
Since the only version of googletest which results in failed mathicgb
unit tests is 1.8.1, we allow older versions instead of forcing the
latest version 1.10.0.

In particular, if the test for >= 1.10.0 fails, then we try to compile
a program using the ScopedTrace utility, which was exposed to the
public interface in 1.8.1.  If this passes, then we know 1.8.1 is
installed, and so we should build our own gtest instead.  If it fails, then
we have an older version, but we will use it since it will be okay for
mathicgb.

Two other changes:

* The version check has been pulled outside of AC_CHECK_FILES.  Since we
  are checking for two files, the check was being run twice, once for
  each file.

* More useful output.  In particular, the results of the version check
  are printed.

### DIFF
--- a/M2/INSTALL
+++ b/M2/INSTALL
@@ -83,7 +83,7 @@ various modern operating systems:
 
   Ubuntu and Debian:
     Install packages as root with:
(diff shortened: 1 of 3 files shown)


d-torrance:
Just added another commit after a discussion with @mahrud in #1104.  Calling `cmake` is unnecessary.  We just need to put the source code in a place for memtailor, mathic, and mathicgb to find.

Issue #1066, opened 2020-01-24
gfan under ubuntu and M2 1.15 (and master) seems incompatible


mikestillman:
I tried to compile M2 master under ubuntu 18.04, and got errors while building (I will have to rerun to see where the problem was).  It appears to be: the version included with ubuntu (after apt-get install) seems to be a version which doesn't work with our interface, and our configure script doesn't check (well, we didn't know we needed to check!).  In fact, our configure script doesn't allow gfan to be built by M2 either.

This description is a bit vague still, I will be more precise when I rebuild M2 on ubuntu.  I just wanted to get this in as an issue, to not forget it.

DanGrayson:
I just looked at my Ubuntu builds for 1.15 on 16.04 and 18.04, and for those, I built gfan.  I see also that INSTALL doesn't recommend installing gfan (with apt-get).  So maybe I also encountered a problem, because it is installed on my 18.04 system.  You could leave it to me for later -- this doesn't get forgotten, because when we make another release, I'll run into it.

d-torrance:
I was having some troubles getting the Debian gfan package to play nicely with Macaulay2, did a search, and found this issue.  I finally fixed my problem when I figured out what that the `path` config option needed to be set to.  It doesn't give a very helpful error message when it's not set correctly!  I wonder if that's what you were getting?

Here's an example (on Ubuntu 19.10) that shows both M2's and Ubuntu's gfan packages working, and also shows the error message that's raised when `path` isn't set correctly:

    Macaulay2, version 1.15
    with packages: ConwayPolynomials, Elimination, IntegralClosure, InverseSystems,
                   LLLBases, PrimaryDecomposition, ReesAlgebra, TangentCone,
                   Truncations
    
    i1 : R = QQ[x, y, z]
    
    o1 = R
    
    o1 : PolynomialRing
    
    i2 : I = ideal(x^2 * y - y^2, y^2 * x - x^2)
    
                 2     2     2    2
    o2 = ideal (x y - y , x*y  - x )
    
    o2 : Ideal of R
    
    i3 : -- M2 copy of gfan --
         loadPackage("gfanInterface",
             Configuration => { "path" => "/usr/libexec/Macaulay2/bin/"},
             Reload => true)
    
    o3 = gfanInterface
    
    o3 : Package
    
    i4 : gfan I
    
[...]

PR #1123, opened 2020-05-06
Print useful message and raise an error when call to gfan fails

d-torrance:
The printing of a message existed in the code but was commented out.
In addition, we raise an error.  If our call to gfan doesn't work,
then the user should be informed right away rather than encountering
a less useful error a bit later when Macaulay2 isn't able to deal
with the output that doesn't exist.

Closes: #1066

### DIFF
--- a/M2/Macaulay2/packages/gfanInterface.m2
+++ b/M2/Macaulay2/packages/gfanInterface.m2
@@ -1016,20 +1016,21 @@ runGfanCommand = (cmd, opts, data) -> (
 
 	if gfanVerbose then << ex << endl;
 	returnvalue := run ex;
+	errorMsg := "";
      	if(not returnvalue == 0) then
 	(
---	     << "GFAN returned an error message.\n";
---	     << "COMMAND:" << ex << endl;
---	     << "INPUT:\n";
---	     << get(tmpFile);
---	     << "ERROR:\n";
---	     << get(tmpFile |".err");
-
+	    errorMsg = "GFAN returned an error message.\n";
+	    errorMsg = errorMsg | "COMMAND:" | ex | "\n";
+	    errorMsg = errorMsg | "INPUT:\n";
+	    errorMsg = errorMsg | get(tmpFile);
+	    errorMsg = errorMsg | "ERROR:\n";
+	    errorMsg = errorMsg | get(tmpFile |".err");
 	     );
 		out := get(tmpFile | ".out");
 	gfanRemoveTemporaryFile tmpFile;
 	gfanRemoveTemporaryFile(tmpFile | ".out");
 	gfanRemoveTemporaryFile(tmpFile | ".err");
+	if length(errorMsg) > 0 then error errorMsg;
 	outputFileName := null;
 	
 	if gfanKeepFiles then outputFileName = tmpFile|".out";
--- a/M2/Macaulay2/packages/gfanInterface.m2
+++ b/M2/Macaulay2/packages/gfanInterface.m2
@@ -88,11 +88,52 @@ export {
 	"multiplicitiesReorder"
 }
 
-gfanPath = gfanInterface#Options#Configuration#"path"
-if gfanPath == "" then gfanPath = prefixDirectory | currentLayout#"programs"
+tryGfanPath = gfanPath -> run(gfanPath | "gfan --help 2> /dev/null")
+
+-- we expect a trailing slash in the path, but the paths given in the
+-- PATH environment variable likely will not have one, so we add one
+-- if needed
+addSlash = gfanPath -> (
+	if last gfanPath != "/" then return gfanPath | "/"
+	else return gfanPath
+)
+
+checkGfanPath = gfanPath -> (
+	if gfanVerbose == true then
+		print("checking for gfan in " | gfanPath | "...");
+	if tryGfanPath(gfanPath) == 0 then (
+		if gfanVerbose == true then print("  found");
+		return true
+	) else (
+		if gfanVerbose == true then print("  not found");
+		return false
+	)
+)
+
+findGfanPath = () -> (
+	-- try user-configured path first
+	gfanPath := gfanInterface#Options#Configuration#"path";
+	if gfanPath != "" then (
+		gfanPath = addSlash(gfanPath);
+		if checkGfanPath(gfanPath) then return gfanPath;
+	);
+	-- now try M2-installed gfan
+	gfanPath = addSlash(prefixDirectory | currentLayout#"programs");
+	if checkGfanPath(gfanPath) then return gfanPath;
+	-- finally, try PATH
+	if getenv "PATH" == "" then error "could not find gfan";
+	paths := apply(separate(":", getenv "PATH"), addSlash);
+	gfanPath = scan(paths, gfanPath ->
+		if checkGfanPath(gfanPath) then break gfanPath
+	);
+	if class(gfanPath) === String then return gfanPath
+	else error "could not find gfan"
+)
 
 fig2devPath = gfanInterface#Options#Configuration#"fig2devpath"
 gfanVerbose = gfanInterface#Options#Configuration#"verbose"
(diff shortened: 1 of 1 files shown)


d-torrance:
Here's an example with the additions:

```
Macaulay2, version 1.15
               Truncations

i1 : R = QQ[x, y, z]

o1 = R

o1 : PolynomialRing

i2 : I = ideal(x^2 * y - y^2, y^2 * x - x^2)

             2     2     2    2
o2 = ideal (x y - y , x*y  - x )

o2 : Ideal of R

i3 : loadPackage("gfanInterface",
         Configuration => { "path" => "foo"},
         Reload => true)

o3 = gfanInterface

o3 : Package

i4 : gfan I
stdio:6:1:(3): error: GFAN returned an error message.
COMMAND:foogfan _bases < /tmp/M2-4272-0/0 > /tmp/M2-4272-0/0.out 2> /tmp/M2-4272-0/0.err
INPUT:
Q[x1,x2,x3]
{-x2^2+x1^2*x2,
-x1^2+x1*x2^2}
ERROR:
sh: 1: foogfan: not found
```

mikestillman:
I will likely do something similar for Topcom.

DanGrayson:
If gfan is not found in the place where we put gfan when we build it ourselves, it would be good if the package would look for it in the PATH before complaining to the user.  That way, if it's installed on the system, it will be found.  To do that, one could forget about the "path" entry in the configuration, and just try to run the program after adding `prefixDirectory | currentLayout#"programs"` to the front of the PATH.

mikestillman:
I think that using a path entry for programs in a configuration file is too clunky.  However, first looking at whether we have it installed isn't always what is desired either: it might be that the version on the path is a newer improved versino we aren't using yet (of course, then the interface might not work so well!).

One way to handle these is to have a mutable exported variable which is the path to the executable, initially set to, say, "" or null.  If this value is not set, when the package tries to find/run the executable, it can first try the `currentLayout` approach that @DanGrayson mentions.  If that doesn't exist, then try it on the path.  If the value is set, just use that.

It would be nice if all of our interface packages had similar interfaces, or ways to handle this.  Similarly, sometimes it is important to be able to see what is actually sent to the program, and see the input and output files.  What would a good standard way to do that be?  I have one method in Topcom, we have methods in the other such packages too...

DanGrayson:
Sounds good, except if you want the user to be able to specify the path, the configuration file is the good way to do that.

d-torrance:
Right now, packages deal with things very differently.  Looking at the ones I've patched to work with Debian packages, there are three different approaches:

* A `path` configuration option like `gfanInterface`.  This is also used by `FourTiTwo`, `Nauty`, and `NautyGraphs`.
* An `executable` configuration option, which also includes the name of the binary and not just the path.  This is used by `CohomCalg`.
* `Normaliz` is hardcoded to only use `prefixDirectory | currentLayout#"programs"`.

DanGrayson:
How about improving it so the package works without modification or user intervention whether gfan is built by us and in its usual place or gfan is part of the OS and found on the PATH.  Prefer the one built by us if both are there, to ensure compatibility.

d-torrance:
I've just added a new commit which runs through the various options to find gfan.

I wonder if this first commit is even necessary now?  With the second commit, we'd get an error right away if gfan doesn't work and the package stops loading, so we'd never get to the point of raising an error later on.  I suppose there's a crazy edge case where the user loads gfanInterface just fine, then goes and uninstalls gfan, and then goes back to M2 and tries making calls to the now-nonexistent gfan...

d-torrance:
Here's some example code showing the new commit in action:

```
Macaulay2, version 1.15
               Truncations

i1 : R = QQ[x];

i2 : -- using user-configured path
     loadPackage("gfanInterface",
         Configuration => {"verbose" => true, "path" => "/usr/bin"}, Reload => true)
checking for gfan in /usr/bin/...
  found

o2 = gfanInterface

o2 : Package

i3 : gfan ideal x
using temporary file /tmp/M2-10614-0/0
/usr/bin/gfan _bases < /tmp/M2-10614-0/0 > /tmp/M2-10614-0/0.out 2> /tmp/M2-10614-0/0.err

o3 = {{(x)}}

o3 : List

i4 : -- user chooses bad path, so we use M2 version
     loadPackage("gfanInterface",
         Configuration => {"verbose" => true, "path" => "foo"}, Reload => true)
checking for gfan in foo/...
  not found
checking for gfan in /usr/libexec/Macaulay2/bin/...
  found

o4 = gfanInterface

o4 : Package

i5 : gfan ideal x
using temporary file /tmp/M2-10614-0/1
/usr/libexec/Macaulay2/bin/gfan _bases < /tmp/M2-10614-0/1 > /tmp/M2-10614-0/1.out 2> /tmp/M2-10614-0/1.err

o5 = {{(x)}}

o5 : List

i6 : -- if user doesn't choose a path, we go straight to M2 version
     loadPackage("gfanInterface",
         Configuration => {"verbose" => true}, Reload => true)
checking for gfan in /usr/libexec/Macaulay2/bin/...
  found

o6 = gfanInterface

o6 : Package

i7 : gfan ideal x
using temporary file /tmp/M2-10614-0/2
/usr/libexec/Macaulay2/bin/gfan _bases < /tmp/M2-10614-0/2 > /tmp/M2-10614-0/2.out 2> /tmp/M2-10614-0/2.err

o7 = {{(x)}}

o7 : List

i8 : -- if M2 version isn't installed, we search PATH
     loadPackage("gfanInterface",
         Configuration => {"verbose" => true}, Reload => true)
checking for gfan in /usr/libexec/Macaulay2/bin/...
  not found
checking for gfan in /home/profzoom/.local/bin/...
  not found
checking for gfan in /usr/local/sbin/...
  not found
checking for gfan in /usr/local/bin/...
  not found
checking for gfan in /usr/sbin/...
  not found
checking for gfan in /usr/bin/...
  found

o8 = gfanInterface

o8 : Package

i9 : gfan ideal x
using temporary file /tmp/M2-10614-0/3
/usr/bin/gfan _bases < /tmp/M2-10614-0/3 > /tmp/M2-10614-0/3.out 2> /tmp/M2-10614-0/3.err

o9 = {{(x)}}

o9 : List

i10 : -- if we can't find gfan in PATH, then raise an error
      loadPackage("gfanInterface",
          Configuration => {"verbose" => true}, Reload => true)
checking for gfan in /usr/libexec/Macaulay2/bin/...
  not found
checking for gfan in /home/profzoom/.local/bin/...
  not found
checking for gfan in /usr/local/sbin/...
  not found
checking for gfan in /usr/local/bin/...
  not found
checking for gfan in /usr/sbin/...
  not found
checking for gfan in /usr/bin/...
  not found
checking for gfan in /sbin/...
  not found
checking for gfan in /bin/...
  not found
checking for gfan in /usr/games/...
  not found
checking for gfan in /usr/local/games/...
  not found
checking for gfan in /snap/bin/...
  not found
../.Macaulay2/code/gfanInterface.m2:136:10:(2):[7]: error: could not find gfan

i11 : gfan ideal x
stdio:21:1:(3): error: no method for adjacent objects:
--            gfan (of class Symbol)
--    SPACE   ideal x (of class Ideal)
```

DanGrayson:
> I've just added a new commit which runs through the various options to find gfan.
> 
> I wonder if this first commit is even necessary now? With the second commit, we'd get an error right away if gfan doesn't work and the package stops loading, so we'd never get to the point of raising an error later on. I suppose there's a crazy edge case where the user loads gfanInterface just fine, then goes and uninstalls gfan, and then goes back to M2 and tries making calls to the now-nonexistent gfan...

Guarding against edge cases is good practice.

DanGrayson:
It's rather noisy to see the "checking for gfan ..." messages every time.  It would be better to hush those and if gfan is not found, then the error message could include the path.

d-torrance:
Those only print if `verbose` is set to `true`.

DanGrayson:
Okay, that's fine then.

PR #1126, opened 2020-05-07
Drop unnecessary executable bit from various .m2 files.

d-torrance:
These were raising a "executable-not-elf-or-script" Lintian warning in
the Debian package:

>   This executable file is not an ELF format binary, and does not start
>   with the #! sequence that marks interpreted scripts. It might be a sh
>   script that fails to name /bin/sh as its shell, or it may be
>   incorrectly marked as executable. Sometimes upstream files developed
>   on Windows are marked unnecessarily as executable on other systems.
> 
>   If you are using debhelper to build your package, running dh_fixperms
>   will often correct this problem for you.
> 
>   Refer to Debian Policy Manual section 10.4 (Scripts) for details.


DanGrayson:
Thanks!

PR #1127, opened 2020-05-07
Wrap some very long lines in Visualize package.

d-torrance:
These were raising a very-long-line-length-in-source-file warning Lintian
warning in the Debian package:

> The source file includes a line length that is well beyond the
> normally human made code line length.
>
> This very long line length does not allow Lintian to do correctly some
> source file checks.
>
> This line could also be the result of some text injected by a computer
> program, and thus could lead to FTBFS bugs.
>
> Last but not least, long line in source code could be used to
> obfuscate the source code and to hide stuff like backdoors or security
> problems.
>
> It could be due to jslint source comments or other build tool comments.
>
> You may report this issue upstream.

Because of the long lines, we also were getting false positives for the
source-contains-prebuilt-javascript-object and source-is-missing Lintian
warnings/errors.

(Also, emacs inserted some newlines at the ends of these files.)


DanGrayson:
thanks!

PR #1140, opened 2020-05-12
Demo for adding a package, do not accept

bstone:
This was as demo for the Workshop. Please delete.


DanGrayson:
There's no such thing as deletion.  If you're giving another demo, you could demonstrate closing a pull request, which you can do yourself.  So I'll leave it open for you.

PR #1141, opened 2020-05-12
Fix bug in some MutableMatrix methods

### LINKED ISSUE #1136: raw determinant
Something is wrong with `rawLinAlgDeterminant`?
```
i1 : R = ZZ/2[a..d]

o1 = R

o1 : PolynomialRing

i2 : M = mutableMatrix genericMatrix(R,2,2)

o2 = | a c |
     | b d |

o2 : MutableMatrix

i3 : det M

o3 = a*d

o3 : R
```

dylanpeifer:
This PR adds checks to the routines `LUdecomposition`, `solve`, `rank`, `determinant`, and `inverse` when called on a `MutableMatrix` to determine if engine linear algebra routines are available. If not available, `LUdecomposition` and `solve` will throw an error, while the other commands will convert to a `Matrix` and try again. Previous implementations would return incorrect results when the base ring was a polynomial ring.

This fixes #1136, which noticed the problem for `determinant`.


pepperhdj:
We were also wondering how we might be able to create tests for this, and in general M2 routines?

DanGrayson:
Thanks!

To add a test for your work, look in M2/Macaulay2/packages/Macaulay2Doc/test/ for a file that already tests your routines, or just add a new file there whose name is of the form *.m2.

DanGrayson:
Let us know when this is in final form.

dylanpeifer:
I'm running some tests now, and will commit, push, and let you know if everything passes on my machine.

dylanpeifer:
Everything passes for me, and this is ready for you to take a look.

We ended up changing the `hasEngineLinearAlgebra` check to an `isField` check since the generic engine LU can handle any field. We also added a few tests and modified a function that uses `LUdecomposition` in the Chordal package after confirming with the author that our fix was correct.

DanGrayson:
Now you have conflicts, probably because of one other pull request just pulled.  Could you take a look and bring your master branch up to date?  No need to delete this pull request.

dylanpeifer:
It looks like the only conflict was the move of the `tests` directory from the `M2/M2/Macaulay2/packages/Macaulay2Doc/` directory to the `M2/M2/Macaulay2/` directory (which is a good change!). I pulled from development and git figured it out automatically.

DanGrayson:
Thanks!

Issue #1037, opened 2019-10-25
Make realPart a method function


jchen419:
Currently `realPart` is a `CompiledFunction`, which does not allow for overloading. It would be useful to make it a `MethodFunction`, to allow functionality such as

`realPart Matrix := Matrix => A -> matrix apply(entries A, r -> r/realPart)` 

Similarly for `imaginaryPart`.

DanGrayson:
Good idea.

pepperhdj:
Dylan and I would like to take a look like this.

DanGrayson:
Great.  What is your name?  (How about putting it and your portrait into your profile?)  What is Dylan's name and github id?  (I'll assign him, too.)

dylanpeifer:
This issue can be closed now. It looks like GitHub recognized the pull request mentioned this issue but not that it fixed it.

DanGrayson:
I think if you had said "resolves #1037" it would have figure it out.

PR #1148, opened 2020-05-13
Make realPart and imaginaryPart method functions

pepperhdj:
This fixes issue #1037.


DanGrayson:
looks good to me, thanks!

I trust it, so I'll pull without testing ...  do `git pull` at your end and let me know if there are problems.

PR #1161, opened 2020-05-15
Fix prompt overriding when read is interrupted

### LINKED ISSUE #1160: Ctrl-C interrupts in the read() function interferes with top-level prompt
`ask` is a function that takes in user input string and returns it. However, if one types Ctrl-C while being prompted, the top-level prompt gets replaced by the string `"Type something: "`

```
i1 : ask := () -> (
       val := read("Type something: ");
       val
     )

o1 = -*Function[stdio:1:11-3:3]*-

o1 : FunctionClosure

i2 : ask()
Type something: hello

o2 = hello

i3 : ask()
Type something: ^C
stdio:2:10:(3):[1]: error: failed to read file : interrupted
Type something: 1+2

o4 = 3
Type something: exit
```

pepperhdj:
This resolves #1160 by changing the `stdIO.prompt` back to `oldprompt` before matching on the user input, so as to ensure that even if an error is thrown, the prompt gets changed back.

### DIFF
--- a/M2/Macaulay2/d/actors4.d
+++ b/M2/Macaulay2/d/actors4.d
@@ -841,10 +841,11 @@ readfun(e:Expr):Expr := (
 	  oldprompt := stdIO.prompt;
 	  stdIO.prompt = readpromptfun;
 	  r := getLine(stdIO);
-	  when r is e:errmsg do buildErrorPacket(e.message)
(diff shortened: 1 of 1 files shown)


DanGrayson:
Great fix, thank you!

PR #1163, opened 2020-05-15
Define ComputationStatusCode and RawStatusCodes directly in engine.h and gb.m2

### LINKED ISSUE #1162: statuscodes no longer installed
I was running some tests on the draft of the Debian package, based on the eigen branch, and got the following error:

```
/usr/share/Macaulay2/Core/gb.m2:4:12:(1):[7]: error: opening input file "/usr/share/Macaulay2/Core/statuscodes" failed: No such file or directory
/usr/share/Macaulay2/Core/setup.m2:113:27:(1):[6]: --back trace--
/usr/share/Macaulay2/Core/setup.m2:106:18:(1):[5]: --back trace--
/usr/share/Macaulay2/Core/setup.m2:126:41:(1):[4]: --back trace--
/usr/share/Macaulay2/Core/setup.m2:129:6:(1):[3]: --back trace--
/usr/share/Macaulay2/Core/setup.m2:147:1:(1):[2]: --back trace--
/build/macaulay2-1.14.0.1.780.g85cb3fc75/M2/Macaulay2/d/startup.m2.in:446:11:(0):[2]: --back trace--
/build/macaulay2-1.14.0.1.780.g85cb3fc75/M2/Macaulay2/d/startup.m2.in:620:14:(0):[1]: --back trace--
/build/macaulay2-1.14.0.1.780.g85cb3fc75/M2/Macaulay2/d/startup.m2.in:617:35:(0):[1]: --back trace--
```

This was the code that was running at the time:
```
skip = {"Macaulay2Doc"}
pkgs = separate_" " version#"packages"
scan(pkgs, pkg -> if not member(pkg, skip) then check pkg)
```
(I'm not sure which package it was checking at the time, and the chroot that was set up for the build has already been deleted...)

It looks like `statuscodes` was moved in 04192c0.

mahrud:
Closes #1162 by eliminating the need for a separate `statuscode` file. @mikestillman given that the status codes haven't changed in over a decade, is this change okay with you?

Commented documentation is added in both `engine.h` and `gb.m2` so that any future chaznges will be made in sync.

Also, I added a line in .clang-format to allow indenting after `extern "C" {`, so that it doesn't ask for all of `engine.h` to be unindented.

### DIFF
--- a/M2/.clang-format
+++ b/M2/.clang-format
@@ -26,6 +26,7 @@ BraceWrapping:
   AfterClass:      true
   AfterControlStatement: true
   AfterEnum:       false
+  AfterExternBlock: true
   AfterFunction:   true
   AfterNamespace:  false
   AfterObjCDeclaration: false
--- a/M2/Macaulay2/d/Makefile.files.in
+++ b/M2/Macaulay2/d/Makefile.files.in
@@ -121,7 +121,6 @@ actors5.o basic.o : \
 	@srcdir@/../e/newdelete.hpp \
 	@srcdir@/../d/M2mem.h \
 	@srcdir@/../d/debug.h \
-	../../include/M2/statuscodes.h \
 	../../include/M2/config.h
 
 # now assemble the files into various categories:
--- a/M2/Macaulay2/d/Makefile.in
+++ b/M2/Macaulay2/d/Makefile.in
@@ -72,11 +72,8 @@ LIBFILES += gmp_aux.o arithmetic.o atomic.o M2.o system.o strings.o varstrin.o e
 libd.a: $(LIBFILES) ; ar -r $@ $?
 rm-libd.a:; rm libd.a
 
-../../include/M2/statuscodes.h: @srcdir@/../m2/statuscodes; (sed 's/\(.*\):\(.*\):\(.*\)/  \2 = \1,/' ; echo "  DUMMY_STATUS_CODE") <$< >$@
-
 clean:: ; rm -f $(M2_MADECFILES)
 xml-c.o: xml-c.h xml-exports.h
-engine.o: ../../include/M2/statuscodes.h
 M2lib.o M2inits.o basic.o equality.o actors4.o actors5.o interface.o interface2.o engine.o : CPPFLAGS += -I@srcdir@/../e
 ifeq (@PARI@,yes)
 pari.o: CFLAGS += -Wno-shadow
--- a/M2/Macaulay2/e/Makefile.in
+++ b/M2/Macaulay2/e/Makefile.in
@@ -27,8 +27,7 @@ TAGS tags: @srcdir@/TAGS
 @srcdir@/TAGS: Makefile.in @srcdir@/*.hpp @srcdir@/*.h @srcdir@/*.cpp @srcdir@/*.c \
 		 @srcdir@/f4/*.hpp @srcdir@/f4/*.cpp \
 		 @srcdir@/bibasis/*.hpp @srcdir@/bibasis/*.cpp \
-		 @srcdir@/unit-tests/*.cpp \
-		 @srcdir@/../m2/statuscodes
+		 @srcdir@/unit-tests/*.cpp
 	@ echo " ** making etags file"
 	@ cd @srcdir@ && @ETAGS@ -o TAGS $(patsubst @srcdir@/%, %, $^)
 
--- a/M2/Macaulay2/e/engine.h
+++ b/M2/Macaulay2/e/engine.h
@@ -1644,30 +1644,43 @@ extern "C" {
   /**** Groebner basis and resolution routines ******/
   /**************************************************/
 
-enum ComputationStatusCode
-{
-  /* include/M2/statuscodes.h is generated from Macaulay2/m2/statuscodes */
-#include "M2/statuscodes.h"
-};
+  enum ComputationStatusCode {
+    /* Keep this enum in sync with RawStatusCodes in Macaulay2/m2/gb.m2 */
+    COMP_NEED_RESIZE = 1,         /* need resize */
+    COMP_ERROR = 2,               /* error */
+    COMP_INTERRUPTED = 3,         /* interrupted */
+    COMP_NOT_STARTED = 4,         /* not started */
+    COMP_INITIAL_STOP = 5,        /* StopBeforeComputation */
+    COMP_DONE = 6,                /* done */
+    COMP_DONE_DEGREE_LIMIT = 7,   /* DegreeLimit */
+    COMP_DONE_LENGTH_LIMIT = 8,   /* LengthLimit */
+    COMP_DONE_SYZYGY_LIMIT = 9,   /* SyzygyLimit */
+    COMP_DONE_PAIR_LIMIT = 10,    /* PairLimit */
+    COMP_DONE_GB_LIMIT = 11,      /* BasisElementLimit */
+    COMP_DONE_SYZ_LIMIT = 12,     /* SyzygyLimit */
+    COMP_DONE_CODIM = 13,         /* CodimensionLimit */
+    COMP_DONE_MIN_GENS = 14,      /* StopWithMinimalGenerators */
+    COMP_DONE_STEPS = 15,         /* StepLimit */
+    COMP_DONE_SUBRING_LIMIT = 16, /* SubringLimit */
+    COMP_COMPUTING = 17,          /* computing */
+    COMP_OVERFLOWED = 18,         /* overflowed */
+  };
 
-enum StrategyValues
-  {
+  enum StrategyValues {
     STRATEGY_LONGPOLYNOMIALS = 1,
     STRATEGY_SORT = 2,
     STRATEGY_USE_HILB = 4,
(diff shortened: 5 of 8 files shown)


mahrud:
Related to this, I've seen enums located in two ways:
1. all enums in the same `enums.h` file
2. enums related to each class in that class file

We could go either way.

DanGrayson:
This change turns an automated task into a manual task, so it's not an improvement.

mahrud:
What is the "manual task"? The enum hasn't changed in 10 years, so why would hardcoding it be a problem? The improvement is reducing the complexity of the build process and increasing readability.

DanGrayson:
The manual task is, when making a change, making it in two spots instead of in one.

DanGrayson:
And you're proposing to eliminate a small, easily understood step in a build process that has worked for decades, at the cost of introducing a new avenue for future human error to introduce a bug.

mikestillman:
I think the reason @mahrud is trying to change this, is because the current code caused a build error in a linux distribution?

mikestillman:
How about instead: define this type in the engine, and have an interface function which obtains the strings and integer values from the engine directly?

DanGrayson:
Sounds good to me.

DanGrayson:
> I think the reason @mahrud is trying to change this, is because the current code caused a build error in a linux distribution?

Maybe : that was #1162.

mahrud:
@mikestillman I thought about adding a raw method implemented in `engine.cpp` that would return the appropriate string for each enum key, which can then be called from `gb.m2`, but that would be just moving the code from `gb.m2` to `e`.

One way this could work is if we changed the top level strings to match the keys in the enum (e.g. use `"COMP_DONE"` instead of `"done"`), then have some macro print the exact key string that is used in `engine.h`. I like this more, but would involve more change in `res.m2` and `gb.m2`.

I still don't understand why this is any different than everything else has to be in sync between `e` and `d` and `m2`.

DanGrayson:
> I still don't understand why this is any different than everything else has to be in sync between `e` and `d` and `m2`.

It's not.  We always try to guard against future error.

mahrud:
@DanGrayson Look at the very next enum in `engine.h`:
https://github.com/Macaulay2/M2/blob/e8b2e860a1638bf5d21a2e1bb0c09984eda0d6a6/M2/Macaulay2/e/engine.h#L1653-L1659
And look at this bit in `gb.m2`:
https://github.com/Macaulay2/M2/blob/e8b2e860a1638bf5d21a2e1bb0c09984eda0d6a6/M2/Macaulay2/m2/gb.m2#L147-L151
`STRATEGY_USE_HILB` is missing. Is it better to add another step to the build process to create `strategycodes.h`, or should we just fix it? I don't think the build system is supposed to be a Rube Goldberg machine.

What if we store all enums like this in `e/enums.h` and `m2/enums.m2` and trust that future programmers would read the comments before making a change?

DanGrayson:
> @mikestillman I thought about adding a raw method implemented in `engine.cpp` that would return the appropriate string for each enum key, which can then be called from `gb.m2`, but that would be just moving the code from `gb.m2` to `e`.

But Mike's idea is to have all the code in the same place, in a single file, which would be fine.  It  might look like this:
```
enum ComputationStatusCode
{
  COMP_NEED_RESIZE = 1,
  COMP_ERROR = 2,
  COMP_INTERRUPTED = 3,
  COMP_NOT_STARTED = 4,
  COMP_INITIAL_STOP = 5,
  COMP_DONE = 6,
  COMP_DONE_DEGREE_LIMIT = 7,
  COMP_DONE_LENGTH_LIMIT = 8,
  COMP_DONE_SYZYGY_LIMIT = 9,
  COMP_DONE_PAIR_LIMIT = 10,
  COMP_DONE_GB_LIMIT = 11,
  COMP_DONE_SYZ_LIMIT = 12,
  COMP_DONE_CODIM = 13,
  COMP_DONE_MIN_GENS = 14,
  COMP_DONE_STEPS = 15,
  COMP_DONE_SUBRING_LIMIT = 16,
  COMP_COMPUTING = 17,
  COMP_OVERFLOWED = 18,
  DUMMY_STATUS_CODE
};

const char * status_description (enum ComputationStatusCode i) {
  switch (i) {
  case COMP_NEED_RESIZE : return "need resize";
  case COMP_ERROR : return "error";
    ...

  default : assert (false) ;
      }
}
```

DanGrayson:
... and the assignments in the enum could be omitted.

mahrud:
Code should always go in `.cpp` file, not header files. This slows down compilation and execution. There are exceptions, like template functions, I don't think this isn't one. Also, `DUMMY_STATUS_CODE` is unnecessary. The last enum can end in comma.

mahrud:
Not sure if you saw my previous comment about `strategyCodes`, but looks like the Hilbert function strategy was removed in commit https://github.com/Macaulay2/M2/commit/ea69a838e397fa08be7d68e0a78886c654508a29.

DanGrayson:
> @DanGrayson Look at the very next enum in `engine.h`:
> https://github.com/Macaulay2/M2/blob/e8b2e860a1638bf5d21a2e1bb0c09984eda0d6a6/M2/Macaulay2/e/engine.h#L1653-L1659
> 
> 
> And look at this bit in `gb.m2`:
> https://github.com/Macaulay2/M2/blob/e8b2e860a1638bf5d21a2e1bb0c09984eda0d6a6/M2/Macaulay2/m2/gb.m2#L147-L151
> 

Let's fix that, too!

> 
> `STRATEGY_USE_HILB` is missing.

That's exactly the kind of error we're trying to guard against.

> Is it better to add another step to the build process to create `strategycodes.h`, or should we just fix it? I don't think the build system is supposed to be a Rube Goldberg machine.
> What if we store all enums like this in `e/enums.h` and `m2/enums.m2` and trust that future programmers would read the comments before making a change?

Those files are too far distant in the directory structure, and it's illogical to have engine data in the m2 directory.  If the data were in e/strategycodes.h and e/strategycodes.cpp, that would be much better.  And a comment in the former referring to the latter.

mikestillman:
What is happening on this request?

DanGrayson:
How about a slight modification of Mahrud's latest proposal: "store all enums like this in e/enums.h and e/enums.m2"?

DanGrayson:
Here's a way to have all the code in a single file, which can be called `e/statuscodes.h`:
```
enum ComputationStatusCode
{
  COMP_NEED_RESIZE,
  COMP_ERROR,
  COMP_INTERRUPTED,
  COMP_NOT_STARTED,
  COMP_INITIAL_STOP,
  COMP_DONE,
  COMP_DONE_DEGREE_LIMIT,
  COMP_DONE_LENGTH_LIMIT,
  COMP_DONE_SYZYGY_LIMIT,
  COMP_DONE_PAIR_LIMIT,
  COMP_DONE_GB_LIMIT,
  COMP_DONE_SYZ_LIMIT,
  COMP_DONE_CODIM,
  COMP_DONE_MIN_GENS,
  COMP_DONE_STEPS,
  COMP_DONE_SUBRING_LIMIT,
  COMP_COMPUTING,
  COMP_OVERFLOWED
};

#ifdef STATUS_DESCRIPTION
const char * status_description (enum ComputationStatusCode i) {
  switch (i) {
  case COMP_NEED_RESIZE : return "need resize";
  case COMP_ERROR : return NULL;
    ...
  default : assert (false) ;
      }
}
#endif
```

Then we write one little interface function in `d/interface.dd`, which also includes the file after defining STATUS_DESCRIPTION.

Having these things in a single file is good for auditing and modifying.  Having these things in just C++ is good because the function can use the identifiers rather than the corresponding numbers.

DanGrayson:
Here's an improvement to what I posted above, which ensures that the compiler will tell us if we omit a case of the enum.  Write the code in this style:
```
enum e {a,b,c};
char * f (int i) {
     switch ((enum e)i) {
	  case a: return "a";
	  case b: return "b";
	  case c: return "c";
	  }
     return 0;
     }
```

DanGrayson:
It would be good to do this now, to simplify the build process, as Mahrud pointed out.

mahrud:
The issue behind this PR should be fixed now, and I don't think the suggestion above is better than what already exists for two reasons:
- it involves creating garbage collected memory to pass a string to Core
- it involves fixing top level strings in the engine (.e.g. `"done"`)
Also, it now has merge conflicts so I'm closing this again. I've implemented the same change in the cmake branch to simplify the build process. I suggest implementing it as it is, but keeping it in mind for future engine improvement.

DanGrayson:
> The issue behind this PR should be fixed now, and I don't think the suggestion above is better than what already exists for two reasons:
> 
> * it involves creating garbage collected memory to pass a string to Core

Top level strings are allocated by gc anyway.

> * it involves fixing top level strings in the engine (.e.g. `"done"`)

Those descriptions are descriptions chosen by the engine designer (Mike), and properly belong to the engine.

>   Also, it now has merge conflicts so I'm closing this again. I've implemented the same change in the cmake branch to simplify the build process. I suggest implementing it as it is, but keeping it in mind for future engine improvement.

What future improvement?

PR #1170, opened 2020-05-18
Output version number with autoconf

d-torrance:
In c0e33c4, the version number was moved to a separate VERSION file and
assigned to a shell variable by autoconf.  However, autoconf will only
substitute output variables during AC_OUTPUT, not all shell variables.
Indeed, from config.log:

```
| #define PACKAGE_VERSION "NO_VERSION_NUMBER"
```

And when running Macaulay2:

```
Macaulay2, version NO_VERSION_NUMBER
```

Instead, we use a solution from https://stackoverflow.com/questions/8559456.
Note that we no longer need to use $srcdir to find VERSION.  It will work
from any build directory.


DanGrayson:
Thanks for catching that!

PR #1172, opened 2020-05-19
Update javascript libraries

d-torrance:
Three of the Javascript libraries used by the Visualize package (BootSideMenu, clipboard.js, and three.js) are far enough out of date that it does not work with the versions of these libraries currently in Debian (or, in the case of BootSideMenu, the version which should be in Debian soon).  Debian does not want embedded copies of libraries included in any of its packages, and so the Macaulay2 package will end up using the Debian packages of these libraries.

I've made patches to Visualize to work with the newer versions of these libraries for the draft of the Debian package and figured I would propose them upstream.

See the individual commit messages for more specific details.


bstone:
These changes look good to me. Thanks. 

@d-torrance , the development repo for visualize is here: 

https://github.com/bstone/Visualize-M2

Would you be willing to fork the repo and copy/paste your `Visualize` folder and `Visualize.m2` file?

d-torrance:
> Would you be willing to fork the repo and copy/paste your `Visualize` folder and `Visualize.m2` file?

Will do!

PR #1131, opened 2020-05-09
Restore missing diagrams from MonomialAlgebras package.

d-torrance:
The links have been broken since at least 2015.  Fortunately, copies from
2013 were still available on the Internet Archive Wayback Machine. We
now include them with the source.

This also has the advantage of fixing a privacy-breach-generic Lintian
warning in the Debian package:

> This package creates a potential privacy breach by fetching data from
> an external website at runtime. Please remove these scripts or
> external HTML resources.

> Please replace any scripts, images, or other remote resources with
> non-remote resources. It is preferable to replace them with text and
> links but local copies of the remote resources are also acceptable as
> long as they don't also make calls to remote services. Please ensure
> that the remote resources are suitable for Debian main before making
> local copies of them.

Closes: #402

PR #1176, opened 2020-05-19
Properly restore missing diagrams from MonomialAlgebras package.

d-torrance:
In the previous attempt (#1131), an absolute path using prefixDirectory
was used for the src attribute of each img tag.  This was hardcoded into
the documentation during build, and so the image links were still broken.

We now skip using prefixDirectory and let html.m2 compute the proper
relative path to the images.


DanGrayson:
Thanks.  I trust this one will work!

d-torrance:
It should!  Here's a snippet from the generated html:

```

<p><img src="../../../../Macaulay2/MonomialAlgebras/projr_33.jpg" alt=""/></p>
--
  | <p><img src="../../../../Macaulay2/MonomialAlgebras/projc_33.jpg" alt=""/></p>
  | <p><img src="../../../../Macaulay2/MonomialAlgebras/projd_33.jpg" alt=""/></p>
  | <p><img src="../../../../Macaulay2/MonomialAlgebras/EGplane_33.jpg" alt=""/></p>
  | <p><img src="../../../../Macaulay2/MonomialAlgebras/EGbar_33.jpg" alt=""/></p>


```

That relative path should always work to get from the html directory to the package directory where the images are located, regardless of where everything is installed.

DanGrayson:
Sounds good.  Do the images appear when you view the html in a browser?

d-torrance:
Yes

DanGrayson:
Great!

PR #1178, opened 2020-05-20
Get the source code for functions defined in startup.m2.in.

d-torrance:
This file only exists in the Macaulay2 source code and is not installed.
However, its contents are available as a string, and we can still
access it.  The code to do this exists in code.m2, but was still
set up for when these functions were defined in layout.m2, which hasn't
existed since fb2a04e.

We also restore the SoureCode option to applicationDirectory's
documentation.  It was removed in 728c632 because of this bug.

### DIFF
--- a/M2/Macaulay2/packages/Macaulay2Doc/doc13.m2
+++ b/M2/Macaulay2/packages/Macaulay2Doc/doc13.m2
@@ -232,7 +232,7 @@ document { Key => {applicationDirectory, "application directory"},
 	  },
      PARA { "The function ", TO "applicationDirectorySuffix", " determines the value of ", TT "applicationDirectory", ", and can be modified by the user." },
      EXAMPLE "applicationDirectory()",
-     SeeAlso => applicationDirectorySuffix}
+     SeeAlso => "applicationDirectorySuffix"}
 
 document {
      Key => installedPackages,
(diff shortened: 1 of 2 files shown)


M2/Macaulay2/packages/Macaulay2Doc/doc13.m2
```diff
@@ -225,6 +225,7 @@ document { Key => {applicationDirectory, "application directory"},
      Headline => "the path to the user's application directory",
      Usage => "applicationDirectory()",
      Outputs => { String => "the path to the user's application directory" },
+     SourceCode => applicationDirectory,
```
DanGrayson, line 228: You can't put this line back in, because browsers can look only in files, not in internal strings.

d-torrance: But isn't the html documentation all static and generated during build?

DanGrayson: Yes, but is the file there at the other end of the link?  Hmm, maybe we could arrange for it to be there...

d-torrance: Do you mean a link to the source file? Right now, there is no link.  It's just the output of `code` inside a `<pre>` environment.

DanGrayson: Oh, you're right, thanks.  It also appears that way here: https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.15/share/doc/Macaulay2/Macaulay2Doc/html/_application__Directory.html

DanGrayson: Ack, I just noticed that `SeeAlso => applicationDirectorySuffix` needs to be changed to  `SeeAlso => "applicationDirectorySuffix"`.  My mistake from long long ago.  Would you like to add that, too?

d-torrance: Sure thing!

PR #1179, opened 2020-05-20
New options added and bugs fixed in ConwayPolynomials

zeromath:
Changes:
- Added new option `Variable` to `conwayPolynomial`. Now it supports to use user-designated variable as the polynomial variable.
- Updated `map(GaloisField,GaloisField)`. Now it works properly between `GaloisFields` defined with user-designated variable.
- Documentation updates.

### DIFF
--- a/M2/Macaulay2/packages/ConwayPolynomials.m2
+++ b/M2/Macaulay2/packages/ConwayPolynomials.m2
@@ -10,19 +10,17 @@ newPackage(
     	Headline => "a database of Conway polynomials"
     	)
 -- the data comes libflint
-export {
-    "conwayPolynomial",
-    -- options to set the variable to use
-    "Variable"
-    }
+export "conwayPolynomial"
 rawConwayPolynomial := value Core#"private dictionary"#"rawConwayPolynomial"
 getCP := (p,n) -> rawConwayPolynomial (p,n,false)
 Ap := memoize((p, a) -> (ZZ/p)(monoid [a]))
 fix := (p,n,co,a) -> sum(#co, i -> co#i * a^i)
-conwayPolynomial = method(Options=>{Variable=>getSymbol "a"})
+conwayPolynomial = method(Options=>{Variable=>null})
 conwayPolynomial(ZZ,ZZ) := opts -> (p,n) -> (
      cp := getCP(p,n);
-     if cp != {} then fix(p,n,cp,(Ap(p, opts.Variable))_0))
+     var := opts.Variable;
+     var = if var === null then getSymbol "a" else baseName var;
+     if cp != {} then fix(p,n,cp,(Ap(p, var))_0))
 conwayPolynomial ZZ := opts -> q -> (
      factors := factor q;
      if #factors =!= 1 or factors#0#0 === -1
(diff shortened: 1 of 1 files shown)


M2/Macaulay2/packages/ConwayPolynomials.m2
```diff
@@ -10,24 +10,28 @@ newPackage(
     	Headline => "a database of Conway polynomials"
     	)
 -- the data comes libflint
-export "conwayPolynomial"
+export {
+    "conwayPolynomial",
+    -- options to set the variable to use
+    "Variable"
```
DanGrayson, line 16: "Variable" is a symbol in the Core package, so just use it -- don't export a new one with the same name.

M2/Macaulay2/packages/ConwayPolynomials.m2
```diff
@@ -10,24 +10,28 @@ newPackage(
...
+    }
 rawConwayPolynomial := value Core#"private dictionary"#"rawConwayPolynomial"
 getCP := (p,n) -> rawConwayPolynomial (p,n,false)
-Ap := memoize(p -> (ZZ/p)(monoid [getSymbol "a"]))
+Ap := memoize((p, a) -> (ZZ/p)(monoid [a]))
 fix := (p,n,co,a) -> sum(#co, i -> co#i * a^i)
-conwayPolynomial = method()
-conwayPolynomial(ZZ,ZZ) := (p,n) -> (
+conwayPolynomial = method(Options=>{Variable=>getSymbol "a"})
```
DanGrayson, line 22: I think that if you run  ' installPackage "ConwayPolynomials" ', this change will lead to a problem -- the error message will be something like " can't determine package of symbol a ".  The solution will be to let the default value be the string "a", and run getSymbol later.

DanGrayson:
This is great, thank you.  What is your name?

Could you add some tests to ensure that everything works as expected?

Actually, we have a big pull request coming up, so it might be good to add the test after we merge it, because it moves all the test files to another directory.

M2/Macaulay2/packages/ConwayPolynomials.m2
zeromath: Gotcha! Just changed!

M2/Macaulay2/packages/ConwayPolynomials.m2
zeromath: I rewrite this part by following the code in `galois.m2`

zeromath:
> This is great, thank you. What is your name?
> 
> Could you add some tests to ensure that everything works as expected?
> 
> Actually, we have a big pull request coming up, so it might be good to add the test after we merge it, because it moves all the test files to another directory.

I'm Zhan Jiang. I was working in the FastLinAlg group during the just past workshop (virtually in Cleveland). I just added one test case.

DanGrayson:
I just remembered that the monoid ring construction function does "getSymbol" for us:
```

i10 : newPackage "Foo" ; export "foo" ; foo = () -> ZZ["a"] ; endPackage "Foo"

o13 = Foo

o13 : Package

i14 : foo()

o14 = ZZ[a]

o14 : PolynomialRing

i15 : baseName oo_0

o15 = a

o15 : Symbol

i16 : package oo

o16 = User

o16 : Package
```

zeromath:
I removed `getSymbol` function and everything seems to work fine.

zeromath:
Here is what I mean in the slack chat
```
i1 : R = GF(4)[x]; S = GF(8)[y]; f = map(S,R,{y})

o3 = map(S,R,{y, a})

o3 : RingMap S <--- R
```

The generator of the `GF(4)` in R is not mapped to S properly by the natural inclusion map `GF(4) -> GF(8)`, it's simply mapped to the generator of `GF(8)`. This is actually a related problem (bug?), but more fundamental.

I think in `ringmap.m2` there is a chunk of code checking whether the source has `coefficientRing` being `GaloisField`, trying to map the `PrimitiveElement` to the correct place. But it ends up using the naive "mapping to the generator" approach. Should we create an issue about this?

DanGrayson:
Ah.  Fixing that would be a good improvement, but beyond the scope of your pull request.  Why don't you create a new issue for that, and volunteer for it, if you like?

DanGrayson:
Pulling now.  Go ahead and make a separate issue for the other problem, and refer to #1179

PR #1191, opened 2020-05-23
Run M2 with -q during 'make check'

d-torrance:
Otherwise, we try to create ~/.Macaulay2 but we may not have permission
to create this directory.  For example, during a test build of the
Debian package:

    ../m2/files.m2:18:30:(1):[16]: error: can't make directory
    "/nonexistent": Permission denied


DanGrayson:
Ah, good catch!  But I think it should be "-q" instead, as that's the option that says : don't use the user's home directory for anything.

d-torrance:
Sounds good!  I think either option works, but `-q` is more consistent will all the other checks anyway.

DanGrayson:
Thanks!

PR #1192, opened 2020-05-24
Updated the FastLinAlg package

kschwede:
In particular, this adds the isCodimAtLeast method, improves documetnation and fixes some minor bugs.


M2/Macaulay2/packages/FastLinAlg.m2
```diff
@@ -15,7 +15,7 @@ Version => "0.1", Date => "October 21st, 2019", Authors => {
     Email=> "yuhuiyao4ever@gmail.com"
     }
 }, --this file is in the public domain
-Headline => "faster linear algebra operations")
+Headline => "A package for faster linear algebra operations.", DebuggingMode => true, Reload=>true)
```
DanGrayson, line 18: Please leave DebuggingMode and Reload off.

M2/Macaulay2/packages/FastLinAlg.m2
```diff
@@ -77,7 +98,7 @@ StrategyDefault := new HashTable from {
 --protect nanosleep;
 --if (not isGlobalSymbol "nanosleep") then protect getSymbol("nanosleep");
 
-StrategyCurrent = new MutableHashTable from {LexLargest=>0, LexSmallestTerm => 50, LexSmallest=>50, GRevLexSmallestTerm=>0, GRevLexSmallest=>0, GRevLexLargest=>0,Random=>0,RandomNonzero=>0};
```
DanGrayson, line 80: Here you intend to override some values from `StrategyDefault`, so it would be good to write 
```
StrategyCurrent = StrategyDefault ++ {LexLargest=>0, LexSmallestTerm => 50, LexSmallest=>50, GRevLexSmallestTerm=>0, GRevLexSmallest=>0, GRevLexLargest=>0,Random=>0,RandomNonzero=>0};
```
, for you could omit from the list those items with the same values.

M2/Macaulay2/packages/FastLinAlg.m2
```diff
@@ -612,6 +649,19 @@ RnReductionP(ZZ, Ring, ZZ):= opts -> (n1, R1, p)-> (
     return Rn(n1, ambRing/Id, opts);
 );
 
+verifyStrategy := (passedStrat) -> (
+    if not ((class passedStrat === HashTable) or ((class passedStrat) === MutableHashTable)) then return false;
+    if not (passedStrat #? LexLargest) then return false;
```
DanGrayson, line 654: Rewrite using the logical operator `and`, without `if` or `return`.

M2/Macaulay2/packages/FastLinAlg.m2
```diff
@@ -710,10 +760,14 @@ Rn(ZZ, Ring) := opts -> (n1, R1) -> (
...
-    if (opts.MaxMinors === null) then (
-        numberOfMinorsCompute = 10*minNumberToCutDown + 8*ceiling(log_1.3(possibleMinors));)
+    if (instance (opts.MaxMinors, Function)) then (
+        numberOfMinorsCompute = opts.MaxMinors(minNumberToCutDown, possibleMinors); )
+        --5*minNumberToCutDown + 8*ceiling(log_1.3(possibleMinors));)
+    else if (instance(opts.MaxMinors, Number)) then (
+        numberOfMinorsCompute = opts.MaxMinors;)
     else (
-        numberOfMinorsCompute = opts.MaxMinors;);
+        numberOfMinorsCompute = 5*minNumberToCutDown + 8*ceiling(log_1.3(possibleMinors));
```
DanGrayson, line 769: At this point you should signal an error if it isn't `null`.  (Users can make mistakes.)

M2/Macaulay2/packages/FastLinAlg.m2
```diff
@@ -781,10 +844,11 @@ chooseGoodMinors(ZZ, ZZ, Matrix) := opts -> (howMany, minorSize, M1) -> (
     i := 0; -- how many times through the loop we go
     k := 0; -- how many actual minors we found
     j := 3; -- controlling when to reset the grevlex matrix
-    maxAttempts := 5*log_2(possibleMinors) + 10; --the absolute max number of attempts
+    maxAttempts := 10*log_1.3(possibleMinors) + 10; --the absolute max number of attempts
     R2 := reorderPolynomialRing(GRevLex, ambR);
     f := map(R2, ambR);
     sumMinors := trim ideal(sub(0, ambR));
+    --1/0;
```
DanGrayson, line 851: This doesn't look useful.

M2/Macaulay2/packages/FastLinAlg.m2
```diff
@@ -799,6 +863,10 @@ chooseGoodMinors(ZZ, ZZ, Matrix) := opts -> (howMany, minorSize, M1) -> (
             );
             i = i+1;
         );
+        if not (opts.PeriodicCheckFunction === null) then (
```
DanGrayson, line 866: You could use `=!=` here.

M2/Macaulay2/packages/FastLinAlg.m2
```diff
@@ -812,26 +880,39 @@ chooseGoodMinors(ZZ, ZZ, Matrix) := opts -> (howMany, minorSize, M1) -> (
...
-    theseMinors := chooseGoodMinors(minorsCount, firstRank, firstDiff, goodMinorsOptions);
+    else if instance(opts.MaxMinors, Function) then (
+        if (debugLevel > 0) or opts.Verbose then print "projDim:  Using passed minors function.";
+        minorsCount = (opts.MaxMinors)(myDim, possibleMinors);)        
+    else (
+        if (debugLevel > 0) or opts.Verbose then print "projDim:  Using default max minors function."; 
+        minorsCount = 5*myDim + 2*log_1.3(possibleMinors); );
+--        print class opts.MaxMinors;
+
+--    1/0;
```
DanGrayson, line 911: This, too.

M2/Macaulay2/packages/FastLinAlg.m2
```diff
@@ -855,6 +940,71 @@ projDim(Module) := opts -> (N1) -> (
     return i;
 );
 
+isGBDone := (myGB) -> (
+    --a temporary function for finding out if a gb computation is done.
+    myStr := status myGB;
+    if (0 < #select("status: done", myStr)) then return true else return false;
```
DanGrayson, line 946: @mikestillman -- is there a better way to do this?  If not, we should invent one.

M2/Macaulay2/packages/FastLinAlg.m2
```diff
@@ -855,6 +940,71 @@ projDim(Module) := opts -> (N1) -> (
...
+);
+
+
+
+isCodimAtLeast = method(Options => optIsCodimAtLeast);
+
+isCodimAtLeast(ZZ, Ideal) := opts -> (n1, I1) -> (
+    R1 := ring I1;
+    S1 := ambient R1;
+    if (not isPolynomialRing(S1)) then error "isCodimAtLeast:  This requires an ideal in a polynomial ring, or in a quotient of a polynomial ring.";
```
DanGrayson, line 956: Lots of redundant parentheses in these lines (and elsewhere).  Example:
```
if (codim monomialIdeal(I1) >= n1) then return true;
```
could be
```
if codim monomialIdeal I1 >= n1 then return true;
```

M2/Macaulay2/packages/FastLinAlg.m2
```diff
@@ -855,6 +940,71 @@ projDim(Module) := opts -> (N1) -> (
...
+        gensList = first entries leadTerm gb(I2, PairLimit=>curLimit);    
+        if (#gensList > 0) then (
+            monIdeal = monomialIdeal(first entries leadTerm myGB);
+            if (opts.Verbose or debugLevel > 0) then print concatenate("isCodimAtLeast: computed gb, now computing codim ");
+            if (codim monIdeal - dAmb >= n1) then return true;
+        );
+        if (isGBDone(myGB)) then i = opts.PairLimit;
+        i = i + 1;
+    );
+    return null;
```
DanGrayson, line 998: This line can be omitted.

M2/Macaulay2/packages/FastLinAlg.m2
```diff
@@ -855,6 +940,71 @@ projDim(Module) := opts -> (N1) -> (
...
+        if (opts.Verbose or debugLevel > 0) then print concatenate("isCodimAtLeast: Computing codim of monomials based on ideal generators.");
+        if (codim monIdeal - dAmb >= n1) then return true;
+        if (opts.Verbose or debugLevel > 0) then print concatenate("isCodimAtLeast: Didn't work, going to find the partial Groebner basis.");
+    );
+    --now we set stuff up for the loop if that doesn't work.  
+    vCount := # first entries vars S1;
+ --   baseDeg := apply(sum(apply(first entries vars S1, t1 -> degree t1)), v -> ceiling(v/vCount)); --use this as the base degree to step by (probably we should use a different value)
+    i := 1;
+    --curLimit := baseDeg;
+    local curLimit;
```
DanGrayson, line 980: An alternative to declaration with `local` is to change the first assignments below to `:=`.

M2/Macaulay2/packages/FastLinAlg.m2
```diff
@@ -1269,8 +1425,8 @@ doc ///
         : List
     Description
         Text
-            This function looks at subtrices of the given matrix, and tries to find
-            one of the specified rank.  If it succeeds, it returns a list of two lists.  
+            This function looks at subtmatrices of the given matrix, and tries to find
```
DanGrayson, line 1428: Still spelled wrong!

M2/Macaulay2/packages/FastLinAlg.m2
```diff
@@ -1299,6 +1455,14 @@ doc ///
         (Rn, ZZ, Ring)
         [Rn, Verbose]
         [Rn, ModP]
```
DanGrayson, line 1457: The name "Rn" is too brief!  Think of something descriptive.

M2/Macaulay2/packages/FastLinAlg.m2
```diff
@@ -1569,8 +1742,8 @@ doc ///
         orderType: Symbol
             a valid monomial order, such as {\tt GRevLex}
     Outputs
-        R1: ZZ
-           a polynomial ring
+        b: Boolean
+           true if
```
DanGrayson, line 1746: The thought here was left incomplete.

M2/Macaulay2/packages/FastLinAlg.m2
```diff
@@ -1614,6 +1787,80 @@ doc ///
...
+doc ///
+    Key
+        isCodimAtLeast
+        (isCodimAtLeast, ZZ, Ideal)
+        [isCodimAtLeast, Verbose]
+        [isCodimAtLeast, SPairsFunction]
+        [isCodimAtLeast, PairLimit]
+        SPairsFunction
+    Headline
+        returns true if we can quickly see if the codim is at least a given number
```
DanGrayson, line 1799: change second "if" to "whether"

M2/Macaulay2/packages/FastLinAlg.m2
kschwede: I agree.  I don't know a better way to do it though.

M2/Macaulay2/packages/FastLinAlg.m2
kschwede: I actually strongly prefer the parenethesed version.  I find it much easier to read.

DanGrayson: Okay, that's fine.

M2/Macaulay2/packages/FastLinAlg.m2
kschwede: I've done this one, I'll reply as I've done the others.

M2/Macaulay2/packages/FastLinAlg.m2
kschwede: I think you are misreading this.  The user can pass a function, a number, or anything else.  If they pass anything else, it reverts to a default function.

M2/Macaulay2/packages/FastLinAlg.m2
kschwede: I've removed the commented out --1/0s, throughout.

M2/Macaulay2/packages/FastLinAlg.m2
kschwede: I believe I've done this throughout.

M2/Macaulay2/packages/FastLinAlg.m2
kschwede: Done.

M2/Macaulay2/packages/FastLinAlg.m2
kschwede: I realize that, but my authors and I thought this way would be easeir to read.

M2/Macaulay2/packages/FastLinAlg.m2
kschwede: The first assignments appear inside a loop (or at least did initially).  Won't it cause issues that way?  (or am I misremembering).  We don't want to do := assignments repeatedly at each step inside the loop, right?

M2/Macaulay2/packages/FastLinAlg.m2
kschwede: RegularInCodimensionN?
That seemed rather long to type.  
I thought we wanted to avoid things like 
isRn
as a matter of style.  Although that would be another option.

I think Rn is the correct and extensively used mathematical name for this though.

M2/Macaulay2/packages/FastLinAlg.m2
kschwede: Worse, the function doesn't return a Boolean, this is now fixed.

M2/Macaulay2/packages/FastLinAlg.m2
kschwede: Done.

M2/Macaulay2/packages/FastLinAlg.m2
kschwede: I believe I've done this.

M2/Macaulay2/packages/FastLinAlg.m2
kschwede: I think you are looking at the replaced code.  The current code creates a Mutable strategy, based on the default strategy, for the user to play with.  There may be other ways to accomplish this, but this seemed reasonable.

M2/Macaulay2/packages/FastLinAlg.m2
kschwede: Fixed.

DanGrayson:
Thanks!

About StrategyCurrent -- Yes, I didn't understand that in your application the mutability is important.

Re "The user can pass a function, a number, or anything else. If they pass anything else, it reverts to a default function."  The problem is that if they make a mistake, and you regard their value as "something else", then they may not get the desired effect.  It would be better to insist on something specific if they want the default.

About ":=" inside a loop.  No, the declaration of the variable as local inside a loop is a static attribute that is determined when the function is parsed and compiled.  The assignment itself is all that happens at runtime.

Another thing I just noticed : if you check user input (as in `class passedStrat === HashTable`) by insisting that the class of something they give you be equal to a certain thing, then that will prevent them from giving you a class that inherits from that class.  I.e., if you ask for an object of class HashTable, and X is a new type of hashtable, then they can't give you an instance of X.  That may not be important for you, but in almost all cases, the use of "instance" instead will look just as good and allow more flexible use of inheritance.

DanGrayson:
How did you make all those replies to my comments appear as subitems of your latest comment?

kschwede:
I made comments by clicking a box that looks like a "text-message" icon in the upper right hand corner of my github page, when viewing the pull request.
WhereIclicked
That let me respond individually.

Good point about the MaxMinors => null.  It now gives an error if the user doesn't provide the correct input.  

I was confused about the := in a loop.  I've changed it.

For the other instance vs class thing, good point.  I've changed it.

M2/Macaulay2/packages/FastLinAlg.m2
mikestillman: How about SerreRn, SerreSn?

kschwede: That's good!  Although I suppose isSerreSn is redundant.  (Since the S is Sn stands for Serre).  I don't think I have any particularly good ideas about how to implement Sn checking though.  There will be another improvement to this package when we get the randomRationalPoints working as well as we want, but that's probably for 1.17.

DanGrayson: "isSerreRn" is bad for two reasons: it contains "R", which is a symbol with no English meaning; it contains "n", which is just an a arbitrary identifier that could have been any other letter of the alphabet.

kschwede: Hi Dan, 
I guess we could just go with 
regularInCodimension
We presumably want to avoid abbreviations like
regInCodim
regularInCodim
On the other hand "Rn" is really the way this is used in mathematical speech I believe.  Specifically, people will really say things like this scheme/variety/ring is R0 / R1 / R2...
Let me know.

mikestillman: I think `regularInCodimension` is a good (self-dcumenting) name.  The documentation could mention the name`Rn`.  I agree that even though it is standard math usage in commutative algebra, it is too short of a name, making it hard to see, search for, etc.  I guess also: the name of the function should then be `R`, not `Rn`, but that isn't allowed either.

kschwede: Ok, I've renamed it.

DanGrayson:
Are you done updating it?

kschwede:
Yes, I'm done.

M2/Macaulay2/packages/FastLinAlg.m2
```diff
@@ -705,15 +754,19 @@ Rn(ZZ, Ring) := opts -> (n1, R1) -> (
...
-        numberOfMinorsCompute = 10*minNumberToCutDown + 8*ceiling(log_1.3(possibleMinors));)
+    if (instance (opts.MaxMinors, Function)) then (
+        numberOfMinorsCompute = opts.MaxMinors(minNumberToCutDown, possibleMinors); )
+        --5*minNumberToCutDown + 8*ceiling(log_1.3(possibleMinors));)
+    else if (instance(opts.MaxMinors, Number)) then (
+        numberOfMinorsCompute = opts.MaxMinors;)
     else (
-        numberOfMinorsCompute = opts.MaxMinors;);
+        --numberOfMinorsCompute = 5*minNumberToCutDown + 8*ceiling(log_1.3(possibleMinors));
+        error "regularInCodimension:  You must pass MaxMinors either a function or a number.";
```
DanGrayson, line 768: Now you don't have a default number (to be used if "null" is there).  Is that intentional?  You could have it.

M2/Macaulay2/packages/FastLinAlg.m2
```diff
@@ -855,6 +936,70 @@ projDim(Module) := opts -> (N1) -> (
     return i;
 );
 
+isGBDone := (myGB) -> (
+    --a temporary function for finding out if a gb computation is done.
+    myStr := status myGB;
+    if (0 < #select("status: done", myStr)) then return true else return false;
```
DanGrayson, line 942: It's a tautology that
```
if X then true else false
```
is equal to X, so the circumlocution is confusing.

DanGrayson, line 942: It's better to replace
```
0 < #select("status: done", myStr)
```
by
```
match("status: done", myStr)
```

M2/Macaulay2/packages/FastLinAlg.m2
```diff
@@ -1614,50 +1784,124 @@ doc ///
...
+        isCodimAtLeast(n, I)
+    Inputs
+        n: ZZ
+            an integer
+        I: Ideal
+            an ideal in a polynomial ring over a field, or a quotient ring
+    Outputs
+        : Boolean
+           {\tt true} if the codimension of I is at least n
+        :
```
DanGrayson, line 1807: This second output item conveys the idea that the function returns 2 outputs, rather than the idea that the function returns 1 output, which could be of two types.

M2/Macaulay2/packages/FastLinAlg.m2
```diff
@@ -1614,50 +1784,124 @@ doc ///
...
+    Inputs
+        n: ZZ
+            an integer
+        I: Ideal
+            an ideal in a polynomial ring over a field, or a quotient ring of such
+    Outputs
+        : Boolean
+           {\tt true} if the dimension of I is at most n
+        : 
+            {\tt null} if the function cannot tell whether the dimension is at most n
```
DanGrayson, line 1853: same here

M2/Macaulay2/packages/FastLinAlg.m2
kschwede: Hi Dan, it is indeed intentional.  The option has a default value (actually, the same as what null used to force).

M2/Macaulay2/packages/FastLinAlg.m2
kschwede: I see, I guess I don't know what the right way to document that is then.

M2/Macaulay2/packages/FastLinAlg.m2
kschwede: Fixed.

kschwede: Fixed.

M2/Macaulay2/packages/FastLinAlg.m2
DanGrayson: Try this:
```
    Outputs
        : Boolean
           {\tt true} or {\tt false}, indicating whether the codimension of I is at least n, if it can be determined, else {\tt null}
```

We cheat a little, in that "null" is not of type Boolean.

kschwede: Ok, I believe I've made this change throughout.

M2/Macaulay2/packages/FastLinAlg.m2
kschwede: Done.

PR #1193, opened 2020-05-24
Don't update git submodules if we aren't building any

d-torrance:
This new feature was causing failures in building the Debian package
as git is not a build dependency, and the Debian packages for
fflas-ffpack, memtailor, mathic, and mathicgb are used instead of
building them from the submodules.

### DIFF
--- a/M2/GNUmakefile.in
+++ b/M2/GNUmakefile.in
@@ -12,6 +12,9 @@ all install:										\
 ifeq "@DEVELOPMENT@" "yes"
 optional-git-submodule-update:
 	: skipping automatic updating of submodules, because --enable-development was specified
+else ifeq "@BUILDSUBLIST@" ""
+optional-git-submodule-update:
+	: skipping automatic updating of submodules, because no submodules are being built
 else
 optional-git-submodule-update:
 	git submodule update


DanGrayson:
Thanks for the improvement.

But what was the error?  I did an experiment with a freshly cloned repository to see the effect of the update command, and it was nil.

d-torrance:
If git isn't installed (e.g., if we're building from a tarball generated from git elsewhere, which is what the Debian packages does), then "git submodule update" would fail.

DanGrayson:
Is the subdirectory `.git` absent in the tarball?  We could use that also as an indication about whether to update.

d-torrance:
That's a good question!  My script uses `git archive` to generate the tarball, but I wasn't sure what all is included.  Apparently the `.git` directory is *not* there:

```
profzoom@gloria:~/tmp/macaulay2$ tar -xf ~/src/macaulay2/macaulay2_1.15.1.0+git67.80c4d54+ds.orig.tar.xz 
profzoom@gloria:~/tmp/macaulay2$ ls -a
.  ..  bugs  .gitignore  .gitmodules  M2  README
```

PR #1195, opened 2020-05-24
Find gftables during check-configure if they were not built

d-torrance:
Previously, make check in M2/check-configure failed if we were using
gftables that were already present in the system, as it was assuming
that they were built and located in @pre_packagesdir@/Core/factory/.

### DIFF
--- a/M2/Macaulay2/d/startup.m2.in
+++ b/M2/Macaulay2/d/startup.m2.in
@@ -286,7 +286,7 @@ if firstTime then (
 	       "programs" => "@programsdir@/",
 	       "program licenses" => "@licensesdir@/",
 	       "package" => "@packagesdir@/PKG/",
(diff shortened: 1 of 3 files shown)


DanGrayson:
Thanks, but now I notice that we have two variables getting into the config.status substitution table: `GFTABLESDIR` and `gftablesdir`.  I think we need only one, and it should be the upper case version.  Could you take care of that, too?

d-torrance:
Done!

DanGrayson:
I think if you declare it with AC_ARG_VAR then you don't also need to declare it with AC_SUBST, you can just set it.

d-torrance:
I've just pushed a couple more commits getting rid of the `AC_SUBST`'s and also fixing a bug in the `--enable-build-libraries` option I discovered testing this stuff.

DanGrayson:
Thanks!

PR #1196, opened 2020-05-25
minor expression update; major WebApp mode update

pzinn:
These are the proposed changes. All the tests ran fine but don't take my word for it :-P
- Tiny expression update
minor changes that didn't get shipped two years ago during the expression upgrade.
mostly, some hold/unhold stuff. no change in the output.
- Major WebApp update (webapp.m2)
in preparation for the launch, see the current beta at
http://115.146.94.251:8002/
- relatedly, new --webapp option for M2
`M2 --webapp` starts M2 with `topLevelMode=WebApp`
- two new (mutable) variables `texMathStart` and `texMathEnd`
that `tex` inserts in math mode.
The reason is that LaTeX/MathJax/KaTeX use the LaTeX delimiters `\(` and `\)` rather than the TeX ones.
Or one could want `$$` or `\[` `\]`. The default remains `$`.
Needed by webapp.m2


M2/Macaulay2/m2/webapp.m2
```diff
@@ -116,3 +129,53 @@ CoherentSheaf#{WebApp,AfterPrint} = F -> (
  )
 
 ZZ#{WebApp,AfterPrint} = identity
+
+if topLevelMode === WebApp then (
+    (webAppEndTag,            -- closing tag
+	webAppHtmlTag,        -- indicates what follows is HTML
+	webAppOutputTag,      -- it's html but it's output
```
DanGrayson, line 136: Why not always set these to these values, in advance?  And not just when topLevelMode is WebApp?

M2/Macaulay2/m2/exports.m2
```diff
@@ -1247,6 +1247,8 @@ exportMutable {
 	"recursionLimit",
 	"scriptCommandLine",
 	"stopIfError",
+	"texMathStart",
```
DanGrayson, line 1250: Why do these have to be mutable?

pzinn: you mean `texMathStart` and `texMathEnd`? the idea is that in different circumstances you may want different delimiters. The direct application here is that webapp.m2 redefines them for its own purposes, but we still want, when the user types
```
tex(ZZ[x,y,z])
```
the output to have the usual delimiters `$`.

M2/Macaulay2/m2/webapp.m2
pzinn: yeah that's the most debatable change I introduced. Let me think about it again. The idea is that internally the actual web app uses those shortened, somewhat obscure tags (ascii characters < 32) which are then turned into normal html tags. Turning on topLevelMode manually would lead to a sort of "debug mode" where the tags appear as (a rough approximation of) their final appearance.
But the truth is, I don't see any real application for this. Let me think about it some more and possibly revert it.

pzinn: reverted

M2/Macaulay2/m2/expressions.m2
```diff
@@ -1174,16 +1168,17 @@ texMath Table := m -> (
...
-    else if m#?0 then if #m#0>10 then "{\\left(" | texMath(new Table from toList m) | "\\right)}" -- the extra {} is to discourage line breaks
+    else if m#?0 then if #m#0>10 then "{\\left(" | texMathTable m | "\\right)}" -- the extra {} is to discourage line breaks
      else concatenate(
       	      "\\begin{pmatrix}" | newline,
      	      between(///\\/// | newline, apply(toList m, row -> concatenate between("&",apply(row,texMath)))),
 	      "\\end{pmatrix}" -- notice the absence of final \\ -- so lame. no newline either in case last line is empty
 	      )
 	  )
-texMath MatrixDegreeExpression := x -> texMath MatrixExpression x#0 -- degrees not displayed atm
+texMath MatrixDegreeExpression := (lookup(texMath,MatrixExpression))@@first -- degrees not displayed atm
```
DanGrayson, line 1181: It might be better to handle this with inheritance by defining
```
MatrixDegreeExpression = new HeaderType of MatrixExpression
```

M2/Macaulay2/m2/exports.m2
DanGrayson: But since you introduce
```
tex Thing := x -> concatenate(texMathStart,texMath x,texMathEnd)
```
, if the user types
```
tex(ZZ[x,y,z])
```
, he will get your delimiters, not the usual ones.

What is the purpose your web app has in mind in trying to redefine that?

pzinn: No, it won't. texMathStart and texMathEnd are only redefined *during the output process*. 
To make sure, I just tried it in the web app, and sure enough
```
i1 : tex(ZZ[x,y,z])
o1 =  ${\mathbb Z}\left[x,y,z\right]$
```

M2/Macaulay2/m2/expressions.m2
pzinn: yeah I don't remember why I made that change 2 years ago.
rethinking and possibly reverting

pzinn: first note that the change, as far as I can tell, is purely cosmetic: the end result is the same.
(I will revert it because I don't remember why I did it).
concerning your remark: I guess the problem is with the `#0` or `@@first`. Ideally you're right we would change the way `MatrixExpression` is internally so `MatrixDegreeExpression` can be a descendant of it. At the moment there's no way to squeeze in the information about the degrees in the way `MatrixExpression` is formatted. Since I didn't want to change `MatrixExpression` but still wanted an expression that preserves degrees (since they are printed in `Standard` mode), I created this distinct type.
Again this is all changes made 2 years ago.
That being said, I'm open to suggestions for an alternative format.

M2/Macaulay2/m2/exports.m2
DanGrayson: What if the user is doing things in various threads?  Then they might be doing something else on another thread during the output process.

Of course, much else in the system is not thread-safe, but every little bit of thread-safety can help eventually.

pzinn: OK that's problem. Is there any canonical solution to this kind of thread-safety issues?
Otherwise I can revert to how it was before: I had my own routine `modifiedTex` (or whatever) which did exactly the same as `tex` except it changed the delimiters. I find this solution much less satisfactory.

DanGrayson: I'm just guessing, but perhaps your webapp top level loop is formatting all the output expressions in tex for display.  So you take an output expression `e` and try to format it as math.  You can use `texMath e` to get the TeX code for it, and then use something else to display it.  Why bother with `tex e` and its behavior at all?

pzinn: that's right. that's how it used to be; but I still need to wrap the tex to let the web app know that the output is tex (as opposed to plain html); that's where the `modifiedTex` (actually I think it was called `texWrap` but anyway) came in. I thought I was being clever merging it with the ordinary `tex` method, but this thread-safety issue can't be avoided, then I suppose I'll have to revert.

pzinn: removed the whole tex redefinition. this PR is getting smaller and smaller!

M2/Macaulay2/m2/webapp.m2
```diff
@@ -1,38 +1,47 @@
...
-htmlWithTex Holder := x -> htmlWithTex x#0
+	  then htmlWithTex net k -- sucks but no choice
+	  else htmlWithTex net k | " : " | htmlWithTex v
+	  ) | "<br/>"), "</pre>")
+
+-- now preparation for output
+
+webAppBegin = (displayStyle) -> (
+    texMathStart = webAppTexTag | (if displayStyle then "\\displaystyle " else "");
+    texMathEnd = webAppTexEndTag;
```
DanGrayson, line 39: What's the point of changing the values of these two variables?

pzinn: it's still convenient to have these variables (which are internal to webapp.m2 now) because Print and AfterPrint have slightly different styles of output. and yes, it makes the output routines themselves not thread-safe, but I really don't see the problem -- certainly output routines should not be called in parallel, it would create a mess anyway.

Issue #1201, opened 2020-05-27
regression: missing gfan causes crash in help


mahrud:
Seems like #1123 causes `needsPackage "gfanInterface"` to fail, which leads to this:
```
i1 : help "running Macaulay2 in emacs"
../../Macaulay2/packages/gfanInterface.m2:132:10:(2):[34]: error: could not find gfan
```
Probably because `help` tries to load `Macaulay2Doc`, which in turn tries to read all other packages.

cc: @d-torrance

d-torrance:
Maybe just print a warning message instead?  And leave the error-raising until the user tries to actually call gfan?

mahrud:
That would be better, though, more generally, must the package find gfan globally as soon as it is loaded? Even a warning when calling `help` is strange. An alternative is to look for it the first time gfan needs to be called and cache the location.

d-torrance:
Good point.

It would be nice if a user who explicitly tried to load `gfanInterface` would know right off the bat that they don't have gfan installed before they start defining their rings and ideals and whatnot.  But random warning messages while running `help` aren't good either.

I'll go ahead and prepare a pull request moving the check until the first actual call to gfan.

mahrud:
The issue doesn't appear to be fixed:
```
macaulay@69cf0245150f:~/M2/M2/BUILD/build-docker$ ./M2 -q
Macaulay2, version 1.15.1.0
...

i1 : help sum
in the evaluation of: (gfanHelp "gfan")
currentString:1:2:(2):[51]: error: could not find gfan

```

d-torrance:
It seems like the underlying issue would be solved if/when #1085 is fixed, right?

I think we want `gfanHelp` to return an error if gfan isn't installed, because it's designed exactly for capturing the output of `gfan --help`.

mahrud:
Sure, but that issue won't be fixed soon. I think for now it would be easier to just print warning messages instead of an outright error, as was the case before #1123.

d-torrance:
Sounds good.  I'll work on a fix.

d-torrance:
I'm a little confused, as I can't reproduce the error using `help` or even `needsPackage "gfanInterface"`.  I don't get it until I try `installPackage "gfanInterface"` with gfan itself uninstalled.  This was running pretty recent build, from 5c60aa0.

If I switch the check for a gfan binary to a warning message, then I still get an error message a bit later:
```
i1 : installPackage "gfanInterface"
warning: could not find gfan
[...]

PR #1202, opened 2020-05-27
Don't check for gfan until the user actually needs it

d-torrance:
This prevents errors from occurring when just loading the
gfanInterface package but not calling any of its functions (e.g., when
running "help") when gfan is not installed.

Close: #1201

### DIFF
--- a/M2/Macaulay2/packages/gfanInterface.m2
+++ b/M2/Macaulay2/packages/gfanInterface.m2
@@ -1062,7 +1062,7 @@ runGfanCommandCaptureBoth = (cmd, opts, data) -> (
 	errorMsg := "";
      	if(not returnvalue == 0) then
 	(
-	    errorMsg = "GFAN returned an error message.\n";
+	    errorMsg = "Gfan returned an error message.\n";
 	    errorMsg = errorMsg | "COMMAND:" | ex | "\n";
 	    errorMsg = errorMsg | "INPUT:\n";
 	    errorMsg = errorMsg | get(tmpFile);
--- a/M2/Macaulay2/packages/gfanInterface.m2
+++ b/M2/Macaulay2/packages/gfanInterface.m2
@@ -1045,6 +1045,11 @@ toPolymakeFormat(Fan) := (F) ->(
 --------------------------------------------------------
 
 runGfanCommand = (cmd, opts, data) -> (
+	(out, err, fileName) := runGfanCommandCaptureBoth(cmd, opts, data);
+	(out, fileName)
+)
+
+runGfanCommandCaptureBoth = (cmd, opts, data) -> (
 	if gfanPath === null then gfanPath = findGfanPath();
 	tmpFile := gfanMakeTemporaryFile data;
 	
@@ -1064,45 +1069,19 @@ runGfanCommand = (cmd, opts, data) -> (
 	    errorMsg = errorMsg | "ERROR:\n";
 	    errorMsg = errorMsg | get(tmpFile |".err");
 	     );
-		out := get(tmpFile | ".out");
-	gfanRemoveTemporaryFile tmpFile;
-	gfanRemoveTemporaryFile(tmpFile | ".out");
-	gfanRemoveTemporaryFile(tmpFile | ".err");
-	if length(errorMsg) > 0 then error errorMsg;
-	outputFileName := null;
-	
-	if gfanKeepFiles then outputFileName = tmpFile|".out";
-	
-	(out, "GfanFileName" => outputFileName)
-	
-)
-
-runGfanCommandCaptureBoth = (cmd, opts, data) -> (
-	tmpFile := gfanMakeTemporaryFile data;
-	args := concatenate apply(keys opts, key -> gfanArgumentToString(cmd, key, opts#key));
-	ex := gfanPath | cmd | args | " < " | tmpFile | " > " | tmpFile | ".out" | " 2> " | tmpFile | ".err";
-	if gfanVerbose then << ex << endl;
-	run ex;
 	out := get(tmpFile | ".out");
 	err := get(tmpFile | ".err");
 	gfanRemoveTemporaryFile tmpFile;
 	gfanRemoveTemporaryFile(tmpFile | ".out");
 	gfanRemoveTemporaryFile(tmpFile | ".err");
+	if length(errorMsg) > 0 then error errorMsg;
 	outputFileName := null;
 	if gfanKeepFiles then outputFileName = tmpFile|".out";
 	(out,err, "GfanFileName"=>outputFileName)
 )
 
 runGfanCommandCaptureError = (cmd, opts, data) -> (
-	tmpFile := gfanMakeTemporaryFile data;
-	args := concatenate apply(keys opts, key -> gfanArgumentToString(cmd, key, opts#key));
-	ex := gfanPath | cmd | args | " < " | tmpFile | " > " | tmpFile | ".out" | " 2> " | tmpFile | ".err";
-	if gfanVerbose then << ex << endl;
-	run ex;
-	err := get(tmpFile | ".err");
-	gfanRemoveTemporaryFile tmpFile;
-	gfanRemoveTemporaryFile(tmpFile | ".out");
-	gfanRemoveTemporaryFile(tmpFile | ".err");
+	(out, err, fileName) := runGfanCommandCaptureBoth(cmd, opts, data);
 	err
 )
 
--- a/M2/Macaulay2/packages/gfanInterface.m2
+++ b/M2/Macaulay2/packages/gfanInterface.m2
@@ -129,7 +129,7 @@ findGfanPath = () -> (
 
 fig2devPath = gfanInterface#Options#Configuration#"fig2devpath"
 gfanVerbose = gfanInterface#Options#Configuration#"verbose"
-gfanPath = findGfanPath()
+gfanPath = null
 
 gfanKeepFiles = gfanInterface#Options#Configuration#"keepfiles"
 gfanCachePolyhedralOutput = gfanInterface#Options#Configuration#"cachePolyhedralOutput"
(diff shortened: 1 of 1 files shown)


mahrud:
Now I get the following:
```
in the evaluation of: (gfanHelp "gfan")
/home/macaulay/M2/M2/Macaulay2/m2/robust.m2:98:26:(1):[53]: error: no method for binary operator | applied to objects:
--            null (of class Nothing)
--      |     "gfan" (of class String)
/home/macaulay/M2/M2/Macaulay2/packages/gfanInterface.m2:1099:24:(2):[52]: --back trace--
/home/macaulay/M2/M2/Macaulay2/packages/gfanInterface.m2:2610:22:(2):[51]: --back trace--
```

mahrud:
Changing `null` to `""` solves it though.

d-torrance:
The underlying problem is that there are three functions which call gfan: `runGfanCommand`, `runGfanCommandCaptureError`, and `runGfanCommandCaptureBoth`.

Somehow I never scrolled down to see the other two, and only the first one got updated in the recent pull requests!  I'll work on a fix.

d-torrance:
I've refactored the three functions in the most recent commit so that gfan is only called in one place, which should fix the problem.

mahrud:
@DanGrayson I'm not experienced enough with this package to know whether the changes will affect anything. Shouldn't we ask one of the authors for review? Probably the previous PR should have gone through that as well.

mahrud:
For instance, previously none of the three functions returned an error value, and now all of them do. I think this makes `runGfanCommandCaptureError` not work, since the program errors out rather than printing the gfan error.

DanGrayson:
Good idea.  @mikestillman?  As author, do you approve the change?

Signaling an error if gfan is not found seems to be a good idea.

d-torrance:
We only get an error if gfan returns something nonzero, which shouldn't happen unless something really bad happens, like it's not installed.

`runGfanCommandCaptureError` just returns gfan's output to `stderr`.  It's only used by `gfanHelp`, since gfan prints its help text there:

```
profzoom@gloria:~$ gfan --help > /dev/null
This is a program for computing all reduced Groebner bases of a polynomial ideal. It takes the ring and a generating set for the ideal as input. By default the enumeration is done by an almost memoryless reverse search. If the ideal is symmetric the symmetry option is useful and enumeration will be done up to symmetry using a breadth first search. The program needs a starting Groebner basis to do its computations. If the -g option is not specified it will compute one using Buchberger's algorithm.
Options:
-g:
 Tells the program that the input is already a Groebner basis (with the initial term of each polynomial being the first ones listed). Use this option if it takes too much time to compute the starting (standard degree lexicographic) Groebner basis and the input is already a Groebner basis.

--symmetry:
 Tells the program to read in generators for a group of symmetries (subgroup of $S_n$) after having read in the ideal. The program checks that the ideal stays fixed when permuting the variables with respect to elements in the group. The program uses breadth first search to compute the set of reduced Groebner bases up to symmetry with respect to the specified subgroup.

-e:
 Echo. Output the generators for the symmetry group.

--disableSymmetryTest:
 When using --symmetry this option will disable the check that the group read off from the input actually is a symmetry group with respect to the input ideal.

--parameters value:
 With this option you can specify how many variables to treat as parameters instead of variables. This makes it possible to do computations where the coefficient field is the field of rational functions in the parameters.
--interrupt value:
 Interrupt the enumeration after a specified number of facets have been computed (works for usual symmetric traversals, but may not work in general for non-symmetric traversals or for traversals restricted to fans).
```

d-torrance:
Also a couple other gfan functions print to both `stdout` and `stderr`, which is why `runGfanCommandCaptureBoth` exists, e.g.:

```
profzoom@gloria:~/src/macaulay2/M2$ gfan _tropicallinearspace -d 1 -n 3 2> /dev/null
{1,2,3}
Q[t,x0,x1,x2]{
t^997*x1+t^998*x0,
t^997*x2+t^999*x0,
t^998*x2+t^999*x1}
profzoom@gloria:~/src/macaulay2/M2$ gfan _tropicallinearspace -d 1 -n 3 > /dev/null
{1,2,3}
0,1,0,=1
1,0,0,=2
0,0,1,=0
1,0,0,=2
0,0,1,=0
0,1,0,=1
```

But it still returns 0, so we shouldn't run into an error in Macaulay2:
```
profzoom@gloria:~/src/macaulay2/M2$ echo $?
0
```

mikestillman:
@dmaclagan: Doug Torrance has made a few changes to gfanInterface (he did a pull request to our M2 Macaulay2 repo).  The changes are basically to give an error if something bad happens, e.g. if the gfan command is not found, or if it returns a non-zero error code.  He also merged 3 functions which call gfan into 1 function (they differ depending on how much output they capture from the gfan output).

These look like good changes to me.  If you have no problem with that, we will accept this pull request.

mikestillman:
> For instance, previously none of the three functions returned an error value, and now all of them do. I think this makes `runGfanCommandCaptureError` not work, since the program errors out rather than printing the gfan error.

@mahrud, is your comment above still an issue?  I think it is better for the function to raise an error if something bad happens.  Is there a case when it should be returning a better error message?  gfan uses data to stderr not as an error indicator (maybe for help strings).  Perhaps I don't understand your comment.

mahrud:
I just meant that I'm not really familiar with gfan or gfanInterface, so I can't review the changes properly (e.g. I didn't know the return code is zero even when gfan gives an error).

mikestillman:
@d-torrance, @dmaclagan : Diane suggested we discuss this package next week at the Warwick workshop, as we will all "be" there.  So I will hold up on accepting this until we all discuss the package next week, if that is ok with everyone.

d-torrance:
> @d-torrance, @dmaclagan : Diane suggested we discuss this package next week at the Warwick workshop, as we will all "be" there. So I will hold up on accepting this until we all discuss the package next week, if that is ok with everyone.

Sounds good to me!

d-torrance:
I figured I'd document the recent history of this package while it's still halfway fresh in my mind.  That way it will all be in one place we can refer to for next week's discussion.

* In my tests while packaging Macaulay2 for Debian, I noticed some strange errors when I didn't have Macaulay2 build and install its own copy of gfan:
```
i8 : gfan I
stdio:17:1:(3): error: expected a list, sequence, hash table, or string
```
* I figured out it was because Macaulay2 was trying to run a copy of gfan that wasn't there, but the error wasn't caught right away.  Rather, the error was cropping up a bit later when it was trying to parse the nonexistent output.  This looked like a duplicate of #1066, so I made some comments there.

* I then opened #1123.  In the first commit (cc7a41c), I changed `runGfanCommand` to raise an error if gfan's return value was nonzero.  Someone had already planned something like this at some point anyway, as my commit was mostly just uncommenting some existing code.

* Dan suggested that we go a step further and have Macaulay2 look several places for gfan.  I implemented this in 43f9681.  In particular, when the package is first loaded, we first check if the user defined a path to gfan.  If they didn't (or we can't find gfan there), then check for the M2-installed copy.  If that's not there, then we check all the directories in `PATH`.  If nothing is found, then we return an error.  Note that this is the current behavior, as these two commits were merged into `development`, which was later merged into `master`.

* In his tests, Mahrud found #1201.  In particular, sometimes (e.g., when running `help`) we load `gfanInterface` when we really don't care to call gfan itself.  If gfan isn't installed, then we'd get an error since the check is done as soon as the package is loaded.

* I attempted to fix this in the first commit of the current pull request (c04801c).  This moves the check for gfan to the first time the user called gfan using `runGfanCommand`.

* Mahrud found a bug in the first commit, as sometimes (e.g., `gfanHelp`, which is used to generate the documentation for the package) , we don't call `runGfanCommand`, but one of two other functions: `runGfanCommandCaptureError` or `runGfanCommandCaptureBoth`.  The code for the three functions was nearly identical, but the return values differed (basically, do we want `stdout`, `stderr`, or both?).  I completely missed these other two functions!

* To fix this bug, in c4a3a8a I refactored the three functions so that the following things occur in one place:
  - If we haven't found gfan yet, then find it.
  - Run gfan
  - Throw an error if gfan's return value is nonzero
  - Record `stdout` and `stderr`

  Then the different functions deal separately with what they want to return.

mahrud:
I don't know if this would ever happen with the gfanInterface design, but errors in gfan sometimes throw nonzero errors:
```
[mahrud@noether ~]$ gfan _bases
N[x]   
PARSER ERROR: Expected: "field"    Found: "N"
gfan: src/parser.cpp:592: Field CharacterBasedParser::parseField(): Assertion `0' failed.
Aborted (core dumped)
[mahrud@noether ~]$ echo $?
134
```
And in the old design this could be captured from M2 interface:
```
i8 : runGfanCommand("gfan _bases", opts, "N[x]")
stdio:8:1:(3): error: GFAN returned an error message.
COMMAND:/usr/local/libexec/Macaulay2/bin/gfan _bases < /tmp/M2-4165622-0/43 > /tmp/M2-4165622-0/43.out 2> /tmp/M2-4165622-0/43.err
INPUT:
N[x]ERROR:
PARSER ERROR: Expected: "field"    Found: "N"
```

d-torrance:
That's using the code from master with the commits from #1123, right?  If I check out the `release-1.15` branch, then the same thing runs without error since the error stuff is still commented out:

```
i4 : runGfanCommand("gfan _bases", hashTable {}, "N[x]")

o4 = (, GfanFileName => /tmp/M2-12830-0/0.out)

o4 : Sequence
```

Then Maculay2 would try to do things with that empty string that's returned and run into the strange errors that prompted all of this. Currently, this is still a problem with the other two functions.

mikestillman:
Perhaps we can fix this stuff next week?

d-torrance:
@mikestillman Just pushed the commit we discussed changing the capitalization for the error message (GFAN -> Gfan).  It should be ready now!

mikestillman:
@dan, I'm going to accept this pull request (into branch development).  OK?  Or, you can do so instead of answering this!

DanGrayson:
> @DanGrayson , I'm going to accept this pull request (into branch development). OK? Or, you can do so instead of answering this!

@mikestillman - Let me take care of that.  I want to see the #1190 pass the overnight test first (on the development branch.)

mikestillman:
@DanGrayson OK, I'll let you handle that.

PR #1204, opened 2020-05-28
Require googletest >= 1.10.0.

d-torrance:
Now that mathicgb's unit tests use the *_TEST_SUITE naming convention
introduced in 1.10.0, we build googletest if the user has a lower version.

### DIFF
--- a/M2/config/gtest.m4
+++ b/M2/config/gtest.m4
@@ -72,7 +72,7 @@ AC_DEFUN([CHECK_GTEST],
   # using TYPED_TEST_SUITE, one of the functions which was introduced in
   # 1.10.0 in the switch from the *_TEST_CASE to the *_TEST_SUITE naming
   # convention.  If this test fails, then an older googletest is
-  # installed, so we instead build our own googletest.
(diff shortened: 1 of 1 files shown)


mahrud:
2 minor things that might as well be fixed:
- we don't build googletest (in the sense of building libgtest), so we can correct the language
- we don't link with libgtest either, so checking for existence of `gtest.h` isn't enough. Mathicgb includes `gtest-all.cc` which is not always installed even when `gtest.h` is there (e.g. with Fedora). The solution would be to build a small program that includes `gtest-all.cc`.

d-torrance:
There's already a test for `gtest-all.cc` which I didn't need to do anything with in this pull request:

https://github.com/Macaulay2/M2/blob/80c4d54d3ead765560c660519ebfaf5a4bc1f830/M2/config/gtest.m4#L58-L66

The `AC_COMPILE_IFELSE` statement is just making sure that we have 1.10.0, since `TYPED_TEST_SUITE` didn't exist before that.

I'll push another commit changing the language.

PR #1205, opened 2020-05-28
Update VirtualResolutions package to v1.2

mahrud:


M2/Macaulay2/packages/VirtualResolutions.m2
```diff
@@ -556,6 +561,6 @@ uninstallPackage "VirtualResolutions"
 restart
 installPackage "VirtualResolutions"
 restart
-needsPackage("VirtualResolutions", FileName => "./VirtualResolutions.m2")
+needsPackage("VirtualResolutions", FileName => "VirtualResolutions.m2")
```
DanGrayson, line 564: Why is the FileName option still needed?

M2/Macaulay2/packages/VirtualResolutions/tests.m2
```diff
@@ -200,6 +211,7 @@ TEST ///
     assert(multigradedRegularity(S, I) == {{1,5},{2,2},{4,1}})
 ///
 
+-- This test currently fails due to a bug in tateResolutions
 ///
```
mahrud, line 215: Ahh, this is fixed in TateOnProducts now. @DanGrayson could you remove the comment and enable that test in a commit to development? I'd rather not make a new PR for one line.

M2/Macaulay2/packages/VirtualResolutions.m2
mahrud: I'm not sure who needed that, but it's in the footer, so does it matter?

M2/Macaulay2/packages/VirtualResolutions/tests.m2
DanGrayson: okay  :  commit 6cbb03398cd607181c78ab1c69663a7e94abac64

PR #1228, opened 2020-06-03
Don't relativize "stdio"

d-torrance:
Current behavior doesn't make sense:

    i1 : relativizeFilename "stdio"

    o1 = ../../../../../../../../../stdio

    i2 : relativizeFilename("/home/m2user", "stdio")

    o2 = ../../../stdio

### DIFF
--- a/M2/Macaulay2/d/stdiop.d
+++ b/M2/Macaulay2/d/stdiop.d
@@ -64,6 +64,7 @@ absoluteFilename(filename:string):string := (
      shorten(filename)
      );
 relativize(cwd:string,filename:string):string := (	    -- purely textual
+     if filename === "stdio" then return filename;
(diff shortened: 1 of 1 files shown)


DanGrayson:
Good idea.  What to do if there really is a file with filename "stdio", though?

d-torrance:
> Good idea. What to do if there really is a file with filename "stdio", though?

We're already making the assumption that there isn't:

https://github.com/Macaulay2/M2/blob/99b7b7e918e9c79dae00b6827284cb77fc7e5bbd/M2/Macaulay2/d/stdiop.d#L78

And `absoluteFilename` eventually does the following check:

https://github.com/Macaulay2/M2/blob/99b7b7e918e9c79dae00b6827284cb77fc7e5bbd/M2/Macaulay2/d/stdiop.d#L46-L51

Issue #1085, opened 2020-04-01
Feature request: allow only reading a package header


mahrud:
This section of Macaulay2Doc, which is responsible for making a list of packages, calls `needsPackage` on every package, but this isn't necessary since all we need is the information in the package header.
https://github.com/Macaulay2/M2/blob/b205e6fc41833e209601704952563cee15cd887d/M2/Macaulay2/packages/Macaulay2Doc/overview_packages.m2#L7-L32

Beside the obvious issues that this is unnecessary, it slows down installation of packages in parallel. For instance, before Macaulay2Doc is installed, if you install any other package (even FirstPackage), somehow* M2 tries to generate that particular documentation node of Macaulay2Doc, and unintentionally runs path checks from SemidefiniteProgramming, which produce a lot of errors if the software aren't installed:
https://github.com/Macaulay2/M2/blob/b205e6fc41833e209601704952563cee15cd887d/M2/Macaulay2/packages/SemidefiniteProgramming.m2#L91-L96

I have been thinking about refactoring packages.m2 (and html.m2) for this and several other reasons, but wanted to make this issue for now in case anyone has other ideas.

*: not sure why or where this happens yet.

mahrud:
Example of the errors produced from SemidefiniteProgramming:
```
--Installing package PackageCitations
--making example results for PackageCitations
--making example results for cite
which: no  in (...)
which: no mosek in (...)
which: no sdpa in (...)
Solvers configured: CSDP
Default solver: CSDP
-- polymake not present
```

DanGrayson:
There's a related issue: https://github.com/Macaulay2/M2/issues/776

How would you get the header without loading the package?

mahrud:
I presume we can stop reading the file when "newPackage(...)" closes. Like a break in the evaluation loop.

DanGrayson:
Good idea.  Set a static flag, and then at the end of "newPackage", conditionally return the symbol "end", which will end the parsing of the file.  Oops, there is a problem: the end of file causes "endPackage" to be run, and the package will be entered into the list of packages as a complete package.  So you have to override that, too, and make sure the next "needsPackage" rereads the file instead of just opening the package's empty dictionary.

mahrud:
Something like that, yes.

Before getting to that, I wanted to see what's really slowing down `installPackage` before Macaulay2Doc is installed and found where `installPackage` calls `help` takes 18 seconds every time:
[...]

PR #1231, opened 2020-06-04
Install FirstPackage before Macaulay2Doc.

d-torrance:
Since we load FirstPackage as an example in the documentation but the
database hasn't been built yet, we currently get a warning message in the
example for "loadPackage":

     |i2 : loadPackage "FirstPackage"                                                                                                                                                               |
     |--database not present: /home/dan/src/M2/M2.git/M2/BUILD/dan/builds.tmp/ubuntu1910-release-1.15/usr-dist/x86_64-Linux-Ubuntu-19.10/lib/Macaulay2/FirstPackage/cache/rawdocumentation-dcba-8.db|

After switching the installation order, the example's behavior is the
normal one when loading a package with notify set to true:

     |i2 : loadPackage "FirstPackage"                                                                                                                                                    |
     |--opening database /build/macaulay2-1.15.1.0+git390.695bfc7+ds/M2/usr-dist/x86_64-Linux-Debian-unknown/lib/x86_64-linux-gnu/Macaulay2/FirstPackage/cache/rawdocumentation-dcba-8.db|

### DIFF
--- a/M2/Macaulay2/packages/=distributed-packages
+++ b/M2/Macaulay2/packages/=distributed-packages
@@ -1,6 +1,6 @@
 Style
-Macaulay2Doc
 FirstPackage
+Macaulay2Doc
 Parsing
 Classic
 Browse
--- a/M2/Macaulay2/packages/Makefile.in
+++ b/M2/Macaulay2/packages/Makefile.in
@@ -91,9 +91,11 @@ $(foreach i, $(PACKAGES), $(eval all: unmark-$i))
 endif
 $(foreach i, $(PACKAGES), $(eval all: all-$i) $(eval install: install-$i))
 
-# make packages Style and Macaulay2Doc get installed before the others:
+# make packages Style, FirstPackage, and Macaulay2Doc get installed
+# before the others:
 $(foreach i, $(patsubst Style,,$(ALL_PACKAGES)), $(eval @pre_libm2dir@/$i/.installed: | @pre_libm2dir@/Style/.installed))
-$(foreach i, $(patsubst Macaulay2Doc,,$(patsubst Style,,$(ALL_PACKAGES))), $(eval @pre_libm2dir@/$i/.installed: | @pre_libm2dir@/Macaulay2Doc/.installed))
+$(foreach i, $(patsubst FirstPackage,,$(patsubst Style,,$(ALL_PACKAGES))), $(eval @pre_libm2dir@/$i/.installed: | @pre_libm2dir@/FirstPackage/.installed))
+$(foreach i, $(patsubst Macaulay2Doc,,$(patsubst FirstPackage,,$(patsubst Style,,$(ALL_PACKAGES)))), $(eval @pre_libm2dir@/$i/.installed: | @pre_libm2dir@/Macaulay2Doc/.installed))
 
 # the Info-validate function doesn't work well enough to be useful:
 # check::check-info


mahrud:
An issue with this is that it would slow down installation, see https://github.com/Macaulay2/M2/issues/1085#issuecomment-610650107. Is there an alternative? e.g. `loadPackage` simply not printing that warning.

On a related note, is `SecondPackage` also needed? If both are needed, perhaps we can remove "docTemplate", "docExample", and "simpleDocFrob" and use those. Reducing overlaps makes things less confusing.

d-torrance:
> An issue with this is that it would slow down installation, see [#1085 (comment)](https://github.com/Macaulay2/M2/issues/1085#issuecomment-610650107). Is there an alternative? e.g. `loadPackage` simply not printing that warning.

The same example is also used for [notify](https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.15/share/doc/Macaulay2/Macaulay2Doc/html/_notify.html), so in its current state, we certainly would like to see some notifications!

I suppose an alternate fix would be to keep `notify` as `false` for the `loadPackage` example and just pick a completely different example for `notify`.

DanGrayson:
Why is the path in `opening database /build/macaulay2-1.15.1.0+git390.695bfc7+ds/M2/usr-dist/x86_64-Linux-Debian-unknown/lib/x86_64-linux-gnu/Macaulay2/FirstPackage/cache/rawdocumentation-dcba-8.db` incorrect?  (There's no directory `/build`.)

DanGrayson:
> An issue with this is that it would slow down installation, see [#1085 (comment)](https://github.com/Macaulay2/M2/issues/1085#issuecomment-610650107). Is there an alternative? e.g. `loadPackage` simply not printing that warning.

Yes, that's a bit noisy.  Perhaps print it only if `notify` is on.  What problem are you trying to solve?  Readers of the documentation seeing  a confusing warning message?  I suppose that's justifiable.

> 
> On a related note, is `SecondPackage` also needed? If both are needed, perhaps we can remove "docTemplate", "docExample", and "simpleDocFrob" and use those. Reducing overlaps makes things less confusing.

The headline for `SecondPackage` says "an example of a package that depends on another" -- that sounds like a justification.  The three things you propose removing are related to SimpleDoc, hence unrelated to the two packages.

d-torrance:
> Why is the path in `opening database /build/macaulay2-1.15.1.0+git390.695bfc7+ds/M2/usr-dist/x86_64-Linux-Debian-unknown/lib/x86_64-linux-gnu/Macaulay2/FirstPackage/cache/rawdocumentation-dcba-8.db` incorrect? (There's no directory `/build`.)

`/build` is the default build directory used by [pbuilder](https://manpages.debian.org/unstable/pbuilder/pbuilder.8.en.html), which is what I use to test builds of the Debian package.  It creates this directory inside a chroot.

d-torrance:
> Yes, that's a bit noisy. Perhaps print it only if `notify` is on. What problem are you trying to solve? Readers of the documentation seeing a confusing warning message? I suppose that's justifiable.

That's the current behavior -- it only prints these messages when `notify` is on.  But this is coming from the documentation for [notify](https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.15/share/doc/Macaulay2/Macaulay2Doc/html/_notify.html).  :)

I thought it was odd that the documentation had a "database not present" warning in an example since the databases should already be generated by the time the user loads a package.

mahrud:
You could also turn that Example into a CannedExample instead and remove
the warning.

DanGrayson:
The fewer canned examples we have, the better.

Issue #1220, opened 2020-06-01
improve package and simpledoc documentation


mikestillman:
The documentation for packages, and the SimpleDoc format could be improved.  I will edit this list with suggestions, things done, etc.  Help would be welcome!

Here are some of the doc nodes for packages and documentation, and sometimes a few suggestions for things that should be changed.  Perhaps these nodes should even be consolidated somehow.

* The node and package `SimpleDoc`
  - Have a getting started node, where one sees the basic use of SimpleDoc.
  - Have a node (currently the `doc` node) that describes the general format, and also how to do some common things (e.g. having a list of items, adding a URL link).
  - How to write math in a doc node
* The node `Package`
  - make sure `SimpleDoc` is linked to!
  - Similarly, `packageTemplate` should be visible from here
  - `packageTemplate` should include the `end--` near the end of the file.
* The node `packages`
* The node `creating a package`
* The node `an example of a package`
  - should link to `packageTemplate` (in `SimpleDoc`)
  - Perhaps use `doc` rather than `multidoc` for doc here
* Node `informing others about your package`
  - Seems out of date.
* The node `writing documentation`
  -- Doesn't even mention `SimpleDoc`, or `docTemplate`.  It should be based on `SimpleDoc` I think.
* The node `conventions for documentation`
  -- Also out of date.

mahrud:
Do all of the above need to be documentation nodes written within Macaulay2? I'm wondering if using a more expressive format would make more sense.

mikestillman:
@mahrud : What do you have in mind?

mahrud:
Nothing specific, other than maybe unifying all the nodes you mentioned into one that is linked on the front page. I was just looking around, and it seems common to have separate formats for _manuals_ (e.g. setup M2 with emacs, start a package, tutorials, etc.), language _reference_, and package _documentation_. It's not the same everywhere, of course. Here are a few examples from [Go](https://www.golang-book.com/books/intro/11), [Julia](https://julialang.github.io/Pkg.jl/v1/creating-packages/), [Rust](https://doc.rust-lang.org/cargo/guide/creating-a-new-project.html), and [GAP](https://www.gap-system.org/Manuals/pkg/example/doc/chapA.html).

DanGrayson:
`TO document` should work, but `(document,List)` is undocumented.

mahrud:
What about `writing documentation`?

DanGrayson:
For that, use `TO "writing documentation"`

[...]

PR #1234, opened 2020-06-04
SimpleDoc v1.2

### LINKED ISSUE #628: contributing authors
We need a standard way to annotate those who are "contributing authors" for a package, since some are resorting to this sort of thing:
```
   Name => "Contributing Author: Anton Leykin",
```

### LINKED ISSUE #926: SimpleDoc should support SYNOPSIS
There is no way to translate this documentation node into SimpleDoc:

```
newPackage ("Foo", Headline => "foo", DebuggingMode => true)
LieAlgebra = new Type of HashTable
basis(ZZ,LieAlgebra) := List => (n,L) -> opts -> {}
basis(ZZ,ZZ,LieAlgebra) := List => (m,n,L) -> opts -> {}
basis(List,LieAlgebra) := List => (s,L) -> opts -> {}
export "LieAlgebra"
beginDocumentation()
document {
     Key => {
	     (basis,ZZ,LieAlgebra),
	     (basis,ZZ,ZZ,LieAlgebra),
	     (basis,List,LieAlgebra)
	  },
     Headline => "make a basis",
     SYNOPSIS {
	  Usage => "basis(m,L)",
     	  Inputs => {
	       "m" => ZZ,
	       "L" => LieAlgebra
	       },
	  Outputs => {
	       List => "a basis of the degree m part of L"
	       }
	  },
     SYNOPSIS {
	  Usage => "basis(m,n,L)",
     	  Inputs => {
	       "m" => ZZ,
	       "n" => ZZ,
	       "L" => LieAlgebra
	       },
	  Outputs => {
	       List => "a basis of the part of L of degree between m and n"
	       }
	  },
     SYNOPSIS {
	  Usage => "basis(s,L)",
     	  Inputs => {
	       "s" => List,
	       "L" => LieAlgebra
	       },
	  Outputs => {
	       List => "a basis of the part of L of degree occurring in the list s"
	       }
	  }
     }
```

... and there should be.

### LINKED ISSUE #975: "help" multiple "Usage" (and viewHelp)
This concerns documentation nodes that list several ways to call the function (`Usage` in SimpleDoc).
```
i1 : installPackage "MonodromySolver"

i2 : help createSeedPair

o2 = createSeedPair -- create initial seed for the homotopy continuation
     *******************************************************************

     Synopsis
     ========

       * Usage: (p0,x0) = createSeedPair PS(p0, x0) = createSeedPair(PS,x0)
...
```
Note: no linebreak above. Also, note that `viewHelp` does put a line break.
 
Another observation:
```
i3 : help (createSeedPair,System)
```
works but 
```
i4 : viewHelp (createSeedPair,System)
```
does **not** since it takes only `String`... but it would be nice if it did.

### LINKED ISSUE #1000: classification of functions in a package
Frank Schreyer and David Eisenbud suggest that a package's documentation should be able to put the documentation nodes for functions into various subheadings, as manually illustrated here: https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.14/share/doc/Macaulay2/K3Carpets/html/.

mahrud:
The improvements are imperceptible to existing packages, but here's a summary:
- [x] the package now is more readable, has better organization, and more comments.
- [x] I implemented some of @mikestillman's documentation suggestions in #1220
- [x] made the documentation for `(doc, String)` much better, with links to the keyword types.
- [x] I also improved the package template and docExample and added testExample.
- [x] Lastly, I added three new keywords:
  - [x] Acknowledgements
  - [x] Contributors (close #628)
  - [x] References
- [x] change `installPackage.m2` to add headings for the keywords above.
- [x] moved `Macaulay2Doc$Wikipedia` to `SimpleDoc$wikipedia`
- [x] added `stacks` and `arXiv` methods.
- [x] close #1000 via the `Tree` keyword
- [x] close #926 and close #975 via the `Synopsis` keyword


mahrud:
I should mention, [this line](https://github.com/mahrud/M2/blob/ca301330d6f23abe36c8b8adceb7c6ddf9379e34/M2/Macaulay2/packages/SimpleDoc.m2#L20):
```
20:    "Node", "Item", "CannedExample", "Pre", "Code", "Acknowledgement", "Contributors", "References" -- temporary nodes
```
Might move to `exports.m2`, depending on where we choose to document these keywords, or `undoc.m2` if we wish not to.

Also, one question is whether, for instance, `Key` on the page for `doc` should have a paragraph of description (as this PR does), or whether the description should move to the documentation node for `Key` itself and on the `doc` page it should only have a link to that node.

M2/Macaulay2/packages/SimpleDoc/doc.txt
```diff
@@ -1,4 +1,21 @@
...
+Node
+  Key
+    SimpleDoc
+  Headline
+    simpler documentation for functions and methods
+  Description
+    Text
+      This package provides a human-readable serialization format for
+      documentation nodes as an alternative to the classic way of
+      @TO "writing documentation"@ using the @TO document@ method.
```
DanGrayson, line 11: Make this link work with : 
```
--- a/M2/Macaulay2/packages/SimpleDoc/doc.txt
+++ b/M2/Macaulay2/packages/SimpleDoc/doc.txt
@@ -8,14 +8,14 @@ Node
     Text
       This package provides a human-readable serialization format for
       documentation nodes as an alternative to the classic way of
-      @TO "writing documentation"@ using the @TO document@ method.
+      @TO "Macaulay2Doc::writing documentation"@ using the @TO document@ method.
       The new format is parsed by either @TO doc@ for writing a single node,
       or @TO multidoc@ for writing multiple documentation nodes in one file
       and adheres off-side rule; i.e., using indentation to separate blocks.
   Contributors
     Andrew Hoefel contributed to this package.
   SeeAlso
-    "writing documentation"
+    "Macaulay2Doc::writing documentation"
 Node
  Key
   multidoc
```

M2/Macaulay2/packages/SimpleDoc/doc.txt
```diff
@@ -12,7 +29,7 @@ Node
  Consequences
   Item
    The string {\tt s} is handled by parsing it the same way @ TO doc @ does,
-   but expecting the least indented keyword to be {\bf Node}, which can occur
+   but expecting the least indented keyword to be @ TO Node @, which can occur
```
DanGrayson, line 32: There's no reason for things such as "Node" to be exported symbols, since they occur only inside strings and not in Macaulay2 code.

M2/Macaulay2/packages/SimpleDoc/doc.txt
mahrud: I thought I tried this, I wonder why it didn't work.

M2/Macaulay2/packages/SimpleDoc/doc.txt
mahrud: It is a keyword in the documentation system, so I think it should be highlighted so users know it's correct. Is there another, easy way to highlight it?

How about we change `Node` to `doc`? Then the format would be:
```
doc
  Key
    blah
  Headline
    blah blah
doc
  Key
    Foo
  Headline
    Foo Foo
```
It is already highlighted, and semantically it makes sense that the input to `multidoc` is made up of a bunch of inputs to `doc`. Also, if you have a documentation file with a bunch of calls to `doc`, you can change it to a text file for `multidoc` just by removing `///` everywhere.

DanGrayson: > It is a keyword in the documentation system, so I think it should be highlighted so users know it's correct. Is there another easy way to highlight it?

Yes, there is: just add them to the list in `emacs/makesyms.m2`.

mahrud: What do you think about changing `Node` to `doc`? There are only about 20 packages that use it, so changing it wouldn't be difficult.

DanGrayson: > What do you think about changing `Node` to `doc`? There are only about 20 packages that use it, so changing it wouldn't be difficult.

I see the following disadvantages:

- "doc" is not a word
- "doc" is not capitalized
- "doc" is already in use for something else in the package
- the word "Node" is descriptive: it introduces a documentation node
- the change would not be backward compatible.

Here's a change I would like to see instead, even though it's backward compatible, if @mikestillman approves:

- get rid of "doc" and "multidoc"
- attach the functionality of "multidoc" as the method for "document String", which is currently empty

It would also be cool to develop an emacs mode for SimpleDoc files.  It could take care of the indentation and helping us know which keywords are valid at the current level of indentation.

eisenbud:
It would be wonderful to have an emacs mode for simpleDoc!!

David
--
David Eisenbud, Director,
Mathematical Sciences Research Institute
and Professor of Mathematics,
University of California, Berkeley

M2/Macaulay2/packages/SimpleDoc/doc.txt
mahrud: I like that. What about also moving the existing `document` method into SimpleDoc?

mahrud:
> It would be wonderful to have an emacs mode for simpleDoc!!

I don't know how to create emacs modes, but I have M2 code for parsing Markdown and I could add functionality to SimpleDoc to read a Markdown file and generate a documentation node from it. The benefit is that Markdown is understood by basically all editors, and even Github can display it natively. Here is an example:
- what github displays: https://github.com/mahrud/m2doc/blob/master/_packages/FGLM.md
- the generated documentation page: https://mahrud.github.io/m2doc/packages/FGLM/
- for comparison in the current system: https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.15/share/doc/Macaulay2/FGLM/html/_fglm.html

M2/Macaulay2/packages/SimpleDoc/doc.txt
DanGrayson: `document` is part of the core, so should stay there.

M2/Macaulay2/m2/document.m2
```diff
@@ -1460,7 +1460,15 @@ tutorial = x -> (
...
 Wikipedia = method(TypicalValue => Hypertext)
-Wikipedia String := s -> PARA { "See ", HREF{ "http://en.wikipedia.org/wiki/" | s }, "."}
+Wikipedia String          :=       title  -> HREF{ "https://en.wikipedia.org/wiki/" | title, title }
+Wikipedia(String, String) := (url, title) -> HREF{ "https://en.wikipedia.org/wiki/" |   url, title }
+
+arXiv = method(TypicalValue => Hypertext)
+arXiv String          :=  ref         -> HREF{ "https://arxiv.org/abs/" | ref, "arXiv:" | ref }
+arXiv(String, String) := (ref, title) -> HREF{ "https://arxiv.org/abs/" | ref, title }
+
+stacks = method(TypicalValue => Hypertext)
```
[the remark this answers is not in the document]
mahrud: Wouldn't any confusion be easily resolved?

Do you have any suggestions?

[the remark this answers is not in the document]
DanGrayson: What about `stacksProject`?

mahrud:
I rebased this branch on top of my refactor/regex branch and got an error, apparently caused by this:
https://github.com/Macaulay2/M2/blob/3a00dcf07869284f10cdc196bce1605c1e47b446/M2/Macaulay2/packages/Polyhedra/documentation/documentation.m2#L1-L33
How did this work before?

mahrud:
I rebased this branch on top of my refactor/regex branch and got an error, apparently caused by this:
https://github.com/Macaulay2/M2/blob/3a00dcf07869284f10cdc196bce1605c1e47b446/M2/Macaulay2/packages/Polyhedra/documentation/documentation.m2#L1-L33
How did this work before

mahrud:
@DanGrayson I added the changes you requested. Couple of outstanding issues:
- [ ] I've defined a new class `Node` in order to facilitate `doc(String)` accepting either a single node or multiple nodes. However, currently the tests fail because `Node` is used in two other places: an example for `tangentSheaf`, and internally in the package `EquivariantGB`. Would you be okay with changing it in those places, so that we can make `Node` a keyword that is both highlighted and meaningful as a type? I think having different names for the type and for the keyword that creates it is kinda ugly.
- [ ] some keywords need to be added to `document.m2` and documented

mahrud:
@DanGrayson is there a reason why `Example` is protected and exported?
https://github.com/Macaulay2/M2/blob/3a00dcf07869284f10cdc196bce1605c1e47b446/M2/Macaulay2/m2/last.m2#L7
Doesn't seem to be defined anywhere.

mahrud:
I think this PR is done.

DanGrayson:
One of the failures involves a regular expression: 
```
make[4]: *** test failed: [Macaulay2/tests/ComputationsBook/geometry]
```

mahrud:
@DanGrayson all green!

.github/workflows/test_build.yml
```diff
@@ -95,7 +95,7 @@ jobs:
           sudo apt-get install -y -q --no-install-recommends libatomic-ops-dev libboost-stacktrace-dev \
                   libncurses-dev libncurses5-dev libreadline-dev libeigen3-dev liblapack-dev libxml2-dev \
                   libgc-dev libgdbm-dev libglpk-dev libgmp3-dev libgtest-dev libmpfr-dev libntl-dev gfan \
-                  libcdd-dev libgivaro-dev fflas-ffpack w3c-markup-validator
+                  libcdd-dev libgivaro-dev libboost-regex-dev fflas-ffpack w3c-markup-validator
```
DanGrayson, line 98: The need for a new library, such as this one, should be documented in the file "INSTALL" for each OS, with the appropriate name there, and the configure script should verify its presence.

(Why didn't I need to install such a thing on Fedora?  Maybe it's just part of libboost.)

M2/Macaulay2/d/boost-regex.cpp
```diff
@@ -0,0 +1,290 @@
+#include "strings-exports.h"
+
+#include <boost/regex.hpp>
+#include <string.h>
+#include <iostream>
+
+#define DEBUG_REGEX 0
+
+/* Keep this enum in sync with RegexFlags in m2/regex.m2 */
+enum RegexFlags {
```
DanGrayson, line 10: This enum is not needed: use boost's `enum flag_type_` instead.

M2/Macaulay2/d/boost-regex.cpp
```diff
@@ -0,0 +1,290 @@
...
+  regex_flags |= flags & literal ? boost::regex::literal : 0;
+  regex_flags |= flags & icase ? boost::regex::icase : 0;
+  regex_flags |= flags & nosubs ? boost::regex::nosubs : 0;
+  regex_flags |= flags & collate ? boost::regex::collate : 0;
+  regex_flags |= flags & no_mod_m ? boost::regex::no_mod_m : 0;
+  regex_flags |= flags & no_mod_s ? boost::regex::no_mod_s : 0;
+  /* the following are on by default on POSIX ERE, so we disable them */
+  regex_flags &= flags & no_escape_in_lists ? regex_flags
+                                            : ~boost::regex::no_escape_in_lists;
+  regex_flags &= flags & no_bk_refs ? regex_flags : ~boost::regex::no_bk_refs;
```
DanGrayson, line 99: It's probably a mistake to complement these, as "no" is part of both names.

M2/Macaulay2/d/boost-regex.cpp
```diff
@@ -0,0 +1,290 @@
...
+  if (hit != cache.end()) return hit->second;
+#if DEBUG_REGEX & 1
+  std::cerr << "flags:\t" << flags << std::endl
+            << "regexp:\t" << M2_tocharstar(pattern) << std::endl;
+#endif
+
+  /* parse and set the flags */
+  boost::regex_constants::syntax_option_type regex_flags =
+      boost::regex::no_except; /* never throw exceptions */
+  regex_flags |= flags & extended ? boost::regex::extended : 0;
```
DanGrayson, line 89: Using boost's enum would eliminate this translation.

M2/Macaulay2/d/boost-regex.cpp
```diff
@@ -0,0 +1,290 @@
...
+  if (expression.status() != 0) return m;
+
+  bool status = false;
+  boost::cmatch matches {};
+
+  auto head = (const char*)&text->array + start;
+  auto tail = (const char*)&text->array + text->len;
+
+  boost::regex_constants::match_flag_type flag = boost::match_default;
+  flag |= flags & match_any ? boost::match_any : flag;
```
DanGrayson, line 147: yes, `flag|flag` yields `flag`, but `flag|0` would be clearer.

M2/Macaulay2/m2/basictests/A06.m2
```diff
@@ -1,13 +1,6 @@
...
-
 assert({1,2,3,4,5} === flatten{1,{},{2},{3,4},5})
 f = { new MutableList from {0,1} }
 assert(f === flatten f)
-
--- test regexp
-assert( regex ( "^(.*[^ ])? *$", " abcdef " ) === {(0, 8), (0, 7)} )
-assert( regex ( "^ *(.*[^ ])? *$", " abcdef " ) === {(0, 8), (1, 6)} )
-assert( regex ( "^ *(.*)$", " abcdef " ) === {(0, 8), (1, 7)} )
-assert( regex ( ".?","" ) === {(0, 0)} )
```
DanGrayson, line 13: Two of these assertions still work, so why delete them.  The two failing ones point out a bug:
```m2
i34 : regex ( "^ *(.*)$", " abcdef " )

o34 = {(0, 8), (0, 8)}

o34 : List

i35 : regex ( "^( *)(.*)$", " abcdef " )

o35 = {(0, 8), (0, 1), (1, 7)}

o35 : List
```
The bug is that addition of parentheses changes the way things match -- in the first one, `.*` matches 8 characters, and in the second one, `.*` matches 7 characters.

It probably has to do with incorrect setting of flags, for the bug goes away if I specify none:
```m2
i36 : regex ( "^ *(.*)$", " abcdef " )

o36 = {(0, 8), (0, 8)}

o36 : List

i37 : regex ( "^ *(.*)$", " abcdef ", Flags => {} )

o37 = {(0, 8), (1, 7)}

o37 : List

i36 : regex ( "^ *(.*)$", " abcdef " )

o36 = {(0, 8), (0, 8)}

o36 : List

i37 : regex ( "^ *(.*)$", " abcdef ", Flags => {} )

o37 = {(0, 8), (1, 7)}

o37 : List
```

All 4 assertions should be restored.

M2/Macaulay2/m2/basictests/regex.m2
```diff
@@ -1,19 +0,0 @@
...
-assert match("a{2}","aa")
-assert match("a[2]","a2")
-assert match("a[^a]","a2")
-assert match("a\\b","a b")
-assert not match("a\\b","ab")
-assert match("a\\>","a b")
-assert match("a\\>","a")
-assert not match("a\\>"," ab")
-
-assert( replaceStrings("^a","x","a \na \naaa a") === "x \nx \nxaa a" )
```
DanGrayson, line 19: Why was this file deleted?  Or are the tests now somewhere else?

M2/Macaulay2/m2/exports.m2
```diff
@@ -328,6 +330,9 @@ export {
 	"Range",
 	"RealField",
 	"Reduce",
+	"References",
+	"RegexPOSIX",
+	"RegexPerl",
```
DanGrayson, line 335: These two won't be needed.

.github/workflows/test_build.yml
mahrud: There's no need, it always comes with boost.

mahrud: Also, the configure script does check for it thanks to @d-torrance.

M2/Macaulay2/d/boost-regex.cpp
mahrud: We use a bitmask to communicate the flags from top level to the engine, so more than the enum, we need to know the actual numbers so they can be written in regex.m2. However, the specific numbers in boost are internal numbers and are not guaranteed to stay the same from version to version. That's why we can't use boost's own enum. Moreover, boost actually has two separate enums, one for regex flags and one for match flags. If we were to use it, we'd need two flags, which is more than we need.

M2/Macaulay2/d/boost-regex.cpp
mahrud: It's not. As the comment line says, these two are enabled by **default** in `extended`, so we need to complement them, since they would introduce breaking changes from GNU regex.

M2/Macaulay2/d/boost-regex.cpp
mahrud: See the comment above on this. The translation is needed since top level doesn't have access to boost's enum.

M2/Macaulay2/d/boost-regex.cpp
mahrud: This wouldn't always compile, since `0` doesn't have the type `match_flag_type`.

M2/Macaulay2/m2/basictests/A06.m2
mahrud: They're not deleted, simply moved to `tests/normal/regex.m2`.

M2/Macaulay2/m2/basictests/regex.m2
mahrud: They're not deleted, simply moved to `tests/normal/regex.m2`.

M2/Macaulay2/m2/exports.m2
mahrud: This is added as a part of the regex PR. They show up in the diff because this PR is based on that one. I'll discuss it there.

.github/workflows/test_build.yml
DanGrayson: > There's no need, it always comes with boost.

If that were true, you wouldn't have to add it to the workflow.

Also, I don't see "libboost-regex-dev" anywhere in the "INSTALL" file.

mahrud: > Also, I don't see "libboost-regex-dev" anywhere in the "INSTALL" file.

I just checked, `libboost-regex-dev` is installed as a recommended package along with `libboost`. That's why it's not mentioned. But sure, I'll add it in the regex PR.

> If that were true, you wouldn't have to add it to the workflow.

Notice that libboost is not installed on the workflow.

mahrud:
@DanGrayson let's keep the discussion related to the regex PR there. The changes appear here because this PR is on top of that one, but that one needs to be merged first.

.github/workflows/test_build.yml
DanGrayson: > Also, the configure script does check for it thanks to @d-torrance.

Oh, good.

DanGrayson: I've just checked: under Fedora, no new library is needed, for `/usr/lib64/libboost_regex.so` is included in `boost-devel`, which we already call for in `INSTALL`.

DanGrayson:
> @DanGrayson let's keep the discussion related to the regex PR there. The changes appear here because this PR is on top of that one, but that one needs to be merged first.

Yes, that confused me.  I'll certainly get confused again.

M2/Macaulay2/m2/exports.m2
```diff
@@ -1253,6 +1256,7 @@ exportMutable {
 	"printingTrailLimit",
 	"programPaths",
 	"recursionLimit",
+	"regexFlags",
```
DanGrayson, line 1259: This variable should be eliminated: having a global variable that changes default basic behavior is a sure way to cause bugs.

M2/Macaulay2/d/boost-regex.cpp
DanGrayson: Re:
```m2
-- Keep this enum in sync with RegexFlags in Macaulay2/e/regex.cpp
RegexFlags = new HashTable from {
...
```

Let's not introduce any more pairs of enums that have to be kept in sync, for it will be a future headache.  Instead, create core variables with names such as Icase and values such as 32.  Then you could write `"Icase"=>Icase` in regex.m2

M2/Macaulay2/m2/exports.m2
mahrud: Again, this is related to the regex PR, not here. And no, it doesn't cause bugs. You can change the default to PosixPerl and M2 still loads and packages work properly.

DanGrayson: It will definitely cause a bug eventually.

M2/Macaulay2/d/boost-regex.cpp
mahrud: That doesn't make sense. Don't you still need to keep the value 32 in sync between regex.m2 and boost-regex.cpp?

DanGrayson: Oh, I should have added that the variable Icase would be defined in regex.dd, where the enum values are available.

DanGrayson: Alternatively, the code for translating the string "Icase" to the number 32 could be entirely in regex.dd.

mahrud: You worry about a "future headache" that seems completely imaginary to me. Plenty of programs pass around status codes between different environments like this. There are tests to make sure that the exported and documented options, namely RegexPOSIX and RegexPerl, are kept in sync.

mahrud:
I added the following to the documentation for regex:
image

M2/Macaulay2/m2/basictests/A06.m2
mahrud: This is fixed in the latest commit. Lists are no longer accepted as input.
The assertions still exist and pass correctly, shall I mark this as resolved?

M2/Macaulay2/d/boost-regex.cpp
mahrud: I considered this, but there's a type error (at least when using clang) when using the suggestion, so I'll mark this as resolved.

M2/Macaulay2/d/boost-regex.cpp
mahrud: Shall I mark this as resolved? As I explained, the two flags are included in `regex::extended` by default, so we remove them to match the behavior of GNU regex.

M2/Macaulay2/d/boost-regex.cpp
mahrud: Is this still not resolved?

M2/Macaulay2/m2/exports.m2
mahrud: I removed this.

M2/Macaulay2/m2/basictests/A06.m2
DanGrayson: I just tried it with you latest commit, and the bug is still present (that's what this conversation is about).  These answers should be the same:
```m2
i1 : (regex(".*(.*)","ab"))#1

o1 = (0, 2)

o1 : Sequence

i2 : (regex("(.*)(.*)","ab"))#2

o2 = (2, 0)

o2 : Sequence
```

mahrud: This is separate from your previous comment here, and we can keep talk about it in the regex PR thread. Shall I close this particular conversation? `Flags => {}` is no longer acceptable.

M2/Macaulay2/packages/SimpleDoc/doc.txt
```diff
@@ -1,192 +1,263 @@
...
+  Headline
+    simpler documentation for functions and methods
+  Description
+    Text
+      This package provides a human-readable serialization format for documentation nodes
+      as an alternative to the classic way of @TO "Macaulay2Doc::writing documentation"@
+      using the @TO document@ method.
+
+      The new format is parsed by @TO (doc, String)@ for writing either a single or
+      multiple documentation nodes from string or file and adheres to the off-side rule;
```
DanGrayson, line 14: change the semicolon to a comma

I don't know what the "off-side rule" is, and neither will the reader, so why introduce the term?

M2/Macaulay2/packages/SimpleDoc/doc.txt
```diff
@@ -1,192 +1,263 @@
...
-    purpose is to allow the documentation of the package to be arranged
-    linearly, as in a book.  The lines in the {\bf Subnodes} section beginning
-    with a colon are subheadings used to classify the subnodes.
+    The string {\tt s} contains a series of sections describing the documentation to be
+    produced. The first line of each section contains only a keyword that specifies the
+    purpose of that section. The keywords must all be indented to the same level, and
+    everything else must be further indented. Lines beginning with white space followed
+    by two hyphens are ignored and can be used for commenting.
+
+    @TO (document, String)@ is a synonyms for @TO doc@.
```
DanGrayson, line 97: change "synonyms" to "synonym"

M2/Macaulay2/packages/SimpleDoc/helpers-doc.txt
```diff
@@ -0,0 +1,81 @@
...
+  Key
+     arXiv
+    (arXiv, String)
+    (arXiv, String, String)
+  Headline
+    link to an arXiv article
+  Usage
+    arXiv ref
+    arXiv(ref, title)
+  Inputs
```
DanGrayson, line 40: How would arXiv be used inside a doc string?  Inside at-signs?

M2/Macaulay2/packages/SimpleDoc/helpers-doc.txt
```diff
@@ -0,0 +1,81 @@
...
+      html oo
+  SeeAlso
+    document
+    stacksProject
+    wikipedia
+Node
+  Key
+     stacksProject
+    (stacksProject, String, String)
+  Headline
```
DanGrayson, line 62: How would this be used in a doc string?

M2/Macaulay2/packages/SimpleDoc/doc.txt
```diff
@@ -1,192 +1,263 @@
...
-Node
-  Key
-    "docExample"
-  Headline
-    an example documentation node
-  Usage
-    docExample
-  Description
+     The variable {\tt testExample} is a @TO String@ containing an example of
+     the use of @TO (TEST, String)@ to write a test case.
```
DanGrayson, line 58: To get `TO (TEST, String)` to work, you have to direct it to the Macaulay2Doc package, which is where that documentation node is.

M2/Macaulay2/packages/SimpleDoc/helpers-doc.txt
```diff
@@ -0,0 +1,81 @@
+-- -*- mode: M2; coding: utf-8 -*-
+Node
+  Key
+     wikipedia
+    (wikipedia, String)
+    (wikipedia, String, String)
+  Headline
```
DanGrayson, line 7: How would this be used in a doc string?

M2/Macaulay2/packages/SimpleDoc/helpers-doc.txt
```diff
@@ -0,0 +1,81 @@
...
+  Key
+     stacksProject
+    (stacksProject, String, String)
+  Headline
+    link to a Stacks Project tag
+  Usage
+    stacksProject(tag, description)
+  Inputs
+    tag:String
+      an alphanumeric Stacks Project tag "TTTT"
```
DanGrayson, line 68: What information is this "TTTT" trying to convey?

M2/Macaulay2/packages/SimpleDoc/example.m2
```diff
@@ -0,0 +1,65 @@
...
+  Consequences
+   Item
+    The first side effect of the function, if any, is described here.
+   Item
+    The second side effect of the function is described here, and so on.
+  Description
+   Text
+     Each paragraph of text begins with \"Text\".  The following
+     line starts a sequence of Macaulay2 example input lines.
+     However, see @TO (matrix,List)@.
```
DanGrayson, line 30: To get `TO (matrix,List)` to work, it has to be directed to Macaulay2Doc

M2/Macaulay2/packages/Macaulay2Doc/functions/replace-doc.m2
```diff
@@ -21,29 +21,40 @@ doc ///
     re:String
       a @TO2 {"regular expressions", "regular expression"}@ describing a pattern
     replacement:String
-      a replacement or regex formatting string, may include backreferences
+      following the @TO2 {"regular expressions", "formatting syntax"}@
```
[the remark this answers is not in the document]
mahrud: I had it that way at first, but the end result of your suggestion is somewhat jarring:
> replacement, a string, a replacement string, following the formatting syntax

Currently it is this, which I think is more readable and carries all the information:
> replacement, a string, following the formatting syntax

Don't you agree that the former is redundant?

M2/Macaulay2/packages/Macaulay2Doc/functions/select-doc.m2
```diff
@@ -55,10 +56,10 @@ doc ///
       The @TT "POSIX => true"@ option can be used to specify the POSIX Extended flavor for the regular
       expression used to match. Note that only the Perl flavor allows the use of lookahead and lookbehinds.
     Example
-      s = "catfish cats dogs"
-      select("cat(?!fish)", s)
+      s = "catfish cats dogs";
```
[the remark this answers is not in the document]
mahrud: I wanted to demonstrate the power of lookarounds. The POSIX option doesn't have anything particular benefits to offer for select.

M2/Macaulay2/packages/SimpleDoc/doc.txt
mahrud: It is a well known term: https://en.wikipedia.org/wiki/Off-side_rule

M2/Macaulay2/packages/SimpleDoc/helpers-doc.txt
mahrud: Yes.

M2/Macaulay2/packages/SimpleDoc/helpers-doc.txt
mahrud: That the tag is a four character string.

M2/Macaulay2/packages/Macaulay2Doc/functions/replace-doc.m2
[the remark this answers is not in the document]
mahrud: Also, my usage matches the one in the `simpleDocFrob` example:
image

M2/Macaulay2/packages/Macaulay2Doc/doc14.m2
```diff
@@ -100,7 +87,7 @@ document { Key => {eint,(eint, QQ),(eint, ZZ),(eint, RR)},
      EXAMPLE lines ///
      	  eint 2
      ///,
-     Wikipedia "Exponential_integral"
+     PARA {"See ", wikipedia "Exponential integral", "."}
```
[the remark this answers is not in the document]
mahrud: What? It still does:
```m2
i1 : wikipedia "Example"

o1 = Example (see https://en.wikipedia.org/wiki/Example )

o1 : HREF
```
The improvement is that it doesn't forcibly enclose the link in a paragraph, so you can use it in a sentence. As an example, I'm using it to link to the off-side rule page in the wikipedia (in my next commit).

M2/Macaulay2/packages/Macaulay2Doc/functions/select-doc.m2
```diff
@@ -26,14 +26,14 @@ doc ///
     re:String
       a @TO2 {"regular expressions", "regular expression"}@ describing a pattern
     replacement:String
-      a replacement or regex formatting string, may include backreferences
+      following the @TO2 {"regular expressions", "formatting syntax"}@
```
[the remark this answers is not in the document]
mahrud: See the response under `replace-doc.m2`.

M2/Macaulay2/packages/Macaulay2Doc/functions/select-doc.m2
```diff
@@ -26,14 +26,14 @@ doc ///
...
     str:String
       a subject string to be searched
     POSIX=>Boolean
       if true, interpret the @TT "re"@ using the POSIX Extended flavor, otherwise the Perl flavor
   Outputs
     :List
-      a list of mutually exclusive substrings of @TT "str"@ matching the pattern @TT "re"@;
+      of mutually exclusive substrings of @TT "str"@ matching the pattern @TT "re"@;
```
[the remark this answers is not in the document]
mahrud: I think I'm following a better convention here. See for instance:
https://github.com/Macaulay2/M2/blob/e38d77162a58137d258a7b8473c0ec57254413b7/M2/Macaulay2/packages/Macaulay2Doc/functions/generators-doc.m2#L104-L106
Resulting in:
image
The comma is extra, but I'd rather endure an extra comma than more redundant words.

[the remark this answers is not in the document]
mahrud: Also, if you grep for `\"of ` you'll find many more examples of this.

M2/Macaulay2/packages/SimpleDoc/doc.txt
mahrud: I'll demonstrate an example of `wikipedia` here.

M2/Macaulay2/packages/SimpleDoc/helpers-doc.txt
mahrud: I'll mention an example in the documentation for `doc(String)`.

M2/Macaulay2/packages/SimpleDoc/helpers-doc.txt
mahrud: I'll mention an example in the documentation for `doc(String)`.

M2/Macaulay2/packages/SimpleDoc/helpers-doc.txt
mahrud: I'll mention an example in the documentation for `doc(String)`.

M2/Macaulay2/packages/Macaulay2Doc/functions/replace-doc.m2
```diff
@@ -21,29 +21,40 @@ doc ///
...
     str:String
       a subject string to be processed
     POSIX=>Boolean
       if true, interpret the @TT "re"@ using the POSIX Extended flavor, otherwise the Perl flavor
   Outputs
     :String
-      the string obtained from @TT "str"@ by replacing its substrings matching @TT "re"@ by copies of @TT "replacement"@.
+      obtained from @TT "str"@ by replacing its substrings matching @TT "re"@ by copies of @TT "replacement"@
```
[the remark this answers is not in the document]
mahrud: Same as the others. This is correct. In the output, it will say "a string, obtained from ...". Declaring the type as String and then immediately saying that it is "a string" is redundant.

M2/Macaulay2/packages/Macaulay2Doc/functions/replace-doc.m2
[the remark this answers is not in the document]
DanGrayson: > I had it that way at first, but the end result of your suggestion is somewhat jarring:
> 
> > replacement, a string, a replacement string, following the formatting syntax
> 
> Currently it is this, which I think is more readable and carries all the information:
> 
> > replacement, a string, following the formatting syntax
> 
> Don't you agree that the former is redundant?

No, because the identifier being described, "replacement", is not part of its own description, so the descriptive text your way contains only "a string, following the formatting syntax", is is both non-grammatical, because of the comma, and incomplete.

M2/Macaulay2/packages/Macaulay2Doc/functions/select-doc.m2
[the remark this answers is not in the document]
DanGrayson: The presence of examples after a paragraph explaining the POSIX option strongly suggests to the reader that the examples there are examples of the POSIX option being used.  The examples are fine -- just add a paragraph above them saying something like "Here are some examples demonstrating the power of lookarounds."

M2/Macaulay2/packages/SimpleDoc/helpers-doc.txt
DanGrayson: Then say something like "an alphanumeric Stacks Project tag of the form "TTTT""

M2/Macaulay2/packages/Macaulay2Doc/functions/replace-doc.m2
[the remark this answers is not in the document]
mahrud: I don't think it's non-grammatical to say, for instance, "X, a Sequence, following the description below". So why is it non-grammatical? They are not supposed to be full sentences. Indeed, the "conventions for documentation" says:
> Inputs, Outputs, and Consequences should not end with periods

M2/Macaulay2/packages/Macaulay2Doc/doc14.m2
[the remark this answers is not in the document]
DanGrayson: > What? It still does:
> 

Oh, you're right.

M2/Macaulay2/packages/SimpleDoc/helpers-doc.txt
mahrud: Sure.

M2/Macaulay2/packages/SimpleDoc/helpers-doc.txt
DanGrayson: Why is "wikipedia" in the SimpleDoc package now?  It has nothing to do with it, so why move it there from Core?  Similarly for the other ones.

M2/Macaulay2/packages/Macaulay2Doc/functions/replace-doc.m2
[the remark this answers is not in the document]
mahrud: Here is a full sentence: "The inputs of this function are: X, a Sequence, following the description above; Y, a String, at most 5 characters; Z, an Integer, positive." Is there a grammatical error here?

M2/Macaulay2/packages/Macaulay2Doc/functions/select-doc.m2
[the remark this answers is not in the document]
mahrud: It is already done. Not sure why it doesn't show here that the file is outdated.

M2/Macaulay2/packages/SimpleDoc/helpers-doc.txt
mahrud: In this package people are a lot more likely to find the documentation than in Macaulay2Doc. I considered doing the same thing as the Text package, which only contains documentation and no definitions, but `wikipedia` is a single line and never used in Core, so there's no point in it being there either.

M2/Macaulay2/packages/Macaulay2Doc/functions/replace-doc.m2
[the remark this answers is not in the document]
DanGrayson: What's incomplete about  "a string, following the formatting syntax"  is that it doesn't say what the string is used for.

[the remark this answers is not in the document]
mahrud: > What's incomplete about "a string, following the formatting syntax" is that it doesn't say what the string is used for.

It doesn't have to, because the description of the inputs only needs to clarify what the input can be, not what it is going to be used for. That's the job of other parts. In particular, the description of the output clarifies it:
> a string, obtained from str by replacing its substrings matching re by copies of replacement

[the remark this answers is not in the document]
DanGrayson: > > What's incomplete about "a string, following the formatting syntax" is that it doesn't say what the string is used for.
> 
> It doesn't have to, because the description of the inputs only needs to clarify what the input can be, not what it is going to be used for. That's the job of other parts. 

That's not the way I think of it.  Oh, well, to each his own.

M2/Macaulay2/packages/Macaulay2Doc/functions/replace-doc.m2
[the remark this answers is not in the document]
DanGrayson: Maybe we should get rid of that pesky comma, which is automatically inserted.

M2/Macaulay2/packages/Macaulay2Doc/functions/replace-doc.m2
```diff
@@ -21,29 +21,42 @@ doc ///
...
+      the corresponding parenthesized subexpression of @TT "re"@.
     Example
       replace("(\\w+)\\.?", "A \\1.", "Dog cat cat.")
     Text
       Special operators such as the lowercase operator @TT "\\\\L"@ may also be used to manipulate the
       replacement substring.
     Example
       replace("(\\w+)\\.?", "A \\L$1.", "Dog cat cat.")
+    Text
+      Lookahead and lookbehinds can be used to precisely control the regular expression matches.
```
DanGrayson, line 49: Make both nouns plural.

M2/Macaulay2/packages/SimpleDoc.m2
```diff
@@ -1,332 +1,291 @@
...
+
+-- Primary functions
+doc = method()
+doc String := str -> (
+    docstring := if fileExists str then get str else str;
+    parsed := toDoc(NodeFunctions, docstring);
+    document \ (
+	if all(parsed, elt -> instance(elt, Node)) then apply(parsed, node -> toList node)
+	else if not any(parsed, elt -> instance(elt, Node)) then {parsed}
+	else error("expected either a single or a list of documentation nodes")))
```
DanGrayson, line 35: This function now returns a list of nulls, but it shouldn't.  (Use `scan` instead of `apply`.)
```m2
Macaulay2, version 1.16.0.2
               MinimalPrimes, PrimaryDecomposition, ReesAlgebra, TangentCone, Truncations

i1 : doc ///
     Key
      a
     ///

o1 = {}

o1 : List

i2 : peek oo

o2 = {null}
```

mahrud: Ah, this is intentional, but it'll make more sense in the next PR, where `document` will actually return the nodes so you can inspect them directly with `show` and other commands. I'll change it for now.

M2/Macaulay2/packages/Macaulay2Doc/functions/replace-doc.m2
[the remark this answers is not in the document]
mahrud: I would be in favor of a different type of output, like:
> X : String -- blah blah

But for now, this seems fine:
image

I think a much needed improvement is more fine grained type checking and type printing, for instance printing "a list of lists of integers" or "a list of ideals" instead of just "a list".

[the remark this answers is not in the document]
DanGrayson: > I think a much needed improvement is more fine grained type checking and type printing, for instance printing "a list of lists of integers" or "a list of ideals" instead of just "a list".

That information, however, is not available.

M2/Macaulay2/packages/Macaulay2Doc/functions/replace-doc.m2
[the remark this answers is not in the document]
mahrud: Shall I mark this, and the other "antecedent missing" comments, as resolved?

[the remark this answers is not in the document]
DanGrayson: Sure.

M2/Macaulay2/packages/Macaulay2Doc/changes.m2
```diff
@@ -36,6 +36,9 @@ document {
      UL {
          LI { "functionality improved",
               UL {
+                   LI { "The ", TO document, " function now accepts three new options: ", TT "Acknowledgement", ", ",
+		       TT "Contributors", ", and ", TT "References", ". The content of each will appear under a subsection
+		       of the same name in the documentation." },
```
DanGrayson, line 41: Mention also that "document" will now accept a string, doing the same thing "doc" would have done.

M2/Macaulay2/packages/SimpleDoc.m2
```diff
@@ -32,7 +32,7 @@ doc String := str -> (
     document \ (
 	if all(parsed, elt -> instance(elt, Node)) then apply(parsed, node -> toList node)
 	else if not any(parsed, elt -> instance(elt, Node)) then {parsed}
-	else error("expected either a single or a list of documentation nodes")))
+	else error("expected either a single or a list of documentation nodes"));)
```
DanGrayson, line 35: "a single" what?  I think this construction doesn't work.  Try "a documentation node or a list of documentation nodes".

M2/Macaulay2/packages/SimpleDoc/helpers-doc.txt
```diff
@@ -65,7 +65,7 @@ Node
     stacksProject(tag, description)
   Inputs
     tag:String
-      an alphanumeric Stacks Project tag "TTTT"
+      an alphanumeric Stacks Project tag of the form "TTTT"
```
DanGrayson, line 68: Good.

DanGrayson (changes requested):

M2/Macaulay2/packages/SimpleDoc.m2
mahrud: how about "one or more documentation nodes"?

M2/Macaulay2/packages/Macaulay2Doc/changes.m2
mahrud: Sure.

M2/Macaulay2/packages/SimpleDoc.m2
DanGrayson: Too ambiguous: it could mean a list of length greater than 0 is the only thing accepted.  Or it could mean a sequence of nodes is accepted.  Or a set of nodes.

DanGrayson (approved):

Issue #1149, opened 2020-05-13
Build path included in example output


d-torrance:
The build path is included in quite a few example output files used in the documentation.  This makes [reproducible builds](https://reproducible-builds.org/) impossible.

For example

```
profzoom@gloria:~$ diff -u /usr/share/doc/Macaulay2/Macaulay2Doc/example-output/___Print.out ~/tmp/m2/usr/share/doc/Macaulay2/Macaulay2Doc/example-output/___Print.out 
--- /usr/share/doc/Macaulay2/Macaulay2Doc/example-output/___Print.out	2019-11-30 13:05:14.000000000 -0500
+++ /home/profzoom/tmp/m2/usr/share/doc/Macaulay2/Macaulay2Doc/example-output/___Print.out	2015-09-06 00:03:15.000000000 -0400
@@ -2,7 +2,7 @@
 
 i1 : code Thing#{Standard,Print}
 
-o1 = /home/dan/src/M2/M2.git/M2/Macaulay2/m2/robust.m2:130:28-146:19: --source code:
+o1 = /build/macaulay2-1.14.0.1.730.g1557b415b/M2/Macaulay2/m2/robust.m2:130:28-146:19: --source code:
      Thing#{Standard,Print} = x -> (
           oprompt := concatenate(interpreterDepth:"o", toString lineNumber, " = ");
           save := printWidth;
```

I suppose this could be fixed using `sed` during build, swapping out the build directory for the eventual installation directory.

DanGrayson:
Reproducible builds might be a good idea -- do you have an application for them?

To fix the problem of including absolute paths in files, we could switching to relative paths in all such output.  The relative path would always be the same.

d-torrance:
Reproducible builds are a big thing for Debian (https://wiki.debian.org/ReproducibleBuilds).  I stumbled on this build path issue because of a bunch of Lintian warnings from my current draft of the Macaulay2 Debian package.  In particular, "file-references-package-build-path":

> The listed file or maintainer script appears to reference the build path used to build the package as specified in the Build-Path field of the .buildinfo file.
> 
> This is likely to cause the package to be unreproducible, but it may also indicate that the package will not work correctly outside of the maintainer's own system.
> 
> Please note that this tag will not appear unless the .buildinfo file contains a Build-Path field. That field is optional. You may have to set DEB_BUILD_OPTIONS=buildinfo=+path or use --buildinfo-option=--always-include-path with dpkg-buildpackage when building.
> 
[...]

PR #1247, opened 2020-06-07
Remove build paths from documentation

d-torrance:
Quite a few times in the documentation, the build path appears.  For example, in [the debugger](https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.15/share/doc/Macaulay2/Macaulay2Doc/html/_the_spdebugger.html):

```
i2 : listUserSymbols

o2 = symbol   class              value
     ------   -----              -----
     g      : FunctionClosure -- g    
     location of symbol
     ------------------                                                      
     /home/dan/src/M2/M2.git/M2/Macaulay2/packages/Macaulay2Doc/demo1.m2:11:1
          
     -11:2
```

This results in builds that aren't [reproducible](https://reproducible-builds.org/), as we will get different documentation when Macaulay2 is built from a different directory or on a different machine. They were also raising many [`file-references-package-build-path`](https://lintian.debian.org/tags/file-references-package-build-path.html) Lintian warnings in the Debian package.

The attached commits remove these build paths.  A summary of the changes:

* Whenever possible, use relative paths instead of absolute paths.
* Set a dummy `HOME` environment variable when building the documentation.  The value is `/home/m2user`, inspired by the [web interface](http://habanero.math.cornell.edu:3690/).
* When relative paths don't make sense (e.g., `toAbsolutePath`), use canned examples.
* Using relative paths for the package installation directory caused build failures, so we just removed the examples that printed this information as they weren't essential.

Closes: #1149

### DIFF
--- a/M2/Macaulay2/d/actors2.dd
+++ b/M2/Macaulay2/d/actors2.dd
@@ -27,14 +27,16 @@ setupfun("reorganize",dbmreorganize);
 dbmopenin(e:Expr):Expr := (
      when e
      is a:stringCell do (
-	  if notify then stderr << "--opening database " << a.v << endl;
+	  if notify then stderr << "--opening database " <<
+	       relativizeFilename(a.v) << endl;
 	  dbmopenin(a.v))
      else buildErrorPacket("expected a string as filename"));
 setupfun("openDatabase",dbmopenin);
 dbmopenout(e:Expr):Expr := (
      when e
      is a:stringCell do (
-	  if notify then stderr << "--opening database for output " << a.v << endl;
+	  if notify then stderr << "--opening database for output " <<
+	       relativizeFilename(a.v) << endl;
 	  dbmopenout(a.v))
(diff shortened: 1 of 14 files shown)


DanGrayson:
It would be better to develop real examples rather than to use canned ones.  However, without a `chdir` command available, it seems impossible to do that for `toAbsolutePath`.  If we had that, the example could be this:
```
     changeDirectory "/tmp"
     toAbsolutePath "."
```
Oh, well.

mahrud:
Instead of the canned examples, can we run the example outputs through a regex replace? We could replace home directory with `$HOME` and build directory with `$BUILDDIR` or something like that. I think it would be relatively simple to do this in `html0.m2`.

DanGrayson:
That's a good idea.  Use "~" for the home directory, though.

Maybe "absoluteFilename" in stdiop.d could even use "~", when possible.

PR #1253, opened 2020-06-09
Quietly return an empty list if gfan is not found when calling gfanHelp

d-torrance:
Closes: #1201

### DIFF
--- a/M2/Macaulay2/packages/gfanInterface.m2
+++ b/M2/Macaulay2/packages/gfanInterface.m2
@@ -85,6 +85,8 @@ export {
 	"multiplicitiesReorder"
 }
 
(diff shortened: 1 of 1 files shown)


d-torrance:
This isn't quite right yet.  It turns out that when I tested it, I still had gfan installed.  :)  I'll force push a new version soon.

d-torrance:
Ok, it's good now :)

DanGrayson:
The real problem is that the `gfanInterface` package runs `gfan` to capture the documentation for `gfan` so it can be inserted into its own documentation.   A lot of that is useless for the user, such as the meaning of various options to `gfan`. 

- insert the documentation into the code, deleting the tidbits about command line options
- omit it entirely

PR #1256, opened 2020-06-10
Add a Github Actions hook for builds

radoslavraynov:
This is to add a [GitHub Actions](https://github.com/features/actions) build hook to the repository. I'll describe the details further down; the gist is as follows:

* Github Actions allows for an easy automation of software workflows — bilding, testing, deployment. Such a tool is an integral part of the so-called [CI/CD model](https://en.wikipedia.org/wiki/CI/CD).
* I find Github Actions quite superior to the ubiquitous [Travis-CI](https://travis-ci.org); Macaulay2 certainly doesn't need [Jenkins](https://www.jenkins.io); and [CircleCI](https://circleci.com) is probably in-between.
* The given workflow builds and tests M2 at every `master` push, `development` pull request, and everyday at 6am EDT. (Note that even if there's no activity in the codebase, the dependencies and environments could still change.)

Given how time-consuming building M2 is (around the 2-hour mark), distinct jobs can be used for the scheduled builds and pull requests, and it certainly doesn't make too much sense to have trigger builds at both `development` pull request and `master` push, although the two could technically be building different code.

Let me know what you think!

## Details
Starting from a fresh environment (macOS or Ubuntu), the workflow consists of the following steps:

1. Install the dev tools
2. Install the libraries
3. Configure (using cmake or autotools)
4. Build (including any remaining libraries and programs)
5. Test

The build matrix consists of:

* Two OSes: latest macOS (10.15) and Ubuntu 18.04. The newest Ubuntu 20.04 is available only in experimental mode.
* Two build systems: the new `cmake` one and the old `autotools` one. In the case of `cmake`, Ninja is used for the actual build. In the case of `autotools`, GNU's Make is used.
* Two compilers: `gcc9` and `default`. The latter just stands for whatever the default complier on the system is; e.g. `AppleClang` on macOS.

This results in just 8 variants, well below the limit of 20 parallel jobs (5 on macOS) for the free accounts.

In the future, a short-term update would be to add two language standards — `C++17/C11` along with the current `C++14/C11`. The switch to C++17 is desirable and should be quite painless. A long-term addition would be to add a specific version of `clang` too. Another optional branch would be to try building on Windows, although I am not a fan of it.

Finally, note that this is just *one* job, `build`, with 8 variants. It would make sense to trigger other kinds of checks when PR for a package is made, etc. The latter, for example, using the latest binary release, is quite simple to do, but not in the scope of the build job.

Note also that I am including a trigger on push to `feature/actions`, which I am happy to remove later. For the sake of reproducing the setup, my `feature/actions` is based off the upstream `master`, not the upstream `development` branch. My `pr/actions` is based off the upstream `development` and introduces a single new commit.

## Status

Perhaps more importantly, in the short-term, the hook should serve as a baseline for what works and what doesn't. In my limited experience, building M2 rarely just works.

In fact, looking at the [success rate](https://github.com/radoslavraynov/M2/actions/runs/130355320) of the builds off the current master, [this](https://giphy.com/gifs/this-is-fine-dumpster-fire-floating-1rNWZu4QQqCUaq434T/fullscreen) captures the state of affairs :) 

I could be wrong with setting up the environments, etc, but the problems I am seeing are consistent with problems I have experienced on my fresh Catalina. To that end, I'm opening #issue to address several issues and inconsistencies I've found.

Thanks!

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -1,0 +1,152 @@
+################################################################################
+#
+#  Github actions trigger (workflow script) for building Macaulay2
+#
+#    See https://help.github.com/en/actions for the documentation.
+#
+################################################################################
+
+name: Build and Test Macaulay2
+
+on:
+  # trigger a test quick build at both PR (on dev)
+  # and pushing to master (after merging or by Dan or Mike)
+  push:
+    branch:
+      - master
+      - feature/actions
+  pull_request:
+    branch:
+      - developement
+  schedule:
+    # cron time in UTC; set to 6a EDT
+    - cron: '0 10 * * *'
+
+env:
+#  CMAKE_VERSION: 3.17
+#  NINJA_VERSION: 1.9.0
+  BUILD_TYPE: Release  # add Buildfast
+  BUILD_DIR: M2/BUILD/cicd
+
+jobs:
+  build:
+    name: ${{matrix.os}}-${{matrix.build-type}}-${{matrix.compiler}}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false  # eventually make this true
+      matrix:
+        os: [macos-latest, ubuntu-latest]
+        build-type: [cmake, autotools]
+        compiler: [gcc9, default]  # add clang
+#       language: [cxx17, cxx14]
+        include:
+          - compiler: gcc9
+            cxx: g++-9
+            cc: gcc-9
+          - os: macos-latest
+            make: gmake
+          - os: ubuntu-latest
+            make: make
+
+
+    steps:
+      - uses: actions/checkout@v1  # this set up the virtual environment, which has a whole lot
+                                   # as of this writing: gcc9, clang10, cmake3.17, homebrew...
+
+# ----------------------
+#   Install missing tools and libraries using homebrew (macOS and Linux)
+# ----------------------
+
+      - name: Install missing dev tools (homebrew)
+        id: get_dev_tools_brew
+        if: ${{ runner.os == 'macOS' }}
+        # autoconf, libtool are already in the virtual env
+        run: |
+          brew config
+          brew install ninja pkg-config autoconf automake gnu-tar libtool yasm make
+
+      - name: Install libraries (homebrew)
+        id: get_dev_libraries_brew
+        if: ${{ runner.os == 'macOS' }}
+        # mpfr, gdbm, libmpc, xz are already in the virtual env
+        run: brew install boost flint ntl mpir bdw-gc libatomic_ops glpk libomp eigen ncurses
+
+# ----------------------
+#   Install missing tools and libraries using apt-get (Debian-based Linux only)
+# ----------------------
+
+      - name: Install missing dev tools (apt-get)
+        id: get_dev_tools_aptget
+        if: ${{ runner.os == 'Linux' }}
+        # make is really gmake
+        run: sudo apt-get install -y ninja-build
+
+      - name: Install libraries (apt-get)
+        id: get_dev_libraries_aptget
+        if: ${{ runner.os == 'Linux' }}
+        # trim all this garbage
+        run: sudo apt-get install -y -q autoconf bison curl emacs fflas-ffpack flex g++ gcc gfortran install-info libatomic-ops-dev libboost-dev libc6-dev libcdd-dev libgc-dev libgdbm-dev libgivaro-dev libglpk-dev libgmp3-dev libgtest-dev liblapack-dev liblzma-dev libmpc-dev libmpfr-dev libncurses-dev libncurses5-dev libntl-dev libreadline-dev libtool libxml2-dev libz-dev make openssh-server patch pinentry-curses pkg-config time unzip xbase-clients yasm zlib1g-dev polymake w3c-markup-validator git dpkg-dev gfan libeigen3-dev libtool-bin
+
+# ----------------------
+#   Steps common to all build variants
+# ----------------------
+
+      - name: Create build directory
+        id: step_build_dir
+        run: mkdir $BUILD_DIR
+
+# ----------------------
+#   Configure and build M2 using cmake
+# ----------------------
+
+      - name: Configure Macaulay2 (cmake)
+        id: step_configure_cmake
+        if: ${{ matrix.build-type == 'cmake' }}
+        working-directory: M2/BUILD/cicd  # use ${{ defaults.dir }} or something along those lines
+        env:
+          CXX: ${{matrix.cxx}}
+          CC: ${{matrix.cc}}
+        run: |
+          echo $CXX $CC
+          cmake -S../.. -B. -G"Ninja" -DCMAKE_BUILD_TYPE=Release
+
+      - name: Build Macaulay2 (cmake)
+        id: step_build
+        if: ${{ matrix.build-type == 'cmake' }}
+        working-directory: M2/BUILD/cicd
+        run: |
+          cmake --build . --target build-libraries  -j4
+          cmake --build . --target build-programs   -j4
+          cmake --build . --target install-packages -j4
+          cmake --build . --target M2-emacs         -j4
+
+# ----------------------
+#   Configure and build M2 the old way
+# ----------------------
+
+      - name: Configure Macaulay2 (autotools)
+        id: step_configure_autotools
+        if: ${{ matrix.build-type == 'autotools' }}
+        env:
+          CXX: ${{matrix.cxx}}
+          CC: ${{matrix.cc}}
+        run: |
+          cd M2
+          ${{matrix.make}} get-tools
+          ${{matrix.make}} -f Makefile
+          cd BUILD/cicd
+          ../../configure --enable-download --enable-build-libraries="readline"
+
+      - name: Build Macaulay2 (autotools)
+        id: step_build_autotools
+        if: ${{ matrix.build-type == 'autotools' }}
+        working-directory: M2/BUILD/cicd
+        run: ${{matrix.make}}
+
+# ----------------------
+#   Run tests (Skip for now given the build failures)
+# ----------------------
+
+      - name: Run Tests
+        id: step_run_tests
+        run: echo "Placeholder. Don't do anything for now"


radoslavraynov:
I've opened #1257 to track the build-related issues.

mahrud:
Thanks a lot for getting the ball rolling on this!

I'll look into the details later, but a few quick comments:
- the `-j4` for cmake builds using Ninja can only slow you down
- use `brew install readline` instead of `ncurses`
- some of the libraries you install on ubuntu are extra, at least for the CMake build. See `BUILD/docker/ubuntu/Dockerfile`
- add `-DWITH_GMP=OFF`, then you can use `gmp`, `mpfr`, and `ntl` from system.
- you don't really need `M2-emacs` ...
- you can use `-DBUILD_SHARED_LIBS=ON` to make shared libraries whenever possible.
- you can use `cpack -G "RPM;DEB"` or `cpack -G "TGZ"` to produce artifacts.
- you can use `ctest -R normal -j4` to run the normal tests, which don't take that long (except for one, that maybe should be moved).

Most important suggestion:
- have you looked into [caching dependencies](https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows)? [This post](https://cristianadam.eu/20191222/using-github-actions-with-c-plus-plus-and-cmake/) and [this one](https://cristianadam.eu/20200113/speeding-up-c-plus-plus-github-actions-using-ccache/) provide more details. I added ccache support in the CMake build specifically for this purpose, but hadn't gotten around to making things work with github yet. We should **not** be wasting time building the `build-libraries build-programs` targets every single time.

mahrud:
> In the future, a short-term update would be to add two language standards — `C++17/C11` along with the current `C++14/C11`. The switch to C++17 is desirable and should be quite painless.

I think there's an issue with Givaro preventing this.

> Another optional branch would be to try building on Windows, although I am not a fan of it.

Given how well WSL2 works, I really don't think it's worth the effort.
 
> Finally, note that this is just _one_ job, `build`, with 8 variants. It would make sense to trigger other kinds of checks when PR for a package is made, etc. The latter, for example, using the latest binary release, is quite simple to do, but not in the scope of the build job.

I think you can specify which job to build based on what files have changed, right?


Oh, also, could you give me push access to your branch? I think I know how to fix the CMake build issues.

radoslavraynov:
Thanks! I am not too concerned about caching right now (and having `-j4` won't slow you down if you have 4 physical threads of execution &mdash; any more and you're paying for context switching). What I am a bit concerned about is that building off the `master` branch failed in 5/8 cases :)

This jobs should not produce artifacts, rather it should ideally quickly validate that a commit does not break building and some minimal set of tests. We can later add jobs for creating artifacts off `release/<tag>` branches, and jobs that do top-level validation when code is pushed to the `Macaulay2/packages` directory, and may be even something else entirely when code is pushed to `Macaulay2/m2`. I can see how you might want to cache the build from the latest `master` core push to use in the latter two, but premature optimization is the root of all evil. ;)

Let's discuss this tomorrow &mdash; may even on Slack. (I have to feed a baby :))

mahrud:
> having `-j4` won't slow you down if you have 4 physical threads of execution — any more and you're paying for context switching.

I have actually confirmed that it slows you down significantly, as it forces ninja to use 4 threads even when using 1 thread is optimal (e.g. if a compilation is too heavy).

> What I am a bit concerned about is that building off the `master` branch failed in 5/8 cases :)

Well, you have the wrong prerequisites installed for CMake, so I'm surprised about the cases that worked! On Catalina, one of the issues is about Normaliz not using OpenMP properly (autotools disables linking with OpenMP), but I'm not sure about the 4ti2 issue.

radoslavraynov:
> Well, you have the wrong prerequisites installed for CMake, so I'm surprised about the cases that worked!

Can you elaborate on this. For the macOS (homebrew) prerequisites, `readline` is clearly already installed by the time one gets to `ncurses`, so that cannot be an issue ([proof](https://github.com/radoslavraynov/M2/runs/757502538?check_suite_focus=true)). What prerequisites are wrong? Same question about Linux (apt-get) prerequisites? (I can clean it up based on `BUILD/docker/ubuntu/Dockerfile` but I think I just copied it from `INSTALL` and `INSTALL-CMake.md`.)

> On Catalina, one of the issues is about Normaliz not using OpenMP properly (autotools disables linking with OpenMP), but I'm not sure about the 4ti2 issue.

So did it build fine on a fresh Catalina for you (or Mike or Dan) then?

mahrud:
@radoslavraynov I don't think using the same setup for both autotools and cmake is a good idea. The build systems are not equivalent in prerequisites. Could you separate the two?

DanGrayson:
> @radoslavraynov I don't think using the same setup for both autotools and cmake is a good idea. The build systems are not equivalent in prerequisites. Could you separate the two?

I don't see that it matters if ninja, for example, is present during the autotools build.  Why would it be bad?

mahrud:
> I don't see that it matters if ninja, for example, is present during the autotools build. Why would it be bad?

Obviously ninja being present but unused doesn't change anything. But a main component that I think should be checked is whether we build libraries and programs correctly, and if they exist on the system we might not realize that there's an issue with the build system.

Unless we don't want to check the library build process at all.

radoslavraynov:
I'll wait for #1260 and then repeat the exercise.

DanGrayson:
I think I'll pull this, as I want to see how it works.  Radoslav, do I have to do anything to enable the actions to proceed?

DanGrayson:
> > I don't see that it matters if ninja, for example, is present during the autotools build. Why would it be bad?
> 
> Obviously ninja being present but unused doesn't change anything. But a main component that I think should be checked is whether we build libraries and programs correctly, and if they exist on the system we might not realize that there's an issue with the build system.
> 
> Unless we don't want to check the library build process at all.

Yes, we do want to check that.  But that's not a reason for "separating" the two build systems, it's a reason for doubling the number of builds, so each one is done with and without the libraries being present.  Actually, we should multiply the number of builds by 2^n, where n is the number of libraries, so we can test all subsets.

radoslavraynov:
@DanGrayson, no &mdash; once committed to the repo, it will start triggering builds as prescribed: at pull request to `development`, push to `master` and every morning at 6:00am EDT.

The trigger at pushing to `master` seems like an overkill but could be very useful for contributors: e.g. I can just keep `radoslavraynov:master` synced with `Macaulay2:master` and push changes to test everything looks good.

I will later this week (or next week, still on paternity leave) add caching.

DanGrayson:
Those actions worked great, Radoslav.  Thanks for setting them up for us!

DanGrayson:
Radoslav, if I push to `development`, will that trigger a build?

Is there a way to manually trigger a build on one of the two branches?

What about when that file gets propagated to the forks belonging to other users?  Will that trigger builds?

radoslavraynov:
It shouldn't. (The reasoning was that any commit on `development` must go thru a PR, thus triggering a build. In that scenario, triggering another build at pushing to `development` would be repetitious. An exception to this is when `development` changes between the PR and merging (due to, say, another commit being merged first), but in that case the original should be rebased, which _should_ trigger another build.)

I am not aware of a manual way to trigger a build. You can trigger a failed build though. Or you send an empty commit.

Each fork would have the same actions associated to it, unless disabled somehow. That was one of the reasons I had push to master trigger a build. Ideally, the `master` should always be in a good state, so if I want to fully test my new code, I would just push it to my forked `master`.

DanGrayson:
We just got a second build in response to me pushing to development.  Take a look.
That's not bad, but it would have better if it would have waited until midnight to test it.

Testing every pull request is a good idea.  Except those that are "drafts".

radoslavraynov:
@DanGrayson, #1262 should fix it.

DanGrayson:
Radoslav, what does "cancelled" mean (see https://github.com/Macaulay2/M2/runs/767252061 ).

DanGrayson:
Radoslav, libncurses exists on the Mac in /usr/lib, and you also use homebrew to install it, so why can't we find it here?: https://github.com/Macaulay2/M2/runs/767252058

radoslavraynov:
> Radoslav, what does "cancelled" mean (see https://github.com/Macaulay2/M2/runs/767252061 )?

The `2020-06-13T01:11:27.0400540Z ##[error]The operation was canceled.` may happen if the job was manually cancelled, if some other variant in the matrix failed (not the case since we have `fail-fast: false`) or if it ran out the allowed 6 hours (not the case either). Or it might just be a bug, `failed` not being propagated.  

> Radoslav, libncurses exists on the Mac in /usr/lib, and you also use homebrew to install it, so why can't we find it here?: https://github.com/Macaulay2/M2/runs/767252058

Not sure. I get the exact same thing happening on my Catalina.

radoslavraynov:
> >Radoslav, libncurses exists on the Mac in /usr/lib, and you also use homebrew to install it, so why can't we find it here?: https://github.com/Macaulay2/M2/runs/767252058
>
> Not sure. I get the exact same thing happening on my Catalina.

But I noticed it only happens with `gcc9`, not if using the default compiler.

DanGrayson:
> > Radoslav, libncurses exists on the Mac in /usr/lib, and you also use homebrew to install it, so why can't we find it here?: https://github.com/Macaulay2/M2/runs/767252058
> 
> Not sure. I get the exact same thing happening on my Catalina.

Good. Can you send me your config.log file after it fails to find tgoto?

DanGrayson:
These actions are turning out to be really useful!

mahrud:
> I haven't seen it fail.

https://github.com/mahrud/M2/runs/806557016?check_suite_focus=true#step:8:641

DanGrayson:
Strange.  I wonder why we aren't getting that.  The first thing I would try is this:
```
diff --git a/.github/workflows/test_build.yml b/.github/workflows/test_build.yml
index debdd43b5..dcf3c74db 100644
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -133,7 +133,6 @@ jobs:
           CC: ${{matrix.cc}}
         run: |
           cd M2
-          ${{matrix.make}} get-tools
           ${{matrix.make}} -f Makefile
           cd BUILD/cicd
           # Specifying /usr as the prefix here allows us to discover gftables if it's installed there.
```

PR #1260, opened 2020-06-11
CMake maintenance for Catalina

mahrud:
The problem with building Normaliz with openmp on Catalina is that libtool is not smart and instead of adding `-Xclang -fopenmp` just adds `-fopenmp` to the end of the link command, which causes an error. I fixed part of this in `Normaliz/patch-3.8.5`, but unfortunately the libtool script that is generated in the build directory of Normaliz is also not smart, requiring a second patch *after configure*. Since the file is generated, the lines are not always a match, which makes this is a *fuzzy* patch.

It's definitely not an ideal solution, but I've tested it on Linux and Catalina and it seems to work fine on both. Suggestions for a better fix are welcome!

### DIFF
--- a/M2/BUILD/docker/ubuntu/Makefile
+++ b/M2/BUILD/docker/ubuntu/Makefile
@@ -12,6 +12,10 @@ BUILD_OPT = -DCMAKE_BUILD_TYPE=Release -DUSING_MPIR=OFF
 ## Script for building Macaulay2
 define M2_BUILD_SCRIPT
 set -xe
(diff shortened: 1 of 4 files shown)


mahrud:
@DanGrayson pull this first to avoid extra Normaliz issues.

mahrud:
#45c2306 fixes the 4ti2 issue on Catalina. Specifically, 4ti2 couldn't find the glpk installed on the system.

PR #1262, opened 2020-06-12
Fix trigger events for Github actions

radoslavraynov:


mahrud:
Why did you remove development? That's actually very useful!

radoslavraynov:
>edit: never mind, I see, it now applies to all non-draft pull requests.
 
that's what I thought too but it behaved a little off, so I brought it back.

mahrud:
Is there a way to prevent all forks from running the same actions as well? The emails are annoying.

radoslavraynov:
You can disable it for your fork from `Settings > Actions > Disable Actions for ...`. It's worth adding this alongside the "How to fork" in the README.

One can add a `if: github.repository == 'macaulay2/M2'` to the build job to prevent it from running anywhere but on the main repo, but I like being able to push to my local `master` to test. (On the other hand, I can just set up my local branch to do it, so I am open to adding the conditional.)

DanGrayson:
Disabling actions on your fork is pretty easy, and people can probably figure it out themselves, even without it being in a README.

mahrud:
I don't think it's easy (e.g: I couldn't figure it out), and I really think
we should consider the environmental impact of having every commit pushed
to any fork of Macaulay2 starting such a computing intensive task.

Maybe a smaller check for packages only would be okay, since most people
only develop packages, but not the whole build.

Another thing to consider is that I have forked and contributed to plenty
of repositories with Actions, but I've never seen one run the Actions in my
fork.

mahrud:
Why not select this option under Actions:
image

DanGrayson:
> Why not select this option under Actions:
> image

I can't figure out what that option means.  If the code is not within the repository or the organization, where is it and why would it be considered?

mahrud:
I presume it means don't run the action on other people's forks, since
those repositories are not within the Macaulay2 organization.

DanGrayson:
It seems that the code for an action can exists elsewhere:

https://help.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#referencing-an-action-from-a-public-repository

So that option is probably an option the owner of a fork could select.  It would work, provided we also refer to the action by location with syntax such as `uses: Macaulay2/M2-actions`.  But you're looking for a solution that doesn't require users to do anything.

DanGrayson:
It will be surprising to those who have forked M2 to see those actions running -- maybe it would be good to add the conditional `if: github.repository == 'Macaulay2/M2'`.  People who want actions of their own to run on their own forks can figure out how to do that by adding a branch for testing that has an extra file.

PR #1266, opened 2020-06-13
Feature: markdown(Hypertext)

mahrud:
This PR ensures the Core satisfies the following:
- [x] all types that inherit from `Hypertext` or `MarkUpType` are in `hypertext.m2` (previously `html0.m2`)
- [x] `net` and `info` methods for such objects are moved in `format.m2` (previously `hypertext.m2`)
- [x] `html` methods for such objects are moved in `html.m2` (previously `hypertext.m2`)
- [x] `tex` methods for such objects are moved in `latex.m2` (previously `hypertext.m2`)
- [x] a new `markdown` method for such objects is added in `markdown.m2`
- [x] the previous content of `html.m2` is temporarily in `installPackage.m2`

To try out the markdown generation, try something like this:
```
debug Core
markdown help sum
```

Issue #1142, opened 2020-05-13
Hypertext types update to allow more attributes


pzinn:
Title says it all. Descendants of `Hypertext`, i.e., types which imitate the structure of html, should be made more useful by adding more allowed attributes.
e.g., `IMG` should have attributes `width`, `height`, `style` and probably plenty more.

I will look into it soon, but this is just so I don't forget.

DanGrayson:
Good idea.  Any volunteers?

DanGrayson:
Oh, I guess you're volunteering, so I'll assign you.

pzinn:
Found a couple of bugs along the way. One of the more annoying ones:
```
i1 : needsPackage "Text"
o1 = Text
o1 : Package

i2 : new LI from "hi" -- should accept but doesn't
stdio:2:1:(3): error: expected new class to be a specialization of the old one

i3 : LI "hi" -- works fine
o3 = hi
o3 : LI

i4 : new LI from 1 -- should probably reject but not with cryptic message
stdio:4:1:(3): error: expected new class to be a specialization of the old one

i5 : LI 1 -- should probably reject but not with cryptic message
stdio:5:1:(3): error: expected new class to be a specialization of the old one
```

pzinn:
I agree with this comment:
```
TEX = withQname_"#PCDATA" new MarkUpType of Hypertext -- TEX really needs to be processed further so its output can be checked, too!

```

pzinn:
... and with this one:
```
DIV1 = withQname_"div" withOptions_{"class"=>"single"} new MarkUpType of HypertextContainer -- phase this one out!
```
once `DIV` has all its options, is there any reason to keep `DIV1 {...}` rather than writing `DIV{"class"=>"single",...}`? I don't think so.

DanGrayson:
> i4 : new LI from 1 -- should probably reject but not with cryptic message
> stdio:4:1:(3): error: expected new class to be a specialization of the old one
> 
> i5 : LI 1 -- should probably reject but not with cryptic message
> stdio:5:1:(3): error: expected new class to be a specialization of the old one
> ```

Maybe the message should be "expected new class to be a specialization of the old one or to find a matching method for `NewFromMethod`".

DanGrayson:
> TABLE {{1},{2,3},{4,5}}

On the other hand, there is the function `validate`:

```
i14 : validate TABLE {{1},{2,3},{4,5}}
[...]

PR #1267, opened 2020-06-13
Hypertext types update

pzinn:
this implements #1142 and fixes a few bugs in the process:
- all `MarkUpType` (that are actual HTML entities) have options, as should be -- they have HTML global options, event options if applicable, and specific options to that tag.
- `new TABLE from List` created
- new function `style` which allows to more easily handle the option "style" => (series of options separated by ;)
- the `html` method is now created by `withQName` rather somewhat cryptically by the global hook.
(also, allows for non globally assigned mark up types)
- bug fixed in `html` in case list is a singleton.
- new types: `OL`, `TH`.
- `DIV1` removed.
- `html` methods for `Matrix` and `Vector` -- there were already methods for `MatrixExpression` and `VectorExpression` but they were not used. in the process, fixed a bug in `expression VectorExpression` (a method I wrote, shame on me).
- cleanup of the start of `html0.m2` with in particular better use of self-initialization.
- package "Text" correspondingly updated.
- though this PR is not directly related to my web app, one may wish to consult the tutorial lesson
http://115.146.94.251:8002/?tutorial=4&lesson=2
which nicely illustrates the new features.
- hopefully no documentation output was modified by any of the above.
- this PR has interactions with #1266.


pzinn:
after discussing the matter with Mahrud, we'll work together on incorporating my changes into his PR. closing

Issue #1092, opened 2020-04-11
Replace Macaulay2/regex with better regex engine


mahrud:
`regex.h` exists on modern systems and is compatible with the rest of our code. Only this chunk would need to be removed or put in a `#ifndef HAVE_REGEX_H` clause:
https://github.com/Macaulay2/M2/blob/002e48263d67673713f9ef42422484181eee1978/M2/Macaulay2/d/scclib.c#L975-L986
See https://github.com/mahrud/M2/commit/57c7ff009a61f308b199df64127f2e6d6c74c2e1 for an example of a fix while keeping the regex directory (note that this also makes building object files in the regex directory unnecessary), though I think we would be fine deleting it.

I tested this on the CMake build and works fine, at least on Linux. A test on Mac would be great.

mahrud:
@DanGrayson, I implemented and tested this on the master branch with this commit: https://github.com/mahrud/M2/commit/111fd551427e844a87151e9bd32711b06e91ac02. Feel free to cherry-pick that commit to the master, though it might be easier to wait and merge the cmake branch, as it already includes these changes.

DanGrayson:
I'd like to get it now.  What OS's have you tested it on?

mahrud:
I've only tested it on Linux.

mahrud:
Continuing the regex discussion from #1095.

> Maybe that's why we include the regex code -- because we like the GNU syntax better than the POSIX syntax. One difference seems to be whether grouping uses `\(...\)` or `(...)`.

[This document](http://web.mit.edu/gnu/doc/html/regex_7.html) describes both APIs. The main advantage of the POSIX API seems to be its portability, so that we don't have to maintain regex code. There's also the [Perl-Compatible Regular Expression](https://www.pcre.org/) library in pcre.h which seems universal. Boost also has a regex library which [supports both](https://www.boost.org/doc/libs/1_72_0/libs/regex/doc/html/boost_regex/syntax.html) POSIX and Perl syntax, so can be a plug-in replacement if neither are present on a system.

DanGrayson:
This code is basically maintenance free, probably less of a headache than writing and maintaining code that searches for compatible code in libraries on the system.

mahrud:
It seems like [lookaheads and lookbehinds](https://www.regular-expressions.info/refadv.html) are not supported:
```
i166 : regex("A(?!B)", "AC AB")
stdio:160:1:(3): error: regex: Invalid preceding regular expression
```
[...]

PR #1278, opened 2020-06-15
New function which escapes special regex characters in a string

d-torrance:
While working on a better solution than using canned examples for some build paths appearing in the documentation, I ran into a issue when using functions involving regular expressions.  Sometimes, we want to match a literal character like "\\" or "+" in a string instead of using its special meaning.

If we know what our string is, then we can just escape the character in question with a "\\".  However, if we don't know the value of the string ahead of time, then it would be great to have a function that does this for us.

These commits add such a function (currently named `toRegex`, but I'm open to suggestions for a better name!) along with corresponding documentation and tests.


mahrud:
I think the standard name for this is "regexp quote", so maybe `regexpQuote` or `regexQuote` is a better option.

d-torrance:
> This is indeed very useful. Thank you! I think the standard name for this is "[regexp quote](https://www.google.com/search?q=regexp+quote)", so maybe `regexpQuote` or `regexQuote` is a better option.

Sounds good to me!  I'll push a new version in a moment using `regexQuote`.

> On a related note, do you know which standardized c/cpp regexp libraries are easily available on Debian? I'm asking for #1092. Boost.Regex, PCRE, or the C++ std::regex are the main ones I'm looking into.

I just checked and all three of those libraries are packaged for Debian.

d-torrance:
I force-pushed a new version with the function renamed to `regexQuote`.  (I chose this over `regexpQuote` since we already have a function named `regex` without a "p".)

PR #1282, opened 2020-06-16
updated package RelativeCanonicalResolution

michaelhoff:
submitted to JSAG 
improvements suggested by the referees


DanGrayson:
@radoslavraynov -- this pull request was originally submitted to `master`, and I edited it and redirected it to `development`.  Why are the actions not happening?

radoslavraynov:
@DanGrayson, only PRs to `development` trigger builds:

```
on:
[...]
   pull_request:
     branches:
       - development
```
Not sure if that's a bug of a featue. I can simply add `master` to it, or you ask the OP to resubmit to `development`, which contributors should start doing at some point. Let me know in case of the former.

The problem remains with not being able to trigger a build manually. (There are many bad ways to achieve that for `master`.) I will address it later when I get the chance, probably by setting up a payload for the REST API.

PR #1285, opened 2020-06-16
Use generic "/path/to/source/" path for cached files

d-torrance:
These files are not installed with Macaulay2, and instead their contents
are cached as strings inside the Macaulay2 binary.  Currently, the
original source directory is displayed when running the basic tests
or locating the code of a function defined in startup.m2.in, which results
in non-reproducible builds and also may be confusing to users.

We instead replace the beginning of these paths with "/path/to/source/"
to indicate that these files are only present in the source code.

### DIFF
--- a/M2/Macaulay2/d/Makefile.in
+++ b/M2/Macaulay2/d/Makefile.in
@@ -224,13 +224,16 @@ startup.c: startup-header.h ../m2/startup.m2 @srcdir@/../m2/basictests/*.m2 Make
 	@(\
 	 cat @srcdir@/startup-header.h; \
 	 echo 'cached_file startupFile = {' ; \
-	     echo @abs_top_srcdir@/Macaulay2/m2/startup.m2.in | sed $(BSTRING) ; \
+	     echo /path/to/source/Macaulay2/m2/startup.m2.in | sed $(BSTRING) ; \
 	     echo  ',' ; \
 	     cat ../m2/startup.m2 | sed $(SSTRING) ; \
 	 echo  '};' ; \
 	 echo 'cached_file testStrings[] = {' ; \
 	 for i in @abs_top_srcdir@/Macaulay2/m2/basictests/*.m2 ; \
-	 do echo '{' ; echo $$i | sed $(BSTRING) ; echo  ',' ; cat $$i | sed $(SSTRING) ; echo  '},' ; \
+	 do echo '{' ; \
+		echo "/path/to/source/Macaulay2/m2/basictests/`basename $$i`" | \
+		sed $(BSTRING) ; echo  ',' ; cat $$i | sed $(SSTRING) ; \
+		echo  '},' ; \
 	 done ; \
 	 echo '};' ; \
 	 cat @srcdir@/startup-trailer.h; \


d-torrance:
Here's an example of the behavior with this commit:

```
profzoom@gloria:~/src/macaulay2/M2/M2/BUILD/doug$ usr-dist/x86_64-Linux-Ubuntu-19.10/bin/M2 --check 1
-- testing /path/to/source/Macaulay2/m2/basictests/A01.m2
-- testing /path/to/source/Macaulay2/m2/basictests/A02.m2
-- testing /path/to/source/Macaulay2/m2/basictests/A03.m2
-- testing /path/to/source/Macaulay2/m2/basictests/A04.m2
-- testing /path/to/source/Macaulay2/m2/basictests/A05.m2
-- testing /path/to/source/Macaulay2/m2/basictests/A06.m2
-- testing /path/to/source/Macaulay2/m2/basictests/A07.m2
-- testing /path/to/source/Macaulay2/m2/basictests/A08.m2
-- testing /path/to/source/Macaulay2/m2/basictests/A10.m2
-- testing /path/to/source/Macaulay2/m2/basictests/A44.m2
-- testing /path/to/source/Macaulay2/m2/basictests/A66.m2
-- testing /path/to/source/Macaulay2/m2/basictests/A77.m2
-- testing /path/to/source/Macaulay2/m2/basictests/D01.m2
-- testing /path/to/source/Macaulay2/m2/basictests/L00.m2
-- testing /path/to/source/Macaulay2/m2/basictests/L00a.m2
-- testing /path/to/source/Macaulay2/m2/basictests/L01.m2
-- testing /path/to/source/Macaulay2/m2/basictests/L02.m2
-- testing /path/to/source/Macaulay2/m2/basictests/W01.m2
-- testing /path/to/source/Macaulay2/m2/basictests/W02-threads.m2
-- testing /path/to/source/Macaulay2/m2/basictests/ascii.m2
-- testing /path/to/source/Macaulay2/m2/basictests/bigreal.m2
-- testing /path/to/source/Macaulay2/m2/basictests/hashcodes.m2
-- testing /path/to/source/Macaulay2/m2/basictests/regex.m2
-- testing /path/to/source/Macaulay2/m2/basictests/version.m2
Macaulay2, version 1.15.1.0
--loading configuration for package "FourTiTwo" from file /home/profzoom/.Macaulay2/init-FourTiTwo.m2
--loading configuration for package "Topcom" from file /home/profzoom/.Macaulay2/init-Topcom.m2
               Truncations
--warning: file doc.css not installed yet in package Style

i1 : code applicationDirectory

o1 = ../../../../../../../../path/to/source/Macaulay2/m2/startup.m2.in:247:32-257:16: --source code:
          applicationDirectory = () -> (
               if getenv "HOME" =!= ""
               then (
                    getenv "HOME" | "/" |
                    -- we allow the user to change applicationDirectorySuffix to a string or a function
                    if instance(applicationDirectorySuffix, Function)
                    then applicationDirectorySuffix()
                    else applicationDirectorySuffix
                    )
               else error "environment variable HOME not set"
               );
```

d-torrance:
Note that currently this is autotools-specific.  It looks like a corresponding cmake version would require patching `cmake/startup.cmake`.

mahrud:
The point of having the path is to help with debugging. Does the path appear in the documentation anywhere?

d-torrance:
> The point of having the path is to help with debugging. Does the path appear in the documentation anywhere?

Yes, e.g., [applicationDirectory](https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.15/share/doc/Macaulay2/Macaulay2Doc/html/_application__Directory.html).  (Note that the webpage shows a relative path, but that's only because that was shorter than the absolute path, and in the stable 1.15 release, the minimum of these two is what is displayed.  One of my recent PR's made it *always* use the relative path, but I'm planning on reverting this soon because using relative paths is pointless if we build the examples from /tmp.)

I can fix the reproducibility in the documentation in another way.  In the draft I'm working on right now, it shows up as `/usr/share/Macaulay2/Core/startup.m2.in`, which is reproducible, but strange because that file doesn't exist.  That's what prompted this PR.

Even for debugging, the current behavior only makes sense if the user is working on the same machine where Macaulay2 was built, which I suppose is decently likely, but may not always be the case.

mahrud:
> Even for debugging, the current behavior only makes sense if the user is working on the same machine where Macaulay2 was built, which I suppose is decently likely, but may not always be the case.

Debugging startup.m2 requires building Macaulay2, so you would always be on the machine where M2 is built. Even if it's not always the case, it certainly doesn't make sense to make debugging harder when it is built on the same machine.

I'm a bit wary of this PR since it diverges the build systems and makes debugging harder.

d-torrance:
> Debugging startup.m2 requires building Macaulay2, so you would always be on the machine where M2 is built. Even if it's not always the case, it certainly doesn't make sense to make debugging harder when it is built on the same machine.
> 
> I'm a bit wary of this PR since it diverges the build systems and makes debugging harder.

Great point!  You convinced me.  :)

Closing.

PR #1289, opened 2020-06-16
Added DecomposableSparseSystems.m2

tbrysiewicz:
Added DecomposableSparseSystems.m2 and appended =distributed-package with that package name


mahrud:
A test for `solveDecomposableSystem` is breaking the build.

mahrud:
Ah, never mind, probably missing phcpack or polyhedra.

PR #1293, opened 2020-06-17
Limit running full builds in Actions

radoslavraynov:
Only run actions in the the upstream repo (Macaulay2/M2) or when pushing to branch `local`.

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -15,6 +15,7 @@ on:
     branches:
       - master
       - development
+      - local
   pull_request:
     branches:
       - development
@@ -30,6 +31,7 @@ env:
 
 jobs:
   build:
+    if: github.repository == 'Macaulay2/M2' || contains(github.ref, "local")
     name: ${{matrix.os}}-${{matrix.build-type}}-${{matrix.compiler}}
     runs-on: ${{ matrix.os }}
     strategy:
@@ -106,9 +108,7 @@ jobs:
         env:
           CXX: ${{matrix.cxx}}
           CC: ${{matrix.cc}}
-        run: |
-          echo $CXX $CC
-          cmake -S../.. -B. -G"Ninja" -DCMAKE_BUILD_TYPE=Release
+        run: cmake -S../.. -B. -G"Ninja" -DCMAKE_BUILD_TYPE=Release
 
       - name: Build Macaulay2 (cmake)
         id: step_build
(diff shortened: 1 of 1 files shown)


DanGrayson:
Why didn't this pull request trigger the actions?

DanGrayson:
Ah, here's why:
```
The workflow is not valid. .github/workflows/test_build.yml (Line: 34, Col: 9): Unexpected symbol: '"local"'. Located at position 61 within expression: github.repository == 'Macaulay2/M2' || contains(github.ref, "local")
```

DanGrayson:
By the way, wouldn't "global" be a better name for that branch than "local"?  Because on it the actions run globally, in everyone's repository.

DanGrayson:
Could you also activate pull requests to "master", since some people direct their requests there.

.github/workflows/test_build.yml
```diff
@@ -117,8 +117,8 @@ jobs:
         run: |
           cmake --build . --target build-libraries  -j4
           cmake --build . --target build-programs   -j4
+          cmake --build . --target M2-core          -j4
```
mahrud, line 120: `M2-core` is a dependency of `install-packages`, so this is unnecessary.

mahrud: Also, again, the `-j4` are unnecessary :)

radoslavraynov: * I like having `M2-core` there explicitly because it's informational.
* I can believe that the `e/` code could not benefit from parallelism. Not sure about `{c,d}/`. But `build-libraries` and `build-programs` are trivially parallelizable, aren't they? (May be with minor exceptions.)

mahrud:
Everything is parallelizable, and all of it is parallelized automatically
by Ninja. That argument is just extra.

.github/workflows/test_build.yml
radoslavraynov: For the `-j4`, I agree Ninja doesn't need it &mdash; it defaults to `(num cores)+2`, which is `2+2=4` for the virtual machines running the builds. I can remove it next time with the next iteration.

mahrud: Weird, my previous message logged as a global comment rather than to this. I'll quote it for the record:
> Everything is parallelizable, and all of it is parallelized automatically
by Ninja. That argument is just extra.

mahrud:
You could also skip on installing things like `emacs` and `polymake` and a bunch of other things.

Issue #1175, opened 2020-05-19
convince mpsolve authors to provide the version number to the program


DanGrayson:
... so we can fix this line in d/version.dd:
```
   "mpsolve version" => "3.1.8", -- TODO: DETERMINE VERSION #??
```
, which hardcodes the version as 3.1.8.

DanGrayson:
I've written to the authors requesting the feature.

d-torrance:
If the command-line tool is available during build, then we could get it that way:

```
profzoom@gloria:~/tmp/mpsolve/usr/bin$ ./mpsolve -v | sed 's/MPSolve //'
3.1.8
```

DanGrayson:
It is, and we're even distributing it, but it will probably be simpler just to wait for the feature.

robol:
We can definitely add that; I opened an issue on MPSolve's repo to track it, I will probably implement it soon.

Also, a slightly unrelated notice: I expect that we may release 3.1.9 with some bugfixes (and likely this change), and later there will be a major version bump because we are in the process of renaming our multiprecision floating point complex type to avoid a name clash with GNU MPC. This will break the API, although anyone not accessing directly multiprecision complex roots won't be affected.

DanGrayson:
Thanks!

robol:
I added new functions `mps_get_major_version()`, `mps_get_minor_version()`, `mps_get_patch_version()` are available in the latest release (3.2.1), along with a string representation with `mps_get_version()`. 

There are also defines statements included in the main header as follows (for 3.2.1)
```
#define MPS_MAJOR_VERSION 3
#define MPS_MINOR_VERSION 2
#define MPS_PATCH_VERSION 1
```
Best, -- Leonardo.

DanGrayson:
Thank you!

PR #1296, opened 2020-06-18
CMake maintenance

mahrud:
- [x] update frobby to v0.9.1
- [x] update mpsolve to v3.2.1
- [x] teach cmake how to check the mpsolve version
- [x] add the mpsolve version to version.dd (see #1175)
- [x] upload the mpsolve archive to Macaulay2 website and update `libraries/mpsolve/Makefile`

@DanGrayson could you do the last item? Also, mpsolve has a COPYING file that probably should be used as the license file rather than the README.

### DIFF
--- a/M2/Macaulay2/d/version.dd
+++ b/M2/Macaulay2/d/version.dd
@@ -19,6 +19,7 @@ header "
    #include <mathic.h>
    #include <gdbm.h>
    #include <mathicgb.h>
+   #include <mps/mps.h>
    #define stringize0(a) #a
    #define stringize(a) stringize0(a)
    namespace constants { extern const char* const version; } /* we have to copy this from frobby until they make it more accessible */
@@ -162,7 +163,7 @@ setupconst("version", Expr(toHashTable(Sequence(
 	       \"not present\"
 	  #endif
 	  "),
-   "mpsolve version" => "3.1.8", -- TODO: DETERMINE VERSION #??
+   "mpsolve version" => Ccode(constcharstar, "stringize(MPS_MAJOR_VERSION) \".\" stringize(MPS_MINOR_VERSION) \".\" stringize(MPS_PATCH_VERSION)" ),
    "python version" => Ccode(constcharstar,"
 	#ifdef WITH_PYTHON
 		 PY_VERSION				      
--- a/M2/Macaulay2/d/version.dd
+++ b/M2/Macaulay2/d/version.dd
@@ -165,7 +165,7 @@ setupconst("version", Expr(toHashTable(Sequence(
 	  "),
    "mpsolve version" => Ccode(constcharstar, "
 	#ifdef MPS_MAJOR_VERSION
-               stringize(MPS_MAJOR_VERSION) \".\" stringize(MPS_MINOR_VERSION) \".\" stringize(MPS_PATCH_VERSION)
+	       stringize(MPS_MAJOR_VERSION) \".\" stringize(MPS_MINOR_VERSION) \".\" stringize(MPS_PATCH_VERSION)
 	#else
 	       \"MPS_MAJOR_VERSION undefined\"
 	#endif
--- a/M2/Macaulay2/d/version.dd
+++ b/M2/Macaulay2/d/version.dd
@@ -163,7 +163,13 @@ setupconst("version", Expr(toHashTable(Sequence(
 	       \"not present\"
 	  #endif
 	  "),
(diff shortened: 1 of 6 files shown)


mahrud:
> * [ ]  upload the mpsolve archive to Macaulay2 website and update `libraries/mpsolve/Makefile`

Alternatively, autotools could also use the github version.

M2/cmake/build-libraries.cmake
```diff
@@ -1066,6 +1072,38 @@ ExternalProject_Add(build-polymake
   )
 #_ADD_COMPONENT_DEPENDENCY(programs polymake "cddlib;lrs;normaliz;nauty" POLYMAKE)
 
+
```
DanGrayson, line 1075: We don't build Bertini, because it is impossible to maintain, due to this clause in the license: "You may modify the Program for your private use only."

mahrud: Sure, which is why the target isn't automatically built. However, providing it is convenient for (my) private use! I read the license, I don't believe adding a build target violates it.

DanGrayson: Okay: if it's not built by default and does not end up in distributions, so that users come to depend on us providing it to them, then that's fine.

This option would also make it easier to test the Bertini package and to rerun its examples in a github action.

mahrud: Exactly.

mahrud:
@DanGrayson feel free to push to this branch to update the MPSolve Makefile, or you can merge this and do that as a direct commit.

M2/Macaulay2/d/version.dd
```diff
@@ -162,7 +163,7 @@ setupconst("version", Expr(toHashTable(Sequence(
 	       \"not present\"
 	  #endif
 	  "),
-   "mpsolve version" => "3.1.8", -- TODO: DETERMINE VERSION #??
+   "mpsolve version" => Ccode(constcharstar, "stringize(MPS_MAJOR_VERSION) \".\" stringize(MPS_MINOR_VERSION) \".\" stringize(MPS_PATCH_VERSION)" ),
```
DanGrayson, line 166: Could you add something like this?:
```
#ifndef MPS_MAJOR_VERSION
#error "MPS_MAJOR_VERSION undefined"
#endif
```
That will make sure we are getting the actual numbers to display.

mahrud: If I add that now it'll lead to errors in the autotools build, since currently it builds an older version of MPSolve that doesn't have those macros. Would you like to add the check along with updating MPSolve?

mahrud: I added a different fix for now.

mahrud:
@lkastner with this PR, there's a `build-polymake` target which also installs the shared libraries, if you would like to experiment with using it in the Polyhedra package. Also let me know if you have any thoughts on the configure options for polymake (around line 1055 of `cmake/build-libraries.cmake`).

DanGrayson:
> @DanGrayson feel free to push to this branch to update the MPSolve Makefile, or you can merge this and do that as a direct commit.

Okay, I updated the mpsolve makefile (by pushing to "development") and included your fix to version.dd, with a slight modification, hence the present conflict.

DanGrayson:
> > @DanGrayson feel free to push to this branch to update the MPSolve Makefile, or you can merge this and do that as a direct commit.
> 
> Okay, I updated the mpsolve makefile (by pushing to "development") and included your fix to version.dd, with a slight modification, and resolved the conflict (on your branch)

DanGrayson:
Once one of the builds completes, I'll pull it.

mahrud:
@DanGrayson I realized I forgot to add a one line typo fix to this PR, could you commit directly to development?
```patch
--- a/M2/cmake/check-libraries.cmake
+++ b/M2/cmake/check-libraries.cmake
@@ -88,7 +88,7 @@ if(USING_MPIR)
 else()
   find_package(GMP     6.0.0 REQUIRED)
   set(MP_LIBRARY GMP)
-  set(MP_ROOT ${GMP_INCLUDE_DIR}/..)
+  set(MP_ROOT ${GMP_INCLUDE_DIRS}/..)
 endif()
 # MP will mask either GMP or MPIR
 foreach(var IN ITEMS FOUND INCLUDE_DIRS LIBRARIES VERSION_OK)
```

DanGrayson:
done in b5083b57b19e01cd38fd7ac338bf4829ad6c1e56

Issue #1295, opened 2020-06-17
package PHCpack


DanGrayson:
The program phcpack isn't installed on every user's system, so detect its presence and enable re-running the examples and running the tests conditionally. This involves the use of the option OptionalComponentsPresent to newPackage. Three packages have started using this option already.

mahrud:
By the way, the link to phcpack in INSTALL is broken.

DanGrayson:
Thanks, I've fixed it.

mahrud:
The bertini link is also incorrect. This is the correct link: https://bertini.nd.edu (the latest version is v1.6 in 2018, but the version you linked has up to v1.4 in 2013).

antonleykin:
PHCpack package header has 
```
  Configuration => { 
    "path" => "",
    "PHCexe"=>"phc", 
    "keep files" => true
  },
```
Should `OptionalComponentsPresent` be checking the presence of the defaults?

DanGrayson:
The idea is that you would have code that tests for the presence of phcpack on the PATH, and, if it's there, provide "true" as the value for OptionalComponentsPresent.  Then the examples would get rerun (automatically) and you could arrange for tests to run, too.

Hmm, maybe the function "check" should be modified so if the optional components are not present, then no tests are run.

antonleykin:
I guess the question is: at what point `OptionalComponentsPresent` is getting used?
If this happens *before* the package is loaded then... e.g., the way `Tropical` is using is likely invalid.

DanGrayson:
But `Tropical` doesn't use it.  (That line is commented out.)

The value is used by `newPackage` to set `UseCachedExampleOutput` at load time.  And then at install time, that determines whether the examples are rerun or the old results are used.

PR #1302, opened 2020-06-18
added OptionalComponentsPresent to PHCpack; fixed one example

antonleykin:
This resolves 
https://github.com/Macaulay2/M2/issues/1295
and fixes one bug in `cascade`.

@janverschelde may want to take a look at this.

### DIFF
--- a/M2/Macaulay2/packages/Bertini.m2
+++ b/M2/Macaulay2/packages/Bertini.m2
@@ -313,6 +313,7 @@ bertiniPosDimSolve Ideal := o -> I -> bertiniPosDimSolve(I_*, o)
 
 
 bertiniSample = method(TypicalValue => List, Options=>{Verbose=>false,
+	BertiniInputConfiguration=>{},
 	IsProjective=>-1
   })
 bertiniSample (ZZ, WitnessSet) := o -> (n, W) -> (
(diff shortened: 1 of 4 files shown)


DanGrayson:
I just tested it and it works great.  There should probably be some tests involving running the program, though.

Also, as long as you're looking at this, could you add this option?:
```
  phc -0 : random numbers with fixed seed for repeatable runs    
```
That way, if we run the examples, the output files don't change and git doesn't notify us that something it knows about has changed.

antonleykin:
Should we just make 
```
  Configuration => { 
    "path" => "",
    "PHCexe"=>"phc -0", 
    "keep files" => true
  },
```
so that `phc` is deterministic by default?

... or do you have some other solution in mind?

DanGrayson:
That would work for us, I believe.

Why not delete the "keep files" option, since it's unused?

antonleykin:
I've derandomized PHCpack (by default) and derandomized the use of Bertini in certain functions of NumericalAG.
@JoseMath, perhaps it is a good idea to make `RandomSeed=>0` default in `Bertini` package itself.

DanGrayson:
> I've derandomized PHCpack (by default) and derandomized the use of Bertini in certain functions of NumericalAG.
> @JoseMath, perhaps it is a good idea to make `RandomSeed=>0` default in `Bertini` package itself.

That would be a good idea.  (When running tests and examples, M2 is careful to use `setRandomSeed 0`, but that wouldn't affect external programs such as `bertini`, which might have their own idea about randomization.)

PR #1303, opened 2020-06-18
new version of NormalToricVarieties

ggsmith:
The new version of the NormalToricVarieties package includes a new datatype "ToricMap" and several methods utilizing it.  There are also various improvements to the documentation.

### DIFF
--- a/M2/INSTALL
+++ b/M2/INSTALL
@@ -540,13 +540,13 @@ we briefly summarize them and the software they depend on:
  	         sudo apt-get install polymake
 
     Bertini needs bertini:
-	    http://www.nd.edu/~sommese/bertini/BertiniSource_v1.4.tar.gz
+	    http://www.nd.edu/~sommese/bertini/
 	which needs gmp and mpfr:
-	    https://gmplib.org/download/gmp/gmp-6.0.0a.tar.xz
-	    http://www.mpfr.org/mpfr-3.1.2/mpfr-3.1.2.tar.xz
+	    https://gmplib.org/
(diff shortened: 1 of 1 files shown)


DanGrayson:
Thanks for the update, Greg, but there's a problem with your pull request: you've added 26000 new files in `M2/tmp` resulting from a build of Macaulay2.  To fix that, I suggest you start a new branch based on our master branch and add the changes for your package to it.

Doing `git rm M2/tmp` on you master branch is not a good idea, as repositories retain all the history and we can do without the bloat.  Eventually you'll want to reset your master branch so it doesn't include those files in its history, too, but for now, you can just use a new branch with a new name.

ggsmith:
Dear Dan,

First, I believe that I have successful reset my branch and just added the new files related to NormalToricVarieties.  I’ve made a new pull request for this.

Second, how to you recommend that I avoid this problem in the future.  Should I create a separate branch when I compile?

Thanks,
Greg.

DanGrayson:
There is no reason to create a separate branch just for compiling.  I guess you used a git command that somehow added all of those files, but I don't have a good guess for what it might have been.  Well, it might have been `git add .` -- that would add all files, even new ones, and it would do it silently (!).  Do you remember the sequence of events?

I'll take a look at your new pull request.

mahrud:
You can also add `M2/tmp` to `.git/info/exclude`, which is your local version of the `.gitignore` file.

ggsmith:
As suggested, I've updated the file `.git/info/exclude`.

PR #1318, opened 2020-06-22
new package Graphics

pzinn:
This PR adds a new package `Graphics`. Note that there is already a package with this name; the suggestion is to rename it `OldGraphics` since it is AFAICT incomplete and unmaintained (NB: the new package `Graphics` is not based in any way on the old one).

The package introduces various graphical primitives which can then be turned into SVG objects using `html` (for e.g. insertion in a web page). `Graphics` handles 2d and 3d graphics, as well as animation; note that manual rotation of 3d graphics and animations require the auxiliary files `Graphics.css` and `Graphics.js` to be added to the web page.

Though this package is not directly related to my new web interface, one may want to consult the lesson
http://115.146.94.251:8002/?tutorial=4&lesson=3
or the documentation page
http://115.146.94.251:8002/usr/share/doc/Macaulay2/Graphics/html/index.html
for examples of use of the package.

### DIFF
--- a/M2/Macaulay2/m2/html.m2
+++ b/M2/Macaulay2/m2/html.m2
@@ -30,7 +30,7 @@ html Hypertext := x -> (
     T := class x;
     qname := T.qname;
     attr := "";
-    cont := if member(Options, keys T) then (
+    cont := if T.?Options then (
 	(op, ct) := try override(options T, toSequence x) else error("markup type ", toString T, ": ",
 	    "unrecognized option name(s): ", toString select(toList x, c -> instance(c, Option)));
 	scanPairs(op, (key, val) -> if val =!= null then attr = " " | key | "=" | format val | attr);
--- a/M2/Macaulay2/m2/expressions.m2
+++ b/M2/Macaulay2/m2/expressions.m2
@@ -364,7 +364,8 @@ Expression + Expression     := Sum => (x,y) -> new Sum from {x,y}
 	   - Minus          := x -> expression x#0
            - Expression     := x -> new Minus from {x}
            - Holder         := x -> new Minus from {x#0}
-Expression - Expression     := Sum => (x,y) -> x + -y
+Expression - Expression     := Sum => (x,y) -> x + Minus y
+Thing - Minus               := Sum => (x,y) -> expression x + y#0
 Product    * OneExpression  :=
 Expression * OneExpression  :=
 Holder     * OneExpression  := (x,y) -> x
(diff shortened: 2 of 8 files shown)


pzinn:
actually, I added to this PR a patch for a bug that's been around forever and that I keep forgetting to include in my PRs. Compare before:
```
i1 : texMath(1-ii)
stdio:1:1:(3): error: recursion limit of 300 exceeded
```
and after:
```
i1 : texMath(1-ii)

o1 = 1-\mathbf{i}
```

DanGrayson:
Thank you for the contribution.  I hesitate to break people's code by renaming an existing package, though -- what about calling this one `Graphics2`?

mikestillman:
or even SVGGraphics?

M2/Macaulay2/m2/html.m2
```diff
@@ -30,7 +30,7 @@ html Hypertext := x -> (
     T := class x;
     qname := T.qname;
     attr := "";
-    cont := if member(Options, keys T) then (
```
mahrud, line 33: Was this a bug?

pzinn: not a bug. I must've included this unintentionally. anyway, `T.?Options` is the standard way to check for the existence of the key `Options`.

DanGrayson:
Paul, what about changing the name of the new package?  (Hopefully today.)

pzinn:
`Graphics2` suggests it's v2 of `Graphics`, which it isn't.
`SVGGraphics` is too long (not to mention, the G is a repeat).
I've ran out of ideas to rename (again) the package.

mahrud:
For what it's worth, I think SVG.m2 is a reasonable name, given that SVG is what people call those objects (e.g. nobody spells out PNG or HTML).

If that's not convincing, VectorGraphics.m2 might be a reasonable alternative, albeit kinda long.

Also, @pzinn, is there anything that the old package does that the new one doesn't? If not, we could move the old one to OldGraphics.m2 instead. In the long term, maybe having a `packages/archive` subdirectory for old packages is clearner.

pzinn:
> Also, @pzinn, is there anything that the old package does that the new one doesn't?
I strongly suspect the answer is, no there isn't, but I haven't read the code of the old package in detail.
I've emailed one of its authors for their input, but no response.

DanGrayson:
> > Also, @pzinn, is there anything that the old package does that the new one doesn't?
> 
> I strongly suspect the answer is, no there isn't, but I haven't read the code of the old package in detail.
> I've emailed one of its authors for their input, but no response.

It's not upward compatible, though -- the new package exports different symbols.  Let's avoid breaking users' code and call the new package "VectorGraphics".

pzinn:
I'll close this PR and create a new one.

PR #1336, opened 2020-06-25
PencilOfQuadrics Package added

foschreyer:


DanGrayson:
Frank, your package has two examples that fail, like this:
```
i10 : C = ciModuleToCliffordModule M
stdio:10:5:(3): error: array index 1 out of bounds 0 .. 0
```

Yeongrak-Kim:
I tried uninstallPackage and installPackage several times, and also check "PencilsOfQuadrics" but I did not find any such an error. Dan, could you please point out where and what errors you have? Of course, I am working in version 1.15.

DanGrayson:
The problem is in this code:
```
fixDegs := d->( -- are the ceiling/floor functions applied correctly?
    d' := S'toS d;
    td := degrees target d;
    sd := degrees source d;
    td':= apply(td, D -> ceiling(3*D_0/2 -D_1));
    sd':= apply(sd, D -> floor(3*D_0/2 -D_1));
    map(S^td',S^sd',d')
    );
```
The value of `td` is `{{-1}, {-1}, {-1}, {-1}, {-3}, {-3}, {-3}, {-3}}`.  The first value of `D` is then `{-1}`.  And then `D_1` fails.
Were you hoping the degrees would be length 2 or greater?

Yeongrak-Kim:
> The problem is in this code:
> 
> ```
> fixDegs := d->( -- are the ceiling/floor functions applied correctly?
>     d' := S'toS d;
>     td := degrees target d;
>     sd := degrees source d;
>     td':= apply(td, D -> ceiling(3*D_0/2 -D_1));
>     sd':= apply(sd, D -> floor(3*D_0/2 -D_1));
>     map(S^td',S^sd',d')
>     );
> ```
> 
> The value of `td` is `{{-1}, {-1}, {-1}, {-1}, {-3}, {-3}, {-3}, {-3}}`. The first value of `D` is then `{-1}`. And then `D_1` fails.
> Were you hoping the degrees would be length 2 or greater?


The inputs d0, d1 for fixDegs are outputs of the method EisenbudShamashTotal (in the pacakge CompleteIntersectionResolutions). David updated this method and this package, and the correct degrees are of length 2, as you expected. However, the package CompleteIntersectionResolution contained in M2-1.15 does not have this update, and the old method EisenbudShamashTotal gives different results.

So I think David should come and update his package first. I am pretty sure our pakcage PencilsOfQuadrics will work then.

DanGrayson:
We have some updates recently, via Mike:
```
commit 35d0ba696095a4c921e7f41622891e284c7f1829
Author: Mike Stillman <mikestillman1@gmail.com>
Date:   Thu Jun 25 14:20:34 2020 -0400

    fix several bugs, including 2 from git issue #1208.

commit 64a33d5aadc692c4f3ae71705162ccb54da283da
Author: Mike Stillman <mikestillman1@gmail.com>
Date:   Wed Jun 24 21:29:29 2020 -0400

    modifications to 2 packages, provided by David EIsenbud.  Also a changes.m2 entry for the Complexes package, added earlier.
```

Maybe that should cover it?

I'll test ...

DanGrayson:
Yes, it seems to work now .

Issue #1257, opened 2020-06-10
Multiple Issues when Building from Scratch


radoslavraynov:
I am creating this ticket to track multiple issues I have had building Macaulay2 from source. Having pushed the GitHub actions PR #1256, one finally has [a way](https://github.com/radoslavraynov/M2/runs/755858392?check_suite_focus=true) to demonstrate those issues. Most but not all have to do with the `cmake` build. All are on `macOS`.  Thanks in advance!

cc @mahrud, @dangrayson

In no particular order,

* Some libraries are clearly present on my system — and are even reported as found, but are still locally built. Current examples are `googletest`, `mpfr`, `ntl`. Previously, the same thing was happening even for `mpir`.
* The `autotools` build uses `GMP` by default, while the `cmake` one uses `mpir`. This introduces an unpleasant inconsistency.
* `Flint`'s required version `2.5.3` looks off. Its [website](http://flintlib.org/downloads.html) lists versions `2.5.2` and `2.6.0`. If that's a hack to use a patched variant, another approach is needed.
* The target `build-programs` has never succeeded for me. It doesn't succeed on github actions either.
* (Omitting `build-programs` or letting it fail should be OK for a very capable M2 binary, so that's what I did.)
* I am unhappy with having to build `M2-core` *after* `M2-binary`, but that's a personal preference. The way I look at it, `M2-core` should refer to a core part and `M2-binary` to the end product.
* I hate the religious adoption of Emacs, and accordingly hate that `M2-emacs` is the default target. In my experience, one should make a software system as modular as possible or useful, (but not more), and systems where logically independent components are too entangled together are never well designed. My view is that Emacs is a front-end and should be kept as one.
* After building `M2-core` (and `M2-binary`), my interactive shell is missing history. For example, ⬆️ (up-key) doesn't bring back anything. I am guessing `readline` is not linked properly. In fact, `otool -L M2-binary` does not list `readline`.
* The default `gmake` step in the `autotools` builds a whole lot. Is there a target that only builds the binary without performing any package installations and tests?

Thanks! I am pasting my `cmake` configuration output for reference.

```
radoslav@home ~/Projects/M2/M2/BUILD/gcc9rel (up-master)$ CXX=g++-9 CC=gcc-9 cmake -S../.. -B. -DCMAKE_BUILD_TYPE=Release
-- The C compiler identification is GNU 9.3.0
-- The CXX compiler identification is GNU 9.3.0
-- Checking whether C compiler has -isysroot
[...]

PR #1337, opened 2020-06-25
Remove build paths from documentation (take 2)

d-torrance:
My first attempt at making Macaulay2's builds more reproducible by removing instances of the build paths from the documentation (#1247) had several problems.  In particular:

* It relied on using relative instead of absolute paths.  But for this to work, we needed to be building the examples from the build directory instead of `/tmp`, or the build path would still appear in the relative path.  However, the commit that made this change (afda220) was causing problems (https://github.com/Macaulay2/M2/issues/1257#issuecomment-643654221) and was reverted in 1a1a009.
* Some examples required the use of absolute paths (e.g., `toAbsolutePath`), and so canned examples were used, which is not ideal.
* A fictional `HOME` environment variable was set for the benefit of examples like `homeDirectory`.  However, this was also causing problems and was reverted in 204983c.

Instead, we take a completely different approach.  Instead of changing the behavior of Macaulay2 so that the examples it produces are reproducible, we process the examples after they are generated with the help of the new `regexQuote` function introduced in #1278.

The only change to the example-generating process itself now is that we use a `printWidth` of 0 (i.e., no wrapping) instead of 77.  Wrapping then takes place after the examples are generated.  The original goal was to make it easier to replace build paths that have been split up between lines, but it also has a nice side effect in that *all* examples are wrapped, instead of just those that previously actually obeyed `printWidth`.  For example, consider the documentation for `HH^ZZ CoherentSheaf`.

In 1.15:

wrap1 15

After this PR:

wrapnew

Closes: #1149

### DIFF
--- a/M2/Macaulay2/packages/Macaulay2Doc/doc.m2
+++ b/M2/Macaulay2/packages/Macaulay2Doc/doc.m2
@@ -274,7 +274,8 @@ document {
 	  prepended to the path, based on the value of the ", TO "applicationDirectory", "
 	  for your system."
 	  },
-     PRE concatenate between_"\n" apply(value Core#"private dictionary"#"userpath",s -> (5,s)),
+     PRE replace(regexQuote homeDirectory, "/home/m2user/",
+	   concatenate between_"\n" apply(value Core#"private dictionary"#"userpath",s -> (5,s))),
      EXAMPLE {
 	  "path",
 	  ///path = append(path, "~/resolutions/")///
--- a/M2/Macaulay2/d/actors2.dd
+++ b/M2/Macaulay2/d/actors2.dd
@@ -27,16 +27,14 @@ setupfun("reorganize",dbmreorganize);
 dbmopenin(e:Expr):Expr := (
      when e
      is a:stringCell do (
-	  if notify then stderr << "--opening database " <<
-	       relativizeFilename(a.v) << endl;
+	  if notify then stderr << "--opening database " << a.v << endl;
 	  dbmopenin(a.v))
      else buildErrorPacket("expected a string as filename"));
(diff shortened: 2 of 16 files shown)


M2/Macaulay2/packages/Macaulay2Doc/doc.m2
```diff
@@ -274,7 +274,8 @@ document {
 	  prepended to the path, based on the value of the ", TO "applicationDirectory", "
 	  for your system."
 	  },
-     PRE concatenate between_"\n" apply(value Core#"private dictionary"#"userpath",s -> (5,s)),
+     PRE replace(regexQuote homeDirectory, "/home/m2user/",
```
mahrud, line 277: Why not use `~` instead here?

d-torrance: It seemed more realistic since running the actual code prints the full path:

```m2
i12 : concatenate between_"\n" apply(value Core#"private dictionary"#"userpath",s -> (5,s))

o12 =      /home/profzoom/.Macaulay2/code/
           /home/profzoom/.Macaulay2/local/share/Macaulay2/
```

d-torrance:
Just found a bug.  https://github.com/Macaulay2/M2/pull/1337/commits/92952d6389867aaebbef2b6cb2e7b7fb23d2cc2e breaks `installPackage` if it's run after Macaulay2 itself has been installed because `topSrcdir` is `null`.  I'll add a new commit fixing this soon.

DanGrayson:
In connection with reproducibility, see 4c5a5d2e682d8986bd870211d9ba1f0f33ae8e06

PR #1338, opened 2020-06-25
Added package ThreadedGB

Sondzus:


Sondzus:
Is it easy to see what actually fails in the build - is it related to threads? I scrolled through the huge printout...and missed something. What should I do at this point? Thanks.

DanGrayson:
Many of the builds fail for reasons unrelated to your package -- I merged it, so I must have been satisfied.  All is well.

PR #1329, opened 2020-06-23
PHCpack fixes

antonleykin:
Should close
https://github.com/Macaulay2/M2/issues/1308
https://github.com/Macaulay2/M2/issues/1315

PHCpack version check implemented; derandomized only if --no-randomize; fixed the line that was breaking either numericalIrreducibleDecomposition or cascade

Issue #1010, opened 2019-09-18
bertiniPosDimSolve returns wrong slice


haerski:
`bertiniPosDimSolve` returns a witness set whose slice doesn't contain the points
```restart
needsPackage "Bertini"

R = CC[x,y,z,t]
I = ideal(x + 3, y+1)

nv = bertiniPosDimSolve(I_*, Verbose => true)

w = first components nv
F = polySystem slice w
pts = w#Points
pts / (p -> norm evaluate(F,p)) -- this value is >> 0
```

This seems to work only when codim I < 2

```
restart
needsPackage "Bertini"

R = CC[x,y,z,t]
I = ideal(x + 3, y+1, z - 2)

nv = bertiniPosDimSolve(I_*, Verbose => true)

w = first components nv
F = polySystem slice w
pts = w#Points
pts / (p -> norm evaluate(F,p)) -- this value is almost 0
```

@JoseMath

DanGrayson:
@antonleykin @JoseMath -- who will look at this?

antonleykin:
This issue got resolved in version 1.16.

PR #1339, opened 2020-06-25
NumericalAG update

antonleykin:
This updates `Bertini` package,  depends on `PHCpack` pull request https://github.com/Macaulay2/M2/pull/1329, and... may need some tweaks in `NumericalAlgebraicGeometry` optional software wrappers once these are settled.


antonleykin:
@JoseMath and @janverschelde may want to keep track of this pull request.

JoseMath:
bertiniPosDimSolve now reads in the slices from a witness_data  file correctly. 
The functionality of the IsProjective option has been fixed in bertiniZeroDimSolve and bertiniPosDimSolve

antonleykin:
> bertiniPosDimSolve now reads in the slices from a witness_data file correctly.
> The functionality of the IsProjective option has been fixed in bertiniZeroDimSolve and bertiniPosDimSolve

The first one addresses https://github.com/Macaulay2/M2/issues/1010

M2/Macaulay2/packages/Python.m2
```diff
@@ -26,10 +26,13 @@ rs = s -> (
      if debugLevel > 0 then stderr << "--python command: " << s << endl; 
      runString s);
 
-numContexts := 0
-nextContext := () -> (
-     numContexts = numContexts + 1;
-     "context" | toString numContexts)
+numContexts = 0
```
DanGrayson, line 29: What's this all about?  This package is for using when python is compiled in with M2, which it never is any more.

DanGrayson:
Anton, the diffs on this pull request are confusing (contradictory), and github thinks there are conflicts.  Could you merge your branch "NAG" with the latest from our "development" branch and push?

M2/Macaulay2/packages/Python.m2
antonleykin: This is a small change that was done when working with the Sage interface group. 
We do want M2 to have python interface. Don't we have `--with-python` anymore?

DanGrayson: By default, we build M2 without python.  I haven't tried it since 2009.  But if you want it for something, that's okay!

DanGrayson:
Thanks, that looks much better!

PR #1340, opened 2020-06-26
Two fixes for finding gftables in the autotools build

d-torrance:
The first fixes a bug that I introduced in 400a7bc and is responsible for the failing "Run Tests" target in the autotools workflows.

The second uses `pkg-config` to find the location of gftables instead of assuming they're under the same prefix we're planning on installing Macaulay2 under.


DanGrayson:
Thanks, it works!

PR #1351, opened 2020-06-28
Add quotes around reference to "break" keyword in docs for scan function

d-torrance:
Without the quotes, "break" is completely missing from the documentation,
likely due to the following:

    i22 : TO break
    stdio:23:4:(3): error: unhandled break command

See https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.15/share/doc/Macaulay2/Macaulay2Doc/html/_scan.html

By adding quotes, it works as intended.  This is already done in the
documentation for "lists and sequences".
See https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.15/share/doc/Macaulay2/Macaulay2Doc/html/_lists_spand_spsequences.html


DanGrayson:
Good catch!

Issue #407, opened 2016-04-12
Wishlist: Use external programs if present


d-torrance:
Currently, packages which use external programs optionally built by M2 use

```
prefixDirectory | currentLayout#"programs"
```

as their path.  Some packages allow the user to specify a different path using init scripts.

It would be great if M2 could detect which of these programs already exist on the system and store that information somewhere for use by these packages.

Note that this is further complicated in Debian by the fact that some of the programs used by M2 packages don't have standard names, e.g., the 4ti2 binaries are prepended with "4ti2-" and the nauty binaries with "nauty-".

DanGrayson:
In addition, the packages containing the files so detected have to be added to the list of dependencies of the Macaulay2 package.

DanGrayson:
Consider `4ti2`, for example.  The script `M2/configure.ac` should check for the presence of
`4ti2` in the system on PATH and not build it.  Then it should inform the package FourTiTwo somehow, which currently always looks for it in ` prefixDirectory | currentLayout#"programs" `.

One idea would be to have FourTiTwo look on the PATH for 4ti2 by default, and to have the directory  ` prefixDirectory | currentLayout#"programs" ` added to the front of the PATH.  That's probably okay -- we won't be installing programs in there with names that conflict with system programs.

When Macaulay2 is installed by the package manager, we need to list the package containing the program 4ti2 as a prerequisite.  The construction of lists of prerequisites is done differently by each entry in `M2/distributions`, since finding the name of that package depends on the type of OS distribution we are running under.

The packages could look for the programs under all popular names, e.g., `dreadnaut` and `nauty-dreadnaut`.

d-torrance:
I've started working on a solution to this issue in the branch https://github.com/d-torrance/M2/tree/programs.

The first few commits basically involve moving and generalizing code from `gfanInterface` to `Core`, but there's still lots of more work to do.  In particular:

* [ ] documentation
* [ ] unit testing
* [ ] `Verbose` option for both `loadProgram` and `runProgram`.  We currently use `notify` for this functionality in `loadProgram`, but I think a separate option would be best for flexibility.  For example, we can set `Verbose => gfanVerbose` in `gfanInterface` so it supports existing code.
* [ ] Support for prefixes, e.g., `topcom-`, `nauty-`, and `4ti2-`, that are prepended to the binary names in some distributions.  This will be especially tricky for topcom, as Gentoo and Fedora only use prefixes for *some* of the binaries.
[...]

PR #1368, opened 2020-07-03
Improving support for external programs, part 1

d-torrance:
This is the first in a planned series of pull requests with the ultimate goal of solving #407.  In it, we introduce two new methods to `Core`: `loadProgram` and `runProgram`, based on code from the `gfanInterface` package, that find and run external programs.  The idea is that eventually all packages that call an external program (and especially, any of the programs which are built by Macaulay2) would use these methods.

This has the advantage of:
* simplifying things for future package authors as they would only have to deal with this higher-level interface
* reducing build time, as we wouldn't need to worry about things like this, i.e., building programs even when they're already present because the packages can't find them: https://github.com/Macaulay2/M2/blob/cc6fe6b580840f7da17dcad9c2915b2254b190c8/M2/configure.ac#L944-L947

* reducing the diff between the upstream Macaulay2 source and the Debian package, as Debian packages of these programs would be supported out of the box with no modification.

As a proof of concept, `gfanInterface` has been ported to the new interface.

Future pull requests will likely include:
* support for dealing with prefixes which may be present for programs that ship with multiple executables, e.g., `4ti2-` in the Debian 4ti2 packages
* support for detecting minimum versions of programs
* porting more existing packages to the new interface

Here's a quick example of these methods in action:

```m2
i4 : programPaths#"gfan" = "/foo/bar"

o4 = /foo/bar

i5 : gfan = loadProgram("gfan", "gfan --help", Verbose => true)
checking for gfan in /foo/bar/...
        not found
checking for gfan in /usr/libexec/Macaulay2/bin/...
        found

o5 = Program{name => gfan                       }
             path => /usr/libexec/Macaulay2/bin/

o5 : Program

i6 : runProgram(gfan, "_version")

o6 = 0

o6 : ProgramRun

i7 : oo#"output"

o7 = Gfan version:
     gfan0.6.2

     Forked from source tree on:
     1506606121 Thu Sep 28 15:42:01 2017

     Linked libraries:
     GMP 6.0.0
     Cddlib       YES
     SoPlex        NO
     Singular      NO
```


mahrud:
Can you instead look for a universally installed program instead of gfan and topcom in the examples for findPrograms and runProgram? Currently none of my builds succeed without rebuilding gfan and topcom, which I don't want to do on every single build.

How about looking for python, or perhaps arranging so the examples don't fail when the program isn't found.

DanGrayson:
This build succeeded and contains this pull request: https://github.com/Macaulay2/M2/actions/runs/175534262

Issue #1153, opened 2020-05-14
feature request: addParallelHook and runParallelHooks


mahrud:
Two requests:
1. add `listHooks` to get a list/location of existing hooks on a methods.
2. possibly implement parallel hooks. For instance, I have two hooks, each have a quick test that checks whether the hook is appropriate. I want both hooks to start working in parallel and once one is finished the result is returned and the other is cancelled.

mahrud:
From Mike:

> Hooks are a useful way to choose algorithms or specific functions to run.
> 
> There are some improvements that should be made though:
> 
> * Improve the documentation.  There should be a doc node that explains how to do this
> * Need a function to get the list of hook functions, and be able to locate their code.
> * Perhaps change `(break result)` to just return something non-null instead.
> * How to find which functions/methods allow hooks.

mahrud:
Status update: the remaining task here is to allow running hooks in parallel.

mahrud:
A general question: what was the idea behind the way hooks are implemented? For instance:
- in (addHook, HashTable, Thing, Thing), is the HashTable supposed to be the type of the input or the output?
- why does the method above save the hooks under the HashTable?

Context: say we're implementing hooks for the quotient and saturate functions, however:
- the meaning of quotient depends on the type of the inputs, which makes the usage `addHook(symbol quotient, algorithm)` undesirable, because that would go through a bunch of unrelated hooks before getting to the ones we really want to try.
- even if we wanted to use the above usage, it wouldn't work:
```m2
i15 : addHook(symbol quotient, ((opts, a, b) -> a*b))
stdio:15:1:(3): error: assignment to protected variable 'quotient'
```
- these functions typically take two inputs, e.g. (quotient, Ideal, Ideal) and (quotient, Module, Ideal), which have exclusive algorithms. This makes installing hooks on a type like `addHook(Module, symbol quotient, func)` not good, because, for instance, (quotient, Module, Module) and (quotient, Module, Ideal) don't even return the same type. We could consider storing the hooks under the return type, but again there are many unrelated meanings of "quotient" that return an ideal, so that doesn't make sense either.

A couple of new items on the wish list:
- [ ] reimplement hooks to dispatch similar to methods, i.e. based on the input types:
```m2
addHook(symbol quotient, (Ideal, Ideal), function)
```
This could, for instance, add `function` to a list under a mutable global dictionary or hash table:
```m2
[...]

PR #1369, opened 2020-07-04
Improvements to hooks

haerski:
Partially addresses #1153

- Added a method `hooks(...)` to list hooks. This returns a `NumberedVerticalList`, and code for each hook can be examined using `code ZZ`
- Hooks can now have optional arguments. The function `runHooks` passes any relevant optional arguments to each hook.
- Hooks no longer need to return their value via `break`, any non-null value will do.
- Improved documentation `runHooks`
- Added a documentation node "using hooks" under the node "The Macaulay2 language"

@pepperhdj


DanGrayson:
Thanks, this looks good!  (It's too late for 1.16, so it will appear in 1.17.)

Issue #1370, opened 2020-07-04
segmentation fault in nets.d: VerticalJoin


mahrud:
Finally managed to get boost to pretty print stacktraces despite ASLR and get some output form the crashes in Ubuntu. Here's the line from the [stacktrace](https://github.com/Macaulay2/M2/pull/1367/checks?check_run_id=836015030#step:9:14044):
```
FAILED: usr-dist/common/share/Macaulay2/Core/tvalues.m2 
cd /home/runner/work/M2/M2/M2/BUILD/cicd/usr-dist/common/share/Macaulay2/Core && /home/runner/work/M2/M2/M2/BUILD/cicd/usr-dist/x86_64-Linux-Ubuntu-18.04/bin/M2-binary -q --silent --stop -e errorDepth=0 --no-preload --no-tvalues /home/runner/work/M2/M2/M2/Macaulay2/m2/tvalues-make.m2 -e "make \"/home/runner/work/M2/M2/M2/Macaulay2/d/\"; exit 0"
-- SIGSEGV
-* stack trace, pid: 119184
 0# stack_trace(std::ostream&, bool) at ../../Macaulay2/bin/main.cpp:124
 1# segv_handler at ../../Macaulay2/bin/main.cpp:241
 2# 0x00007FAFE6274F20 in /lib/x86_64-linux-gnu/libc.so.6
 3# nets_VerticalJoin at /home/runner/work/M2/M2/M2/Macaulay2/d/nets.d:132
 4# evaluate_evalraw at /home/runner/work/M2/M2/M2/Macaulay2/d/evaluate.d:1293
...
```

@DanGrayson Any ideas why this might be happening?

DanGrayson:
Line 132 in nets.d is 
```
	  leng = leng + length(n.body);
```
, which translates to the following C code:
```C
  leng_1 = (leng_1 + tmp__79->array[tmp__80]->body->len);
```
There is no function call to a function in libc on that line, so 
maybe line 2 of the stack trace is an interrupt handler routine, too.  In that case, one of the three memory
accesses must be out of bounds.  If so, the most likely explanation for it is that we have a systematic
screw-up in the handling of libgc memory, and some corruption has occurred.  In that case, a lengthy
session with a debugger is called for.  Such a screw-up is more likely, since the eigen branch was merged
not so long ago. An example of a screw-up would be storing a pointer to libgc memory in malloc memory, and
then using it after the garbage collector has collected it.  That could be anywhere else in the code, for after
[...]

Issue #1357, opened 2020-06-30
Add GCC to macOS-based Github Actions test builds


DanGrayson:
... so maybe I'll eliminate 2 of our 8 builds from the workflow, now that the other 6 are working:

https://github.com/Macaulay2/M2/actions/runs/152706463

@radoslavraynov 
@mahrud

mahrud:
Why not? I see a ThreadedGB test is failing, so we should probably look into it.

DanGrayson:
I wonder why ThreadedGB is timing out for you.  I just tried "tgb" code manually here and it took 27 seconds, and the limit is 27 seconds.

Here's the error on the autotools build:
```
2020-06-30T12:22:51.1928230Z /bin/sh ../../libtool  --tag=CXX   --mode=link g++-9  -I./..   -D__STDC_LIMIT_MACROS -DNDEBUG -std=gnu++14 -g3 -O2  -w -Wall -std=c++0x -L../4ti2 -R/Users/runner/runners/2.169.1/work/M2/M2/M2/BUILD/cicd/usr-host/lib -l4ti2common -no-undefined -L/Users/runner/runners/2.169.1/work/M2/M2/M2/BUILD/cicd/usr-host/lib  -L/Users/runner/runners/2.169.1/work/M2/M2/M2/BUILD/cicd/submodules/memtailor/.libs  -L/Users/runner/runners/2.169.1/work/M2/M2/M2/BUILD/cicd/submodules/mathic/.libs  -L/Users/runner/runners/2.169.1/work/M2/M2/M2/BUILD/cicd/submodules/mathicgb/.libs -L/usr/lib -L/usr/local/lib  -g3 -headerpad_max_install_names -o 4ti2int32 4ti2int32-circuits_main.o 4ti2int32-groebner_main.o 4ti2int32-markov_main.o 4ti2int32-minimize_main.o 4ti2int32-normalform_main.o 4ti2int32-qsolve_main.o 4ti2int32-rays_main.o 4ti2int32-walk_main.o 4ti2int32-zbasis_main.o 4ti2int32-main.o lib4ti2int32.la -lglpk -lgmp -lgmpxx -lgmp
[...]

PR #1371, opened 2020-07-04
CMake maintenance

mahrud:


mahrud:
@DanGrayson okay, this is ready now.

DanGrayson:
Ready for pulling, or ready to see if the tests work out?

mahrud:
Well, it fixes the cddgmp bug that you emailed me about, so it can be merged on its own. I thought I fixed that in the first commit of the last PR but somehow a commit was overwritten I think.

There are still two issues, #1370 and #1357 that break two builds, but I can fix those later.

Issue #1173, opened 2020-05-19
Failing check in StatePolytope


d-torrance:
```
i2 : check_0 StatePolytope
--loading configuration for package "StatePolytope" from file /home/profzoom/.Macaulay2/init-StatePolytope.m2
--running test 0 of package StatePolytope on line 212 in file ./StatePolytope.m2
--    rerun with: check_0 "StatePolytope"
--making test results in file /tmp/M2-3023-0/0.out
ulimit -c unlimited; ulimit -t 700; ulimit -m 850000; ulimit -v 850000; ulimit -s 8192; ulimit -n 512; cd /tmp/M2-3023-0/1-rundir/; GC_MAXIMUM_HEAP_SIZE=400M /usr/bin/M2-binary --silent --print-width 77 --stop --int --no-readline --no-randomize -e 'needsPackage("StatePolytope", Reload => true, FileName => "/home/profzoom/src/macaulay2/M2/M2/Macaulay2/packages/StatePolytope.m2")' <"/tmp/M2-3023-0/0.m2" >>"/tmp/M2-3023-0/0.tmp" 2>&1
/tmp/M2-3023-0/0.tmp:0:1: (output file) error: Macaulay2 exited with status code 1
/tmp/M2-3023-0/0.m2:0:1: (input file)
M2: *** Error 1
stdio:2:6:(3): error: error occurred running test for package StatePolytope: 0
```

The check is supposed to assert that the following two are the same, but they definitely aren't!
```
i3 : set initialIdeals ideal(a*c-b^2,a*d-b*c,b*d-c^2)
----------------------------------------------------------------------------
This is the Gfan program for computing Groebner fans and tropical varieties.
     Webpage: http://home.math.au.dk/jensen/software/gfan/gfan.html
----------------------------------------------------------------------------

To run a Gfan subcommand do for example:
	gfan _groebnerfan
in your shell and type in your input, for example:
	Q[x,y]{x^2-y-1,y^2-xy-2/3}

Options for controlling the main program:
 _list             List all subcommands.
 _version          Display version of gfan and compiled libraries.
 _test             Run selftest. Use --suite to specify test suite.
                   Creates/overwrites files in current folder.
 _documentation    Generate .tex file for manual.

General useful options for subcommands:
 --help            Print the help text associated to the subcommand.
 --log1            Set the log level for comments written to stderr while computing (--log2,--log3).
 --stdin FILENAME  Use the specified file as input for the program rather than reading from stdin.
[...]

Issue #507, opened 2016-08-02
statePolytope vs StatePolytope.m2


lkastner:
In Polyhedra.m2 there is a very long method named statePolytope that computes the state polytope of an ideal. There is also the package StatePolytope.m2. We should probably get rid of one of these two, if they really do the same. Dan, could you please add Dave Swinarski to this issue to discuss this?

DanGrayson:
Here's the way you get someone involved in a discussion: @dswinarski

lkastner:
Ah, thank you. I didn't know that.

PR #1373, opened 2020-07-06
StatePolytope updates

d-torrance:
This is a short followup to #1368, specifically dealing with the `StatePolytope` package.

It calls gfan for the the `initialIdeals` method, which is used by its other methods (which call polymake).  We update it to use the `gfanInterface` package, and therefore, indirectly, the new `loadProgram` and `runProgram` methods so it can use any gfan, either installed by Macaulay2 or already present on the system.

We also:
* Rename `statePolytope` to `polymakeStatePolytope` to avoid an error (see also #507).
* Fix the failing test from #1173.
* Update the cached examples.


DanGrayson:
@d-torrance -- how could changing the name of a function prevent an error?

mahrud:
Perhaps in light of #507, this package should be removed from distributed packages?

mahrud:
Also, are we not concerned about involving package authors in changes to their packages anymore?

DanGrayson:
> Also, are we not concerned about involving package authors in changes to their packages anymore?

We are.  That's why I've (already) asked the author to review this.

DanGrayson:
> Perhaps in light of #507, this package should be removed from distributed packages?

That would premature, since the question in #507 has not been answered.

dswinarski:
Thanks everyone for your comments on this--I'll review it this week.  I
predict that it may indeed be time to remove this package from the
distributed packages, but I would like to test a few examples using the
Polyhedra package first to see if there is any major difference in
performance anymore.  Many years ago, there would have been, but I suspect
that gap may have closed with recent work on Polyhedra.

Best,

Dave
*_______________________________________*

*David Swinarski, Ph.D.*
Associate Dean, Faculty of Arts and Sciences
Fordham University
718.817.3034

dswinarski@fordham.edu

(he/him/his)
*_______________________________________*

d-torrance:
> @d-torrance -- how could changing the name of a function prevent an error?

`Polyhedra` also has a `statePolytope` method, and it's exported by `gfanInterface`.  So if we try to define a new `statePolytope` method, we get an error:

    error: assignment to protected built-in global variable 'statePolytope'

DanGrayson:
> > @d-torrance -- how could changing the name of a function prevent an error?
> 
> `Polyhedra` also has a `statePolytope` method, and it's exported by `gfanInterface`. So if we try to define a new `statePolytope` method, we get an error:
> 
> ```
> error: assignment to protected built-in global variable 'statePolytope'
> ```

Why not just not try to define a new `statePolytope` function?  (That's the error here.)

d-torrance:
> Why not just not try to define a new `statePolytope` function? (That's the error here.)

That was my goal.  By renaming `statePolytope` to something else, then we wouldn't be trying to define it a second time.

DanGrayson:
I don't get any errors when I build Macaulay2, so we aren't trying to define it a second time.

d-torrance:
> I don't get any errors when I build Macaulay2, so we aren't trying to define it a second time.

Not currently, no.  But this PR adds the following line:
https://github.com/Macaulay2/M2/blob/7785cd656d011450a49f8df49e14c42792de9cbe/M2/Macaulay2/packages/StatePolytope.m2#L16

If I try to load the package with that line added, I get:
```m2
i1 : loadPackage("StatePolytope",
         FileName => "~/src/macaulay2/M2/M2/Macaulay2/packages/StatePolytope.m2")
../../StatePolytope.m2:126:15:(2):[7]: error: assignment to protected built-in global variable 'statePolytope'
```

DanGrayson:
Ah, finally I catch on, thanks!  If Dave likes the name the way it is, then inside the package you could refer to the shadowed symbol by its synonym, `StatePolytope$statePolytope`.

d-torrance:
I just noticed a further complication.  This package requires polymake < 4.0, as the command line interface changed with the version bump:
https://polymake.org/doku.php/user_guide/tutorials/legacy#command_line

I tested this pull request initially in Ubuntu 18.04, which has polymake 3.2, and everything worked fine.  But in Debian unstable, which has polymake 4.0, it fails:

```m2
i8 : R=QQ[a..d];

i9 : I=ideal(a*c-b^2,a*d-b*c,b*d-c^2);

o9 : Ideal of R

i10 : L = {{9, 6, 6, 9}, {9, 3, 12, 6}, {7, 5, 14, 4}, {5, 8, 14, 3}, {3, 12, 12,3}, {6, 12, 3, 9}, {4, 14, 5, 7}, {3, 14, 8, 5}};

i11 : statePolytope(3, I)
----------------------------------------------------------------------------
This is the Gfan program for computing Groebner fans and tropical varieties.
     Webpage: http://home.math.au.dk/jensen/software/gfan/gfan.html
----------------------------------------------------------------------------

To run a Gfan subcommand do for example:
	gfan _groebnerfan
in your shell and type in your input, for example:
	Q[x,y]{x^2-y-1,y^2-xy-2/3}

Options for controlling the main program:
 _list             List all subcommands.
 _version          Display version of gfan and compiled libraries.
 _test             Run selftest. Use --suite to specify test suite.
                   Creates/overwrites files in current folder.
 _documentation    Generate .tex file for manual.

General useful options for subcommands:
 --help            Print the help text associated to the subcommand.
 --log1            Set the log level for comments written to stderr while computing (--log2,--log3).
 --stdin FILENAME  Use the specified file as input for the program rather than reading from stdin.
 --stdout FILENAME Use the specified file as output for the program rather than writing to stdout.

The command "gfan" is deprecate for computing all Groebner bases of an ideal.
Please use subcommand "gfan _bases" instead.
To keep compatibility with older versions, gfan _bases is awaiting input.
<Ctrl>-D to end.
usage: polymake [-dv] [-A|-a <application>]
                [--reconfigure] [--config-path PATH ... | --no-config | --ignore-config]
                [ --script | --iscript <script_file> arg ... ] | script_file | 'one-liner' |
                - | [ --connect SOCKETFILE | HOST:PORT ] |
                --touch <file> ... | --help | --version
stdio:11:1:(3): error: close: process exited with code 256
```

DanGrayson:
The interaction you posted seems to be with gfan, not polymake.

If we want to distribute M2 on both systems (once Debian unstable is incorporated into a release), we might have to arrange for some packages to run polymake both ways.  That'll be a pain.

How many packages are running polymake themselves?

d-torrance:
> The interaction you posted seems to be with gfan, not polymake.

It interacts with both.  First it calls gfan to get the initial ideals, and then it calls polymake to find the vertices of a polytope.   gfan was just a lot more verbose in the example above.  But polymake is there -- check out the last few lines.

> If we want to distribute M2 on both systems (once Debian unstable is incorporated into a release), we might have to arrange for some packages to run polymake both ways. That'll be a pain.

Fortunately, the new command-line interface was introduced in 3.x (or maybe even earlier?).  They carried compatibility with both for a while and dropped the old one in 4.0.

I've added a new commit porting `StatePolytope` to the new interface. It also uses `loadProgram` and `runProgram` now, too.

> How many packages are running polymake themselves?

Certainly `Polymake`, but glancing at it, it also appears to use the old interface.  I imagine a better solution in the future would be to port it to the new interface and have `StatePolytope` import it instead of calling polymake directly.

d-torrance:
I also wrote a patch for the Debian package that updates `StatePolytope` to use modern gfan and polymake in a more minimal way, i.e., no importing `gfanInterface` or calling `loadProgram`/`runProgram`:
https://salsa.debian.org/science-team/macaulay2/-/blob/master/debian/patches/fix-state-polytope.patch

d-torrance:
Rebased this on top of `development` since it's affected by the `loadProgram` → `findProgram` change from #1389.

d-torrance:
Just added a new commit adding the `--no-config` argument to polymake.  Without it, polymake tries to create `~/.polymake`, which we may not have permission to do.

DanGrayson:
@dswinarski -- did this fall off your radar?

DanGrayson:
@dswinarski -- did this fall off your radar?

d-torrance:
I've rebased this branch on top of `development` and added a new commit adding support for the `OptionalComponentsPresent` option.

DanGrayson:
@dswinarski -- did this fall off your radar?

dswinarski:
Yes, sorry. I'll invest some time in it this weekend.

Best,

Dave
*_______________________________________*

*David Swinarski, Ph.D.*
Associate Dean, Faculty of Arts and Sciences
Fordham University
718.817.3034

dswinarski@fordham.edu

(he/him/his)
*_______________________________________*

DanGrayson:
@dswinarski  --  why don't I go ahead and pull this so it gets into 1.17?  You can always look at it later.

Or take a look during the next week or so.

PR #1378, opened 2020-07-07
Improving support for external programs, part 2

### LINKED ISSUE #1121: Topcom package should raise an error when it can't find TOPCOM
Currently, there's no indication that the call to the TOPCOM program failed.  Here's an example:

    Macaulay2, version 1.15
                   Truncations
    
    i1 : loadPackage("Topcom", Reload=>true)
    
    o1 = Topcom
    
    o1 : Package
    
    i2 : debugLevel = 5
    
    o2 = 5
    
    i3 : A = transpose matrix {{0,3},{0,1},{-1,-1},{1,-1},{-4,-2},{4,-2}};
    
                  2        6
    o3 : Matrix ZZ  <--- ZZ
    
    i4 : tri = {{0,1,2}, {1,3,5}, {2,3,4}, {0,1,5}, {0,2,4}, {3,4,5}, {1,2,3}};
    
    i5 : isRegularTriangulation(A,tri)
    -- calling topcom
    -- checkregularity --checktriang -v: using temporary file prefix /tmp/M2-12003-0/0
    -- checkregularity --checktriang -v: executing /usr/libexec/Macaulay2/bin/checkregularity --checktriang -v </tmp/M2-12003-0/0.in 1>/tmp/M2-12003-0/0.out 2>/tmp/M2-12003-0/0.err
    -- checkregularity --checktriang -v: output = {{0,1,2},{1,3,5},{2,3,4},{0,1,5},{0,2,4},{3,4,5},{1,2,3}}
                                                  is non-regular.
    
    
    o5 = false
    
    i6 : topcompath = "foo"
    
    o6 = foo
    
    i7 : isRegularTriangulation(A,tri)
    -- calling topcom
    -- checkregularity --checktriang -v: using temporary file prefix /tmp/M2-12003-0/1
    -- checkregularity --checktriang -v: executing foocheckregularity --checktriang -v </tmp/M2-12003-0/1.in 1>/tmp/M2-12003-0/1.out 2>/tmp/M2-12003-0/1.err
    -- checkregularity --checktriang -v: output = 
    
    o7 = false

d-torrance:
Continuing #1368, this pull request primarily introduces the `Prefix` option to `loadProgram`, which checks various prefixes that distributions often add to programs that include lots of executables.  Then, behind the scenes, `runProgram` adds these prefixes where necessary.

The `Topcom` package is modified to use the new methods.  It was a good test case since Debian, Fedora, and Gentoo all use different prefix schemes!  Since `loadProgram` throws an error when it can't find a program, this closes #1121.

We also add some unit tests for the new methods.


M2/Macaulay2/packages/Macaulay2Doc/functions/loadProgram-doc.m2
```diff
@@ -40,27 +45,48 @@ document {
...
+	Prefix => List => {
+	    "a list of sequences containing three strings identifying a ",
+	    "prefix that is added to the executable binaries belonging to the ",
+	    "program by some distributions.  These sequences should be of the ",
+	    "form ", TT "(distribution, regex, prefix)", " where: ",
+		UL {
+		    {TT "distribution", " identifies the name of the ",
+			"distribution that uses the prefix, e.g., ",
+			"\"debian\".  It is only used for printing messages ",
+			"when ", TT "Verbose", " is set to ", TO "true", "."},
```
DanGrayson, line 75: "distribution" is vague.  What about defining it to be the initial alphanumeric part of `version#"issue"`?  We could even add an entry `version#"distribution"` based on that definition.

d-torrance: The idea is that the package author would set this when they write a package that interfaces with an external program.  They'd need to do the legwork to see what the various prefixes are in various distributions.  It shouldn't be automatically generated.  For example:

```M2
	topcomProgram = loadProgram("topcom","cube 3", Prefix => {
	    ("debian", ".*", "topcom-"),
	    ("fedora", "^(cross|cube|cyclic|hypersimplex|lattice)$", "TOPCOM-"),
	    ("gentoo", "^cube$", "topcom_")});
```

DanGrayson: But how does the author know whether to write "debian" or "Debian" for the distribution?  That's what's vague.

DanGrayson: Also vague is whether, under Ubuntu, to write "Ubuntu" or "Debian".

d-torrance: It doesn't really matter, as it's just there to add some readability to the verbose output.  `loadProgram` just tries every path/prefix combination by brute force until it finds one that works.  It doesn't actually try to detect what the system actually is.  For example:

```m2
i3 : loadProgram("topcom", "cube 3", Verbose => true, Prefix => {
         ("debian", ".*", "topcom-"),
         ("fedora", "^(cross|cube|cyclic|hypersimplex|lattice)$", "TOPCOM-"),
         ("gentoo", "^cube$", "topcom_")})
checking for topcom in /usr/libexec/Macaulay2/bin/...
         trying default prefix...
        not found
         trying debian prefix...
        not found
         trying fedora prefix...
        not found
         trying gentoo prefix...
        not found
checking for topcom in /home/profzoom/.local/bin/...
         trying default prefix...
        not found
         trying debian prefix...
        not found
         trying fedora prefix...
        not found
         trying gentoo prefix...
        not found
checking for topcom in /usr/local/sbin/...
         trying default prefix...
        not found
         trying debian prefix...
        not found
         trying fedora prefix...
        not found
         trying gentoo prefix...
        not found
checking for topcom in /usr/local/bin/...
         trying default prefix...
        not found
         trying debian prefix...
        not found
         trying fedora prefix...
        not found
         trying gentoo prefix...
        not found
checking for topcom in /usr/sbin/...
         trying default prefix...
        not found
         trying debian prefix...
        not found
         trying fedora prefix...
        not found
         trying gentoo prefix...
        not found
checking for topcom in /usr/bin/...
         trying default prefix...
        not found
         trying debian prefix...
        found
```

I could put a suggestion in the documentation as to what it should be, e.g., the "parent" distribution in lowercase or something, so that packages are consistent.

DanGrayson: Oh!  It would be better, instead of demanding what is essentially just a comment from the user, to just display the prefix itself that you're trying.  It will give complete information, after all.

d-torrance: Ah, gotcha.  I think my thought was that some distributions may end up using the same prefix, but not on the same binaries.  Say, Debian ships `foo-bar` and `foo-baz`, but Fedora ships `foo-bar` and `baz`.  So just the prefix may not be sufficient to disambiguate the different cases.

This may never happen in reality, though.  I'll double-check nauty and 4ti2 to see what the different distributions use as prefixes.  If it's not an issue, then I'll drop that first string.

DanGrayson: To handle the case you're worried about, when looking for `bar`, look for both `foo-bar` and `bar`.  Then also, when looking for `baz`, look for both `foo-baz` and `baz`.  That makes the regular expression redundant, too.

d-torrance:
Hrm, looks like the examples for `loadProgram` are failing in the workflows:

```
--making example results for loadProgram
 ulimit -c unlimited; ulimit -t 700; ulimit -m 850000; ulimit -s 8192; ulimit -n 512;  cd /tmp/M2-8022-0/507-rundir/; GC_MAXIMUM_HEAP_SIZE=400M "/home/runner/work/M2/M2/M2/BUILD/cicd/usr-dist/x86_64-Linux-Ubuntu-18.04/bin/M2-binary" -q --int --no-randomize --no-readline --silent --stop --print-width 77 <"/tmp/M2-8022-0/90_load__Program.m2" >>"/home/runner/work/M2/M2/M2/BUILD/cicd/usr-dist/common/share/doc/Macaulay2/Macaulay2Doc/example-output/_load__Program.errors" 2>&1
/home/runner/work/M2/M2/M2/BUILD/cicd/usr-dist/common/share/doc/Macaulay2/Macaulay2Doc/example-output/_load__Program.errors:0:1: (output file) error: Macaulay2 exited with status code 1
/tmp/M2-8022-0/90_load__Program.m2:0:1: (input file)
M2: *** Error 1
```
My guess is that it's not finding topcom in the second example:
```m2
i3 : loadProgram("topcom", "cube 3", Verbose => true, Prefix => {
         ("debian", ".*", "topcom-"),
         ("fedora", "^(cross|cube|cyclic|hypersimplex|lattice)$", "TOPCOM-"),
         ("gentoo", "^cube$", "topcom_")})
```
I'll try and diagnose what's going on.

M2/Macaulay2/packages/Macaulay2Doc/functions/loadProgram-doc.m2
d-torrance: If we remove the regular expression, then `runProgram` would have to worry about trying both things, too.  My goal was for `loadProgram` to deal with everything up front so that `runProgram` would have it easy.

d-torrance: It looks like the nauty and 4ti2 prefixes are all more sane than the topcom ones (either everything is prefixed with `nauty-`, `4ti2-`, or `4ti2_`, or nothing is), so it should be fine to drop the distribution string.

Whenever I figure out the failing workflow, I'll push a couple more commits.

d-torrance: Even if we do run into that case at some point, I suppose it's not the end of the word if the verbose output looks a little redundant, even though behind the scenes the checks are slightly different:

```m2
checking for foo in /usr/bin/...
         trying prefix ""...
        not found
         trying prefix "foo-"...
        not found
         trying prefix "foo-"...
        not found
```

d-torrance: Or even better, I'll just include the regex in the message, too.

DanGrayson: > If we remove the regular expression, then `runProgram` would have to worry about trying both things, too. My goal was for `loadProgram` to deal with everything up front so that `runProgram` would have it easy.

Not if `runProgram` records where each program was found, so `runProgram` knows where and under what name it was found.

d-torrance: > > If we remove the regular expression, then `runProgram` would have to worry about trying both things, too. My goal was for `loadProgram` to deal with everything up front so that `runProgram` would have it easy.
> 
> Not if `loadProgram` records where each program was found, so `runProgram` knows where and under what name it was found.

But `loadProgram` currently doesn't check each program -- only the minimal number of them needed to figure out the path and prefix.  Even for TOPCOM, we only need to call one thing -- the prefix of `cube` gives us enough information to figure out all of the prefixes, provided we know where to get the information, which is what the regexes are doing.

TOPCOM comes with 42 little programs, so I'd rather not check each one for each prefix/path pair.

d-torrance:
Just tried a build using Macaulay2-built TOPCOM instead of using the Debian package.  It looks like the TOPCOM files aren't present when building the documentation:

```
(sid-amd64)root@orange:~/M2/M2/usr-dist/x86_64-Linux-Debian-unknown/libexec/Macaulay2/bin# ls cube
ls: cannot access 'cube': No such file or directory
```

d-torrance:
Aha!  It's not that the TOPCOM files aren't installed, it's just that only some of them are installed.  In particular, we're missing `cube`, which is used in the example:
https://github.com/Macaulay2/M2/blob/6a08eb37987df136fb4db0109a8cc133ac30ffb8/M2/libraries/topcom/Makefile.in#L9-L12

Is there a reason that we're only installing these?  Apparently, these are the only ones the current packages use since none of the other tests are failing.

If we're 100% certain we'll never use any other TOPCOM programs, then I could certainly make the prefix thing much simpler, as none of the current ones have prefixes on Fedora or Gentoo.

DanGrayson:
I'm sure that list of programs was the complete list at some point in the past.  Now it seems the complete list would be this:
```
B_A B_A_center B_D chiro2allfinetriangs chiro2alltriangs chiro2circuits
chiro2cocircuits chiro2dual chiro2finetriang chiro2finetriangs chiro2mintriang
chiro2nallfinetriangs chiro2nalltriangs chiro2nfinetriangs chiro2ntriangs
chiro2placingtriang chiro2triangs cocircuits2facets cross cube cyclic
hypersimplex lattice points2allfinetriangs points2alltriangs points2chiro
points2facets points2finetriang points2finetriangs points2flips
points2nallfinetriangs points2nalltriangs points2nfinetriangs points2nflips
points2ntriangs points2placingtriang points2triangs points2volume
santos_22_triang santos_dim4_triang santos_triang
```

Feel free to install them all.

But one thing I don't understand: how can you be getting an error in an example when I'm not?

d-torrance:
> But one thing I don't understand: how can you be getting an error in an example when I'm not?

The new example for the `loadPackage` documentation added in this PR (fec655b) assumes that TOPCOM's `cube` is available.  (It's the best choice for identifying the prefix, in particular because it's the only one that Gentoo even gives a prefix to.)  I didn't have any problems before I submitted the PR because I was using the Debian package, which includes `cube`.

I didn't see the error until I got an email about all 8 workflows failing!  I finally figured out that `cube` wasn't being installed into `pre_programsdir`.

mikestillman:
I would be happy to have more (or all) of the topcom programs installed with M2.  It is true that we currently need only a few, but I do plan on using more of them (unless polymake becomes available to us).  However, some are very basis, doing things that can be easily done in M2 (probably `cube` is one of these).

d-torrance:
Just added two new commits, dropping the "distribution" string from the prefix option and installing the rest of the TOPCOM.

d-torrance:
It looks like the `loadProgram` documentation has some invalid html:

```
 *** invalid HTML: /home/runner/work/M2/M2/M2/BUILD/cicd/usr-dist/common/share/doc/Macaulay2/Macaulay2Doc/html/_load__Program.html
error: line 50: document type does not allow element "ul" here; missing one of "button", "map", "object", "ins", "del", "noscript" start-tag
8499 HTML files checked; 1 invalid
```
I'll take a look at it.

DanGrayson:
`B_A` is an ugly name for a program.  Eh.

Issue #1227, opened 2020-06-03
Fedora build -- unique pointers


DanGrayson:
on Fedora with gcc version 10.1.1 on the release-1.15 branch:
```
make: Entering directory '/home/dan/src/M2/M2/BUILD/dan/builds.tmp/fedora-release-1.15/Macaulay2/e'
CXX f4/res-moninfo-dense.cpp
In file included from /usr/include/c++/10/memory:83,
                 from /home/dan/src/M2/M2/Macaulay2/e/f4/res-moninfo-dense.hpp:13,
                 from /home/dan/src/M2/M2/Macaulay2/e/f4/res-moninfo-dense.cpp:3:
/usr/include/c++/10/bits/unique_ptr.h: In instantiation of ‘constexpr std::unique_ptr<_Tp [], _Dp>::unique_ptr() [with _Del = std::default_delete<int []>; <template-parameter-2-2> = void; _Tp = int; _Dp = std::default_delete<int []>]’:
/home/dan/src/M2/M2/Macaulay2/e/f4/res-moninfo-dense.cpp:10:65:   required from here
/usr/include/c++/10/bits/unique_ptr.h:530:9: error: no matching function for call to ‘std::__uniq_ptr_data<int, std::default_delete<int []>, true, true>::__uniq_ptr_data()’
  530 |  : _M_t()
      |         ^
/usr/include/c++/10/bits/unique_ptr.h:209:40: note: candidate: ‘std::__uniq_ptr_data<int, std::default_delete<int []>, true, true>::__uniq_ptr_data(std::__uniq_ptr_impl<int, std::default_delete<int []> >::pointer) [inherited from std::__uniq_ptr_impl<int, std::default_delete<int []> >]’
  209 |       using __uniq_ptr_impl<_Tp, _Dp>::__uniq_ptr_impl;
      |                                        ^~~~~~~~~~~~~~~
/usr/include/c++/10/bits/unique_ptr.h:209:40: note:   candidate expects 1 argument, 0 provided
/usr/include/c++/10/bits/unique_ptr.h:209:40: note: candidate: ‘template<class _Del> std::__uniq_ptr_data<int, std::default_delete<int []>, true, true>::__uniq_ptr_data(std::__uniq_ptr_impl<int, std::default_delete<int []> >::pointer, _Del&&) [inherited from std::__uniq_ptr_impl<int, std::default_delete<int []> >]’
[...]

PR #1379, opened 2020-07-08
CMake maintenance

### LINKED ISSUE #1376: CMake does not detect OpenMP flags with clang-10 on macOS
@moorewf reported that using clang-10 installed via `brew install llvm libomp`, CMake doesn't seem to be able to find OpenMP flags, which causes csdp build to fail. My various attempts at giving CMake the prefix to find libomp.dylib failed.

The autotools build seems to be unaffected.

mahrud:
- [x] close #1376
- [x] add clang10 to github action tests
  - [x] tests pass with cmake
  - [ ] tests pass with autotools
- [x] various improvements to libraries linking with gmp/mpir


mahrud:
Seems like the workaround for llvm-clang breaks AppleClang!

.github/workflows/test_build.yml
```diff
@@ -40,18 +38,20 @@ jobs:
       matrix:
         os: [macos-latest, ubuntu-latest]
         build-type: [cmake, autotools]
-        compiler: [gcc9, default]  # add clang
+        compiler: [clang10, gcc9, default]
```
DanGrayson, line 41: What's the point of testing clang10, when we're already up to clang11 on the Mac?

mahrud: clang11 doesn't exist yet, as far as I know. The 11 in AppleClang is the version of Xcode.

mahrud:
@DanGrayson shouldn't all autotools targets use the M2 wrapper instead of M2-binary?
```
2020-07-10T13:41:29.2161390Z /home/runner/work/M2/M2/M2/BUILD/build/usr-dist/x86_64-Linux-Ubuntu-18.04/bin/M2 -q --silent --stop -e errorDepth=0 --no-preload --no-tvalues ../../../../Macaulay2/m2/tvalues-make.m2 -e 'make "../../../../Macaulay2/m2/../d/"; exit 0'
2020-07-10T13:41:29.2210430Z /home/runner/work/M2/M2/M2/BUILD/build/usr-dist/x86_64-Linux-Ubuntu-18.04/bin/M2-binary: error while loading shared libraries: libgivaro.so.9: cannot open shared object file: No such file or directory
2020-07-10T13:41:29.2211296Z Makefile:43: recipe for target 'tvalues.m2' failed
```

DanGrayson:
> @DanGrayson shouldn't all autotools targets use the M2 wrapper instead of M2-binary?
> 
> ```
> 2020-07-10T13:41:29.2161390Z /home/runner/work/M2/M2/M2/BUILD/build/usr-dist/x86_64-Linux-Ubuntu-18.04/bin/M2 -q --silent --stop -e errorDepth=0 --no-preload --no-tvalues ../../../../Macaulay2/m2/tvalues-make.m2 -e 'make "../../../../Macaulay2/m2/../d/"; exit 0'
> 2020-07-10T13:41:29.2210430Z /home/runner/work/M2/M2/M2/BUILD/build/usr-dist/x86_64-Linux-Ubuntu-18.04/bin/M2-binary: error while loading shared libraries: libgivaro.so.9: cannot open shared object file: No such file or directory
> 2020-07-10T13:41:29.2211296Z Makefile:43: recipe for target 'tvalues.m2' failed
> ```

Either that or it should set LD_LIBRARY_PATH.  In any case, in this case, it is running `M2`, as indicated on line 1.

DanGrayson:
> > @DanGrayson shouldn't all autotools targets use the M2 wrapper instead of M2-binary?
> > ```
> > 2020-07-10T13:41:29.2161390Z /home/runner/work/M2/M2/M2/BUILD/build/usr-dist/x86_64-Linux-Ubuntu-18.04/bin/M2 -q --silent --stop -e errorDepth=0 --no-preload --no-tvalues ../../../../Macaulay2/m2/tvalues-make.m2 -e 'make "../../../../Macaulay2/m2/../d/"; exit 0'
> > 2020-07-10T13:41:29.2210430Z /home/runner/work/M2/M2/M2/BUILD/build/usr-dist/x86_64-Linux-Ubuntu-18.04/bin/M2-binary: error while loading shared libraries: libgivaro.so.9: cannot open shared object file: No such file or directory
> > 2020-07-10T13:41:29.2211296Z Makefile:43: recipe for target 'tvalues.m2' failed
> > ```
> 
> Either that or it should set LD_LIBRARY_PATH. In any case, in this case, it is running `M2`, as indicated on line 1.

It was relatively recently that I converted givaro from a library to a submodule, as regarded by the configure script (June 12).  Maybe this is the first time we've had to build it, and after building it, we don't put the library where it can be found.  (?)  That's what I should test.

DanGrayson:
Meanwhile, adding `libgivaro-dev` to the list of packages installed with `apt-get` would fix that problem.

mahrud:
@DanGrayson I intentionally removed fflas-ffpack and givaro pending #1227, since it was breaking cmake builds. Apparently it was a good idea to try that and reveal this bug.

DanGrayson:
> @DanGrayson I intentionally removed fflas-ffpack and givaro pending #1227, since it was breaking cmake builds. Apparently it was a good idea to try that and reveal this bug.

Yup.  I've started the Ubuntu 20.04 build with those two not installed, so I'll be able to fix it.

mahrud:
@DanGrayson I added caching for CMake jobs, which speeds up builds significantly without missing changes to the code. However, I didn't enable it for Autotools for two reasons:
1. currently the size of the usr-host directory after building with Autotools is more than 1G (compare with ~400M for CMake), which is way too much and would cause github to delete catch archives for other jobs (maximum is 5G, shared between 8 or so jobs!)
2. the autotools build also builds *too much* for each library, resulting in the `libraries` directory to get to more than 5G (compare with 1G for CMake). This means ccache is probably overwhelmed and takes up more space.

I also disabled gcc7 on ubuntu and gcc9 on macOS. We can revisit this later, but I figured we're covering three compilers (clang, gcc, xcode) across the builds, so that's good enough.

I also added tests for CMake jobs.

mahrud:
Oh, also, I couldn't figure out why autotools can't use clang10 on ubuntu.

DanGrayson:
> Oh, also, I couldn't figure out why autotools can't use clang10 on ubuntu.

I'll take a look ...

DanGrayson:
I think commit 0a38544ced5b59c649f87e4124fab144300051d8 will fix the problem with ubuntu-latest-autotools-gcc9

mahrud:
@DanGrayson why is it building glpk rather than finding the version installed by linuxbrew?

DanGrayson:
> @DanGrayson why is it building glpk rather than finding the version installed by linuxbrew?

Good question, but I can't tell without the config.log artifacts, which you disabled.

mahrud:
I think the configure log is enough in this case:
```
configure: using BUILDLIBLIST  =  gc mpfr readline ntl flint factory mpsolve frobby glpk cddlib gtest
configure: using BUILDSUBLIST  =  memtailor mathic mathicgb fflas_ffpack givaro
configure: using BUILDPROGLIST =  4ti2 gfan normaliz csdp nauty lrslib gftables topcom cohomcalg
configure: using BUILDLIST     =  gc mpfr readline ntl flint factory mpsolve frobby glpk cddlib gtest 4ti2 gfan normaliz csdp nauty lrslib gftables topcom cohomcalg
configure: using BUILD_ALWAYS  =  readline gfan cddlib normaliz
```

I don't think it's looking under `brew --prefix` at all, hence missing all of `bdw-gc mpfr glpk flint ntl cddlib`.

mahrud:
Also, I don't think gfan and cddlib need to be built anymore. I don't think readline needs to be built either. In particular, cmake never builds readline.

DanGrayson:
> I think the configure log is enough in this case:
> 
> ```
> configure: using BUILDLIBLIST  =  gc mpfr readline ntl flint factory mpsolve frobby glpk cddlib gtest
> configure: using BUILDSUBLIST  =  memtailor mathic mathicgb fflas_ffpack givaro
> configure: using BUILDPROGLIST =  4ti2 gfan normaliz csdp nauty lrslib gftables topcom cohomcalg
> configure: using BUILDLIST     =  gc mpfr readline ntl flint factory mpsolve frobby glpk cddlib gtest 4ti2 gfan normaliz csdp nauty lrslib gftables topcom cohomcalg
> configure: using BUILD_ALWAYS  =  readline gfan cddlib normaliz
> ```
> 
> I don't think it's looking under `brew --prefix` at all, hence missing all of `bdw-gc mpfr glpk flint ntl cddlib`.

Brew's prefix is usually /usr/local, so it should be looking there anyway.

mahrud:
> Brew's prefix is usually /usr/local, so it should be looking there anyway.

This is only true on macOS, and even then if the user doesn't customize it. Here is mine:
```
[mahrud@noether M2]$ brew --prefix
/home/mahrud/.linuxbrew
```
This is the reason why I don't just hardcode `/usr/local` in cmake, by the way. A small advantage of supporting brew on linux is that while the ubuntu package for flint is still 2.5.2, flint via brew is already at 2.6.0 and cmake can take advantage of it.

DanGrayson:
Oh, yes, and I just remembered (and verified) that gcc-9 doesn't look in /usr/local.

Adding something like
```
   CPPFLAGS=-I`brew --prefix`/include  LDFLAGS=-L`brew --prefix`/lib
```
to the configure script command line might help, in that case.

mahrud:
Sure, but perhaps a solution so the user doesn't need to do this is better. On that note, the first part of this is also incorrect:
https://github.com/Macaulay2/M2/blob/e78fd58f92763fde048a16af40ab2045b0c3cfc0/M2/configure.ac#L61
On fedora, the path is `/usr/lib64` or `/lib64`, on ubuntu, the path is something like `/usr/lib/<architecture>`. The way cmake handles all of these situations is [documented here](https://cmake.org/cmake/help/latest/command/find_library.html).

mahrud:
Seems like it is more complicated than that; now factory can't find ntl:
```
+ set -x
+ grep ^.define HAVE_NTL 1 _config.h
../Makefile.library:180: recipe for target '.configured-4.1.1' failed
gmake[2]: Leaving directory '/home/runner/work/M2/M2/M2/BUILD/build/libraries/factory'
Makefile:7: recipe for target 'all-in-factory' failed
```
I'll revert the last change.

DanGrayson:
> Sure, but perhaps a solution so the user doesn't need to do this is better.

It's only the user who knows that brew has put some interesting files somewhere.

mahrud:
> It's only the user who knows that brew has put some interesting files somewhere.

It is also a prerequisite of brew packages to work with arbitrary prefixes, but perhaps only cmake will support that.

DanGrayson:
>  I don't think readline needs to be built either. In particular, cmake never builds readline.

I don't know why Radoslav decided to build readline every time.

DanGrayson:
> It found flint:
> 
> ```
> checking for flint >= 2.6.0... yes
> ```
> 
> However, seems like it is more complicated than that, since now factory can't find ntl:
> 
> ```
> + set -x
> + grep ^.define HAVE_NTL 1 _config.h
> ../Makefile.library:180: recipe for target '.configured-4.1.1' failed
> gmake[2]: Leaving directory '/home/runner/work/M2/M2/M2/BUILD/build/libraries/factory'
> Makefile:7: recipe for target 'all-in-factory' failed
> ```
> 
> I'll revert the last change for now. This enhancement for the autotools build probably needs its own PR.

I wonder why you think commit 49341303950ed6b742f51bdc230914044f00cca8 changed anything.  Isn't the /usr/local the brew prefix on our github action machines?

DanGrayson:
> > It found flint:
> > ```
> > checking for flint >= 2.6.0... yes
> > ```
> > 
> > 
> > However, seems like it is more complicated than that, since now factory can't find ntl:
> > ```
> > + set -x
> > + grep ^.define HAVE_NTL 1 _config.h
> > ../Makefile.library:180: recipe for target '.configured-4.1.1' failed
> > gmake[2]: Leaving directory '/home/runner/work/M2/M2/M2/BUILD/build/libraries/factory'
> > Makefile:7: recipe for target 'all-in-factory' failed
> > ```
> > 
> > 
> > I'll revert the last change for now. This enhancement for the autotools build probably needs its own PR.
> 
> I wonder why you think commit [4934130](https://github.com/Macaulay2/M2/commit/49341303950ed6b742f51bdc230914044f00cca8) changed anything. Isn't /usr/local the brew prefix on our github action machines?

Oh, apparently it's `/home/linuxbrew/.linuxbrew`.

DanGrayson:
Commit 49341303950ed6b742f51bdc230914044f00cca8 seems to be a good change.  It would be good to make it work.

DanGrayson:
> > It's only the user who knows that brew has put some interesting files somewhere.
> 
> It is also a prerequisite of brew packages to work with arbitrary prefixes, but perhaps only cmake will support that.
> 

What do you mean by this?

mahrud:
> Oh, apparently it's `/home/linuxbrew/.linuxbrew`.

Yes, see https://github.com/Macaulay2/M2/pull/1379#issuecomment-657559250

> Commit [4934130](https://github.com/Macaulay2/M2/commit/49341303950ed6b742f51bdc230914044f00cca8) seems to be a good change. It would be good to make it work.

Yes, but like I said, we should work on that in a separate issue or PR, since more changes with library Makefiles are required to get it to work, and I'd like to keep this PR focused on fixing clang10 build on CMake and adding the tests to the actions.

> > It is also a prerequisite of brew packages to work with arbitrary prefixes, but perhaps only cmake will support that.
> 
> What do you mean by this?

I have a working brew formula for building M2, and for that we need the build process to be able to find every brew package rather than build them, since then homebrew will not accept the formula. What I meant is that one build system supporting this is enough.

DanGrayson:
> > Oh, apparently it's `/home/linuxbrew/.linuxbrew`.
> 
> Yes, see [#1379 (comment)](https://github.com/Macaulay2/M2/pull/1379#issuecomment-657559250)
> 
> > Commit [4934130](https://github.com/Macaulay2/M2/commit/49341303950ed6b742f51bdc230914044f00cca8) seems to be a good change. It would be good to make it work.
> 
> Yes, but like I said, we should work on that in a separate issue or PR, since more changes with library Makefiles are required to get it to work, and I'd like to keep this PR focused on fixing clang10 build on CMake and adding the tests to the actions.

On second thought, it should be up to the user whether we peek into the homebrew location.  After all, how would the user make us *not* look there, if desired?

> 
> > > It is also a prerequisite of brew packages to work with arbitrary prefixes, but perhaps only cmake will support that.
> > 
> > 
> > What do you mean by this?
> 
> I have a working brew formula for building M2, and for that we need the build process to be able to find every brew package rather than build them, since then homebrew will not accept the formula. What I meant is that one build system supporting this is enough.

I'm not ready to give up on the makefiles.

mahrud:
@DanGrayson this PR is ready. The main remaining build that fails is autotools with clang, which fails because it can't find libomp. Do you know how to fix that? Installing libomp-dev didn't help.

DanGrayson:
Let's make all the builds succeed before we pull it.

DanGrayson:
> @DanGrayson this PR is ready. The main remaining build that fails is autotools with clang, ...

I see 4 builds failing, not 1.

DanGrayson:
> @DanGrayson this PR is ready. The main remaining build that fails is autotools with clang, which fails because it can't find libomp. Do you know how to fix that? Installing libomp-dev didn't help.

I remember I once succeeded with adding `-lgomp` somewhere ...  Point me to the corresponding `config.log` artifact.

mahrud:
@DanGrayson I think the other fails were minor issues because I hadn't merged from development. Here is the failing autotools+clang: https://github.com/Macaulay2/M2/suites/917497000/artifacts/11176441

mahrud:
@DanGrayson I don't understand why the autotools build is failing on mpsolve, but it probably has to do with the get-tools target, which just doesn't work with clang. Can you suggest a fix?

DanGrayson:
> @DanGrayson I think the other fails were minor issues because I hadn't merged from development.

I don't see how that could be relevant, because doesn't github merge with the target branch when running the actions?

mahrud:
I just put the get-tools target back in. The autotools+clang build will fail, but that should probably be addressed in a different PR.

DanGrayson:
If the error is the mysterious failure to compile m4 with a message about porting "fseek", then why not just skip building m4 and get just the other tools?:
```
make get-libtool get-automake get-autoconf
```

mahrud:
Is m4 available on macOS?

DanGrayson:
> Is m4 available on macOS?

Yes, all the tools are available.

DanGrayson:
Libtool is the only one on the mac that isn't modern enough for the makefile.

mahrud:
No they're not. It requires get-tools, otherwise it fails with "libtool version is not GNU"

DanGrayson:
Ah, right.  Libtool is available, but it isn't acceptable.

mahrud:
Strange, gctest is failing on autotools+gcc

mahrud:
I give up on making autotools use brew, and just opened an issue to deal with that. This might cause issues down the line because on ubuntu builds we have two installations of a bunch of libraries in two different prefixes.

mahrud:
I'm gonna close this PR and split it into two.

Issue #1380, opened 2020-07-08
printWidth not obeyed consistently


d-torrance:
For example, `Model` objects from `RandomMonomialIdeals` don't wrap.  But when we `peek` at them, they do:

```m2
i1 : loadPackage "RandomMonomialIdeals"

o1 = RandomMonomialIdeals

o1 : Package

i2 : ER(2,2,0.5)

o2 = Model{Generate => -*Function[/usr/share/Macaulay2/RandomMonomialIdeals.m2:129:23-129:47]*-}
           Name => Erdos-Renyi
           Parameters => (2, 2, .5)

o2 : Model

i3 : peek oo

o3 = Model{Generate => -*Function[/usr/share/Macaulay2/RandomMonomialIdeals.m2:
           Name => Erdos-Renyi
           Parameters => (2, 2, .5)
     --------------------------------------------------------------------------
     129:23-129:47]*-}
```

DanGrayson:
In this case, the non-wrapped version looks better.  Users who want to see the part off the screen to the right can scroll horizontally.

d-torrance:
I agree!  Also, the first one gets fixed by the `reproduciblePaths` function but the second one doesn't now that 3a686bc has been reverted.

DanGrayson:
So what's the issue about?

d-torrance:
It seems odd that some things wrap when others don't, especially when they're essentially the same thing.  In this case, `peek` shows us exactly the same information, but wrapped for some reason.

And it's not consistent from hash table to hash table, either.  A hash table full of long strings doesn't wrap whether we `peek` or not:

```m2
i1 : hashTable apply(10, i -> i => concatenate apply(10, i -> apply(10, toString)))

o1 = HashTable{0 => 0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789}
               1 => 0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789
               2 => 0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789
               3 => 0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789
               4 => 0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789
               5 => 0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789
               6 => 0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789
[...]

Issue #1372, opened 2020-07-06
line wrapping has changed for the worse


DanGrayson:
Compare https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.15/share/doc/Macaulay2/Macaulay2Doc/html/_elimination_spof_spvariables.html to https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.16/share/doc/Macaulay2/Macaulay2Doc/html/_elimination_spof_spvariables.html : 

<img width="488" alt="Screen Shot 2020-07-06 at 10 24 27 AM" src="https://user-images.githubusercontent.com/700228/86604122-95453200-bf6a-11ea-906e-3a16d37131a3.png">

<img width="639" alt="Screen Shot 2020-07-06 at 10 24 47 AM" src="https://user-images.githubusercontent.com/700228/86604136-9aa27c80-bf6a-11ea-89ac-30130df4c3e1.png">

DanGrayson:
@d-torrance -- this could be from your commit 3a686bc6b0d960f979fc72c4bc475abe765b2c02

DanGrayson:
I'd like to get this fixed before releasing 1.16.  Thanks to @giovannistagliano for letting me know about the problem.

d-torrance:
Yikes!  Yeah, it's definitely that commit.  Feel free to revert.

In many instances, I think the commit provides an improvement, especially when using `help` inside Macaulay2 (e.g., the screenshots I provided in https://github.com/Macaulay2/M2/pull/1337#issue-440148436).  But certainly anything like the example above involving printing the net form of a polynomial with the exponents on a separate line will look horrible!

The number of build paths that are split between lines when wrapping the examples (the original reason for the commit) is very small.  We'll still catch most of them without this commit.

d-torrance:
In fact, even without 3a686bc, I don't get any build paths in the documentation.  Some of the examples look a little odd because long build paths have been replaced with shorter paths and it looks like things were wrapped a bit too much.  But nowhere near as bad as the first example above!

```
i1 : path

o1 = {./, /usr/share/Macaulay2/Core/,
     ------------------------------------------------------------------------
     /usr/share/Macaulay2/,
     ------------------------------------------------------------------------
     /usr/share/
     ------------------------------------------------------------------------
     Macaulay2/Core/, /usr/
     ------------------------------------------------------------------------
     common/share/Macaulay2/}

o1 : List

[...]

PR #1381, opened 2020-07-08
Remove build paths from documentation (take 3)

d-torrance:
One of the commits from #1337 (3a686bc) has been reverted due to #1372.  Its goal was to wrap examples after they were generated so that `reproduciblePaths` could swap out any build paths without worrying about whether they had been split between two lines due to wrapping.

It turns out that even when wrapping the examples during generation as was done before #1337, only two examples were missed.  (At least on my system -- there may be some variation from system to system depending on the length of the build path.)

The goal of this pull request is to tweak these examples so this is not an issue.

### Example 1
In `RandomMonomialIdeals`, when describing the `Generate` key of a `Model` object, we `peek`'ed at the model, but this provided no extra information, as the exact same information was present when printing a `Model` without `peek`.  Furthermore, since `peek` results in a net, it was getting wrapped, splitting up a build path.  (Thanks again, Dan, for providing an explanation in #1380!)

So we just remove the `peek` line.

*Before*:
```m2
i1 : myModel = ER(2,2,0.5)

o1 = Model{Generate => -*Function[/usr/share/Macaulay2/RandomMonomialIdeals.m2:129:23-129:47]*-}
           Name => Erdos-Renyi
           Parameters => (2, 2, .5)

o1 : Model

i2 : peek myModel

o2 = Model{Generate => -*Function[/build/macaulay2-1.16+git41.3c2eac5+ds/M2/
           Name => Erdos-Renyi
           Parameters => (2, 2, .5)
     Macaulay2/packages/RandomMonomialIdeals.m2:129:23-129:47]*-}
```

*After*:
```m2
i1 : myModel = ER(2,2,0.5)

o1 = Model{Generate => -*Function[/usr/share/Macaulay2/RandomMonomialIdeals.m2:129:23-129:47]*-}
           Name => Erdos-Renyi
           Parameters => (2, 2, .5)

o1 : Model
```

### Example 2
In `NormalToricVarieties`, the code of the `(monomialIdeal, NormalToricVariety)` method is displayed.  Since it's a one-liner, we get a net with 3 rows (the ` -- code for ...` comment, the path, and the code itself).  Any net of 3 or fewer rows will be wrapped.

So we add a newline in the middle of the code so we end up with a net with 4 rows, which won't be wrapped.

*Before*:
```m2
i20 : code (monomialIdeal, NormalToricVariety)

o20 = -- code for method: monomialIdeal(NormalToricVariety
      /build/macaulay2-1.16+git41.3c2eac5+ds/M2/Macaulay2/
      monomialIdeal NormalToricVariety := MonomialIdeal =>
      -----------------------------------------------------------------------
      )
      packages/NormalToricVarieties/Sheaves.m2:33:56-33:79: --source code:
       X -> monomialIdeal ideal X
```

*After*:
```m2
i20 : code (monomialIdeal, NormalToricVariety)

o20 = -- code for method: monomialIdeal(NormalToricVariety)
      /usr/share/Macaulay2/NormalToricVarieties/Sheaves.m2:33:56-34:25: --source code:
      monomialIdeal NormalToricVariety := MonomialIdeal => X ->
          monomialIdeal ideal X
```

### DIFF
--- a/M2/Macaulay2/m2/installPackage.m2
+++ b/M2/Macaulay2/m2/installPackage.m2
@@ -439,9 +439,8 @@ reproduciblePaths = outf -> (
      if topSrcdir === null then return outstr;
      srcdir := regexQuote toAbsolutePath topSrcdir;
      prefixdir := regexQuote prefixDirectory;
-     builddir := substring(0, #prefixdir - 9, prefixdir); -- 9 = #"usr-dist/"
-     homedir := regexQuote homeDirectory;
-     homedir = substring(0, #homedir - 1, homedir); -- remove trailing "/"
+     builddir := replace("usr-dist/?$", "", prefixdir);
+     homedir := replace("/$", "", regexQuote homeDirectory);
      if any({srcdir, builddir, homedir}, dir -> match(dir, outstr))
      then (
 	 -- .m2 files in source directory
--- a/M2/Macaulay2/m2/installPackage.m2
+++ b/M2/Macaulay2/m2/installPackage.m2
@@ -438,25 +438,27 @@ reproduciblePaths = outf -> (
      outstr := get outf;
      if topSrcdir === null then return outstr;
      srcdir := regexQuote toAbsolutePath topSrcdir;
-     builddir := regexQuote prefixDirectory;
+     prefixdir := regexQuote prefixDirectory;
+     builddir := substring(0, #prefixdir - 9, prefixdir); -- 9 = #"usr-dist/"
      homedir := regexQuote homeDirectory;
-     if last homedir == "/" then homedir = substring(0, #homedir - 1, homedir);
+     homedir = substring(0, #homedir - 1, homedir); -- remove trailing "/"
      if any({srcdir, builddir, homedir}, dir -> match(dir, outstr))
      then (
-	 outstr = replace(srcdir | "Macaulay2/m2/startup.m2.in",
-	     "/path/to/source/Macaulay2/m2/startup.m2.in", outstr);
+	 -- .m2 files in source directory
 	 outstr = replace(srcdir | "Macaulay2/m2",
 	     finalPrefix | Layout#1#"packages" | "Core", outstr);
 	 outstr = replace(srcdir | "Macaulay2/packages/",
 	     finalPrefix | Layout#1#"packages", outstr);
-	 outstr = replace(builddir | Layout#2#"bin",
-	     finalPrefix | Layout#1#"bin", outstr);
-	 outstr = replace(builddir | Layout#2#"data",
-	     finalPrefix | Layout#1#"data", outstr);
-	 outstr = replace(builddir | Layout#2#"lib",
-	     finalPrefix | Layout#1#"lib", outstr);
+	 -- generated .m2 files in build directory (tvalues.m2)
+	 outstr = replace(builddir | "Macaulay2/m2",
+	     finalPrefix | Layout#1#"packages" | "Core", outstr);
+	 -- everything in staging area
+	 scan({"bin", "data", "lib", "program licenses", "programs"}, key ->
+	     outstr = replace(prefixdir | Layout#2#key,
+		 finalPrefix | Layout#1#key, outstr));
+	 outstr = replace(prefixdir, finalPrefix, outstr);
+	 -- home directory
 	 outstr = replace(homedir, "/home/m2user", outstr);
-	 outstr = replace(builddir, finalPrefix, outstr);
 	 outf << outstr << close;
 	 );
      outstr
--- a/M2/Macaulay2/m2/installPackage.m2
+++ b/M2/Macaulay2/m2/installPackage.m2
@@ -434,8 +434,7 @@ dispatcherMethod := m -> m#-1 === Sequence and (
      f := lookup m;
      any(dispatcherFunctions, g -> functionBody f === functionBody g))
 
-reproduciblePaths = outf -> (
-     outstr := get outf;
+reproduciblePaths = outstr -> (
      if topSrcdir === null then return outstr;
      srcdir := regexQuote toAbsolutePath topSrcdir;
      prefixdir := regexQuote prefixDirectory;
@@ -458,7 +457,6 @@ reproduciblePaths = outf -> (
 	 outstr = replace(prefixdir, finalPrefix, outstr);
 	 -- home directory
 	 outstr = replace(homedir, "/home/m2user", outstr);
-	 outf << outstr << close;
 	 );
      outstr
     )
@@ -656,7 +654,8 @@ installPackage Package := opts -> pkg -> (
 			 );
 		    -- read, separate, and store example output
 		    if fileExists outf then (
-			 outstr := reproduciblePaths outf;
+			 outstr := reproduciblePaths get outf;
+			 outf << outstr << close;
 			 pkg#"example results"#fkey = drop(separateM2output outstr,-1)
 		    )
 		    else (
--- a/M2/Macaulay2/packages/NormalToricVarieties/Sheaves.m2
+++ b/M2/Macaulay2/packages/NormalToricVarieties/Sheaves.m2
@@ -30,8 +30,8 @@ ideal NormalToricVariety := Ideal => (
 	    )
 	)
     );
-monomialIdeal NormalToricVariety := MonomialIdeal => X -> monomialIdeal ideal X
-
+monomialIdeal NormalToricVariety := MonomialIdeal => X ->
+    monomialIdeal ideal X
 
 ------------------------------------------------------------------------------
 -- sheaves
--- a/M2/Macaulay2/m2/document.m2
+++ b/M2/Macaulay2/m2/document.m2
@@ -671,7 +671,7 @@ fixupTable := new HashTable from {
 				   if f === null then error("SourceCode: ", toString m, ": not a method");
 				   c := code f;
 				   if c === null then error("SourceCode: ", toString m, ": code for method not found");
-				   c)))}}
+				   reproduciblePaths toString c)))}}
 	  else v),
      Subnodes => v -> (
 	  v = nonnull enlist v;
--- a/M2/Macaulay2/packages/Macaulay2Doc/doc.m2
+++ b/M2/Macaulay2/packages/Macaulay2Doc/doc.m2
@@ -277,8 +277,8 @@ document {
      PRE replace(regexQuote homeDirectory, "/home/m2user/",
 	   concatenate between_"\n" apply(value Core#"private dictionary"#"userpath",s -> (5,s))),
      EXAMPLE {
-	  "path",
-	  ///path = append(path, "~/resolutions/")///
+	  "stack path",
+	  ///path = append(path, "~/resolutions/"); stack path///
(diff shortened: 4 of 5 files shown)


M2/Macaulay2/packages/NormalToricVarieties/Sheaves.m2
```diff
@@ -30,8 +30,8 @@ ideal NormalToricVariety := Ideal => (
 	    )
 	)
     );
-monomialIdeal NormalToricVariety := MonomialIdeal => X -> monomialIdeal ideal X
```
mahrud, line 33: This doesn't seem necessary.

d-torrance:
I need to do some more work on the `reproduciblePaths` function.  While it catches all except for the above two examples with the build paths I've been using, it didn't do such a good job with Dan's!  From the new Ubuntu 18.04 package of 1.16:

```
profzoom@orange:/usr/share/doc/Macaulay2$ grep -r "BUILD/dan/builds.tmp/" | wc -l
118
```

DanGrayson:
> I need to do some more work on the `reproduciblePaths` function. While it catches all except for the above two examples with the build paths I've been using, it didn't do such a good job with Dan's! From the new Ubuntu 18.04 package of 1.16:
> 
> ```
> profzoom@orange:/usr/share/doc/Macaulay2$ grep -r "BUILD/dan/builds.tmp/" | wc -l
> 118
> ```

Could you show us a few examples?

d-torrance:
> Could you show us a few examples?

```
profzoom@orange:/usr/share/doc/Macaulay2$ grep -r "BUILD/dan/builds.tmp/" | tail
Macaulay2Doc/example-output/_prefix__Path.out:o2 = /home/m2user/src/M2/M2/BUILD/dan/builds.tmp/ubuntu1804-release-1.16/usr-
Macaulay2Doc/example-output/_prefix__Path.out:o3 = /home/m2user/src/M2/M2/BUILD/dan/builds.tmp/ubuntu1804-release-1.16/usr-
Macaulay2Doc/example-output/_loaded__Files.out:                      93 => /home/m2user/src/M2/M2/BUILD/dan/builds.tmp/ubuntu1804-release-1.16/Macaulay2/m2/tvalues.m2
Macaulay2Doc/example-output/_copyright.out:     For licenses see /home/m2user/src/M2/M2/BUILD/dan/builds.tmp/ubuntu1804-release-1.16/usr-dist/x86_64-Linux-Ubuntu-18.04/libexec/Macaulay2/program-licenses/
Macaulay2Doc/example-output/_path.out:     /home/m2user/src/M2/M2/BUILD/dan/builds.tmp/ubuntu1804-release-1.16/
Macaulay2Doc/example-output/_path.out:     Macaulay2/m2/, /home/m2user/src/M2/M2/BUILD/dan/builds.tmp/ubuntu1804-
Macaulay2Doc/example-output/_path.out:     /home/m2user/src/M2/M2/BUILD/dan/builds.tmp/ubuntu1804-release-1.16/usr-
Macaulay2Doc/example-output/_path.out:     /home/m2user/src/M2/M2/BUILD/dan/builds.tmp/ubuntu1804-release-1.16/
Macaulay2Doc/example-output/_path.out:     Macaulay2/m2/, /home/m2user/src/M2/M2/BUILD/dan/builds.tmp/ubuntu1804-
Macaulay2Doc/example-output/_path.out:     /home/m2user/src/M2/M2/BUILD/dan/builds.tmp/ubuntu1804-release-1.16/usr-
```

You can see these particular examples on the webpage:
https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.16/share/doc/Macaulay2/Macaulay2Doc/html/_prefix__Path.html
https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.16/share/doc/Macaulay2/Macaulay2Doc/html/_loaded__Files.html
https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.16/share/doc/Macaulay2/Macaulay2Doc/html/_copyright.html
https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.16/share/doc/Macaulay2/Macaulay2Doc/html/_path.html

I think swapping these two lines will fix many of these.  We'd been swapping out the home directory before the final swap of the build path, and so if the build happened inside the home directory (which I hadn't been doing), the build path would be all mangled with the fake home directory before that final swap.
https://github.com/Macaulay2/M2/blob/818b61cb76df7035f6011ed860d0024fcd91e6f8/M2/Macaulay2/m2/installPackage.m2#L458-L459

d-torrance:
We should also call `reproduciblePaths` when we encounter the `SourceCode` key, and not just for examples:

```html
profzoom@orange:/usr/share/doc/Macaulay2/Macaulay2Doc/html$ grep -C 1 integers.m2 _zero.html 
          <h2>Code</h2>
          <pre>/home/dan/src/M2/M2/Macaulay2/m2/integers.m2:47:10-47:18: --source code:
zero = x -> x == 0                                          -- we use == so this can apply to all types of things</pre>
```

I hadn't noticed this because `minimizeFilename` must have decided relative paths were the best for my builds:
```html
profzoom@orange:~/tmp/m2/usr/share/doc/Macaulay2/Macaulay2Doc/html$ grep -C 1 integers.m2 _zero.html
          <h2>Code</h2>
          <pre>../../Core/integers.m2:47:10-47:18: --source code:
zero = x -> x == 0                                          -- we use == so this can apply to all types of things</pre>
```

M2/Macaulay2/m2/installPackage.m2
```diff
@@ -438,25 +438,27 @@ reproduciblePaths = outf -> (
      outstr := get outf;
      if topSrcdir === null then return outstr;
      srcdir := regexQuote toAbsolutePath topSrcdir;
-     builddir := regexQuote prefixDirectory;
+     prefixdir := regexQuote prefixDirectory;
+     builddir := substring(0, #prefixdir - 9, prefixdir); -- 9 = #"usr-dist/"
```
mahrud, line 442: (edit: moved the comment to the PR rather than the commit)
`prefixDirectory` only ends in `usr-dist` when M2 is run from a build directory. Otherwise it's `/usr` or `/usr/local`.

d-torrance: Correct, but we won't get this far in the code if we're not running it from the build directory since `topSrcdir` will be `null`, so we'll leave in line 439.

d-torrance:
After the last two commits, I'm down to just two examples that aren't working well:

```m2
      i1 : stack apply(prefixPath, p -> p | Layout#1#"packages" | "Foo.m2")

      o1 = /home/m2user/src/M2/M2/BUILD/dan/builds.tmp/ubuntu1804-release-1.16/usr-
           dist/share/Macaulay2/Foo.m2

      i2 : stack apply(prefixPath, p -> p | replace("PKG","Foo",Layout#1#"packagehtml") | "bar.html")

      o2 = /home/m2user/src/M2/M2/BUILD/dan/builds.tmp/ubuntu1804-release-1.16/usr-
           dist/share/doc/Macaulay2/Foo/html/bar.html

      i3 : stack apply(prefixPath, p -> p | Layout#1#"info" | "Foo.info")

      o3 = /home/m2user/src/M2/M2/BUILD/dan/builds.tmp/ubuntu1804-release-1.16/usr-
           dist/share/info/Foo.info
```

```m2
      i1 : cohomCalgExecutable

      o1 = /home/m2user/src/M2/M2/BUILD/dan/builds.tmp/ubuntu1804-release-1.16/usr-
           dist/x86_64-Linux-Ubuntu-18.04/libexec/Macaulay2/bin/cohomcalg
```

The problem is that the build path is so long that `usr-dist` is getting split up between two lines, and  so the build path isn't getting noticed by `reproduciblePaths`.

I see two possible solutions:
* Increase the `printWidth` for examples to something bigger than 77.  So many of the examples already aren't being wrapped anyway (nets of 4 or more rows) that this shouldn't cause too many problems.
* Do one last-ditch replacement of any instances of the build path but *without* `usr-dist`.

I think I'd prefer the first option, as the second option would really only help in situations where the build path is about the same length as Dan's.  If it were a bit longer, we might end up with something like this:
```m2
      i1 : cohomCalgExecutable

      o1 = /home/m2user/src/M2/M2/BUILD/dan/builds.tmp/ubuntu1804-release-1.16/this-
           path-is-longer/usr-dist/x86_64-Linux-Ubuntu-18.04/libexec/Macaulay2/bin/cohomcalg
```

And even if we did remove the build path, the example would look strange:

```m2
      i1 : cohomCalgExecutable

      o1 = usr-
           dist/x86_64-Linux-Ubuntu-18.04/libexec/Macaulay2/bin/cohomcalg
```

Any thoughts?

DanGrayson:
Yes, I wonder why you are building in `M2/BUILD/dan/builds.tmp/ubuntu1804-release-1.16`.  (I  assume you're the one in control.)  Why not just build in something short, such as `M2/x`?

d-torrance:
> Yes, I wonder why you are building in `M2/BUILD/dan/builds.tmp/ubuntu1804-release-1.16`. (I assume you're the one in control.) Why not just build in something short, such as `M2/x`?

I just ran the updated `reproduciblePaths` on the examples that were already generated during your build of the Ubuntu 18.04 package.  The build path I've been using isn't quite as short as `M2/x`, but it's short enough that I'm not running into to these particular issues.

DanGrayson:
> > Yes, I wonder why you are building in `M2/BUILD/dan/builds.tmp/ubuntu1804-release-1.16`. (I assume you're the one in control.) Why not just build in something short, such as `M2/x`?
> 
> I just ran the updated `reproduciblePaths` on the examples that were already generated during your build of the Ubuntu 18.04 package. The build path I've been using isn't quite as short as `M2/x`, but it's short enough that I'm not running into to these particular issues.

As long as printWidth is finite, a long enough path will foil your technique, so why worry further?  Isn't the important thing for Debian that Debian builds are reproducible?

d-torrance:
> As long as printWidth is finite, a long enough path will foil your technique, so why worry further? Isn't the important thing for Debian that Debian builds are reproducible?

Fair enough.  :)

M2/Macaulay2/m2/installPackage.m2
mahrud: Still, something like `replace("usr-dist/?", "", s)` would be safer.

d-torrance: Good point!  I was emulating the corresponding code in `startup.m2.in`:
https://github.com/Macaulay2/M2/blob/e78fd58f92763fde048a16af40ab2045b0c3cfc0/M2/Macaulay2/m2/startup.m2.in#L332-L336

I'll push a new commit changing it to use `replace` soon.

d-torrance:
Fixed the `SourceCode` issue:
```html
profzoom@orange:/tmp/m2/usr-dist/common/share/doc/Macaulay2/Macaulay2Doc/html$ grep -C 1 integers.m2 _zero.html
          <h2>Code</h2>
          <pre>/usr/share/Macaulay2/Core/integers.m2:47:10-47:18: --source code:
zero = x -> x == 0                                          -- we use == so this can apply to all types of things</pre>
```

This PR is ready now.

PR #1386, opened 2020-07-10
Update SparseResultants

giovannistagliano:
Is it possible to merge in the 1.16 version?


DanGrayson:
> Is it possible to merge in the 1.16 version?

No, it's too late for that: many distributions have already been made.

DanGrayson:
Giovanni, the master branch currently succeeds 6/8 of the time (see https://github.com/Macaulay2/M2/actions/runs/167196616 ).  Could you merge your branch with our current master branch, so we can see whether it improves your build results?

giovannistagliano:
Sorry, how do I do this? I just merged a pull request of my branch. Is this?

DanGrayson:
> Giovanni, the master branch currently succeeds 6/8 of the time (see https://github.com/Macaulay2/M2/actions/runs/167196616 ). Could you merge your branch with our current master branch, so we can see whether it improves your build results?

This command ought to bring you up to date on your computer:
```
git pull https://github.com/Macaulay2/M2.git master
```
And then this will bring your clone up to date:
```
git push
```

giovannistagliano:
I think I did it.

DanGrayson:
I think you did, too.

DanGrayson:
If you'd like an abbreviation for the URL of a remote repository, see the command 'git remote', which is described in the man page for 'git-remote'.

giovannistagliano:
Thank you for your suggestion. I should start to understand Github better ...

M2/Macaulay2/packages/SparseResultants.m2
```diff
@@ -462,6 +463,210 @@ genericMultihomogeneousPolynomial (VisibleList,VisibleList) := o -> (k,d) -> (
...
+MultidimensionalMatrix == MultidimensionalMatrix := (M,N) -> (
+   if M#"format" != N#"format" then return false;
+   if M#"ring" =!= N#"ring" then return false;
+   M#"multilinearForm" == N#"multilinearForm"
+);
+
+transposition = method();
+transposition (MultidimensionalMatrix,List) := (M,l) -> (
+    n := M#"format";
+    try assert(ring matrix {l} === ZZ and sort l == toList(0 .. #n-1)) else error("expected a permutation of the set "|toString toList(0 .. #n-1));
```
DanGrayson, line 522: It's better to use `if not ... then ... ` here instead of `try assert ... else ...`, because it's more logical, and because other errors will be masked.  For example, what if n is not a list, so #n gives an error?  Then you'll never see it.

M2/Macaulay2/packages/SparseResultants.m2
```diff
@@ -462,6 +463,210 @@ genericMultihomogeneousPolynomial (VisibleList,VisibleList) := o -> (k,d) -> (
...
+    );
+    for i to #n-2 do L = flatten L;
+    K := ring matrix{L};
+    gensRing0(K,n)
+);
+
+makeMultilinearForm = method();
+makeMultilinearForm (List,List) := (L,X) -> (
+    if not instance(first L,List) then return ((matrix{last X})*transpose(matrix{L}))_(0,0);
+    X' := X_{0 .. #X-2};
```
DanGrayson, line 550: It's better to use `take` here, for speed.

M2/Macaulay2/packages/SparseResultants.m2
```diff
@@ -462,6 +463,210 @@ genericMultihomogeneousPolynomial (VisibleList,VisibleList) := o -> (k,d) -> (
...
+    return makeMultilinearForm(apply(L,l -> makeMultilinearForm(l,X')),X);
+);
+
+makeList = method();
+makeList (RingElement,List) := (F,X) -> (
```
DanGrayson, line 555: This is not a descriptive name for this function.  Can you think of a better one?

M2/Macaulay2/packages/SparseResultants.m2
```diff
@@ -462,6 +463,210 @@ genericMultihomogeneousPolynomial (VisibleList,VisibleList) := o -> (k,d) -> (
...
+coeffs = method();
+coeffs (RingElement) := (F) -> flatten entries sub(last coefficients(matrix F,Monomials=>vars ring F),coefficientRing ring F);
+coeffs (List) := (L) -> apply(L,F -> coeffs F);
+
+applyToEntries = method();
+applyToEntries (List,Function) := (M,f) -> (
+    ff := method();
+    ff Thing := t -> f(t);
+    ff List := l -> apply(l,ff);
+    ff(M)
```
DanGrayson, line 572: This is too heavy a use of "method".  Use something like `if instance(M,List) then ... else ...`.

M2/Macaulay2/packages/SparseResultants.m2
```diff
@@ -462,6 +463,210 @@ genericMultihomogeneousPolynomial (VisibleList,VisibleList) := o -> (k,d) -> (
...
+
+genericMultidimensionalMatrix = method(TypicalValue => MultidimensionalMatrix, Options => {CoefficientRing => ZZ});
+genericMultidimensionalMatrix (VisibleList,Boolean) := o -> (n,b) -> (
+    F := genericMultihomogeneousPolynomial(toSequence n,#n:1,CoefficientRing=>o.CoefficientRing);
+    K := coefficientRing coefficientRing ring F;
+    if b then F = sub(sub(F,apply(gens coefficientRing ring F,u -> u => random(K))),K[gens ring F]);
+    x := gens ring F;
+    X := {};
+    for i to #n-1 do (
+        X = append(X,x_{0 .. n_i -1});
```
DanGrayson, line 584: Never make a list by using "append" to add one element at a time -- it's the standard way to convert an order N algorithm to an order N^2 algorithm, by copying those lists so often.  Use one of the many list processing functions instead.  Also use `take` to select the elements of the list `x`.

M2/Macaulay2/packages/SparseResultants.m2
```diff
@@ -462,6 +463,210 @@ genericMultihomogeneousPolynomial (VisibleList,VisibleList) := o -> (k,d) -> (
...
+
+randomMultidimensionalMatrix = method(TypicalValue => MultidimensionalMatrix, Options => {CoefficientRing => ZZ});
+randomMultidimensionalMatrix (VisibleList) := o -> (n) -> genericMultidimensionalMatrix(n,true,CoefficientRing=>o.CoefficientRing);
+
+GENSRING0 := local GENSRING0;
+
+gensRing0 = method();
+gensRing0 (Ring,List) := (K,n) -> (
+    if instance(GENSRING0_(K,n),List) then return GENSRING0_(K,n);
+    x := apply(#n,i -> getSymbol("x"|toString(i)));
```
DanGrayson, line 599: All this stuff with `GENSRING` can probably be replaced by a single use of `memoize`.

M2/Macaulay2/packages/SparseResultants.m2
```diff
@@ -462,6 +463,210 @@ genericMultihomogeneousPolynomial (VisibleList,VisibleList) := o -> (k,d) -> (
...
+    }
+);
+
+Thing * MultidimensionalMatrix := (a,M) -> (
+    applyToEntries(M,m -> a*m)
+);
+
+RingElement * MultidimensionalMatrix := (a,M) -> (
+    applyToEntries(M,m -> a*m)
+);
```
DanGrayson, line 497: This method is redundant, since is does the same thing as the more general one above.  It can be removed.

M2/Macaulay2/packages/SparseResultants.m2
```diff
@@ -462,6 +463,210 @@ genericMultihomogeneousPolynomial (VisibleList,VisibleList) := o -> (k,d) -> (
...
+    F := M#"multilinearForm";
+    K := M#"ring";
+    X := M#"varsMultilinearForm";    
+    G := sub(F,(K[X_i0])[flatten X_(toList delete(i0,0..#n-1))]);
+    N := multidimensionalMatrix makeList(G,X_(toList delete(i0,0..#n-1)));
+    detN := schlafliMethod(N,Strategy=>o.Strategy);
+    if detN == 0 then return 0_(M#"ring");
+    d := first degree detN;
+    detN' := sub(sub(detN,first gens ring detN => 1),K[(gens ring detN)_{1 .. n_i0-1}]);
+    try assert(d == first degree detN') else error "unhomogenization failed";
```
DanGrayson, line 645: Replace `try` by `if` here, also.

giovannistagliano:
Many thanks for all the detailed comments on the code. They will be very useful to me and I'll apply them soon.

M2/Macaulay2/packages/SparseResultants.m2
giovannistagliano: Now I left only the following `try assert(ring matrix {l} === ZZ) else error ... ` It should fail exactly when l is not a list of integers.

M2/Macaulay2/packages/SparseResultants.m2
giovannistagliano: Done.

M2/Macaulay2/packages/SparseResultants.m2
giovannistagliano: Renamed into `makeHyperrectangularArray`.

M2/Macaulay2/packages/SparseResultants.m2
giovannistagliano: Done.

M2/Macaulay2/packages/SparseResultants.m2
giovannistagliano: Ok, thanks. Changed.

M2/Macaulay2/packages/SparseResultants.m2
giovannistagliano: Yes, changed. I wasn't aware of this useful method.

M2/Macaulay2/packages/SparseResultants.m2
giovannistagliano: Actually the method `Thing * MultidimensionalMatrix` is hidden and undocumented. Otherwise I should also define `ZZ * MultidimensionalMatrix`, etc. The only documented method is `RingElement * MultidimensionalMatrix`.

M2/Macaulay2/packages/SparseResultants.m2
giovannistagliano: Done.

giovannistagliano:
I would have finished.

M2/Macaulay2/packages/SparseResultants.m2
DanGrayson: The purpose of that line will be unclear to the reader.  Just use something like
```m2
if not all(l, i -> instance(i,ZZ)) then error ...
```

M2/Macaulay2/packages/SparseResultants.m2
DanGrayson: Why not document the actual behavior of your code, so people can understand it better?  When documenting `Thing * MultidimensionalMatrix` you can mention that not all rings are engine rings, and so not all elements of rings inherit from class RingElement.

DanGrayson:
> I would have finished.

"would have"?

M2/Macaulay2/packages/SparseResultants.m2
giovannistagliano: Ok, changed.

M2/Macaulay2/packages/SparseResultants.m2
giovannistagliano: Thinking about it, the method `Thing * Mult..` was not very logical since `Mult.. * Mult..` also exists with a completely different definition. I have added the methods `ZZ * Mult..` and `QQ * Mult..`.

giovannistagliano:
If there are no other issues or suggestions to the code, I am done (although I'll probably update it before the next release of M2).

PR #1389, opened 2020-07-10
Rename loadProgram to findProgram

d-torrance:
As suggested by Dan in https://github.com/Macaulay2/M2/commit/a3212614f019837d26925cfbd9316d2feac8cc61#r40517131


DanGrayson:
Could you merge with our master branch to see whether we can get 6 of the builds to succeed?

d-torrance:
> Could you merge with our master branch to see whether we can get 6 of the builds to succeed?

Done.  I ended up rebasing it on top of `development`, because using `master` would have added a couple of your commits to this PR, which seemed strange.  It looks like the only difference between the two branches right now is some PHCpack example output, so I don't think that will affect the builds.

DanGrayson:
Oh, yeah, maybe I should also merge "master" into "development" once in a while!

mahrud:
@DanGrayson the practice of rebasing on master for PRs to development is making the history more nonlinear than usual, since every time there has to be a merge. Why not just rebase on development? If a PR is breaking tests, it shouldn't be merged with development, so if rebasing on development causes tests to break it's not from the development branch itself.

DanGrayson:
Not sure what your question is about -- I've never rebased anything, and I've never asked anyone to rebase anything.

mahrud:
Sorry, I meant merging from master. I edited the comment.

DanGrayson:
> @DanGrayson the practice of merging from master for PRs to development is making the history more nonlinear than usual, since every time there has to be a merge. Why not just rebase on (or merge from) development? If a PR is breaking tests, it wouldn't be merged with development, so if rebasing on development causes tests to break it's not from the development branch itself.

The idea of the "development" branch is that things there may not have been tested, and that's where they get tested: work from Mike and me.  It's best for users with code under development not be plagued by possible bugs.  The idea of the "master" branch is that its tip always has passed tests, both the github actions and the overnight test on habanero.

mahrud:
That was before we had the workflow tests. Now, I don't think PRs should be merged even with development before they pass the github tests (minus the few that are broken even on master). This way if the builds on a PR (which is on top of development, as always) fail, it's because of the PR itself.

mahrud:
And later, when they also pass overnight tests on habanero, development can be merged with the primary branch.

DanGrayson:
> That was before we had the workflow tests. Now, I don't think PRs should be merged even with development before they pass the github tests (minus the few that are broken even on master). This way if the builds on a PR (which is on top of development, as always) fail, it's because of the PR itself.

Well, if we make the "development" branch that stable, we might as well get rid of it and just submit pull requests to "master".

By the way, do the github actions always merge the proposed pull request with the current tip of the target branch?  Maybe we should display a few log messages during the actions so we can see which code is being tested.

d-torrance:
> By the way, do the github actions always merge the proposed pull request with the current tip of the target branch?

That's what it looks like.  From https://github.com/Macaulay2/M2/runs/870684912?check_suite_focus=true, at the very bottom of the "Run actions/checkout@v1" log:

```
HEAD is now at 35dd31090 Merge 3edc32e42063fece3f99b7209c13b971753a57e9 into 74c3aa6389cbcb5bc2997e384f97f088a6b93efb
```
3edc32e42063fece3f99b7209c13b971753a57e9 is the commit at the HEAD of my branch and 74c3aa6389cbcb5bc2997e384f97f088a6b93efb is at the HEAD of `development`.

PR #1390, opened 2020-07-11
Add option to use generic path name for cached source files

d-torrance:
The paths to `startup.m2.in` and the sources of all of the basic tests
are stored as strings in `M2-binary`, preventing reproducible builds.

In #1285, it was proposed to replace these paths with the generic
"/path/to/source" for reproducibility.  However, this was closed due
after two concerns were pointed out:

* Developers may want the original paths for debugging.
* The original proposal was autotools-only.

This new commit addresses both issues:

* Replacing the path is now an option set at the beginning of the build.
  The original behavior of keeping the original path is the default.
  So developers interested in having these paths for debugging purposes
  don't need to do anything.  Downstream package maintainers interested
  in reproducible builds can set the option.
* It supports both autools and cmake:

      ./configure --enable-reproducible-cached-file-paths

  or

      cmake -DREPRODUCIBLE_CACHED_FILE_PATHS=ON


mahrud:
I think this is the wrong approach for doing this. See #1391 for an alternative implemented for the CMake build. I can implement the same for the autotools build in a few days, but if you'd like to help with that approach instead, I'm happy to give you push access to that branch.

d-torrance:
Closing in favor of #1391.

PR #1391, opened 2020-07-11
Tell 'code' how to locate startup.m2

mahrud:
This PR partially fixes reproducibility issues related to startup.m2 and basic tests by:
- [x] adding a link `Macaulay2/Core -> Macaulay2/m2`
- [x] changing the addresses for these files to be relative to `topSrcdir`
- [x] modify code.m2 to look for startup.m2.in if `topSrcdir` is define, or else look for startup.m2 under the installation directory.

Possible remaining tasks:
- [x] implement the same with autotools
- [ ] also distribute the basic tests
- [ ] modify the error function to do the same with the address of a basic test that fails
- [ ] possibly modify the `locate` function in `actors4.d` to do the above instead
- [ ] completely move `m2` directory to `Core` in the repository

And in the long term:
- [ ] reduce the contents of `startupString`
- [ ] allow running basic tests from real files


mahrud:
@DanGrayson @mikestillman would you be okay to moving `m2` to `Core`?

d-torrance:
Awesome, this is way more elegant than my solution!

I made the corresponding change to the autotools build in https://github.com/d-torrance/M2/commit/337a9f043bbc5e254fe02ef58cb17e50e8aeaf74.  Feel free to either merge it into your branch or give me push access -- whatever's easiest.

mahrud:
You should have push access now. Hopefully your commit fixes the build errors.

d-torrance:
Pushed, thanks!

DanGrayson:
> @DanGrayson @mikestillman would you be okay to moving `m2` to `Core`?

I don't see any advantage, but if the point is that it would make `debug Core` more sensible, we should change Macaulay2Doc to Core and move all the files `m2/*.m2` to there.  One could also add a file `packages/Core.m2` to complete the sham.

In general I don't like renaming, though, because it makes it hard to follow the history.

d-torrance:
This is causing an error running `code applicationDirectory` after using the autotools build because `/usr/share/Macaulay2/Core/startup.m2` doesn't exist:

```m2
/usr/share/Macaulay2/Core/debugging.m2:20:6:(1):[40]: error: couldn't find file /usr/share/Macaulay2/Core/startup.m2
/usr/share/Macaulay2/Core/code.m2:15:53:(1):[39]: --back trace--
/usr/share/Macaulay2/Core/code.m2:5:33:(1):[38]: --back trace--
/usr/share/Macaulay2/Core/code.m2:47:16:(1):[36]: --back trace--
/usr/share/Macaulay2/Core/code.m2:64:23:(1):[35]: --back trace--
/usr/share/Macaulay2/Core/document.m2:672:41:(1):[33]: --back trace--
/usr/share/Macaulay2/Core/document.m2:668:40:(1):[32]: --back trace--
/usr/share/Macaulay2/Core/document.m2:1010:44:(1):[32]: --back trace--
/usr/share/Macaulay2/Core/document.m2:1010:6:(1):[31]: --back trace--
/usr/share/Macaulay2/Core/methods.m2:119:80:(1):[30]: --back trace--
/usr/share/Macaulay2/Macaulay2Doc/doc13.m2:224:1:(2):[28]: --back trace--
...
```
The fix should be to just install it, which I'm assuming it's already doing in the CMake build.  I have a patch that I think fixes it, but I need to test it before I push it.

d-torrance:
It works!  Just pushed it.

PR #1398, opened 2020-07-14
Feature: package review workflow

mahrud:
The added action and workflow allow running the tests and examples from any package from either web interface, like this:
image
or a `POST` request like this:
```
curl \
  -u $owner:$token -X POST \
  -H "Accept: application/vnd.github.v3+json" \
  -d "{\"inputs\":{\"package\":\"$package\"}, \"ref\":\"$ref\"}" \
  https://api.github.com/repos/$owner/M2/actions/workflows/1877698/dispatches
```

### DIFF
--- a/.github/actions/package-review/action.yml
+++ b/.github/actions/package-review/action.yml
@@ -1,0 +1,29 @@
+# Relevant documentation:
+# https://docs.github.com/en/actions/creating-actions/creating-a-docker-container-action
+# https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions
+
+name: Package Review Action
+description: 'Run all examples and tests in a package'
+
+inputs:
+  package:
+    description: 'Package name'
+    required: true
+    default: 'FirstPackage'
+outputs:
+  numExampleErrors:
+    description: 'Number of errors occured running examples'
+  numTestErrors:
+    description: 'Number of errors occured running tests'
+runs:
+  using: 'docker'
+  image: 'docker://mahrud/macaulay2:v1.16.0.1'
+  entrypoint: 'M2'
+  args: ["-q", "--stop", "--silent", "--srcdir", "/github/workspace/M2",
+         "-e", "debug Core;
+           pkg = installPackage(\"${{ inputs.package }}\",
+               IgnoreExampleErrors => true, InstallPrefix => \"/github/workspace/\");
+           stderr << \"::set-output name=numExampleErrors::\" << numErrors << endl;
+           check pkg;
+           stderr << \"::set-output name=numTestErrors::\" << numErrors << endl;
+           exit 0"]
--- a/M2/BUILD/mahrud/package-review.sh
+++ b/M2/BUILD/mahrud/package-review.sh
@@ -1,0 +1,26 @@
+owner=mahrud
+token=`cat ~/.github_token` # https://github.com/settings/tokens
+
+# List workflows:
+#curl -H "Accept: application/vnd.github.v3+json" https://api.github.com/repos/$owner/M2/actions/workflows
+
+if [[ -z "$1" ]];
+then package="FirstPackage"
+else package=$1
+fi
+
+shift
+
+if [[ -z "$1" ]];
+then ref="refs/master"
+else ref=$1
+fi
+
+echo package: $package
+echo     ref: $ref
+
+curl \
+  -u $owner:$token -X POST \
+  -H "Accept: application/vnd.github.v3+json" \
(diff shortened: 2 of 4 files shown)


mahrud:
Using the `M2/BUILD/mahrud/package-review.sh` script added, you can do this like:
```
sh package-review.sh LocalRings refs/master
```
Note that you need to set your username and get a Github token with `repo` permissions

mahrud:
Here's an example of the triggered action:
https://github.com/mahrud/M2/runs/868553980?check_suite_focus=true
image
Note that it took 49 seconds as opposed to hours.

M2/Macaulay2/m2/run.m2
```diff
@@ -111,7 +111,7 @@ runFile = (inf, inputhash, outf, tmpf, desc, pkg, announcechange, usermode, exam
      cmd = cmd | readmode(ArgSilent,      "--silent");
      cmd = cmd | readmode(ArgStop,        "--stop");
      cmd = cmd | readmode(ArgPrintWidth,  "--print-width 77");
-     cmd = cmd | concatenate apply(srcdirs, d -> (" --srcdir",format d));
+     cmd = cmd | concatenate apply(srcdirs, d -> (" --srcdir ", format d));
```
DanGrayson, line 114: That space needs to be there.

DanGrayson: ..., so good change.

.github/actions/package-review/action.yml
```diff
@@ -0,0 +1,29 @@
+# Relevant documentation:
```
DanGrayson, line 1: What's the difference between an "action" and a "workflow"?

.github/actions/package-review/action.yml
```diff
@@ -0,0 +1,29 @@
...
+    required: true
+    default: 'FirstPackage'
+outputs:
+  numExampleErrors:
+    description: 'Number of errors occured running examples'
+  numTestErrors:
+    description: 'Number of errors occured running tests'
+runs:
+  using: 'docker'
+  image: 'docker://mahrud/macaulay2:v1.16.0.1'
```
DanGrayson, line 20: By the way, if this turns out to be used often, we'll want a "Macaulay2" docker account.

M2/BUILD/mahrud/package-review.sh
```diff
@@ -0,0 +1,26 @@
...
+fi
+
+echo package: $package
+echo     ref: $ref
+
+curl \
+  -u $owner:$token -X POST \
+  -H "Accept: application/vnd.github.v3+json" \
+  -d "{\"inputs\":{\"package\":\"$package\"}, \"ref\":\"$ref\"}" \
+  https://api.github.com/repos/$owner/M2/actions/workflows/1877698/dispatches
```
DanGrayson, line 26: What's the explanation of this URL?

.github/actions/package-review/action.yml
mahrud: https://docs.github.com/en/actions/getting-started-with-github-actions/core-concepts-for-github-actions#action

M2/BUILD/mahrud/package-review.sh
mahrud: It's the API gateway for triggering a workflow dispatch event: https://docs.github.com/en/rest/reference/actions#create-a-workflow-dispatch-event

.github/actions/package-review/action.yml
mahrud: It might already be [taken](https://hub.docker.com/u/macaulay2), unless @mikestillman owns that. Regardless, github has its own docker registry. I just didn't use it because it requires an authentication step.

mahrud:
Note: you can actually run package tests from a given commit or git reference, so this action can be used as a check for PRs that touch packages, so they don't take many hours.

.github/actions/package-review/action.yml
DanGrayson: Or could it be Lars Kastner?  We should find out.

mahrud:
@DanGrayson this PR is ready.

M2/BUILD/mahrud/package-review.sh
DanGrayson: Where did the number 1877698 come from?

mahrud: That's the workflow id. The commented out command a few lines above lists available workflow ids.

This script is for advanced use, hence why it's in this directory and not advertised. The web interface is the intended way to trigger this workflow.

PR #1404, opened 2020-07-15
exclude some builds

DanGrayson:
until we can figure out how to make them work


DanGrayson:
Two builds are still failing.  I may have to disable those, too.

PR #1406, opened 2020-07-15
fix "setup()" function

### LINKED ISSUE #1139: "setup()" should deal with "zsh"
"zsh" doesn't read ".profile", so "setup()" doesn't result in adding anything to the PATH if "zsh" is your shell.

That shell reads "$ZDOTDIR/.zprofile", but the manual page gives no hint about the value of "ZDOTDIR"

DanGrayson:
resolves #1139

fixes bug noticed by Mahrud, see https://github.com/Macaulay2/M2/commit/4a66fd902328c1642298254215014d82bcb49caf#r40507676

### DIFF
--- a/M2/Macaulay2/m2/files.m2
+++ b/M2/Macaulay2/m2/files.m2
@@ -322,7 +322,7 @@ mungeFile = (filename, headerline, trailerline, text) -> (
      insert := headerline | text | trailerline;
      local action;
      if fileExists filename then (
-     	  filename = realpath filename;	-- no other editor does this, but it seems like a good idea...
+     	  filename = realpath filename;	-- if filename is a symbolic link, we want to preserve that link and modify the underlying file
 	  hdr := "^" | regexpString headerline;
 	  tlr := "^" | regexpString trailerline;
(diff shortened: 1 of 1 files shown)


mahrud:
Does this apply the zsh changes even if .bash_profile is found? I would have thought in "A or B" if A is evaluated to true, B is not evaluated

DanGrayson:
That's true, but the only reason for "mungeFile" to return false is if the user says to quit.

mahrud:
so .. then this would not apply the changes to zsh if .bash_profile is found and the user doesn't quit, right?

mahrud:
In fact, we need the first one to return false for the zsh one to go through.
```m2
i2 : (print "bash"; true) or (print "zsh"; true)
bash

o2 = true
```

I don't think this resolves the issue. Probably the `setup("bash")` vs. `setup("zsh")` solution is the most straightforward way to go.

DanGrayson:
> so .. then this would not apply the changes to zsh if .bash_profile is found and the user doesn't quit, right?

oops, I said it the wrong way around: "true" means "quit"

PR #1407, opened 2020-07-15
Include error output in message when examples have errors

d-torrance:
This will make build logs more useful, especially when only the log and no
artifacts are available.

### DIFF
--- a/M2/Macaulay2/m2/installPackage.m2
+++ b/M2/Macaulay2/m2/installPackage.m2
@@ -664,7 +664,16 @@ installPackage Package := opts -> pkg -> (
 		    ));
 
 	  if not opts.IgnoreExampleErrors
-	  then if hadError then error(toString numErrors, " error(s) occurred running examples for package ", pkg#"pkgname");
+	  then if hadError then error(toString numErrors,
+	       " error(s) occurred running examples for package ",
+	       pkg#"pkgname",
+	       if verbose then ":" | newline | newline |
+	            concatenate apply(select(readDirectory(exampleOutputDir),
+		         file -> match("\\.errors$", file)), err ->
+		              err | newline |
+			      concatenate(width err : "*") | newline |
+			      get("!tail " | exampleOutputDir | err))
+	       else "");
 
 	  -- if no examples were generated, then remove the directory
 	  if length readDirectory exampleOutputDir == 2 then


d-torrance:
Here's an example from a [failing build of an Ubuntu package of the `development` branch](https://launchpadlibrarian.net/488677476/buildlog_ubuntu-groovy-amd64.macaulay2_16.0.1+git11929.74c3aa6-0ppa11681~ubuntu20.10.1_BUILDING.txt.gz) I wrote this patch to try and diagnose:

```
/<<BUILDDIR>>/macaulay2-16.0.1+git11929.74c3aa6/M2/usr-dist/common/share/doc/Macaulay2/Nauty/example-output/_relabel__Bipartite.errors:0:1: (output file) error: Macaulay2 exited with status code 1
/tmp/M2-367-0/0_relabel__Bipartite.m2:0:1: (input file)
M2: *** Error 1
currentString:1:1:(3):[2]: error: 1 error(s) occurred running examples for package Nauty:

_relabel__Bipartite.errors
**************************
-- -*- M2-comint -*- hash: 284388156

i1 : R = QQ[a..f];

i2 : G = graph flatten apply({a,c,e}, v->v*{b,d,f})

o2 = Graph{edges => {{a, b}, {b, c}, {a, d}, {c, d}, {b, e}, {d, e}, {a, f}, {c, f}, {e, f}}}
           ring => R
           vertices => {a, b, c, d, e, f}

o2 : Graph

i3 : relabelBipartite G
stdio:3:1:(3): error: callNauty: nauty terminated with the error [>E Sorry, this program doesn't support digraphs yet.].

make[3]: *** [Makefile:92: /<<BUILDDIR>>/macaulay2-16.0.1+git11929.74c3aa6/M2/usr-dist/x86_64-Linux-Ubuntu-20.10/lib/x86_64-linux-gnu/Macaulay2/Nauty/.installed] Error 1
```

DanGrayson:
Good idea.

Github seems to be having problems today ...

mahrud:
I think this would make developing packages harder, since if there are a few example errors the output would be very lengthy. We can just upload the artifacts instead: https://github.com/Macaulay2/M2/pull/1379/files#diff-b844ab5da048a39243d76cae5b7d8153R208-R209

d-torrance:
Maybe add an option to `installPackage` for whether this output is displayed that defaults to `false` for package developers, but is set to `true` when building Macaulay2 itself?

d-torrance:
If we have example errors, then the build logs would still be shorter overall since we wouldn't get as far as if the build were successful.

mahrud:
> If we have example errors, then the build logs would still be shorter overall since we wouldn't get as far as if the build were successful.

I don't think so. The same number of tests are run and the same error is returned; only the error message includes the content of *all* example errors, which can be huge.

d-torrance:
At least in the autotools build, the build errors out as soon as one package has example errors.  So we'd be getting at most one package's worth of example errors, and nothing else that would usually come after it (building the rest of the packages' examples, running tests, etc.)

mahrud:
I'm only talking about what this PR changes. There is no limit to the size of a test, and the error might occur 20 lines into the test. Or there could be 20 tests failing from one package. This would make the log longer. I'm not opposed to it, just saying it shouldn't be the default.

d-torrance:
Yeah, that is a good point.  It could print, say, the last 10 lines of the output, mimicking `tail _foo.errors`, instead.

d-torrance:
Alrighty, I've pushed a new version of this commit.  It only prints the error message if the local variable `verbose` (which is `true` when the `Verbose` option is `true` or `debugLevel > 0`) is `true` and uses `tail` to likely only print pertinent parts of the output.

d-torrance:
Here's an example of a package with a long example that errors out (30 lines and then an error), but it's not so bad:

```m2
i7 : installPackage("Foo")
--installing package Foo in /home/profzoom/.Macaulay2/local/ with layout 1
--using package sources found in /home/profzoom/.Macaulay2/code/
--storing raw documentation in /home/profzoom/.Macaulay2/local/lib/Macaulay2/Foo/cache/rawdocumentation-dcba-8.db
--new raw documentation, not already in database, for foo
--closed the database
--running tests that are functions
--making example result files in /home/profzoom/.Macaulay2/local/share/doc/Macaulay2/Foo/example-output/
--making example results for foo in file /home/profzoom/.Macaulay2/local/share/doc/Macaulay2/Foo/example-output/_foo.out
 ulimit -c unlimited; ulimit -t 700; ulimit -m 850000; ulimit -s 8192; ulimit -n 512;  cd /tmp/M2-26944-0/11-rundir/; GC_MAXIMUM_HEAP_SIZE=400M "/usr/bin/M2-binary" --int --no-randomize --no-readline --silent --stop --print-width 77 -e 'needsPackage("Foo", Reload => true, FileName => "/home/profzoom/.Macaulay2/code/Foo.m2")' <"/tmp/M2-26944-0/10_foo.m2" >>"/home/profzoom/.Macaulay2/local/share/doc/Macaulay2/Foo/example-output/_foo.errors" 2>&1
 ulimit -c unlimited; ulimit -t 700; ulimit -m 850000; ulimit -s 8192; ulimit -n 512;  cd /tmp/M2-26944-0/11-rundir/; GC_MAXIMUM_HEAP_SIZE=400M "/usr/bin/M2-binary" --int --no-randomize --no-readline --silent --stop --print-width 77 -e 'needsPackage("Foo", Reload => true, FileName => "/home/profzoom/.Macaulay2/code/Foo.m2")' <"/tmp/M2-26944-0/10_foo.m2" >>"/home/profzoom/.Macaulay2/local/share/doc/Macaulay2/Foo/example-output/_foo.errors" 2>&1
/home/profzoom/.Macaulay2/local/share/doc/Macaulay2/Foo/example-output/_foo.errors:0:1: (output file) error: Macaulay2 exited with status code 1
/tmp/M2-26944-0/10_foo.m2:0:1: (input file)
M2: *** Error 1
stdio:7:1:(3): error: 1 error(s) occurred running examples for package Foo:

_foo.errors
***********
i29 : 28

o29 = 28

i30 : 29

o30 = 29

i31 : error("foo")
stdio:31:1:(3): error: foo
```

Issue #1330, opened 2020-06-23
RunExternalM2 tests failing (sometimes)


d-torrance:
I just got the following while running `make check`:

```shell
--making test results
if ! grep "CacheExampleOutput => true" ./RunExternalM2.m2 >/dev/null ; then /build/macaulay2-1.15.1.0+git575.6c49cc1+ds/M2/usr-dist/x86_64-Linux-Debian-unknown/bin/M2 -q --stop --silent -e "needsPackage(\"RunExternalM2\",LoadDocumentation=>true,DebuggingMode=>true); check(RunExternalM2,UserMode=>false); exit 0" ; fi 
--running test 0 of package RunExternalM2 on line 909 in file ./RunExternalM2.m2
--    rerun with: check_0 "RunExternalM2"
--making test results
 ulimit -c unlimited; ulimit -t 700; ulimit -m 850000; ulimit -s 8192; ulimit -n 512;  cd /tmp/M2-3914-0/1-rundir/; GC_MAXIMUM_HEAP_SIZE=400M "/build/macaulay2-1.15.1.0+git575.6c49cc1+ds/M2/usr-dist/x86_64-Linux-Debian-unknown/bin/M2-binary" -q --int --no-randomize --no-readline --silent --stop --print-width 0 -e 'needsPackage("RunExternalM2", Reload => true, FileName => "/build/macaulay2-1.15.1.0+git575.6c49cc1+ds/M2/Macaulay2/packages/RunExternalM2.m2")' <"/tmp/M2-3914-0/0.m2" >>"/tmp/M2-3914-0/0.tmp" 2>&1
/tmp/M2-3914-0/0.tmp:0:1: (output file) error: Macaulay2 exited with status code 1
/tmp/M2-3914-0/0.m2:0:1: (input file)
M2: *** Error 1

[snip]

../m2/debugging.m2:20:6:(1):[6]: error: 5 error(s) occurred running tests for package RunExternalM2
../m2/installPackage.m2:565:30:(1):[5]: --back trace--
../m2/methods.m2:119:80:(1):[4]: --back trace--
../m2/option.m2:16:8:(1):[3]: --back trace--
currentString:1:76:(3):[2]: --back trace--
../m2/startup.m2.in:557:33:(0):[1]: --back trace--
../m2/startup.m2.in:668:6:(0): --back trace--
make[4]: *** [Makefile:39: check-RunExternalM2] Error 1
```

[...]

PR #1410, opened 2020-07-15
Call external M2 with -q in RunExternalM2 if parent M2 was

d-torrance:
This prevents errors if we don't have permission to write to the home
directory.

Closes: #1330

### DIFF
--- a/M2/Macaulay2/packages/RunExternalM2.m2
+++ b/M2/Macaulay2/packages/RunExternalM2.m2
@@ -85,7 +85,6 @@ export {
 	"runExternalM2ReturnAnswer",
 	-- Various Options:
 	"M2Location",
-	"KeepFiles",
 	"KeepStatistics",
 	"KeepStatisticsCommand",
 	"PreRunScript"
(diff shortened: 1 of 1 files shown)


d-torrance:
Apparently the fact that I added `debug Core` and we now have a `KeepFiles` symbol there was resulting in build errors:
```m2
../../../../../Macaulay2/packages/RunExternalM2.m2:895:1:(2):[47]: error: item to be documented comes from another package: Macaulay2Doc :: KeepFiles
```

I've pushed a new commit which should fix this.

DanGrayson:
@bapike --- do these changes look good to you, as author of the package?

bapike:
This seems like a great solution!  Thank you @d-torrance .

If running with a non-writable home directory is useful, maybe 'M2 --help' should say that '-q' allows that.

Just for some background:
I think 'M2Options' here did not include -q (but includes the rest of --script) so that a child M2 process can find locally installed packages.  This was especially important when I was developing and testing RunExternalM2, since I would install it locally and the child must be able to find RunExternalM2.  It's less vital now that RunExternalM2 is installed with Macaulay2, but I like the idea of continuing to normally omit -q so that users could use RunExternalM2 with locally installed packages.  And if the parent M2 is run with '-q' then users should never expect to use locally installed packages (in the parent or the child), so I think this fix does the right thing in all cases.

PR #1400, opened 2020-07-14
Improving support for external programs, part 3

d-torrance:
Some distributions may install programs in a non-standard location, e.g., Fedora installs the various executables that come with 4ti2 in `/usr/lib/4ti2/bin` or `/usr/lib64/4ti2/bin`, depending on the architecture.

We add a `Path` option to `findProgram` which allows the user to specify additional paths to search to find a program in situations like this.

As a proof of concept, the `FourTiTwo` package has been ported to the new `findProgram`/`runProgram` interface.

PR #1413, opened 2020-07-16
Improving support for external programs, part 4

d-torrance:
This PR has two main parts:

* Use `findProgram`/`runProgram` in the two nauty packages, `Nauty` and `NautyGraphs`.
* Now that the appropriate packages can find gfan, topcom, and pending #1400 and this PR, 4ti2 and nauty, out of the box, we modify the `configure` script to allow the autotools build to use pre-installed versions of these programs.  For example:

```
checking whether the package 4ti2 is installed... yes, using prefix "4ti2-"
checking whether the package cohomcalg is installed... yes, will build anyway
checking whether the package gfan is installed... yes
checking whether the package lrs is installed... yes
checking whether the package csdp is installed... yes, will build anyway
checking whether the package nauty is installed... yes, using prefix "nauty-"
checking whether the package topcom is installed... yes, using prefix "topcom-"

...

configure: using BUILDLIBLIST  =  flint mpsolve gtest
configure: using BUILDSUBLIST  = 
configure: using BUILDPROGLIST =  normaliz csdp cohomcalg
configure: using BUILDLIST     =  flint mpsolve gtest normaliz csdp cohomcalg
configure: using BUILD_ALWAYS  =  normaliz
configure: using FILE_PREREQS  =  /usr/bin/4ti2-circuits /usr/bin/gfan /usr/bin/lrs /usr/bin/nauty-complg /usr/bin/topcom-points2finetriangs
```

### DIFF
--- a/M2/Macaulay2/packages/Nauty.m2
+++ b/M2/Macaulay2/packages/Nauty.m2
@@ -36,10 +36,9 @@ newPackage select((
 -- Configuration
 -------------------
 
--- Check the ~/.Macaulay2/init-Nauty.m2 file for the absolute path.
--- If it's not there, then use the M2-Programs directory.
-nauty'path = (options Nauty).Configuration#"path";
-if nauty'path == "" then nauty'path = prefixDirectory | currentLayout#"programs";
+-- for backward compatibility
+if not programPaths#?"nauty" and Nauty#Options#Configuration#"path" != ""
+    then programPaths#"nauty" = Nauty#Options#Configuration#"path"
 
 -------------------
 -- Exports
@@ -458,6 +457,8 @@ stringToGraph (String, PolynomialRing) := Graph => (str, R) -> graph stringToEdg
 -- Local-Only Code
 -------------------
 
+nauty = null
+
 -- Sends a command and retrieves the results into a list of lines.
 -- If ReadError is set to true and the command is successfully executed,
 -- then the data from stderr is returned (filterGraphs and removeIsomorphs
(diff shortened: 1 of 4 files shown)


DanGrayson:
Some builds have failed.

d-torrance:
Yeah:

```
../../../../../Macaulay2/tests/normal/schenck-book-3.m2:4:1:(3):[4]: error: assertion failed
```

The corresponding line is:
```m2
assert( (gfan(I)) === {new MarkedPolynomialList from {{y^7,x},{y^7-1,-y^3+x}},new MarkedPolynomialList from {{x^3,x^2*y,y^3},{x^3-y^2,x^2*y-1,y^3-x}},new MarkedPolynomialList from {{x^5,x^2*y,y^2},{x^5-y,x^2*y-1,-x^3+y^2}},new MarkedPolynomialList from {{x^7,y},{x^7-1,-x^5+y}}} );
```

Github is installing the Ubuntu gfan package, but it's older since it's running 18.04:

```
Preparing to unpack .../027-gfan_0.5+dfsg-6_amd64.deb ...
Unpacking gfan (0.5+dfsg-6) ...
```

I'll investigate the difference in the output when using 0.5 v 0.6.2.  I'm guessing this is the opposite to #1173, where a test was written for an older gfan and we were running it on a newer  one.  Now we're trying to run a test written for a newer gfan with an older one.

d-torrance:
I'm going to remove the gfan-specific commit from this PR for now, as the rest should be ok.  We may need to still build gfan until `findPackage` can check minimum versions.

d-torrance:
Yup, that's the problem:

```m2
i1 : R = QQ[x, y];

i2 : I = ideal(x^7 - 1, y - x^5);

o2 : Ideal of R

i3 : loadPackage("gfanInterface"); -- gfan 0.6.2

i4 : gfan I

o4 = {{(y^7) -1, (x) -y^3}, {(x^3) -y^2, (x^2*y) -1, (y^3) -x}, {(x^5) -y,
     --------------------------------------------------------------------------
     (x^2*y) -1, (y^2) -x^3}, {(x^7) -1, (y) -x^5}}

o4 : List

i5 : loadPackage("gfanInterface", Configuration => {"path" => "/usr/bin"},
         Reload => true); -- gfan 0.5

i6 : gfan I

o6 = {{(y^7) -1, (x) -y^3}, {(y^3) -x, (x^2*y) -1, (x^3) -y^2}, {(y^2) -x^3,
     --------------------------------------------------------------------------
     (x^2*y) -1, (x^5) -y}, {(y) -x^5, (x^7) -1}}

o6 : List
```

DanGrayson:
Is one of those answers wrong?

d-torrance:
No, the ordering of each term has just been reversed.

DanGrayson:
If both answers are right, then why should we care which version of gfan is out there?

d-torrance:
We probably shouldn't.  Perhaps a `==` method for `MarkedPolynomialList` objects could identify when two of them are the same despite different orderings.

d-torrance:
Ok, these lastest commits should fix the gfan problem.  I've added the `==` method I mentioned and modified the failing test to use it instead of `===`.  (It seems like `===` should still only be `true` if the order is the same.)

M2/Macaulay2/packages/gfanInterface.m2
```diff
@@ -206,6 +206,9 @@ MarkedPolynomialList.synonym = "marked polynomial list";
   -- the monomial inL#i is the marked monomial, which
   -- should occur with the same coefficient in L#i.
 
+MarkedPolynomialList == MarkedPolynomialList := (L1, L2) ->
+	set transpose L1 === set transpose L2
+
```
DanGrayson, line 211: The word "List" in "MarkedPolynomialList" implies that for objects of this type, the order matters, so introducing an equality test like this into the package will result in cognitive dissonance.  Could you instead just fix the test, so it does the appropriate thing?

d-torrance: Makes sense -- will do!

d-torrance:
This is tagged as "waiting for update", but I believe the update that was being waited on (changing the test that was failing due to older versions of gfan instead of adding a `==` method for `MarkedPolynomialList`'s) has been addressed.

DanGrayson:
@dcookii -- are these changes to Nauty okay with you?

dcookii:
These are all good changes.  Thanks @d-torrance!

PR #1416, opened 2020-07-17
CMake maintenance

mahrud:
Includes:
- many important improvements to how CMake finds libraries
- add Clang10 test for ubuntu and macOS, but only for CMake
- add github action tests for CMake
- improve github action artifact uploads for both cmake and autotools to include example errors, if any occurred

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -91,13 +91,11 @@ jobs:
       - name: Install requirements for Linux
         if: runner.os == 'Linux'
         run: |
-          # TODO: add libgivaro-dev and fflas-ffpack once they are updated
           sudo apt-get install -y -q --no-install-recommends clang-10 gfortran libtool-bin ninja-build yasm ccache
           sudo apt-get install -y -q --no-install-recommends libatomic-ops-dev libboost-stacktrace-dev \
                   libncurses-dev libncurses5-dev libreadline-dev libeigen3-dev liblapack-dev libxml2-dev \
                   libgc-dev libgdbm-dev libglpk-dev libgmp3-dev libgtest-dev libmpfr-dev libntl-dev gfan \
-                  libcdd-dev w3c-markup-validator
-                  # libz-dev zlib1g-dev liblzma-dev fflas-ffpack libgivaro-dev
+                  libcdd-dev libgivaro-dev fflas-ffpack w3c-markup-validator
           # TODO: help autotools detect flint from brew
           brew install flint
 
@@ -131,11 +129,11 @@ jobs:
           path: |
             ~/.ccache
             ~/M2/BUILD/build/usr-host
-          key: usr-host-${{ runner.os }}-${{ matrix.compiler }}-${{ matrix.build-system }}-${{ hashFiles('**/cmake/*-libraries.cmake') }}
+          key: build-cache-${{ runner.os }}-${{ matrix.compiler }}-${{ matrix.build-system }}-${{ hashFiles('**/cmake/*-libraries.cmake') }}
           restore-keys: |
-            usr-host-${{ runner.os }}-${{ matrix.compiler }}-${{ matrix.build-system }}-
-            usr-host-${{ runner.os }}-${{ matrix.compiler }}-
-            usr-host-${{ runner.os }}-
+            build-cache-${{ runner.os }}-${{ matrix.compiler }}-${{ matrix.build-system }}-
+            build-cache-${{ runner.os }}-${{ matrix.compiler }}-
+            build-cache-${{ runner.os }}-
 
 
 # ----------------------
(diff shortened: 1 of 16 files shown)


mahrud:
I also had to remove givaro and fflas-ffpack from the installed packages, since they include a problematic `-fabi-version=6` flag that causes problems with CMake. This forces us to build givaro and fflas-ffpack, and there seems to be a test failing with givaro.

mahrud:
@DanGrayson the two failed builds are caused by a failing givaro test [here](https://github.com/Macaulay2/M2/pull/1416/checks?check_run_id=880869264#step:13:12986):
```
PASS: test-ruint_square
============================================================================
Testsuite summary for Givaro 4.1.1
============================================================================
# TOTAL: 67
# PASS:  66
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0
```
Since the build itself is passing and the tests step is failing, can we finally merge this?

DanGrayson:
> @DanGrayson the two failed builds are caused by a failing givaro test [here](https://github.com/Macaulay2/M2/pull/1416/checks?check_run_id=880869264#step:13:12986):
> 
> ```
> PASS: test-ruint_square
> ============================================================================
> Testsuite summary for Givaro 4.1.1
> ============================================================================
> # TOTAL: 67
> # PASS:  66
> # SKIP:  0
> # XFAIL: 0
> # FAIL:  1
> # XPASS: 0
> # ERROR: 0
> ```
> 
> Since the build itself is passing and the tests step is failing, can we finally merge this?
> Config logs are available for both of those if you'd like to debug them.

I don't want to have builds failing every time, for otherwise, why run them every time?  Let's just disable the testing of givaro.

mahrud:
I installed fflas_ffpack and givaro and just forced cmake to build them. This should cause autotools not to test those.

mikestillman:
@mahrud why would that cause autotools not to test those?

> I installed fflas_ffpack and givaro and just forced cmake to build them. This should cause autotools not to test those.

DanGrayson:
> @mahrud why would that cause autotools not to test those?
> 
> > I installed fflas_ffpack and givaro and just forced cmake to build them. This should cause autotools not to test those.

We test only those libraries that we build.  (Testing requires the source code.)

.github/workflows/test_build.yml
```diff
@@ -25,143 +25,188 @@ on:
...
+          make check
+          make -C Macaulay2/html-check-links check
+          make validate-html
+
+# ----------------------
+#   Upload build logs
+# ----------------------
+
+      - name: Upload build logs
+        if: failure()
```
DanGrayson, line 199: We probably want the build logs always, not just after a failure.  For example, what if we want to determine why a configure script didn't detect a certain package?

DanGrayson:
The 1 failure is due to a crash while running ThreadedGB examples:
```
i8 : tgb({f0,f1},2,Verbose=>true)
M2-binary: tpp.c:84: __pthread_tpp_change_priority: Assertion `new_prio == -1 || (new_prio >= fifo_min_prio && new_prio <= fifo_max_prio)' failed.
Aborted (core dumped)
```

mahrud:
This has been happening at random since ThreadedGB was merged. I haven't
been able to reproduce it yet, but I'm certain that it's not related to
changes in this PR (e.g that build worked last time!)

mahrud:
For that purpose, I think we should add tests that fail if a certain
library isn't found, so that we notice and get the logs.

You can also remove the "failure ()" condition in a PR to get all the logs.

Alternatively, I can add a manual dispatch trigger for this workflow that
keeps logs for everything. Would that work?

If not, feel free to remove the condition and merge if that's the only
remaining issue.

DanGrayson:
Rerunning ...

DanGrayson:
> For that purpose, I think we should add tests that fail if a certain library isn't found, so that we notice and get the logs. You can also remove the "failure ()" condition in a PR to get all the logs. Alternatively, I can add a manual dispatch trigger for this workflow that keeps logs for everything. Would that work? If not, feel free to remove the condition and merge if that's the only remaining issue

I think we should just change `if: failure()` to `if: always()` -- that's the way I currently have it set up, so we always get the logs.

mahrud:
I'm fairly sure this is the very first, all green build.

> I think we should just change `if: failure()` to `if: always()` -- that's the way I currently have it set up, so we always get the logs.

@DanGrayson I don't have access to my laptop this weekend. Could you push the changes you'd like and merge this? As always, I leave pushes from maintainers to my PRs open.

DanGrayson:
That's worth commemorating:
Screen Shot 2020-07-18 at 9 18 14 AM

DanGrayson:
Not quite the first one:
Screen Shot 2020-07-18 at 9 28 39 AM

mikestillman:
@mahrud , @DanGrayson Good job!

PR #1417, opened 2020-07-19
Improving support for external programs, part 5

d-torrance:
*I'm leaving this as a draft pull request pending #1400 and #1413, as they touch the same files and there will need to be some fixing up to do before this is merged.  But it's ready aside from that, so I figured I'd open it for review.*

This PR introduces the following:

* A new `versionCompare` method for comparing version strings.
* A `MinimumVersion` option to `findProgram`, which uses the above to determine whether the user has a new enough version of a program on their system.
* A `RunDirectory` option to `runProgram`, which gives the user the option of running their program in a directory other than the current working directory.
* Updates to the `Normaliz` package.  It uses the above options, as it requires at least version 2.11 of normaliz and needs to run in the same directory as its input and output files.

### DIFF
--- a/M2/Macaulay2/m2/programs.m2
+++ b/M2/Macaulay2/m2/programs.m2
@@ -10,14 +10,24 @@ addSlash = programPath -> (
     else return programPath
 )
 
-checkProgramPath = (name, cmds, opts) -> (
-    if all(cmds, cmd -> run(cmd | " >/dev/null 2>&1") == 0) then (
-	if opts.Verbose == true then print("    found");
-	return true;
-    ) else (
-	if opts.Verbose == true then print("    not found");
-	return false;
-    )
+checkProgramPath = (name, cmds, pathToTry, prefix, opts) -> (
+    found := all(cmds, cmd ->
+	run(pathToTry | addPrefix(cmd, prefix) | " >/dev/null 2>&1") == 0);
+    msg := "";
+    if found then (
+	if opts.MinimumVersion === null then msg = "    found"
+	else (
+	    thisVersion := get("!" | pathToTry |
+		addPrefix(opts.MinimumVersion_1, prefix));
+	    found = found and thisVersion >= opts.MinimumVersion_0;
+	    if found then msg = "    found version " | thisVersion | " >= " |
+		opts.MinimumVersion_0
+	    else msg = "   found, but version " | thisVersion | " < " |
+		opts.MinimumVersion_0;
+	)
+    ) else msg = "    not found";
+    if opts.Verbose == true then print(msg);
+    found
 )
 
 addPrefix = (cmd, prefix) ->
@@ -44,8 +54,8 @@ getProgramPath = (name, cmds, opts) -> (
 	    if opts.Verbose == true and #prefixes > 1 then
 		print("  trying prefix \"" | prefix_1 |
 		    "\" for executables matching \"" | prefix_0 | "\"...");
-	    if checkProgramPath(name, apply(cmds, cmd ->
-		pathToTry | addPrefix(cmd, prefix)), opts) then break prefix)
+	    if checkProgramPath(name, cmds, pathToTry, prefix, opts) then
+	        break prefix)
 	);
 	if prefix =!= null then break (pathToTry, prefix)
     ))
@@ -56,7 +66,8 @@ findProgram = method(TypicalValue => Program,
 	RaiseError => true,
 	Verbose => false,
 	Prefix => {},
-	AdditionalPaths => {}
+	AdditionalPaths => {},
+	MinimumVersion => null
     })
 findProgram(String, String) := opts -> (name, cmd) ->
     findProgram(name, {cmd}, opts)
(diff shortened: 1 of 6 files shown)


DanGrayson:
What's the point of "versionCompare"? Doesn't the usual string comparison work for you?

d-torrance:
> What's the point of "versionCompare"? Doesn't the usual string comparison work for you?

I honestly didn't even think of that because string comparison won't work on other platforms, e.g., in Python:

```python
>>> '1.9' < '1.10'
False
```
But it does the right thing in Macaulay2, which I didn't try until your comment:
```m2
i14 : "1.9" < "1.10"

o14 = true
```
One nice thing about using the algorithm from `apt` is we can insert `~` if we want to sort an alpha, beta, etc. release lower:

```m2
i19 : versionCompare("1.0~beta", "1.0")

o19 = -1

i20 : "1.0~beta" < "1.0"

o20 = false

i21 : versionCompare("1.0~beta", "1.0") < 0

o21 = true
```
But of course, programs don't usually have `~` already in there, and the user would need to worry about using regular expressions to insert them before sending the strings to `versionCompare`.  

As far as I know, this `~` convention is only used in Debian.  Maybe there's a reason no one else seems to have adopted it!"

I think the `~` might be the only argument to use this algorithm instead of string comparison, but it's not super important since I don't know how many of the programs we use bother with things like alpha and beta releases.  I'm ok with dropping this commit if you'd prefer.

DanGrayson:
Well, I purposely made string comparison do version number comparison, so it would be applicable here.  It would be great if you would use it.

d-torrance:
Alrighty, I've removed the commit that added `versionCompare` and switched the `MinimumVersion` check to use string comparison.

d-torrance:
Just pushed a new version of this PR which bumps the minimum version of normaliz to 3.5.2, closing #1439.

d-torrance:
I went ahead and removed the draft status from this PR.  There will be some conflicts with some of my other outstanding PR's, but I'll just fix whichever ones need fixing if/when they are merged.

M2/Macaulay2/m2/programs.m2
```diff
@@ -42,15 +50,26 @@ getProgramPath = (name, cmds, opts) -> (
 	    if opts.Verbose == true and #prefixes > 1 then
 		print("  trying prefix \"" | prefix_1 |
 		    "\" for executables matching \"" | prefix_0 | "\"...");
+	    newOpts := if opts.MinimumVersion =!= null then
+		new OptionTable from apply(keys opts, key -> key =>
+		    if key === MinimumVersion then (opts#key_0,
+			pathToTry | addPrefix(opts#key_1, prefix))
+		    else opts#key)
```
DanGrayson, line 57: using `merge` here would be better

M2/Macaulay2/m2/programs.m2
```diff
@@ -42,15 +50,26 @@ getProgramPath = (name, cmds, opts) -> (
 	    if opts.Verbose == true and #prefixes > 1 then
 		print("  trying prefix \"" | prefix_1 |
 		    "\" for executables matching \"" | prefix_0 | "\"...");
+	    newOpts := if opts.MinimumVersion =!= null then
```
DanGrayson, line 53: Why not just assign the new value to `opts`?

DanGrayson:
@csoeger - are these changes to the Normaliz package okay with you?

M2/Macaulay2/m2/programs.m2
d-torrance: The new value contains the path and prefix information specific to the particular iteration we're working on, so if we use `opts`, then we'd lose the original value without that stuff.

For example, let's say we want to find the program `foo` with version at least 1.0.  And maybe some Linux distribution prepends all the various components of the `foo` program with `bar-`.  We might call something like:
```m2
findProgram("foo", "foo --help", Prefix => (".*", "bar-"), MinimumVersion => ("1.0", "foo --version"))
```

But to check the version number for a particular path/prefix combo, we want to insert that information into the `MinimumVersion` sequence.  So when checking, say, `/usr/local/bin` with the `bar-` prefix, we'd want `MinimumVersion` to be `("1.0", "/usr/local/bin/bar-foo --version")`.  That's what `newOpts` contains.  It doesn't remember what the original sequence in `opts` looks like, which is why I went with a new variable.

Looking at the code again, I think it may be simpler to just send the prefix and path information to the `checkProgramPath` function and let it do everything.

M2/Macaulay2/m2/programs.m2
d-torrance: Cool, I wasn't aware of `merge` -- I'll check it out.  It may be a moot point if I refactor `checkProgramPath` to deal with the path/prefix substitution as we wouldn't need to do this anyway.

d-torrance:
I've just force-pushed a new version of this branch with the `MinimumVersion` commit changed so that we don't have to deal with `newOpts`.  The path and prefix information are now passed to `checkProgramPath`, and it deals with inserting them.

d-torrance:
I've rebased this branch to fix a conflict that arose after #1447 was merged.

DanGrayson:
Now there are conflicts.  Could you merge?

DanGrayson:
@csoeger - are these changes to the Normaliz package okay with you?

d-torrance:
> Now there are conflicts. Could you merge?

Done

DanGrayson:
@csoeger - are these changes to the Normaliz package okay with you?

d-torrance:
Closing since #1513 was merged.

DanGrayson:
Closing, because #1513 was pulled.

DanGrayson:
Wow, you beat me to it by 0.5 seconds.

PR #1419, opened 2020-07-20
resolve #1418

### LINKED ISSUE #1418: problem with 3edc32e42063fece3f99b7209c13b971753a57e9
the recent commit 3edc32e42063fece3f99b7209c13b971753a57e9 breaks a hack used by http://habanero.math.cornell.edu:3690/ which is to redefine `open` for its own purposes. now I don't necessarily condone this hack but I do think the user should be given a choice as to which program should be launched in such circumstances; a simple fix would be to move the line https://github.com/Macaulay2/M2/blob/1b1df7a362b8ecb008c5397513c93c831d4bd647/M2/Macaulay2/m2/html.m2#L48 
two lines up as the first test of `browser()`?
incidentally I don't think https://github.com/Macaulay2/M2/blob/1b1df7a362b8ecb008c5397513c93c831d4bd647/M2/Macaulay2/m2/html.m2#L39
is correct.

pzinn:
seems everyone agrees on this simple fix.

### DIFF
--- a/M2/Macaulay2/m2/html.m2
+++ b/M2/Macaulay2/m2/html.m2
@@ -36,16 +36,16 @@ popIndentLevel  = (n, s) -> (indentLevel = indentLevel - n; s)
 -- whether fn exists on the path
 -- TODO: check executable
 runnable := fn -> (
-    if fn == "" then false;
+    if fn == "" then return false;
     if isAbsolutePath fn then fileExists fn
     else 0 < # select(1, apply(separate(":", getenv "PATH"), p -> p|"/"|fn), fileExists))
 
 -- preferred web browser
 -- TODO: cache this value
 browser := () -> (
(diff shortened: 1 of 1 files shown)


mahrud:
@d-torrance does ubuntu 18.04 have a working `xdg-open`? I'm wondering if we could check for `xdg-open` first, hoping to catch most linux varieties, and when that is not present, check for `open`. The main reason is that `open` is useful beyond macOS systems, and if this way we would remove the OS check, that would be great.

d-torrance:
> @d-torrance does ubuntu 18.04 have a working `xdg-open`?

Yup!  `xdg-utils`, the package containing `xdg-open`, has been in Debian and its derivatives since [2006](https://tracker.debian.org/news/541553/accepted-xdg-utils-0910beta1-1-source-all/).

PR #1428, opened 2020-07-23
Modify findProgram/runProgram examples to not need gfan/topcom installed

d-torrance:
We set RaiseError to false for the findProgram examples.  For the
runProgram examples (which necessarily need a successful call to
findProgram), we switch from gfan to perl, which should be present on
most systems.

### DIFF
--- a/M2/Macaulay2/packages/Macaulay2Doc/functions/findProgram-doc.m2
+++ b/M2/Macaulay2/packages/Macaulay2Doc/functions/findProgram-doc.m2
@@ -97,15 +97,15 @@ document {
     PARA {"Note that if a program consists of a single executable binary ",
 	"file, then ", TT "name", " should coincide with the name of this ",
 	"file."},
-    EXAMPLE lines ///
-	programPaths#"gfan" = "/path/to/gfan/"
-	gfan = findProgram("gfan", "gfan --help", Verbose => true)///,
+    EXAMPLE {///programPaths#"gfan" = "/path/to/gfan/"///,
+	///gfan = findProgram("gfan", "gfan --help", Verbose => true,
+	    RaiseError => false)///},
     PARA {"One program that is shipped with a variety of prefixes in ",
 	"different distributions and for which the ", TT "Prefix",
 	" option is useful is TOPCOM:"},
     EXAMPLE {///findProgram("topcom", "cube 3", Verbose => true, Prefix => {
     (".*", "topcom-"),
     ("^(cross|cube|cyclic|hypersimplex|lattice)$", "TOPCOM-"),
-    ("^cube$", "topcom_")})///},
+    ("^cube$", "topcom_")}, RaiseError => false)///},
     SeeAlso => {"Program", "programPaths", "runProgram"}
 }
--- a/M2/Macaulay2/packages/Macaulay2Doc/functions/runProgram-doc.m2
+++ b/M2/Macaulay2/packages/Macaulay2Doc/functions/runProgram-doc.m2
@@ -53,11 +53,10 @@ document {
     PARA {"This method runs an external program which has already been ",
 	"loaded using ", TO "findProgram", ".  The results of this run are ",
 	"available in a ", TO "ProgramRun", " object."},
-    EXAMPLE lines ///
-    	gfan = findProgram("gfan", "gfan --help")
-	runProgram(gfan, "_version")
-	oo#"output"
-	runProgram(gfan, "_foo", RaiseError => false)
-	oo#"error"///,
+    EXAMPLE {///ls = findProgram("ls", "ls .")///,
+	///runProgram(ls, "--version")///,
+	///oo#"output"///,
+	///runProgram(ls, "this-file-does-not-exist", RaiseError => false)///,
+	///oo#"error"///},
     SeeAlso => {"ProgramRun", "findProgram"}
 }


d-torrance:
This is in response to https://github.com/Macaulay2/M2/pull/1368#issuecomment-662795338.  I went with perl instead of python, as there's been some discussion about not shipping `/usr/bin/python` in the next Debian release, and instead only shipping `/usr/bin/python2` and `/usr/bin/python3` and letting the user decide if they want to symlink one of them.

DanGrayson:
> We set RaiseError to false for the findProgram examples.
> For the
> runProgram examples (which necessarily need a successful call to
> findProgram), we switch from gfan to perl, which should be present on
> most systems.

Why?  `gfan` and `topcom` will always be found, because we require them to be there.

DanGrayson:
Having an example fail if `perl` is not found is a good idea only if we add `perl` to the prerequisites of Macaulay2.  We don't check for its presence at configure time, and we don't have a good way to inform downstream distributers that it's a requirement -- they'll just get an error deep in the build process if `perl` is missing.

d-torrance:
> Why? `gfan` and `topcom` will always be found, because we require them to be there.

In https://github.com/Macaulay2/M2/pull/1368#issuecomment-662795338, Mahrud mentioned he'd like to be able to do some builds without needing them.

> Having an example fail if `perl` is not found is a good idea only if we add `perl` to the prerequisites of Macaulay2. We don't check for its presence at configure time, and we don't have a good way to inform downstream distributers that it's a requirement -- they'll just get an error deep in the build process if `perl` is missing.

I was under the impression that perl was always present, but I did a bit more googling after your comment, and apparently that's not the case for the BSD's (see https://unix.stackexchange.com/questions/568058/is-there-a-unix-like-system-that-doesnt-come-with-perl).  I could switch it to `awk` instead, which should always be available.

DanGrayson:
> > Why? `gfan` and `topcom` will always be found, because we require them to be there.
> 
> In [#1368 (comment)](https://github.com/Macaulay2/M2/pull/1368#issuecomment-662795338), Mahrud mentioned he'd like to be able to do some builds without needing them.

He actually said without building them, which seems to imply that they were found to exist on the system.  @marhud  -- can you clarify?

> 
> > Having an example fail if `perl` is not found is a good idea only if we add `perl` to the prerequisites of Macaulay2. We don't check for its presence at configure time, and we don't have a good way to inform downstream distributers that it's a requirement -- they'll just get an error deep in the build process if `perl` is missing.
> 
> I was under the impression that perl was always present, but I did a bit more googling after your comment, and apparently that's not the case for the BSD's (see https://unix.stackexchange.com/questions/568058/is-there-a-unix-like-system-that-doesnt-come-with-perl). I could switch it to `awk` instead, which should always be available.

Or better, `ls`, since our users will have heard of it.

d-torrance:
> Or better, `ls`, since our users will have heard of it.

Sounds good!  I've pushed a new version of the commit using `ls`.

mahrud:
This is great, thank you!

DanGrayson:
I still don't see the point of setting ` RaiseError => false `, since gfan will always be found.  We want the example output in the documentation to be the same on all systems as that would allow the text in the documentation to discuss the output.  Leaving that option out will also help ensure, as a check, that gfan is findable, which we want it to be.

mahrud:
@d-torrance why not look for M2 itself, since that will always be available?

> I still don't see the point of setting `RaiseError => false`, since gfan will always be found.

This isn't true for me.

> We want the example output in the documentation to be the same on all systems as that would allow the text in the documentation to discuss the output.

The path of some programs will inevitably be different on different systems, so this will never be true.

> Leaving that option out will also help ensure, as a check, that gfan is findable, which we want it to be.

gfan is findable for me when I use M2 because I set the path in my init file, but we install packages with -q, which causes the error. Besides, Macaulay2Doc isn't the right place to check this, gfanInterface is. All I want to do is install Macaulay2Doc without a hassle, because otherwise M2 runs slower and `help` takes 20 seconds.

DanGrayson:
> @d-torrance why not look for M2 itself, since that will always be available?
> 
> > I still don't see the point of setting `RaiseError => false`, since gfan will always be found.
> 
> This isn't true for me.

Why not?

> 
> > We want the example output in the documentation to be the same on all systems as that would allow the text in the documentation to discuss the output.
> 
> The path of some programs will inevitably be different on different systems, so this will never be true.

Well, sure, but gfan is a requirement

> 
> > Leaving that option out will also help ensure, as a check, that gfan is findable, which we want it to be.
> 
> gfan is findable for me when I use M2 because I set the path in my init file, but we install packages with -q, which causes the error. 

How is the -q option related to finding gfan?

> Besides, Macaulay2Doc isn't the right place to check this, gfanInterface is. All I want to do is install Macaulay2Doc without a hassle, because otherwise M2 runs slower and `help` takes 20 seconds.

I still don't get it.

mahrud:
Dan, I don't understand why this is such a contentious point for you. I'm describing a problem that affects me as someone both developing M2 and doing math with M2.

I'll reiterate: the location of gfan, 4ti2, etc. on my system is set in `~/.Macaulay2/init-gfanInterface.m2`. When we run `installPackage("Macaulay2Doc",...)`, we pass in `-q`:
https://github.com/Macaulay2/M2/blob/6e1ed058f91c1a4be70436e130254fc754ddefd2/M2/Macaulay2/packages/Makefile.in#L72
This causes my init file not to be read, and therefore gfan isn't found, which causes build to fail.

To emphasize, I don't want the build system to rebuild gfan for me, because it is already available when I need it.

If you don't consider my problem as a valid problem, you're welcome to close this PR or put a wontfix label on it, but I think that's the wrong approach in leading an open source project. I've even provided a safe alternative, which is to look for `M2` and run `M2 --version` instead, which would produce the same output on every system.

d-torrance:
I don't believe `/<builddir>/usr-dist/<multiarch-triplet>/bin/` is added to `PATH` when the examples are generated, so I don't think `M2` would be the best choice.  (We could add the appropriate directory to `prefixPaths` first, but I think that might give us another build path in the documentation to swap out...)  We could go with `ls` for `findProgram`, just like `runProgram`, for a nice simple program which is always present.

gfan was just an arbitrary choice (it was the first program I dealt with in this series of pull requests, so I used it...)  I do like using topcom  because it gives the most comprehensive example of the `Prefix` option, but we could always just refer users to the source code for the `Topcom` package instead.

DanGrayson:
(Yes, "M2" as an example for "findProgram" would not work, because it's not installed in a standard location.)

I don't understand the underlying issue yet.

Mahrud, I'll reiterate.  The program "gfan" is a build requirement, so it has to be available during the build, either because at the start it is found in one of the standard locations, or because we build it and put it in a standard location.  During the build, the -q option is always used, so any information in your file `~/.Macaulay2/init-gfanInterface.m2` about the location of "gfan" will not be available.  When the documentation for packages is built, it has to be there, in one of the standard locations, so the documentation examples will work.  So how is it possible that it is not there when we are running the examples for "findProgram"?

You have emphasized: "I don't want the build system to rebuild gfan for me, because it is already available when I need it."  I infer from that that you have installed gfan in a standard location where the build system can find it at the start, as a consequence of which it will determine that it doesn't have to build it.  And thus the examples for "findProgram" should be able to find it, too.

mahrud:
Ah, you're right about M2. I think the central issue here is that I see value in reducing unnecessary dependencies within a system, and since gfan is only a requirement for three packages `gfanInterface Tropical StronglyStableIdeals`, it seems to me like making Macaulay2Doc depend on gfan is unnecessary. What about "mgb" which comes with Mathicgb instead?

The CMake build, for instance, has a separate targets `build-programs` and `install-package`, and if you just want to build "M2-binary" to test "regex", as is the case currently for me, you don't need to build those extra targets. However, the command `help` stops working normally if `Macaulay2Doc` isn't installed (which is a separate issue #1085), so I install that too, but now `Macaulay2Doc` also depends on `gfan`, which is unnecessary.

> The program "gfan" is a build requirement

Can we change this, so "gfan" is only a build requirement for the three packages I mentioned? A full M2 build would still include everything.

DanGrayson:
> Ah, you're right about M2. I think the central issue here is that I see value in reducing unnecessary dependencies within a system, and since gfan is only a requirement for three packages `gfanInterface Tropical StronglyStableIdeals`, it seems to me like making Macaulay2Doc depend on gfan is unnecessary. What about "mgb" which comes with Mathicgb instead?
> 
> The CMake build, for instance, has a separate targets `build-programs` and `install-package`, and if you just want to build "M2-binary" to test "regex", as is the case currently for me, you don't need to build those extra targets. However, the command `help` stops working normally if `Macaulay2Doc` isn't installed (which is a separate issue #1085), so I install that too, but now `Macaulay2Doc` also depends on `gfan`, which is unnecessary.
> 
> > The program "gfan" is a build requirement
> 
> Can we change this, so "gfan" is only a build requirement for the three packages I mentioned? A full M2 build would still include everything.

No, I see no practical way to keep track of which packages depend on gfan.  And gfan is exactly the sort of program that findProgram was designed to find, so it makes an excellent example to illustrate it.


> > infer from that that you have installed gfan in a standard location where the build system can find it at the start
> 
> No, it's not installed in a standard location.

Then how does cmake know not to build gfan for you?

Wouldn't findProgram find it if you would simply put gfan's directory on your PATH?  If so, that would solve your problem trivially.

mahrud:
> No, I see no practical way to keep track of which packages depend on gfan. And gfan is exactly the sort of program that findProgram was designed to find, so it makes an excellent example to illustrate it.

This is not that hard really, but it's a separate task.

> Then how does cmake know not to build gfan for you?

I simply skip `build-programs` and `install-package`. It makes developing M2 much faster. This is my entire point. A complete build can build gfan and everything else, okay, but if I want to just build "M2-binary" and install "Macaulay2Doc" to try out a change I made, there's no real need to build an entirely separate program.

> Wouldn't findProgram find it if you would simply put gfan's directory on your PATH? If so, that would solve your problem trivially.

I wish, but no. I have gfan available there on my own machine for doing math, but when I need to test a different configurations using docker I can't use the gfan executable built for my machine because it is not compatible with the system running inside docker.

Do you find "mgb" not agreeable? Mathicgb is a library requirement, so it is already a real requirement of "M2-binary" with no way around it.

DanGrayson:
> > No, I see no practical way to keep track of which packages depend on gfan. And gfan is exactly the sort of program that findProgram was designed to find, so it makes an excellent example to illustrate it.
> 
> This is not that hard really, but it's a separate task.
> 
> > Then how does cmake know not to build gfan for you?
> 
> I simply skip `build-programs` and `install-package`. It makes developing M2 much faster. This is my entire point. A complete build can build gfan and everything else, okay, but if I want to just build "M2-binary" and install "Macaulay2Doc" to try out a change I made, there's no real need to build an entirely separate program.
> 
> > Wouldn't findProgram find it if you would simply put gfan's directory on your PATH? If so, that would solve your problem trivially.
> 
> I wish, but no. I have gfan available there on my own machine for doing math, but when I need to test a different configurations using docker I can't use the gfan executable built for my machine because it is not compatible with the system running inside docker.

And you don't want to install gfan on your docker image?

> 
> Do you find "mgb" not agreeable? Mathicgb is a library requirement, so it is already a real requirement of "M2-binary" with no way around it.

No, we build mgb only incidentally, we don't put it with the other programs where it can be run, and we don't distribute it -- it would not be found by findProgram.

mahrud:
> And you don't want to install gfan on your docker image?

No. I don't need gfan when I'm testing something like regex. Can we please reduce unnecessary dependencies? gfanInterface already depends on gfan, why would we make Macaulay2Doc also depend on it?

> No, we build mgb only incidentally, we don't put it with the other programs where it can be run, and we don't distribute it -- it would not be found by findProgram.

Oh, that's a bug then. I believe Mike uses mgb. It comes with the mathicgb [package](https://packages.debian.org/sid/amd64/mathicgb/filelist) on Debian and the CMake build installs and distributes it as well.

DanGrayson:
> > No, we build mgb only incidentally, we don't put it with the other programs where it can be run, and we don't distribute it -- it would not be found by findProgram.
> 
> Oh, that's a bug then. I believe Mike uses mgb. It comes with the mathicgb [package](https://packages.debian.org/sid/amd64/mathicgb/filelist) on Debian and the CMake build installs and distributes it as well.

No, nothing in Macaulay2 uses mgb.

DanGrayson:
> > And you don't want to install gfan on your docker image?
> 
> No. I don't need gfan when I'm testing something like regex. Can we please reduce unnecessary dependencies? gfanInterface already depends on gfan, why would we make Macaulay2Doc also depend on it?
> 

We would make Macaulay2Doc depend on it, because it's an ideal example for findProgram.

When I wanted to test your regex PR I started with a complete build, switched to your branch, recompiled in the d and e directories, relinked in the bin directory, and was ready to go in about 3 minutes.  It sounds like you insist on rebuilding everything from scratch.  If so, that's a waste of time.

mahrud:
> We would make Macaulay2Doc depend on it, because it's an ideal example for findProgram.

Okay, can we set RaiseError to false at least? If, as you say, gfan is a build requirement, then it'll always be found. Setting it to false has two fold benefits: demonstrate the use of the option, and also helps me (if nobody else). Seems completely harmless.

> When I wanted to test your regex PR I started with a complete build, switched to your branch, recompiled in the d and e directories, relinked in the bin directory, and was ready to go in about 3 minutes. It sounds like you insist on rebuilding everything from scratch. If so, that's a waste of time.

Of course not! Like I said, I want to test in docker environments to test in ubuntu, arch, etc. with different compilers, and I don't have 6 different complete builds. Testing this way was useful in finding out that std::regex was not reliable, for instance.

DanGrayson:
> > We would make Macaulay2Doc depend on it, because it's an ideal example for findProgram.
> 
> Okay, can we set RaiseError to false at least? If, as you say, gfan is a build requirement, then it'll always be found. Setting it to false has two fold benefits: demonstrate the use of the option, and also helps me (if nobody else). Seems completely harmless.

The thing is, this would help you not at all.  You have two totally trivial ways to get around a failing example in Macaulay2Doc: you could build the programs yesterday if it's too much trouble to build them today; or you could set IgnoreExampleErrors to true when you build Macaulay2Doc.  Either of those would work in your 6 different docker images.

The purpose of the example is to show findProgram successfully finding a program, so insisting on success is a good way to ensure that occurs.  That's why not setting RaiseError to false is a good thing.  Demonstrate the use of the option with a second example on a program that doesn't exist.

mahrud:
> or you could set IgnoreExampleErrors to true when you build Macaulay2Doc. Either of those would work in your 6 different docker images.

This is what I currently do, but I was hoping this change is harmless enough that I could avoid doing that.

> The purpose of the example is to show findProgram successfully finding a program, so insisting on success is a good way to ensure that occurs.

My point here was that the examples in Macaulay2Doc are not the right place for assertions that things work as intended. We can test that gfan exists and works in `gfanInterface` or even a new `tests/programs` subdirectory, perhaps.

DanGrayson:
> > or you could set IgnoreExampleErrors to true when you build Macaulay2Doc. Either of those would work in your 6 different docker images.
> 
> This is what I currently do, but I was hoping this change is harmless enough that I could avoid doing that.

The problem is that you're proposing a policy for the future, that Macaulay2Doc should not run programs, without
a way to enforce it at build time.  Authors should be confident that they aren't contravening any conventions once
the github actions succeed.  Receiving a complaint through the transom weeks later will just annoy.

> 
> > The purpose of the example is to show findProgram successfully finding a program, so insisting on success is a good way to ensure that occurs.
> 
> My point here was that the examples in Macaulay2Doc are not the right place for assertions that things work as intended. We can test that gfan exists and works in `gfanInterface` or even a new `tests/programs` subdirectory, perhaps.

No, examples are a great place to ensure that things work without error, and adding that option distracts the eye of the reader from what's really going on.

d-torrance:
I'll go ahead and close this PR

PR #1442, opened 2020-07-28
Improving support for external programs, part 6

d-torrance:
This PR ports the `CohomCalg` package to the new `findProgram`/`runProgram` interface.

For `findProgram` to work, we need a command that will return 0.  Somewhat frustratingly, the natural choice doesn't:

```console
profzoom@peg-amy:~$ cohomcalg --help

    cohomCalg v0.32
    (compiled for Linux/Unix x86-64 / 64 bit)
    author: Benjamin Jurke (mail@benjaminjurke.com)
    Based on the algorithm detailed in arXiv:1003.5217


Syntax:  cohomcalg [--option1] [--option2] ... InputFileName [> OutputFileName]

Command line options:
  --in="..."      Treats the text between parentheses like additional
                  input data, i.e. like appended content of the input file.
  --nomonomfile   Prohibits usage and generation of monomial file.
  --checkserre    Computes the Serre dual cohomology for comparison.
  --noreduction   Deactivates the Serre self-duality reduction for ambiguous
                  monomials (may increase computation time dramatically!)
  --hideinput     Does not print the input data read from the input file.
  --showtime      Shows timing stats even for application runs < 1 second.
  --showbits      Prints the internally used bitmasks for debug output.
  --mathematica   Gives formattet output for the Mathematica script version.
  --integrated    Produces minimalistic output for application integration.
  --verboseN      Provides debug output at level N=1..6, e.g. --verbose3.
  --maxX=N        Defines limits for X=verts,srgens,cohoms to N.

The order of options is irrelevant, but later commands always overwrite
prior options, e.g. from '--verbose4 --verbose1' only the later one will
have any effect.

The program automatically tries to add the file extension '.in' to the
InputFileName. See the package manual for further information.

profzoom@peg-amy:~$ echo $?
1
```

So instead, we include a small example file from the upstream source (the [del Pezzo surface](https://github.com/BenjaminJurke/cohomCalg/blob/master/bin/dP1.in)) and use it as an input file for the call to `cohomcalg` used by `findPackage`.

We also remove the now deprecated mutable exported string `cohomCalgExecutable`, since its purpose is now covered by setting `programPaths#"cohomcalg"`.

Note that this PR shouldn't have any conflicts with any of the others in this series.

### DIFF
--- a/M2/Macaulay2/packages/CohomCalg.m2
+++ b/M2/Macaulay2/packages/CohomCalg.m2
@@ -8,6 +8,7 @@ newPackage(
              HomePage => "http://www.math.cornell.edu/People/Faculty/stillman.html"
              }},
         Headline => "interface to CohomCalg software for computing cohomology of torus invariant divisors on a toric variety",
+        AuxiliaryFiles => true,
         PackageExports => {"NormalToricVarieties"},
         Configuration => {
             "executable" => "",
@@ -17,7 +18,6 @@ newPackage(
 
 -- TODO:
 --   add in documentation for these "monomialfile" routines
---   document cohomCalgExecutable
 --   document Silent
 --   add in use of --Verbose1, ..., Verbose6.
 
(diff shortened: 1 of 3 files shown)


mahrud:
You could try something like this: `cohomCalg --help || [[ "$?" == "1" ]]`. I think that syntax for `if` works in all versions of bash, but I'd check zsh and mac also.

d-torrance:
> You could try something like this: `cohomCalg --help || [[ "$?" == "1" ]]`. I think that syntax for `if` works in all versions of bash, but I'd check zsh and mac also.

I did consider doing something like that, but I was worried that there might be false positives, where `cohomcalg --help` returns 1 for some reason other than "that's what `--help` is supposed to return."   Maybe there's some other error that occurs that causes a return value of 1.  I'm not really sure what that might be -- the most common return value `findProgram` likely sees is 127 when the command isn't found -- but I'd rather get a 0 from the program itself than hoping that a particular nonzero return value actually meant success.

mahrud:
cohomCalg doesn't have any distributed Linux packages, right? In this case
findPrograms is probably less useful, but maybe we can patch it.

d-torrance:
> cohomCalg doesn't have any distributed Linux packages, right? In this case findPrograms is probably less useful, but maybe we can patch it.

It's in [Debian](https://tracker.debian.org/pkg/cohomcalg) and [Fedora](https://src.fedoraproject.org/rpms/cohomCalg).

mahrud:
Oh nevermind then.

Issue #1165, opened 2020-05-16
finding external programs


DanGrayson:
Ubuntu installs the programs of 4ti2 with the prefix "4ti2-" attached to their names.  Teach the package FourTiTwo how to find them there or under the non-prefixed name.

(Maybe test for the presence of "4ti2-markov" and "markov" on the PATH, and record the result for subsequent use.)

d-torrance:
This is also the case for nauty and (once it enters Debian) topcom.  "nauty-" and "topcom-" prefixes, respectively.

DanGrayson:
It's also the case for gfan and normaliz.

DanGrayson:
Maybe we need a top-down change for the way all interface packages find their programs: look for the program on the PATH after looking in the internal M2 place.  Packages could implement that by having the shell modify the PATH first when they try to run the external program, putting the internal M2 place at the start.  We could make that easier by having a core M2 function that does that simple thing.

mikestillman:
I think this is a very good idea.  At the moment, everyone has to discover what to do (e.g. by looking at one of the current interface packages), and the resulting interfaces are inconsistent.

However, a version on the system can either be too old (and so we don't want to use it), or newer (perhaps fixing bugs in the distributed version).  We should allow the user to select somehow between the system version, and the M2-distributed version.

mikestillman:
@DanGrayson here is the sort of warning I am receiving now, while compiling the e directory.  Either we need to quiet this one warning down, or (perhaps better): `scc` should define this as a class?  Actually, I'm not sure how to fix this, as currently, `engine-exports.h` needs to work with both C and C++.  Perhaps have `scc` place an ifdef for c++?  Or perhaps have these types included from one .h file?

```
CXX x-ringmap.cpp
In file included from ../../../../../../Macaulay2/e/x-ringmap.cpp:1:
../../../../../../Macaulay2/e/engine.h:14:1: warning: class 'Monomial' was previously declared as a struct [-Wmismatched-tags]
class Monomial;
^
../d/engine-exports.h:106:22: note: previous use is here
typedef const struct Monomial * engine_RawMonomial;
                     ^
In file included from ../../../../../../Macaulay2/e/x-ringmap.cpp:1:
../../../../../../Macaulay2/e/engine.h:15:1: warning: class 'Monoid' was previously declared as a struct [-Wmismatched-tags]
class Monoid;
^
../d/engine-exports.h:111:22: note: previous use is here
[...]

PR #1447, opened 2020-07-29
Improving support for external programs, part 7

### LINKED ISSUE #407: Wishlist: Use external programs if present

d-torrance:
This is the last pull request in this series.  If it and the others in the series are merged, then we can close #407 and #1165, as every one of Macaulay2's build dependencies on programs could be satisfied by using existing installations.

The focus is on the `SemidefiniteProgramming` package, which uses csdp (and mosek and sdpa, but only csdp is a build dependency).

Since csdp needs to be run in the same directory as the `params.csdp` file containing various parameters, we need the `RunDirectory` option to `runProgram`.  However, this was introduced in #1417, which hasn't been merged yet, so I went ahead and also added the corresponding commit to this branch, too.  If one of the PR's is merged, then I'll remove this commit from the other so they don't conflict.

### DIFF
--- a/M2/Macaulay2/m2/exports.m2
+++ b/M2/Macaulay2/m2/exports.m2
@@ -342,6 +342,7 @@ export {
 	"RingFamily",
 	"RingMap",
 	"RowExpression",
+	"RunDirectory",
 	"RunExamples",
 	"SPACE",
 	"SVD",
--- a/M2/Macaulay2/m2/programs.m2
+++ b/M2/Macaulay2/m2/programs.m2
@@ -63,14 +63,23 @@ findProgram(String, List) := opts -> (name, cmds) -> (
 )
 
 runProgram = method(TypicalValue => ProgramRun,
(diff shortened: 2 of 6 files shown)


DanGrayson:
It's one commit, so even if it's in two pull requests, it won't cause a conflict.

DanGrayson:
@diegcif @tom111 -- are these changes to the package SemidefiniteProgramming okay with you?

diegcif:
Yes, it seems fine. Thanks!

tom111:
Yes, looks like a nice improvement.

Issue #1408, opened 2020-07-15
Nauty error: Sorry, this program doesn't support digraphs yet


d-torrance:
This is from a build of the `development` branch on Ubuntu 20.10 using the Ubuntu nauty package.  

```
/<<BUILDDIR>>/macaulay2-16.0.1+git11929.74c3aa6/M2/usr-dist/common/share/doc/Macaulay2/Nauty/example-output/_relabel__Bipartite.errors:0:1: (output file) error: Macaulay2 exited with status code 1
/tmp/M2-367-0/0_relabel__Bipartite.m2:0:1: (input file)
M2: *** Error 1
currentString:1:1:(3):[2]: error: 1 error(s) occurred running examples for package Nauty:

_relabel__Bipartite.errors
**************************
-- -*- M2-comint -*- hash: 284388156

i1 : R = QQ[a..f];

i2 : G = graph flatten apply({a,c,e}, v->v*{b,d,f})

o2 = Graph{edges => {{a, b}, {b, c}, {a, d}, {c, d}, {b, e}, {d, e}, {a, f}, {c, f}, {e, f}}}
           ring => R
           vertices => {a, b, c, d, e, f}

o2 : Graph

i3 : relabelBipartite G
stdio:3:1:(3): error: callNauty: nauty terminated with the error [>E Sorry, this program doesn't support digraphs yet.].

make[3]: *** [Makefile:92: /<<BUILDDIR>>/macaulay2-16.0.1+git11929.74c3aa6/M2/usr-dist/x86_64-Linux-Ubuntu-20.10/lib/x86_64-linux-gnu/Macaulay2/Nauty/.installed] Error 1
```
Full build log at https://launchpadlibrarian.net/488677476/buildlog_ubuntu-groovy-amd64.macaulay2_16.0.1+git11929.74c3aa6-0ppa11681~ubuntu20.10.1_BUILDING.txt.gz:

I'm very confused as to what's happening here, as I don't have any problems with Nauty on Debian unstable, and the [diff between the Debian and Ubuntu packages](http://launchpadlibrarian.net/466865125/nauty_2.6r10+ds-1_2.6r10+ds-1ubuntu1.diff.gz) is very small and only affects the CI tests.

DanGrayson:
How would I get Ubuntu 20.10?  It doesn't exist yet.

d-torrance:
It's under development and won't be officially released until October.  But you can still get it:
[...]

PR #1452, opened 2020-07-31
Use AX_COMPARE_VERSION for the autotools nauty version check

d-torrance:
We previously we were using bc, but this adds an unnecessary build
dependency.

### DIFF
--- a/M2/configure.ac
+++ b/M2/configure.ac
@@ -1043,13 +1043,12 @@ for nauty_prefix in "" "nauty-"
 do
     if test "`type -t ${nauty_prefix}complg`" = file
     then nauty_version=`${nauty_prefix}complg --version | cut -d " " -f 3`
-         if test `echo "$nauty_version >= 2.7000" | bc -l` = 1
-         then AC_MSG_RESULT([yes, using prefix "$nauty_prefix"])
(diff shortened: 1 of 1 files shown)


d-torrance:
Without `bc`, we get things like this (from https://launchpadlibrarian.net/491022267/buildlog_ubuntu-focal-amd64.macaulay2_16.0.1+git11949.c25ac45-0ppa11703~ubuntu20.04.1_BUILDING.txt.gz):

```
checking whether the package nauty (>= 2.7) is installed... ./configure: line 11547: bc: command not found
./configure: line 11547: test: =: unary operator expected
```

DanGrayson:
Are you implying that we have a macro that depends on bc being present, but does not signal an error when bc is absent?  If so, that should be fixed, too.

By the way, bc is light, as a dependency.

d-torrance:
> Are you implying that we have a macro that depends on bc being present, but does not signal an error when bc is absent? If so, that should be fixed, too.

No, my code in aa51757 for fixing #1408 introduced the dependency on `bc`.  `AX_COMPARE_VERSION` doesn't use it.

DanGrayson:
Maybe if you merge with the development branch, the builds will start succeeding.

DanGrayson:
...never mind, enough builds succeeded.

PR #1465, opened 2020-08-16
`RRi` --- interval arithmetic via `mpfi`

antonleykin:
This is a draft pull request. 
The development of `RRi` has been started at the 3rd internals meeting --- see the notes:
https://github.com/Macaulay2/M2/wiki/Add-a-library-for-interval-arithmetic

### DIFF
--- a/M2/Macaulay2/d/expr.d
+++ b/M2/Macaulay2/d/expr.d
@@ -295,6 +295,7 @@ export QQClass := newnumbertype();
 export ringFamilyClass := newtypeof(typeClass);
 export inexactNumberTypeClass := newtypeof(ringFamilyClass);
        newbignumbertype():HashTable := newHashTable(inexactNumberTypeClass,inexactNumberClass);
+export RRiClass := newbignumbertype();
 export RRClass := newbignumbertype();
 export CCClass := newbignumbertype();
 
--- a/M2/Macaulay2/d/gmp.d
+++ b/M2/Macaulay2/d/gmp.d
@@ -39,6 +39,14 @@ export QQorNull := QQ or null;
 
 export QQcell := {+v:QQ};
 
+export RRimutable := Pointer "mpfr_ptr"; -- !!! needs to change
+
+export RRi := Pointer "mpfr_srcptr"; -- !!! needs to change
+
+export RRiorNull := RRi or null;
+
+export RRicell := {+v:RRi};
+
 export RRmutable := Pointer "mpfr_ptr";
 
 export RR := Pointer "mpfr_srcptr";
--- a/M2/Macaulay2/d/actors2.dd
+++ b/M2/Macaulay2/d/actors2.dd
@@ -581,6 +581,9 @@ leftshiftfun(e:Expr):Expr := (
 installMethod(Expr(LessLessS),ZZClass,ZZClass,
      Expr(CompiledFunction(leftshiftfun,nextHash()))
      );
+installMethod(Expr(LessLessS),RRiClass,ZZClass,
+     Expr(CompiledFunction(leftshiftfun,nextHash()))
+     );
 installMethod(Expr(LessLessS),RRClass,ZZClass,
      Expr(CompiledFunction(leftshiftfun,nextHash()))
      );
--- a/M2/Macaulay2/d/basic.d
+++ b/M2/Macaulay2/d/basic.d
@@ -29,6 +29,7 @@ export hash(e:Expr):int := (
      is x:DictionaryClosure do x.dictionary.hash -- there may be many dictionary closures with the same dictionary and different frames, too bad
      is x:QQcell do hash(x.v)
      is x:RRcell do hash(x.v)
+     is x:RRicell do hash(x.v)
      is x:CCcell do hash(x.v)
      is x:Sequence do (
 	  -- the numbers here are the same as in binary lookup() in objects.d!!
--- a/M2/Macaulay2/d/classes.dd
+++ b/M2/Macaulay2/d/classes.dd
@@ -20,6 +20,7 @@ setupconst("ZZ",Expr(ZZClass));
 setupconst("QQ",Expr(QQClass));
 setupconst("RingFamily",Expr(ringFamilyClass));
 setupconst("InexactFieldFamily",Expr(inexactNumberTypeClass));
+setupconst("RRi",Expr(RRiClass));
 setupconst("RR",Expr(RRClass));
 setupconst("CC",Expr(CCClass));
 setupconst("RingElement",Expr(ringElementClass));
@@ -138,6 +139,7 @@ export Class(e:Expr):HashTable := (
      is CompiledFunctionClosure do compiledFunctionClosureClass
      is s:SymbolClosure do if s.symbol.word.parse == parseWORD then symbolClass else keywordClass
      is SymbolBody do symbolBodyClass
+     is RRicell do RRiClass
      is RRcell do RRClass
      is CCcell do CCClass
      is RawComputationCell do rawComputationClass
--- a/M2/Macaulay2/d/interface.dd
+++ b/M2/Macaulay2/d/interface.dd
@@ -673,6 +673,12 @@ export rawConwayPolynomial(e:Expr):Expr := (
     );
 setupfun("rawConwayPolynomial",rawConwayPolynomial);
 
+export rawRRi(e:Expr):Expr := (
+     when e is prec:ZZcell do if !isInt(prec) then WrongArgSmallInteger(1)
+     else toExpr(Ccode(RawRingOrNull, "rawRingRRi(",toInt(prec),")" ))
+     else WrongArgZZ(1));
+setupfun("rawRRi",rawRRi);
+
 export rawRR(e:Expr):Expr := (
      when e is prec:ZZcell do if !isInt(prec) then WrongArgSmallInteger(1)
      else toExpr(Ccode(RawRingOrNull, "IM2_Ring_RRR(",toInt(prec),")" ))
--- a/M2/Macaulay2/d/parse.d
+++ b/M2/Macaulay2/d/parse.d
@@ -364,6 +364,7 @@ export TaskCell := {+ body:TaskCellBody };
 export Expr := (
      CCcell or
      RRcell or
+     RRicell or
      Boolean or
      CodeClosure or
      CompiledFunction or
--- a/M2/Macaulay2/e/CMakeLists.txt
+++ b/M2/Macaulay2/e/CMakeLists.txt
@@ -122,6 +122,7 @@ set(SRCLIST
   SLP
   ZZ
   ZZp
+  aring-RRi
   aring-CC
   aring-CCC
   aring-RR
--- a/M2/Macaulay2/e/Makefile.files
+++ b/M2/Macaulay2/e/Makefile.files
@@ -62,6 +62,7 @@ INTERFACE = \
 	aring-zz-flint \
 	aring-qq-flint \
 	aring-zzp-flint \
+	aring-RRi \
 	aring-RR \
 	aring-CC \
 	aring-RRR \
--- a/M2/Macaulay2/e/aring-RRi.cpp
+++ b/M2/Macaulay2/e/aring-RRi.cpp
@@ -1,0 +1,31 @@
+#include "aring-RRi.hpp"
+
+namespace M2 {
+
+void ARingRRi::text_out(buffer &o) const { o << "ARRi_" << mPrecision; }
+void ARingRRi::elem_text_out(buffer &o,
+                             const ElementType &ap,
+                             bool p_one,
+                             bool p_plus,
+                             bool p_parens) const
+{
+  mpfr_ptr a = &const_cast<ElementType &>(ap);
+  M2_string s = (*gmp_tostringRRpointer)(a);
+  bool prepend_plus = p_plus && (s->array[0] != '-');
+  bool strip_last =
+      !p_one && ((s->len == 1 && s->array[0] == '1') ||
+                 (s->len == 2 && s->array[1] == '1' && s->array[0] == '-'));
+
+  if (prepend_plus) o << "+";
+  if (strip_last)
+    o.put(s->array, s->len - 1);
+  else
+    o.put(s->array, s->len);
+}
+
+};  // end namespace M2
+
+// Local Variables:
+// compile-command: "make -C $M2BUILDDIR/Macaulay2/e  "
+// indent-tabs-mode: nil
+// End:
--- a/M2/Macaulay2/e/aring-RRi.hpp
+++ b/M2/Macaulay2/e/aring-RRi.hpp
@@ -1,0 +1,297 @@
+// Copyright 2012 Michael E. Stillman
+
+#ifndef _aring_RRi_hpp_
+#define _aring_RRi_hpp_
+
+#include <mpfi.h>
+#include "aring.hpp"
+#include "buffer.hpp"
+#include "ringelem.hpp"
+#include "ringmap.hpp"
+
+class RRi;
+class RingMap;
+
+namespace M2 {
+/**
+\ingroup rings
+*/
+class ARingRRi : public RingInterface
+{
+  // Higher precision real numbers
+
+ public:
+  static const RingID ringID = ring_RRi;
+
+  typedef __mpfr_struct elem;
+  typedef elem ElementType;
+
+  ARingRRi(unsigned long precision) : mPrecision(precision) {
+    // !!!this is just SOME code to test that mpfi library links fine 
+       mpfi_t x, y;
+       mpfi_init (x);			/* use default precision */
+       mpfi_init2 (y, 256);		/* precision _exactly_ 256 bits */
+       /* Unless the program is about to exit, do ... */
+       mpfi_clear (x);
+       mpfi_clear (y);
+  }
+  // ring informational
+  size_t characteristic() const { return 0; }
+  unsigned long get_precision() const { return mPrecision; }
+  void text_out(buffer &o) const;
+
+  unsigned int computeHashValue(const elem &a) const
+  {
+    double d = mpfr_get_d(&a, GMP_RNDN);
+    return static_cast<unsigned int>(d);
+  }
+
+  /////////////////////////////////
+  // ElementType informational ////
+  /////////////////////////////////
+
+  bool is_unit(const ElementType &f) const { return !is_zero(f); }
+  bool is_zero(const ElementType &f) const { return mpfr_cmp_si(&f, 0) == 0; }
+  bool is_equal(const ElementType &f, const ElementType &g) const
+  {
+    return mpfr_cmp(&f, &g) == 0;
+  }
+
+  int compare_elems(const ElementType &f, const ElementType &g) const
+  {
+    int cmp = mpfr_cmp(&f, &g);
+    if (cmp < 0) return -1;
+    if (cmp > 0) return 1;
+    return 0;
+  }
+
+  ////////////////////////////
+  // to/from ringelem ////////
+  ////////////////////////////
+  // These simply repackage the element as either a ringelem or an
+  // 'ElementType'.
+  // No reinitialization is done.
+  // Do not take the same element and store it as two different ring_elem's!!
+  void to_ring_elem(ring_elem &result, const ElementType &a) const
+  {
+    mpfr_ptr res = getmemstructtype(mpfr_ptr);
+    mpfr_init2(res, mPrecision);
+    mpfr_set(res, &a, GMP_RNDN);
+    result = ring_elem(moveTo_gmpRR(res));
+  }
+
+  void from_ring_elem(ElementType &result, const ring_elem &a) const
+  {
+    mpfr_set(&result, a.get_mpfr(), GMP_RNDN);
+  }
+
+  // 'init', 'init_set' functions
+
+  void init(ElementType &result) const { mpfr_init2(&result, mPrecision); }
+  void init_set(ElementType &result, const ElementType &a) const
+  {
+    init(result);
+    mpfr_set(&result, &a, GMP_RNDN);
+  }
+
+  void set(ElementType &result, const ElementType &a) const
+  {
+    mpfr_set(&result, &a, GMP_RNDN);
+  }
+
+  void set_zero(ElementType &result) const
+  {
+    mpfr_set_si(&result, 0, GMP_RNDN);
+  }
+
+  void clear(ElementType &result) const { mpfr_clear(&result); }
+  void copy(ElementType &result, const ElementType &a) const
+  {
+    mpfr_set(&result, &a, GMP_RNDN);
+  }
+
+  void set_from_long(ElementType &result, long a) const
+  {
+    mpfr_set_si(&result, a, GMP_RNDN);
+  }
+
+  void set_var(ElementType &result, int v) const
+  {
+    mpfr_set_si(&result, 1, GMP_RNDN);
+  }
+
+  void set_from_mpz(ElementType &result, mpz_srcptr a) const
+  {
+    mpfr_set_z(&result, a, GMP_RNDN);
+  }
+
+  bool set_from_mpq(ElementType &result, mpq_srcptr a) const
+  {
+    mpfr_set_q(&result, a, GMP_RNDN);
+    return true;
+  }
+
+  bool set_from_double(ElementType &result, double a) const
+  {
+    mpfr_set_d(&result, a, GMP_RNDN);
+    return true;
+  }
+  bool set_from_BigReal(ElementType &result, gmp_RR a) const
+  {
+    mpfr_set(&result, a, GMP_RNDN);
+    return true;
+  }
+
+  // arithmetic
+  void negate(ElementType &result, const ElementType &a) const
+  {
+    mpfr_neg(&result, &a, GMP_RNDN);
+  }
+
+  void invert(ElementType &result, const ElementType &a) const
+  // we silently assume that a != 0.  If it is, result is set to a^0, i.e. 1
+  {
+    mpfr_si_div(&result, 1, &a, GMP_RNDN);
+  }
+
+  void add(ElementType &result,
+           const ElementType &a,
+           const ElementType &b) const
+  {
+    mpfr_add(&result, &a, &b, GMP_RNDN);
+  }
+
+  void addMultipleTo(ElementType &result,
+                     const ElementType &a,
+                     const ElementType &b) const
+  {
+    mpfr_fma(&result, &a, &b, &result, GMP_RNDN);
+  }
+
+  void subtract(ElementType &result,
+                const ElementType &a,
+                const ElementType &b) const
+  {
+    mpfr_sub(&result, &a, &b, GMP_RNDN);
+  }
+
+  void subtract_multiple(ElementType &result,
+                         const ElementType &a,
+                         const ElementType &b) const
+  {
+    // result -= a*b
+    ElementType ab;
+    init(ab);
+    mult(ab, a, b);
+    subtract(result, result, ab);
+    clear(ab);
+  }
+
+  void mult(ElementType &result,
+            const ElementType &a,
+            const ElementType &b) const
+  {
+    mpfr_mul(&result, &a, &b, GMP_RNDN);
+  }
+
+  void divide(ElementType &result,
+              const ElementType &a,
+              const ElementType &b) const
+  {
+    mpfr_div(&result, &a, &b, GMP_RNDN);
+  }
+
+  void power(ElementType &result, const ElementType &a, int n) const
+  {
+    mpfr_pow_si(&result, &a, n, GMP_RNDN);
+  }
+
+  void power_mpz(ElementType &result, const ElementType &a, mpz_srcptr n) const
+  {
+    mpfr_pow_z(&result, &a, n, GMP_RNDN);
+  }
+
+  void swap(ElementType &a, ElementType &b) const { mpfr_swap(&a, &b); }
+  void elem_text_out(buffer &o,
+                     const ElementType &a,
+                     bool p_one,
+                     bool p_plus,
+                     bool p_parens) const;
+
+  void syzygy(const ElementType &a,
+              const ElementType &b,
+              ElementType &x,
+              ElementType &y) const  // remove?
+  // returns x,y s.y. x*a + y*b == 0.
+  // if possible, x is set to 1.
+  // no need to consider the case a==0 or b==0.
+  {
+    set_var(x, 0);  // set x=1
+    if (!is_zero(b))
+      {
+        set(y, a);
+        negate(y, y);
+        divide(y, y, b);
+      }
+  }
+
+  void random(ElementType &result) const  // redo?
+  {
+    randomMpfr(&result);
+  }
+
+  void eval(const RingMap *map,
+            ElementType &f,
+            int first_var,
+            ring_elem &result) const
+  {
+    if (!map->get_ring()->from_BigReal(&f, result))
+      {
+        result = map->get_ring()->from_long(0);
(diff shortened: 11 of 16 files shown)


antonleykin:
@DanGrayson, @MichaelABurr, the project doesn't build in the current state:
```
 SCC1 -dep basic.d
../../../../Macaulay2/d/basic.d:32:22: error: unsuitable arguments
```
but it did build before our group started on trying to define `RRicell`.

DanGrayson:
I'll take a look.

DanGrayson:
I think you have to set up the mpfi analog of this function:
```
export hash(x:RR):int := int(precision0(x)) + Ccode(int, 
     "mpfr_hash(",					    -- see gmp_aux.c for this function
          x, 
     ")"
    );
```

DanGrayson:
Oh, and then you have to implement that function.  I think I already have code for computing a hash code of an mfpr number, so you could define the hash code of an mpfi number to be a linear combination of the hash codes of the upper and lower limits.  It could even just be the sum, since commutativity is irrelevant -- the top and bottom never trade places.

DanGrayson:
Should I explain "hash codes"?

M2/Macaulay2/d/gmp.d
```diff
@@ -160,10 +162,19 @@ init(x:RRmutable,prec:ulong):RRmutable := (
...
+Ccode( RRimutable, "(mpfr_init2(", x, ",(mpfr_prec_t)",prec,"),",x,")" )
+); -- End added for MPFI
+
 export newRRmutable(prec:ulong):RRmutable := init(GCmalloc(RRmutable),prec);
 
+export newRRimutable(prec:ulong):RRimutable := init(GCmalloc(RRimutable),prec); -- Added for MPFI
+
 clear(x:RRmutable) ::= Ccode( void, "mpfr_clear(",  x, ")" );
 
+clear(x:RRimutable) ::= Ccode( void, "mpfr_clear(",  x, ")" ); -- Added for MPFI
```
DanGrayson, line 176: This looks wrong, because x is an RRi but mpfr_clear takes an RR.

M2/Macaulay2/d/gmp_aux.c
```diff
@@ -21,6 +21,15 @@ int mpfr_hash(mpfr_srcptr x) {
   return 777 + h * 3737 + x->_mpfr_exp + 11 * x->_mpfr_sign;
 }
 
+int mpfi_hash(mpfr_srcptr x) { // Added for MPFI
```
DanGrayson, line 24: This looks wrong: "mpfi_hash" should take an mpfi, not an mpfr.

DanGrayson: Why not define mpfi_hash the sum of the results of mpfr_hash on the two endpoints?

M2/Macaulay2/d/gmp_aux.h
```diff
@@ -2,3 +2,4 @@
 
 extern int mpz_hash(mpz_srcptr x);
 extern int mpfr_hash(mpfr_srcptr x);
+extern int mpfi_hash(mpfr_srcptr x);  // Added for MPFI
```
DanGrayson, line 5: This looks wrong -- same type mismatch as above.

DanGrayson:
All those comments "Added for mpfi" just add noisy clutter to the code.

MichaelABurr:
Right now, the code is just a copy of mpfr, we haven't put mpfi into the mix.  All the comments, etc., are about just marking what is minimally needed to get a sucessful compilation.

MichaelABurr:
I'm working on adapting "moveToRR" function for mpfi's.  Could someone give a brief overview of how this works?

DanGrayson:
> I'm working on adapting "moveToRR" function for mpfi's. Could someone give a brief overview of how this works?

There are two types: RR and RRmutable.  Each contains pointers to allocated memory areas.  For RR, the memory is allocated with libgc, our garbage collector, so it gets automatically collected.  For RRmutable, the memory is allocated with the allocator used by mpfr and gmp, which is usually the standard routine "malloc".  The purpose of "moveToRR", after a computation is over, is to move the data into memory allocated by libgc.

> 
> If I'm reading the moveToRR function correctly, first there's a malloc call, then some c code that makes a copy of the mpfr at a low level. Finally, does the Ccode(RR,y) cast y in to an RR?

Yes, in effect.  The "Ccode" macro takes the type RR as its first argument, and that's the type of the result.  The rest of the arguments are either strings containing literal C code or names of variables in the D language.  The latter are replaced by the corresponding variable in the C language.

> 
> For mpfi, the struct just consists of two MPFR's. I'd like to leverage the moveToRR function for the moveToRRi, but I'm not sure how to mix the c and d code (I'm not sure how to call moveToRR on mpfi->left, for instance). Or, do I just need to copy the MPFR low-level copy twice (for left and right)?

Yes, you would like to call "moveToRR" twice, but there is a complication.  Here are the relevant declarations from "mpfi.h":
```c
typedef struct {
  __mpfr_struct left;
  __mpfr_struct right;
}__mpfi_struct;

typedef __mpfi_struct mpfi_t[1];
typedef __mpfi_struct *mpfi_ptr;
typedef const __mpfi_struct *mpfi_srcptr;
```

And here are the relevant declarations from "mpfr.h":
```c
typedef struct {
  mpfr_prec_t  _mpfr_prec;
  mpfr_sign_t  _mpfr_sign;
  mpfr_exp_t   _mpfr_exp;
  mp_limb_t   *_mpfr_d;
} __mpfr_struct;

typedef __mpfr_struct mpfr_t[1];
typedef __mpfr_struct *mpfr_ptr;
typedef const __mpfr_struct *mpfr_srcptr;
```

And here's what "RRmutable" and "RR" are declared to be:
```
export RRmutable := Pointer "mpfr_ptr";
export RR := Pointer "mpfr_srcptr";
```

If z is an mpfi pointer, then the D code ` Ccode(RRmutable, "&", z, "->left") ` ought to get you the left end point, and this can be passed to "moveToRR".  

Oh, but if you pass that to "moveToRR", then the first thing it will do is this:
```
     y := GCmalloc(RRmutable);
```
That will allocate space for a struct of type __mpfr_struct, as declared above.  But that's not what we want, because an mpfi struct consists of two structs of type __mpfr_struct, side by side, in the same block of memory. 

So the function moveToRRi has to duplicate what moveToRR does; in particular, it should allocate memory for a struct of type __mpfi_struct.  It should then do what moveToRR does for the left endpoint and also do what moveToRR does for the right endpoint.

antonleykin:
@DanGrayson, we get
```

i1 : i = interval (1,2)

o1 = [1,2]

o1 : RRi (of precision 53)

i2 : i+i
stdio:2:2:(3): error: no method for binary operator + applied to objects:
--            [1,2] (of class RRi)
--      +     [1,2] (of class RRi)

```
despite `+` being defined on line 1126 of `d/gmp.d`.
We do see the similarly defined `+` when we `peek RR`... but it's not there when we `peek RRi`.
What are we missing?

MichaelABurr:
@DanGrayson, Anton and I did some searching and we got addition to work.  We were missing the connection in actors.d

antonleykin:
@MichaelABurr, I've updated https://github.com/Macaulay2/M2/wiki/Internals:-Adding-a-Library-for-Interval-Arithmetic with a detailed list of changes so far (in `d/` and `m2/`).
Please check if complete. Also there is one comment about an unused (possibly not needed) function.

MichaelABurr:
@DanGrayson could you describe the role that the typical values comments make in actors.d?  Are they just comments or do they come into the compilation/linking/installation process somehow?  I think that I put one in incorrectly and got an error (installation error) in the last stages of gmake.

mahrud:
> @DanGrayson could you describe the role that the typical values comments make in actors.d? Are they just comments or do they come into the compilation/linking/installation process somehow? I think that I put one in incorrectly and got an error (installation error) in the last stages of gmake.

Take a look here:
https://github.com/Macaulay2/M2/blob/7ba26782adc33cb23f4ca3522130721f53b93fb8/M2/Macaulay2/m2/tvalues-make.m2#L1-L18
This file is read here:
https://github.com/Macaulay2/M2/blob/7ba26782adc33cb23f4ca3522130721f53b93fb8/M2/Macaulay2/m2/typicalvalues.m2#L167-L181
You can test it out with a command like this from the build directory:
```
./M2 --no-preload --no-tvalues ../../Macaulay2/m2/tvalues-make.m2 -e 'print make "../../Macaulay2/d/"; exit 0'
```

DanGrayson:
> @DanGrayson could you describe the role that the typical values comments make in actors.d? Are they just comments or do they come into the compilation/linking/installation process somehow? I think that I put one in incorrectly and got an error (installation error) in the last stages of gmake.

See https://m2internals.slack.com/archives/C016HLWBWV7/p1601841519001700?thread_ts=1593206752.008500&cid=C016HLWBWV7

MichaelABurr:
@DanGrayson As I understand it, there's a logic error in the compare function when dealing with intervals.  The code:

```
x = interval(1,3) -- creates interval [1,3] 
0 < x -- returns true  
0 <= x -- returns true  
1 < x -- returns false  
1 <= x -- returns false, but should be true
```

Although I might misunderstand it, `<=` is parsed as `< or =`.  In other words, the code tests `1 < x`, which is correctly false, and `1 == x`, which is also correctly false, but `1 <= x` is true.  The problem is that there isn't enough data in the three returns of `LessE`, `GreaterE`, `EqualEqualE` to identify this case.  Note that `2 < x` and `2 == x` are both false, but `2 <= x` is false.

One possible fix would be to add the expressions `LessEqualE` and `GreaterEqualE` as possible returns.  Alternatively, `1 <= leftRR(x)` does give the correct return, but I'm not sure where compare is called from in this case, and, if we're at the expression level, this might not make sense.

Do you have any suggestions on how to fix this error that would be M2-appropriate?

DanGrayson:
> @DanGrayson As I understand it, there's a logic error in the compare function when dealing with intervals. ...

This is a really interesting problem, since these interval numbers don't form a partially ordered set.  I like your proposal about adding LessEqualE and GreaterEqualE as possible return values for the comparison operation `x ? y` : you'll have to fix our generic code for x <= y and x >= y to take that into account by reacting properly when the new values appear.  The code for `x <= y` should still call the comparison operation, which will be the only place where mpfi is called.

What will you do when comparing the interval [1,3] to the interval [2,4] or to the number 2?  Maybe you need another new return value : `Undecidable`.  Even comparing [1,3] to [1,3] should result in Undecidable, I guess!  So you'll never return Equal.

antonleykin:
https://github.com/Macaulay2/M2/pull/1465/commits/e6b4585ff93500d25567a05fa9d5a4726c0bd517 makes changes to build mpfi from sources.

@DanGrayson, could you check that I didn't miss anything?
Also, should the tar file be placed here: https://faculty.math.illinois.edu/Macaulay2/Downloads/OtherSourceCode/ ?

@mahrud, I did only `autotools`. I expected 3 (out of 8) checks to pass, but the two `ubuntu` checks fail and the reason doesn't seem to be M2 related.

mahrud:
> @mahrud, I did only `autotools`. I expected 3 (out of 8) checks to pass, but the two `ubuntu` checks fail and the reason doesn't seem to be M2 related.

I've wanted c2c0b2a2fd to be reverted before, since it wastes resources updating prebuilt images that github provides. Apparently it doesn't always solve the issue that it was supposes to solve either. cc: @d-torrance

d-torrance:
The build failures indeed seemed to be from the `apt-get update` step:

```
E: Failed to fetch https://packagecloud.io/github/git-lfs/ubuntu/dists/bionic/InRelease  402  Payment Required [IP: 54.193.34.251 443]
E: The repository 'https://packagecloud.io/github/git-lfs/ubuntu bionic InRelease' is no longer signed.
```

This issue appeared to affect pretty much everyone (https://github.com/git-lfs/git-lfs/issues/4294) but has now been closed.  I think the builds should work again now.

I think situations like this (where `apt-get update` fails) are likely rarer than situations that prompted c2c0b2a in the first place (where GitHub's Ubuntu packages are out-of-date).  But my feelings certainly won't be hurt if that commit is reverted.  :)

DanGrayson:
> I've wanted [c2c0b2a](https://github.com/Macaulay2/M2/commit/c2c0b2a2fd2e915d45a084c0ce2da1f1b8063988) to be reverted before, since it wastes resources updating prebuilt images that github provides. Apparently it doesn't always solve the issue that it was supposes to solve either. cc: @d-torrance

Why revert that?  The "apt-get update" just gets the list of packages from the server, it doesn't upgrade any.

mahrud:
Ah, never mind, somehow repeatedly my eyes read "update" as "upgrade".

antonleykin:
I made the `cmake` build work on my Mac, but all automatic cmake checks fail. 
@mahrud, could you see what I missed in  https://github.com/Macaulay2/M2/pull/1465/commits/1bedff37c29747a96468d609efe271664a2f219d ?
It seems the the `autotools` check fails on `ubuntu` for reasons unrelated to this branch.

mahrud:
It built fine locally, but neither the ubuntu nor mac environments on github have mpfi, so I pushed a commit to install mpfi with brew and added `FindMPFI.cmake` so it can find it. Hopefully that should resolve it.

mahrud:
If you remove the "restart" line from this test, I believe all CMake builds should succeed:
https://github.com/Macaulay2/M2/blob/09d205ae632a8835f76b74b5996c9646f034dede/M2/Macaulay2/tests/normal/RRi.m2#L3

mahrud:
Is this PR not going to make it into this release? I think the failing test is just due to changed hash codes, which happens when you introduce new objects in the interpreter and can just be updated. Right?

DanGrayson:
> Is this PR not going to make it into this release? I think the failing test is just due to changed hash codes, which happens when you introduce new objects in the interpreter and can just be updated. Right?

If tests are failing because hash codes have changed due to new internal objects in the interpreter, the right way to fix that is to move the construction of the new internal objects a little later in the sequence of events.  This will work, as mutable objects, including types, are assigned serial numbers as their hash codes.

MichaelABurr:
Hi Mahrud,
	I'm not planning on adding RRi to this release, but the next one.  Even though there's not too much to do to finish getting it ready, I'm really overwhelmed with the end of the semester and don't want to promise something that I can't deliver.
	Best,
	Michael

mahrud:
@antonleykin @MichaelABurr would either of you mind if I rebase this branch on top of development to see if the tests still pass? If not, could you commit any uncommitted changes and push local commits?

It would be easiest to rebase, and the history of changes will be easier to track, but if you prefer I can just merge with development. The only downside of rebasing is that you might need to do a hard reset to pull the changes, whereas with a merge you would only pull.

mahrud:
Also, I'm kind of interested in trying out Arb and ANTIC, both of which have very similar API to FLINT, and are used, for instance, in Calcium. Do you know whether switching to Arb would be a very significant change at this point, or just switching some low-level functions?

DanGrayson:
> @antonleykin @MichaelABurr would either of you mind if I rebase this branch on top of development to see if the tests still pass? If not, could you commit any uncommitted changes and push local commits?
> 
> It would be easiest to rebase, and the history of changes will be easier to track, but if you prefer I can just merge with development. The only downside of rebasing is that you might need to do a hard reset to pull the changes, whereas with a merge you would only pull.

It's probably best if Anton merges with our master branch.

MichaelABurr:
I'm ok with either (although I don't have a lot of experience with rebasing).  I'm currently working on my own branch, moving some of my work out of the d-language to the m2-language (and similar).

mahrud:
@MichaelABurr I rebased your branch on top of development and corrected a ton of conflicts. It seems to be building fine (though, `CC_53[x]` gives an error!). I can push to this branch if Anton approves, or you can continue working from https://github.com/mahrud/M2/tree/feature/RRi

antonleykin:
I see no problem --- please push the rebased version.

mahrud:
Done. Here is an example of an error that were currently getting:
```m2
i5 : CC[x];
stdio:5:1:(3): error: expected new class to be a specialization of the old one
```
I'm not sure when it was introduced.

MichaelABurr:
The last time that I encountered that error, the issue was in reals.m2.  
CC[x] is working in my working version, so I'll look for differences and get back to you if I find anything.

mahrud:
> The last time that I encountered that error, the issue was in reals.m2. CC[x] is working in my working version, so I'll look for differences and get back to you if I find anything.

Sounds good! By the way, since this branch has been rebased, this is how you can change your local fork to this branch so you can push to this branch:
```
git fetch antonleykin # fetch recent changes
git pull antonleykin internals3-mpfi --rebase -Xours # pull while rebasing
git reset antonleykin/internals3-mpfi # reset to that commit
```
At the end, there may be small differences, which you can review using `git status` and `git diff`.

MichaelABurr:
Can you check reals.m2?  Line 74 shouldn't have RRi listed in the baseRings for CC.  The "correct" version should be:

new ComplexField of Nothing' from ZZ := memoize( 
	(ComplexField,Nothing',prec) -> newClass(ComplexField,Nothing', 
	hashTable { 
	symbol precision => prec, 
	symbol Engine => true, 
	symbol isBasic => true, 
	symbol baseRings => {ZZ,QQ,RR}, 
	symbol RawRing => rawCC prec }))
	
I'm currently dealing with some other bugs with the rebased version.  Once I've dealt with those, I'll be able to push the fix (if you haven't already done it).

mahrud:
I'll try later tonight. Are you having issues with the rebase (e.g. do you get any conflicts?) or is it just bugs that appeared after rebasing?

mahrud:
Hmm, I thought I fixed those, but maybe I missed some?

MichaelABurr:
I'm preparing a pull request since my solutions to the errors may not be optimal.

	- Main Issue: In Macaulay2/packages/Macaulay2Doc/shared.m2 , the key "intersection" on line 7 was giving issues due to intersection not being previously exported.
		
	- Solution: I added the following to Macaulay2/m2/shared.m2
		
		intersection = method(Options => true) 
		export "intersection"

	- Subsequent issue: Macaulay2/packages/ReesAlgebra.2 uses a variable name "intersection" in one of its examples.  I changed the name to intersectionReesAlgebra .

mahrud:
I meant to ask this earlier: could you use intersect instead?

MichaelABurr:
I'm not quite sure what you're asking.  I'm sure that you could use the name intersect instead of intersection, but this isn't a conflict with RRi, it's an issue that appeared when I pulled the rebased version.

mahrud:
> I'm not quite sure what you're asking. I'm sure that you could use the name intersect instead of intersection, but this isn't a conflict with RRi, it's an issue that appeared when I pulled the rebased version.

I see. There was a conflict involving intersection with NAGtypes, and I thought that was the only one. Moving the declaration of intersection into either `methods.m2` or `shared.m2` seems reasonable to me.

Parallel to that, I think using intersect might be better, as long as there isn't a particular need to have separate methods.

mahrud:
Okay, I made some changes to `intersect`, renamed the interval intersection code to `intersect`, but left the `intersection` function as commented, in case you want to revert it.

I thought this is easier than changing all the places where `intersection` is defined, but eventually perhaps `intersection` should be a synonym of `intersect`, or vice versa.

Is that good? Feel free to revert any of that, I didn't mean to interrupt your work.

MichaelABurr:
I've been working to move some of the interval functions out of the d-language and into the m2-language.  This should alleviate these issues with naming (at least somewhat).  I'll review your changes today.

M2/Macaulay2/d/actors4.d
```diff
@@ -1313,24 +1313,24 @@ toRRi(e:Expr):Expr := (
                	       when s.1 is y:ZZcell do toExpr(toRRi(x.v,y.v))
                                 is y:QQcell do toExpr(toRRi(x.v,y.v))
                                 is y:RRcell do toExpr(toRRi(x.v,y.v))
-                    	        else WrongArg(1,"a pair of integral, rational, or real numbers"))
+                    	        else WrongArg(1,"a pair of integral, rational, or real numbers, with a precision"))
```
DanGrayson, line 1316: Adding "with a precision" makes it sound like the user should provide a real number and a precision, two things.  I think it was fine the way it was.

M2/Macaulay2/d/actors4.d
```diff
@@ -1399,39 +1399,6 @@ subsetRRi(e:Expr):Expr := (
          else WrongArg(1,"a pair of integral, rational, real numbers or intervals"));
```
DanGrayson, line 1399: An "interval" is primarily a real number, so such things are implicitly included in the phrase "real numbers".

M2/Macaulay2/m2/exports.m2
```diff
@@ -739,6 +739,7 @@ export {
 	"id",
 	"ideal",
 	"identity",
+    "identical",
```
DanGrayson, line 742: Does this `identical(x,y)` differ from `x===y` somehow for objects of type RRi?  I would expect it not to.

M2/Macaulay2/m2/exports.m2
```diff
@@ -793,6 +794,7 @@ export {
 	"isInputFile",
 	"isIsomorphism",
 	"isListener",
+    "isMember",
```
DanGrayson, line 797: It's going to be pretty confusing to have both `isMember` and `member` defined.

MichaelABurr:
In general, I prefer to prefix functions whose return value is a boolean with "is" (I find it a useful mnemonic).
I could change the name to member and there shouldn't be a problem.
Do you have an alternate suggestion (other than member) that might prevent confusion?

MichaelABurr:
For RRi's there are several equalities.

As coded right now: 

== is typically false for intervals (it can only be true when both intervals have diameter 0)
	[2,2] == [2,2] is true
	[2,3] == [2,3] is false

=== is likely currently broken.  As I understand it, it should be a pointer check.
	
	typical behavior:
	x = interval[2,3]
	y = interval[2,3]
	x===x is true
	x===y is false

	undesirable current behavior:
	x = interval(2)
	y = interval(2)
	x === y is true

I need a third type of equality, a set-theoretic equality.  
	I had originally called it sameInterval, but recently changed it to identical.
	If you think that sameInterval would be a better name, I can change it back 
	(or perhaps you have an alternate suggestion).

M2/Macaulay2/m2/exports.m2
mahrud: Dan, you've mentioned in the past that we should rename `mutable -> isMutable` and similar functions, I think `member -> isMember` would be reasonable as well.

MichaelABurr:
toRRi is mostly a back-end function.  
It shouldn't really be called by the user (I do export it, but perhaps I shouldn't).

The behavior of the function is:
	One input, an interval of default precision of diameter 0.
	Two inputs, an interval of default precision with given endpoints.
	Three inputs, an interval of specified precision with given endpoints.

I intended the error messages to indicate which branch of toRRi the user ended up in since this is more like debugging error messages than typical user-level error messages.

MichaelABurr:
I can make this change, but this is not an exported function.

MichaelABurr:
I would very much like to have an alternate name for this function:
	At the internals meeting, I felt like there was a suggestion to avoid function names that include their input/output type as that clutters the namespace.
	On the other hand, identical is a function that returns a Boolean that doesn't begin with "is", and I'd prefer such a function to start with "is"
	Using a function name like "isSame" seems like a bad idea with "same" already defined.

DanGrayson:
>  === is likely currently broken. As I understand it, it should be a pointer check.

No, x===y should check whether x and y have the same contents and behave the same way.  For RRi it should check that the two endpoints are the same (exactly).

DanGrayson:
>  As coded right now: == is typically false for intervals (it can only be true when both intervals have diameter 0) [2,2] == [2,2] is true [2,3] == [2,3] is false


What about making x==y mean that the secret numbers represented by the intervals x and y **could** be the same?  That might make x==y useful.  For example, if x could be 0, you don't want to divide by it, and if x couldn't be 0, you may.


Perhaps there are other sensible choices.

M2/Macaulay2/m2/exports.m2
DanGrayson: Yes, that would be quite reasonable.

M2/Macaulay2/m2/exports.m2
DanGrayson: Let's settle what x===y does first -- maybe it will end up doing what "identical" now does.

DanGrayson:
> I would very much like to have an alternate name for this function: At the internals meeting, I felt like there was a suggestion to avoid function names that include their input/output type as that clutters the namespace. On the other hand, identical is a function that returns a Boolean that doesn't begin with "is", and I'd prefer such a function to start with "is" Using a function name like "isSame" seems like a bad idea with "same" already defined.
> […](#)
> On Feb 2, 2021, at 8:30 AM, Daniel R. Grayson ***@***.***> wrote: @DanGrayson commented on this pull request. In M2/Macaulay2/m2/exports.m2: > @@ -739,6 +739,7 @@ export { "id", "ideal", "identity", + "identical", Does this identical(x,y) differ from x===y somehow for objects of type RRi? I would expect it not to. — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub, or unsubscribe.

Yes, we can't have both `isSame` and `same`.  However, it might be reasonable to rename `same` now.  

Oh, but when I look at `same{a,b,c}` it seems to me that we are testing whether a, b, and c are the same.  Perhaps someone will now lobby for `areSame` as the name!

mahrud:
> Oh, but when I look at `same{a,b,c}` it seems to me that we are testing whether a, b, and c are the same. Perhaps someone will now lobby for `areSame` as the name!

I'd much prefer not touching `all`, `any`, `same`, etc. I don't see anything wrong with using `same` for checking if two (or more) intervals are the same, but `identical` is fine too imo.

MichaelABurr:
I read more into the help of === and I'll need to fix it, but === should replace identical.
I didn't realize that === had different behaviors for mutable and immutable objects.
Since intervals are immutable, it makes sense to use === for having the same contents, 
which is what identical does.  
I'll make this change.

MichaelABurr:
A (somewhat long) alternative to determine if x and y could represent the same number is
	not isEmpty intersect(x,y)
I think that I like using == and === for statements about the intervals themselves (as sets)
and using other commands for interpretations.

There are times when an interval is not used to represent a secret point, but to evaluate 
a function over a region.  I'd like to leave interpretations up to the user and provide the 
tools so that the user can interpret the results appropriately.

Ex: In root finding, intervals can be used to exclude regions from having a root
	If f is a univariate polynomial and I is an interval, 
	then interval arithmetic computes an interval J containing f(I),
	If J doesn't contain 0, then f doesn't have a root in I.

M2/Macaulay2/m2/exports.m2
DanGrayson: But changing "member" to "isMember" should be done separately, not in this pull request.

DanGrayson:
> A (somewhat long) alternative to determine if x and y could represent the same number is not isEmpty intersect(x,y) I think that I like using == and === for statements about the intervals themselves (as sets) and using other commands for interpretations. 

But aren't you explicitly using intervals just as a way to represent approximate real numbers, not as sets?  Hence the name "RRi", and that's how we're going to have matrices whose entries are such things.  The corresponding interval is something else, even mathematically.  Indeed, if you want to treat intervals as mathematical objects, you might introduce a new type, `Interval`, and provide methods for computing with it.

MichaelABurr:
The name RRi does seem to lend to the idea that RRi's are supposed to represent RR's.
One could do what you suggest, but I don't think that I'm in favor of it (at least not yet).
If we created a separate type called interval, it would create a lot of duplication for only a few benefits.
Almost all of the functions and operations would be identical - it would be like creating a different type for polynomials as algebraic objects and as functions.

I prefer the conservative approach, in the following way:
Suppose that X is an interval.  Then X/X is interpreted as {a/b : a,b in X}.
This is the least lossy value for X/X.  If the user has knowledge about the application, 
e.g., that the interval represents a point, then the user should explicitly replace X/X with [1,1] (we shouldn't do it for the user).
I don't want to make a specialized choice for the user, since I think that it's easier for a user to ignore or throw away data to make a specialized choice, 
than for a user to recover data that we've already discarded.

DanGrayson:
I also prefer the conservative approach, but ignoring the fact that this implementation uses intervals.  If x and y are objects in RRi representing approximations to abstract real numbers x' and y', then x/y is the most accurate approximation to x'/y' available to us in RRi.  This more abstract point of view makes sense for any way of approximating real numbers, and dictates the implementation.

MichaelABurr:
I believe that this draft pull request is ready to be a regular pull request.  I don't think that I can change it from a draft.  

One of the build and tests failed on the hash check, but I'm not sure why (or how to fix it).

DanGrayson:
I pushed the "ready for review" button, and that changed it into a real pull request.

DanGrayson:
This test fails:

https://github.com/Macaulay2/M2/pull/1465/checks?check_run_id=2522569259#step:13:722

M2/Macaulay2/d/expr.d
```diff
@@ -295,6 +295,7 @@ export QQClass := newnumbertype();
 export ringFamilyClass := newtypeof(typeClass);
 export inexactNumberTypeClass := newtypeof(ringFamilyClass);
        newbignumbertype():HashTable := newHashTable(inexactNumberTypeClass,inexactNumberClass);
+export RRiClass := newbignumbertype();
```
DanGrayson, line 298: To fix the assertion error, simply move this new line above the line below that says "-- all new types, dictionaries, and classes go just above this line, if possible, so hash codes don't change gratuitously!
"

MichaelABurr: Thanks - I made the change.  Obviously, I missed the comment about where to add new types.

MichaelABurr:
Does anyone have any ideas on the failed tests?  It looks like cmake-ubuntu failed once and succeeded once (on NoetherianOperators), so that might not be a problem.  The seg-fault on autotools-ubuntu is more mysterious to me.  I can't reproduce the error locally (although I'm not running ubuntu).  I'm also not seeing anything in SuperLinearAlgebra that should cause a problem.

d-torrance:
> Does anyone have any ideas on the failed tests? It looks like cmake-ubuntu failed once and succeeded once (on NoetherianOperators), so that might not be a problem. The seg-fault on autotools-ubuntu is more mysterious to me. I can't reproduce the error locally (although I'm not running ubuntu). I'm also not seeing anything in SuperLinearAlgebra that should cause a problem.

They both look like known issues, unrelated to your pull request.  The autotools build failed at `check(34, "MinimalPrimes")` and the cmake build making the example results for `getIdealFromNoetherianOperators`, both of which are known instances of #1742.

MichaelABurr:
> > Does anyone have any ideas on the failed tests? It looks like cmake-ubuntu failed once and succeeded once (on NoetherianOperators), so that might not be a problem. The seg-fault on autotools-ubuntu is more mysterious to me. I can't reproduce the error locally (although I'm not running ubuntu). I'm also not seeing anything in SuperLinearAlgebra that should cause a problem.
> 
> They both look like known issues, unrelated to your pull request. The autotools build failed at `check(34, "MinimalPrimes")` and the cmake build making the example results for `getIdealFromNoetherianOperators`, both of which are known instances of #1742.

I see - it looks like the tests were run twice (I'm showing two sets of checks when I click on details).  In one of the tests, the autotools-ubuntu run got through the MinimalPrimes issue, but seg-faulted on SuperLinearAlgebra.

d-torrance:
> I see - it looks like the tests were run twice (I'm showing two sets of checks when I click on details). In one of the tests, the autotools-ubuntu run got through the MinimalPrimes issue, but seg-faulted on SuperLinearAlgebra.

That one looks like it may be related to #1984.

DanGrayson:
The build failures seem unrelated.

PR #1475, opened 2020-08-23
Use KaTeX for web documentation

### LINKED ISSUE #1288: Switch to KaTeX for rendering math in documentation
[KaTeX](https://katex.org/) is a javascript library for rendering math. We could use it two ways:
- load the javascript library in html files and let the browser handle it.
- use server side rendering, which would mean adding a dependency on Node.js.

I think the first option is probably best.

With this change, I think we should also remove the `TEX` type (or make it idempotent) and delete `m2/texhtml.m2` and other `TEX` related methods.

mahrud:
- [x] Close #1288 
- [x] translate basic TeX sequences like `{\tt/bf/em/it ...}` and `\url{...}` to Hypertext types in the constructor of `TEX` using regex (this way they show up nicely in net and info as well)
- [x] add synonym `BF` and `IT` for `BOLD` and `ITALICS` to simplify translating TeX to Markup

This PR is on top of PR #1454 because it includes changes in a few places that would cause conflicts.


DanGrayson:
I get a problem while making the top level help file with "-q" not on the command line:
```m2
+ /Users/dan/src/M2/M2.git/M2/BUILD/dan/builds.tmp/gallium-1475/M2 --no-readline --print-width 196 --no-debug
Macaulay2, version 1.16.0.2
with packages: ConwayPolynomials, Elimination, IntegralClosure, InverseSystems, LLLBases, MinimalPrimes, PrimaryDecomposition, ReesAlgebra, TangentCone, Truncations
M2/Macaulay2/m2/debugging.m2:20:6:(1):[21]: error: "null" is of an unrecognized type "Nothing" with no qname, appearing in hypertext during validation
M2/Macaulay2/m2/validate.m2:23:10:(1):[20]: --back trace--
M2/Macaulay2/m2/validate.m2:74:26:(1):[19]: --back trace--
M2/Macaulay2/m2/validate.m2:83:45:(1):[18]: --back trace--
M2/Macaulay2/m2/validate.m2:83:5:(1):[17]: --back trace--
M2/Macaulay2/m2/validate.m2:40:33:(1):[16]: --back trace--
M2/Macaulay2/m2/validate.m2:42:5:(1):[14]: --back trace--
M2/Macaulay2/m2/validate.m2:62:29:(1):[12]: --back trace--
M2/Macaulay2/m2/validate.m2:42:5:(1):[10]: --back trace--
M2/Macaulay2/m2/validate.m2:49:5:(1):[8]: --back trace--
M2/Macaulay2/m2/installPackage.m2:280:27:(1):[7]: --back trace--
M2/Macaulay2/m2/installPackage.m2:269:52:(1):[5]: --back trace--
M2/Macaulay2/m2/last.m2:61:16:(1):[3]: --back trace--
Macaulay2/Core/startup.m2.in:71:58:(0):[2]: --back trace--
Macaulay2/Core/startup.m2.in:71:32:(0):[1]: --back trace--
```

mahrud:
What command did you enter? I don't see one here.

DanGrayson:
It's on line 1:
```
/Users/dan/src/M2/M2.git/M2/BUILD/dan/builds.tmp/gallium-1475/M2 --no-readline --print-width 196 --no-debug
```

mahrud:
I don't see anything about building the top level help here. Is this a fresh build?

DanGrayson:
I'm referring to the page that `viewHelp` displays:
Screen Shot 2020-09-25 at 6 58 13 PM

mahrud:
Should be fixed now.

M2/Macaulay2/m2/hypertext.m2
```diff
@@ -130,6 +130,8 @@ STRONG     = new MarkUpType of Hypertext
 SUB        = new MarkUpType of Hypertext
 SUP        = new MarkUpType of Hypertext
 TT         = new MarkUpType of Hypertext
+BF = BOLD
+IT = ITALIC
```
DanGrayson, line 134: These are not needed.

M2/Macaulay2/packages/Text.m2
```diff
@@ -6,8 +6,8 @@ newPackage("Text",
      )
 
 exportFrom_Core {
-     "ANCHOR", "BLOCKQUOTE", "BODY", "BOLD", "BR", "CDATA", "CODE", "COMMENT", "DD", "DIV", "DL", "DT", "EM", "ExampleItem", "HEAD", "HEADER1",
-     "HEADER2", "HEADER3", "HEADER4", "HEADER5", "HEADER6", "HR", "HREF", "HTML", "Hypertext", "HypertextContainer", "HypertextParagraph", "IMG", "ITALIC",
+     "ANCHOR", "BLOCKQUOTE", "BODY", "BOLD", "BF", "BR", "CDATA", "CODE", "COMMENT", "DD", "DIV", "DL", "DT", "EM", "ExampleItem", "HEAD", "HEADER1",
+     "HEADER2", "HEADER3", "HEADER4", "HEADER5", "HEADER6", "HR", "HREF", "HTML", "Hypertext", "HypertextContainer", "HypertextParagraph", "IMG", "ITALIC", "IT",
```
DanGrayson, line 10: remove BF and IT

DanGrayson (changes requested):

M2/Macaulay2/m2/exports.m2
```diff
@@ -60,6 +60,7 @@ export {
 	"AssociativeExpression",
 	"Authors",
 	"AuxiliaryFiles",
+	"BF",
```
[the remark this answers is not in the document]
mahrud: I'd rather keep this, as well as `IT`, because it makes substituting `\bf{...}` to `BF{...}` very straightforward:
https://github.com/mahrud/M2/blob/b6b561ba1155b9201b558873839737641f7e2d21/M2/Macaulay2/m2/hypertext.m2#L228-L237

I tried to do this another way without exporting BF and IT, perhaps with locally defining them, but it didn't work. If you have a suggestion to make that work then we can unexport them.

M2/Macaulay2/packages/Text.m2
mahrud: I think remembering `@BF "..."@` and `@IT "..."@` is much easier for users who know latex, but the reason I mentioned above I mentioned why I added them.

mahrud: > I think remembering `@BF "..."@` and `@IT "..."@` is much easier for users who know latex, but I mentioned above the reason why I added them.

M2/Macaulay2/packages/Macaulay2Doc/changes.m2
```diff
@@ -66,6 +66,10 @@ document {
 	       },
 	 LI { "functionality changed:",
 	      UL {
+		   LI {
+		       "$\\LaTeX$ typesetting for the web documentation is now rendered using ",
+		       HREF{"https://katex.org/","$\\KaTeX$"}, ", a JavaScript math typesetting library for browsers."
```
DanGrayson, line 71: Shouldn't this go in the section called "functionality improved:" and include information about how the appearance is better and more tex macros and features are supported?  For example, subscripts of subscripts presumably are now okay.

M2/Macaulay2/tests/normal/formatting.m2
```diff
@@ -8,3 +8,9 @@ assert (
 QQ[a][x]
 assert( toString (x-a) == "x-a" )
 assert( class expression (-a) === Minus )
+
+-- html formatting
+debug Core
```
DanGrayson, line 13: Why is this line needed?

DanGrayson (changes requested):

M2/Macaulay2/packages/Macaulay2Doc/changes.m2
mahrud: I think this is fine. Feel free to move it.

M2/Macaulay2/tests/normal/formatting.m2
mahrud: TEX is not exported by default, and `debug Core` is faster than `needsPackage "Text"`.

M2/Macaulay2/packages/Macaulay2Doc/changes.m2
DanGrayson: It's an important improvement, so its advantages need to be described and bragged about.  You're the one who can do that, not I.

mahrud: I prefer brief messages here rather than long ones, because longer messages are probably ignored, and I'm sure users will discover the improvements on their own very quickly.

Do you have a suggestion for what else the description should say? I'm trying to stay away from coding for a little bit, so please feel free to move it or add to it as you see fit.

DanGrayson: I wouldn't depend on users figuring that all out for themselves -- get them eager to try it.  And another destination for entries in our change log is our annual progress reports to the NSF, so it's worthwhile saying something that sounds positive rather than neutral.  What about something this, with the false bits corrected and more true bits added?:

$\\LaTeX$ typesetting for the web documentation is now rendered using ", HREF{"https://katex.org/","$\\KaTeX$"}, ", a JavaScript math typesetting library for browsers.  As a result, the subset of TeX that is correctly rendered in the HTML version of the documentation is substantially enlarged.  Matrices are now rendered as TeX would, rather than as stodgy old HTML tables.  Subscripts can now have subscripts, and fractions with horizontal bars work.  Many more characters in many more fonts are available.

mahrud:
@DanGrayson this is ready by the way.

PR #1481, opened 2020-08-29
fix printing in threads, part 1

DanGrayson:
some assertion statements have been put into stdio.d

Currently working on getting this code to work:
```m2
n = 0 ; schedule( () -> while true do (n=n+1; << "hi there hi there hi there hi there hi there hi there  " << n << endl << flush ));
```

The problem seems to be that the FOSS obtained by the master thread and the scheduled thread is the same, so somehow the "outindex" field in it is altered by the master thread when it prints the next prompt.  That happens after the scheduled thread has printed its message about 500 times, and thread context switching most often occurs while the scheduled thread is in the "write" system call.

The next step would be to check whether the two FOSS structures really are the same, perhaps by writing a little code so the address can be printed from top level.

@mahrud

### DIFF
--- a/M2/Macaulay2/d/Makefile.in
+++ b/M2/Macaulay2/d/Makefile.in
@@ -238,7 +238,7 @@ redo-startup:clean-startup.c startup.c cat-startup.c
 
 all: $(M2_LIBDEPS)
 tags: @srcdir@/TAGS
-@srcdir@/TAGS: $(M2_SRCFILES); cd @srcdir@ && @ETAGS@ $(M2_SRCFILES)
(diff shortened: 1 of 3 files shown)


mahrud:
I don't understand how the M2 code you shared above is intended to work. stdout is a resource which cannot be shared by multiple threads. There are three ways to handle this:
- only the main thread has access to stdout, other threads return an error on print
- stdout is protected by a mutex, so threads have to wait to get access
- every thread has direct access to stdout and the characters get jumbled together

The last option is clearly bad. The first option is the rather limiting, but easiest to get working. The middle one is the reasonable option, but I can't see any mutexes used in the current code.

DanGrayson:
Yes, it seems printing doesn't lock any mutexes, at least, that's what setting a breakpoint on `pthread_mutex_lock` indicates.  Probably I didn't finish developing the code and never managed to put that in.

PR #1495, opened 2020-09-11
This adds the RandomRationalPoints package, and SwitchingFields.  It also updates FastLinAlg to use RandomRationalPoints in some cases.

kschwede:

Issue #1181, opened 2020-05-21
missing documentation from other packages when installing Macaulay2Doc


mahrud:
Macaulay2Doc installs many documentation nodes from Core, but, strangely, it also gives warnings for missing docs from `SumsOfSquares`, `LLLBases`, `IntegralClosures`, and `null`. I managed to get rid of the SOS ones by removing Macaulay2Doc from `ImportedPackages` in `SumsOfSquares.m2`, but these persist:
```
--warning: tag has no documentation: Macaulay2Doc :: .Macaulay2/, key ".Macaulay2/", package null
--warning: tag has no documentation: Macaulay2Doc :: integralClosures(Ideal), key (integralClosures,Ideal), package IntegralClosure
--warning: tag has no documentation: Macaulay2Doc :: gramm(Matrix), key (gramm,Matrix), package LLLBases
--warning: tag has no documentation: Macaulay2Doc :: hermite(Matrix), key (hermite,Matrix), package LLLBases
--warning: tag has no documentation: Macaulay2Doc :: kernelLLL(Matrix), key (kernelLLL,Matrix), package LLLBases
```
Why is this happening? I can't find any reference to any of these nodes in Macaulay2Doc. Is it because `LLLBases` and `IntegralClosure` are preloaded? Even so, why would they tag along when Macaulay2Doc is being installed?

DanGrayson:
The first of those is the one you just fixed for me by adding quotation marks.

DanGrayson:
... ".Macaulay2/" is the value of the symbol "applicationDirectorySuffix".

mahrud:
I fixed? When?

DanGrayson:
Oops, not you, sorry.  It was @d-torrance at https://github.com/Macaulay2/M2/pull/1178#discussion_r428387580

DanGrayson:
See https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.15/share/doc/Macaulay2/Macaulay2Doc/html/___Ideal.html , where it shows this:

<img width="844" alt="Screen Shot 2020-05-20 at 9 09 46 PM" src="https://user-images.githubusercontent.com/700228/82515637-4fc5c600-9ade-11ea-806b-980676f895cb.png">

[...]

PR #1502, opened 2020-09-14
Documentation improvements

### LINKED ISSUE #23: automate the list of contributors
The page at http://www.math.illinois.edu/Macaulay2/Contributors/ should be generated in an automatic way from the current list of packages.  For hints, see how the page at http://www.math.illinois.edu/Macaulay2/Packages/ has been automated recently.

### LINKED ISSUE #1301: replace subversion references with github commit numbers
```
Binomials.m2:41:	     "repository code URI" => "svn://svn.macaulay2.com/Macaulay2/trunk/M2/Macaulay2/packages/Binomials.m2",
EdgeIdeals.m2:24:		"repository code URI" => "svn://svn.macaulay2.com/Macaulay2/trunk/M2/Macaulay2/packages/EdgeIdeals.m2",
HodgeIntegrals.m2:13:	"repository code URI" => "svn://svn.macaulay2.com/Macaulay2/trunk/M2/Macaulay2/packages/HodgeIntegrals.m2",
KustinMiller.m2:24:	     "repository code URI" => "svn://svn.macaulay2.com/Macaulay2/trunk/M2/Macaulay2/packages/KustinMiller.m2",
ModuleDeformations.m2:20:	  "repository code URI" => "svn://svn.macaulay2.com/Macaulay2/trunk/M2/Macaulay2/packages/ModuleDeformations.m2",
Nauty.m2:27:	 "repository code URI" => "svn://svn.macaulay2.com/Macaulay2/trunk/M2/Macaulay2/packages/Nauty.m2",
Normaliz.m2:29:		"repository code URI" => "svn://svn.macaulay2.com/Macaulay2/trunk/M2/Macaulay2/packages/Normaliz.m2",
NumericalAlgebraicGeometry.m2:29:	  "repository code URI" => "svn://svn.macaulay2.com/Macaulay2/trunk/M2/Macaulay2/packages/NumericalAlgebraicGeometry.m2",
OldPolyhedra.m2:22:	 "repository code URI" => "svn://svn.macaulay2.com/Macaulay2/trunk/M2/Macaulay2/packages/Polyhedra.m2",
OldToricVectorBundles.m2:28:	 "repository code URI" => "svn://macaulay2.math.uiuc.edu/Macaulay2/trunk/M2/Macaulay2/packages/ToricVectorBundles.m2",
PieriMaps.m2:22:	       "repository code URI" => "svn://svn.macaulay2.com/Macaulay2/trunk/M2/Macaulay2/packages/PieriMaps.m2",
SimplicialDecomposability.m2:25:	"repository code URI" => "svn://svn.macaulay2.com/Macaulay2/trunk/M2/Macaulay2/packages/SimplicialDecomposability.m2",
ToricVectorBundles.m2:28:	 "repository code URI" => "svn://macaulay2.math.uiuc.edu/Macaulay2/trunk/M2/Macaulay2/packages/ToricVectorBundles.m2",
VersalDeformations.m2:23:	 "repository code URI" => "svn://svn.macaulay2.com/Macaulay2/trunk/M2/Macaulay2/packages/VersalDeformations.m2",
```

Should be easy to do, by referring to the log, e.g.:
```
commit d814bd64858da074bc73c626d27ff5494dfb7d4f
Author: Daniel Grayson <dan@math.uiuc.edu>
Date:   Sat Aug 21 10:25:26 2010 +0000

    package Normaliz:
            update to version published by JSAG
            add certification
    
    
    git-svn-id: svn://svn.macaulay2.com/Macaulay2/trunk@11859 a8012f80-5f13-0410-8dc0-eb77bd940c0d
```

### LINKED ISSUE #1459: Keywords for packages
A feature that has been frequently requested at workshops is having keywords for packages and group them together in the list of packages.

I can include this as a part of #1454, which also adds a page for package authors. My current suggestion is to have `newPackage` accept a list of strings as the option Keyword. The first one would be used to group the packages together, the rest would only be displayed next to the package or on the top page of the package.

@DanGrayson @mikestillman how does this sound?

### LINKED ISSUE #1501: Set a length limit for Headline
> Yes, good idea, but why does it say "package null"? -- that's not informative.

If I could figure out why those calls to `document` are being processed, then I would figure out why `currentPackage` is null, but I can't figure out either.

On a related note, should "Headline" have a length limit? Pages like this look funny:
https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.16/share/doc/Macaulay2/IntegralClosure/html/_idealizer_lp..._cm_sp__Index_sp_eq_gt_sp..._rp.html

_Originally posted by @mahrud in https://github.com/Macaulay2/M2/issues/1181#issuecomment-670385672_

Response by Dan:
> Oh, my goodness, yes, let's set a length limit. The headline is intended to be a brief subheading used in menus and lists of links.

mahrud:
- [x] close #23 
  - [x] finish correcting duplicate names
  - [x] finish removing "contributing authors" and "special thanks"
- [x] close #1301 
  - [x] modify package documentation to link to original commit
- [x] close #1459 (finish adding keywords for packages)
- [x] close #1501 (added a warning that causes error when `debugLevel>0`)
- [x] mention `readPackage` in `changes.m2`
- [x] make sure installing `Complexes` package tells you that "DDD" is missing
- [x] Ways to use of `help toString` should not include undocumented items


mahrud:
@DanGrayson regarding [this comment of yours](https://github.com/Macaulay2/M2/issues/1181#issuecomment-674510102):

> > I think adding `--no-preload` when installing packages is the solution to this issue.
> 
> Even if it doesn't solve this issue, it's probably worth doing anyway, since the preloaded packages are there for the convenience of the ultimate user, not for the other packages.

The following examples of Macaulay2Doc fail with `--no-preload`, mostly because of Eliminate, PrimaryDecomposition, or Classic.
```m2
(degrees, Ring)
(indices, RingElement)
(lcm, MonomialIdeal)
(presentation, Module)
(symbol**, Matrix, RingElement)
(symbol/, Ideal, Function)
(symbol|, GradedModuleMap, GradedModuleMap)
(symbol||, GradedModuleMap, GradedModuleMap)
(syz, GroebnerBasis)
(syz, Matrix)
Elementary uses of Groebner bases I. Math 634 Fall 2005
FastNonminimal
LUdecomposition
MonomialIdeal
Schreyer orders
Singular Book 1.3.15
Singular Book 1.7.10
Singular Book 1.8.11
Singular Book 1.8.18
Singular Book 1.8.4
Singular Book 1.8.6
Singular Book 1.8.9
[associatedPrimes, Strategy]
[saturate, MinimalGenerators]
associated primes of an ideal
basis
elimination of variables
independentSets
minimalPrimes
partition
primary decomposition
solve
standardPairs
submatrixByDegrees
support
```

I can see a few solutions:
- move those functions to Core (don't like this)
- add the preloaded packages to the Macaulay2Doc package exports (don't like this either, since they would be loaded on hundreds of examples in Macaulay2Doc, until we make `capture` work)
- add "needsPackage" lines in the beginning of those examples
- do some sort of autoload thing (could run into exported symbol issues)
- forget about adding --no-preload, just get `capture` to work

Any thoughts?

DanGrayson:
> @DanGrayson regarding [this comment of yours](https://github.com/Macaulay2/M2/issues/1181#issuecomment-674510102):
> 

It's early in the morning!


> > > I think adding `--no-preload` when installing packages is the solution to this issue.
> > 
> > 
> > Even if it doesn't solve this issue, it's probably worth doing anyway, since the preloaded packages are there for the convenience of the ultimate user, not for the other packages.
> 
> The following examples of Macaulay2Doc fail with `--no-preload`, mostly because of Eliminate, PrimaryDecomposition, or Classic.
> 
> ```m2
> (degrees, Ring)
> (indices, RingElement)
> (lcm, MonomialIdeal)
> (presentation, Module)
> (symbol**, Matrix, RingElement)
> (symbol/, Ideal, Function)
> (symbol|, GradedModuleMap, GradedModuleMap)
> (symbol||, GradedModuleMap, GradedModuleMap)
> (syz, GroebnerBasis)
> (syz, Matrix)
> Elementary uses of Groebner bases I. Math 634 Fall 2005
> FastNonminimal
> LUdecomposition
> MonomialIdeal
> Schreyer orders
> Singular Book 1.3.15
> Singular Book 1.7.10
> Singular Book 1.8.11
> Singular Book 1.8.18
> Singular Book 1.8.4
> Singular Book 1.8.6
> Singular Book 1.8.9
> [associatedPrimes, Strategy]
> [saturate, MinimalGenerators]
> associated primes of an ideal
> basis
> elimination of variables
> independentSets
> minimalPrimes
> partition
> primary decomposition
> solve
> standardPairs
> submatrixByDegrees
> support
> ```
> 
> I can see a few solutions:
> 
> * move those functions to Core (don't like this)

I don't like that.

> * add the preloaded packages to the Macaulay2Doc package exports (don't like this either, since they would be loaded on hundreds of examples in Macaulay2Doc, until we make `capture` work)

There shouldn't be hundreds of examples that load Macaulay2Doc -- just the few that show example of "help" or something similar.

> * add "needsPackage" lines in the beginning of those examples

I don't like that because it will confuse the user.  The preloaded packages, from the user's point of view, are part of Macaulay2.

> * do some sort of autoload thing (could run into exported symbol issues)
> * forget about adding --no-preload, just get `capture` to work

I prefer forgetting about adding --no-preload.

> 
> Any thoughts?

Let's go back to my ideas in https://github.com/Macaulay2/M2/issues/1181#issuecomment-632018275 .

Here is a way I just thought of to get the hashcodes of the packages in the right order: we update the hashcode in the endPackage function.  To make a new Package object from an old one this code suffices: `new Package from new MutableHashTable from Foo`.  There shouldn't be any pointers to the old package lying around, but in case there are, it might be good instead to have "newPackage" make the package of type MutableHashTable instead of being of type Package, and then in "endPackage" we could do `new Package from Foo` instead.

mahrud:
> I prefer forgetting about adding --no-preload.

I don't like this as much, because I think packages that, say, required "elimination" should be forced to put "Elimination" in their PackageExports. Otherwise they'll break if the preloaded packages change at some point.

> Let's go back to my ideas in #1181 (comment) .

> Here is a way I just thought of to get the hashcodes of the packages in the right order: we update the hashcode in the endPackage function. To make a new Package object from an old one this code suffices: new Package from new MutableHashTable from Foo. There shouldn't be any pointers to the old package lying around, but in case there are, it might be good instead to have "newPackage" make the package of type MutableHashTable instead of being of type Package, and then in "endPackage" we could do new Package from Foo instead.

This is irrelevant here. That issue is already resolved and closed.

DanGrayson:
> > I prefer forgetting about adding --no-preload.
> 
> I don't like this as much, because I think packages that, say, required "elimination" should be forced to put "Elimination" in their PackageExports. Otherwise they'll break if the preloaded packages change at some point.

Right, but that's already true!  I did that not too long ago.

> 
> > Let's go back to my ideas in #1181 (comment) .
> 
> > Here is a way I just thought of to get the hashcodes of the packages in the right order: we update the hashcode in the endPackage function. To make a new Package object from an old one this code suffices: new Package from new MutableHashTable from Foo. There shouldn't be any pointers to the old package lying around, but in case there are, it might be good instead to have "newPackage" make the package of type MutableHashTable instead of being of type Package, and then in "endPackage" we could do new Package from Foo instead.
> 
> This is irrelevant here. That issue is already resolved and closed.

Then why do you care about "--no-preload"?

mahrud:
> Right, but that's already true! I did that not too long ago.

Okay, then adding `--no-preload` will speed up installing almost all packages, because they don't need to load a bunch of stuff that they don't use.

> Then why do you care about "--no-preload"?

I literally said it at the very beginning, regarding this [comment of yours](https://github.com/Macaulay2/M2/issues/1181#issuecomment-674510102):

> **Even if it doesn't solve this issue, it's probably worth doing anyway**, since the preloaded packages are there for the convenience of the ultimate user, not for the other packages.

DanGrayson:
> > Right, but that's already true! I did that not too long ago.
> 
> Okay, then adding `--no-preload` will speed up installing almost all packages, because they don't need to load a bunch of stuff that they don't use.

What about having `--no-preload` for all packages except for "Macaulay2Doc"?

> 
> > Then why do you care about "--no-preload"?
> 
> I literally said it at the very beginning, regarding this [comment of yours](https://github.com/Macaulay2/M2/issues/1181#issuecomment-674510102):
> 
> > **Even if it doesn't solve this issue, it's probably worth doing anyway**, since the preloaded packages are there for the convenience of the ultimate user, not for the other packages.

mahrud:
> What about having --no-preload for all packages except for "Macaulay2Doc"?

That could partially help, but still, Macaulay2Doc has hundreds of examples, more than any other package, each of which loads all the preloaded packages. And note that since Macaulay2Doc is installed before all the other packages, their documentation is not catched, so ALL of the packages are read, not just the stuff before `beginDocumentation`.

This would be resolved if we fixed capture, so I guess it's not a high priority.

mahrud:
> > I don't like this as much, because I think packages that, say, required "elimination" should be forced to put "Elimination" in their PackageExports. Otherwise they'll break if the preloaded packages change at some point.
> 
> Right, but that's already true! I did that not too long ago.

This is in fact not true! For instance, at least one example in the RandomRationalPoints package added in #1495 does not work with `--no-preload`.

DanGrayson:
> > > I don't like this as much, because I think packages that, say, required "elimination" should be forced to put "Elimination" in their PackageExports. Otherwise they'll break if the preloaded packages change at some point.
> > 
> > 
> > Right, but that's already true! I did that not too long ago.
> 
> This is in fact not true! For instance, at least one example in the RandomRationalPoints package added in #1495 does not work with `--no-preload`.

To see that it's true, install this package and then evaluate "foo":
```m2

i1 : loadPackage "Foo"

o1 = Foo

o1 : Package

i2 : foo

o2 = {Foo#"private dictionary", Foo.Dictionary, Core.Dictionary, OutputDictionary,
     PackageDictionary}

o2 : List
```

So you example must have failed for some other reason.

mahrud:
What? That didn't make any sense. Try this yourself:
```m2
needsPackage "RandomRationalPoints"
R = ZZ/5[x,y,z];
I = ideal(random(3,R)-2, random(2,R));
M = jacobian(I);
extendIdealByNonZeroMinor(2,M,I, Strategy => LinearIntersection)
T = ZZ/101[x1,x2,x3,x4,x5,x6,x7];
I =  ideal(x5*x6-x4*x7,x1*x6-x2*x7,x5^2-x1*x7,x4*x5-x2*x7,x4^2-x2*x6,x1*x4-x2*x5,x2*x3^3*x5+3*x2*x3^2*x7+8*x2^2*x5+3*x3*x4*x7-8*x4*x7+x6*x7,x1*x3^3*x5+3*x1*x3^2*x7+8*x1*x2*x5+3*x3*x5*x7-8*x5*x7+x7^2,x2*x3^3*x4+3*x2*x3^2*x6+8*x2^2*x4+3*x3*x4*x6-8*x4*x6+x6^2,x2^2*x3^3+3*x2*x3^2*x4+8*x2^3+3*x2*x3*x6-8*x2*x6+x4*x6,x1*x2*x3^3+3*x2*x3^2*x5+8*x1*x2^2+3*x2*x3*x7-8*x2*x7+x4*x7,x1^2*x3^3+3*x1*x3^2*x5+8*x1^2*x2+3*x1*x3*x7-8*x1*x7+x5*x7);
M = jacobian I;
i = 0;
J = I;
elapsedTime(while (i < 10) and dim J > 1 do (i = i+1; J = extendIdealByNonZeroMinor(4, M, J)) );
```
First with:
```
./M2 --stop -q -e errorDepth=0
```
Then with:
```
cat test.m2 | ./M2 --stop -q -e errorDepth=0 --no-preload
...
i11 : elapsedTime(while (i < 10) and dim J > 1 do (i = i+1; J = extendIdealByNonZeroMinor(4, M, J)) );
stdio:11:59:(3): error: key not found in hash table
     -- 0.788107 seconds elapsed
```
The error goes away if you add `PrimaryDecomposition` to the PackageExports, but I'm not sure why.

DanGrayson:
All my 3 line example demonstrates is that the symbols in the dictionaries of preloaded packages are not available while the package is loading.  Unfortunately, the various methods installed by preloaded packages, if they are attached to items that are in the Core package, or at least in other packages, are still available.  There's no way to easily remove them.  So the package could have a secret dependency on a preloaded package.

The problem there is that the generators of an object of type Ideal are stored inside it under the key `generators`, but the same is not true for objects of type MonomialIdeal, even though MonomialIdeal inherits methods from Ideal.  So without `--preload`, some method for monomial ideals must be missing, and the corresponding method for ideals is triggered instead.

The problem can be traced to this line in the MinimalPrimes package:
```m2
    comps := decompose monomialIdeal monoms;
```
Notice that both `decompose` and `MonomialIdeal` come from Core.  The preloaded package that provides the standard method for that is `PrimaryDecomposition`.  Thus MinimalPrimes has a secret dependency on PrimaryDecomposition.  The solution is to add the latter to the PackageImports of the former.
```m2
diff --git a/M2/Macaulay2/packages/MinimalPrimes.m2 b/M2/Macaulay2/packages/MinimalPrimes.m2
index 895f9aed4..7524d7120 100644
--- a/M2/Macaulay2/packages/MinimalPrimes.m2
+++ b/M2/Macaulay2/packages/MinimalPrimes.m2
@@ -15,7 +15,7 @@ newPackage(
             {Name => "Franziska Hinkelmann"}
             },
         AuxiliaryFiles=>true,
-        PackageImports => {"Elimination"}
+        PackageImports => {"Elimination","PrimaryDecomposition"}
         )
 
 -- << "warning!  This package is experimental.  The interface will change, and although" << endl;
```

mahrud:
> All my 3 line example demonstrates is that the symbols in the dictionaries of preloaded packages are not available while the package is loading.

Your 3 lines aren't even reproducible on my machine. What is "foo"?? Also, `PackageDictionary` is right there in your three lines, and if you look into it you'll see the preloaded packages. So I really don't see your point.

Thanks for the explanation but I understand what was missing, that wasn't the issue! (And please don't fix it now, there's a PR for that.) The issue is that missed dependencies on preloaded pakages currently goes unnoticed, and running of the examples is slowed down because all of those packages must be preloaded in order to run completely unrelated examples.

DanGrayson:
Oops, I forgot to give `Foo` to you:
```m2
newPackage ("Foo", Headline => "Foo")
export "foo"
foo = dictionaryPath
```

DanGrayson:
> The issue is that missed dependencies on preloaded pakages currently goes unnoticed, and running of the examples is slowed down because all of those packages must be preloaded in order to run completely unrelated examples.

That's what I said, right?

mahrud:
> That's what I said, right?

Sure, I was referring to your first comment.

By the way, at least ThreadedGB requires "Classic". Not sure what else needs it.

mahrud:
Looks like there are more and more missed dependencies, and not all because of secret dependencies: some simply have examples that need another package. Function names in examples appear as strings, so that might explain why you missed them.

mahrud:
@DanGrayson do you agree with the following rules of thumb:
- if package A uses something from package B in its examples, then it should include "B" in PackageExports, even if it is not used within the package itself.
- if a package is listed in PackageExports, it doesn't need to also be on PackageImports

DanGrayson:
If we're going to use `--no-preload` when running package examples, then we have to give package authors an effective way to find out which of the preloaded packages their package depends on.

Maybe using `--no-preload` is not a good idea.  After all, the preloaded packages are preloaded because we regard the functionality they provide as being part of the system.

DanGrayson:
> @DanGrayson do you agree with the following rules of thumb:
> 
> * if package A uses something from package B in its examples, then it should include "B" in PackageExports, even if it is not used within the package itself.

No, the author can just add `needsPackage "B"` to the example.

> * if a package is listed in PackageExports, it doesn't need to also be on PackageImports

Yes, because exported packages are also imported by "newPackage".  We don't have an option for _just_ exporting.

mahrud:
> > * if package A uses something from package B in its examples, then it should include "B" in PackageExports, even if it is not used within the package itself.
> 
> No, the author can just add `needsPackage "B"` to the example.

Sorry, I meant for preload packages. In that case, needsPackage would be confusing for readers, but I think authors should still list them.

Anyway, this doesn't seem feasible right now as there are too many missing dependencies. I reverted changes related to preload.

This PR is ready.

M2/Macaulay2/packages/Macaulay2Doc/overview_packages.m2
```diff
@@ -37,17 +37,19 @@ Node
...
+      The following people have authored packages that are distributed with Macaulay2:
     Code
-      authors := new HashTable from flatten apply(separate_" " version#"packages", pkgname -> (
+      authors := new MutableHashTable from flatten apply(separate_" " version#"packages", pkgname -> (
               pkgopts := readPackage pkgname;
               apply(pkgopts.Authors, author -> (
                       author = new OptionTable from author;
                       if author.?Name and (not authors#?(author.Name) or not instance(authors#(author.Name), HREF))
                       then author.Name => if author.?HomePage then HREF {author.HomePage, author.Name} else author.Name))));
+      scan(select(keys authors, match_{"(C|c)ontribut", "(M|m)aintain", "(A|a)uthor", "(T|t)hank", "(Michael|Mike) Stillman"}),
```
DanGrayson, line 48: It looks like you're removing Mike as author from all the packages he's an author of.

mahrud: No, "Michael E. Stillman" is still listed.

DanGrayson: What's the purpose of matching on Stillman, then?

mahrud: I don't want to exclude all other Mikes or Michaels. Is there a problem here?

DanGrayson: What is the purpose of removing authors matching "(Michael|Mike) Stillman" from the list of authors?

mahrud: Clearly we don't want Mike's name appearing in three forms in the list of authors. This removes two of them so the list of authors is not redundant. Of all the changes in this PR, is this the only thing you have questions about?

DanGrayson: Wouldn't it be more uniform if you would edit the occurrences of his name where he appears as an author, the way you do for everyone else?  As here:
```m2
diff --git a/M2/Macaulay2/packages/BGG.m2 b/M2/Macaulay2/packages/BGG.m2
index 28ec23e28..af0131a63 100644
--- a/M2/Macaulay2/packages/BGG.m2
+++ b/M2/Macaulay2/packages/BGG.m2
@@ -6,11 +6,12 @@ newPackage(
 	     {Name => "Hirotachi Abo", Email => "abo@uidaho.edu", HomePage => "http://www.webpages.uidaho.edu/~abo/"},
 	     {Name => "Wolfram Decker", Email => "decker@math.uni-sb.de", HomePage => "http://www.math.uni-sb.de/ag/decker/"},
 	     {Name => "David Eisenbud", Email => "de@msri.org", HomePage => "http://www.msri.org/~de/"},	     
-	     {Name => "Frank Schreyer", Email => "schreyer@math.uni-sb.de", HomePage => "http://www.math.uni-sb.de/ag/schreyer/"},
+	     {Name => "Frank-Olaf Schreyer", Email => "schreyer@math.uni-sb.de", HomePage => "http://www.math.uni-sb.de/ag/schreyer/"},
 	     {Name => "Gregory G. Smith", Email => "ggsmith@mast.queensu.ca", HomePage => "http://www.mast.queensu.ca/~ggsmith/"},
 	     {Name => "Mike Stillman", Email => "mike@math.cornell.edu", HomePage => "http://www.math.cornell.edu/~mike/"}
```

mahrud: This is a temporary patch, not a uniform change. Mike's name appears in 57 packages and he didn't have a clear preference to change them all into. If you have time, feel free to change them all when he has decided, but I wanted to get it done quickly and get to more important things.

DanGrayson: Oh, okay.   Meanwhile, the list of authors doesn't have all the information that the one I created manually does, so #23 should not be closed.  Compare:
--
Screen Shot 2020-10-08 at 8 51 41 AM
--
Screen Shot 2020-10-08 at 8 51 26 AM
--
Is that on your radar eventually?

mahrud: > Is that on your radar eventually?

No. I think that page should be removed because it excludes many many many contributors. The versions are too old and many links and emails don't even work anymore either.

I tried listing packages that people have contributed to, but this is not feasible anymore when there are ~200 packages. Some packages are repeated many times and some names will have way too many packages listed in front of them (e.g. Mike is listed in 57 packages!)

mahrud:
@DanGrayson I just pushed a commit that lists packages authored by each person if you hold your mouse over it. Hopefully that's enough to retire the old [contributors list](https://faculty.math.illinois.edu/Macaulay2/Contributors/).

DanGrayson (changes requested):

DanGrayson:
> @DanGrayson I just pushed a commit that lists packages authored by each person if you hold your mouse over it. Hopefully that's enough to retire the old [contributors list](https://faculty.math.illinois.edu/Macaulay2/Contributors/).

For now, but I would like the new one to look like the old one eventually.

M2/Macaulay2/m2/document.m2
```diff
@@ -521,7 +521,11 @@ getUsage := val -> (
     if #val === 0 then error "Usage: expected content";
     DL flatten { "class" => "element", DT "Usage: ", DD \ TT \ val } )
 
-getHeadline   := (val, key)   -> if instance(val, String) then fixup val else error("expected ", toString key, " option to be a string")
+getHeadline   := (val, key)   -> (
+    title := if instance(val, String) then fixup val else error("expected ", toString key, " option to be a string");
+    if #title > 100 then warning("document: documentation headlines must be less than 100 characters long:\n  " | format title);
+    if #title > 200 then   error("document: documentation headlines must be less than 100 characters long:\n  " | format title);
```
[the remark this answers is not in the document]
mahrud: Okay. I don't want to push a commit just to swap two lines, so I'll wait for your other comments and fix them together.

M2/Macaulay2/packages/Macaulay2Doc/overview_packages.m2
mikestillman: > This is a temporary patch, not a uniform change. Mike's name appears in 57 packages and he didn't have a clear preference to change them all into. If you have time, feel free to change them all when he has decided, but I wanted to get it done quickly and get to more important things.

I do want to have my name appear in only one way.  Perhaps "Michael Stillman"?  If you want, I can go change them all.

mahrud:
> For now, but I would like the new one to look like the old one eventually.

For the record, this wouldn't look good.

DanGrayson:
> > For now, but I would like the new one to look like the old one eventually.
> 
> For the record, this wouldn't look good.

For the record, it has more functionality.

M2/Macaulay2/m2/document.m2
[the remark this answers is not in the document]
DanGrayson: I'm done, and that's all I have.

mahrud:
> > > For now, but I would like the new one to look like the old one eventually.
> > 
> > 
> > For the record, this wouldn't look good.
> 
> For the record, it has more functionality.

What is the additional functionality? With mouseover you can tell who helped with what package.

The only things not included are:
- contributions to Core, which are outdated anyway, but have to be listed manually somewhere
- whether they "authored" or "coauthored" which is not that important imo.

DanGrayson:
> > > > For now, but I would like the new one to look like the old one eventually.
> > > 
> > > 
> > > For the record, this wouldn't look good.
> > 
> > 
> > For the record, it has more functionality.
> 
> What is the additional functionality? With mouseover you can tell who helped with what package.
> 
> The only things not included are:
> 
> * contributions to Core, which are outdated anyway, but have to be listed manually somewhere
> * whether they "authored" or "coauthored" which is not that important imo.

No, there's also being able to see the packages when viewing with "help" or "info", and also there're the links to the packages.

mahrud:
> No, there's also being able to see the packages when viewing with "help" or "info", and also there're the links to the packages.

No, that old page isn't even a documentation node, hence not available via help or info. I fail to see when someone might be interested in a package because someone contributed to it, but maybe what you really want is a node for each user, listing their contribution. This would be fine, but I think the right place for it is in a wiki style place where users can regularly make edits themselves rather than submitting pull requests.

Regardless, I think after this PR you should replace the link to the old contributors page, because it will confuse people and result in people not finding the right list of contributors.

DanGrayson:
Mahrud, can you reproduce this problem on your branch?

Former behavior:
```m2
+ /Applications/Macaulay2-1.16/bin/M2 --no-readline --print-width 74
Macaulay2, version 1.16
with packages: ConwayPolynomials, Elimination, IntegralClosure,
               InverseSystems, LLLBases, MinimalPrimes,
               PrimaryDecomposition, ReesAlgebra, TangentCone, Truncations

i1 : help "syz(Matrix)"

o1 = syz(Matrix) -- compute the syzygy matrix
     ****************************************

     +----------------------------------------------------+
     |i1 : R = QQ[a..g];                                  |
     +----------------------------------------------------+
     |i2 : I = ideal"ab2-c3,abc-cef,ade-cfg"              |
     |                                                    |
     |               2    3                               |
     |o2 = ideal (a*b  - c , a*b*c - c*e*f, a*d*e - c*f*g)|
     |                                                    |
     |o2 : Ideal of R                                     |
     +----------------------------------------------------+
     |i3 : syz gens I                                     |
     |                                                    |
     |o3 = {3} | -abc+cef 0        -ade+cfg de2f-bcfg  |  |
     |     {3} | ab2-c3   -ade+cfg 0        -c2de+b2fg |  |
     |     {3} | 0        abc-cef  ab2-c3   bc3-b2ef   |  |
     |                                                    |
     |             3       4                              |
     |o3 : Matrix R  <--- R                               |
     +----------------------------------------------------+

     See also
     ========

       * "gb" -- compute a Gröbner basis

o1 : DIV

i2 : 

Process M2<1> finished
```

Current behavior:
```m2
+ M2 --no-readline --print-width 90
Macaulay2, version 1.16.0.2
               MinimalPrimes, PrimaryDecomposition, ReesAlgebra, TangentCone, Truncations

i1 : help "syz(Matrix)"
 -- warning: there is no documentation for (syz,Matrix)

o1 = (syz,Matrix)
     ************



o1 : DIV

i2 : 

Process M2<1> finished
```

mahrud:
Seems to be a bug in (toExternalString, DocumentTag). I'll push a commit to fix it in a minute.

Issue #1504, opened 2020-09-18
global variable lastMatch


pzinn:
`lastMatch` is a global variable. this, we can all agree on, is bad.
One particular problem is that if output routines use `regex`, then the content of `lastMatch` will not be preserved (unless we do some ugly backup process which doesn't resolve the underyling thread-safety issue).
This is unfortunately the case in WebApp mode:
```
i9 : topLevelMode=WebApp

o9 =  WebApp 

o9 :   Symbol  

i10 : match ("asd", "--asd--")


o10 =   true  

i11 : lastMatch
```
(i.e., `lastMatch` is null instead of the expected answer).

I would urge to deprecate `lastMatch` and find some cleaner alternative.

pzinn:
incidentally the latest version of the `match` help example won't run at all in web app mode because it asks for `substring(first lastMatch, s)` (and `lastMatch` is always `null`).

mahrud:
It would be good to make this a caveat of `lastMatch`, but perhaps not deprecate it. Does webapp work if you don't use `match` and instead just use `regex`?

DanGrayson:
Yes, it's bad -- in particular, it's not thread safe.  It should go.

mahrud:
perhaps there should be a way to make top level variables thread local without going into `d`.

pzinn:
> It would be good to make this a caveat of `lastMatch`, but perhaps not deprecate it. Does webapp work if you don't use `match` and instead just use `regex`?

I'll look into it.

pzinn:
> > It would be good to make this a caveat of `lastMatch`, but perhaps not deprecate it. Does webapp work if you don't use `match` and instead just use `regex`?
> 
> I'll look into it.

As a workaround, replacing
```
htmlLiteral = s -> if s === null or not match("<|&|]]>|\42", s)l then s else (
```
with
```
htmlLiteral = s -> if s === null or regex("<|&|]]>|\42", s) === null then s else (
```
solves the webapp issue, yes. but not the underlying issue.

pzinn:
I think we all agree on the underlying issue, but it is unlikely to get fixed any time soon, since there are plenty of other thread-unsafe code in M2. closing.

DaveBarton:
[...]

Issue #1492, opened 2020-09-11
replace now a MethodFunctionWithOptions


pzinn:
apparently `replace` is now a `MethodFunctionWithOptions`. unfortunately this breaks existing code, e.g.,
```
i1: replace(0, a=>1, {b=>2,c=>3}) -- should (and used to) output {a=>1,c=>3}

stdio:34:1:(3): error: no method found for applying replace to:
     argument 1 :  0 (of class ZZ)
     argument 2 :  {b => 2, c => 3} (of class List)
```
any reason this change has been made?
because if so, a lot of code needs checking...

mahrud:
The reason was to take options for regex:
```m2
i2 : options (replace, String, String, String)

o2 = OptionTable{POSIX => false}
```
Also for `(select, String, String)` and `(select, , String, String, String)`.

It's annoying that parsing options is not done at the top level, but the interpreter, since the interpreter doesn't know what are the options for the method, otherwise `a=>1` would not be consumed as an option for a method that doesn't take any options.

pzinn:
sure, it's easy to fix, but there may be many such issues in the existing code.
an example:
```
i54 : style( SPAN { "hello", "style"=>"background:yellow" }, "font-weight" => "bold" )

stdio:57:1:(3): error: no method found for applying replace to:
     argument 1 :  1 (of class ZZ)
     argument 2 :  {hello, style => background:yellow} (of class List)
```

mahrud:
Yes, that's always the case with any change, which is why there should be tests written for all exported functions.

In any case, I don't think use of class `Option` in a list as opposed to an OptionTable or HashTable is correct, and probably all Hypertext objects should use option tables rather than a list of options. Then you can use `override` for changing an option rather searching in a list.

pzinn:
I don't disagree, the current internal structure of Hypertext objects is not great. But barring (another!) complete rewrite...
I will fix that particular instance of the issue (style) when I have time, and add a test as suggested.

DanGrayson:
To get this to work:
```m2
i2 : replace(0, a=>1, {b=>2,c=>3})

o2 = {a => 1, c => 3}
```
the way it used to, it would be sufficient to fix the way method functions with optional named arguments work, so that the named arguments have to come at the end.  Then here, where the object of type Option is in the middle, it would be dispatched as before.

This wouldn't be a great loss -- I think in the vast majority of the code written, the options go at the end of the list
of arguments, anyway.

pzinn:
[...]

PR #1507, opened 2020-09-20
workarounds for #1492 and #1504 and misc

pzinn:
- workaround for issue #1492 -- rewrote `style`. I'm still not too happy with `replace` having both a `Thing` argument and optional arguments.
A side effect is that I also had to slightly rewrite the part of `VectorGraphics` that called `style`, which means I threw in a couple of minor patches of `VectorGraphics`.
- took this opportunity to add doc for `style`.
- workaround for issue #1504 -- `htmlLiteral` now uses `regex` rather than `match`.
- minor `webapp` update.


M2/Macaulay2/packages/VectorGraphics.m2
```diff
@@ -1,7 +1,7 @@
 -- -*- coding: utf-8 -*-
 newPackage(
         "VectorGraphics",
-        Version => "0.9",
+        Version => "0.91",
```
mahrud, line 4: Could you also remove "A package to " from the headline, just so it matches other packages?

M2/Macaulay2/m2/html.m2
```diff
@@ -21,7 +21,7 @@ defaultHEAD = title -> HEAD splice { TITLE title, defaultCharset(), defaultStyle
 -----------------------------------------------------------------------------
 
 -- TODO: urlEncode
-htmlLiteral = s -> if s === null or not match("<|&|]]>|\42", s) then s else (
+htmlLiteral = s -> if s === null or regex("<|&|]]>|\42", s) === null then s else (
```
DanGrayson, line 24: Was this change needed for some reason??

DanGrayson:
> I'm still not too happy with `replace` having both a `Thing` argument and optional arguments.

Maybe we should discuss in a separate whether to go back to `replace` taking no options, for the sake of upward compatibility.  The only option is for POSIX regular expressions.

M2/Macaulay2/m2/html.m2
mahrud: #1504.

DanGrayson: Is this change related to `lastMatch`?  It doesn't appear in the code.

pzinn: `lastMatch` is modified by `match`. `htmlLiteral` is called in `WebApp` mode all the time (to convert text to HTML). That means `lastMatch` can't be used in `WebApp` mode. If `lastMatch` becomes deprecated that won't be an issue, but for now I followed Mahrud's suggestion to replace `match` with `regex` (which does the exact same except doesn't update `lastMatch`).

pzinn:
> > I'm still not too happy with `replace` having both a `Thing` argument and optional arguments.
> 
> Maybe we should discuss in a separate whether to go back to `replace` taking no options, for the sake of upward compatibility. The only option is for POSIX regular expressions.

sure, that's what #1492 is about.

M2/Macaulay2/m2/webapp.m2
```diff
@@ -3,28 +3,37 @@
...
+	)=webAppTags;
 
 texMathStart := "\\(";
 texMathEnd := "\\)";
-texWrap := x -> texMathStart | texMath x | texMathEnd; -- similar to 'tex', but only used by webapp.m2 to avoid thread-safety issues
+texWrap := x -> (  -- similar to 'tex', but only used by webapp.m2 to avoid thread-safety issues -- TODO: rewrite in a thread-safe way
+    y := texMath x;
+    if class y =!= String then error "invalid texMath output";
+    texMathStart | y | texMathEnd
+    )
```
DanGrayson, line 24: What aspect of this is not thread safe?

M2/Macaulay2/m2/webapp.m2
```diff
@@ -3,28 +3,37 @@
...
 
 htmlWithTex Thing := texWrap -- by default, we use tex (as opposed to html)
 
+stripTags := s -> replace(concatenate("[",webAppTags,"]"),"",s)
+
```
DanGrayson, line 29: You may wish to precompute the string `concatenate("[",webAppTags,"]")`.

M2/Macaulay2/packages/Text.m2
```diff
@@ -816,6 +816,16 @@ document { Key => {(validate, Hypertext),validate},
...
+document { Key => {(style, Hypertext),style},
+    Usage => "style (x,opts)",
+    Inputs => { "x" => { TO "hypertext" } },
+    Consequences => { { "Returns a new Hypertext object with its style option updated." }},
+    EXAMPLE lines ///
+    d = DIV { "Hello" }
+    d = style(d, "font-weight" => "bold")
+    peek d
+    ///
+    }
```
DanGrayson, line 828: I guess the code for `style` could also be in the Text package.

pzinn: Oops! Will fix that

M2/Macaulay2/m2/webapp.m2
pzinn: We discussed this during my previous PR (can't find the conversation for some reason); you're the one who convinced me non thread-safe code was bad. I'm merely writing a note to myself to eventually fix this; no actual change here.

M2/Macaulay2/packages/Text.m2
pzinn: edited: oh. I thought the comment was about the documentation. let me think about it carefully before I make this change.

M2/Macaulay2/packages/VectorGraphics.m2
pzinn: missed that. so what's the correct formulation? just "produce SVG graphics"? I'm confused

M2/Macaulay2/packages/Text.m2
pzinn: do you have an opinion Mahrud? since you're the one who restructured `hypertext.m2`.
In principle it would make sense to move `style` to `Text`, but it's a bit weird since at the moment `Text` is a completely empty package which only contains exports & documentation.

mahrud: I do not.

pzinn: OK then I'd rather not move style.

M2/Macaulay2/packages/VectorGraphics.m2
pzinn: I followed the `FirstPackage` template which says `an example Macaulay2 package`...

Issue #1509, opened 2020-09-21
Failing run program test


dylanpeifer:
A test in `tests/normal/programs.m2` is failing for me. The test itself fails on line 19 (when checking the output). I think a simpler version of the same failure is

```macaulay2
i1 : run("echo -n Hello World")
-n Hello World

o1 = 0
```

which ignores the `-n` flag that removes the trailing newline when I run `echo -n Hello World` on a separate terminal. I'm using `MacOS-10.15.6` and have compiled off the latest version of the master branch. I wasn't able to track down exactly what `run` is doing, though it looks like it is setup on line 1053 of `d/actors3.d` and might be imported from somewhere on line 117 in `d/system.d`. Any suggestions?

DanGrayson:
That's probably because you have your SHELL environment variable set to "/bin/sh", whose built-in "echo" command doesn't support "-n".

For predictability of the behavior of the "run" command, we should set SHELL to "/bin/sh" in the script "M2".

mahrud:
> For predictability of the behavior of the "run" command, we should set SHELL to "/bin/sh" in the script "M2".

This would not make it predictable, because on Fedora `/bin/sh` is a symlink to bash while on Ubuntu/Debian it's a symlink to dash, which has a smaller syntax.

d-torrance:
Since `echo` isn't portable, I've just opened #1510 switching this test to use `printf`, which is.

PR #1510, opened 2020-09-21
Use printf instead of echo for findProgram/runProgram test

d-torrance:
It is more portable.  From [1]:

> It is not possible to use echo portably across all POSIX systems
> unless both -n (as the first argument) and escape sequences are
> omitted.
> ...
> New applications are encouraged to use printf instead of echo.

Closes: #1509

[1] https://pubs.opengroup.org/onlinepubs/007904975/utilities/echo.html

### DIFF
--- a/M2/Macaulay2/tests/normal/programs.m2
+++ b/M2/Macaulay2/tests/normal/programs.m2
@@ -1,5 +1,5 @@
 fn = temporaryFileName()
-fn << "echo -n Hello $1!" << close
+fn << ///printf "Hello %s!" "$1"/// << close
 name = baseFilename fn
 dir = replace(name | "$", "", fn)
 programPaths#name = dir


DanGrayson:
A simple alternative would be to remove the "-n".

d-torrance:
> A simple alternative would be to remove the "-n".

Yeah, but that would have meant changing the expected behavior of the tests.  Both of the following would need newlines:

https://github.com/Macaulay2/M2/blob/837c8c6c4a2c64ed7415d485c5682416b11b77a1/M2/Macaulay2/tests/normal/programs.m2#L19

https://github.com/Macaulay2/M2/blob/837c8c6c4a2c64ed7415d485c5682416b11b77a1/M2/Macaulay2/tests/normal/programs.m2#L22

I can switch it if you'd prefer.

DanGrayson:
> I can switch it if you'd prefer.

No, it's okay, as long as "printf" is found everywhere.

d-torrance:
> No, it's okay, as long as "printf" is found everywhere.

It should be.  From https://en.wikipedia.org/wiki/Printf_(Unix):

> `printf` is part of the X/Open Portability Guide since issue 4 of 1992.

Issue #1439, opened 2020-07-27
Normaliz output file formats


d-torrance:
From https://github.com/Normaliz/Normaliz/commit/3661d6d6e9b022181a22a35dc4fcda041649bc65 (included in 3.4.0) until https://github.com/Normaliz/Normaliz/commit/6a7a243d6497f97ed074c80e11d85acb9ec30216 (included in 3.5.2), Normaliz didn't include row and column numbers when it output matrices.  In particular, version 3.5.1, which is the version found in Ubuntu 18.04 (EOL April 2028), falls in this window.  For example:

```console
profzoom@orange:~/tmp$ /usr/bin/normaliz --version
Normaliz 3.5.1
Copyright (C) 2007-2017  The Normaliz Team, University of Osnabrueck.
This program comes with ABSOLUTELY NO WARRANTY; This is free software,
and you are welcome to redistribute it under certain conditions;
See COPYING for details.
profzoom@orange:~/tmp$ /usr/bin/normaliz -f foo
profzoom@orange:~/tmp$ cat foo.gen
0 1 0 
1 0 0 
profzoom@orange:~/tmp$ /usr/libexec/Macaulay2/bin/normaliz --version
Normaliz 3.7.2
Copyright (C) 2007-2018  The Normaliz Team, University of Osnabrueck.
This program comes with ABSOLUTELY NO WARRANTY; This is free software,
and you are welcome to redistribute it under certain conditions;
See COPYING for details.
profzoom@orange:~/tmp$ /usr/libexec/Macaulay2/bin/normaliz -f foo
profzoom@orange:~/tmp$ cat foo.gen
2
3
0 1 0 
1 0 0 
```
The `Normaliz` package in Macaulay2 expects the latter format and fails miserably when encountering the former, e.g.,:

```m2
...

--making example results for diagInvariants
[...]

PR #1513, opened 2020-09-25
Alternate fix for differences between normaliz output file formats

d-torrance:
As reported in #1439, depending on the version, normaliz may or may not include the number of rows and columns in a matrix in its output.  Currently, the `Normaliz` package expects these numbers to be present.  If using an older normaliz (e.g., from Ubuntu 18.04, which is an LTS release), then the package fails.

One proposal, used in #1417, is to just require a new enough version of normaliz to guarantee the expected format.  In this pull request, which contains many of the same commits as #1417 (in particular, adding a `MinimumVersion` option to `findProgram` and porting the `Normaliz`package to use `findProgram` and `runProgram`), we also modify the matrix-parsing algorithm so that it will work with either output format and we don't need to worry about the version of normaliz (aside from requiring at least version 2.11, which was already the case).

If this solution is preferable, then I'll close #1417.  Otherwise, I'll close this.

### DIFF
--- a/M2/Macaulay2/m2/exports.m2
+++ b/M2/Macaulay2/m2/exports.m2
@@ -255,6 +255,7 @@ export {
 	"MinimalGenerators",
 	"MinimalMatrix",
 	"Minimize",
+	"MinimumVersion",
 	"Minus",
 	"Module",
 	"ModuleMap",
--- a/M2/Macaulay2/m2/programs.m2
+++ b/M2/Macaulay2/m2/programs.m2
@@ -10,14 +10,24 @@ addSlash = programPath -> (
     else return programPath
 )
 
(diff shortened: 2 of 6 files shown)


DanGrayson:
This looks okay to me.

@csoeger - Do you and Gesa agree with these changes?

DanGrayson:
Doug, which of the two pull requests do you prefer?  I don't have an opinion yet.

d-torrance:
> Doug, which of the two pull requests do you prefer? I don't have an opinion yet.

I prefer this one since it lets us build Macaulay2 on Ubuntu 18.04 using the normaliz package that's already available.

DanGrayson:
It looks like we're not in contact with the authors, so I'll pull.

Issue #1423, opened 2020-07-22
integralClosure out of memory example error on i386


d-torrance:
On a test build of the Debian package in an i386 Debian unstable chroot, I got the following:

```
--making example results for integralClosure(..., Strategy => ...)
 ulimit -c unlimited; ulimit -t 700; ulimit -m 850000; ulimit -s 8192; ulimit -n 512;  cd /tmp/M2-32642-0/13-rundir/; GC_MAXIMUM_HEAP_SIZE=400M "/build/macaulay2-1.16+git55.94c4b7d+ds/M2/usr-dist/i686-Linux-Debian-unknown/bin/M2-binary" -q --int --no-randomize --no-readline --silent --stop --print-width 77 -e 'needsPackage("IntegralClosure", Reload => true, FileName => "/build/macaulay2-1.16+git55.94c4b7d+ds/M2/Macaulay2/packages/IntegralClosure.m2")' <"/tmp/M2-32642-0/0_integral__Closure_lp..._cm_sp__Strategy_sp_eq_gt_sp..._rp.m2" >>"/build/macaulay2-1.16+git55.94c4b7d+ds/M2/usr-dist/common/share/doc/Macaulay2/IntegralClosure/example-output/_integral__Closure_lp..._cm_sp__Strategy_sp_eq_gt_sp..._rp.errors" 2>&1
/build/macaulay2-1.16+git55.94c4b7d+ds/M2/usr-dist/common/share/doc/Macaulay2/IntegralClosure/example-output/_integral__Closure_lp..._cm_sp__Strategy_sp_eq_gt_sp..._rp.errors:0:1: (output file) error: Macaulay2 exited with status code 1
/tmp/M2-32642-0/0_integral__Closure_lp..._cm_sp__Strategy_sp_eq_gt_sp..._rp.m2:0:1: (input file)
M2: *** Error 1
--making example results for ringFromFractions
--making example results for icFractions
--making example results for integralClosure(Ideal,RingElement,ZZ)
--making example results for integralClosure(..., Verbosity => ...)
--making example results for icPIdeal
--making example results for integralClosure(..., Variable => ...)
currentString:1:1:(3):[2]: error: 1 error(s) occurred running examples for package IntegralClosure
[...]

PR #1537, opened 2020-10-08
Use example ideal from earlier version of IntegralClosure

d-torrance:
The newer example, introduced in f1cea48, was causing memory errors on
i386 systems.  The examples run without problems after reverting to the
ideal that had been used previously.

Closes: #1423

### DIFF
--- a/M2/Macaulay2/packages/IntegralClosure.m2
+++ b/M2/Macaulay2/packages/IntegralClosure.m2
@@ -1343,40 +1343,40 @@ doc ///
      to instead compute Hom(J^-1, J^-1).  This is usually a more time consuming
      computation, but it does potentially get to the answer in a smaller number of steps.
    Example
-     S = QQ[x,y]
-     f = ideal (y^4-2*x^3*y^2-4*x^5*y+x^6-x^7)
+     S = QQ[x,y,z]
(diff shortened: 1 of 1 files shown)


DanGrayson:
@mikestillman -- is this okay with you?

mahrud:
I didn't follow the issue earlier, but isn't this a bug that should be fixed?

mikestillman:
> The newer example, introduced in [f1cea48](https://github.com/Macaulay2/M2/commit/f1cea48e204bf4534ea5f76f81369c61ec355bca), was causing memory errors on
> i386 systems. The examples run without problems after reverting to the
> ideal that had been used previously.
> 
> Closes: #1423

@d-torrance What kind of memory errors?  Actual ones, or ulimit based?

d-torrance:
> @d-torrance What kind of memory errors? Actual ones, or ulimit based?

All the info is over in #1423.  It's been a couple months, but I think I was able to reproduce them without doing anything with ulimit.

d-torrance:
> I didn't follow the issue earlier, but isn't this a bug that should be fixed?

Yeah, the underlying issue should totally be fixed.  This commit is basically a band-aid to prevent build failures on i386 because of these particular examples.

PR #1544, opened 2020-10-18
Add --no-preload for all examples, except in Macaulay2Doc

### LINKED ISSUE #1543: Examples being run with --no-preload
Maybe there's already an issue out there mentioning these problems, but since I haven't found any, here it is.
Some examples seem to rely on certain packages. Here's one: `resolution-doc.m2` tries
`ideal"a3-ab-c2,abc-d2-1, b3-b2-b"` which fails with
```
currentString:10:5:(3):[25]: error: no method found for applying ideal to:
     argument   :  "a3-ab-c2,abc-d2-1, b3-b2-b" (of class String)
```
because the package `Classic` has not been loaded.

mahrud:
- [x] always preload classic
- [x] only load other preloaded packages for Macaulay2Doc examples and tests (close #1543)
- [ ] fix missing dependencies in other packages


mahrud:
Closing this, will eventually include it in #1546.

Issue #963, opened 2019-06-27
'help "RingElement"' fails after defining a polynomial ring


jkyang92:
From the command line, running the following sequence

```
QQ[x]
help "RingElement"
```
Produces the following error 

> stdio:2:1:(3): error: makeDocumentTag: package cannot be determined: 'o1' : here is the first use of 'o1'

Poking around in the documentation code suggests that this error occurs when trying to populate the "Types of ring element" subsection in the documentation, since o1 is now an subtype of the RingElement type, but the code can't figure out what package o1 is from.

Amusingly, if you instead run,

```
QQ[x]
o1 = null
help "RingElement"

```
there is no error, and both oo and ooo show up in the the "Types of ring element", which seems odd since it would suggest that the documentation code is perfectly happy with oo and ooo, but not o1.

DanGrayson:
Thanks!  That's a bug.
```
+ M2 --no-readline --print-width 111
Macaulay2, version 1.14
--loading configuration for package "FourTiTwo" from file /Users/dan/Library/Application Support/Macaulay2/init-FourTiTwo.m2
--loading configuration for package "Topcom" from file /Users/dan/Library/Application Support/Macaulay2/init-Topcom.m2
with packages: ConwayPolynomials, Elimination, IntegralClosure, InverseSystems, LLLBases, PrimaryDecomposition,
               ReesAlgebra, TangentCone, Truncations

i1 : QQ[x]

o1 = QQ[x]

o1 : PolynomialRing

i2 : help "RingElement"
stdio:2:1:(3): error: makeDocumentTag: package cannot be determined: 'o1'
: here is the first use of 'o1'

i3 : errorDepth = 0

o3 = 0

i4 : help "RingElement"
/Applications/Macaulay2-1.14/share/Macaulay2/Core/debugging.m2:20:6:(1):[14]: error: makeDocumentTag: package cannot be determined: 'o1'
: here is the first use of 'o1'
/Applications/Macaulay2-1.14/share/Macaulay2/Core/debugging.m2:20:6:(1):[14]: --entering debugger (type help to see debugger commands)
/Applications/Macaulay2-1.14/share/Macaulay2/Core/debugging.m2:20:6-20:27: --source code:
     olderror processArgs args)

ii5 : 
/Applications/Macaulay2-1.14/share/Macaulay2/Core/document.m2:171:27:(1):[13]: --entering debugger (type help to see debugger commands)
[...]

PR #1546, opened 2020-10-18
Fix capture + several minor bug fixes

### LINKED ISSUE #1480: Speeding up running Macaulay2Doc examples
Once #1469 is fixed and #1454 is merged, we can run examples from Macaulay2Doc in the same process. This would cut at least 35 minutes from the install time (there are 1185 documentation keys, each starting a new M2 process which takes at least 1.8s on my relatively fast machine).
 
However, even after that some of the tests seem to take a lot longer, and it would be good to reduce them. For instance, here is the line that contributes to the fifth longest example from the node `"communicating with programs"`, which takes about 8s:
https://github.com/Macaulay2/M2/blob/65332ed07101a51255182fb5a16e4e6e20809292/M2/Macaulay2/packages/Macaulay2Doc/overview2.m2#L826

Here is the top percentile of examples by running time:
```m2
16.5518 seconds elapsed -- isPseudoprime(ZZ)
10.88 seconds elapsed -- minimalBetti(Ideal)
9.45877 seconds elapsed -- FastNonminimal
8.25365 seconds elapsed -- benchmark
8.0775 seconds elapsed -- communicating with programs
6.79511 seconds elapsed -- computing resolutions
6.21707 seconds elapsed -- betti(..., Minimize => ...)
4.50856 seconds elapsed -- quotient(..., Strategy => ...)
4.07333 seconds elapsed -- variety(Ring)
3.90818 seconds elapsed -- cancelTask(Task)
3.83857 seconds elapsed -- parallel programming with threads and tasks
3.65652 seconds elapsed -- Tutorial: Divisors
3.51705 seconds elapsed -- HH^ZZ CoherentSheaf
3.17934 seconds elapsed -- randomKRationalPoint
3.1685 seconds elapsed -- computing Groebner bases
3.09443 seconds elapsed -- elimination of variables
3.03655 seconds elapsed -- Tutorial: Fano varieties
```

PS: parallelization with `capture` is probably going to take a lot more work, since the task supervisor needs fixes before it can handle IO access in parallel. That's not the topic of this issue, however.

mahrud:
This PR began with the intention of reducing the time it takes to install Macaulay2Doc from >1hr to <15min, but that feature is disabled temporarily until this line is removed:
https://github.com/Macaulay2/M2/blob/073fca9783f346179055e6178eaf1b11a6433cd1/M2/Macaulay2/m2/examples.m2#L139


mahrud:
It went decently well, till this happened:
```m2
 -- capturing example results for index
../../../../Macaulay2/m2/methods.m2:37:35:(1):[15]: error: no method found for applying index to:
     argument   :  a (of class Symbol)
../../../../Macaulay2/m2/methods.m2:133:19:(1):[14]: --back trace--
currentString:2:1:(3):[13]: --back trace--
 -- making example results for index
 -- capturing example results for forceGB(..., SyzygyMatrix => ...)

   -- registering polynomial ring 212 at 0x7fd17524c100

   -- registering gb 312 at 0x7fd174f89540

   -- [gb]{2}(1)m{3}(1)m{4}(1)m{5}(1)z{6}(1)z{7}(1)znumber of (nonminimal) gb elements = 3
   -- number of monomials                = 9
   -- #reduction steps = 6
   -- #spairs done = 6
   -- ncalls = 0
   -- nloop = 0
   -- nsaved = 0
   --  -- capturing example results for Symbol _ Ring

   -- registering polynomial ring 213 at 0x7fd17524c000

   -- registering polynomial ring 214 at 0x7fd174d3af00
 -- capturing example results for pairs
 -- capturing example results for end

 -- removing polynomial ring 76 at 0x7fd175084d00
--loaded ../../../../Macaulay2/packages/Macaulay2Doc/demo3.m2
--loaded ../../../../Macaulay2/packages/Macaulay2Doc/demo1.m2
../../../../Macaulay2/packages/Macaulay2Doc/demo1.m2:8:12:(3):[15]: error: division by zero
../../../../Macaulay2/packages/Macaulay2Doc/demo1.m2:8:12:(3):[15]: --entering debugger (type help to see debugger commands)

i5 : 
```

mahrud:
This error appears at the very end of installing `Macaulay2Doc`, but running it again doesn't produce any errors!
```m2
../../../../Macaulay2/m2/debugging.m2:20:6:(1):[12]: error: internal error: documentation node not processed yet: Grassmannian2
../../../../Macaulay2/m2/document.m2:349:10:(1):[11]: --back trace--
../../../../Macaulay2/m2/installPackage.m2:434:21:(1):[10]: --back trace--
../../../../Macaulay2/m2/installPackage.m2:419:5:(1):[9]: --back trace--
../../../../Macaulay2/m2/installPackage.m2:753:31:(1):[8]: --back trace--
../../../../Macaulay2/m2/methods.m2:119:80:(1):[7]: --back trace--
../../../../Macaulay2/m2/option.m2:16:8:(1):[6]: --back trace--
../../../../Macaulay2/m2/installPackage.m2:572:5:(1):[5]: --back trace--
../../../../Macaulay2/m2/methods.m2:119:80:(1):[4]: --back trace--
../../../../Macaulay2/m2/option.m2:16:8:(1):[3]: --back trace--
currentString:1:1:(3):[2]: --back trace--
../../../../Macaulay2/m2/debugging.m2:20:6:(1):[3]: error: package not current: User
../../../../Macaulay2/m2/packages.m2:455:76:(1):[2]: --back trace--
../../../../Macaulay2/m2/packages.m2:274:18:(1):[1]: --back trace--

1 :
```
The entire process takes under 10 minutes.

mahrud:
@DanGrayson I think this should be working now, except `capture` isn't storing stderr. Do you know what is the change required to make that work?

DanGrayson:
> @DanGrayson I think this should be working now, except `capture` isn't storing stderr. Do you know what is the change required to make that work?

Yes, somehow we have to set it up so that the output sent to either stdError or to stdIO is buffered in the same buffer, for use by the "capture" function.

mahrud:
@DanGrayson do you have any thoughts on this PR so far? In particular, I had to add a `-* no-capture-flag *-` to some examples to avoid them being run with capture. _Eventually_ I'd like that to be unnecessary, but I think the remaining action in this PR is getting stderr to be captured.

DanGrayson:
Well, how many examples actually print to stderr?  Probably not many, and you could find them and mark them, too.  Just search for "stderr", and also mark the ones that demonstrate the use of the debugger after an error.

By the way, is the no-capture-flag filtered out so it doesn't appear in the documentation?

mahrud:
> Well, how many examples actually print to stderr? Probably not many, and you could find them and mark them, too. Just search for "stderr", and also mark the ones that demonstrate the use of the debugger after an error.

That was the cause of most of the ones that I had to exclude.

> By the way, is the no-capture-flag filtered out so it doesn't appear in the documentation?

No, but I can do it.

DanGrayson:
> > By the way, is the no-capture-flag filtered out so it doesn't appear in the documentation?
> 
> No, but I can do it.

Something like that would be good.  What about instead letting EXAMPLE be a function with options, and introducing an option, such as `Capture => false` ?

mahrud:
> Something like that would be good. What about instead letting EXAMPLE be a function with options, and introducing an option, such as `Capture => false` ?

I'd prefer to eventually make capture work for everything. I don't see any intrinsic value in wanting an example to be run externally.

DanGrayson:
> > Something like that would be good. What about instead letting EXAMPLE be a function with options, and introducing an option, such as `Capture => false` ?
> 
> I'd prefer to eventually make capture work for everything. I don't see any intrinsic value in wanting an example to be run externally.

Okay, good point.  But, hmm, do these examples still work when capturing?: https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.16/share/doc/Macaulay2/Macaulay2Doc/html/_the_spdebugger.html

mahrud:
> Okay, good point. But, hmm, do these examples still work when capturing?: https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.16/share/doc/Macaulay2/Macaulay2Doc/html/_the_spdebugger.html

Not currently, which is why I added the no-capture-flag for it. But once a few issues are resolved, I don't see why it wouldn't?

DanGrayson:
> > By the way, is the no-capture-flag filtered out so it doesn't appear in the documentation?
> 
> No, but I can do it.

... waiting for this update.

mahrud:
> > > By the way, is the no-capture-flag filtered out so it doesn't appear in the documentation?
> > 
> > 
> > No, but I can do it.
> 
> ... waiting for this update.

I'm waiting for capturing stderr to work. Without it, you'd have to check all packages to see if any examples print errors that should be captured.

DanGrayson:
> > > > By the way, is the no-capture-flag filtered out so it doesn't appear in the documentation?
> > > 
> > > 
> > > No, but I can do it.
> > 
> > 
> > ... waiting for this update.
> 
> I'm waiting for capturing stderr to work. Without it, you'd have to check all packages to see if any examples print errors that should be captured.

Why wait for that?  What I meant was, that while processing the documentation, we strip such comments out, so they don't appear in the documentation, because the comments will confuse the readers.

mahrud:
> Why wait for that? What I meant was, that while processing the documentation, we strip such comments out, so they don't appear in the documentation, because the comments will confuse the readers.

That's a one line change, but this PR isn't complete even with that. Currently the documentation for any node that intentionally prints to stderr is incomplete, and it's too much work and maintenance to regularly check which examples print to stderr.

DanGrayson:
> > Why wait for that? What I meant was, that while processing the documentation, we strip such comments out, so they don't appear in the documentation, because the comments will confuse the readers.
> 
> That's a one line change, but this PR isn't complete even with that. Currently the documentation for any node that intentionally prints to stderr is incomplete, and it's too much work and maintenance to regularly check which examples print to stderr.

If the PR isn't complete, why pull it?  Won't that make the documentation of nodes that print to stderr also look bad?

mahrud:
I never said pull this now! It's not done!

DanGrayson:
> I never said pull this now! It's not done!

Oops, okay.

mahrud:
@DanGrayson I think this is ready now, actually

DanGrayson:
It's too noisy:
```m2
 -- capturing example results for listUserSymbols
     -- 0.0326315 seconds elapsed
 -- capturing example results for quotient(..., MinimalGenerators => ...)
     -- 0.0207978 seconds elapsed
 -- making example results for communicating with programs
     -- 6.26688 seconds elapsed
 -- capturing example results for standardPairs
     -- 0.118061 seconds elapsed
 -- capturing example results for maps between chain complexes
     -- 0.0485607 seconds elapsed
```

Remove the time commands.

mahrud:
> It's too noisy:
> Remove the time commands.

It's useful information, it should stay. Even in your examples, it's important to know this:
```m2
 -- making example results for communicating with programs
     -- 6.26688 seconds elapsed
```

mikestillman:
Or perhaps put the time command output on the same line, so each example is only one line long in the output?

mahrud:
> Or perhaps put the time command output on the same line, so each example is only one line long in the output?

That would be a good idea, but I don't know how to do that.

mahrud:
Actually, maybe this PR isn't quite ready yet:
```m2
i9 : last capture "print random(ZZ^2,ZZ^2)"

o9 =                             | 4 0 |
     i1 : print random(ZZ^2,ZZ^2)| 0 8 |


     i2 : 
```
I'm guessing not too many examples in the documentation print a Net, but I don't know how to fix this in the interpreter, so I'll leave it for @DanGrayson to fix.

DanGrayson:
> > It's too noisy:
> > Remove the time commands.
> 
> It's useful information, it should stay. Even in your examples, it's important to know this:
> 
> ```m2
>  -- making example results for communicating with programs
>      -- 6.26688 seconds elapsed
> ```

Why do you think it is important?

DanGrayson:
It's not just nets, it's any printing:
```m2

i6 : last capture /// print "hi there" ///

o6 =                        hi there
     i1 :  print "hi there" 


     i2 : 
```

M2/Macaulay2/m2/examples.m2
```diff
@@ -39,8 +39,9 @@ capture String := s -> (
     User#"private dictionary" = new Dictionary;
     OutputDictionary = new GlobalDictionary;
     dictionaryPath = {
-	currentPackage.Dictionary,
 	User#"private dictionary",
+	currentPackage.Dictionary,
```
DanGrayson, line 43: I'm suspicious of the idea of setting `currentPackage` while running examples.  Its purpose is to inform the code which package is currently being loaded, so documentation and tests can get stored inside the package.  Thus, if you have an example whose purpose is to demonstrate documentation or tests, that junk will get stored inside the package, instead of inside `User`, where its presence will not be noticed.

mahrud: Where am I setting currentPackage?

DanGrayson: Hmm, you're right: I thought you were, for some reason.

mahrud: Oh, I can unset it right before calling capture though. That might help remove a few `no-capture-flag`ged nodes. Sorry if you meant this I misunderstood your point.

DanGrayson: No, I was just confused.

But actually, by the way, isn't CurrentPackage always set to User when examples are being run with "capture"?  So it seems that adding `currentPackage.Dictionary` to the path serves no purpose.

mahrud: It depends. When you run capture from the interpreter, it is set to User, but when you run it while installing a package, it is set to that package, so it does serve a purpose.

mahrud:
> It's not just nets, it's any printing:

Yes, seems like `print` and `File << Thing` somehow get captured incorrectly, or their output isn't combined with the current output stored in the buffer correctly. Do you have any ideas for fixing it?

Also, for capturing stderr, is it possible to redirect stderr to stdout from the interpreter temporarily?

M2/Macaulay2/m2/examples.m2
DanGrayson: Ah, okay.

DanGrayson:
Maybe try this:
```diff
diff --git a/M2/Macaulay2/d/interp.dd b/M2/Macaulay2/d/interp.dd
index d88d21a57..d38e7a83c 100644
--- a/M2/Macaulay2/d/interp.dd
+++ b/M2/Macaulay2/d/interp.dd
@@ -529,6 +529,9 @@ capture(e:Expr):Expr := (
      when e
      is s:stringCell do (
      	  flush(stdIO);
+	  flush(stdError);
+	  oldStdError := stdError;
+	  stdError = stdIO;
 	  oldDebuggingMode := debuggingMode;
 	  setDebuggingMode(false);
           foss := getFileFOSS(stdIO);
@@ -552,6 +555,7 @@ capture(e:Expr):Expr := (
 	  foss.outindex = oldindex;
 	  setLineNumber(oldLineNumber);
 	  setDebuggingMode(oldDebuggingMode);
+	  stdError = oldStdError;
 	  previousLineNumber = -1;
 	  Expr(Sequence( when r is err:Error do True else False, toExpr(out) )))
      else WrongArg(1,"a string"));
```

mahrud:
> Maybe try this:

Didn't work:

```m2
i1 : capture "stderr << \"hi\" << endl"
hi

o1 = (false,                            )
             i1 : stderr << "hi" << endl

             o1 = stderr

             o1 : File

             i2 : 


o1 : Sequence

i2 : capture "error \"hi\""
currentString:1:1:(3):[1]: error: hi

o2 = (true,                )
            i1 : error "hi"


o2 : Sequence
```

M2/Macaulay2/m2/examples.m2
```diff
@@ -28,9 +28,35 @@ capture List   := s -> capture demark_newline s
...
+    OutputDictionary = new GlobalDictionary;
+    dictionaryPath = {
+	User#"private dictionary",
+	currentPackage.Dictionary,
+	oldPrivateDictionary,
+	Core.Dictionary,
+	OutputDictionary,
+	PackageDictionary};
+    scan(Core#"pre-installed packages", needsPackage);
+    scan(currentPackage.Options.PackageExports, needsPackage);
```
DanGrayson, line 49: Here we could simply do `needsPackage currentPackage`, and also remove `currentPackage.Dictionary,` from the statement above.

mahrud: The problem, I think, is that if the package is already loaded (which it always is!), the operation doesn't do anything, in particular it doesn't add the contents of PackageExports to the PackageDictionary.

DanGrayson:
> > Maybe try this:
> 
> Didn't work:

Ah, that's because at top level, the old value is stored in an expression object:
```
export stderrS := setupconst("stderr",Expr(stdError));
```

M2/Macaulay2/m2/examples.m2
DanGrayson: This experiment seems to show that it does, since the way it tells whether a package is loaded is by looking in `loadedPackages`:
```m2
Macaulay2, version 1.16.0.2
               TangentCone, Truncations

i1 : save = (dictionaryPath,loadedPackages);

i2 : needsPackage "BGG"

o2 = BGG

o2 : Package

i3 : (dictionaryPath,loadedPackages) = save;

i4 : needsPackage "BGG"

o4 = BGG

o4 : Package

i5 : dictionaryPath

o5 = {BGG.Dictionary, BoijSoederberg.Dictionary, Truncations.Dictionary,
     ---------------------------------------------------------------------------------
     SimpleDoc.Dictionary, InverseSystems.Dictionary, ConwayPolynomials.Dictionary,
     ---------------------------------------------------------------------------------
     ReesAlgebra.Dictionary, TangentCone.Dictionary, Classic.Dictionary,
     ---------------------------------------------------------------------------------
     PrimaryDecomposition.Dictionary, IntegralClosure.Dictionary,
     ---------------------------------------------------------------------------------
     MinimalPrimes.Dictionary, LLLBases.Dictionary, Elimination.Dictionary,
     ---------------------------------------------------------------------------------
     User#"private dictionary", User.Dictionary, Core.Dictionary, OutputDictionary,
     ---------------------------------------------------------------------------------
     PackageDictionary}

o5 : List
```

mahrud: Oh, sure.

mahrud:
@DanGrayson the CMake builds finish in under 1hr. I excluded a bunch of examples involving printing and stderr from being captured. If you have some time to take a look at why `print` and `File << Thing` don't seem to behave well with capture (and also missing stderr), this PR could go into this release, otherwise probably would be safer to merge after the release so there's more time to look for potential documentation issues.

mahrud:
Oh, actually, since this PR solves a few other issues as well, a far better thing to do is to temporarily add `false and` on this line:
https://github.com/Macaulay2/M2/blob/f2054a597c58add216f785c6cd00db286e262c79/M2/Macaulay2/m2/examples.m2#L129-L130
and merge this now. This was the documentation gets built without capture, but capture is available for use and those other bugs are (like #963) are also resolved.

DanGrayson:
> @DanGrayson the CMake builds finish in under 1hr. I excluded a bunch of examples involving printing and stderr from being captured. If you have some time to take a look at why `print` and `File << Thing` don't seem to behave well with capture (and also missing stderr), this PR could go into this release, otherwise probably would be safer to merge after the release so there's more time to look for potential documentation issues.

I'll take a look...

mahrud:
@DanGrayson I manually disabled capturing example results, so it can be merged without any risk of harming the documentation.

As a bonus, I managed to do what @mikestillman had asked for, so running the examples looks like this:
```m2
[536/545] Installing package Matroids
 -- making example results for "hyperplanes"                                 -- 1.48583 seconds elapsed
 -- making example results for "representationOf"                            -- 1.12504 seconds elapsed
 -- making example results for "getIsos"                                     -- 1.01062 seconds elapsed
 -- making example results for "independenceComplex(Matroid)"                -- 1.27673 seconds elapsed
 -- making example results for "quickIsomorphismTest"                        -- 1.61963 seconds elapsed
```

DanGrayson:
> @DanGrayson I manually disabled capturing example results, so it can be merged without any risk of harming the documentation.
> 
> As a bonus, I managed to do what @mikestillman had asked for, so running the examples looks like this:
> 
> ```m2
> [536/545] Installing package Matroids
>  -- making example results for "hyperplanes"                                 -- 1.48583 seconds elapsed
>  -- making example results for "representationOf"                            -- 1.12504 seconds elapsed
>  -- making example results for "getIsos"                                     -- 1.01062 seconds elapsed
>  -- making example results for "independenceComplex(Matroid)"                -- 1.27673 seconds elapsed
>  -- making example results for "quickIsomorphismTest"                        -- 1.61963 seconds elapsed
> ```

So this pull request offers no advantages over the status quo?  In that case, you could submit it when it's ready.

mahrud:
> So this pull request offers no advantages over the status quo? In that case, you could submit it when it's ready.

There are several issues that this PR fixes, even without using capture! Take a look at the diff.

Issue #1298, opened 2020-06-18
check should do nothing if the optional components are not present


DanGrayson:
Hmm, maybe the function "check" should be modified so if the optional components are not present, then no tests are run.

_Originally posted by @DanGrayson in https://github.com/Macaulay2/M2/issues/1295#issuecomment-645949651_

PR #1551, opened 2020-10-22
Skip check for packages when optional components aren't present.

d-torrance:
Closes: #1298


d-torrance:
Here's an example of this commit in action:

```m2
i1 : check "PHCpack"
--warning: optional components required for PHCpack tests are not present; skipping

i2 : 
```

Issue #1314, opened 2020-06-21
incorrect inverse in quotient ring


pzinn:
what the heck?

```
i1 : R=QQ[t]

o1 = R

o1 : PolynomialRing

i2 : S=R/ideal(t^3)

o2 = S

o2 : QuotientRing

i3 : t+t^2

      2
o3 = t  + t

o3 : S

i4 : oo^(-1)

o4 = - t + 1

o4 : S

i5 : o3*o4

o5 = t

o5 : S

i6 : t^(-1)

o6 = 1

o6 : S 
```

DanGrayson:
Probably in the following code we should check that `ret` is 1:
```c++
ring_elem PolyRingQuotient::invert(const ring_elem f) const
{
  if (nvars_ == 1 && n_quotients() == 1 && K_->is_field() && ! K_->is_fraction_field())
    {
      ring_elem g = quotient_element(0);

      RingElement *f1 = RingElement::make_raw(getAmbientRing(), f);
      RingElement *g1 = RingElement::make_raw(getAmbientRing(), g);
      const RingElement *u1;
      const RingElement *v1;
      const RingElement *ret = rawExtendedGCDRingElement(f1, g1, &u1, &v1);
      if (ret == NULL)
        {
          // one reason this might return NULL is if the coefficient ring is not
          // ZZ/n, ZZ, or QQ
          // now what do we do?
          // we can't return NULL
          INTERNAL_ERROR("ring element gcd computation failed");
        }
      return u1->get_value();
    }
  else if (M_->getNonTermOrderVariables()->len == 0)
    return ann(from_long(1), f);
  else
    {
      // An error message is generated higher up
      return from_long(0);
    }
}
```

PR #1556, opened 2020-10-24
fix #1214 and #1314

### LINKED ISSUE #1214: override not accepting "true" as defaults
the documentation of `override` says
```
override(defaults,args) overrides default values for optional arguments present in the argument sequence args.
[...]
A second possibility is for the argument defaults to be "true"
```
however:
```
i2 : override(true,(a=>1,b=>2))
stdio:2:1:(3): error: expected argument 1 to be a hashtable
```
what am I missing?

pzinn:
The recent `override` discussion reminded me that the doc needed fixing. Done.


pzinn:
I added a fix for #1314

DanGrayson:
With this fix, does running the code in #1314 now result in an error message?

pzinn:
It does:
```
i1 : R=QQ[t]

o1 = R

o1 : PolynomialRing

i2 : S=R/ideal(t^3)

o2 = S

o2 : QuotientRing

i3 : t+t^2

      2
o3 = t  + t

o3 : S

i4 : oo^(-1)
stdio:4:3:(3): error: either element not invertible, or no method available to compute its inverse
```
which is the same message as in the multi-generator case:
```
i5 : R=QQ[t,s]

o5 = R

o5 : PolynomialRing

i6 : S=R/ideal(t^3,1+s^2+t^2)

o6 = S

o6 : QuotientRing

i7 : t+t^2

        2
o7 = - s  + t - 1

o7 : S

i8 : oo^(-1)
stdio:8:3:(3): error: either element not invertible, or no method available to compute its inverse
```

PR #1559, opened 2020-10-27
Update SpecialFanoFourfolds

giovannistagliano:


mahrud:
@giovannistagliano what keywords would you associate to this package? In the next version the packages list is organized by keywords.

giovannistagliano:
> @giovannistagliano what keywords would you associate to this package? In the next version the packages list is organized by keywords.

I do not know. But I've seen that many packages have "Commutative algebra" as their keywords, so "Commutative algebra" might be fine. Better perhaps it would be "Algebraic geometry".

mahrud:
I mostly added "Commutative Algebra" as a placeholder for packages that I didn't know much about. "Algebraic Geometry" sounds good, but if you can think of other keywords let me know, it would help users who know about another package with that keyword find out about your package as well.

Issue #1540, opened 2020-10-09
Narrowing conversion error on i386


d-torrance:
When running `make check` after building Macaulay2 on an i386 chroot, I get the following:

```shell
CXX basics-test.cpp
../../../../../Macaulay2/e/unit-tests/basics-test.cpp: In member function 'virtual void Util_m2array2stdvec_big_Test::TestBody()':
../../../../../Macaulay2/e/unit-tests/basics-test.cpp:59:50: error: narrowing conversion of '-1453853049583' from 'long long int' to 'long int' [-Wnarrowing]
   59 |   std::vector<long> a{-1453853049583, 3, 6, 4, -2};
      |                                                  ^
```

On i386 systems, longs are only 4 bytes, so -1453853049583 won't fit.

PR #1560, opened 2020-10-27
Use long long instead of long for m2array2stdvec_big unit test

d-torrance:
On systems where longs are 32 bits (e.g., i386), the test was failing with
a "narrowing conversion" error as the previous value was less than -2^31.

Closes: #1540

### DIFF
--- a/M2/Macaulay2/e/unit-tests/basics-test.cpp
+++ b/M2/Macaulay2/e/unit-tests/basics-test.cpp
@@ -56,9 +56,9 @@ TEST(Util, m2arrayint_zero)
 
 TEST(Util, m2array2stdvec_big)
 {
-  std::vector<long> a{-1453853049583, 3, 6, 4, -2};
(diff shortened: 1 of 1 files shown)


d-torrance:
Did the previous value have any special meaning, or was it just an arbitrary `long`?

I just lopped off the last 3 digits so it would still line up with the `m2array2stdvec_check` test, which also uses a value beginning with the same digits.

d-torrance:
I opened this too quickly.  With this change, the test compiles, but it fails:
```
basics-test.cpp:62: Failure
Value of: a == c
  Actual: true
Expected: false
[  FAILED  ] Util.m2array2stdvec_big (0 ms)
```

d-torrance:
I figured it out.  We *want* a big value that changes when we cast to an `int`.  But on i386, `long`s and `int`s are *both* 32 bits, so we need to bump up to `long long`.

PR #1565, opened 2020-10-28
Minimize filename for package test inputs

d-torrance:
This improves the readability of the output from the "check" function.
An (extreme) example of the current behavior:

```m2
i1 : get "foo.m2"

o1 = newPackage "Foo"

     input "~/tmp/bar.m2"
     input "~/tmp/./././././bar.m2"

i2 : get "bar.m2"

o2 = TEST "assert(1 + 1 == 2)"

i3 : loadPackage("Foo", FileName => "~/tmp/foo.m2")

ii4 : TEST "assert(1 + 1 == 2)"

ii5 :

ii6 : TEST "assert(1 + 1 == 2)"

ii7 :

o7 = Foo

o7 : Package

i8 : check "Foo"
-- running test 0 of package Foo on line 2 in file /home/dtorrance/tmp/bar.m2
--    rerun with: check_0 "Foo"
 -- making test results
-- running test 1 of package Foo on line 2 in file /home/dtorrance/tmp/./././././bar.m2
--    rerun with: check_1 "Foo"
 -- making test results
```

And after this commit:

```m2
i8 : check "Foo"
-- running test 0 of package Foo on line 2 in file bar.m2
--    rerun with: check_0 "Foo"
 -- making test results
-- running test 1 of package Foo on line 2 in file bar.m2
--    rerun with: check_1 "Foo"
 -- making test results
```

### DIFF
--- a/M2/Macaulay2/m2/testing.m2
+++ b/M2/Macaulay2/m2/testing.m2
@@ -33,8 +33,9 @@ prep := pkg -> (
 
 onecheck = (n, pkg, usermode) -> (
      (filename, lineno, teststring) := pkg#"test inputs"#n;
-     stderr << "-- running test " << n << " of package " << pkg << " on line " << lineno << " in file " << filename << endl;
-     stderr << "--    rerun with: check_" << n << " \"" << pkg << "\"" << endl;
+     stderr << "-* running test " << n << " of package " << pkg << " in file:" << endl;
+     stderr << "   " << filename << ":" << lineno << ":1:" << endl;
+     stderr << "   rerun with: check_" << n << " \"" << pkg << "\" *-" << endl;
      runString(teststring, pkg, usermode);
      )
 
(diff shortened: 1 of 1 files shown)


mahrud:
The repeated `/./` is annoying, but this makes it harder to find the file, especially if it's a generic name like `test.m2`. The emacs-clickable format would be even better:
```m2
/path/to/file.m2:[row]:[col]::
```

d-torrance:
> The repeated `/./` is annoying, but this makes it harder to find the file, especially if it's a generic name like `test.m2`. The emacs-clickable format would be even better:
> 
> ```m2
> /path/to/file.m2:[row]:[col]::
> ```

I like that idea!  I've added a new commit so now the output looks like this:


```m2
i6 : check "Foo"
-* running test 0 of package Foo in file:
   bar.m2:2:1:
   rerun with: check_0 "Foo" *-
 -- making test results
-* running test 1 of package Foo in file:
   bar.m2:2:1:
   rerun with: check_1 "Foo" *-
 -- making test results
```

PR #1568, opened 2020-10-30
Increased degree limit for tangent cone computations

nilten:
The previous degree limit was missing some lowest order terms of obstruction equations. The change should fix this.

### DIFF
--- a/M2/Macaulay2/packages/VersalDeformations.m2
+++ b/M2/Macaulay2/packages/VersalDeformations.m2
@@ -6,13 +6,13 @@
 ---------------------------------------------------------------------------
 newPackage("VersalDeformations",
     Headline => "versal deformations and local Hilbert schemes",
-    Version => "2.1",
-    Date => "April 10, 2020",
+    Version => "2.2",
+    Date => "October 30, 2020",
     Authors => {
         {Name => "Nathan Ilten",
 	  HomePage => "http://www.sfu.ca/~nilten/",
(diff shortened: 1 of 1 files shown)


M2/Macaulay2/packages/VersalDeformations.m2
```diff
@@ -6,13 +6,13 @@
...
-    Version => "2.1",
-    Date => "April 10, 2020",
+    Version => "2.2",
+    Date => "October 30, 2020",
     Authors => {
         {Name => "Nathan Ilten",
 	  HomePage => "http://www.sfu.ca/~nilten/",
 	  Email => "nilten@sfu.ca"}},
-    Keywords => {"Deformation Theory"},
+--  Keywords => {"Deformation Theory"},
```
DanGrayson, line 15: Perhaps you commented this out so it would run with an older version of M2?

nilten: Keywords is giving an error with version 1.16, which is the latest distributed version.

DanGrayson: Well, as this is the code to be incorporated into 1.17, let's leave the line in.

Or you could write an "if" statement based on testing the version number that would run newPackage both ways.  Then your source code would work with both versions.

nilten:
I'm closing this pull request, and will make a new one shortly: I've made a significant number of other changes.

PR #1572, opened 2020-10-31
Remembering formation of modules

### LINKED ISSUE #1222: Tensor does not support arbitrary number of arguments
Unlike directSum, tensor does not support a sequence as an input. It would be nice to add this functionality.

### LINKED ISSUE #1569: components for other bilinear functors
We currently keep track of the components of direct sums of modules.  It would be good to do the same thing for `M**N` and `Hom(M,N)`.    @mikestillman  and @eisenbud want this facility.  Maybe also generate the induced sum components if M and/or N have sum components.  That might also make it possible to generate the associativity isomorphisms in a robust new way.

It should also be possible to attach labels to modules, and have them get used.  For example, if M and N are labelled x and y, then x and y should be usable for getting the components of M++N.  (Hmm?  What does getting M from x have to do with M++N?)

Also, it would be good to have tensor product of multiple modules, rather than of just 2.  That would be analogous to `directSum`.

DanGrayson:
resolves #1569 
resolves #1222

### DIFF
--- a/M2/Macaulay2/m2/chaincomplexes.m2
+++ b/M2/Macaulay2/m2/chaincomplexes.m2
@@ -305,6 +305,7 @@ ChainComplexMap ++ ChainComplexMap := ChainComplexMap => (f,g) -> (
      complete g;
      scan(union(spots f, spots g), i -> h#i = f_i ++ g_i);
      h.cache.components = {f,g};
+     h.cache.formation = BinaryOperation { symbol ++, f, g };
      h)
 
 isHomogeneous ChainComplexMap := f -> (complete f; all(spots f, i -> isHomogeneous f_i))
@@ -552,11 +553,12 @@ chainComplex List := {} >> opts -> maps -> (
 directSum ChainComplex := C -> directSum(1 : C)
 ChainComplex.directSum = args -> (
      C := new ChainComplex;
-     C.cache.components = toList args;
      C.ring = ring args#0;
      scan(args,D -> (complete D; complete D.dd;));
      scan(unique flatten (args/spots), n -> C#n = directSum apply(args, D -> D_n));
      scan(spots C, n -> if C#?(n-1) then C.dd#n = directSum apply(args, D -> D.dd_n));
+     C.cache.components = toList args;
+     C.cache.formation = FunctionApplication { directSum, args };
      C)
 ChainComplex ++ ChainComplex := ChainComplex => (C,D) -> directSum(C,D)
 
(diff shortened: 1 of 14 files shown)


mahrud:
Should it also be `{directSum, (f, g)}`? In a few places the components are not in a sequence.

DanGrayson:
> Should it also be `{directSum, (f, g)}`? In a few places the components are not in a sequence.

I'm not sure what you mean.  Could you point me to those places?

M2/Macaulay2/m2/chaincomplexes.m2
```diff
@@ -305,6 +308,7 @@ ChainComplexMap ++ ChainComplexMap := ChainComplexMap => (f,g) -> (
      complete g;
      scan(union(spots f, spots g), i -> h#i = f_i ++ g_i);
      h.cache.components = {f,g};
+     h.cache.formation = BinaryOperation { symbol ++, f, g };
```
mahrud, line 311: Should this be `{directSum, (f, g)}`?

M2/Macaulay2/m2/gradedmodules.m2
```diff
@@ -239,6 +244,7 @@ GradedModuleMap ++ GradedModuleMap := GradedModuleMap => (f,g) -> (
      h.degree = f.degree;
      scan(union(spots f, spots g), i -> h#i = f_i ++ g_i);
      h.cache.components = {f,g};
+     h.cache.formation = BinaryOperation { symbol ++, f, g };
```
mahrud, line 247: Same here.

M2/Macaulay2/m2/gradedmodules.m2
```diff
@@ -316,6 +323,7 @@ GradedModule ++ GradedModule := GradedModule => (C,D) -> (
      if R =!= D.ring then error "expected graded modules over the same ring";
      scan(union(spots C, spots D), i -> E#i = C_i ++ D_i);
      E.cache.components = {C,D};
+     E.cache.formation = BinaryOperation { symbol ++, C, D };
```
mahrud, line 326: And here.

M2/Macaulay2/m2/chaincomplexes.m2
DanGrayson: No, because the result was formed by evaluating `f ++ g`, not be evaluating `directSum(f,g)`.

mahrud: Are those different?

DanGrayson: Hmm -- okay, in the cases where we have code for both, I'll just replace the superfluous code for `f++g` by a call to `directSum`.

PR #1574, opened 2020-11-01
Add "Verbose" option to "check"

d-torrance:
This pull request is similar to #1407, but for running `check` on a package rather than building its examples.  It's especially useful when building on a system where the user doesn't have access to any build artifacts to aid in diagnosing failing tests.

In particular, we add a `Verbose` option to `check` that displays the output of a failing test when `true`.  I've already applied this patch to my PPA builds, and here's an example of the output (from https://launchpadlibrarian.net/504684149/buildlog_ubuntu-bionic-i386.macaulay2_1.16.0.2+git202010291401-0ppa202010311520~ubuntu18.04.1_BUILDING.txt.gz):

```m2
-* running test 12 of package DeterminantalRepresentations in file:
   DeterminantalRepresentations.m2:1527:1:
   rerun with: check_12 "DeterminantalRepresentations" *-
 -- making test results
-* running test 13 of package DeterminantalRepresentations in file:
   DeterminantalRepresentations.m2:1539:1:
   rerun with: check_13 "DeterminantalRepresentations" *-
 -- making test results
../m2/debugging.m2:20:6:(1):[9]: error: test #2 of package DeterminantalRepresentations failed:

     1.81712x1+1.81712x2+.757134x3 |  | .312572x3                    
     .438081x3                     .312572x3                     |}
     1.81712x1+3.63424x2+1.21141x3 .0920576x3                    |
     .0920576x3                    1.81712x1+1.81712x2+.757134x3 |

o7 : List

i8 : assert(all(reps, L -> clean(1e-10, fRR - det L) == 0))
stdio:10:1:(3): error: assertion failed

../m2/testing.m2:59:22:(1):[8]: --back trace--
../m2/methods.m2:119:80:(1):[7]: --back trace--
../m2/option.m2:16:8:(1):[6]: --back trace--
../m2/testing.m2:44:33:(1):[5]: --back trace--
../m2/methods.m2:119:80:(1):[4]: --back trace--
../m2/option.m2:16:8:(1):[3]: --back trace--
currentString:1:91:(3):[2]: --back trace--
Macaulay2/Core/startup.m2.in:563:33:(0):[1]: --back trace--
Macaulay2/Core/startup.m2.in:674:6:(0): --back trace--
Makefile:39: recipe for target 'check-DeterminantalRepresentations' failed
make[4]: *** [check-DeterminantalRepresentations] Error 1
make[4]: Leaving directory '/<<BUILDDIR>>/macaulay2-1.16.0.2+git202010291401/M2/Macaulay2/packages'
Makefile:14: recipe for target 'check-in-packages' failed
make[3]: *** [check-in-packages] Error 2
```

Note that we've also refactored `check` so that all of the code is in one place, instead of having to make sure that `check Package` and `check(ZZ, Package)` stay in sync with one another.

*Note*:  I haven't actually tested the CMake commit.

### DIFF
--- a/M2/Macaulay2/m2/testing.m2
+++ b/M2/Macaulay2/m2/testing.m2
@@ -39,7 +39,7 @@ onecheck = (n, pkg, usermode) -> (
      runString(teststring, pkg, usermode);
      )
 
-check = method(Options => {UserMode => null})
+check = method(Options => {UserMode => null, Verbose => false})
 check String  := opts -> pkg -> check(-1, pkg, opts)
 check Package := opts -> pkg -> check(-1, pkg, opts)
 
@@ -58,4 +58,12 @@ check(ZZ, Package) := opts -> (n, pkg) -> (
                 errorList = append(errorList, k)));
     if hadError then error("test", if numErrors > 1 then "s" else "",
         " #", demark(", ", toString \ errorList),
-        " of package ", toString pkg, " failed");)
+        " of package ", toString pkg, " failed",
+        if opts.Verbose then (
+            numTests := if n == -1 then pkg#"test number" else 1;
+            ":" | newline |
+            concatenate(apply(errorList, k ->
+                newline | get("!tail " | temporaryDirectory() |
+                toString(temporaryFilenameCounter + 2 * (k - numTests)) |
+                ".tmp")))
+        ) else "");)
--- a/M2/Macaulay2/m2/testing.m2
+++ b/M2/Macaulay2/m2/testing.m2
@@ -40,15 +40,8 @@ onecheck = (n, pkg, usermode) -> (
      )
 
 check = method(Options => {UserMode => null})
-check String  := opts -> pkg -> check(needsPackage (pkg, LoadDocumentation => true), opts)
-check Package := opts -> pkg -> (
-    if not pkg.Options.OptionalComponentsPresent then (
-        stderr << "--warning: optional components required for " <<
-            toString pkg << " tests are not present; skipping" << endl;
-        return);
-    pkg = prep pkg;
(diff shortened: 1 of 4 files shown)


mahrud:
Why not use `debugLevel>0` instead of Verbose? The benefit of debugLevel is that it's an integer, so it's more fine grained. I'd personally prefer to deprecate "Verbose" and "Verbosity", so prefer not to add new uses.

DanGrayson:
> Why not use `debugLevel>0` instead of Verbose? The benefit of debugLevel is that it's an integer, so it's more fine grained. I'd personally prefer to deprecate "Verbose" and "Verbosity", so prefer not to add new uses.

The "Verbose" option is better, because it targets a specific function.  Also, "debugLevel" is a global variable, it's not "functional".

mahrud:
> The "Verbose" option is better, because it targets a specific function. Also, "debugLevel" is a global variable, it's not "functional".

As you said, Macaulay2 is not a functional programming language, and in this case the more verbose output of check is specifically for the purpose of debugging, so debugLevel seems very apt here.

DanGrayson:
> > The "Verbose" option is better, because it targets a specific function. Also, "debugLevel" is a global variable, it's not "functional".
> 
> As you said, Macaulay2 is not a functional programming language, and in this case the more verbose output of check is specifically for the purpose of debugging, so debugLevel seems very apt here.

It's true that Macaulay2 is not a functional programming language, but functional programming is still a good thing, and can be pursued in a limited way by avoiding assignment statements.

Also, who wants to see all the junk that all the other functions start spitting out when debugLevel is on?  Better to target it with an option.

mahrud:
Okay, then let Verbose print the output of all tests. Why just failing tests?

PR #1588, opened 2020-11-11
Update list of Ubuntu/Debian packages providing build dependencies

d-torrance:
It's now possible to build Macaulay2 using only Ubuntu/Debian packages
in Ubuntu 21.04 (Hirsute Hippo) and Debian 11 (Bullseye), so we update
the list to reflect this.

### DIFF
--- a/M2/INSTALL
+++ b/M2/INSTALL
@@ -83,11 +83,15 @@ various modern operating systems:
 
   Ubuntu and Debian:
     Install packages as root with:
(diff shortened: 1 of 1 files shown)


DanGrayson:
Does the same list of packages work in older versions of ubuntu and debian?

d-torrance:
> Does the same list of packages work in older versions of ubuntu and debian?

Yes, I was careful to only add packages to the first `apt-get install` line if they've been around for some time (e.g., 4ti2 was introduced in 2015 and has been a part of the past two Debian stable releases).  The newer packages are only mentioned in the comments below that line.

PR #1593, opened 2020-11-15
Update Cremona.m2

giovannistagliano:


giovannistagliano:
Hi. I'd like to know if there is a date planned for the new version of M2.

DanGrayson:
> Hi. I'd like to know if there is a date planned for the new version of M2.

Yes, see the announcement on the google group at https://groups.google.com/g/macaulay2/c/XFFbr8OdAe4

giovannistagliano:
OK, thanks.

M2/Macaulay2/packages/Cremona.m2
```diff
@@ -54,7 +54,8 @@ export{
    "abstractRationalMap",
    "specialCubicTransformation",
    "isMorphism",
-   "exceptionalLocus"
+   "exceptionalLocus",
+   "inverse'"
```
DanGrayson, line 58: The introduction of  `inverse'` will cause confusion with `inverse`.  Just add a new method to `inverse` instead.

DanGrayson:
Could you add a comment at the top of the package about copyright and license?  See the last bit of https://github.com/Macaulay2/M2/blob/master/M2/Macaulay2/packages/README

M2/Macaulay2/packages/Cremona.m2
giovannistagliano: I already added the method inverse(RationalMap). Maybe I could replace inverse'(RationalMap) with something like inverse(RationalMap,Nothing)... or rename inverse' into inverse2 ?

DanGrayson: Let me help you find a better way.  What's the difference between `inverse(RationalMap)` and `inverse'(RationalMap)`?

giovannistagliano: Both compute the inverse in the same way, but neither verifies whether the map is really birational (as this requires non-trivial computations). What change is just the test done at the end to verify that the output is correct and a fortiori that the input map was birational. inverse' performs a probabilistic but very fast test.

DanGrayson: And your method for `inverse` performs no test at all?

giovannistagliano: `inverse` performs a non-probabilistic test but sometimes very hard.

DanGrayson: In any case, this sort of thing is exactly why we have optional named arguments to some method functions.  Writing something like `inverse(f,Verify=>false)`, `inverse(f,Verify=>Probabilistic)`, or `inverse(f,Verify=>true)` would make everything clear to the reader.  Unfortunately, `inverse` doesn't accept optional arguments, but you can simulate it by installing a method for `inverse(RationalMap,Option)`, anticipating the ultimate day when we take the time to convert `inverse` to a function that does accept options.

Wouldn't that be better?

giovannistagliano: `inverseMap(...,MathMode=>false)` performs no test at all, and  `inverse` is basically the same of `inverseMap(...,MathMode=>true)`, which performs a full test. So I see `inverse'` as a middle ground approach between MathMode=>true and MathMode=>false.

DanGrayson: That's even more confusing.  Wouldn't it be a good idea to put the functionality of `inverseMap` into a method for `inverse`?

giovannistagliano: > In any case, this sort of thing is exactly why we have optional named arguments to some method functions. Writing something like `inverse(f,Verify=>false)`, `inverse(f,Verify=>Probabilistic)`, or `inverse(f,Verify=>true)` would make everything clear to the reader. Unfortunately, `inverse` doesn't accept optional arguments, but you can simulate it by installing a method for `inverse(RationalMap,Option)`, anticipating the ultimate day when we take the time to convert `inverse` to a function that does accept options.
> 
> Wouldn't that be better?

I'll try to do as you say, thanks.

giovannistagliano: > That's even more confusing. Wouldn't it be a good idea to put the functionality of `inverseMap` into a method for `inverse`?

I agree with you. The same type of problem occurs with other functions, such as degreeOfRationalMap which could be degree(RationalMap, MathMode => false/true). But in the beginning I had implemented all the methods only for the RingMaps class (interpreted as rational maps) and methods such as degree(RingMap) and inverse(RingMap) already existed. Now, considering that the package is several years old, I would prefer to continue accepting ring maps as input.

DanGrayson: Okay.   But now what about the name `MathMode`?  That's not descriptive, because all computations in M2 are done with "math".  What about Verify instead?

giovannistagliano: > Okay. But now what about the name `MathMode`? That's not descriptive, because all computations in M2 are done with "math". What about Verify instead?

Yes, but for this version it would take me too much effort to change that name. However, the name was chosen to be the same on all methods of the package. It can mean checking the output (as in inverseMap) but also performing the calculation using a non-probabilistic algorithm (as in projectiveDegrees).

giovannistagliano: > In any case, this sort of thing is exactly why we have optional named arguments to some method functions. Writing something like `inverse(f,Verify=>false)`, `inverse(f,Verify=>Probabilistic)`, or `inverse(f,Verify=>true)` would make everything clear to the reader. Unfortunately, `inverse` doesn't accept optional arguments, but you can simulate it by installing a method for `inverse(RationalMap,Option)`, anticipating the ultimate day when we take the time to convert `inverse` to a function that does accept options.
> 
> Wouldn't that be better?

I did this and it looks nicer indeed.

giovannistagliano:
I think I'm done.

M2/Macaulay2/packages/Cremona.m2
```diff
@@ -583,18 +650,30 @@ RationalMap ^ ZZ := (Phi,j) -> (
...
+   if Phi#"inverseRationalMap" === null and Psi#"inverseRationalMap" === null then forceInverseMap(Phi,Psi); 
+   Psi
+);
+
+inverse (RationalMap) := (Phi) -> inverse(Phi,MathMode=>false);
 
 RationalMap ! := (Phi) -> (
      toMap Phi;
-     try Phi^(-1);
+     try inverse Phi;
```
DanGrayson, line 676: Using "try" is unreliable, because if it may catch an error of a type you didn't expect, which you would want to find out about.  Why even run this code, if the result is not needed?

M2/Macaulay2/packages/Cremona.m2
```diff
@@ -884,13 +963,13 @@ inverseMapInt (RationalMap,Nothing) := o -> (Phi,nothing) -> (
...
    T := transpose mingens kernel transpose Sub submatrix(jacobian Bl,{0..n},);
-   psi := try map rationalMap(target Phi,source Phi,submatrix(T,{0},)) else error "not able to obtain an inverse rational map";
+   psi := try map rationalMap(target Phi,source Phi,submatrix(T,{0},)) else error "not able to obtain an inverse rational map; make sure the input map is birational";
    psiAll := apply(numRows T,i -> map(target Phi,source Phi,submatrix(T,{i},)));
    if not o.MathMode then return (psi,psiAll);
    if (if n - Phi#"dimTarget" <= Phi#"dimAmbientSource" - Phi#"dimSource" then isInverseMap(map Phi,psi) else isInverseMap(psi,map Phi)) then (
         if o.Verbose then <<certificate; return (psi,psiAll);
    ) else (
-        error "not able to obtain an inverse rational map";
+        error "not able to obtain an inverse rational map; make sure the input map is birational";
```
DanGrayson, line 972: Could the code fail to find the inverse of a birational map?  If so, you might say "not able to obtain an inverse rational map; the map may not be birational".  If not, say "map is not birational".

The problem with the current way of saying it is: what should the user do if he/she has already ensured the map is birational?  It seems to imply that the code would always succeed in finding the inverse, but it's a bit ambiguous.

DanGrayson (changes requested):

M2/Macaulay2/packages/Cremona.m2
giovannistagliano: OK, thanks. I correct it immediately.

M2/Macaulay2/packages/Cremona.m2
giovannistagliano: Ok, but this is really a very secondary function (mostly used for testing). I would leave it unchanged if possible.

M2/Macaulay2/packages/Cremona.m2
giovannistagliano: > Okay. But now what about the name `MathMode`? That's not descriptive, because all computations in M2 are done with "math". What about Verify instead?

A better name than `MathMode` might be for example be `Certify`. But I would like not to deviate too much from the certified version of the package. However I really appreciate the comments.

M2/Macaulay2/packages/Cremona.m2
giovannistagliano: This function is only useful for some simple maps where you can calculate everything easily. It returns the same rational map, i.e. the same mutable hash table, but first fills almost all the entries (such as "projectiveDegrees", "isBirational", etc). After this run, the method "describe(RationalMap)" yields a larger output.
Example:
`i1 : needsPackage "Cremona";

i2 : f = rationalMap map specialCremonaTransformation 5;

o2 : RationalMap (quadratic rational map from PP^5 to PP^5)

i3 : describe f

o3 = rational map defined by forms of degree 2
     source variety: PP^5
     target variety: PP^5
     coefficient ring: QQ

i4 : describe f!

o4 = rational map defined by forms of degree 2
     source variety: PP^5
     target variety: PP^5
     dominance: true
     birationality: true (the inverse map is already calculated)
     projective degrees: {1, 2, 4, 4, 2, 1}
     number of minimal representatives: 1
     dimension base locus: 2
     degree base locus: 4
     coefficient ring: QQ
`
The user should perform the functions he needs, not all of them in the order I choose. So I consider this function quite irrelevant.

DanGrayson: What if you just remove "try"?  Won't it still work?

giovannistagliano: In that case, it would give an error when the map `Phi` is not birational. One could replace `try inverse Phi` with `if isBirational Phi then inverse Phi`, but this test requires non trivial computation (and useless if in the case of birationality one also want the inverse map).

DanGrayson: If `inverse Phi` fails more quickly than `isBirational Phi` does when `Phi` is not birational, why not include the code that fails at the start of `isBirational`?  Then `isBirational Phi` will return `false` more quickly.

giovannistagliano: This could be done, but I have some concerns about it. I already use this approach. If `inverse Phi` succeeds then `isBirational Phi` will return `true` quickly; if `degree Phi` returns a value different from 1 then `isBirational Phi` will return `false` quickly and `inverse Phi` will give an error quickly, etc. But I don't trust that a particular code fails only in the case the answer doesn't exist. In the future, the code of `inverse(RationalMap)` could be rewritten keeping track of where it fails (if it does).

DanGrayson: Okay.

Issue #1529, opened 2020-10-03
Simplify output methods


mahrud:
Currently we have the following output methods:
```m2
net, info, html, htmlWithTex, markdown, mathML, tex, texMath
```
There are also some internal ones that are not exported, like `texWrap`.

Now that math in the web documentation is rendered with KaTeX, I think we can simplify some of these. For instance, `html` applied to a non-hypertext object should simply call `tex` on the object and wrap it in appropriate delimiters (`$`/`$$`, or maybe `\(`/`\[`). This is basically what `htmlWithTex` does, so I think we can basically merge them by fold in the before/after printing features of `htmlWithTex` into `html`. In other words, all HTML is HTML with tex now, so there's no distinction.

Is there a reason not to do this?

cc: @pzinn

pzinn:
besides my answer in another thread:
> we definitely should discuss this in more detail at some point, but my short answer: I think html should indeed produce html 
> code, not tex code. if one wants mixed output, a method like htmlWithTex would be appropriate, yes. or the user should be 
> allowed to directly insert tex code into his HTML using TEX or something like that, but it's a different story. I didn't follow your
> recent changes, so am not sure if they contradict the philosophy above.

I want to say one word about `texWrap`: right now the main use of `htmlWithTex` is for my web app. unfortunately the web app cannot use standard delimiters $ for tex math, it uses instead its own tags (ascii<32). That's why in my original version of the code I had made the delimiters used by `tex` mutable objects, so `htmlWithTex` just changed these before calling `tex`. Dan protested (rightly) that this made the code not reentrant or thread-safe or what not, so instead of really fixing the problem I temporarily introduced `texWrap` which is... the same as `tex` except with modifyable delimiters. This code will go, eventually, but before that I want to clarify some of the other related issues, such as the `html` issue that started this conversation.

pzinn:
before I forget, there's also one side issue which is that `htmlWithTex String` and `html String` differ, as well as `htmlWithTex Net` and `html Net`.

mahrud:
> before I forget, there's also one side issue which is that `htmlWithTex String` and `html String` differ, as well as `htmlWithTex Net` and `html Net`.

Yeah, I was hoping to talk about those. I think htmlWithTex does the right thing for Net, but I'm not sure about String. In particular, why does it put it in <pre>?

Descent is also something that I think htmlWithTex does right, and I'm wondering why the code isn't just implemented for html.

[...]

Issue #1594, opened 2020-11-16
height of empty Net


pzinn:
something I don't understand:
```
i1 : a=net ""

o1 = 

i2 : b=stack{}

o2 = 

i3 : height a

o3 = 1

i4 : height b

o4 = 0
```
why are the results different? and which one is the "correct" one?

DanGrayson:
I think they're both correct.  All strings convert to nets of height 1.

pzinn:
OK!

Issue #1218, opened 2020-06-01
toString vs net Type inconsistency


pzinn:
another minor issue (and easily fixable), but it's the season for bug fixing...
compare the two codes below. there are slight differences (besides the normal `toString`/`net` differences): `toString` shows `parent X`, whereas `net` doesn't; `net` has this extra `if #X > 0` test. I think they should be uniformized. (in which case, should `parent X` be shown? I'd say yes).

```
i1 : code(toString,Type)

o1 = -- code for method: toString(Type)
     /usr/share/Macaulay2/Core/nets.m2:28:20-33:81: --source code:
     toString Type := X -> (
          if hasAnAttribute X then (
               if hasAttribute(X,PrintNames) then return getAttribute(X,PrintNames);
               if hasAttribute(X,ReverseDictionary) then return toString getAttribute(X,ReverseDictionary);
               );
          concatenate(toString class X, " of ", toString parent X, "{...", toString(#X), "...}"))

i2 : code(net,Type)

o2 = -- code for method: net(Type)
     /usr/share/Macaulay2/Core/nets.m2:213:15-219:70: --source code:
     net Type := X -> (
          if hasAnAttribute X then (
               if hasAttribute(X,PrintNet) then return getAttribute(X,PrintNet);
               if hasAttribute(X,PrintNames) then return net getAttribute(X,PrintNames);
               if hasAttribute(X,ReverseDictionary) then return toString getAttribute(X,ReverseDictionary);
               );
          horizontalJoin ( net class X, if #X > 0 then ("{...", toString(#X), "...}") else "{}" ))
```

pzinn:
also, there's no `texMath Type` (it uses `texMath HashTable` which again has a similar but subtly different output than the above)

pzinn:
might want to add a `texMath Type` too if showing `parent X`

DanGrayson:
Sounds good to me!

pzinn:
just to remember what the issue when the time comes to fix it:
currently both the `net` and `toString` output are ugly:
```
i1 : new Type of List

o1 = Type{}

o1 : Type

i2 : toString oo

o2 = Type of List{...0...}
```
Probably something in between, like `Type of List`, would be preferrable.

PR #1595, opened 2020-11-16
fix minor documentation/html issues

pzinn:
- `TO` got double `htmlLiteral`ed, resulting in the documentation problems mentioned in #1529 -- fixed.
- A bug was introduced in the recent rewrite in format.m2:
```
i1 : needsPackage "Text"; net DIV {}
stdio:1:22:(3): error: expected a list of integers
```
fixed:
```
i1 : needsPackage "Text"; net DIV {}

o2 = 
```
- Now we allow `html` to process anything, this shouldn't cause an error:
```
i3 : net DIV { det }
stdio:2:1:(3): error: no method found for applying -*Function*- to:
     argument   :  determinant (of class MethodFunctionWithOptions)
```
fixed:
```
i3 : net DIV { det }

o3 = determinant
```

### DIFF
--- a/M2/Macaulay2/m2/format.m2
+++ b/M2/Macaulay2/m2/format.m2
@@ -118,7 +118,7 @@ scan({net, info},
 	-- object or a sequence of such objects and BKs or SKs for later splicing.
 	parser' := value (toString parser | "'") <- method(Dispatch => Thing);
 	-- setup default rendering methods
-	parser' Hypertext := parser;
+	parser' Thing := parser;
 	-- { } indicates wrapping is already done or is not desired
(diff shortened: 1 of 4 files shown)


pzinn:
I added a minor fix of `html Net` in the case where the net has zero depth and height -- cf #1594

pzinn:
added a simple fix for #1218, producing the desired output:
```
i1 : new Type of List

o1 = Type of List

o1 : Type
```
also, `texMath Type` no longer needed because `html Type` takes over; minor fix for `html Monoid` added.

pzinn:
there was a minor typo in the doc which prevented KaTeX from compiling successfully. fixed.

Issue #1598, opened 2020-11-18
line wrapping of documentation is screwed up


DanGrayson:
Look at this:

<img width="737" alt="Screen Shot 2020-11-18 at 6 10 07 AM" src="https://user-images.githubusercontent.com/700228/99529049-d1e75800-2964-11eb-9ca6-4384b2a8f139.png">

Maybe @mahrud , @d-torrance can help.

d-torrance:
The problem appears to be with nested lists, e.g.:

```m2
i1 : debug Core

i2 : UL LI {"foo", UL LI ("f" | concatenate apply(100, i -> "o"))}

o2 =   * foo  * foooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooo
                ooooooooooooooooooooooooooooooooooo
         ooo

o2 : UL
```
We're assuming that we only need to indent 4 spaces, but that's only okay for the top level list.
https://github.com/Macaulay2/M2/blob/2c9e746102cd4d91c114887eaf10723e0642e5e2/M2/Macaulay2/m2/format.m2#L177-L182

d-torrance:
Never mind my previous comment -- the indentation is just fine.  `printWidth` with will decrease appropriately thanks to recursion.  For example:

```m2
i1 : debug Core

i2 : UL UL UL UL UL UL UL apply(2, i -> "f" | concatenate apply(100, i -> "o"))

o2 =   *   *   *   *   *   *   * fooooooooooooooooooooooooooooooooooooooooooooo
                                 oooooooooooooooooooooooooooooooooooooooooooooo
                                 ooooooooo
                               * fooooooooooooooooooooooooooooooooooooooooooooo
                                 oooooooooooooooooooooooooooooooooooooooooooooo
                                 ooooooooo
```

The problem is mixing non-`UL`'s and `UL`'s together in the same `LI` object.  They're getting horizontally joined, but they should probably be stacked instead.

```m2
i3 : LI {"foo", UL "bar"}

o3 = foo  * bar
```

DanGrayson:
This bug was introduced recently:

<img width="289" alt="Screen Shot 2020-11-18 at 9 32 02 AM" src="https://user-images.githubusercontent.com/700228/99551000-f5200080-2980-11eb-9fb0-ec8c86c31387.png">

DanGrayson:
Maybe commit c7f4e4c9f0beeb208f779cc374266cda26f48f3b did it:

```
[...]

PR #1599, opened 2020-11-18
Stack nested lists vertically

d-torrance:
Previously, each element of an LI object was joined horizontally.
This is usually what we want, unless we encounter a nested OL or UL
object.  In this case, we ended up with:

```m2
i1 : beginDocumentation()

i2 : LI{"a", "b", UL{"c", "d"}, "e", "f", OL{"g", "h"}}

o2 = ab  * cef  1. g
         * d    2. h

o2 : LI
```

We now separate these out and stack them vertically:

```m2
i1 : beginDocumentation()

i2 : LI{"a", "b", UL{"c", "d"}, "e", "f", OL{"g", "h"}}

o2 = ab
       * c
       * d
     ef
       1. g
       2. h

o2 : LI
```

Closes: #1598

### DIFF
--- a/M2/Macaulay2/m2/format.m2
+++ b/M2/Macaulay2/m2/format.m2
@@ -174,6 +174,18 @@ info PRE := x -> wrap(printWidth, "-", net concatenate noopts x)
 net  CODE :=
 info CODE := x -> stack lines concatenate noopts x
 
+LIop := op -> x -> (
+     -* we want to join each element of our li horizontally, unless
+        we encounter a nested list, in which case we want to stack
+        vertically.  so first identify these *-
+     toStack := sublists(sublists(toList noopts x,
+          i -> member(class i, {OL, UL})), j -> not instance(j, List));
+     stack apply(toStack, y -> wrapHorizontalJoin(op \ y))
+     )
+
+info LI := LIop info
+net  LI := LIop net
+
 ULop := op -> x -> (
      s := "  * ";
      printWidth = printWidth - #s;


d-torrance:
This bug has been bothering me for a while!  Here's this commit in action:

```m2
i4 : help combine

o4 = combine -- combine hash tables
     ******************************

     Synopsis
     ========

       * Usage: 
             z = combine(x,y,f,g,h)
       * Inputs:
           * x, a hash table
           * y, a hash table of the same class as x
           * f, a function of two variables to be used for combining a key of x
             with a key of y to make a new key for z.
           * g, a function of two variables to be used for combining a value of
             x with a value of y to make a new value for z.
           * h, a function of two variables to be used for combining two values
             returned by g when the corresponding keys returned by f turn out to
             be equal.  Its first argument will be the value accumulated so far,
             and its second argument will be a value just provided by g.
       * Outputs:
           * z, a new hash table, of the same class as x and y, containing the
             pair f(p,q) => g(b,c) whenever x contains the pair p => b and y
             contains the pair q => c, except that h is used to combine values
             when two keys coincide.  If f or g evaluates "continue", then
             nothing is contributed to the resulting hash table.  If h evaluates
             "continue", then, at that point, the entry stored under the key
             f(p,q) in the hash table under construction is removed.
```

DanGrayson:
This fix looks complicated.  Why not just undo commit c7f4e4c9f0beeb208f779cc374266cda26f48f3b?

d-torrance:
> This fix looks complicated. Why not just undo commit [c7f4e4c](https://github.com/Macaulay2/M2/commit/c7f4e4c9f0beeb208f779cc374266cda26f48f3b)?

That works, too!  I figured there must have been a good reason for wanting an `LI` to be `HypertextParagraph` instead of a `HypertextContainer`.

DanGrayson:
Let's just undo that commit instead.

d-torrance:
> Let's just undo that commit instead.

Sounds good!  Closing.

DanGrayson:
Shall I undo it, or would you like to?

d-torrance:
I already have a branch with that commit reverted, so I'll go ahead and push it and open a PR.

PR #1612, opened 2020-11-24
Update poset-example.html

fchapoton:
trying to fix the broken web page https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2/share/Macaulay2/Visualize/poset-example.html

### DIFF
--- a/M2/Macaulay2/packages/Visualize/digraph-example.html
+++ b/M2/Macaulay2/packages/Visualize/digraph-example.html
@@ -1,14 +1,14 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
   <head>
-    <meta charset="utf-8">
+    <meta charset="utf-8"/>
     <title>Macaulay2: Digraph Visualization</title>
-    <link href="css/bootstrap.min.css" rel="stylesheet">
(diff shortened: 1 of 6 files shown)


DanGrayson:
The new version still has an error.  You can verify that by uploading it to https://validator.w3.org/.

fchapoton:
Here are some more fixes that make the validator happy. Thanks for the feedback.

DanGrayson:
What about the other html files in the package?  Does the validator like them?

fchapoton:
They should be given the same treatment, yes. Should I do so here ?

DanGrayson:
Yes, that would be great, it you have the spare time to donate.  Time is tight, though, since we are about to release 1.17.

Maybe, in the future, we should validate such html files included in packages.  We already validate the ones we generate.

DanGrayson:
Authors ( @bstone, ... ) : are these changes okay with you?

fchapoton:
ok, all done. I did not check wether the validator is fully happy, but it is certainly less angry.

DanGrayson:
Okay, thanks!

bstone:
I don't have any problems with these changes.  Thanks for your work on this.

PR #1615, opened 2020-11-24
Add new packages: EigenSolver, MultiplicitySequence; Update NumericalImplicitization

jchen419:


mahrud:
Did you mean to update PrimaryDecomposition here? I think some old version of it was accidentally included in this PR.

jchen419:
OK, reverted the changes to PrimaryDecomposition now

M2/Macaulay2/m2/methods.m2
```diff
@@ -277,7 +277,7 @@ length = method(TypicalValue => ZZ, Dispatch => Thing)
 codim = method( Options => true )
 radical = method( Options=>{ Unmixed=>false, CompleteIntersection => null, Strategy => Decompose } )
 regularity = method( TypicalValue => ZZ, Options => { Weights => null } )
-associatedPrimes = method( TypicalValue => List, Options =>{ Strategy => 1 } )
+associatedPrimes = method( TypicalValue => List, Options =>{ Strategy => 1, CodimensionLimit => -1 } )
```
DanGrayson, line 280: You probably didn't intend to make this change.

M2/Macaulay2/packages/MultiplicitySequence.m2
```diff
@@ -0,0 +1,866 @@
...
+        multiplicitySequence(i, I)
+    Inputs
+        I:Ideal
+        i:ZZ
+    Outputs
+        :HashTable
+	    the multiplicity sequence of I
+    Description
+        Text 
+            Given a (graded)-ideal I, this function computes
```
DanGrayson, line 372: no hyphen here

M2/Macaulay2/packages/NumericalImplicitization.m2
```diff
@@ -64,9 +65,16 @@ newPackage("NumericalImplicitization",
         "generalCombinations",
         "imageSlice",
         "witnessPointPairs",
-	"isOnImage"
+	"isOnImage",
+        "realPoint",
+        "nelderMead",
```
DanGrayson, line 70: "nelderMead" is a clumsy name.  What about "optimize" instead?  It's descriptive.

DanGrayson (changes requested):

M2/Macaulay2/m2/methods.m2
jchen419: This was leftover from an old commit for PrimaryDecomposition-related code - I'll revert it.

M2/Macaulay2/packages/MultiplicitySequence.m2
jchen419: OK.

M2/Macaulay2/packages/NumericalImplicitization.m2
jchen419: I'm a little hesitant to refer to this method by a totally generic name, since the function is exactly implementing the Nelder-Mead algorithm. Calling it "optimize" might conflict with other optimization methods added in the future. (Also, I was not really intending for users to call this function directly, and only exposed it via the user-facing, documented method `realPoint`.)

DanGrayson: If you add other optimization methods in the future, the choice can be offered to the user like this:
```m2
optimize(..., Strategy=>NelderMead)
```
, with one of them being the default.  It's better if function names are verbs describing what is to be done.  The choice of algorithm can be described in the documentation.

jchen419: Hi Dan, I'm still not entirely sure about this name change. I am perfectly happy with the possibility of a (future) general optimization function to call this as a strategy (or even a hook), but I don't know if this function should play that role. Would you find it acceptable to just not export this function (so it becomes an internal function)? Alternatively, what about something like "optimizeNelderMead"?

DanGrayson: Okay.  That's a bit prettier, because both of the names get capitalized, and there is something descriptive.

jchen419: OK, great! Marking as resolved.

PR #1616, opened 2020-11-25
Update VirtualResolutions package to v1.3

mahrud:
The changes are on top of #1385, hence bloating the change log.


mahrud:
@DanGrayson have you had a chance to take a look at this pull request?

By the way, did you know on github you can limit the pull request diff to only a subset of the commits? This is useful here, for instance, where only the last 8 commits are original to this pull request, and you can see only their diff here:
https://github.com/Macaulay2/M2/pull/1616/files/baf468c3547ab96a71d5ae8a49dfd5ca64506b03..77e80d0526b96eb29ad8a9327f57e209e739dd18?file-filters%5B%5D=
This is where you can do this:
image

DanGrayson:
Thanks for the pointer to the controller of the diffs, but it doesn't seem easy to determine the common ancestor of this pull request with #1385 .   I might as well continue waiting for that pull.

mahrud:
> Thanks for the pointer to the controller of the diffs, but it doesn't seem easy to determine the common ancestor of this pull request with #1385 . I might as well continue waiting for that pull.

I already determined it for you in the link in my previous comment, where you can start leaving your comments. The common ancestor is baf468c3547ab96a71d5ae8a49dfd5ca64506b03

You can also use the `git merge-base` tool:
```
[mahrud@noether clang]$ git merge-base VirtualResolutions PrimaryDecomposition 
baf468c3547ab96a71d5ae8a49dfd5ca64506b03
```

DanGrayson:
Meh, what a pain.  I might as well just use `git diff`.

mahrud:
> Meh, what a pain.  I might as well just use `git diff`.

With the link I sent you can leave your review via github, but sure, now that you know the common ancestor you can use whatever.

M2/Macaulay2/packages/VirtualResolutions.m2
```diff
@@ -137,46 +137,30 @@ resolveViaFatPoint(Ideal, Ideal, List) := ChainComplex => (J, irr, A) -> (
...
+isVirtual(Ideal,              ChainComplex) := opts -> (irr, C) -> (
+    S := ring irr;
+    if S =!= ring C then error "isVirtual: expected objects in the same ring";
 -- if strategy "determinantal is selected, the method checks virtuality
 -- via the depth criterion on the saturated ideals of minors
+    debugInfo := if debugLevel < 1 then identity else i -> printerr("isVirtual failed at homological degree ", toString i);
     if opts.Strategy === "Determinantal" then (
         for i from 1 to length(C) do (
             if rank(source(C.dd_i)) != (rank(C.dd_i) + rank(C.dd_(i+1))) then (
-                if debugLevel >= 1 then print "isVirtual failed at homological degree " | toString i;
```
DanGrayson, line 147: Why does failure not result in an error?

M2/Macaulay2/packages/VirtualResolutions.m2
```diff
@@ -137,46 +137,30 @@ resolveViaFatPoint(Ideal, Ideal, List) := ChainComplex => (J, irr, A) -> (
...
-                return false;
-                );
-            );
+                debugInfo i; return false));
         for i from 1 to length(C) do (
             minor := minors(rank(C.dd_i),C.dd_i);
             minorSat := ourSaturation(minor,irr);
-            if depth(minorSat,ring(minorSat)) < i then (
-                if debugLevel >= 1 then print "isVirtual failed at homological degree " | toString i;
```
DanGrayson, line 155: here, too

M2/Macaulay2/packages/VirtualResolutions/doc.m2
```diff
@@ -1,7 +1,4 @@
-undocumented {
-    (multigradedRegularity, Thing, Ideal),
-    (multigradedRegularity, Thing, Thing, Module)
-    }
+undocumented { "LowerLimit", "UpperLimit" }
```
DanGrayson, line 1: document these

If they are options to just one method, just point the documentation to the same node.  If not, just say "This is a symbol used as an optional argument name."  Or something like that.

DanGrayson (changes requested):

M2/Macaulay2/packages/VirtualResolutions.m2
mahrud: `isVirtual` returns a boolean. Better would be to say "virtuality failed" ...

mahrud:
Done. Now the common ancestor is d54ff8f

M2/Macaulay2/packages/VirtualResolutions.m2
DanGrayson: Hmm?  If you mean `isVirtual` returned `false`, just say that -- nothing failed.  Unless you think virtuality is a virtue and lack of virtuality is a vice.

mahrud: You're looking at the old version!

DanGrayson: No, I'm responding to this comment of yours: "Better would be to say "virtuality failed"".

mahrud: This conversation is moot. If you look at the current version, you'll instead see this:
https://github.com/Macaulay2/M2/blob/071bc5d6f6da278c356d3446b10ed30c1d70233c/M2/Macaulay2/packages/VirtualResolutions.m2#L147

Being a virtual resolution is equivalent to being exact as a locally-free complex of sheaves. In many areas of math people say something "fails to be exact" or "exactness fails". If the complex was found not to be virtual at that a degree, then it failed to be virtual. There's no vice is virtue in failure.

mahrud:
I accidentally pushed a commit here, then reverted it, but the previous tests are still running (successfully) here: https://github.com/Macaulay2/M2/actions/runs/394328398

M2/Macaulay2/packages/VirtualResolutions.m2
mahrud: > here, too

This is also old code, by the way.

DanGrayson: Do you mean that it is never entered?  If so, we could get rid of it.

mahrud: I don't understand what you mean. You left this review directly on the old code, and so there's nothing I can do about this. This segment of the new code has been majorly changed. Have you seen the new code?

mahrud:
> I accidentally pushed a commit here, then reverted it, but the previous tests are still running (successfully) here: https://github.com/Macaulay2/M2/actions/runs/394328398

The tests have passed.

M2/Macaulay2/packages/VirtualResolutions.m2
DanGrayson: Oh, now I see what you mean!

mahrud: Great! Any other suggestions on this pull request?

DanGrayson (approved):

mahrud:
@DanGrayson should I merge this before midnight? Everything is approved and tested.

DanGrayson:
> @DanGrayson should I merge this before midnight? Everything is approved and tested.

No.

PR #1620, opened 2020-11-25
Package NoetherianOperators v1.0

haerski:
@jchen419 @kroner @antonleykin @ycid


M2/Macaulay2/packages/NoetherianOperators.m2
```diff
@@ -0,0 +1,3649 @@
...
+    "numericalNoetherianOperators",
+    "noethOpsFromComponents",
+    "coordinateChangeOps",
+    "rationalInterpolation",
+    "InterpolationTolerance",
+    "InterpolationDegreeLimit",
+    "NoetherianDegreeLimit",
+    "DependentSet",
+    "KernelStrategy",
+    "BM",
```
DanGrayson, line 59: Can you come up with a more descriptive name?

M2/Macaulay2/packages/NoetherianOperators.m2
```diff
@@ -0,0 +1,3649 @@
...
+	   In case of an ideal supported at one point a set of Noetherian operators forms a Macaulay inverse system that spans the dual space of the ideal. 
+	   These notions relate to the work of 
+	   Macaulay ("The algebraic theory of modular systems", Cambridge Press, (1916)). 
+	    
+	   In this package, we implement several (exact symbolic and approximate numerical) algorithms for the computation of a set of Noetherian operators describing a primary ideal. 
+    	           	   
+           Methods and types for computing and manipulating Noetherian operators:
+
+           @UL {
+               {TO noetherianOperators, UL {
```
DanGrayson, line 1497: How does this list look in the html form of the documentation?  I'm suspicious, because the items in a list should each be inside an `LI` item.

M2/Macaulay2/packages/NoetherianOperators.m2
jchen419: Hi Dan, just to clarify: did you want a more descriptive name just for the option "BM", or the other ones as well? (Would something like "useBMStrategy" work?)

M2/Macaulay2/packages/NoetherianOperators.m2
jchen419: At least on my computer, it seems to look OK (ignore the dark mode theme): 

noethOps_doc

mahrud: By the way, you can also use the "Tree" keyword of SimpleDoc for this purpose.

M2/Macaulay2/packages/NoetherianOperators.m2
DanGrayson: My point was that the acronym "BM" is itself mysterious, so encapsulating it in something larger doesn't seem to be an improvement.

M2/Macaulay2/packages/NoetherianOperators.m2
DanGrayson: Okay, it seems UL is eager to wrap items inside LI, so there is no problem.

DanGrayson:
Let me know when you'd like this pull request to be considered again, after your updates.

M2/Macaulay2/packages/NoetherianOperators.m2
haerski: @DanGrayson Would `IntegralStrategy` be an acceptable replacement for `BM` ?

DanGrayson: Yes, unless you expect the user to type `Strategy => IntegralStrategy`, in which case `Strategy => Integral` would look better.

haerski:
@DanGrayson The package is updated. I changed the symbol `BM` to `IntegralStrategy`. It has to be a separate from `Strategy` since the choice of `Strategy` and `IntegralStrategy` are independent.

PR #1628, opened 2020-11-28
Change uses of \bf in math mode

mahrud:
See https://github.com/Macaulay2/M2/issues/1529#issuecomment-735258064


mahrud:
cc: @pzinn

DanGrayson:
@pzinn -- does this work for you?

@mahrud -- I don't see `\PP` at https://katex.org/docs/supported.html -- doesn't that mean it won't work with Katex?

mahrud:
> @mahrud -- I don't see `\PP` at https://katex.org/docs/supported.html -- doesn't that mean it won't work with Katex?

We can define arbitrary macros in the header:

https://github.com/Macaulay2/M2/blob/855e1c3f83a51f061ff07227a8df3b88a4e50cd2/M2/Macaulay2/m2/html.m2#L14-L23

The choice of which macros to support there was by you:
https://github.com/Macaulay2/M2/blob/8dde6c2b228e073809c5ff1d5e9e3f0cd4a62516/M2/Macaulay2/m2/texhtml.m2#L114-L121

We can always add more or remove the unused ones. If the list starts to get too long, we can also move it to an external file so the browser can cache it and not retrieve it repeatedly (though with modern browser compression, this is really negligible)

DanGrayson:
Okay, good, it will work with katex.  But this sentence from your documentation should probably mention the other macros we define: "See the list of supported functions and symbols for more information, or this page for an introduction to math mode in $\LaTeX$", since it gives the impression that that list is complete.

pzinn:
It works fine. Though I don't think one-letter macros such as `\R` and `\C` are such a great idea, especially since they're redundant with `\RR` and `\CC`.

DanGrayson:
I agree -- there's no reason to have both `\R` and `\RR`.

mahrud:
Does arXiv still use `\R` and `\C`? That was the comment that used to be there. Either way, feel free to make the changes you wish. I probably won't have time for it before this release.

DanGrayson:
My only application for the Arxiv is our citations on the web site -- I can find a work-around.  Coincidentally, the machine I process those on is running M2 version 1.6 still, from May, 2013.  And it works.

Issue #1631, opened 2020-11-29
check_0 "EigenSolver" fails on i386


d-torrance:
I got a build failure on a PPA build of the `development` branch on Ubuntu 18.04 (i386) after #1615 was merged:

```m2
if ! grep "CacheExampleOutput => true" ./EigenSolver.m2 >/dev/null ; then /<<BUILDDIR>>/macaulay2-1.16.0.2+git202011281728/M2/usr-dist/i686-Linux-Ubuntu-18.04/bin/M2 -q --no-preload --stop --silent -e "needsPackage(\"EigenSolver\",LoadDocumentation=>true,DebuggingMode=>true); check(EigenSolver,UserMode=>false,Verbose=>true); exit 0" ; fi 
-* running test 0 of package EigenSolver in file:
   EigenSolver.m2:142:1:
   rerun with: check_0 "EigenSolver" *-
 -- making test results
 ulimit -c unlimited; ulimit -t 700; ulimit -m 850000; ulimit -s 8192; ulimit -n 512;  cd /tmp/M2-7955-0/2-rundir/; GC_MAXIMUM_HEAP_SIZE=400M "/<<BUILDDIR>>/macaulay2-1.16.0.2+git202011281728/M2/usr-dist/i686-Linux-Ubuntu-18.04/bin/M2-binary" -q --int --no-randomize --no-readline --silent --stop --print-width 77 -e 'needsPackage("EigenSolver", Reload => true, FileName => "/<<BUILDDIR>>/macaulay2-1.16.0.2+git202011281728/M2/Macaulay2/packages/EigenSolver.m2")' <"/tmp/M2-7955-0/1.m2" >>"/tmp/M2-7955-0/1.tmp" 2>&1
/tmp/M2-7955-0/1.tmp:0:1: (output file) error: Macaulay2 exited with status code 1
/tmp/M2-7955-0/1.m2:0:1: (input file)
M2: *** Error 1
-* running test 1 of package EigenSolver in file:
   EigenSolver.m2:149:1:
   rerun with: check_1 "EigenSolver" *-
 -- making test results
-* running test 2 of package EigenSolver in file:
   EigenSolver.m2:171:1:
   rerun with: check_2 "EigenSolver" *-
 -- making test results
-* running test 3 of package EigenSolver in file:
   EigenSolver.m2:191:1:
   rerun with: check_3 "EigenSolver" *-
 -- making test results
[...]

PR #1633, opened 2020-11-29
Hardcode the multiplier used in an EigenSolver test

d-torrance:
Otherwise, one is randomly chosen.  The one chosen with random seed 0
on i386 systems results in a failing test, as it finds 4 real points
instead of the expected 6.

We use the multiplier that would be chosen with a random seed of 1 instead,
which works on both i386 and amd64.

Closes: #1631


DanGrayson:
Good fix, but let's figure out also how to get floating point behavior to be identical on the two systems.

jchen419:
Hi Doug, thanks for the fix, and the diligence in finding these bugs! (I have no access to other hardware, and thus appreciate these robust checks.)

mahrud:
> Good fix, but let's figure out also how to get floating point behavior to be identical on the two systems.

Is this worth the effort? Ubuntu doesn't have 32-bit releases anymore.

d-torrance:
> > Good fix, but let's figure out also how to get floating point behavior to be identical on the two systems.
> 
> Is this worth the effort? Ubuntu doesn't have 32-bit releases anymore.

Ubuntu 18.04 did, and its end of life isn't until April 2028 (https://wiki.ubuntu.com/Releases).  Debian supports three 32-bit architectures officially (i386, armhf, and mipsel), and several more unofficially.

PR #1641, opened 2020-12-01
fewer github builds

DanGrayson:


mahrud:
I think you should put cmake-macos-latest-clang10 back in. Both clang builds were nontrivial to get working, and I would like them not to break unintentionally.

If you'd like to reduce the number of builds, you should make a separate workflow that maybe runs 1 or 2 jobs, and perhaps when it is merged run a full workflow with 8 jobs.

.github/workflows/test_build.yml
```diff
@@ -194,7 +182,7 @@ jobs:
         if: always()
         uses: actions/upload-artifact@v2
         with:
-           name: ${{ matrix.os }}-${{ matrix.build-system }}-${{ matrix.compiler }}-logs
+           name: ${{ matrix.build-system }}-${{ matrix.os }}-${{ matrix.compiler }}-logs
```
mahrud, line 185: No, this is very much incorrect and should be fixed!

This implies that caches for cmake-macos are compatible with caches for cmake-linux, which is not the case!

mahrud: Also, would be good if you allowed some time for review before merging.

mahrud: In particular, this is completely incompatible with this:
https://github.com/Macaulay2/M2/blob/1e47260755100936bafd60d64013672dedd3941f/.github/workflows/test_build.yml#L125-L129

mahrud: Ah, never mind, I forgot that the build cache key is defined separately from the log key. It's not a big deal, but it is confusing that the keys are ordered differently.

DanGrayson: I'd be happy to reorganize the lines concerned with caching, if you propose a change.

PR #1647, opened 2020-12-03
Only require one variable for verbosity in autotools build

d-torrance:
Previously, the autotools build had two variables controlling the
verbosity of the build:

* `VERBOSE`: See `M2/include/config.Makefile.in`.  Controls the output during
  compilation, e.g., `CXX foo.cc` v.
  `g++ (lots of build flags) foo.cc -o foo.o`.

* `Verbose`: See `M2/Macaulay2/packages/Makefile.in`.  Controls the
  output when building examples and testing packages.  In particular,
  whether to display what happened when these fail.

We make it so that only the former is necessary.  This brings the
autotools build in line with the behavior of the CMake build, which
uses `$<IF:$<BOOL:${VERBOSE}>,true,false>` to determine the value of
`Verbose`.


d-torrance:
I haven't seen this failing test before:

```m2
 -- making test results../../../../Macaulay2/m2/debugging.m2:20:6:(1):[9]: error: test #3 of package MultiGradedRationalMap failed
../../../../Macaulay2/m2/testing.m2:59:22:(1):[8]: --back trace--
../../../../Macaulay2/m2/methods.m2:119:80:(1):[7]: --back trace--
../../../../Macaulay2/m2/option.m2:16:8:(1):[6]: --back trace--
../../../../Macaulay2/m2/testing.m2:44:33:(1):[5]: --back trace--
../../../../Macaulay2/m2/methods.m2:119:80:(1):[4]: --back trace--
../../../../Macaulay2/m2/option.m2:16:8:(1):[3]: --back trace--
currentString:1:85:(3):[2]: --back trace--
Macaulay2/Core/startup.m2.in:563:33:(0):[1]: --back trace--
Macaulay2/Core/startup.m2.in:674:6:(0): --back trace--
make[2]: *** [check-MultiGradedRationalMap] Error 1
Makefile:39: recipe for target 'check-MultiGradedRationalMap' failed
```

mahrud:
> I haven't seen this failing test before:

I've seen it a few times. Seems completely random. Probably worth making an issue to investigate.

d-torrance:
It also happened in #1648!  The artifacts don't include the output of the failed check...

I'm trying to reproduce the issue locally, but no luck yet. Over 7000 tries and counting!

PR #1653, opened 2020-12-04
Drop executable bits on two new package source files

d-torrance:


mahrud:
This one also seems to have +x: `Macaulay2/packages/Visualize/css/BootSideMenu.css`

d-torrance:
> This one also seems to have +x: `Macaulay2/packages/Visualize/css/BootSideMenu.css`

Nice catch -- thanks!

Issue #787, opened 2018-05-24
Posets package: Bug in rankFunction method


bbarwick:
A poset P is **ranked** if there exists an integer-valued function r : V(P) -> ZZ such that

1. The function r respects the partial order (i.e., if u < v, then r(u) < r(v)).
2. If v covers u, then r(v) = r(u)+1.

The following example demonstrates a bug in the `Posets` package in which the rank function returned by `rankFunction` violates part 2 of the definition above.

`i18 : P = poset({3, 4, 5, 6, 7, 8, 9, 10, 11},{{4, 7}, {4, 8}, {5, 11}, {5, 9}, {6, 4}, {6, 9}, {7, 3}, {8, 3}, {9, 7}, {9, 8}, {10, 5}, {11, 4}})`

`o18 = P`

`o18 : Poset`

`i19 : isRanked P`

`o19 = true`

`i20 : vertices P`

`o20 = {3, 4, 5, 6, 7, 8, 9, 10, 11}`

`o20 : List`

`i21 : rankFunction P -- Notice that 6 and 9 both have rank 2.`

`o21 = {5, 3, 1, 2, 4, 4, 2, 0, 2}`

`o21 : List`

`i22 : coveringRelations P -- But 9 covers 6.`

`o22 = {{4, 7}, {4, 8}, {5, 11}, {5, 9}, {6, 4}, {6, 9}, {7, 3}, {8, 3}, {9, 7}, {9, 8}, {10, 5}, {11, 4}}`

`o22 : List`

DanGrayson:
@dcookii -- who should be looking at this?

dcookii:
{Name => "Sonja Mapes", Email => "smapes1@nd.edu"} or {Name => "Gwyn
Whieldon", Email => "whieldon@hood.edu"}


On Thu, May 24, 2018 at 2:34 PM, Daniel R. Grayson <notifications@github.com

bbarwick:
Here is a slightly simpler example to demonstrate the bug using less vertices and such that the vertex labels correspond to their indices.  The `isRanked` method returns that this poset is ranked, but it actually isn't (this is easy to see since it has maximal chains of different lengths).

`i2 : P = poset({0,1,2,3,4},{{0,2},{2,3},{1,4},{0,4},{1,3}})`
`o2 = P`

`o2 : Poset`
`i3 : isRanked P`
`o3 = true`

`i4 : vertices P`
`o4 = {0, 1, 2, 3, 4}`
[...]

PR #1657, opened 2020-12-04
fixing isRanked in Posets

fchapoton:
fix the issue as suggested in #787


DanGrayson:
@dcookii @smapes @gwynwhieldon -- are these changes okay with you?

gwynwhieldon:
Looks good to me.

dcookii:
Fine with me.  Thanks for the fix @fchapoton!

Issue #1649, opened 2020-12-03
Popular symbols


antonleykin:
This is a proposal to
1. find symbols that are frequently used by package authors (e.g., for method/option names) 
2. define them in the `Core` (in case of methods we need to think about a default way of doing this)
3. standardize our approach to this

As two examples I have 
- `Jacobian` defined in `ReesAlgebra`
- `truncate` defined in `Truncations`
In both cases I'm forced to "export" the corresponding packages just in order to use these symbols... or face the "shadow" warning (and be forced to use prefixes) --- these two packages load at startup.

DanGrayson:
see commit f3db59552f1bc1fbc8eba7bf3504502c639a40ff

antonleykin:
Thanks! 
One minor issue: documentation for the symbol, say, `Jacobian` is still defined in `ReesAlgebra`.
Ideally, it'd be good to have the main node for `Jacobian` say something like "an option used in packages" such and such.
Even better would be to automate creation of such documentation nodes.

mahrud:
1. Here's a not yet fully formed idea I've had for a while: can we modify the code that handles optional arguments to not require the symbols to be exported at all? The reason I ask this is that for functions, for instance, the dictionary has a pointer to the function itself:
```m2
i3 : value Core.Dictionary#"gb"

o3 = gb

o3 : MethodFunctionWithOptions
```
But for symbols, it's just a symbol, not pointing to anything else:
```m2
i4 : value Core.Dictionary#"Generic"

o4 = Generic

o4 : Symbol
```
So what's the point of having it there at all? In other words, I'm wondering if there's any advantage in having non-equal symbols with the same name. The only thing I can think of is that by exporting, symbols become protected, but again, what's the benefit of that?

I realize that this is not a fully formed idea, but perhaps relaxing the requirement to export symbols can simplify this issue.

2. Why should symbols have documentation? Consider this hypothetical example: suppose `(resolution, Module)` can take an option `Strategy => FastNonminimal`. I think only the following documentation nodes would need to be created:
```m2
  Module
  resolution
 (resolution, Module)
[(resolution, Module), Strategy]
[(resolution, Module), Strategy => FastNonminimal]
```
That's it. In particular, I don't see the point of forcing package authors to add a node for `Strategy` or `FastNonminimal`, because they have no intrinsic meaning outside the context of this method, other than their name.

[...]

Issue #1377, opened 2020-07-07
Relocate the portion of Truncations that uses Polyhedra


mahrud:
When M2 loads, a number of packages are preloaded. I believe this is done by `last.m2` when it reads [m2/installedPackages](https://github.com/Macaulay2/M2/blob/master/M2/Macaulay2/m2/installedpackages). This lists corresponds to the initial value of `loadedPackages` (minus Core).

However, this doesn't quite correspond to the list of packages printed when M2 starts (Classic and SimpleDoc are missing) nor does it include about 7 packages that are loaded as dependencies and visible via `loadedFiles`: Parsing, Text, FourTiTwo, FourierMotzkin, Normaliz, Topcom, and Polyhedra.

In particular, in order to define Ext and cohomology, `varieties.m2` and `local.m2` call needsPackage on Truncations, which in turn loads Polyhedra's ~60 files. This could be simplified to speed up load times. Here are three ways I can think of:
- move Ext and cohomology elsewhere (e.g. Complexes?);
- use autoload instead of needsPackage;
- relocate the portion of Truncations that uses Polyhedra to the Polyhedra package.

mahrud:
> * use autoload instead of needsPackage;

Related to this, perhaps it would be good to have a version of [`autoload`](http://www2.macaulay2.com/Macaulay2/doc/Macaulay2-1.16/share/doc/Macaulay2/Macaulay2Doc/html/_autoload_lp__Symbol_cm__String_rp.html) for method functions, in particular `truncate`.

mahrud:
@DanGrayson the suggestion about would be good to implement, if you can think of a way to do it. Or any of the alternatives in the top comment.

DanGrayson:
Loading Polyhedra takes only 0.13 seconds:
```m2
+ M2 --no-readline -q --print-width 218 --no-preload
Macaulay2, version 1.16.0.2

i1 : time loadPackage "Polyhedra"
     -- used 0.130304 seconds

o1 = Polyhedra

o1 : Package
```

Is it worth worrying about?

mahrud:
Yes, try uninstalling all packages first! It takes about 4 seconds on my machine.
The user would obviously get M2 with all packages preinstalled, but the problem is that while examples are being generated, new instances of M2 are starting and loading about 70 files (Polyhedra alone has 60 files) that they don't need, unless they call the `truncate` command.

DanGrayson:
[...]

PR #1661, opened 2020-12-06
Relax requirements for documentation nodes

### LINKED ISSUE #192: exporting methods
It would be nice if packages could add specified methods to the automatically generated lists of hyperlinks even when the methods are installed on previously existing classes and method functions.

eg 

(resolutions, ChainComplex)

as an element of the export list.

### LINKED ISSUE #818: determining the package of a method function
Here is an example of a common problem: the package Functoriality defines a method for Tor(ZZ, Matrix, Module), but all of the elements of that are defined in the Core package, so we don't permit the package to document it, and we can't tell that the method function comes from the package.

Even more odd is that the Core defines a method for Tor(ZZ, Matrix, Ideal), which doesn't work unless Functoriality is added.

### LINKED ISSUE #1219: configuration notification at start.
```
+ M2 --no-readline --print-width 212
Macaulay2, version 1.15
--loading configuration for package "FourTiTwo" from file /Users/dan/Library/Application Support/Macaulay2/init-FourTiTwo.m2
--loading configuration for package "Topcom" from file /Users/dan/Library/Application Support/Macaulay2/init-Topcom.m2
with packages: ConwayPolynomials, Elimination, IntegralClosure, InverseSystems, LLLBases, PrimaryDecomposition, ReesAlgebra, TangentCone, Truncations
```

Let's hush notifications about configurations being loaded, unless `notify` is true.

Suggested by Mike.

### LINKED ISSUE #1528: To figure out that the proper package for the documentation of "integralClosure(Ideal)" is "IntegralClosure" presents a problem.  From
This code is wrong:
```
     /Applications/Macaulay2-1.15/share/Macaulay2/Core/packages.m2:470:23-470:44: --source code:
     package Sequence := s -> youngest (package\s)
```
because it depends on the hash code of the package object.  Hash codes always increase as time goes on, but objects of type Package may not be created in dependency order, chronologically.  The Package object is created by "newPackage", and *after* that, other packages may be loaded by the code inside the package, or by the newPackage function taking care of the PackageExports and PackageImports options.

Here is a way to get the hashcodes of the packages in the right order: we update the hashcode in the endPackage function. To make a new Package object from an old one this code suffices: new Package from new MutableHashTable from Foo. There shouldn't be any pointers to the old package lying around, but in case there are, it might be good instead to have "newPackage" make the package of type MutableHashTable instead of being of type Package, and then in "endPackage" we could do new Package from Foo instead.

mahrud:
- Vaguely related to the discussion in #1649:
  - allow documenting `[quotient, Strategy => Iterate]`
  - allow documenting `[(quotient, Ideal, Ideal), Strategy => Iterate]`
- removes extra spaces from the formatted key for optional arguments
- corrects two or three minor bugs
- close #1528
- close #1219
- close #818
- probably also close #192
- partially alleviate #1377, but probably keeping that one open for a better fix


mahrud:
Also, the links at the bottom here weren't working before:
image

M2/Macaulay2/packages/Macaulay2Doc.m2
```diff
@@ -12,6 +12,7 @@ newPackage ("Macaulay2Doc",
 	  {Name => "Michael E. Stillman", Email => "mike@math.cornell.edu", HomePage => "http://www.math.cornell.edu/People/Faculty/stillman.html"}
 	  },
      Keywords => {"Miscellaneous"},
+     PackageExports => Core#"pre-installed packages",
```
mahrud, line 15: This is necessary in order to get the correct "ways to use" links, but it also slows down installing the examples it seems like:
```m2
 -- making example results for "isModule"                                    -- 4.83991 seconds elapsed
 -- making example results for "leadCoefficient"                             -- 4.83778 seconds elapsed
 -- making example results for "gradedModule"                                -- 4.82856 seconds elapsed
```

@DanGrayson is there a way to force `beginDocumentation()` to end loading the file even if the package isn't installed yet?

mahrud: Or maybe I should add a global variable or command line option for that purpose?

mahrud: I remembered that the main culprit is the Truncation package loading Polyhedra, and given the recent changes with `shared.m2`, I removed Truncation from preloaded packages. Now it should run fast again.

mahrud:
I added `Macaulay2Doc` examples only to be generated using capture. I haven't looked at every generated example output, but at least there are no error messages missing and no errors apparent.

M2/Macaulay2/packages/Macaulay2Doc.m2
DanGrayson: > @DanGrayson is there a way to force `beginDocumentation()` to end loading the file even if the package isn't installed yet?

Yes, it could return `end`.  But to what purpose?

mahrud: To make the examples run faster. Currently, when Macaulay2Doc examples are being generated, no other packages are installed, so in particular the "Truncation" package causes a whole bunch of other packages to be read, _including_ their documentation. This PR makes reading the documentation somewhat slower, specifically because `makeDocumentTag` and `(package, Sequence)` are more thorough. These changes are necessary because they fix various issues (see the top comment), so instead removing "Truncations" made the situation is mostly bearable again.

DanGrayson:
Mahrud, can you figure out why one of the tests for the "Saturation" package failed?

mahrud:
> Mahrud, can you figure out why one of the tests for the "Saturation" package failed?

Because I had forgotten to load Truncations in a test that used it.

M2/Macaulay2/m2/local.m2
```diff
@@ -4,8 +4,8 @@
 
 truncatedDual := (M,e) -> (
      -- find (k-dual M), truncated in degrees >= e.
+     -- depends on truncate methods
      needsPackage "Truncations";
```
DanGrayson, line 8: This call to `needsPackage` is redundant, because Truncations is a pre-loaded package.

And it's a hack.

It would be good to move this code into a package that explicitly imports "Truncations".

mahrud: Sorry if I wasn't clear, but this PR removes Truncations from the preloaded list in order to increase speed.

mahrud: See #1377.

PR #1664, opened 2020-12-06
Replace build paths that appear in docs for option default values

d-torrance:
Otherwise, we will have things like:

    InstallPrefix => a string, default value
      /build/macaulay2-1.16.0.2+git815.1b4bbe4+ds/debian/.debhelper/
      generated/_source/home/.Macaulay2/local/, ...

This results in non-reproducible builds.


d-torrance:
With this commit:

image

PR #1667, opened 2020-12-08
check for Prem from Factory (HAVE_FACTORY_PREM)

mahrud:
See https://github.com/Macaulay2/M2/pull/1538#discussion_r537901750.


mahrud:
@d-torrance @DanGrayson I believe this can be reverted, now that we require Factory 4.2.0.

PR #1678, opened 2020-12-10
Update emacs and vim files for v1.17

mahrud:


DanGrayson:
I don't understand this pull request, because the repository it appears to come from is not a fork of this one:

Screen Shot 2020-12-10 at 6 51 10 AM

PR #1679, opened 2020-12-10
Include flint version in "version" hash table

d-torrance:
We have been marking it as "not present" if the HAVE_FLINT macro wasn't
defined, but it hasn't been since 2014!  See 3e2fd2e.

### DIFF
--- a/M2/Macaulay2/d/version.dd
+++ b/M2/Macaulay2/d/version.dd
@@ -55,13 +55,11 @@ header "
        #include <linbox/linbox-config.h>
    #endif
    // HAVE_FROBBY is in factory/factoryconf.h
-   #ifdef HAVE_FLINT
-       #include <M2/gc-include.h>
-       #pragma GCC diagnostic push
-       #pragma GCC diagnostic ignored \"-Wconversion\"
-       #include <flint/flint.h>
-       #pragma GCC diagnostic pop
-   #endif
+   #include <M2/gc-include.h>
+   #pragma GCC diagnostic push
+   #pragma GCC diagnostic ignored \"-Wconversion\"
+   #include <flint/flint.h>
+   #pragma GCC diagnostic pop
    ";
 gcversion():constcharstar := Ccode(returns,"
      static char buf[40];
@@ -192,13 +190,7 @@ setupconst("version", Expr(toHashTable(Sequence(
 	"),
    "ntl version" => Ccode(constcharstar,"NTL_VERSION"),
    "frobby version" => Ccode(constcharstar,"constants::version"),
-   "flint version" => Ccode(constcharstar,"
-	 #if HAVE_FLINT
-	   flint_version
-	 #else
-	   \"not present\"
-	 #endif
-	 "),
+   "flint version" => Ccode(constcharstar,"flint_version"),
    "scscp version" => Ccode(constcharstar,"
 	 #ifdef HAVE_SCSCP
 	   stringize(SCSCP_VERSION_MAJOR) \".\" stringize(SCSCP_VERSION_MINOR) \".\" stringize(SCSCP_VERSION_PATCH)


DanGrayson:
Hmm, I see it:
```m2

+ M2-1.16 --no-readline --print-width 218 -q
Macaulay2, version 1.16
with packages: ConwayPolynomials, Elimination, IntegralClosure, InverseSystems, LLLBases, MinimalPrimes, PrimaryDecomposition, ReesAlgebra, TangentCone, Truncations

i1 : version#"flint version"

o1 = 2.6.0

i2 : 

+ M2-1.15 --no-readline -q --print-width 218
Macaulay2, version 1.15
with packages: ConwayPolynomials, Elimination, IntegralClosure, InverseSystems, LLLBases, PrimaryDecomposition, ReesAlgebra, TangentCone, Truncations

i1 : version#"flint version"

o1 = 2.5.2

i2 : 

Process M2 finished

+ M2 --no-readline --print-width 107 -q
Macaulay2, version 1.16.0.2
with packages: ConwayPolynomials, Elimination, IntegralClosure, InverseSystems, LLLBases, MinimalPrimes,
               PrimaryDecomposition, ReesAlgebra, Saturation, TangentCone

i1 : version#"flint version"

o1 = 2.6.3
```

DanGrayson:
I figured it out: HAVE_FLINT is defined for us by factory, in `factoryconf.h`.

d-torrance:
That makes sense -- the only other places that I was seeing `HAVE_FLINT` in the source was where we were dealing with building factory.  But when we don't build factory, it's not defined.  For example, from my local installation, which uses factory from the Ubuntu 20.10 singular package:

```m2
Macaulay2, version 1.16.0.2
               TangentCone, Truncations

i1 : version#"flint version"

o1 = not present
```

d-torrance:
And here's the corresponding lines from the Ubuntu `factoryconf.h`: 

```
/* Define if FLINT is installed */
/* #undef HAVE_FLINT */
```

DanGrayson:
So I guess that version of factory is not using flint, but we are (somewhere else).  Makes sense.

DanGrayson:
So, if a user runs `apt-get install macaulay2`, will all of singular get installed?  If so, that's not a good look.

d-torrance:
> So, if a user runs `apt-get install macaulay2`, will all of singular get installed? If so, that's not a good look.

Unless they specify `--no-install-recommends`, then yes.  The [`macaulay2`](https://packages.debian.org/sid/macaulay2) package depends on [`singular-data`](https://packages.debian.org/sid/singular-data) (for gftables), which recommends [`singular-ui`](https://packages.debian.org/sid/singular-ui), which contains `/usr/bin/Singular`.

d-torrance:
I could probably drop the `singular-data` dependency by just including `gftables/961` in the `debian` directory and pointing `configure` there during build.  Would that be better?

d-torrance:
> I could probably drop the `singular-data` dependency by just including `gftables/961` in the `debian` directory and pointing `configure` there during build. Would that be better?

Hrm, but it looks like not having `singular-data` installed would slow down factory.  From its Makefile:

```
####################################################
# Support for building GF(q)-tables
#
# gengftables needs libfactory.la and factory.h AND is needed for
# generating gftable but the tables are not necessarily needed to
# compile and use libfactory (though it will be a lot slower).
```

DanGrayson:
We definitely need all the gf tables.

Ideally factory and gftables would be debian packages, too, and we could just use them.  But maybe now is not the time to arrange that, if time is tight.

mikestillman:
It looks like flint will include multivariate polynomial factorization in this upcoming release, although perhaps it is still experimental.  Perhaps that means factory will be subsumed by this (Bill Hart works with the Singular group).

d-torrance:
> We definitely need all the gf tables.
> 
> Ideally factory and gftables would be debian packages, too, and we could just use them. But maybe now is not the time to arrange that, if time is tight.

That should be pretty easy to do.  We could still just have one singular source package, but split out the factory stuff into a few binary packages (probably something like `libsingular-factory4`, `libsingular-factory-dev`, and maybe `singular-factory-data`) that Macaulay2 could depend on.

Because of the new binary packages, it would have to go through the NEW queue again.  So it's probably best to wait until after bullseye is released and shoot for Debian 12 (bookworm).  I'd also have to make sure it's okay with the other interested parties -- in particular the Sage maintainers.

DanGrayson:
That sounds good!  How about making an issue for that and assigning yourself to it?

d-torrance:
> That sounds good! How about making an issue for that and assigning yourself to it?

Done.  I don't think I have the privileges to assign myself to an issue, though.

Issue #1685, opened 2020-12-12
check_18 "GraphicalModelsMLE" failing on several architectures


d-torrance:
I've seen the following on several architectures (i386, arm64, [ppc64el](https://launchpadlibrarian.net/510941465/buildlog_ubuntu-groovy-ppc64el.macaulay2_1.16.0.2+git202012111652-0ppa202012101516~ubuntu20.10.1_BUILDING.txt.gz), and [s390x](https://launchpadlibrarian.net/510933088/buildlog_ubuntu-groovy-s390x.macaulay2_1.16.0.2+git202012111652-0ppa202012101516~ubuntu20.10.1_BUILDING.txt.gz) so far).

```m2
-* running test 18 of package GraphicalModelsMLE in file:
   GraphicalModelsMLE.m2:1797:1:
   rerun with: check_18 "GraphicalModelsMLE" *-

 ulimit -c unlimited; ulimit -t 700; ulimit -m 850000; ulimit -s 8192; ulimit -n 512;  cd /tmp/M2-4090381-0/38-rundir/; GC_MAXIMUM_HEAP_SIZE=400M "/build/macaulay2-1.16.0.2+git1077.7e49bfd+ds/M2/usr-dist/i686-Linux-Debian-unstable/bin/M2-binary" -q --int --no-randomize --no-readline --silent --stop --print-width 129 -e 'needsPackage("GraphicalModelsMLE", Reload => true, FileName => "/build/macaulay2-1.16.0.2+git1077.7e49bfd+ds/M2/Macaulay2/packages/GraphicalModelsMLE.m2")' <"/tmp/M2-4090381-0/37.m2" >>"/tmp/M2-4090381-0/37.tmp" 2>&1
/tmp/M2-4090381-0/37.tmp:0:1: (output file) error: Macaulay2 exited with status code 1
/tmp/M2-4090381-0/37.m2:0:1: (input file)
M2: *** Error 1
../m2/debugging.m2:20:6:(1):[9]: error: test #18 of package GraphicalModelsMLE failed:

               | 2.26215      0      3.26493      14.6143 |

o3 : Sequence

i4 : assert(round(5,mx)==-8.4691)

i5 : assert(ML==1)

i6 : assert(MLE_(2,0)==0)
stdio:8:1:(3): error: assertion failed

../m2/testing.m2:59:22:(1):[8]: --back trace--
../m2/methods.m2:119:80:(1):[7]: --back trace--
[...]

PR #1687, opened 2020-12-12
Round matrix entry in GraphicalModelsMLE test

d-torrance:
Due to variations between floating point operations between different
architectures, this entry may not be exactly 0, e.g., it's
-2.03212e-17 on i386.

Closes: #1685


DanGrayson:
The floating point architecture should be the IEEE standard, so it's probably compiler bugs.

Issue #1686, opened 2020-12-12
"Verbose" option to "check" doesn't work when running a single test


d-torrance:
I discovered this while investigating #1685.

Suppose we have a package with at least two tests like the following in `bar.m2`:

```m2
newPackage "Bar";
TEST "assert(1 + 1 == 2)"
TEST "assert(2 + 2 == 5)"
```

Then we get the following:

```m2
i1 : loadPackage("Bar", FileName => "~/tmp/bar.m2");

i2 : check("Bar", Verbose => true)
-* running test 0 of package Bar in file:
   ../tmp/bar.m2:3:1:
   rerun with: check_0 "Bar" *-
-* running test 1 of package Bar in file:
   ../tmp/bar.m2:4:1:
   rerun with: check_1 "Bar" *-

 ulimit -c unlimited; ulimit -t 700; ulimit -m 850000; ulimit -s 8192; ulimit -n 512;  cd /tmp/M2-3540425-0/3-rundir/; GC_MAXIMUM_HEAP_SIZE=400M "/usr/bin/M2-binary" --int --no-randomize --no-readline --silent --stop --print-width 129 -e 'needsPackage("Bar", Reload => true, FileName => "/home/profzoom/tmp/bar.m2")' <"/tmp/M2-3540425-0/2.m2" >>"/tmp/M2-3540425-0/2.tmp" 2>&1
/tmp/M2-3540425-0/2.tmp:0:1: (output file) error: Macaulay2 exited with status code 1
/tmp/M2-3540425-0/2.m2:0:1: (input file)
M2: *** Error 1
stdio:2:1:(3): error: test #1 of package Bar failed:

-- -*- M2-comint -*- hash: 1661191464

i1 : --/home/profzoom/tmp/bar.m2:4: location of test code
     assert(2 + 2 == 5)
stdio:2:1:(3): error: assertion failed
```
It works as expected.  But let's try taking another look at just that one test that failed:

```m2
i3 : check(1, "Bar", Verbose => true)
-* running test 1 of package Bar in file:
   ../tmp/bar.m2:4:1:
   rerun with: check_1 "Bar" *-

[...]

PR #1688, opened 2020-12-12
Find location of failing test output correctly when running a single test

d-torrance:
`temporaryFilenameCounter` is incremented twice every time we run a
test, so we use this fact to compute the location of the temporary
file containing the output of failing tests when the `Verbose` option
to `check` is `true`.

However, this was done incorrectly when a single test was checked.  In
particular, the number of the test was being used to compute the
location of the temporary file, but this number is meaningless in this
case.  Instead, we always want to subtract 2 from the current value of
`temporaryFilenameCounter`.

Closes: #1686


mahrud:
Could you also document [check, Verbose]?

d-torrance:
> Could you also document [check, Verbose]?

Done!  Thanks for catching this.  I documented it in the original PR but forgot to add the key!

PR #1699, opened 2020-12-15
Add gftables path to FILE_PREREQS if found

d-torrance:
The 1.16.99 Ubuntu package was apparently built with the singular-data
package installed, as it points to the singular copy of these tables
instead of shipping its own.  Indeed:

```m2
Macaulay2, version 1.16.99
               Saturation, TangentCone

i1 : currentLayout#"factory gftables"

o1 = /usr/share/singular/factory/
```

However, the singular-data package is not listed as a dependency,
potentially resulting errors when running Macaulay2 after installing
the package if singular-data is not already present.  For example, see
https://groups.google.com/g/macaulay2/c/1Rb7fYsJMME/m/OVhjRvMZAwAJ.

We add the gftables path to FILE_PREREQS so that singular-data will be
included as a dependency to future builds of the package.


DanGrayson:
Thanks!

PR #1700, opened 2020-12-15
Merge 'release-1.16' branch

d-torrance:
The `release-1.16` branch has one extra commit (94c4b7d, updating some Debian packages in `M2/INSTALL`) that was never merged back into `master`.  I'd like to update  my `upstream` branch in the [Debian package repository](https://salsa.debian.org/science-team/macaulay2), which currently tracks `release-1.16`, to track `release-1.17`.  But I can't fast-forward and I'd prefer not to force push.

The commit in question actually conflicts with 92d795b, which similarly updated the list of Debian packages to install.  These conflicts have been resolved so that the branch can be merged back in.


DanGrayson:
Thanks for noticing!

PR #1704, opened 2020-12-16
Simplify GIT_DESCRIPTION

mahrud:
The value of `GIT_DESCRIPTION` will be computed as `version-X-Y-Z[-dirty]` where:
- `X = ` content of `VERSION` file
- `Y = ` number of commits since `VERSION` file was changed
- `Z = ` short commit hash of `HEAD`
- `-dirty` is added if any files tracked by git have been modified

For example: `version-1.16.99-51-418b9a61c`. If git is not found, gives `version-X-unknown`.

Also removes redundant factory patch file.


DanGrayson:
Thanks!

Is that pesky "g" going to be there?  (I tested the code here, and it seems to be, but I'm mystified.)

mahrud:
> Is that pesky "g" going to be there? (I tested the code here, and it seems to be, but I'm mystified.)

I don't see it anymore, did you do `make reconfigure`?

mahrud:
@DanGrayson https://faculty.math.illinois.edu is timing out.

DanGrayson:
> @DanGrayson https://faculty.math.illinois.edu is timing out.

Yeah, that's strange.  When it stops doing that I'll restart the jobs.

mahrud:
> > @DanGrayson https://faculty.math.illinois.edu is timing out.
> 
> Yeah, that's strange.  When it stops doing that I'll restart the jobs.

No need to restart, it's just that one build.

PR #1713, opened 2020-12-18
Fixes in MLdegree. Add output in scoreEquations with documentation.

roserhp:
@DanGrayson, I apologize in advance, but we've come accross with an issue in GraphicalModelsMLE. The way random data was generated in MLdegree gave an incorrect answer in more occasions than expected. This is fixed in this pull request. The changes are very local, it shouldn't cause any problems in the merge. Can we still make it for the release?


DanGrayson:
Yes, you're in time for 1.17.  Bug fixes are welcome!

PR #1718, opened 2020-12-19
GraphicalModelsMLE

roserhp:
Hi @DanGrayson, this is a follow-up from yesterday's pull request:
https://github.com/Macaulay2/M2/pull/1713
We realized that we forgot to set the debugging mode to false. This is the only change in this pull request, maybe it is easier to fix directly in release-1.17. Sorry about that.

### DIFF
--- a/M2/Macaulay2/packages/GraphicalModelsMLE.m2
+++ b/M2/Macaulay2/packages/GraphicalModelsMLE.m2
@@ -43,7 +43,7 @@ newPackage(
 	  },
      Headline => "maximum likelihood estimates for graphical statistical models",
      --Keywords => {"Algebraic Statistics"},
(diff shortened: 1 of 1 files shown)


M2/Macaulay2/packages/GraphicalModelsMLE.m2
```diff
@@ -42,7 +42,7 @@ newPackage(
 	   HomePage=>"http://faculty.fordham.edu/dswinarski"}
 	  },
      Headline => "maximum likelihood estimates for graphical statistical models",
-     Keywords => {"Algebraic Statistics"},
+     --Keywords => {"Algebraic Statistics"},
```
DanGrayson, line 45: Do you really want to comment out the keywords!

Issue #1639, opened 2020-12-01
Github actions 6hr time limit


mahrud:
The Github Actions that use autotools and run `make check` are starting to run into Github's 6hr time limit (e.g. [Complexes](https://github.com/Macaulay2/M2/actions/runs/392154407)). There are a few solutions:
- make autotools compile faster (it takes ~1hr longer just to build libraries, Macaulay2, and install packages than CMake)
- [x] don't run the library checks with autotools (really wasteful to run all tests for almost all libraries twice on every push, more than those libraries get tested on their own!)
- run package tests with CMake instead, which can run them in parallel
- finish and merge #1546

All are good solutions, we should probably do all, but we should do at least one very soon.

DanGrayson:
Also check whether we are installing all the Ubuntu packages we could be installing.

DanGrayson:
Good points, all of them.  I've just implemented one of them:
```diff
commit 1e47260755100936bafd60d64013672dedd3941f (HEAD -> development, upstream/development)
Author: Daniel R. Grayson <dan@math.uiuc.edu>
Date:   Mon Nov 30 19:49:40 2020 -0600

    omit checking the libraries

diff --git a/.github/workflows/test_build.yml b/.github/workflows/test_build.yml
index 0f851c497..aae8c12c5 100644
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -182,7 +182,7 @@ jobs:
       - name: Run Tests using Autotools
         if: matrix.build-system == 'autotools' && matrix.os == 'ubuntu-latest'
         run: |
-          make check
+          make check -o check-in-libraries
           make -C Macaulay2/html-check-links check
           make validate-html
 
```

mahrud:
> Also check whether we are installing all the Ubuntu packages we could be installing.

Yes, I think CMake is using all of them + flint from Linuxbew (because the new version isn't on Ubuntu yet), and #1412 is about allowing the same with Autotools. Flint is the biggest bottleneck for Autotools, because factory also depends on it.

DanGrayson:
I just checked: flint is still at 2.5.2 on both Ubuntu 18.04 and 20.04 (to which github will soon switch).  But we should be getting factory from `singular-factory`.

d-torrance:
[...]

PR #1722, opened 2020-12-21
build and upload tarfile from actions

mahrud:


DanGrayson:
I wonder how annoyed with us github will be when we starting leaving such gigantic artifacts lying around all over.

mahrud:
> I wonder how annoyed with us github will be when we starting leaving such gigantic artifacts lying around all over.

1. it's not so heavy, here is on my machine:
> Finished uploading artifact artifact. Reported size is 39536572 bytes. There were 0 items that failed to upload

2. they get discarded after 90 days by default, but I can also reduce it by `retention-days: 5`.
3. compute costs are a much more significant cost for github than storage costs, mainly because storage doesn't produce heat

DanGrayson:
40 mb is pretty heavy, especially when we're making 6-10 of those per day

DanGrayson:
What purpose do these tarfiles serve?

mahrud:
> 40 mb is pretty heavy, especially when we're making 6-10 of those per day

Each github build is downloading and uploading hundreds of megabytes. Again, storage is nearly free! A public repository can be up to 100GB and each file up to 2GB. 40MB is really nothing.

> What purpose do these tarfiles serve?

I want to download them and make sure the documentation is generated okay.

DanGrayson:
> > 40 mb is pretty heavy, especially when we're making 6-10 of those per day
> 
> Each github build is downloading and uploading hundreds of megabytes. Again, storage is nearly free! A public repository can be up to 100GB and each file up to 2GB. 40MB is really nothing.
> 
> > What purpose do these tarfiles serve?
> 
> I want to download them and make sure the documentation is generated okay.

Ah, that's worthwhile.

mahrud:
Should I also omit the html validation check, as mentioned in #1639?

DanGrayson:
> Should I also omit the html validation check, as mentioned in #1639?

No, that's important to check.

mahrud:
> No, that's important to check.

Sure, but why not just do that on the nightly builds that you run? Certainly not every commit pushed warrants an hour of checking thousands of links across Macaulay2Doc!

Also, are you checking your slack messages?

DanGrayson:
> > No, that's important to check.
> 
> Sure, but why not just do that on the nightly builds that you run? Certainly not every commit pushed warrants an hour of checking thousands of links across Macaulay2Doc!
> 

Well, that's a good idea.  Weekly should be enough.  We could also check various compilers weekly.

> Also, are you checking your slack messages?

The slack notification on my laptop pops up for 1 second and then disappears.  After that I don't know how to find it.  Just send me email instead.

mahrud:
> Well, that's a good idea. Weekly should be enough. We could also check various compilers weekly.

So should I remove it from the github builds and have you do it once a week, or setup a system to do it once a week on github?

If the latter, can you show me how to excluded a target from a subdirectory? Does `-o run-html-check` work even if that target is inside `Macaulay2/html-check-links`?

DanGrayson:
> > Well, that's a good idea. Weekly should be enough. We could also check various compilers weekly.
> 
> So should I remove it from the github builds and have you do it once a week, or setup a system to do it once a week on github?
> 
> Eitherway, can you show me how to excluded a target from a subdirectory? Does `-o run-html-check` work even if that target is inside `Macaulay2/html-check-links`?

Checking the links is quick, I believe, so don't worry about that.  The (slow) validation step can be omitted with `-o validate-html` at top level.

mahrud:
Sorry, that was silly of me: I didn't see there was just a `make validate-html` that I needed to erase. I also updated the CPack scripts and changed it to create an ubuntu package. It should automatically detect dependencies, I believe, so it should be usable by developers at least.

PR #1730, opened 2020-12-23
Fix elapsedTime and elapsedTiming on 32-bit systems

d-torrance:
We introduce a new `longlong` (and, for good measure, `ulonglong`) data type to the D language which is guaranteed to hold a 64-bit integer.  Using this, we won't overflow while timing calculations on 32-bit architectures after just over 2 seconds!


DanGrayson:
Shouldn't we use `uint64_t` from `stdint.h` instead?

d-torrance:
> Shouldn't we use `uint64_t` from `stdint.h` instead?

Will do!

d-torrance:
I've pushed a new version of these commits, using `uint64_t` now.  I also went ahead and added `long long` and all the fix-width integer types to the D language, both signed and unsigned, in case we need any them in the future.

mahrud:
Maybe we should drop all data types like `long` that don't explicitly mention the bit length. Safer to see an error and realize it should be avoided!

mahrud:
Correction: we should probably keep `int` as the representative of the size that the CPU can most efficiently use, but even though `long long` is standard, it's much less informative compared to `int64_t`.

d-torrance:
> Correction: we should probably keep `int` as the representative of the size that the CPU can most efficiently use, but even though `long long` is standard, it's much less informative compared to `int64_t`.

It's actually `long` that varies by architecture.  The others should be the same regardless (`short` is 16 bits, `int` 32 bits, and `long long` 64 bits -- see https://wiki.debian.org/ArchitectureSpecificsMemo).

mahrud:
Ah, so maybe let's just remove `long` and `unsigned long` to avoid any other confusion.

d-torrance:
> Ah, so maybe let's just remove `long` and `unsigned long` to avoid any other confusion.

Sounds like a good idea -- but probably for another pull request.  There's likely some false positives, but it looks nontrivial:

```
profzoom@peg-amy:~/src/macaulay2/M2/M2/Macaulay2/d$ grep long * | wc -l
286
```

Issue #1728, opened 2020-12-23
Too many heap sections capturing check(8, "DiffAlg") on armhf


d-torrance:
From PPA builds on Ubuntu [18.04](https://launchpadlibrarian.net/512788986/buildlog_ubuntu-bionic-armhf.macaulay2_1.16.99+git202012211524-0ppa202012211755~ubuntu18.04.1_BUILDING.txt.gz) and [20.04](https://launchpadlibrarian.net/512800594/buildlog_ubuntu-focal-armhf.macaulay2_1.16.99+git202012211524-0ppa202012211755~ubuntu20.04.1_BUILDING.txt.gz) armf last night:

```m2
 -- capturing check(8, "DiffAlg") from source:
   DiffAlg.m2:1324:1:                                                       Too many heap sections: Increase MAXHINCR or MAX_HEAP_SECTS
Aborted (core dumped)
make[4]: *** [Makefile:39: check-DiffAlg] Error 134
make[4]: Leaving directory '/<<BUILDDIR>>/macaulay2-1.16.99+git202012211524/M2/Macaulay2/packages'
make[3]: *** [Makefile:14: check-in-packages] Error 2
make[3]: Leaving directory '/<<BUILDDIR>>/macaulay2-1.16.99+git202012211524/M2/Macaulay2'
make[2]: *** [GNUmakefile:229: check-in-Macaulay2] Error 2
make[2]: Leaving directory '/<<BUILDDIR>>/macaulay2-1.16.99+git202012211524/M2'
```

The PPA builds for Ubuntu 20.10 and 21.04 armhf finished without problems.

It's interesting that we didn't get to the `test #8 of package DiffAlg failed` message.  I suppose this is because of the switch to `capture`, and the parent `M2` aborted, not some child one that it called just for this test?

DanGrayson:
Yes, if the program crashes, then it doesn't get to capture the next test.

Another thing about capturing is that all the garbage memory from the previous tests is still there.

mahrud:
Interesting. Let me think about this a bit. If it's unavoidable, we can disable that for the build.

> Another thing about capturing is that all the garbage memory from the previous tests is still there.

I added garbage collection after capturing in #1716

mahrud:
@d-torrance does this happen on top of #1716 as well?

d-torrance:
> @d-torrance does this happen on top of #1716 as well?

I'll test it out and see.

mahrud:
[...]

PR #1732, opened 2020-12-23
Don't capture check(8, "DiffAlg")

d-torrance:
Band-aid for #1728


mahrud:
(I cancelled the tests since it's literally not a change from the perspective of github actions)

Issue #1736, opened 2020-12-24
Wishlist: A family of "assert" functions


d-torrance:
For helping to diagnose test failures from build logs, it would be nice if we could see *why* an `assert` statement threw an error.  Right now, `assert` just takes a boolean expression as input and throws an error if it's anything but `true`.  Instead, we could have a family of functions that know a little bit more about the input and could include that information in the error message, e.g., something like:

```m2
i1 : assertEqual(2 + 2, 5)
stdio:1:1:(3): error: assertion failed: 4 != 5
```

DanGrayson:
Can't you just look at the source code to see the statement that was false?

d-torrance:
> Can't you just look at the source code to see the statement that was false?

Sure, but that only tells part of the story story.  I got this idea looking at a PPA build failure on Ubuntu 21.04 arm64 from last night:

```m2
testing: timing-quotient.m2
timing-quotient.m2:223:1:(3):[4]: error: assertion failed
```

Here's the code:

```m2
standardSecond = first timing (value Core#"private dictionary"#"spin") 2833 -- about 1 second on a Macbook Pro in 2015

...

P=QQ[x,y,z,MonomialOrder=>Lex];
d=z^4+z^2*x*y^9+z*x^9*y+x^5*y^5;
phi=map(P,P,matrix{{x^13*y^4,x^3*y,x^20*y^6*z}});
tim = timing factor(phi(d));
assert(tim#0 < .05 * standardSecond)
```

From the source code alone, I have no idea what `tim` and `standardSecond` actually were for this build.

DanGrayson:
Maybe a better way to accomplish the same objective would be to change `assert` to a method function that converts expressions to values using `value`.  Then the code could say something like `assert BinaryOperation {symbol < , 4 , 3}` and the error message could display the expression, not its value.  You'd have to make `value BinaryOperation {symbol < , 4 , 3}` work, too, but that would be good, anyway.

mahrud:
@d-torrance sorry if this is too late, but another option is to have assert also accept a string, then feed it to `capture` for evaluation and check the result, then if it fails it can actually print the string! Something like this:
```m2
assert' = method()
assert' String := s -> if first capture concatenate("assert(", s, ")") then error s
```
This is what it would look like:
```m2
i3 : assert'("2 == 2")

i4 : assert'("2 != 2")
currentString:1:1:(3):[5]: error: assertion failed
[...]

PR #1738, opened 2020-12-24
"assert" improvements

d-torrance:
We convert `assert` into a method and add `(assert, BinaryOperation)` with various wrappers for more informative error messages.  For example:

```m2
i1 : assertEqual(2 + 2, 5)
stdio:1:1:(3): error: assertion failed:
4==5 is false

i2 : assertGreaterThan(rank matrix{{1, 2}, {2, 4}}, 2)
stdio:2:1:(3): error: assertion failed:
1>2 is false
```

Still to do:

- [x] Documentation
- [x] Unit tests
- [x] Convert some existing tests as a proof of concept

Closes: #1736

### DIFF
--- a/M2/Macaulay2/m2/basictests/W02-threads.m2
+++ b/M2/Macaulay2/m2/basictests/W02-threads.m2
@@ -1,3 +1,5 @@
+assert = x -> if not x then error "assertion failed"
+
 assert( getGlobalSymbol "stopIfError" === symbol stopIfError )
 threadVariable foo
 assert( getGlobalSymbol "foo" === symbol foo )
--- a/M2/Macaulay2/m2/expressions.m2
+++ b/M2/Macaulay2/m2/expressions.m2
@@ -64,6 +64,12 @@ expressionValue Symbol := value
 
 value Expression := expressionValue
 
+assert Expression := v -> (
+    val := value v;
+    if not instance(val, Boolean) then error "assert: expected true or false";
+    if not val then error toString("assertion failed:" || net v | " is false")
+)
+
 --Holder2 = new WrapperType of Expression			    -- Holder{ printable form, value form }
 --Holder2.synonym = "holder"
 --Holder = new WrapperType of Holder2			    -- Holder{ printable form, value form }, with printable form === value form
--- a/M2/Macaulay2/m2/expressions.m2
+++ b/M2/Macaulay2/m2/expressions.m2
@@ -470,7 +470,13 @@ binaryOperatorFunctions := new HashTable from {
      symbol != => ((x,y) -> x != y),
      symbol and => ((x,y) -> x and y),
      symbol or => ((x,y) -> x or y),
-     symbol ^** => ((x,y) -> x^**y)
+     symbol ^** => ((x,y) -> x^**y),
+     symbol === => ((x,y) -> x === y),
+     symbol =!= => ((x,y) -> x =!= y),
+     symbol < => ((x,y) -> x < y),
+     symbol <= => ((x,y) -> x <= y),
+     symbol > => ((x,y) -> x > y),
+     symbol >= => ((x,y) -> x >= y)
      }
 
 expressionBinaryOperators =
--- a/M2/Macaulay2/m2/methods.m2
+++ b/M2/Macaulay2/m2/methods.m2
@@ -1,9 +1,9 @@
 --		Copyright 1993-2002 by Daniel R. Grayson
 
 -- temporary definitions to get error messages to work before methods are working, so we can debug methods
-assert( class between === Symbol )
+assert'( class between === Symbol )
 between = (m,v) -> mingle(v,#v-1:m)			    -- provisional
-assert( class toString === Symbol )
+assert'( class toString === Symbol )
 toString = x -> (					    -- provisional
      if hasAttribute(x,ReverseDictionary) then simpleToString getAttribute(x,ReverseDictionary)
      else if class x === Net then concatenate between("\n",unstack x)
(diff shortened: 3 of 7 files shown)


DanGrayson:
I see no reason to introduce scads of new functions, such as `assertEqual`, which you can write something like `assert (hold 4 == 5)` or `assert (hold 4 > 5)`.

mahrud:
> I see no reason to introduce scads of new functions, such as `assertEqual`, when you can write something like `assert (hold 4 == 5)` or `assert (hold 4 > 5)`.

The latter doesn't work:
```m2
i6 : hold 4 > 5
stdio:6:8:(3): error: no method for binary operator ? applied to objects:
--            4 (of class Holder)
--      ?     5 (of class ZZ)
```
How does `hold` know where the expression stops?

d-torrance:
> I see no reason to introduce scads of new functions, such as `assertEqual`, when you can write something like `assert (hold 4 == 5)` or `assert (hold 4 > 5)`.

Functions like these are pretty common in other platforms, e.g., in R:

``` r
> library(testthat)
> expect_equal(2 + 2, 5)
Error: 2 + 2 not equal to 5.
1/1 mismatches
[1] 4 - 5 == -1
```
and Python:
```python
>>> import unittest
>>> class MyTest(unittest.TestCase):
	def test_foo(self):
		self.assertEqual(2 + 2, 5)

		
>>> unittest.main()
F
======================================================================
FAIL: test_foo (__main__.MyTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "<pyshell#4>", line 3, in test_foo
AssertionError: 4 != 5

----------------------------------------------------------------------
Ran 1 test in 0.002s

FAILED (failures=1)
```

I figured it would be nice for readability.  I'd rather type `assertEqual(2 + 2, 5)` than `assert(symbol ==, 2 + 2, 5)` or something similarly strange-looking.  But I'm happy to remove them if you'd rather not have them!

mahrud:
I agree with Doug, but if that `assert hold` can work, it would be amazing! I don't know of any other languages that can evaluate "4 == 5" not just as true, but parse it into an expression and manipulate it.

ps: I wonder if it would be beneficial to extract testing related functions into a `Testing` package, with more functionality.

mikestillman:
Actually, I believe that the c++ unit testing library catch2 does that (parses input in a call).

DanGrayson:
> > I see no reason to introduce scads of new functions, such as `assertEqual`, when you can write something like `assert (hold 4 == 5)` or `assert (hold 4 > 5)`.
> 
> The latter doesn't work:
> 
> ```m2
> i6 : hold 4 > 5
> stdio:6:8:(3): error: no method for binary operator ? applied to objects:
> --            4 (of class Holder)
> --      ?     5 (of class ZZ)
> ```
> 

Right, we'd have to make it work.  Actually, we'd have to install method lookup in the d directory first, because the code for `x>y` always just runs `x?y` to see what the answer is.  If we did that, then this  would work:
```m2
Thing > Expression := Expression > Thing := (x,y) -> BinaryOperation {symbol >, x, y}
```

> How does `hold` know where the expression stops?

In `hold x == y`, we're comparing `hold x` to `y`.  But there is a method for doing that:
```m2

i5 : code lookup (symbol ==, Expression, Thing)

o5 = ../../../../../../Macaulay2/m2/expressions.m2:490:45-490:64: --source code:
         installMethod(op,Expression,Thing,(x,y) ->  g(x,expression y));
     | symbol   class              value location of symbol
     | ------   -----              ----- ------------------                                         
     | f      : FunctionClosure -- ...   ../../../../../../Macaulay2/m2/expressions.m2:485:5-485:6  
     | g      : FunctionClosure -- ...   ../../../../../../Macaulay2/m2/expressions.m2:489:5-489:6  
     | op     : Keyword         -- ==    ../../../../../../Macaulay2/m2/expressions.m2:484:33-484:35
```

mikestillman:
Having both assert and assertEqual sounds good to me (my 2 cents)

DanGrayson:
> Having both assert and assertEqual sounds good to me (my 2 cents)

I think that's name space bloat.  Especially when we don't need it, if we make `assert (hold x == hold y)` work.

d-torrance:
Perhaps `assertEqual` and friends could go into a new `Testing` package, as Mahrud suggested?

DanGrayson:
> > Having both assert and assertEqual sounds good to me (my 2 cents)
> 
> I think that's name space bloat. Especially when we don't need it, if we make `assert (hold x == hold y)` work.

And by name bloat, consider how it will look after you have added `assertEqual`, `assertUnequal`, `assertLessThan`, `assertLessThanOrEqual`, `assertGreater`, `assertGreaterThanOrEqual`, `assertNot`, `assertIsSubset`, `assertIsHomogeneous`, `assertIsNonsingular`, etc., etc.  All of that is already taken care of by "expressions".

DanGrayson:
No reason for it.  The slight improvements I've suggested are enough, and moving the namespace bloat to a package still counts as namespace bloat.

d-torrance:
Just pushed a new version fixing the failing test and reverting the addition of the new functions.  I've also generalized `(assert, BinaryOperation)` to `(assert, Expression)`, so any expression that evaluates to a boolean should work.

The following works, which is nice:

```m2
i1 : assert Equation(2 + 2, 5)
stdio:1:1:(3): error: assertion failed:
4 == 5 is false
```

d-torrance:
I think this PR is ready.  The basic functionality of `(assert, Expression)` works, and `assert Equation(x, y)` is close enough to `assertEqual(x, y)` that I'm pretty happy on that front.

I'll save `Expression < Thing` and friends for later.

PR #1741, opened 2020-12-26
Load test directory before determining which tests to run

d-torrance:
Otherwise, we won't run any tests in the directory until "check" is
called a second time.

We also drop the check for "Core" -- it's the only package that has a
test directory, so it was redundant.


d-torrance:
I moved this commit on top of #1750 since they deal with essentially the same thing.

PR #1755, opened 2020-12-29
Add "no-check-flag" comment for tests

d-torrance:
I recently added this feature to the draft of the Debian package to simplify skipping all the package tests that have been failing on various architectures (see [this patch](https://salsa.debian.org/science-team/macaulay2/-/blob/debian/development/debian/patches/skip-failing-package-tests.patch)).  Would there be interest in having it upstream as well?  Note that this PR only enables the feature, and doesn't actually skip any tests.

Here's an example of what it looks like when a test is skipped:

```m2
 -- capturing check(10, "DeterminantalRepresentations")                      -- 1.46659 seconds elapsed
 -- capturing check(11, "DeterminantalRepresentations")                      -- 0.0888114 seconds elapsed
 -- skipping check(12, "DeterminantalRepresentations")                       -- 0.000079951 seconds elapsed
 -- capturing check(13, "DeterminantalRepresentations")                      -- 0.0776138 seconds elapsed
```

### DIFF
--- a/M2/Macaulay2/m2/testing.m2
+++ b/M2/Macaulay2/m2/testing.m2
@@ -29,18 +29,21 @@ TEST(String, String) := (title, teststring) -> (
 -- check
 -----------------------------------------------------------------------------
 
+checkmsg := (verb, desc) ->
+    stderr << commentize pad(pad(verb, 10) | desc, 72) << flush;
+
 captureTestResult := (desc, teststring, pkg, usermode) -> (
     stdio << flush; -- just in case previous timing information hasn't been flushed yet
     if match("no-check-flag", teststring) then (
-	stderr << commentize pad("skipping " | desc, 72) << flush;
+	checkmsg("skipping", desc);
 	return true);
     -- try capturing in the same process
     if isCapturable(teststring, pkg, true) then (
-	stderr << commentize pad("capturing " | desc, 72) << flush;
+	checkmsg("capturing", desc);
 	(err, output) := capture(teststring, PackageExports => pkg, UserMode => usermode);
 	if err then printerr "capture failed; retrying ..." else return true);
     -- fallback to using an external process
-    stderr << commentize pad("running " | desc, 72) << flush;
+    checkmsg("running", desc);
     runString(teststring, pkg, usermode))
 
 loadTestDir := pkg -> (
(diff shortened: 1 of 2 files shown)


DanGrayson:
Well, that's better than just commenting the test out.  It will always be skipped, right?

d-torrance:
> Well, that's better than just commenting the test out. It will always be skipped, right?

Correct

mahrud:
I disliked adding `no-capture-flag`, and hope to eventually remove it from everywhere in the code, and feel bad that it opened the door to other flags like this.

I see the point of why you need it, and this is probably an okay solution for now, but I suggest a better solution for future: instead of storing tests as just a string, store them as a list or hashtable with more fields, one of which is the test itself, and the others can be information like a description of what it tests or whether it's a fast test or slow test. Then, add an option to check to limit which tests to run, and for purposes like testing on Debian builds, only run the normal tests and skip the slow or "64bit only" tests.

Also, add an extra space after "Skipping" just so all the "check(..."s line up.

d-torrance:
> I disliked adding `no-capture-flag`, and hope to eventually remove it from everywhere in the code, and feel bad that it opened the door to other flags like this.

I totally agree.  Both comments are super hacky, and I'd be content with leaving this one in the Debian package until a better solution comes along.  But I figured I'd see if there was interest.

> I see the point of why you need it, and this is probably an okay solution for now, but I suggest a better solution for future: instead of storing tests as just a string, store them as a list or hashtable with more fields, one of which is the test itself, and the others can be information like a description of what it tests or whether it's a fast test or slow test. Then, add an option to check to limit which tests to run, and for purposes like testing on Debian builds, only run the normal tests and skip the slow or "64bit only" tests.

I've had similar ideas -- this is definitely something to work towards!
 
> Also, add an extra space after "Skipping" just so all the "check(..."s line up.

`running` has the same problem.  I pushed a new commit that adds a helper function to print all three messages so they line up nicely.  For example:

```m2
 -- capturing check(0, "AlgebraicSplines")                                   -- 0.0181995 seconds elapsed
 -- skipping  check(1, "AlgebraicSplines")                                   -- 0.00004813 seconds elapsed
 -- running   check(2, "AlgebraicSplines")                                   -- 1.54114 seconds elapsed
 -- capturing check(3, "AlgebraicSplines")                                   -- 0.0424775 seconds elapsed
```

Issue #1604, opened 2020-11-21
missing methods


mahrud:
Here are various things that are incomplete:
```m2
i2 : 0 < #methods(resolution, Ideal)

o2 = false

i9 : #methods symbol= === #methods sequence symbol=

o9 = false

i11 : code methods (symbol!, symbol=)

o11 = -- no method function found: = !
```
There are more. I'll fix these in #1385

DanGrayson:
You probably meant `0 < # methods(resolution, Ideal)`

mahrud:
Yes. Also, looking at the code for methods, I don't understand the point of these lines:
- This one implies there are types `T` with a hash table entry `K => f` where `K` is a keyword and `f` is a function. Is there any such an example? I tried adding such a key to a type, but it wasn't callable.
https://github.com/Macaulay2/M2/blob/8219a0a0a89cb4d626f1d6776bb4c9164feb126b/M2/Macaulay2/m2/code.m2#L109
- This line implies the existence of types `T1` with a hash table entry `T2 => f` where `T2` is also a type. Is there such an example? I also couldn't construct anything like this.
https://github.com/Macaulay2/M2/blob/8219a0a0a89cb4d626f1d6776bb4c9164feb126b/M2/Macaulay2/m2/code.m2#L120

DanGrayson:
The code is strange, because `instance(key, Symbol) and instance(key,Keyword)` is equivalent to `instance(key,Keyword)`.

I managed it this way:
```m2

i2 : R = QQ[x]

o2 = R

o2 : PolynomialRing

i3 : R#symbol ! = p -> p^3;

i4 : x!

      3
o4 = x

o4 : R
```

Perhaps `if key === F then found#(F,x) = true ` is not needed and could go.

mahrud:
Well, I changed and simplified it quite a bit. I think everything that was returned before is still returned, and a bunch of things that were returned incorrectly (e.g `code` didn't work on them) are now corrected.

Try it out and let me know if anything is missing.

mahrud:
Were you able to create anything of the second type? Like `T1.T2`?

DanGrayson:
I think you mean `T1#T2` -- in any case, no, I think it's not a thing.

PR #1758, opened 2020-12-30
Properly list methods from a package

mahrud:
The fix was nearly trivial thanks to my changes in #1604.


M2/Macaulay2/m2/format.m2
```diff
@@ -62,7 +62,7 @@ info TITLE :=
 net  COMMENT :=
 info COMMENT :=
 net  LITERAL :=
-info LITERAL := x -> ""
+info LITERAL := x -> ()
```
DanGrayson, line 65: `info` is to return a net, so this can't be right.

mahrud: > info is to return a net, so this can't be right.

The typical value of `info` is String, perhaps that should be fixed. When combined in any other hypertext object, it still returns a Net:
```m2
i9 : info DIV{"a", LITERAL "b", "c"}

o9 = ac

i10 : class oo

o10 = Net

o10 : Type

i11 : peek ooo

      +--+
o11 = |ac|
      +--+
```

Here is the reason for the change:
old behavior:
```m2
i1 : R = kk[x,y];

i2 : info help R

o2 = R 
     **



     For the programmer
     ==================

     The object R (missing documentation) is a *note polynomial ring: (Macaulay2Doc)PolynomialRing,, with ancestor classes *note
     RingElement: (Macaulay2Doc)RingElement, < *note BasicList: (Macaulay2Doc)BasicList, < *note Thing: (Macaulay2Doc)Thing,.

i3 : net help R

o3 = R 
     **



     For the programmer
     ==================

     The object "R" is a polynomial ring, with ancestor classes "RingElement" < "BasicList" < "Thing".
```
New behavior:
```m2
i1 : R = kk[x,y];

i2 : info help R

o2 = R 
     **

     For the programmer
     ==================

     The object R (missing documentation) is a *note polynomial ring: (Macaulay2Doc)PolynomialRing,, with ancestor classes *note
     RingElement: (Macaulay2Doc)RingElement, < *note BasicList: (Macaulay2Doc)BasicList, < *note Thing: (Macaulay2Doc)Thing,.

i3 : net help R

o3 = R 
     **

     For the programmer
     ==================

     The object "R" is a polynomial ring, with ancestor classes "RingElement" < "BasicList" < "Thing".
```
Notice the extra line not being there.

DanGrayson:
Yes, the typical value should be fixed.  It would be better to return an empty net of height 0 than an empty sequence, because the sequence probably requires special handling.

mahrud:
> Yes, the typical value should be fixed. It would be better to return an empty net of height 0 than an empty sequence, because the sequence probably requires special handling.

Okay, I'll fix those in a second.

mahrud:
> It would be better to return an empty net of height 0 than an empty sequence, because the sequence probably requires special handling.

I don't know how to return a net of height 0 that would fix this. Do you have a suggestion?

mahrud:
Nevermind, figured out a different way.

d-torrance:
> > It would be better to return an empty net of height 0 than an empty sequence, because the sequence probably requires special handling.
> 
> I don't know how to return a net of height 0 that would fix this. Do you have a suggestion?

`horizontalJoin()` does this

DanGrayson:
> > It would be better to return an empty net of height 0 than an empty sequence, because the sequence probably requires special handling.
> 
> I don't know how to return a net of height 0 that would fix this. Do you have a suggestion?

```m2

i1 : horizontalJoin ()

o1 = 

i2 : height oo

o2 = 0
```

mahrud:
> horizontalJoin() does this

I'm aware. This doesn't solve the issue:
```m2
i11 : info DIV{"A", DIV{COMMENT()}, "B"}

o11 = A

      B
```
Regardless, it is fixed a better way.

mahrud:
> horizontalJoin() does this

I'm aware. This doesn't solve the issue:
```m2
i11 : info DIV{"A", DIV{COMMENT horizontalJoin()}, "B"}

o11 = A

      B
```
Regardless, it is fixed a better way.

PR #1760, opened 2020-12-30
add preliminary version of the package MultiprojectiveVarieties

giovannistagliano:


giovannistagliano:
Thanks for merging. I'll just do one last update today.

PR #1762, opened 2020-12-30
concatenate/commentize/pad updates

d-torrance:
After initially agreeing to dropping the filenames and line numbers of tests during `check` recently, I've found myself missing them when, for example, we get a segfault during `capture` and never get to `check`'s message at the end with information about the failing tests.  It's possible to do some sleuthing to figure out the location of the tests, but it would be nice if we had that information in the log, at least when check's `Verbose` option is set to `true`.

This PR accomplishes this.  For example:
```m2
i1 : check(0, "Core")
 -- capturing check(0, "Core")                                               -- 0.0316604 seconds elapsed

i2 : check(0, "Core", Verbose => true)
 -* capturing check(0, "Core") in file:                                     
    /usr/share/doc/Macaulay2/Core/tests/0-homog.m2:0:1:                      *- -- 0.0315109 seconds elapsed
```

Although the actual change to `check` itself was just a couple of lines, I ended up needing to do a fair amount of work changing `pad` and `commentize` to play nicely with multiline strings and nets so we wouldn't end up with a `--` in front of the filename and preventing us from jumping to it quickly in Emacs.

In particular:

## `pad`
Now works with multiline strings and nets in (what I would consider) the expected way:

*before*
```m2
i1 : pad(4, "foo\nbar")

o1 = foo
     bar
```
Nothing happened since the string was longer than 4 characters, regardless of the fact that each line only has 3.

*after*
```m2
i1 : pad(4, "foo\nbar")

o1 =  foo
      bar
```
## `commentize`
Now uses `-* ... *-` comments for multiline strngs and nets, and also preserves the baseline of nets:

*before*
```m2
i4 : commentize "foo\nbar"

o4 =  -- foo
      -- bar

i5 : commentize "foo"^1

o5 =  -- foo
```

*after*
```m2
i4 : commentize "foo\nbar"

o4 =  -* foo
         bar *-

i5 : commentize "foo"^1

      -- foo
o5 = 
```

## `concatenate`
Now accepts negative integers (returning the empty string), which is consistent with its current behavior when given a list containing a negative integer.  I needed this for an early draft of the `pad` updates that I ended up changing, but it seemed like a good idea, so I kept it in.

*before*
```m2
i6 : concatenate {-1}

o6 = 

i7 : concatenate(-1)
stdio:7:1:(3): error: encountered negative integer
```

*after*
```m2
i1 : concatenate(-1)

o1 = 
```

### DIFF
--- a/M2/Macaulay2/d/actors4.d
+++ b/M2/Macaulay2/d/actors4.d
@@ -615,10 +615,8 @@ stringcatfun(e:Expr):Expr := (
      is y:SymbolClosure do toExpr(y.symbol.word.name)
      is n:ZZcell do (
 	  if isInt(n) then (
-	       m := toInt(n);
-	       if m >= 0
-	       then toExpr(new string len m do provide ' ')
-	       else buildErrorPacket("encountered negative integer")
+	       m := max(toInt(n), 0);
+	       toExpr(new string len m do provide ' ')
 	       )
 	  else buildErrorPacket("encountered a large integer")
 	  )
--- a/M2/Macaulay2/m2/nets.m2
+++ b/M2/Macaulay2/m2/nets.m2
@@ -262,7 +262,9 @@ commentize = method(Dispatch => Thing)
 commentize Nothing   := s -> ""
 commentize BasicList := s -> commentize horizontalJoin s
 commentize String    := s -> concatenate(" -- ", between("\n -- ", separate s))
-commentize Net       := S -> stack(commentize \ unstack S)
+commentize Net       := S -> (
+    baseline := height S - if height S == -depth S then 0 else 1;
+    (stack(commentize \ unstack S))^baseline)
 
 printerr = msg -> (stderr << commentize msg << endl;) -- always return null
 warning  = msg -> if debugLevel > 0 then (
--- a/M2/Macaulay2/m2/nets.m2
+++ b/M2/Macaulay2/m2/nets.m2
@@ -66,7 +66,9 @@ toString Manipulator := f -> (
 toExternalString String := format
 
 toString Net := x -> demark("\n",unstack x)
-toExternalString Net := x -> concatenate(format toString x, "^", toString(height x - 1))
+toExternalString Net := x -> if height x + depth x == 0 then
+     concatenate("(horizontalJoin())", "^", toString height x) else
+     concatenate(format toString x, "^", toString(height x - 1))
 
 toExternalString MutableHashTable := s -> (
      if hasAttribute(s,ReverseDictionary) then return toString getAttribute(s,ReverseDictionary);
--- a/M2/Macaulay2/m2/printing.m2
+++ b/M2/Macaulay2/m2/printing.m2
@@ -4,6 +4,8 @@ pad = method()
 
 pad(String,ZZ) := String => (s,n) -> concatenate(s,n-# s)
 pad(ZZ,String) := String => (n,s) -> concatenate(n-# s,s)
+pad(Net, ZZ) := (S, n) -> S | (concatenate(n - width S))^(-depth S)
+pad(ZZ, Net) := (n, S) -> (concatenate(n - width S))^(height S - 1) | S
 
 columnate = (wid,items) -> (
      lens := apply(items,i -> #i);
--- a/M2/Macaulay2/packages/Macaulay2Doc/doc12.m2
+++ b/M2/Macaulay2/packages/Macaulay2Doc/doc12.m2
@@ -550,16 +550,32 @@ document {
      	  }
      }
 
-document {
-     Key => {pad,(pad, String, ZZ),(pad, ZZ, String)},
-     Headline => "pad a string with spaces",
-     TT "pad(s,n)", " -- pads the string ", TT "s", " to length ", TT "n", " with
-     spaces on the right.",
-     BR{},
- 
-     TT "pad(n,s)", " -- pads the string ", TT "s", " to length ", TT "n", " with
-     spaces on the left."
-     }
+doc ///
+  Key
+    pad
+    (pad, String, ZZ)
+    (pad, ZZ, String)
+    (pad, Net, ZZ)
+    (pad, ZZ, Net)
+  Headline
+    pad a string or net with spaces
+  Usage
+    pad(s,n)
+    pad(n,s)
+  Inputs
+    s:Net
+    n:ZZ
+  Description
+    Text
+      @TT "pad(s,n)"@ pads the string or net @TT "s"@ to length @TT
+      "n"@ with spaces on the right.
+
+      @TT "pad(n,s)"@ pads the string or net @TT "s"@ to length @TT
+      "n"@ with spaces on the left.
+    Example
+      pad(6, "foo")
+      pad("foo", 6) | "bar"
+///
 
 document {
      Key => columnate,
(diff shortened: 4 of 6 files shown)


mahrud:
Please don't change commentize.

mahrud:
Also, the filenames were partially removed because they appeared as errors in Emacs compile-mode (see ##1719). I removed it knowing that errors, we still get the test location in the error with `Verbose => true`:

```m2
i1 : check("Doo", Verbose => true)
 -- capturing check(0, "Doo")                                                -- capture failed; retrying ...
 -- running check(0, "Doo")                                                 
 ulimit -c unlimited; ulimit -t 700; ulimit -m 850000; ulimit -s 8192; ulimit -n 512;  cd /tmp/M2-21-0/1-rundir/; GC_MAXIMUM_HEAP_SIZE=400M "/usr/local/bin/M2-binary" --int --no-randomize --no-readline --silent --stop --print-width 77 -e 'needsPackage("Doo",Reload=>true,FileName=>"/home/macaulay/Doo.m2")' <"/tmp/M2-21-0/0.m2" >>"/tmp/M2-21-0/0.tmp" 2>&1
/tmp/M2-21-0/0.tmp:0:1: (output file) error: Macaulay2 exited with status code 1
/tmp/M2-21-0/0.m2:0:1: (input file)
M2: *** Error 1
 -- 0.578492 seconds elapsed
Doo.m2:7:1: error:
 -- -- -*- M2-comint -*- hash: 1169441718
 -- 
 -- i1 : --/home/macaulay/Doo.m2:8: location of test code
 --      assert false
 -- stdio:2:1:(3): error: assertion failed
 -- 
stdio:1:1:(3): error: test(s) #0 of package Doo failed.
```

See the line "Doo.m2:7:1: error:", which is also detected by compile-mode as an error so you can jump to it.

d-torrance:
> Please don't change commentize.

What about an option?  Default is the current behavior, but if given, uses the proposed behavior?

d-torrance:
> Also, the filenames were partially removed because they appeared as errors in Emacs compile-mode (see ##1719).

Gotcha -- I'll take a look at that.

> I removed it knowing that errors, we still get the test location in the error with `Verbose => true`:

Yes, I'm aware of that.  But we may not get that far, as I explained above.

mahrud:
> What about an option? Default is the current behavior, but if given, uses the proposed behavior?

The point of commenting with `-* .. *-` is either commenting a portion of a line or a block of code that you might want to uncomment later. Printed lines, especially stderr lines, should be individually commented. What happens, for instance, if printing a commentized block is interrupted? Then emacs' highlighting will be messed up, because `*-` isn't found.

d-torrance:
> > What about an option? Default is the current behavior, but if given, uses the proposed behavior?
> 
> The point of commenting with `-* .. *-` is either commenting a portion of a line or a block of code that you might want to uncomment later. Printed lines, especially stderr lines, should be individually commented. What happens, for instance, if printing a commentized block is interrupted? Then emacs' highlighting will be messed up, because `*-` isn't found.

Excellent point!

Maybe a better solution would be to fix the Emacs mode so that it will still jump to a file even with `--` in front?

mahrud:
> Maybe a better solution would be to fix the Emacs mode so that it will still jump to a file even with -- in front?

Sure, that would be nice. Then `code` can also put `--` before the source code, so the output can be correctly evaluated.

mahrud:
> > I removed it knowing that errors, we still get the test location in the error with `Verbose => true`:
> 
> Yes, I'm aware of that. But we may not get that far, as I explained above.

You always know the test number where the failure occurred, why not add a helper function to tell you the test location given the test number so you can debug it? Then you can add a method to `edit`, or perhaps a new function `editTest` or `viewTest` so you can automatically open an editor to that location.

d-torrance:
> > > I removed it knowing that errors, we still get the test location in the error with `Verbose => true`:
> > 
> > 
> > Yes, I'm aware of that. But we may not get that far, as I explained above.
> 
> You always know the test number before the failure occurs, why not add a helper function to tell you the test location given the test number so you can debug it? Then you can add a method to `edit`, or perhaps a new function `editTest` or `viewTest` so you can automatically open an editor to that location.

Good idea!

mahrud:
Lastly, regarding your changes to `pad`, I like the improvement, however:
1. they can be significantly simplified with `horizontalJoin`:
```m2
i33 : pad' = (n, N) -> horizontalJoin((n - width N):" ", N);

i34 : << pad'(4, net "abc\nd")
 abc
 d
```
2. I'm not sure whether a string "abc\nd" should be automatically assumed to be a net. What do I do if I actually want to prepend or append spaces to the string? Since anyone who wants it to be assumed as a net can just do `net "abc\nd"`, I'm inclined to think it shouldn't be assumed a net.

d-torrance:
> Lastly, regarding your changes to `pad`, I like the improvement, however:
> 
> 1. they can be significantly simplified with `horizontalJoin`:
> 
> ```m2
> i33 : pad' = (n, N) -> horizontalJoin((n - width N):" ", N);
> 
> i34 : << pad'(4, net "abc\nd")
>  abc
>  d
> ```

My first draft was essentially that, I believe:

```m2
pad (ZZ, Net) := (n, N) -> concatenate(n - width N) | N
```
which prompted the `concatenate` change.  However, this ended up not doing exactly what I expected.  Indeed:

```m2
i2 : elapsedTime << horizontalJoin(net "foo\nbar", "    ") << flush
foo    
bar -- 0.00009597 seconds elapsed
```
Where are the spaces after `bar`?  It's because we're only padding the first row:
```m2
i3 : unstack horizontalJoin(net "foo\nbar", "    ")

o3 = {foo    , bar}
```
The current version pads each row, giving us the expected behavior:
```m2
i5 : elapsedTime << stack apply(unstack net "foo\nbar", s -> pad(s, 6)) << flush
foo   
bar    -- 0.000101591 seconds elapsed
```

> 1. I'm not sure whether a string "abc\nd" should be automatically assumed to be a net. What do I do if I actually want to prepend or append spaces to the string? Since anyone who wants it to be assumed as a net can just do `net "abc\nd"`, I'm inclined to think it shouldn't be assumed a net.
> 
> In other words, you're assuming that any string containing `\r\n` or `\n` is a net, and I don't think that's true.

Fair point.  I'll revert that change.

mahrud:
> The current version pads each row, giving us the expected behavior:

Oh, pad the last row then:
```m2
i12 : pad' = (N, n) -> horizontalJoin(N, (concatenate(n - width N))^(-height N));

i13 : elapsedTime << pad'("foo\nbar", 10) << flush
foo
bar    -- 0.000068067 seconds elapsed
```
My point was that unstack and then stack is costly.

d-torrance:
Alrighty, I've force-pushed a new version of this PR with Mahrud's suggested changes.  In particular:

* No change to `check`.
* The only change to `commentize` is preserving the baseline of nets.
* No change to `pad` for strings, and simplification of `pad` for nets.

I'll use another PR for any future changes to `edit` or similar dealing with locating tests.

d-torrance:
Just added two more commits dealing with #1763.

d-torrance:
The test failure in the `autotools-ubuntu-latest-default` workflow looks like another instance of https://github.com/Macaulay2/M2/issues/1742#issuecomment-751502158.

Issue #1772, opened 2021-01-01
infoHelp not working in Emacs


d-torrance:
Running Macaulay2 (from a PPA build of the `release-1.17` branch) on Ubuntu 20.10 in Emacs 26.3:

```m2
i1 : infoHelp(rank, Matrix)
info: Terminal type 'dumb' is not smart enough to run Info
```
Doing the same thing from a terminal window opens `info` to the correct documentation page as expected.

mahrud:
Also, I think the PPA versions don't install the emacs package at all.

d-torrance:
> Also, I think the PPA versions don't install the emacs package at all.

They're in the `elpa-macaulay` package, which should be installed automatically unless you don't install `Recommends`.

mahrud:
Ah, I see, the problem is that they're being installed in `site-lisp/elpa/macaulay2-...` instead of `site-lisp/Macaulay2`, so `setupEmacs()` isn't putting the correct address in `.emacs-Macaulay2`.

mahrud:
Also, why is it prefixed with ELPA and put in that directory at all? It's not a GNU ELPA package (the list is [here](https://elpa.gnu.org/packages/), and ours isn't among them) at all.

Sorry about hijacking this issue, feel free to move it to a separate issue. I just remembered it when I saw this issue.

d-torrance:
> Ah, I see, the problem is that they're being installed in `site-lisp/elpa/macaulay2-...` instead of `site-lisp/Macaulay2`, so `setupEmacs()` isn't putting the correct address in `.emacs-Macaulay2`.

That isn't necessary for Emacs packages which are built using the `dh_elpa` debhelper addon.  M2-mode works out of the box once `elpa-macaulay2` is installed.

mahrud:
> That isn't necessary for Emacs packages which are built using the dh_elpa debhelper addon. M2-mode works out of the box once elpa-macaulay2 is installed.

I'm getting an error that "M2-init" isn't found.

d-torrance:
> Also, why is it prefixed with ELPA and put in that directory at all? It's not a GNU ELPA package (the list is [here](https://elpa.gnu.org/packages/), and ours isn't among them) at all.
> 
> Sorry about hijacking this issue, feel free to move it to a separate issue. I just remembered it when I saw this issue.

[...]

PR #1775, opened 2021-01-02
Add Emacs support to infoHelp

d-torrance:
If we detect that Macaulay2 is running inside Emacs, then print a
message instead of running info.  M2-mode will then use this message
to open the corresponding documentation node in Info mode.

Closes: #1772


d-torrance:
No hurry on this -- it can wait until after 1.17.

d-torrance:
Here's a screenshot:

image

Issue #1777, opened 2021-01-03
html HREF v. net HREF


d-torrance:
`net HREF` ignores all arguments after the second and doesn't properly format the arguments it doesn't ignore:

```m2
i1 : beginDocumentation()

i2 : href = HREF("https://foo.com", EM "foo", SPAN "bar", BOLD "baz");

i3 : html href

o3 = <a href="https://foo.com"><em>foo</em><span>bar</span><b>baz</b></a>

i4 : net href

o4 = EM{foo} (see https://foo.com )
```
This is visible when you run `help` on a package with a gold star, e.g., from `help "Cremona"`.  See the link to the git repository in the final sentence:

```m2
     Certification a gold star
     =========================

     Version 4.2.2 of this package was accepted for publication in volume 8
     (see https://msp.org/jsag/2018/8-1/ ) of The Journal of Software for
     Algebra and Geometry (see http://j-sag.org/ ) on 11 June 2018, in the
     article A Macaulay2 package for computations with rational maps (see
     https://msp.org/jsag/2018/8-1/p06.xhtml ). That version can be obtained
     from the journal (see https://msp.org/jsag/2018/8-1/jsag-v8-n1-x06-Cremona
     .zip ) or from (the ,EM{Macaulay2}, source code repository) (see
     http://github.com/Macaulay2/M2/blob/
     2e87a29e4b5b68af1bd8917a9c76d4008ff9fc5b/M2/Macaulay2/packages/Cremona.m2
     ).
```

DanGrayson:
HREF should be given just 2 arguments, and they should both be strings.  The first is the URL, and the second is something to display instead of the URL.  So the problem goes back to commit 17fcfe4bee.

d-torrance:
It looks like `HREF`'s behavior changed to allow more arguments in https://github.com/Macaulay2/M2/commit/8bfd3ca68cbf1f793c93a2512d4a756452b3f7b7.

mahrud:
Why can't HREF contain other hypertext objects? In almost all places `net` is trying to replicate how browsers present `html`, so I would think the fix would be to put `html r` instead here:
https://github.com/Macaulay2/M2/blob/2a7d81a7f4443b855de7f095a8b51ab4ce3c956c/M2/Macaulay2/m2/html.m2#L140

DanGrayson:
[...]

PR #1781, opened 2021-01-03
Update net/info HREF to match html HREF behavior

d-torrance:
Closes: #1777

### DIFF
--- a/M2/Macaulay2/m2/format.m2
+++ b/M2/Macaulay2/m2/format.m2
@@ -210,10 +210,8 @@ info IMG := x -> (
 
 net  HREF :=
 info HREF := x -> (
-     if #x === 1 then x#0
-     else if match ("^mailto:",x#0) then toString x#1
-     -- x#0 is sometimes the relative path to the file, but not from the current directory
-     else toString x#1 | " (see " | x#0 | " )")
+     r := horizontalJoin(net \ toList splice if #x > 1 then drop(x, 1) else x);
+     r | " (see " | x#0 | " )")
 
 net TABLE :=  x -> (
      (op,ag) := override(options TABLE, toSequence x);
--- a/M2/Macaulay2/m2/format.m2
+++ b/M2/Macaulay2/m2/format.m2
@@ -210,8 +210,14 @@ info IMG := x -> (
 
 net  HREF :=
 info HREF := x -> (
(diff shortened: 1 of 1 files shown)


d-torrance:
Hrm, I probably shouldn't have overwritten the `mailto` stuff.  Let me add another commit...

d-torrance:
Here is is in action:

```m2
i1 : beginDocumentation()

i2 : HREF{"http://foo.com", "look", SUP "at", SUB "this", "link!"}

         at
o2 = look      link! (see http://foo.com )
           this
```

And now `help` for gold star packages looks fine:

```m2
     Certification a gold star
     =========================

     Version 4.2.2 of this package was accepted for publication in volume 8
     (see https://msp.org/jsag/2018/8-1/ ) of The Journal of Software for
     Algebra and Geometry (see http://j-sag.org/ ) on 11 June 2018, in the
     article A Macaulay2 package for computations with rational maps (see
     https://msp.org/jsag/2018/8-1/p06.xhtml ). That version can be obtained
     from the journal (see https://msp.org/jsag/2018/8-1/jsag-v8-n1-x06-Cremona
     .zip ) or from the Macaulay2 source code repository (see
     http://github.com/Macaulay2/M2/blob/
     2e87a29e4b5b68af1bd8917a9c76d4008ff9fc5b/M2/Macaulay2/packages/Cremona.m2
     ).
```

mahrud:
Whenever you push a commit, Github Actions starts a whole build, and since Github gives each repository a finite number of concurrent builds, it slows subsequent actions. But actions on superseded commits aren't automatically cancelled, which means someone has to manually cancel them. All this to say that it's better to push in groups rather than a commit at a time.

d-torrance:
> Whenever you push a commit, Github Actions starts a whole build, and since Github gives each repository a finite number of concurrent builds, it slows subsequent actions. But actions on superseded commits aren't automatically cancelled, which means someone has to manually cancel them. All this to say that it's better to push in groups rather than a commit at a time.

Yeah, I totally understand, and generally try to do that and push all at once.  I just didn't notice these two things until after I'd pushed...

d-torrance:
Perhaps [this](https://github.com/marketplace/actions/cancel-workflow-action) could be useful to prevent having to deal with that manually?

PR #1784, opened 2021-01-04
Cancel previous workflow runs when pushing new commits to a branch

d-torrance:
This way, we don't have to cancel them manually.

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -61,6 +61,11 @@ jobs:
           #  cc: clang
 
     steps:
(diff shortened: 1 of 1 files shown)


d-torrance:
Hrm, this isn't working:

```
Error while cancelling workflow_id 1539317: Resource not accessible by integration
```

d-torrance:
Ah, it looks like it might always fail for users with read-only permissions: https://github.com/actions/first-interaction/issues/10.  So pretty much anyone that might submit a PR!

Closing -- sorry for the noise.

PR #1769, opened 2021-01-01
update M2-emacs submodule commit

mahrud:

PR #1786, opened 2021-01-04
Update M2-emacs submodule commit

d-torrance:
As requested in https://github.com/Macaulay2/M2/pull/1769#issuecomment-754040096


d-torrance:
I went ahead and merged in the commit with the corresponding `infoHelp` Macaulay2 code.  It hasn't been merged into `release-1.17` yet, and we need both pieces for this feature to work.  :)

PR #1789, opened 2021-01-05
Use AC_CHECK_PROGS to find git during configure.

d-torrance:
This prevents "git: command not found" errors when determining
GIT_DESCRIPTION if git is not present.  These don't cause configure to
fail, but are flagged as errors in Emacs compilation mode.

### DIFF
--- a/M2/configure.ac
+++ b/M2/configure.ac
@@ -560,7 +560,7 @@ fi
 AC_SUBST(GIT_DESCRIPTION)
 # TODO: currently finds the last commit that changed the VERSION file
 # but ideally it should get the last release commit instead
-NEW_COMMITS=`$GIT rev-list $(git rev-list -1 HEAD $srcdir/VERSION)..HEAD --count`
+NEW_COMMITS=`$GIT rev-list $($GIT rev-list -1 HEAD $srcdir/VERSION)..HEAD --count`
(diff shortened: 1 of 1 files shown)


d-torrance:
No rush -- this can wait until after 1.17.

d-torrance:
Oops -- I missed the second `git` in the definition of `NEW_COMMITS`.

M2/configure.ac
```diff
@@ -559,8 +560,8 @@ fi
 AC_SUBST(GIT_DESCRIPTION)
 # TODO: currently finds the last commit that changed the VERSION file
 # but ideally it should get the last release commit instead
-NEW_COMMITS=`git rev-list $(git rev-list -1 HEAD $srcdir/VERSION)..HEAD --count`
-GIT_COMMIT=`git describe --dirty --always --match HEAD`
+NEW_COMMITS=`$GIT rev-list $($GIT rev-list -1 HEAD $srcdir/VERSION)..HEAD --count`
```
mahrud, line 563: wouldn't it now give "rev-list: command not found" errors?

d-torrance: No, `$GIT` will be `false` if git isn't found, and `false rev-list ...` just silently fails.

mahrud: Ah, nice!

PR #1794, opened 2021-01-06
just a little bunch of typos in some packages

fchapoton:
found using codespell

in packages Posets, Graphs, Polyhedra, Polymake


fchapoton:
should I go on with some other packages ? or is this too much trouble for the gain ?

DanGrayson:
> should I go on with some other packages ? or is this too much trouble for the gain ?

Well, you're the only one who can judge whether it's too much trouble, but in my opinion, the corrections are welcome and we're grateful.

PR #1795, opened 2021-01-06
"infoHelp" updates

d-torrance:
We add some new features to `infoHelp` to bring it more in line with `help` and `viewHelp`:

* It is now a method function.  `(infoHelp, Thing)` has the old behavior and `(infoHelp, ZZ)` plays nicely with `about`.
* It is now a command.  `infoHelp` with no argument is equivalent to `infoHelp "Macaulay2Doc"`.

The documentation has also been updated to reflect these changes (as well as the previous recent change adding Emacs support).

### DIFF
--- a/M2/Macaulay2/m2/help.m2
+++ b/M2/Macaulay2/m2/help.m2
@@ -453,10 +453,12 @@ setAttribute(viewHelp#0, ReverseDictionary, symbol viewHelp)
 
 infoHelp = method(Dispatch => Thing)
 infoHelp Thing := key -> (
+    if key === () then infoHelp "Macaulay2Doc";
     tag := infoTagConvert makeDocumentTag(key, Package => null);
     if getenv "INSIDE_EMACS" == "" then chkrun ("info " | format tag)
     else print("-* infoHelp: " | tag | " *-");)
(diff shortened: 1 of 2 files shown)


d-torrance:
From the documentation for `infoHelp`:

> If you do so, we recommend configuring the value of the Emacs
variable Info-hide-note-references to hide in order to prevent Emacs from
inserting a superfluous See or see in front of the hyperlinks.

It might be nice to just do this for the user when they run `setupEmacs()`.  I could add a commit to this PR doing this if there's interest.

DanGrayson:
Does this point emacs at the info files of the version of M2 currently running?  Or does emacs just display the info files it's already got listed in `Info-default-directory-list`?

I'll merge, anyway.

d-torrance:
> Does this point emacs at the info files of the version of M2 currently running? Or does emacs just display the info files it's already got listed in `Info-default-directory-list`?

The latter -- it just fires up Info mode, which wouldn't be aware of, say, .info files that just got built if you're running from a build directory or something.

DanGrayson:
By contrast, viewHelp points to the documentation of the running M2.  Something to think about in the future.

PR #1797, opened 2021-01-07
Include "Synopsis" SimpleDoc keyword in syntax highlighting

d-torrance:
It's currently missing.  See, for example, Github's own Markdown syntax highlighting:

```m2
       Synopsis
        Heading
         a different way to use this method
        Usage
         simpleDocFrom_n M
        Inputs
         :ZZ
         :Matrix
        Outputs
         :Matrix
        Consequences
         Item
          There may be specific consequences.

```

PR #1801, opened 2021-01-07
CMake maintenance

mahrud:
This includes the commit from #1797, but both can be merged. Git should be able to fast-forward.

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -90,9 +90,8 @@ jobs:
                   libgc-dev libgdbm-dev libglpk-dev libgmp3-dev libgtest-dev libmpfr-dev libntl-dev gfan \
                   libcdd-dev libgivaro-dev libboost-regex-dev fflas-ffpack w3c-markup-validator \
                   cohomcalg libflint-dev libmps-dev topcom \
-                  libsingular-dev libgtest-dev libmemtailor-dev libmathic-dev \
+                  libsingular-dev libfrobby-dev libgtest-dev libmemtailor-dev libmathic-dev \
                   libmathicgb-dev 4ti2 normaliz coinor-csdp nauty lrslib singular-data
-                  # TODO: add libfrobby-dev again when it is updated
 
 # ----------------------
 #   Steps common to all build variants
@@ -140,8 +139,6 @@ jobs:
       - name: Build libraries using Ninja
         if: matrix.build-system == 'cmake' # && steps.restore-cache.outputs.cache-hit != 'true'
         run: |
-          # TODO: remove this line soon
-          cmake --build . --target build-frobby
           cmake --build . --target build-libraries build-programs
 
       - name: Build Macaulay2 using Ninja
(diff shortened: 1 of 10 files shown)


mahrud:
@d-torrance now that https://github.com/Macaulay2/frobby/pull/5 is merged, do you know when the PPA package will be updated?

@DanGrayson how do I turn frobby into a submodule, again?

PS: don't wait on this PR for v1.17. I'm just using the release-1.17 branch as the target because development seems to be outdated.

d-torrance:
> @d-torrance now that [Macaulay2/frobby#5](https://github.com/Macaulay2/frobby/pull/5) is merged, do you know when the PPA package will be updated?

frobby isn't currently in the PPA since it's in the official Ubuntu repositories.  I suppose I could update it, but ideally the `version.dd` code should be backwards compatible and use `constants::version` if `frobby_version` isn't present.  The older versions will be around for a while.  For example, 0.9.0 is in Ubuntu 20.04 LTS, and its end of life isn't until [2032](https://ubuntu.com/about/release-cycle)!

DanGrayson:
> @DanGrayson how do I turn frobby into a submodule, again?
> 

I think you already did that in commit 7f58b0bef0814dcf4a0522aa7b4e00eea330bf14

mahrud:
> > @DanGrayson how do I turn frobby into a submodule, again?
> 
> I think you already did that in commit [7f58b0b](https://github.com/Macaulay2/M2/commit/7f58b0bef0814dcf4a0522aa7b4e00eea330bf14)

I meant for autotools. It is currently downloading an older tarfile.

DanGrayson:
> > > @DanGrayson how do I turn frobby into a submodule, again?
> > 
> > 
> > I think you already did that in commit [7f58b0b](https://github.com/Macaulay2/M2/commit/7f58b0bef0814dcf4a0522aa7b4e00eea330bf14)
> 
> I meant for autotools. It is currently downloading an older tarfile.

Look in configure.ac.  There are two lists, LIBLIST and SUBLIST.  The former is the list of library names, and the latter is the list of submodule names.  Move frobby from the former to the latter.  Then you might have to add some targets to GNUmakefile.in, using the other submodules as models, to get the right license files copied, and to provide a target that the configure makes, so we know whether to run the configure script.

mahrud:
@DanGrayson the brew bottle is based on this branch, so it would be good to merge it, but either is fine.

DanGrayson:
I think the diff should show 17 files changed, not 73.  Is github slow to recompute diffs after the target branch is updated?

mahrud:
@DanGrayson there, I rebased it.

mahrud:
The failure is the same monomial ideal bug, again in ComputationsBook/monomialIdeals/test.m2.

mahrud:
Actually, it's probably better to close this one and just merge #1803. Feel free to reopen and merge this first if you can't continue your review on that PR (you should be able to, since these commits are there too).

PR #1819, opened 2021-01-16
another wave of typos in packages (between L* and V*, not R*)

fchapoton:
in the packages with names starting with a letter between L and V excluding R


DanGrayson:
Thanks!

PR #1827, opened 2021-01-17
add manual trigger for test builds and remove pre-master

mahrud:


DanGrayson:
I like the idea of having a manual trigger, so I've just put that on the development branch to try it out.  But I think we'll steer clear of the cache until after a successful build on the master branch.

PR #1829, opened 2021-01-18
Don't capture DiffAlg tests

d-torrance:
As discussed in #1728, there appears to be a memory leak.  The test
failed on a recent build of the Debian package on powerpc.


mahrud:
I wonder if it would be more efficient to add DiffAlg to the list of packages not to use capture with:
https://github.com/Macaulay2/M2/blob/9b0b7f0c72ad4b5d1ad7f3814786736c35c87022/M2/Macaulay2/m2/examples.m2#L118

d-torrance:
> I wonder if it would be more efficient to add DiffAlg to the list of packages not to use capture with:
> https://github.com/Macaulay2/M2/blob/9b0b7f0c72ad4b5d1ad7f3814786736c35c87022/M2/Macaulay2/m2/examples.m2#L118

Makes sense -- I'll push a new version.

PR #1809, opened 2021-01-11
Upgrade to FLINT 2.7.0 and Factory 4.2.0

mahrud:
- [x] Fixes #1808 for CMake
- [x] update factory build for autotools.

@DanGrayson: this only updates the CMake build. The builds won't fail, because flint is installed from either brew or apt currently, but could you fix the autotools build and add a commit to this PR?

PR #1832, opened 2021-01-18
autotools: update flint to 2.7.0 and factory to 4.2.0, part 2

DanGrayson:
should have been done as part of commit 1494b858e0afd4280154a6390d5fb6ab056ec9d4 for https://github.com/Macaulay2/M2/pull/1809

(This fixes the configure script to look for those versions in the system.)

### DIFF
--- a/M2/configure.ac
+++ b/M2/configure.ac
@@ -937,17 +937,14 @@ dnl factory needs to know:
 AC_SUBST(BUILD_flint)
 if test $BUILD_flint = no
 then AC_LANG(C)
-     AC_MSG_CHECKING([for flint >= 2.6.0 and < 2.7.0])
+     AC_MSG_CHECKING([for flint >= 2.7.0])
      AC_COMPILE_IFELSE(
         [AC_LANG_PROGRAM(
 	  [#include <flint/flint.h>
 	  ],
 	  [
-#if __FLINT_VERSION < 2 || (__FLINT_VERSION == 2 && __FLINT_VERSION_MINOR < 6)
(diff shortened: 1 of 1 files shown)


d-torrance:
flint >= 2.7.0 is only necessary for building factory, right?  Could we keep the old >= 2.6.0 check if we're not building factory?

d-torrance:
Also, is it necessary to require factory 4.2.0?

mahrud:
I think we must require or build Factory 4.2.0, since it is not compatible with Flint 2.7.0, but it is compatible with lower versions of Flint, so we could keep the minimum for Flint at 2.6.3.

DanGrayson:
> I think we must require or build Factory 4.2.0, since it is not compatible with Flint 2.7.0, but it is compatible with lower versions of Flint, so we could keep the minimum for Flint at 2.6.3.

Hmm?  I think factor 4.2.0 is compatible with flint 2.7.0.  At least I can build them together.  Is there a run-time failure?
```m2
Macaulay2, version 1.17.1.1
with packages: ConwayPolynomials, Elimination, IntegralClosure, InverseSystems, LLLBases, MinimalPrimes, PrimaryDecomposition, ReesAlgebra, Saturation, TangentCone

i1 : version#"factory version"

o1 = 4.2.0

i2 : version#"flint version"

o2 = 2.7.0
```

mahrud:
Sorry, typo. Corrected now.

DanGrayson:
> I think we must require or build Factory 4.2.0, since older versions of Factory are not compatible with Flint 2.7.0, but Factory 4.2.0 is compatible with lower versions of Flint as well, so we could keep the minimum for Flint at 2.6.3.

will do

Issue #1834, opened 2021-01-18
check_254 "Core" failing during "M2 --check 2" on i386


d-torrance:
This has happened twice so far ([here](https://ci.debian.net/data/autopkgtest/unstable/i386/m/macaulay2/9795053/log.gz) and [here](https://ci.debian.net/data/autopkgtest/unstable/i386/m/macaulay2/9794142/log.gz)) on continuous integration tests of the 1.17.1 package in Debian unstable for i386:

```m2
 -- capturing check(252, "Core")                                             -- 0.21414 seconds elapsed
 -- capturing check(253, "Core")                                             -- 3.12982 seconds elapsed
 -- capturing check(254, "Core")                                            

 *** out of memory trying to allocate 10000 bytes, exiting ***

 cd /tmp/M2-2214-0/1-rundir/; GC_MAXIMUM_HEAP_SIZE=400M "/usr/bin/M2-binary" -q --int --no-randomize --no-readline --silent --stop --print-width 129 <"/tmp/M2-2214-0/0.m2" >>"/tmp/M2-2214-0/0.tmp"
/tmp/M2-2214-0/0.tmp:0:1: (output file) error: Macaulay2 exited with status code 1
/tmp/M2-2214-0/0.m2:0:1: (input file)
M2: *** Error 1
```

It doesn't appear to be a problem when just running the test by itself, so I think it's probably just a result of having just run 254 other tests right before it!  From an i386 chroot:

```m2
i1 : check_254 "Core"
 -- capturing check(254, "Core")                                             -- 0.000356972 seconds elapsed
```

d-torrance:
I tried adding `-* no-capture-flag *-` to that test, and it just ran out of memory in the next test!

```m2
 -- capturing check(254, "Core")                                             -- 0.137893 seconds elapsed
 -- capturing check(255, "Core")                                            

 *** out of memory trying to allocate 10000 bytes, exiting ***

 cd /tmp/M2-2006253-0/1-rundir/; GC_MAXIMUM_HEAP_SIZE=400M "/usr/bin/M2-binary" -q --int --no-randomize --no-readline --silent --stop --print-width 129 --srcdir "macaulay2-1.17.1+ds/M2/" <"/tmp/M2-2006253-0/0.m2" >>"/tmp/M2-2006253-0/0.tmp"
/tmp/M2-2006253-0/0.tmp:0:1: (output file) error: Macaulay2 exited with status code 1
[...]

PR #1835, opened 2021-01-18
Don't use capture during check "Core"

d-torrance:
Otherwise, we run out of memory on i386.

Closes: #1834

### DIFF
--- a/M2/Macaulay2/m2/examples.m2
+++ b/M2/Macaulay2/m2/examples.m2
@@ -115,7 +115,7 @@ isCapturable = (inputs, pkg, isTest) -> (
     inputs = replace("-\\*.*?\\*-", "", inputs);
     -- TODO: remove this when the effects of capture on other packages is reviewed
     (isTest or match({"FirstPackage", "Macaulay2Doc"},            pkg#"pkgname"))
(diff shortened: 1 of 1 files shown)


M2/Macaulay2/m2/examples.m2
```diff
@@ -115,7 +115,7 @@ isCapturable = (inputs, pkg, isTest) -> (
     inputs = replace("-\\*.*?\\*-", "", inputs);
     -- TODO: remove this when the effects of capture on other packages is reviewed
     (isTest or match({"FirstPackage", "Macaulay2Doc"},            pkg#"pkgname"))
-    and not match({"EngineTests", "ThreadedGB", "RunExternalM2", "DiffAlg"}, pkg#"pkgname")
+    and not match({"EngineTests", "ThreadedGB", "RunExternalM2", "DiffAlg", "Core"}, pkg#"pkgname")
```
mahrud, line 118: If you must, add another line to disable capture based on architecture, but I strictly disagree with this.

mahrud (changes requested):

d-torrance:
Closing -- see https://github.com/Macaulay2/M2/issues/1834#issuecomment-762387080

PR #1837, opened 2021-01-18
Angle bar list

DanGrayson:
add a new type of visible list of a type called AngleBarList
    
Example:

        i11 : <| x,y,z |>
    
        o11 = <|x, y, z|>

Intended use: associative algebras, by Stillman and Moore

### DIFF
--- a/M2/configure.ac
+++ b/M2/configure.ac
@@ -1139,8 +1139,8 @@ AC_LANG(C++)
 AC_CHECK_HEADER(frobby.h,LIBS="-lfrobby $LIBS",BUILD_frobby=yes)
 if test $BUILD_frobby = yes
 then BUILTLIBS="-lfrobby $BUILTLIBS"
-    dnl change this to [1] when frobby is built from a submodule
-    AC_DEFINE([HAVE_FROBBY_VERSION], [0], [whether frobby has frobby_version])
+    dnl uncomment this to [1] when frobby is built from a submodule (?)
+    dnl AC_DEFINE([HAVE_FROBBY_VERSION], [0], [whether frobby has frobby_version])
 else
     AC_MSG_CHECKING([whether frobby has frobby_version])
     AC_COMPILE_IFELSE(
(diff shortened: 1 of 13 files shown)


DanGrayson:
@mikestillman @moorewf

M2/configure.ac
```diff
@@ -1139,8 +1139,8 @@ AC_LANG(C++)
 AC_CHECK_HEADER(frobby.h,LIBS="-lfrobby $LIBS",BUILD_frobby=yes)
 if test $BUILD_frobby = yes
 then BUILTLIBS="-lfrobby $BUILTLIBS"
-    dnl change this to [1] when frobby is built from a submodule
-    AC_DEFINE([HAVE_FROBBY_VERSION], [0], [whether frobby has frobby_version])
+    dnl uncomment this to [1] when frobby is built from a submodule (?)
+    dnl AC_DEFINE([HAVE_FROBBY_VERSION], [0], [whether frobby has frobby_version])
```
mahrud, line 1143: This is not going to work with both the new and old version of frobby.

mahrud:
Is this different from the [bra-ket notation](https://en.wikipedia.org/wiki/Bra%E2%80%93ket_notation)?

M2/configure.ac
mahrud, line 1143: Did you test this with both the old version and the new version of Frobby?

DanGrayson:
> Is this different from the [bra-ket notation](https://en.wikipedia.org/wiki/Bra%E2%80%93ket_notation)?

Yes.  Bra-ket notation is the binary operator `<x|y>`, together with the convention that `<x|` and `|y>` can stand alone.

M2/configure.ac
DanGrayson: There's nothing to test -- the line is active only when we are building frobby, and the version to build is hard-coded in frobby/Makefile.in.

Also, `#define HAVE_FROBBY_VERSION 0` doesn't have the desired effect on your line `#ifdef HAVE_FROBBY_VERSION` in `version.dd`.

Issue #1841, opened 2021-01-19
findProgram can't handle empty entries in PATH


DanGrayson:
```m2
+ PATH=:/Applications/Macaulay2-1.17/bin:/Users/dan/src/ProofChecking/agda/.cabal-sandbox/bin:/usr/local/texlive/2018/bin/x86_64-darwin:/Users/dan/src/M2/qesource/bin:/Users/dan/.opamroot/default/bin:/usr/local/opt/libxml2/bin:/Users/dan/bin:/Users/dan/local/bin:/Applications/DjView.app/Contents/bin:/usr/local/brew-root/sbin:/usr/local/brew-root/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/./sbin:/./bin:/opt/X11/bin:/nowhere:/Users/dan/.rvm/bin:/Applications/Emacs-27.1-1.app/Contents/MacOS/bin-x86_64-10_14:/Applications/Emacs-27.1-1.app/Contents/MacOS/libexec-x86_64-10_14
+ /Users/dan/src/M2/M2.git/M2/BUILD/dan/builds.tmp/gallium-development/M2 --no-readline --print-width 197
Macaulay2, version 1.17.1.1
with packages: ConwayPolynomials, Elimination, IntegralClosure, InverseSystems, LLLBases, MinimalPrimes, PrimaryDecomposition, ReesAlgebra, Saturation, TangentCone
Macaulay2/Core/startup.m2.in:121:19:(0):[26]: error: string index out of bounds
Macaulay2/Core/startup.m2.in:121:19:(0):[26]: --entering debugger (type help to see debugger commands)
Macaulay2/Core/startup.m2.in:121:18-121:21: --source code:
     last = x -> x#-1;

i1 : 
```

An empty entry indicates that the current directory should be searched.

d-torrance:
> An empty entry indicates that the current directory should be searched.

Is that true?

From https://pubs.opengroup.org/onlinepubs/000095399/basedefs/xbd_chap08.html#tag_08_03:

> A strictly conforming application shall use an actual pathname (such as .) to represent the current working directory in PATH .

And from https://en.wikipedia.org/wiki/PATH_(variable):

> System administrators as a rule do not include it in $PATH in order to prevent the accidental execution of scripts residing in the current directory, such as may be placed there by a malicious tarbomb.

DanGrayson:
> > An empty entry indicates that the current directory should be searched.
> 
[...]

PR #1843, opened 2021-01-19
Interpret an empty string in PATH as "." in findProgram

d-torrance:
Closes: #1841


d-torrance:
For example:

```m2
i1 : getenv "PATH"

o1 = :/usr/bin

i2 : findProgram("gfan", "gfan --help", Verbose => true)
checking for gfan in /usr/lib/x86_64-linux-gnu/Macaulay2/bin/...
    not found
checking for gfan in ./...
    not found
checking for gfan in /usr/bin/...
    found

o2 = gfan

o2 : Program
```

PR #1844, opened 2021-01-20
Show correct minimum version of flint in autoconf check

d-torrance:
We're actually checking for a minimum version of 2.6.3, but the
message was stating a lower bound of 2.7.0.

### DIFF
--- a/M2/configure.ac
+++ b/M2/configure.ac
@@ -937,7 +937,7 @@ dnl factory needs to know:
 AC_SUBST(BUILD_flint)
 if test $BUILD_flint = no
 then AC_LANG(C)
-     AC_MSG_CHECKING([for flint >= 2.7.0])
+     AC_MSG_CHECKING([for flint >= 2.6.3])
      AC_COMPILE_IFELSE(
         [AC_LANG_PROGRAM(
 	  [#include <flint/flint.h>


d-torrance:
I was really confused as to why it was saying that >= 2.7.0 had been found during my build when only 2.6.3 was present!

I'm still puzzled as to why we're requiring factory 4.2.0 now.  Everything was working fine for me with flint 2.6.x and factory 4.1.1.  It seems like maybe the check should be a little more granular:

* If both flint and factory are on the system, what are the minimum versions?
* If only flint is on the system, what is the minimum version to build factory?
* If neither are on the system, go ahead and build 2.7.0 and 4.2.0.

mahrud:
In general, I think it's fair to require that for building Macaulay2 from source you must have the latest version of all the requirements, and if not we will build them. For distributions, however, the packaged Macaulay2 may only use the latest version available on the distribution.

DanGrayson:
If it works fine with factory 4.1.1, and that's all that's available under Debian, then indeed, why not use it?  Should I relax that check?

d-torrance:
> If it works fine with factory 4.1.1, and that's all that's available under Debian, then indeed, why not use it? Should I relax that check?

That would be great!  Everything works at least as far back as 4.1.0, which is what's in Ubuntu 18.04.

DanGrayson:
done in commit ef1ebded5f42854e992043b54d267832cd15f98e

PR #1845, opened 2021-01-20
once again typos in packages, last such branch maybe

fchapoton:
one last (?) branch for fixing typos in packages/


DanGrayson:
Thanks for all these fixes!

Issue #1787, opened 2021-01-04
checking a package changes the namespace


mikestillman:
The following code appears to assign a value to `f`.

```
needsPackage "Complexes"
check Complexes
f
```

mahrud:
Ah, it's because I'm passing the `UserMode` option of `check` to `capture`:
https://github.com/Macaulay2/M2/blob/9a7832bc6d77020eb146e93899afc8617644dee7/M2/Macaulay2/m2/testing.m2#L43

If you do `check(Complexes, UserMode => false)` it should go away.

But perhaps we should change this. Either default value of `[check, UserMode]` should be false, or `check` should always pass `UserMode => false` to capture. Which would you prefer?

PR #1847, opened 2021-01-20
CMake maintenance

mahrud:


M2/configure.ac
```diff
@@ -2035,23 +2035,23 @@ AC_MSG_NOTICE([staging area for architecture dependent files: $pre_exec_prefix])
 
 AC_SUBST(packagesdir,$datadir/Macaulay2)
 AC_SUBST(libm2dir,$libdir/Macaulay2)
-AC_SUBST(emacsdir,$datadir/emacs/site-lisp/Macaulay2)
+AC_SUBST(emacsdir,$datadir/emacs/site-lisp/macaulay2)
```
mahrud, line 2038: These changes were made to make the directory more aligned with the location of M2-mode in the brew and debian packages. Apparently the directories should always be lowercase and not abbreviated.

DanGrayson: Unless it's actually mandatory, I think it would be better for Macaulay2 always to be capitalized.  After all, "Macaulay" is the name of a person.

mahrud: I would have preferred that too, but it is mandatory on brew, debian, and even MELPA, so I think the upstream should match the convention as closely as possible.

DanGrayson: That's going to be impossible to obey.  In particular, the names of the auxiliary directories for the packages are all capitalized.

Can you point  me to statements of the requirement in those three systems?

mahrud: I'm talking about emacs packages only! What auxiliary directories do you mean?

DanGrayson: Auxiliary directories such as `M2/Macaulay2/packages/Bertini`.

So, when you said "the directories should always be lowercase", you actually meant "the emacs packages should always be lowercase"?

In that case, there is no problem, because `emacs/site-lisp/Macaulay2` is not an emacs package -- it's a directory name.

mahrud: Sometimes I feel like you intentionally misread my comments ...
When creating a package for debian or a bottle for brew, it is mandatory that if they install an emacs major mode the directory name for the major mode be lowercase. @d-torrance could you elaborate for debian?

It doesn't have anything to do with the contents of `share/Macaulay2`. Only int the contents of `share/emacs`, the directory names need to be lowercase. Not even the filenames, as we have `M2-mode.el` and that is fine.

DanGrayson: Oh, okay.  Thanks for adding details.

DanGrayson: I just ran lintian against my latest distribution, and `share/site-lisp/Macaulay2` was not flagged, not even as a warning.

```sh
ubuntu2010$ lintian Macaulay2-1.17-common.deb
E: macaulay2-common: install-info-used-in-maintainer-script postinst:7
E: macaulay2-common: install-info-used-in-maintainer-script postinst:8
E: macaulay2-common: install-info-used-in-maintainer-script prerm:7
E: macaulay2-common: latest-changelog-entry-without-new-date
W: macaulay2-common: debian-changelog-has-wrong-day-of-week 2017-06-10 is a Saturday
W: macaulay2-common: description-synopsis-starts-with-article
W: macaulay2-common: duplicate-font-file usr/share/Macaulay2/Visualize/fonts/glyphicons-halflings-regular.ttf also in fonts-glewlwyd
W: macaulay2-common: embedded-javascript-library usr/share/Macaulay2/Visualize/js/jquery-1.11.3.min.js please use libjs-jquery
W: macaulay2-common: embedded-javascript-library usr/share/Macaulay2/Visualize/js/underscore.js please use libjs-underscore
W: macaulay2-common: package-contains-timestamped-gzip usr/share/doc/macaulay2-common/changelog.Debian.gz 2021-01-05T20:17:50
W: macaulay2-common: package-contains-timestamped-gzip usr/share/info/AbstractToricVarieties.info.gz 2021-01-08T04:16:10
W: macaulay2-common: package-contains-timestamped-gzip usr/share/info/AdjointIdeal.info.gz 2021-01-08T03:40:48
W: macaulay2-common: package-contains-timestamped-gzip ... use --no-tag-display-limit to see all (or pipe to a file/program)
W: macaulay2-common: priority-extra-is-replaced-by-priority-optional
W: macaulay2-common: truetype-font-prohibits-installable-embedding [preview/print only] usr/share/Macaulay2/Visualize/fonts/glyphicons-halflings-regular.ttf
W: macaulay2-common: wrong-name-for-changelog-of-native-package usr/share/doc/macaulay2-common/changelog.Debian.gz
ubuntu2010$ mkdir tmp
ubuntu2010$ cd tmp
ubuntu2010$ ar x ../Macaulay2-1.17-common.deb
ubuntu2010$ tar tzf data.tar.gz | grep site-lisp
usr/share/emacs/site-lisp/
usr/share/emacs/site-lisp/Macaulay2/
usr/share/emacs/site-lisp/Macaulay2/M2-symbols.el
usr/share/emacs/site-lisp/Macaulay2/M2-emacs-help.txt
usr/share/emacs/site-lisp/Macaulay2/M2.el
usr/share/emacs/site-lisp/Macaulay2/M2-emacs.m2
usr/share/emacs/site-lisp/Macaulay2/M2-mode.el
usr/share/emacs/site-lisp/Macaulay2/README.md
usr/share/emacs/site-lisp/Macaulay2/M2-init.el
```

mahrud: So? This check is not a comprehensive check of requirements to be distributed in debian or ubuntu. It's not even the full list of checks available through lintian:

M2-mode is not distributed with @d-torrance's Macaulay2 package. It's in the elpa-macaulay2 package. The lisp files are bytecompiled, and the directory name is `/usr/share/emacs/site-lisp/elpa/macaulay2-1.17`. Just because you can make a `.deb` package doesn't mean it is acceptable.

d-torrance: The directory containing the Emacs files in the Debian package is lowercase as a consequence of several things.

Debian package names must be all lowercase ([Policy 5.6.1](https://www.debian.org/doc/debian-policy/ch-controlfields.html#source)) and the [Debian Emacsen Team's policy](https://wiki.debian.org/Teams/DebianEmacsenTeam) is that Emacs packages should be of the form `elpa-foo`.  So I chose `elpa-macaulay2` for consistency with `macaulay2` and `macaulay2-common`.

The default behavior of [dh_elpa](https://manpages.debian.org/unstable/dh-elpa/dh_elpa.1.en.html), which does the heavy lifting of creating a nice Emacs binary package, is to use the `foo` part of `elpa-foo` as the Emacs package name, and that in turn is used to create the directory name, which is also appended with the version number.  So that's where the `/usr/share/emacs/site-lisp/elpa/macaulay2-1.17` is coming from.

It looks like I could change it to `Macaulay2` by setting the `ELPA_NAME` environment variable during build, but the path would be different enough from upstream anyway (with the `elpa` subdirectory and the version number) that I haven't bothered.

Also, even though I haven't found it explicitly stated anywhere as policy, I can't find any packages which ship a subdirectory of `/usr/share/emacs/site-lisp/elpa` containing a capital letter.

mahrud: To the last point, I think it's because emacs modes are case insensitive, so to avoid conflicts they want everything to be lowercase.

Here is where brew requires that the directory name match the package name:
https://github.com/Homebrew/brew/blob/6fc1570134f852748f5848dddc29b554c6bb2fd3/Library/Homebrew/formula_cellar_checks.rb#L135-L151
And the packages are lower case, so it has to be `macaulay2`.

DanGrayson: Okay, thanks.

M2/Macaulay2/m2/testing.m2
```diff
@@ -40,7 +40,8 @@ captureTestResult := (desc, teststring, pkg, usermode) -> (
     -- try capturing in the same process
     if isCapturable(teststring, pkg, true) then (
 	checkmsg("capturing", desc);
-	(err, output) := capture(teststring, PackageExports => pkg, UserMode => usermode);
+	-- TODO: adjust and pass argumentMode, instead. This can be done earlier, too.
+	(err, output) := capture(teststring, PackageExports => pkg, UserMode => false);
```
DanGrayson, line 44: Suppose a user gets the source code for two packages that are not part of Macaulay2 and wants to install and test them.  Suppose one depends on the other.  The test will not work if UserMode is false here.  So, I think we have to keep this the way it was.

mahrud: That's not what `UserMode` does for `capture`. It only affects whether the symbols from the test end up in the User dictionary, see #1787. The interpreter inside `capture` uses the same `path` variable as the M2 interpreter that called it, so in your scenario the packages will be loaded fine, as long as `-q` isn't given to M2.

mahrud: But I think you're right that there's an edge case that I had missed. I justpushed a commit to avoid a situation where the command line has `-q` but `UserMode => true` is given, so `capture` isn't used.

M2/Macaulay2/m2/CMakeLists.txt
```diff
@@ -29,4 +30,8 @@ foreach(filename IN LISTS M2LIST ITEMS
   file(COPY ${filename} DESTINATION ${CORE_DIR})
 endforeach()
 
+## Copy Core tests from Macaulay2/tests/normal to share/doc/Macaulay2/Core/tests
+file(COPY ${CMAKE_SOURCE_DIR}/Macaulay2/tests/normal/ DESTINATION ${CORE_TESTS}
```
d-torrance, line 34: I'd tried to do this in https://github.com/Macaulay2/M2/commit/7a55f01d5c7cc3c89612b168a4ff82e652aa03ad, but I guess it wasn't working properly?  We probably should revert that commit, too.

PR #1861, opened 2021-01-22
Noncommutative Algebra Functionality

moorewf:
This is the first pull request adding some engine functionality for noncommutative algebras, e.g. two-sided Gröbner bases for ideals, kernels of graded ring homomorphisms, etc.


M2/CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ cmake_minimum_required(VERSION 3.15...3.19)
 cmake_policy(VERSION 3.14)
 
 ## Use the C++14 standard
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
```
mahrud, line 22: Nice! Whenever you get a chance, could you update the commented line above too?

M2/Macaulay2/e/CMakeLists.txt
```diff
@@ -34,6 +35,9 @@ set(HFILES
 
 # these files are include only.
 set(HPPONLYFILES
+  	MemoryBlock.hpp
+	NCAlgebras/Range.hpp
+	VectorArithmetic2.hpp
```
mahrud, line 40: Not urgent, but could you keep to 2 spaces for indentation? That's the common size for CMake.

M2/Macaulay2/e/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ set(UNDETERMINED
 
 # c files
 # TODO: convert to cpp or simplify if not needed
+# MES TODO:   monordering.c removed this, is that correct?
```
mahrud, line 18: What was removed? I remember I merged some monomial ordering stuff by hand, did I miss something?

M2/Macaulay2/d/interface.dd
```diff
@@ -770,6 +770,76 @@ export rawSolvableAlgebra(e:Expr):Expr := (
...
+      "rawRingM2FreeAlgebraQuotient(",
+        a, ",", b, 
+      ")"
+    ))
+  ) else WrongArgSmallInteger(1)
+  ) else WrongArgZZ(1)
+  ) else WrongArg(0,"a raw matrix")
+  ) else WrongNumArgs(2)
+  ) else WrongNumArgs(2)
+  );
```
mahrud, line 816: Some functions seem to be indented at 2, some at 5. It might be good to convert all to 4, since it divides 8 tab-size and is the same as in M2-mode.

M2/Macaulay2/e/M2FreeAlgebraQuotient.cpp
```diff
@@ -0,0 +1,378 @@
...
+
+#include <iostream>
+#include <memory>
+#include <utility>
+#include <vector>
+
+#include "NCAlgebras/FreeAlgebra.hpp"
+#include "matrix.hpp"
+#include "ring.hpp"
+#include "stdinc-m2.hpp"
```
mahrud, line 11: What is `stdinc-m2.hpp`? I forgot to ask this when I was merging the code. Should it go in `engine-includes.hpp`?

M2/Macaulay2/e/util.cpp
```diff
@@ -2,7 +2,17 @@
 
 #include "stdinc-m2.hpp"
 #include "util.hpp"
-#include <string>
+
```
mahrud, line 5: I think `#include <string>` should still appear here, since this file uses it. (It should appear in `util.hpp` as well)

M2/Macaulay2/e/unit-tests/RingTowerTest.cpp
```diff
@@ -10,6 +10,7 @@
 
 #include "RingTest.hpp"
 #include "tower.hpp"
+#include "../util.hpp"
```
mahrud, line 13: I don't believe you need `../` here. See `tower.hpp` one line above.

M2/CMakeLists.txt
```diff
@@ -19,10 +19,12 @@ cmake_minimum_required(VERSION 3.15...3.19)
...
 ## Use the C11 (C0X) standard
 set(CMAKE_C_STANDARD 11)
 
+set(WITH_TBB ON CACHE BOOL "whether to use TBB")
```
mahrud, line 26: This is already in `cmake/configure.cmake`

M2/Macaulay2/e/NCAlgebras/FreeAlgebra.hpp
```diff
@@ -0,0 +1,323 @@
...
+
+#include "Polynomial.hpp"             // for Poly, Monom
+#include "NCAlgebras/FreeMonoid.hpp"  // for FreeMonoid
+#include "NCAlgebras/Word.hpp"        // for Word
+#include "newdelete.hpp"              // for our_new_delete
+#include "ring.hpp"                   // for Ring (ptr only), SumCollector
+#include "ringelem.hpp"               // for ring_elem
+#include "style.hpp"                  // for GEOHEAP_SIZE
+
+#include <gmp.h>                      // for mpz_srcptr, mpq_srcptr
```
mahrud, line 12: I must have missed this, but I think `M2/math-include.h` should be used instead.

M2/Macaulay2/e/NCAlgebras/FreeAlgebraQuotient.hpp
```diff
@@ -0,0 +1,150 @@
...
+#define _free_algebra_quotient_hpp_
+
+#include "NCAlgebras/FreeAlgebra.hpp" // for FreeAlgebra
+#include "NCAlgebras/FreeMonoid.hpp"  // for FreeMonoid
+#include "NCAlgebras/NCGroebner.hpp"  // for NCGroebner
+#include "Polynomial.hpp"             // for Poly, ConstPolyList
+#include "newdelete.hpp"              // for our_new_delete
+#include "ringelem.hpp"               // for ring_elem
+
+#include <gmp.h>                      // for mpz_srcptr, mpq_srcptr
```
mahrud, line 11: Same here.

M2/Macaulay2/e/NCAlgebras/NCGB.m2
```diff
@@ -0,0 +1,93 @@
+restart
```
mahrud, line 1: Should this file be in the `e/` directory?

M2/Macaulay2/e/NCAlgebras/OverlapTable.cpp
```diff
@@ -0,0 +1,109 @@
+#include "NCAlgebras/OverlapTable.hpp"
+#include "engine-exports.h"  // for M2_gbTrace
```
mahrud, line 2: Include `engine-includes.hpp` instead.

DanGrayson:
Frank, try merging with the master branch first.  Then the build might work.

M2/Macaulay2/e/NCAlgebras/SuffixTree.cpp
```diff
@@ -0,0 +1,713 @@
+#include "NCAlgebras/SuffixTree.hpp"
+#include <algorithm>            // for copy
+#include <iterator>             // for back_insert_iterator, back_inserter
+#include "NCAlgebras/Word.hpp"  // for Word, operator<<
```
mahrud, line 4: This comment applies to almost all the new sources, but I didn't notice it till now:
I think we should stick to the [convention here](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) for order of inclusions. This file seems to be correct (except newlines), but all the others may need to be reordered. It's not hugely urgent though.

M2/Macaulay2/e/NCAlgebras/roos.m2
```diff
@@ -0,0 +1,52 @@
+restart
```
mahrud, line 1: Should this file be in the `e/` directory?

M2/Macaulay2/e/aring-zzp-ffpack.hpp
```diff
@@ -17,8 +17,9 @@
 
 // this fixes a weird problem in the package "openblas" of Arch Linux, which somehow fails to declare this function:
 extern "C" void openblas_set_num_threads(int num_threads);
```
mahrud, line 19: Not relevant to your changes, but seeing this reminded me that we should probably also call [`flint_set_num_threads`](http://flintlib.org/doc/threading.html) to use threading there as well.

M2/Macaulay2/e/interface/polyroots.cpp
```diff
@@ -1,7 +1,9 @@
 #include "interface/factory.h"
 #include "interface/ring.h"
 
+#define register
```
mahrud, line 4: What does this do? Perhaps it could benefit from a short comment.

M2/Macaulay2/e/matrix-ncbasis.hpp
```diff
@@ -0,0 +1,30 @@
+// TODO: should this be in the interface?
```
mahrud, line 1: Should this file be in `e/interface`?

M2/Macaulay2/e/stdinc-m2.hpp
```diff
@@ -3,6 +3,8 @@
 #endif
 #define M2_STDINC_M2_GUARD
 
+#warning TODO: remove e/stdinc-m2.hpp
```
mahrud, line 6: This file probably can go in `engine-includes.hpp`

M2/Macaulay2/e/todo-ncalgebra.m2
```diff
@@ -0,0 +1,242 @@
+19 May 2020  -- what needs to be done?
```
mahrud, line 1: The changelog here was very useful to read for me when I was trying to figure out this branch, but perhaps it should move to `NCAlgebras/README.md` or somewhere like that.

mahrud:
Try adding `libtbb-dev` here:
https://github.com/Macaulay2/M2/blob/244e8ea71ec6fe665e4741d2b1708b520173b1b7/.github/workflows/test_build.yml#L96

mahrud:
With the fix above, my build kept going, until this point:
```
CXX VectorArithmetic.cpp
../../../../Macaulay2/e/VectorArithmetic.cpp:11:10: fatal error: execution: No such file or directory
 #include <execution>
          ^~~~~~~~~~~
compilation terminated.
```

moorewf:
Sorry, we can remove <execution>.  That was from an old attempt at parallelization.

M2/CMakeLists.txt
moorewf: Fixed in https://github.com/Macaulay2/M2/pull/1861/commits/9b2791241e2a5762d6faae940ca695b841bf9a77.

M2/Macaulay2/e/CMakeLists.txt
moorewf: Fixed in https://github.com/Macaulay2/M2/pull/1861/commits/57afec66919d43b7380dcf1ec09ccb3b6ae580b4.

M2/Macaulay2/e/interface/polyroots.cpp
moorewf: Fixed in https://github.com/Macaulay2/M2/pull/1861/commits/7d57cb6e6ea50494e7771a43642b815e02a8569d.

moorewf:
@mahrud How do I add libtbb-dev to the line you reference above?

mahrud:
> @mahrud How do I add libtbb-dev to the line you reference above?

Add it to `.github/workflows/test_build...` in your fork.

DanGrayson:
Frank, it's probably best to use "git push" rather infrequently, because each one starts a 3-hour build on github's machines, as the branch is involved in an active pull request.

moorewf:
@DanGrayson Yes, I noticed that. I'll make sure they are more substantial going forward.

The autotools build crash above is a strange one.  Not sure if it is related, but [this](https://github.com/facebook/folly/issues/487) seems to indicate that something is strange with gcc and the c++14 headers.

mahrud:
Wasn't this related to a patch for givaro to work with C++17? I think we might need to force building givaro with a patch.

moorewf:
There was different issue with using C++17 with givaro, but this is a new one.  I build with apple-clang usually; I am having difficulty getting brew's clang to find libomp again, so I'm having difficulty troubleshooting this one.

mahrud:
> There was different issue with using C++17 with givaro, but this is a new one. I build with apple-clang usually; I am having difficulty getting brew's clang to find libomp again, so I'm having difficulty troubleshooting this one.

So, clang (from brew's llvm) ships with OpenMP, but you need to run ``export LIBRARY_PATH=`llvm-config --libdir` `` then call `cmake` for it to find the correct library. But Xcode doesn't ship with OpenMP, so you need to run ``cmake -DCMAKE_SYSTEM_PREFIX_PATH=`brew --prefix libomp` ``. What error do you get?

moorewf:
So... I tried fixing my LIBRARY_PATH as in our earlier conversations and it wasn't working -- turns out I was editing the command for the wrong build target :/

After adding LIBRARY_PATH=`/usr/local/opt/llvm/bin/llvm-config --libdir` to just before cmake the compilation runs to completion on my machine on brew clang.  Now I will try to get it to build with gcc.

mahrud:
I think you'd want to try building on Ubuntu 18.04 with the default gcc, which is gcc7. You can try building on Docker, from this Dockerfile, which roughly mimics what's running on the github actions:
[Dockerfile.txt](https://github.com/Macaulay2/M2/files/5860737/Dockerfile.txt)

mikestillman:
@mahrud can you remind me where all of the docker files are located?

mahrud:
https://github.com/Macaulay2/M2/tree/development/M2/BUILD/docker

mikestillman:
@mahrud never mind: I found them: in BUILD/docker

mahrud:
Be careful though, most of those Dockerfiles don't install tbb, so you'd have to edit the Dockerfiles (usually it's there, just commented). Try editing the one in [`BUILD/docker/ubuntu`](https://github.com/Macaulay2/M2/tree/development/M2/BUILD/docker/ubuntu), then running `make build`. That should be a straightforward CMake build on ubuntu.

mikestillman:
@mahrud Will do, thanks!

mahrud:
For the record, I had to edit some things in the engine to be able to compile locally. Here is the patch of my changes:
[0001-fix-build-errors-on-cmake.patch.txt](https://github.com/Macaulay2/M2/files/5860801/0001-fix-build-errors-on-cmake.patch.txt)

mahrud:
One more thing, now that tbb is required with M2, we would get the warnings in https://github.com/Macaulay2/mathicgb/issues/23 all the time. They could be silenced, but I think it's better to follow their recommendation and only include the submodules of tbb that we use, rather than `tbb.h` wholesale. This would probably speed up compilation by a tiny bit, too.

mahrud:
Does anybody mind if I push a few commits to fix the CMake build here? It looks like we might need to switch Github Actions to Ubuntu 20.04, because the TBB on Ubuntu 18.04 is too old. At least most of the builds [here](https://github.com/mahrud/M2/runs/1756744572?check_suite_focus=true) compiled okay.

moorewf:
@mahrud That is fine with me, thanks for the help!

mikestillman:
> Does anybody mind if I push a few commits to fix the CMake build here? It looks like we might need to switch Github Actions to Ubuntu 20.04, because the TBB on Ubuntu 18.04 is too old. At least most of the builds [here](https://github.com/mahrud/M2/runs/1756744572?check_suite_focus=true) compiled okay.

I guess the reason TBB on ubuntu 18.04 is now too old, is that the 18.04 version doesn't work under compilation with c++17?  Or are we using features of TBB that don't exist in that version (which seems unlikely to me)?

mikestillman:
@mahrud The build script in .github (at least on our branch) has the line:
```
brew install gmp mpfr ntl flint@2.6.3 factory@4.1.3 mpsolve glpk
```
Should we be changing that to 2.7.1 and 4.2.0?  Or is that done somewhere else, and we just haven't merged that in yet?

d-torrance:
> I guess the reason TBB on ubuntu 18.04 is now too old, is that the 18.04 version doesn't work under compilation with c++17? Or are we using features of TBB that don't exist in that version (which seems unlikely to me)?

It would be nice if we could still build in Ubuntu 18.04.  It's an LTS release with an end of life in 2028!

mikestillman:
> > I guess the reason TBB on ubuntu 18.04 is now too old, is that the 18.04 version doesn't work under compilation with c++17? Or are we using features of TBB that don't exist in that version (which seems unlikely to me)?
> 
> It would be nice if we could still build in Ubuntu 18.04. It's an LTS release with an end of life in 2028!

I agree.  Unless there is an easy fix, I see 2 solutions: 
  - Make TBB optional (but then these versions won't get the speed increase from using TBB)
  - Somehow recompile TBB (I'm less enthused about doing this).

@DanGrayson @mahrud @d-torrance @moorewf what do you all think?

d-torrance:
> * Somehow recompile TBB (I'm less enthused about doing this).

I could look into backporting a more modern version to 18.04 and adding it the Macaulay2 PPA.  That would fix the issue for both the GitHub builds (which already uses the PPA for a few of the build dependencies that aren't in 18.04) and the PPA builds.

mikestillman:
@d-torrance Perhaps this is relevant? https://askubuntu.com/questions/1170054/install-newest-tbb-thread-building-blocks-on-ubuntu-18-04

d-torrance:
> @d-torrance Perhaps this is relevant? https://askubuntu.com/questions/1170054/install-newest-tbb-thread-building-blocks-on-ubuntu-18-04

I'm not a huge fan of the suggestion given in the answer there.  By adding a more recent Ubuntu to `/etc/apt/sources.list`, we'd end up downloading a bunch of newer versions of dependencies during the build, and the resulting binary likely would depend on versions of libraries that aren't in 18.04.

d-torrance:
I was able to backport the TBB package that's currently in Debian unstable to Ubuntu 18.04 with just a couple minor changes.  It's currently building in the PPA and should be available for future pushes!

mikestillman:
@d-torrance Excellent, thanks Doug!

DanGrayson:
> > I guess the reason TBB on ubuntu 18.04 is now too old, is that the 18.04 version doesn't work under compilation with c++17? Or are we using features of TBB that don't exist in that version (which seems unlikely to me)?
> 
> It would be nice if we could still build in Ubuntu 18.04. It's an LTS release with an end of life in 2028!

I agree.  We probably have users on that system.

DanGrayson:
Why are these builds getting cancelled??

mikestillman:
I was wondering that too.  If one build fails, it seems like it is triggering the rest to be cancelled?

DanGrayson:
> I was wondering that too. If one build fails, it seems like it is triggering the rest to be cancelled?

Maybe, but it seems doubtful.

mikestillman:
thanks @mahrud for the updates!  Here is another thing, which is I believe not limited to this branch, but that is where I noticed it.  I get many warnings about ffpack and openmp suport.  The compilation runs to completion fine.  This occurs under both appleclang, and latestclang (from brew llvm) on macos catalina. 
```
[118/132] Building CXX object Macaulay2/e/CMakeFiles/M2-engine.dir/interface/groebner.cpp.o
In file included from ../../../../Macaulay2/e/interface/groebner.cpp:24:
In file included from ../../../../Macaulay2/e/NCAlgebras/NCF4.hpp:12:
In file included from ../../../../Macaulay2/e/VectorArithmetic2.hpp:7:
In file included from ../../../../Macaulay2/e/aring-glue.hpp:7:
In file included from ../../../../Macaulay2/e/aring-translate.hpp:18:
In file included from ../../../../Macaulay2/e/aring-zzp-ffpack.hpp:21:
In file included from /Users/mike/src/M2-development/M2/BUILD/mike/builds.tmp/cmake-latestclang/usr-host/include/fflas-ffpack/ffpack/ffpack.h:39:
/Users/mike/src/M2-development/M2/BUILD/mike/builds.tmp/cmake-latestclang/usr-host/include/fflas-ffpack/fflas-ffpack-config.h:102:2: warning: "openmp was not detected correctly at configure time, please report this bug" [-W#warnings]
#warning "openmp was not detected correctly at configure time, please report this bug"
 ^
1 warning generated.
```

mahrud:
I've reported that issue to fflas_ffpack, but they have yet to fix it.

mikestillman:
> I've reported that issue to fflas_ffpack (see [linbox-team/fflas-ffpack#309](https://github.com/linbox-team/fflas-ffpack/issues/309)), but they have yet to fix it.

@mahrud, Thanks!

mahrud:
Have you tried using the brew package instead of building it? All M2 dependencies are in brew now.

moorewf:
The remaining errors seem to not be specific to this pull request.  I'm not sure why some of the tests are failing.

mahrud:
Do computations over the exterior algebra use any of the added functionality? I'm wondering if the TateOnProducts example that is running out of memory on cmake-ubuntu-20.04-clang10 is related to this PR or not: [_action__On__Direct__Image.errors.txt](https://github.com/Macaulay2/M2/files/5870519/_action__On__Direct__Image.errors.txt)

mikestillman:
Exterior algebra computations should not be related to anything we have done (he says naively...!).   But seriously, I think they should not be related at all (unless by adding in tbb things changed).

mikestillman:
Or that we are now compiing with c++17.  But I don't think that should be related either.

mahrud:
Seems like `actionOnDirectImage` running out of memory is a recurring issue on Ubuntu with clang. It could be related to #1557, but someone probably should look into it.

mikestillman:
Here is the error happening under `cmake-ubuntu-20.04-clang10`.  I don't see why this would differ in behavior from e.g. `cmake-ubuntu-20.04-gcc9`, which passes all tests.  

I also don't see how the ncEngine branch is related at all.  Perhaps there is a problem using c++17, rather than c++14 (or c++17 with gcc)?  @DanGrayson @d-torrance @mahrud Any suggestions?
```
home/runner/work/M2/M2/M2/BUILD/build/usr-dist/common/share/doc/Macaulay2/Core/tests/programs.m2:0:1: error:
 -- 
 -- i6 : 
 --      program = findProgram(name, name, RaiseError => false)
 -- 
 -- o6 = 0
 -- 
 -- o6 : Program
 -- 
 -- i7 : assert(program === null)
 -- stdio:9:1:(3): error: assertion failed
 -- 
stdio:1:1:(3): error: test(s) #203 of package Core failed.

 cd /tmp/M2-151017-0/1-rundir/; GC_MAXIMUM_HEAP_SIZE=400M "/home/runner/work/M2/M2/M2/BUILD/build/usr-dist/x86_64-Linux-Ubuntu-20.04/bin/M2-binary" -q --int --no-randomize --no-readline --silent --stop --print-width 77 <"/tmp/M2-151017-0/0.m2" >>"/tmp/M2-151017-0/0.tmp"
/tmp/M2-151017-0/0.tmp:0:1: (output file) error: Macaulay2 exited with status code 1
/tmp/M2-151017-0/0.m2:0:1: (input file)
M2: *** Error 1
Error: Process completed with exit code 1.
```

mikestillman:
Looking at the code for `findProgram`, I don't even see how `findProgram` can return 0...

mahrud:
> Here is the error happening under `cmake-ubuntu-20.04-clang10`. I don't see why this would differ in behavior from e.g. `cmake-ubuntu-20.04-gcc9`, which passes all tests.

The difference is that the former runs the tests and the latter doesn't. In particular, `autotools-ubuntu-20.04-default` uses gcc9 and also runs the tests, and also fails:
```m2
testing: ../../../../../Macaulay2/tests/normal/programs.m2
../../../../../Macaulay2/tests/normal/programs.m2:8:1:(3):[4]: error: assertion failed
programs.errors:0: error output left here for the errors above:
```

d-torrance:
> Looking at the code for `findProgram`, I don't even see how `findProgram` can return 0...

It's returning a `Program` object, and the name of the program is `0`, which was the basename of the file returned by `temporaryFileName()`.

It's not supposed to successfully run a program yet, because we haven't run `fileMode` yet to make it executable.  But maybe the file already existed from the earlier `make check`, and now we're running the test again in `M2 --check 2`?

Deleting the file first, it if exists, should fix this.

d-torrance:
> > Here is the error happening under `cmake-ubuntu-20.04-clang10`. I don't see why this would differ in behavior from e.g. `cmake-ubuntu-20.04-gcc9`, which passes all tests.
> 
> The difference is that the former runs the tests and the latter doesn't. In particular, `autotools-ubuntu-20.04-default` uses gcc9 and also runs the tests, and also fails:
> 
> ```m2
> testing: ../../../../../Macaulay2/tests/normal/programs.m2
> ../../../../../Macaulay2/tests/normal/programs.m2:8:1:(3):[4]: error: assertion failed
> programs.errors:0: error output left here for the errors above:
> Command exited with non-zero status 1
> ```

Hrm, wait, this would be from the earlier `make check` run of the test.  I guess the file is created with an executable bit for some reason? I guess maybe run `fileMode` to make it non-executable.  Something like:

```diff
diff --git a/M2/Macaulay2/tests/normal/programs.m2 b/M2/Macaulay2/tests/normal/programs.m2
index 9f3c812c5..04b812aac 100644
--- a/M2/Macaulay2/tests/normal/programs.m2
+++ b/M2/Macaulay2/tests/normal/programs.m2
@@ -4,6 +4,7 @@ name = baseFilename fn
 dir = replace(name | "$", "", fn)
 programPaths#name = dir
 
+fileMode(6*64 + 4*8 + 4, fn)
 program = findProgram(name, name, RaiseError => false)
 assert(program === null)
```

mikestillman:
> I was able to backport the TBB package that's currently in Debian unstable to Ubuntu 18.04 with just a couple minor changes. It's currently building in the PPA and should be available for future pushes!

@d-torrance Doug.  Once this branch is ready to be pushed, your builds on older ubuntu's will now be fine? (I expect that once we fix that findPrograms example as you suggest, there is a chance all the tests will run to completion).

d-torrance:
> > I was able to backport the TBB package that's currently in Debian unstable to Ubuntu 18.04 with just a couple minor changes. It's currently building in the PPA and should be available for future pushes!
> 
> @d-torrance Doug. Once this branch is ready to be pushed, your builds on older ubuntu's will now be fine? (I expect that once we fix that findPrograms example as you suggest, there is a chance all the tests will run to completion).

If TBB was the only issue, then we should be fine in 18.04, yes.

mikestillman:
I'll try to add in `fileMode(6*64 + 4*8 + 4, fn)`, and check it in, to the ncEngine branch, to see if this time the tests will complete.

mikestillman:
Alas, the change to `test/programs.m2` didn't work (the same 2 builds failed (autotools, and cmake, ubuntu.  Here is the part of the .errors file.  Changing the mode didn't seem to work.  Any suggestions for what we should try next? @d-torrance @DanGrayson ? Also, now there are conflicts.  I'll merge in development again.

```
ii6 : programPaths#name = dir

oo6 = /tmp/M2-158651-0/

ii7 : 
      fileMode(6*64 + 4*8 + 4, fn)

ii8 : program = findProgram(name, name, RaiseError => false)

oo8 = 0

oo8 : Program

ii9 : assert(program === null)
../../../../../Macaulay2/tests/normal/programs.m2:9:1:(3):[4]: error: assertion failed
0.30user 0.02system 0:00.31elapsed 104%CPU (0avgtext+0avgdata 93520maxresident)k
0inputs+8outputs (0major+4721minor)pagefaults 0swaps
```

d-torrance:
> Alas, the change to `test/programs.m2` didn't work (the same 2 builds failed (autotools, and cmake, ubuntu. Here is the part of the .errors file. Changing the mode didn't seem to work. Any suggestions for what we should try next?

Hmm, interesting -- I'll clone this branch and play around with it locally.

mahrud:
I'm kind of suspicious that this doesn't have anything to do with this PR, because this error didn't occur in [this builds](https://github.com/Macaulay2/M2/runs/1771543033), which also uses gcc9 on ubuntu-20.04

mikestillman:
> I'm kind of suspicious that this doesn't have anything to do with this PR, because this error didn't occur in [this builds](https://github.com/Macaulay2/M2/runs/1771543033), which also uses gcc9 on ubuntu-20.04

@mahrud Do you mean you think it is related to this PR, or that it isn't?

mahrud:
That it does have to do with this PR. Initially I thought it had to do with transition to ubuntu-20.04, but that doesn't seem to be the case at least.

d-torrance:
Feature request for this branch:  add a check for TBB in `configure.ac`.  I didn't have it installed and `configure` didn't warn me.

mikestillman:
It should have warned you, I thought.  I added some lines to `configure.ac` where if tbb isn't found it should give an error.  But I agree, if it isn't doing that, it should.

d-torrance:
Looking back, it did check, saw that I didn't have it, and then happily continued onward without raising an error:

```
checking tbb/tbb.h usability... no
checking tbb/tbb.h presence... no
checking for tbb/tbb.h... no
checking for herror... yes
checking for error... yes
```

d-torrance:
> It should have warned you, I thought. I added some lines to `configure.ac` where if tbb isn't found it should give an error. But I agree, if it isn't doing that, it should.
> 
> (edit): Here are the relevant lines in `configure.ac`:
> 
> ```
> AC_SUBST(HAVE_LIBTBB,no)
> AC_SUBST(LIBTBB)
>    AC_LANG(C++)
>    AC_CHECK_HEADER(tbb/tbb.h,
>       [AC_SEARCH_LIBS(TBB_runtime_interface_version,"tbb",
> 	 LIBTBB=$ac_cv_search_TBB_runtime_interface_version
> 	 HAVE_LIBTBB=yes,
> 	 [AC_MSG_ERROR([tbb not found])])])
>    AC_LANG(C)
> ```

I think there needs to be a second `AC_MSG_ERROR`, if `AC_CHECK_HEADER` turns up empty.  I think it only gets to `AC_SEARCH_LIBS` (and therefore potentially `AC_MSG_ERROR`) if it finds the headers first.

mikestillman:
> Looking back, it did check, saw that I didn't have it, and then happily continued onward without raising an error:
> 
> ```
> checking tbb/tbb.h usability... no
> checking tbb/tbb.h presence... no
> checking for tbb/tbb.h... no
> checking for herror... yes
> checking for error... yes
> ```

Ahh, perhaps that code is incorrect, if tbb/tbb.h isn't found, it doesn't give an error message.  Only if that is OK, but it can't find the library is an error being given.  Sorry, my ability in autoconf stuff is limited...

d-torrance:
Hmm, getting a linking error in Ubuntu 20.10...

```
/usr/bin/ld: ../e/VectorArithmetic.o: undefined reference to symbol '_ZN3tbb13queuing_mutex11scoped_lock7releaseEv'
/usr/bin/ld: /lib/x86_64-linux-gnu/libtbb.so.2: error adding symbols: DSO missing from command line
```

d-torrance:
I got it to build finally in a 20.04 chroot, but can't reproduce the failing test.  :(

mikestillman:
Do you know why the link failed?

I've been mainly using macos, but I will build using a few ubuntu's as well...

d-torrance:
> Do you know why the link failed?
> 
> I've been mainly using macos, but I will build using a few ubuntu's as well...

20.10 was the first Ubuntu to use gcc 10 as the default compiler, but I'm wondering if maybe the 20.10 package was built with gcc 9 before the bump.  So maybe the .so file that Ubuntu ships is missing some symbols that appear if compiling something against TBB with gcc 10?  Just a theory.

d-torrance:
> > Do you know why the link failed?
> 
> 20.10 was the first Ubuntu to use gcc 10 as the default compiler, but I'm wondering if maybe the 20.10 TBB package was built with gcc 9 before the bump. So maybe the .so file that Ubuntu ships is missing some symbols that appear if compiling something against TBB with gcc 10? Just a theory.

d-torrance:
No, that's not it -- found the [build log](https://launchpadlibrarian.net/495225948/buildlog_ubuntu-groovy-amd64.tbb_2020.3-1_BUILDING.txt.gz), and the 20.10 TBB package was built with gcc 10.  Maybe some other dependency changed since then?

mahrud:
I have no idea what was causing it, but merging with development seems to have fixed it. The single remaining failure is a memory exhaustion in SpecialFanoFourfolds.

d-torrance:
> I have no idea what was causing it, but merging with development seems to have fixed it. The single remaining failure is a memory exhaustion in SpecialFanoFourfolds.

Well, that explains why I wasn't able to reproduce it!  I'd fetched the branch after the merge.

mikestillman:
> I have no idea what was causing it, but merging with development seems to have fixed it. The single remaining failure is a memory exhaustion in SpecialFanoFourfolds.

@mahrud @d-torrance @DanGrayson Is this something we need to handle in this branch.  It is also a problem in development?  Is someone planning on removing or simplifying that test?

DanGrayson:
Let's ask the author:  

@giovannistagliano -- we get the following failure:
```
 -- capturing check(2, "SpecialFanoFourfolds")                              

 *** out of memory trying to allocate 262208 bytes, exiting ***

 cd /tmp/M2-153037-0/373-rundir/; GC_MAXIMUM_HEAP_SIZE=400M "/home/runner/work/M2/M2/M2/BUILD/build/usr-dist/x86_64-Linux-Ubuntu-20.04/bin/M2-binary" -q --int --no-randomize --no-readline --silent --stop --print-width 77 <"/tmp/M2-153037-0/372.m2" >>"/tmp/M2-153037-0/372.tmp"
/tmp/M2-153037-0/372.tmp:0:1: (output file) error: Macaulay2 exited with status code 1
/tmp/M2-153037-0/372.m2:0:1: (input file)
M2: *** Error 1
```

Can we just delete that test?  Or replace it by one that uses much less memory?

mahrud:
My response was very long, so I made it a new issue: #1890.
Short version: if we remove or simplify it, we should move it to `tests/quarantine` so it can be eventually re-enabled, and can serve as a goal post for whether a memory leak has been fixed or not.

M2/Macaulay2/d/interface.dd
```diff
@@ -770,6 +770,80 @@ export rawSolvableAlgebra(e:Expr):Expr := (
...
+	            else WrongArgSmallInteger(1)
+		)
+	        else WrongArgZZ(1)
+    	    )
+	    else WrongArg(0,"a raw matrix")
+	)
+        else WrongNumArgs(2)
+    )
+    else WrongNumArgs(2)
+);
```
mahrud, line 820: I think my comment in https://github.com/Macaulay2/M2/pull/1861/files/45ecc939be495767363d9f1dc844fa54fc74163a#r562943180 was misinterpreted. The fixed indentation before ed25e7e664ebaac9bf367de0f473da55840b0142 was correct, I just meant that it should be 4 spaces rather than two. It's not very important tho.

M2/Macaulay2/e/CMakeLists.txt
```diff
@@ -128,7 +132,25 @@ set(CPPONLYFILES
   )
 
 # these files all have .hpp and .cpp files
+# MES TODO: put these NCAlgebra files (and friends) into alpha order in list
```
mahrud, line 135: Should this be done now? Fixing the indentation also would be good.

DanGrayson:
> My response was very long, so I made it a new issue: #1890.
> Short version: if we remove or simplify it, we should move it to `tests/quarantine` so it can be eventually re-enabled, and can serve as a goal post for whether a memory leak has been fixed or not.

That's good, too.  But meanwhile, Giovanni (@giovannistagliano) may wish to have an alternative test that runs every time, if quarantining that test results in incomplete coverage.

giovannistagliano:
> Let's ask the author:
> 
> @giovannistagliano -- we get the following failure:
> 
> ```
>  -- capturing check(2, "SpecialFanoFourfolds")                              
> 
>  *** out of memory trying to allocate 262208 bytes, exiting ***
> 
>  cd /tmp/M2-153037-0/373-rundir/; GC_MAXIMUM_HEAP_SIZE=400M "/home/runner/work/M2/M2/M2/BUILD/build/usr-dist/x86_64-Linux-Ubuntu-20.04/bin/M2-binary" -q --int --no-randomize --no-readline --silent --stop --print-width 77 <"/tmp/M2-153037-0/372.m2" >>"/tmp/M2-153037-0/372.tmp"
> /tmp/M2-153037-0/372.tmp:0:1: (output file) error: Macaulay2 exited with status code 1
> /tmp/M2-153037-0/372.m2:0:1: (input file)
> M2: *** Error 1
> ```
> 
> Can we just delete that test? Or replace it by one that uses much less memory?

Maybe that test is hard because it is really 21 tests together. 
To simplify it you could comment the lines: [2780](https://github.com/Macaulay2/M2/blob/a9b2a9c3af77eec2d660de1320da32b980eac518/M2/Macaulay2/packages/SpecialFanoFourfolds.m2#L2780), [2785](https://github.com/Macaulay2/M2/blob/a9b2a9c3af77eec2d660de1320da32b980eac518/M2/Macaulay2/packages/SpecialFanoFourfolds.m2#L2785), and [2786](https://github.com/Macaulay2/M2/blob/a9b2a9c3af77eec2d660de1320da32b980eac518/M2/Macaulay2/packages/SpecialFanoFourfolds.m2#L2786).

DanGrayson:
Is this pull request ready to be pulled?

giovannistagliano:
> Is this pull request ready to be pulled?

Yes, thanks.  For now I have no updates to make.

mikestillman:
> Is this pull request ready to be pulled?

I have not changed the test that is failing in SpecialFanoFourfolds.  Is that failing also on development? (cmake-ubuntu-20.04-clang10?)

I can change the test, unless someone has already been doing that.

One more thing that should be done is to fix the TBB check code to give an error if TBB isn't present.  I thought I had done that, but I missed something (I checked for the lib, gave an error message, but Doug noticed that if the header isn't found it is not giving an error message).

So: @DanGrayson should I fix these 2 things, then we can merge it?

DanGrayson:
Sure, Mike, it would be great if you would simplify the test on this branch.  On the other, it's unrelated and it will be easy enough to do later, as I made an issue at  https://github.com/Macaulay2/M2/issues/1896

mahrud:
I implemented my suggestions for this PR (mainly header management, e.g. removing stdinc-m2.h and not directly including tbb.h in order to prevent the warnings) in another PR for Frank's branch: https://github.com/moorewf/M2-frank/pull/2. It passed the same tests here: https://github.com/mahrud/M2/actions/runs/517463819
But I can wait till this is merged and open it against development instead.

mikestillman:
@mahrud, I've done some of this already in a branch that will be considered after ncEngine.  I'm sorry that we aren't coordinating our effort here...!  I'll take a look at your pull request.

mahrud:
No worries. Most of the changes (e.g. improvements to IWYU, or removing M2mem from cpp files) were actually in our old engine branch that didn't make it into that PR, so I think the intersection is probably small.

mikestillman:
I have just pushed this change, and also merged in development again.  The one thing I have not done is to fix configure.ac to error if TBB is not found (I know it is simple, but not for me (I do not really understand autotools for the most part!), but will need to sit down and do it soon).  @mahrud If TBB is not found, does the cmake version give an error right away?

Anyway, if the test builds work, this is ready for merging.

mahrud:
Yes:
https://github.com/Macaulay2/M2/blob/a9b2a9c3af77eec2d660de1320da32b980eac518/M2/cmake/check-libraries.cmake#L71

DanGrayson:
Mike, we can fix configure.ac together on Tuesday.

d-torrance:
The following patch fixes the TBB check for me:

```diff
diff --git a/M2/configure.ac b/M2/configure.ac
index 32ce61dcb..288c614fe 100644
--- a/M2/configure.ac
+++ b/M2/configure.ac
@@ -369,7 +369,8 @@ AC_SUBST(LIBTBB)
       [AC_SEARCH_LIBS(TBB_runtime_interface_version,"tbb",
         LIBTBB=$ac_cv_search_TBB_runtime_interface_version
         HAVE_LIBTBB=yes,
-        [AC_MSG_ERROR([tbb not found])])])
+        [AC_MSG_ERROR([tbb not found])])],
+      [AC_MSG_ERROR([tbb not found])])
    AC_LANG(C)
 
 # the following line remove -ltbb from LIBS (only thing in LIBS so far).  It is not needed
```

After removing the `libtbb-dev` Ubuntu package and running `configure`, I get the following, as expected:

```
checking tbb/tbb.h usability... no
checking tbb/tbb.h presence... no
checking for tbb/tbb.h... no
configure: error: tbb not found
```

mikestillman:
@DanGrayson @d-torrance Thanks for your suggestion and offer to help!

moorewf:
Huzzah!  Thanks for all the help @DanGrayson @mahrud @d-torrance and @giovannistagliano.

d-torrance:
Just started my first test build of the Debian package using the `development` branch and noticed a few things:

```
NCAlgebras/NCReduction.cpp: In function 'std::string getHeapName(HeapType)':
NCAlgebras/NCReduction.cpp:1061:1: warning: control reaches end of non-void function [-Wreturn-type]
 1061 | }
      | ^
```

There should be a `default` option for the switch statement:

https://github.com/Macaulay2/M2/blob/35a63ff88d3fc5ae252f5c11b1a49a0178ad0d5c/M2/Macaulay2/e/NCAlgebras/NCReduction.cpp#L1041-L1061

Some unused variables:

```
NCAlgebras/FreeMonoid.cpp: In constructor 'FreeMonoid::FreeMonoid(const std::vector<std::__cxx11::basic_string<char> >&, const PolynomialRing*, const std::vector<int>&, const std::vector<int>&, const std::vector<int>&)':
NCAlgebras/FreeMonoid.cpp:37:8: warning: unused variable 'nvars' [-Wunused-variable]
   37 |   auto nvars = variableNames.size();
      |        ^~~~~
NCAlgebras/FreeMonoid.cpp: In member function 'void FreeMonoid::mult3(const Monom&, const Monom&, const Monom&, FreeMonoid::MonomialInserter&) const':
NCAlgebras/FreeMonoid.cpp:92:7: warning: unused variable 'sz' [-Wunused-variable]
   92 |   int sz = m1[0] + wordLength(m2) + wordLength(m3);
      |       ^~
```
Not sure what's going on here...
```
NCAlgebras/NCGroebner.cpp:249:49: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
  249 |   return reinterpret_cast<const ConstPolyList&>(mGroebner);
      |                                                 ^~~~~~~~~

...

./NCAlgebras/NCF4.hpp: In member function 'const ConstPolyList& NCF4::currentValue() const':
./NCAlgebras/NCF4.hpp:153:51: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
  153 |     return reinterpret_cast<const ConstPolyList&>(mGroebner);
      |                                                   ^~~~~~~~~

...

In file included from interface/groebner.cpp:24:
./NCAlgebras/NCF4.hpp: In member function 'const ConstPolyList& NCF4::currentValue() const':
./NCAlgebras/NCF4.hpp:153:51: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
  153 |     return reinterpret_cast<const ConstPolyList&>(mGroebner);
      |                                                   ^~~~~~~~~
```

moorewf:
After looking into it, it seems the aliasing warning is quite bad - making such a cast is undefined behavior in the standard, so we should either make a copy or rework the data types to be PolyLists.

PR #1862, opened 2021-01-22
Development

DanGrayson:


mahrud:
When the tests pass, instead of clicking "Merge pull request", could you do this instead?
image

Ideally this should be done on all PRs, so that commits from one PR appear next to each other, rather than being mingled with commits from other PRs based on the time of commit.

DanGrayson:
No, because that renders the user's branch unusable.

mahrud:
It doesn't affect the user's branch as far as I know.

DanGrayson:
> It doesn't affect the user's branch as far as I know.

It renders the user's branch unusable, in the sense that it contains commits that did not get into the upstream repository.  They'll have to abandon that branch, because another pull request from the same branch will still appear to have those old commits as new content, at least on the list of pull request commits (not on the diff).

mahrud:
That's not true. They can add commits to the branch and make a new PR and you will just rebase when it is accepted again. Try it if you want.

Issue #1672, opened 2020-12-08
Bug in package RationalPoints


jieaosong:
There is a bug in the package RationalPoints, which can for example be reproduced by the following code
```macaulay2
loadPackage "RationalPoints"
I = Grassmannian(1,3,CoefficientRing=>ZZ/2)
rationalPoints I
```
The error reads
```
error: array index 2 out of bounds 0 .. 1
```
I think it's because in the following line, the created lists are not flattened
https://github.com/Macaulay2/M2/blob/8ecbb08b154be4e1a39e83faf8576112311e322d/M2/Macaulay2/packages/RationalPoints.m2#L175
which should probably be corrected to
```macaulay2
newstuff := apply(toList (set els)^**n, deepSplice);
```

8d1h:
I tried to read the source code of the package which is rather cryptic for me... (though I'm admittedly a beginner)

Thus I wrote a rather straightforward version which naively test all the points, coordinate by coordinate, until one encounters a single variable polynomial which we then factorize and only enumerate its roots instead of the entire finite field *k*.

Surprisingly this runs quite well: even faster than the original version in some cases, when the later doesn't report bug...

So how can I submit my code?

DanGrayson:
Are you thinking of modifying the package?  If so, perhaps first contact the author(s).

For submitting changes, see https://github.com/Macaulay2/M2/wiki/Git-for-Macaulay2-Contributors

mahrud:
@DanGrayson the new package entirely supersedes the old package. I think we should retire the old one and name this just `RationalPoints`. I don't see a reason to keep distributing a 12 year old package that doesn't seem to be maintained anymore when there's a better alternative.

DanGrayson:
I'll turn your comment into an issue -- it's worth discussing.

d-torrance:
> **Written by Claude** (Claude Opus 5, via Claude Code), not by @d-torrance, whose account posted it -- please weigh it accordingly.

Closing this: it was fixed by the reporter, five years ago, and nobody came back to say so.

The diagnosis in the body was right, and @jieaosong applied it themselves in
[#1989](https://github.com/Macaulay2/M2/pull/1989) —
[`4b2aa89ff4`](https://github.com/Macaulay2/M2/commit/4b2aa89ff4), merged 2021-03-11 — turning
`RationalPoints.m2:175` into

```m2
newstuff := apply(toList (set els)^**n, deepSplice);
```

[...]

PR #1863, opened 2021-01-23
New RationalPoints package

8d1h:
The package `RationalPoints` no longer works properly for some ideals (see #1672). Also the algorithm is very basic and runs rather slow once the characteristic/dimension gets big. For example it won't be able to find the points of a zero dimensional ideal in some large characteristic/dimension affine space.

Therefore I've written a new version from scratch, with some improvements in performance. It is now relatively easy to compute the rational points for a zero dimensional variety.
**Update**: I added the algorithm by Frank Sottile in the Macaulay2 book for finding roots of 0-dimensional varieties using elimination. This significantly improved the overall performance.
**Update2**: Elimination works also over infinite fields, so the method `rationalPoints` now works for 0-dimensional ideals over rational numbers. And it is possible to specify the coefficient field.

Here are some test cases.
```macaulay2
needsPackage \ ("Points", "RationalPoints2");
R = ZZ/101[x_0..x_9];
M = randomPointsMat(R, 20);
time rationalPoints variety points M;
```
This can finish in around 0.2 second (tested on the web version). Also I expanded the interface so if a variety is plugged in, it will automatically calculate the homogeneous coordinates in the projective space (which can also be toggled using the option `Projective`, i.e. `rationalPoints(I, Projective=>true)`.

For general varieties, the performance is also improved.
```macaulay2
needsPackage \ ("RandomSpaceCurves", "RationalPoints2");
R = ZZ/101[x_0..x_3];
I = (random spaceCurve)(4, 1, R);
time rationalPoints variety I;
```
This can finish in less than 1 second.

Now we can do the following over number fields.
```macaulay2
R = QQ[a,b,c,d];
I = ideal(a^3+b^3+c^3+d^3);
#rationalPoints variety Fano_1 I -- 3 -- used 0.182874 seconds
F = toField(QQ[w]/(w^2+w+1));
#rationalPoints_F variety Fano_1 I -- 27 -- used 0.566278 seconds
```
We can use `rationalPoints(F, I)` to specify the coefficient field `F` for finding `F`-rational points.

Obviously this is not on par with Magma for example. But at least it is usable now in some small cases, and hopefully will get more powerful in the future.

And since this is not just a modification of the old package, I renamed it to `RationalPoints2` to avoid replacing it for now.


8d1h:
I messed up my local branch (shouldn't have used master...) so I created a new pull request #1911. It's ready for review.

PR #1864, opened 2021-01-23
some typos (not all of them) in d/ and e/

fchapoton:
some more typos in the folders d/ and e/

there remains a few ones


DanGrayson:
Thanks!

Issue #1871, opened 2021-01-24
"displayGraph" not working for digraphs


d-torrance:
`displayGraph` is defined for digraphs, but it calls `writeDotFile`, but this only works for graphs:

```m2
i14 : needsPackage "Graphs"

o14 = Graphs

o14 : Package

i15 : D = digraph {(0, 1)}

o15 = Digraph{0 => {1}}
              1 => {}

o15 : Digraph

i16 : displayGraph D
stdio:16:1:(3): error: no method found for applying writeDotFile to:
     argument 1 :  "/tmp/M2-2729721-0/4.dot" (of class String)
     argument 2 :  Digraph{0 => {1}} (of class Digraph)
                           1 => {}

i17 : methods writeDotFile

o17 = {0 => (writeDotFile, String, Graph)}

o17 : NumberedVerticalList
```

PR #1872, opened 2021-01-24
Add digraph support to "writeDotFile" in Graphs package

d-torrance:
As a consequence, "displayGraph" and "showTikZ" also now support
digraphs.

Closes: #1871


pzinn:
interesting. I had already implemented something similar on the version of M2 that runs on the web server, see
https://www.unimelb-macaulay2.cloud.edu.au/?user=severian666#tutorial-0-5
(and replace `graph` -> `digraph`)
testing your version now

pzinn:
Interesting. Your version produces a different result for ordinary graphs.
my version:
graphs1
your version:
graphs2

pzinn:
incidentally there's an extra bit I wrote which you might as well add (defines the `html` output):
```
html Digraph := G -> if G.cache#?"svg" then G.cache#"svg" else (
     dotfilename := temporaryFileName() | ".dot";
     writeDotFile(dotfilename, G);
     svgfilename := temporaryFileName() | ".svg";
     runcmd(graphs'DotBinary  | " -Tsvg " | dotfilename | " -o " | svgfilename);
     G.cache#"svg" = get svgfilename
     )
```

d-torrance:
> Interesting. Your version produces a different result for ordinary graphs.

I think this is because edges are represented as sets for undirected graphs, and since I switched to using `edges` instead of dealing with the adjacency matrix (so it could handle both graphs and digraphs and to avoid duplicating code -- the adjacency matrix stuff is already in `edges` anyway), the order is arbitrary:

```m2
i10 : G = graph {{a,b}, {b,c}, {b,d}, {d,e}};

i11 : edges G

o11 = {set {a, b}, set {b, c}, set {b, d}, set {d, e}}

o11 : List

i12 : toList \ oo

o12 = {{a, b}, {c, b}, {d, b}, {d, e}}

```

graphviz must handle things a bit differently when it sees something like `d -- b` instead of `b -- d`.



> incidentally there's an extra bit I wrote which you might as well add (defines the `html` output):

Cool!  I've pushed this (plus some documentation for it).

pzinn:
slightly off-topic, but has anyone managed to make the package `Visualize` work in my web interface?
in principle it would be pretty natural since we're already in the browser. Don't know much about how it works though.

PR #1874, opened 2021-01-24
some more typos in e/

fchapoton:


mahrud:
There are currently two large pull requests (#1465 and #1861) with significant changes in `e/` and `d/`. It would be better to hold off on correcting typos in these two directories in order to avoid unnecessary merge conflicts. Also, checking for typos after those two are merged might save you time as well. Thanks!

fchapoton:
ok, then

DanGrayson:
#1861 has been merged, but #1465 is still just a draft.   I think I'll merge now and let the authors of #1465 take care of any conflicts, which will be minor.  (This pull request consists of just spelling corrections.)

PR #1887, opened 2021-01-27
master <= development

mahrud:
Currently the last tag is at `1.17.1`. I made this PR so the checks run for the purpose of tagging `1.17.1.1` or `1.17.2` and bumping up the version.

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -17,7 +17,6 @@ on:
     branches:
       - master
       - development
-      - release-1.17
   schedule:
     # cron time in UTC
     # a scheduled job runs on the default branch: 'master'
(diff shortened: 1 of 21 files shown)


DanGrayson:
Should we ignore the failure in SpecialFanoFourfolds ?

DanGrayson:
Just to confirm -- you'd like 1.17.2 to be based on commit b6a0555dc3779163feac0e7052a2d23f3e734b3e, the current commit at the head of the development branch ?

I'll suspend merging pull requests for the moment while we take care of that.

mahrud:
> Should we ignore the failure in SpecialFanoFourfolds ?

Up to you, but given that it seems to be limited, I'm fine to ignore it for the moment.

> Just to confirm -- you'd like 1.17.2 to be based on commit b6a0555, the current commit at the head of the development branch ?

That seems reasonable to me. By the way, if you prefer not to build a new binary distribution, 1.17.1.1 is sufficient for a new brew build.

mahrud:
1.17.1.1 is actually probably better, so you can continue merging PRs as soon as this PR is merged and a tag is created.

DanGrayson:
I like 1.17.2, because it's intended to appear in at least one distribution, the brew distribution.  I think of numbers with 4 parts as labels for versions under ongoing development.  Then after I make the tag, I'll bump it immediately to 1.17.2.1.  Are you okay with that?

d-torrance:
> I like 1.17.2, because it's intended to appear in at least one distribution, the brew distribution.

I can release PPA builds for it too.  (I think I mentioned this elsewhere, but I'm going to hold off trying to get it into Debian/Ubuntu since we're so close to the Debian 11 freeze.)

mahrud:
Sure.

DanGrayson:
Tagged, see https://github.com/Macaulay2/M2/releases/tag/release-1.17.2

DanGrayson:
Hmm, merging the pull request on github did not do a fast-forward, so now we have a merge commit on master that is not present on development.  That's a reason not to use this mechanism for updating master.  I think I'll stick with merging at home.

Issue #1186, opened 2020-05-22
Definition of some unicode symbols


pzinn:
I'm about to do a PR and found this in my code:
```
export { "ℚ","ℝ","ℤ","ℂ","∞" }
ℚ=QQ
ℝ=RR
ℤ=ZZ
ℂ=CC
∞=infinity
```
This wasn't meant to be shipped, so a priori, I'll remove it.
But I was wondering, do people think it's a good idea?

DanGrayson:
That would be fine, but let's make them synonyms, rather than separate symbols, as here:
```
export {
	-- synonyms:
        "ann" => "annihilator",
        "ass" => "associatedPrimes",
        "coker" => "cokernel",
        "det" => "determinant",
        "gens" => "generators",
        "ker" => "kernel",
        "minPres" => "minimalPresentation",
        "numcols" => "numColumns",
        "numrows" => "numRows",
        "res" => "resolution",
        "sub" => "substitute",
        "top" => "topComponents"
}
```

Hmm, how can the user control which one is displayed on the screen when printing?  We probably don't want the unicode to be displayed by default, but it would be great to have the option available.

There are also various unicode operators we could consider introducing as synonyms for existing ones, such as `⊗` for `**`.

pzinn:
Ah yes, synonyms of course.
Because of #1069 it is indeed not a good idea to use the unicode by default.
Leaving this aside for now, will get back to it when I look into #1069.

pzinn:
Aha! as to ⊗ for **, I tried but failed.
I had indeed added as well
```
	"⊗" => "**"
```
but then `x ⊗ x` will work but not `x⊗x` ...

DanGrayson:
[...]

Issue #1069, opened 2020-01-29
utf8 characters in M2 and Emacs


mahrud:
Is there any interest in using utf8 (or even unicode) characters more in Emacs? Recent versions of Emacs as well as most terminal emulators support unicode (not sure since which version). M2 itself also supports utf8 characters for variable names:
```
S = QQ[θ]; -- in Emacs use C-x 8 [Enter] 03B8 [Enter] to type θ
gens S
θ^2
S_0^2
```

Entering them is not super easy, but for instance `S_0` is a an easy shortcut. I don't think allowing `\theta` as variable name would be possible.

As an example of an alternative way, some Sage functions have a [`latex_name`](https://doc.sagemath.org/html/en/reference/calculus/sage/calculus/var.html) option, which just tells Sage to print the LaTeX name of the variable or function when using `latex(variable)`. Example:
```
sage: function('riemann', latex_name="\\mathcal{R}")
riemann
sage: latex(riemann(x))
\mathcal{R}\left(x\right)
```

We could have a similar option so that on viewers supporting utf8 we can print the utf8 character and on viewers supporting LaTeX, for instance in the documentation or the interactive shell, use the LaTeX code and let MathJax or KaTeX do the conversion.

This is somewhat related to #522

DanGrayson:
The Agda input method, available as part of the agda package in homebrew, allows you to type `\theta` to get `θ`.  If I use that with M2 I can easily use that in commands:
```

i1 : R = QQ[θ]

o1 = R

o1 : PolynomialRing

i2 : θ^6

       6
o2 = θ

o2 : R

i3 : θ + 1

o3 = θ + 1

o3 : R
```

The standard emacs function `insert-char`, available on the key sequence `C-x 8 RET`, allows you to type the unicode name of any unicode character to get it.  These are the ones involving "theta" in the name:
```
GREEK CAPITAL LETTER THETA (Θ)
GREEK CAPITAL THETA SYMBOL (ϴ)
GREEK SMALL LETTER SCRIPT THETA (ϑ)
GREEK SMALL LETTER THETA (θ)
GREEK THETA SYMBOL (ϑ)
MATHEMATICAL BOLD CAPITAL THETA (𝚯)
MATHEMATICAL BOLD CAPITAL THETA SYMBOL (𝚹)
MATHEMATICAL BOLD ITALIC CAPITAL THETA (𝜣)
[...]

PR #1897, opened 2021-01-29
Iterating over strings

d-torrance:
One nice thing in languages like Python is the ability to iterate over strings.  After all, they're just arrays of characters!

I've started implementing some string iteration features in the D language.  In particular, `toList`, `toSequence`, `for` loops, and `scan`.

Still to do:
- [x] `apply` (this is trivial if we want the output to be a `Sequence`, but a little trickier if we want the output to be another `String`)
- [ ] any other methods that might make sense?
- [x] documentation
- [x] unit testing

Opening a draft PR for comments in the meantime.

Looking forward into the future, it would be cool if we could define arbitrary types to be iterable so these things would work out of the box.

### DIFF
--- a/M2/Macaulay2/d/actors2.dd
+++ b/M2/Macaulay2/d/actors2.dd
@@ -62,7 +62,8 @@ toList(e:Expr):Expr := (
 	       	    List(listClass, if b.Mutable then copy(b.v) else b.v,
 		    	 0, false),
 	       	    false)))
-     else WrongArg("list, sequence, or set"));
+     is s:stringCell do list(strtoseq(s))
+     else WrongArg("list, sequence, set, or string"));
 setupfun("toList",toList);
 values(e:Expr):Expr := (
      when e
--- a/M2/Macaulay2/d/actors3.d
+++ b/M2/Macaulay2/d/actors3.d
@@ -2029,7 +2029,8 @@ scan(e:Expr,f:Expr):Expr := (
 	  if !isInt(i)
 	  then WrongArgSmallInteger(1)
 	  else scan(toInt(i),f))
-     else buildErrorPacket("scan expects a list"));
+     is s:stringCell do scan(strtoseq(s), f)
+     else buildErrorPacket("scan expects a list, sequence, integer, or string"));
 scan(e:Expr):Expr := (
      when e is a:Sequence do (
 	  if length(a) == 2
--- a/M2/Macaulay2/d/actors3.d
+++ b/M2/Macaulay2/d/actors3.d
@@ -2057,7 +2057,8 @@ toSequence(e:Expr):Expr := (
 	  then Expr(new Sequence len length(b.v) do foreach i in b.v do provide i)
 	  else Expr(b.v)
 	  )
-     else WrongArg("a list or sequence"));
+     is s:stringCell do Expr(strtoseq(s))
+     else WrongArg("a list, sequence, or string"));
 setupfun("toSequence",toSequence);
 
 sequencefun(e:Expr):Expr := (
--- a/M2/Macaulay2/d/actors3.d
+++ b/M2/Macaulay2/d/actors3.d
@@ -1554,7 +1554,8 @@ map(e:Expr,f:Expr):Expr := (
 	  if !isInt(i)
 	  then WrongArgSmallInteger()
 	  else map(toInt(i),f))
-     else WrongArg(1,"a list, sequence, or an integer"));
+     is s:stringCell do map(strtoseq(s), f)
+     else WrongArg(1,"a list, sequence, integer, or string"));
 map(e1:Expr,e2:Expr,f:Expr):Expr := (
      when e1
      is a1:Sequence do (
@@ -1571,7 +1572,8 @@ map(e1:Expr,e2:Expr,f:Expr):Expr := (
 	       is v:Sequence do list(b2.Class,v,b2.Mutable)
 	       else nullE		  -- will not happen
 	       )
-	  else WrongArg(2,"a list or sequence"))
+	  is s2:stringCell do map(a1, strtoseq(s2), f)
+	  else WrongArg(2,"a list, sequence, or string"))
      is b1:List do (
 	  when e2
 	  is a2:Sequence do (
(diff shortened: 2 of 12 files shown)


d-torrance:
A quick example:


```m2
i1 : toList "foo"

o1 = {f, o, o}

o1 : List

i2 : toSequence "foo"

o2 = (f, o, o)

o2 : Sequence

i3 : for c in "foo" do print c
f
o
o

i4 : scan("foo", print)
f
o
o
```

mahrud:
Have you seen `ascii`?
```m2
i1 : for i in ascii "foo" do print ascii i
f
o
o

i2 : scan(ascii "foo", print @@ ascii)
f
o
o
```

Here's a cool feature to have, though: the disadvantage here is that `ascii` pre-allocates the memory for an integer array, but a cool feature of more developed languages like Python is that they have iterators. It would be a very cool feature if M2 supported something like the `yield` keyword in Python, so you could have something like `for c in nextChar str` or `for p in nextPrime N`. This would require the interpreter to make a new call to that function on every cycle to get the next character or prime.

DanGrayson:
Doug, I wonder whether your iteration will support unicode with utf8 encoding.  Mahrud has mentioned "ascii" -- there is also "utf8":
```m2

i14 : for i in utf8 "我是美国人。" do print i
25105
26159
32654
22269
20154
12290

i15 : for i in utf8 "我是美国人。" do print utf8 i
我
是
美
国
人
。
```

d-torrance:
> PS: maybe this is what you're suggesting as well, in which case sorry for the explanation! I thought I'd mention it since `strtoseq` in the current draft also pre-allocates memory.

Yes, that's what I was thinking, too.  No worries!

> Doug, I wonder whether your iteration will support unicode with utf8 encoding. Mahrud has mentioned "ascii" -- there is also "utf8":

It doesn't yet.  :(

```m2
i1 : toList "我是美国人。"

o1 = {�, �, �, �, �, �, �, �, �, �, �, �, �, �, �, �, �, �}

o1 : List
```
For that matter, neither does subscripting:

```m2
i2 : "我是美国人。"_0

o2 = �
```

DanGrayson:
Maybe we should have various subtypes of String: UnicodeString and AsciiString.

d-torrance:
Multiple types sound confusing to me...  Since ASCII is a subset of Unicode, wouldn't just the latter suffice?

`length` also doesn't return the expected thing for Unicode strings:

```m2
i3 : #"我是美国人。"

o3 = 18
```

DanGrayson:
> Multiple types sound confusing to me... Since ASCII is a subset of Unicode, wouldn't just the latter suffice?

You may actually wish to know whether a string consists just of ascii characters.  Your remark is useful, however,
for it suggests that AsciiString should be a subtype of UnicodeString.

You remark also suggests sticking with just String and regarding all strings as sequences of unicode characters.  However, there might be a use for sequences of bytes.  Hmm, maybe not.

> 
> `length` also doesn't return the expected thing for Unicode strings:
> 
> ```m2
> i3 : #"我是美国人。"
> 
> o3 = 18
> ```

Fixing that would be possible if UnicodeString were introduced.

mahrud:
There are two stagnant issues regarding unicode: #1069, #1186
I'm certain that focusing on having allocation-less iterators would be far more useful to virtually all M2 users than supporting Chinese characters.

pzinn:
maybe not chinese characters but I do like using greek letters for my ring variables.

mahrud:
> maybe not chinese characters but I do like using greek letters for my ring variables.

You can already do that.

Seems like the idea I described above is not the same as Doug's intention for this PR.

pzinn:
> > maybe not chinese characters but I do like using greek letters for my ring variables.
> 
> You can already do that.
> 
> Seems like the idea I described above is not the same as Doug's intention for this PR.

you can, but as mentioned in https://github.com/Macaulay2/M2/issues/1069, formatting will be incorrect. In Standard of course -- in WebApp mode everything looks perfect...

mahrud:
I didn't mean to sidetrack the point of this PR. I mentioned those issues so this discussion of unicode characters can move there, not the other way around!

d-torrance:
> Seems like the idea I described above is not the same as Doug's intention for this PR.

Initially, not for this PR, no.  Adding support for iterables does seem like the best way to make these features work for unicode strings, though, so I've been considering starting over from scratch.

But at this point, I think I'm going to keep it essentially the way it is, for consistency with the current behavior of subscripting and `length` with strings, and how `for`, `scan`, and `apply` are currently implemented (they need sequences).

But looking forward, whenever #1904 is implemented, then I would absolutely like to see string iteration use that instead, possibly using [CharacterIterator](https://unicode-org.github.io/icu/userguide/strings/characteriterator.html) from ICU.

d-torrance:
I realized that it made no sense for `apply` to return a string when it had a string argument as there's no guarantee that the function we're applying will returns characters/strings that could be glued back together into a string!  So it returns a sequence.

I think this is in pretty good shape now, so I've removed the draft status.

mahrud:
Is there anything that isn't accomplished by inserting `ascii` or `utf8` by the user? If not, I feel like you could simply mention in the documentation that strings can be turned into a list using those functions.

d-torrance:
> Is there anything that isn't accomplished by inserting `ascii` or `utf8` by the user? If not, I feel like you could simply mention in the documentation that strings can be turned into a list using those functions.

It's true that those accomplish the task, but it seems unnecessarily complicated since we need to call them twice -- once to convert each character to an integer and again to convert back -- `ascii \ ascii "foo"` vs. just `toList "foo"`.

mahrud:
You can define only `toList String := List => s -> utf8 \ utf8 s`, then the rest would follow by adding `toList`.

d-torrance:
> You can define only `toList String := List => s -> utf8 \ utf8 s`, then the rest would follow by adding `toList`.

That doesn't work since `toList` is defined in the interpreter:

```m2
i1 : toList String := List => s -> utf8 \ utf8 s

o1 = List => -*Function[stdio:1:28-1:43]*-

o1 : Option

i2 : toList "foo"
stdio:2:1:(3): error: expected list, sequence, or set

```

And even if it did, we'd still need to include `toList` in `for`, `scan`, and `apply`, which is what I was trying to avoid in the first place.

mahrud:
Ah, I forgot that it's not an actual method. I'm only skeptical that the improvement of saving the user the trouble of typing `toList` (assuming it is defined in the interpreter) is worth the added complexity. In particular, the proposed solution is strictly less time & memory efficient than using an index to loop over a string. That's just my preference though, so feel free to ignore.

d-torrance:
I agree that the implementation isn't ideal, but the current code for `for`, `scan`, and `apply` code all expect sequences, which are tagged arrays.  Strings aren't tagged array, though, so copying them over into a new sequence seemed to be the simplest way to add the feature without rewriting a bunch of other things.  (Which should definitely be rewritten in the long term for iterators!)

DanGrayson:
If scanning over strings is useful, then modifying "for" to accept strings would be useful, too.

d-torrance:
> If scanning over strings is useful, then modifying "for" to accept strings would be useful, too.

Yup!  That's https://github.com/Macaulay2/M2/pull/1897/commits/c38b45bfcd070c356515b5b79b88d446b89f82a5

PR #1907, opened 2021-01-31
"infoHelp" updates

d-torrance:
We more closely mimic the behavior of `viewHelp`.  In particular:

* Add a new `(infoHelp, DocumentTag)` method that ends up doing most of the work.
* Generate info pages on the fly for unknown documentation.

### Before
```
i1 : infoHelp foo
stdio:1:1:(3): error: no method found for applying format to:
     argument   :  foo (of class Symbol)
```
(We actually raised an error while creating our error message, since `infoHelp`'s error message was trying to call `format` on a `Thing`!)

### After
```m2
i1 : infoHelp foo
```
brings up:

```info
Up: (Macaulay2Doc)Top

foo
***

For the programmer
==================

The object foo (missing documentation) is a *note symbol: (Macaulay2Doc)Symbol,\
.












-----Info: (foo)Top, 10 lines --All---------------------------------------------
Welcome to Info version 6.7.  Type H for help, h for tutorial.

```

### DIFF
--- a/M2/Macaulay2/m2/help.m2
+++ b/M2/Macaulay2/m2/help.m2
@@ -455,15 +455,17 @@ setAttribute(viewHelp#0, ReverseDictionary, symbol viewHelp)
 
 infoHelp = method(Dispatch => Thing)
 infoHelp Thing := key -> (
-    if key === () then return infoHelp "Macaulay2Doc";
-    rawdoc := fetchAnyRawDocumentation makeDocumentTag(key, Package => null);
-    if (tag := getOption(rawdoc, symbol DocumentTag)) =!= null
+    if key === () then infoHelp "Macaulay2Doc"
+    else infoHelp makeDocumentTag key)
+infoHelp DocumentTag := tag -> (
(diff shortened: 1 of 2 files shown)


d-torrance:
Hmm, this is broken in Emacs right now in certain situations:

```m2
i4 : infoHelp "foo (bar)"
-* infoHelp: /tmp/M2-3662639-0/foo (bar).info *-
```

This gives:
```
error in process filter: user-error: No such node or anchor: /tmp/M2-3662639-0/foo (bar)
error in process filter: No such node or anchor: /tmp/M2-3662639-0/foo (bar)
```
I'll fix this later to use the same file naming convention we use for html files.

I suppose we could also use the default integer name given by `temporaryFileName`, but then the node would be displayed as, say, `(0)Top` instead of `(foo)Top`.

d-torrance:
Force-pushed a new version that uses `toFilename` when creating the temporary .info file.  It also checks if the file exists already -- no sense in re-creating it a second time.

PR #1908, opened 2021-02-01
Don't bother checking if temporary info file already exists

d-torrance:
Its contents may have changed, so we create a new version each time.

### DIFF
--- a/M2/Macaulay2/m2/help.m2
+++ b/M2/Macaulay2/m2/help.m2
@@ -455,7 +455,7 @@ setAttribute(viewHelp#0, ReverseDictionary, symbol viewHelp)
 
 makeInfo := tag -> (
     infoFile := temporaryDirectory() | toFilename format tag | ".info";
-    if not fileExists infoFile then infoFile << "\037" << endl <<
+    infoFile << "\037" << endl <<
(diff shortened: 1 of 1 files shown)


d-torrance:
I realized after #1907 was merged that my idea to only generate each temporary .info file once wasn't great, since the contents might change in between calls to `infoHelp`.  For example:

```m2
i1 : infoHelp f
```
brings up:

```
Node: Top, Up: (Macaulay2Doc)Top

f 
**

For the programmer
==================

The object f (missing documentation) is a *note symbol: (Macaulay2Doc)Symbol,.













-----Info: (_f)Top, 9 lines --All-----------------------------------------------
Welcome to Info version 6.7.  Type H for help, h for tutorial.

```

and then:

```m2
i2 : f = x -> x^2

o2 = f

o2 : FunctionClosure

i3 : infoHelp f
```
brings up:

```
Node: Top, Up: (Macaulay2Doc)Top

f 
**

For the programmer
==================

The object f (missing documentation) is a *note function closure:
(Macaulay2Doc)FunctionClosure,.












-----Info: (_f)Top, 10 lines --All----------------------------------------------
Welcome to Info version 6.7.  Type H for help, h for tutorial.

```

Previously, we still would have gone the old, out-of-date symbol version of the documentation.

Issue #1899, opened 2021-01-30
Feature request: singularLocus efficiency


jieaosong:
The method `singularLocus` in M2 is not very efficient. I recently looked at the file [`bir.m2`](http://goo.gl/eT4rCR) written by Giovanni Staglianò for his [article](http://dx.doi.org/10.1016/j.jsc.2015.11.004), where one gets the option to use `Singular` to carry out this computation. For example, here is a test case.
```macaulay2
R = ZZ/11[x_0..x_3];
I = ideal apply(10, i->random_3 R);
elapsedTime isSmooth(I, Use=>Singular);   -- 0.0754029 seconds elapsed
elapsedTime isSmooth(I, Use=>M2);         -- 3.26353 seconds elapsed
```
Is it possible to make this an official feature, which will make `singularLocus` much more usable?
**Update**: The above example is a bit degenerate. Here is a better one.
```macaulay2
I = Grassmannian(1,4,CoefficientRing=>ZZ/101);
R = ring I;
J = I + ideal(random_1 R, random_2 R);
elapsedTime isSmooth(J, Use=>Singular);  -- 2.72101 seconds elapsed
elapsedTime isSmooth(J, Use=>M2);        -- 18.1278 seconds elapsed
```

DanGrayson:
No, we can't require Singular as a prerequisite for Macaulay2, but we could try to devise a faster algorithm.

8d1h:
That's very reasonable.

Actually I looked at the code for `singularLocus`, the most compute-heavy parts are the call of `minors` and the construction of a quotient ring. Both seem to be rather basic methods...

Also I noticed that by changing the strategy of `minors` to `Cofactor` seems to improve for this particular case (and several others). Though I'm not sure that this is always the case.
```macaulay2
singularLocus(Ring) := QuotientRing => (R) -> (
     f := presentation R;
     A := ring f;
     c := codim(R,Generic=>true);
     A / trim (ideal f + minors(c, jacobian f, Strategy=>Cofactor)))
```

DanGrayson:
Notice also that forming a quotient ring involves computing a Groebner basis, which is perhaps not needed if all you want to do is to determine whether the Jacobian ideal is the unit ideal, especially in the homogeneous case.  I'm not sure whether the code for `I == 1` is optimized in that sense yet.

DanGrayson:
Anyway, if you come up with changes that are improvements, feel free to submit a pull request.

8d1h:
[...]

PR #1909, opened 2021-02-02
Minor edits related to singularLocus

8d1h:
Related to #1899 
- Changed the `minors` strategy to `Cofactor` in `singularLocus`;
- Removed one useless test in ringmap.m2 (`promote` cannot be used this way so it will always fail) that can slow down computation;
- Added `cacheValue` to several properties.

### DIFF
--- a/M2/Macaulay2/m2/matrix1.m2
+++ b/M2/Macaulay2/m2/matrix1.m2
@@ -491,11 +491,11 @@ Ideal#{Standard,AfterPrint} = Ideal#{Standard,AfterNoPrint} = (I) -> (
 Ideal ^ ZZ := Ideal => (I,n) -> ideal symmetricPower(n,generators I)
 Ideal * Ideal := Ideal => ((I,J) -> ideal flatten (generators I ** generators J)) @@ samering
 Ideal * Module := Module => ((I,M) -> subquotient (generators I ** generators M, relations M)) @@ samering
-dim Ideal := I -> dim cokernel generators I
+dim Ideal := (cacheValue symbol dim) (I -> dim cokernel generators I)
 Ideal + Ideal := Ideal => ((I,J) -> ideal (generators I | generators J)) @@ tosamering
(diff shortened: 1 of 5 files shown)


M2/Macaulay2/m2/matrix1.m2
```diff
@@ -491,11 +491,11 @@ Ideal#{Standard,AfterPrint} = Ideal#{Standard,AfterNoPrint} = (I) -> (
 Ideal ^ ZZ := Ideal => (I,n) -> ideal symmetricPower(n,generators I)
 Ideal * Ideal := Ideal => ((I,J) -> ideal flatten (generators I ** generators J)) @@ samering
 Ideal * Module := Module => ((I,M) -> subquotient (generators I ** generators M, relations M)) @@ samering
-dim Ideal := I -> dim cokernel generators I
+dim Ideal := (cacheValue symbol dim) (I -> dim cokernel generators I)
```
mahrud, line 494: It's probably unnecessary to cache here and for `dim Module`. Since `cokernel` and `generators` are both cached, caching at `dim Module` should suffice.

The reason I say this is that `CacheFunctions` make debugging a bit complicated, so trying to use them only at the bottlenecks is better.

DanGrayson:
That use of "try" was indeed worth removing, but maybe now it's not getting the degree map set correctly, as there are failures when building.

M2/Macaulay2/m2/matrix1.m2
8d1h: You are right, this does not seem to improve much.

DanGrayson:
Multiple builds are failing because of an error in running the example code for the documentation of `symmetricPower`.

DanGrayson:
Can this pull request be revived?

8d1h:
Sorry, I almost forgot about this...

Maybe that `try` should not be removed after all, since the problem is with `try` itself, and I have no idea how to fix that. I can close this PR if you want.

DanGrayson:
To avoid confusion, I'll close it for now.  Feel free to reopen it or to make another one later on, and thanks for your work.

PR #1911, opened 2021-02-04
added package RationalPoints2

8d1h:
The package `RationalPoints` no longer works properly for some ideals (see #1672). Also the algorithm is very basic and runs rather slow once the characteristic/dimension gets big. For example it won't be able to find the points of a zero dimensional ideal in some large characteristic/dimension affine space.

This is a new version written from scratch using elimination. The package runs much faster for large examples. Also the interface is expanded so we can treat homogeneous ideals as projective varieties, and also specify the coefficient field.

Since this is not just a modification of the old package, I'm submitting it as a new package under the name `RationalPoints2` to avoid replacing the old one. (I contacted the author but received no response.)


DanGrayson:
Thank you for the contribution!

PR #1912, opened 2021-02-04
Fix an assertation related to typecodes in scc1

### LINKED ISSUE #1134: Bad error message running scc1 (chk.c:314: assertion failed)
For the following d file.

```
Y := {+ a:int, b:int, c:bool};
Z := {+ d:int, e:int};
Z2 := {+ d:int, e:int, f:int};
U := Y or Z;
W2 := {  x:int, av:tarray(Z)};
W3 := {  x:int, a:(Y or Z2)};
W4 := {  x:int, a:{+ aa:Y, bb:Z2}};
U2 := Y or Z or tarray(Z);
U3 := Y or Z or {+ x:int, y:bool };
-- Following code fails to compile with assertion failure in chk.c:314
bad () : U2 := (U2(new tarray(Z) len 1 do provide Z(11,22)));

-- The following code also fails with an assertion failure
-- bad2(e:U2) : U2 := (
--  when e
--     is b:(tarray(Z)) do e 
--     else e
--  );

-- Following code works
-- A := tarray (Z);
-- good () : U2 := (U2(new tarray(Z) len 1 do provide Z(11,22)));
-- good2(e:U2) : U2 := (
--  when e
--    is b:(tarray(Z)) do e 
--    else e
-- );
```

Fails to compile with "chk.c:314: assertion failed". Without the `bad` function, the code compiles, although it generates some pretty interesting structs. With the `A:=tarray(Z)` the function `good` (which is identical to `bad`) compiles.

~~The most straightforward fix would be to forbid `tarray(Z)` in `new`, but should it also be forbidden in the definition of `U2` for example? and what about the definitions of `U3`,`W2`,`W3`,`W4` Also that would make `tarray` and `array` inconsistent in how they are used with `new`.~~

Edit: I made an actual syntax error in my original code, hiding the fact that `bad2` in fact fails with an assertion failure rather than a syntax error. This makes the question of what the correct fix is less obvious to me. I also added a `good2` which is identical to `bad2`.

jkyang92:
This fixes #1134. In detail, first it stops scc1 from identifying tagged types with each other. Then for tagged types that don't have a name, it generates a name of the form `_<typecode>_unnamed` when a typecode is requested. The value of the typecodes for the unnamed types should never matter, since being unnamed, these types cannot occur in more than one place. 

An alternative strategy would be to place the warning about unnamed types where the type is constructed and create a dummy name there, rather than when the typecode is requested, but that requires a more significant change to the code.

At this point, the code for identifying identical types is still used for non-tagged structs and arrays.

This now makes the dummy fields in types like `GlobalQuote` in Macaulay2/d/parse.d redundant, but this pull request does not remove those yet.


DanGrayson:
Thanks, it's good to have someone looking at that old code.

Issue #1914, opened 2021-02-05
check_3 "SparseResultants" takes a long time


d-torrance:
It was killed during an [Ubuntu continuous integration test on ppc64el](https://objectstorage.prodstack4-5.canonical.com/v1/AUTH_77e2ada1e7a84929a74ba3b87153c0ac/autopkgtest-hirsute/hirsute/ppc64el/m/macaulay2/20210130_190812_e47f3@/log.gz), which is currently preventing the most recent version of the Macaulay2 package from migrating from "proposed" to "release" for Ubuntu 21.04:

```m2
SparseResultants
****************
 -- capturing check(0, "SparseResultants")                                   -- 0.198924 seconds elapsed
 -- capturing check(1, "SparseResultants")                                   -- 12.8166 seconds elapsed
 -- capturing check(2, "SparseResultants")                                   -- 3.69444 seconds elapsed
 -- capturing check(3, "SparseResultants")                                  Killed

 cd /tmp/M2-4875-0/369-rundir/; GC_MAXIMUM_HEAP_SIZE=400M "/usr/bin/M2-binary" -q --int --no-randomize --no-readline --silent --stop --print-width 129 <"/tmp/M2-4875-0/368.m2" >>"/tmp/M2-4875-0/368.tmp"
/tmp/M2-4875-0/368.tmp:0:1: (output file) error: Macaulay2 killed by signal 9
/tmp/M2-4875-0/368.m2:0:1: (input file)
M2: *** Error 9
```
It takes a long time in general.  For example, from [the most recent amd64 build in Debian](https://buildd.debian.org/status/fetch.php?pkg=macaulay2&arch=amd64&ver=1.17.1%2Bds-2&stamp=1611607466&raw=0):

```m2
-- capturing check(3, "SparseResultants")                                   -- 188.982 seconds elapsed
```

d-torrance:
This test checks five matrices, and the 4th one is the slow one:

```m2
i10 : elapsedTime property {{0,0,0},{1,0,0},{0,1,0},{0,0,1},{0,1,2}};
phi: PP^3 --> PP^5, degree of image:3
det A = -15737
-*An example of sparse resultant*-
 -- 0.0739692 seconds elapsed

i11 : elapsedTime property {{0,0,0},{1,0,0},{1,1,0},{0,1,0},{0,0,1},{0,1,2}};
phi: PP^3 --> PP^6, degree of image:5
det A = 4955
-*An example of sparse resultant*-
[...]

PR #1915, opened 2021-02-05
Remove slow case from SparseResultants test

d-torrance:
Closes: #1914


mahrud:
It's better not to outright remove a test. At least comment it out and mention how long it is taking now so it can be compared and potentially re-enabled in the future.

d-torrance:
> It's better not to outright remove a test. At least comment it out and mention how long it is taking now so it can be compared and potentially re-enabled in the future.

Fair enough!  I've force-pushed a new version.  I didn't mention a specific time since it will vary from system to system (~2 minutes on my office desktop, but presumably we hit 700 seconds on ppc64el in Ubuntu for it to get killed.)

mahrud:
You should mention a specific time and a rough description of your machine, otherwise we wouldn't know if it has gotten better or worse. Alternatively, you can use `cpuTime` before and after.

d-torrance:
> You should mention a specific time and a rough description of your machine, otherwise we wouldn't know if it has gotten better or worse. Alternatively, you can use `cpuTime` before and after.

Done!

PR #1927, opened 2021-02-09
Set "Up" node to "(dir) for "Top" nodes of package info files

d-torrance:
Previously, they served as their own "Up" node, but the convention
among other info files on the system (e.g., the Emacs documentation)
is to point to the directory node.

### DIFF
--- a/M2/Macaulay2/m2/installPackage.m2
+++ b/M2/Macaulay2/m2/installPackage.m2
@@ -354,7 +354,8 @@ installInfo := (pkg, installPrefix, installLayout, verboseLog) -> (
 	    if PREV#?tag then infofile << ", Prev: " << infoTagConvert' format PREV#tag;
 	    -- nodes without an Up: link tend to make the emacs info reader unable to construct the table of contents,
 	    -- and the display isn't as nice after that error occurs
-	    infofile << ", Up: " << if UP#?tag then infoTagConvert' format UP#tag else "Top";
+	    infofile << ", Up: " << if UP#?tag then infoTagConvert' format UP#tag else
+		if fkey == infotitle then "(dir)" else "Top";
 	    infofile << endl << endl << info fetchProcessedDocumentation(pkg, fkey) << endl));
     infofile << "\037" << endl << "Tag Table:" << endl;
     scan(values byteOffsets, b -> infofile << b << endl);


d-torrance:
### Before

```
Up: Top

FirstPackage -- an example Macaulay2 package
********************************************

Description
===========

FirstPackage is a basic package to be used as an example.

...
```

Pressing "U" does nothing.

### After

```
Up: (dir)

FirstPackage -- an example Macaulay2 package
********************************************

Description
===========

FirstPackage is a basic package to be used as an example.

...
```

Pressing "U" takes us to the top level info directory:


```
File: dir,      Node: Top,      This is the top of the INFO tree.

This is the Info main menu (aka directory node).
A few useful Info commands:

  'q' quits;
  'H' lists all Info commands;
  'h' starts the Info tutorial;
  'mTexinfo RET' visits the Texinfo manual, etc.

* Menu:

Basics
* Common options: (coreutils)Common options.
* Coreutils: (coreutils).       Core GNU (file, text, shell) utilities.
* Date input formats: (coreutils)Date input formats.
...
```

mahrud:
Is the convention rigid? I wonder if jumping to Macaulay2Doc, or better yet "packages provided with Macaulay2" would be more helpful.

d-torrance:
> Is the convention rigid? I wonder if jumping to Macaulay2Doc, or better yet "packages provided with Macaulay2" would be more helpful.

I haven't seen it written down anywhere -- it was just something I noticed.  Regardless, I like your idea!

I like "packages provided by Macaulay2", but it's a node of Macaulay2Doc, so we'd end up with an acyclic graph, which doesn't seem ideal.

Perhaps a new .info file, equivalent to the index file used by `infoHelp`, that can serve as the top node for all Macaulay2 info files?  It's "Up" node could then be "(dir)".

d-torrance:
> > Is the convention rigid? I wonder if jumping to Macaulay2Doc, or better yet "packages provided with Macaulay2" would be more helpful.
> 
> I haven't seen it written down anywhere -- it was just something I noticed. Regardless, I like your idea!
> 
> I like "packages provided by Macaulay2", but it's a node of Macaulay2Doc, so we'd end up without an acyclic graph, which doesn't seem ideal.
> 
> Perhaps a new .info file, equivalent to the index file used by `infoHelp`, that can serve as the top node for all Macaulay2 info files? It's "Up" node could then be "(dir)".

mahrud:
Currently `info M2` brings up the manual page, which imo is pretty useless. Perhaps that can be replaced? Alternatively, `help Macaulay2` points to `help Macaulay2Doc`, so `info Macaulay2` is the other option.

d-torrance:
Perhaps `info M2` and `info Macaulay2` could both point to the same place, this proposed index node?  I'll look into it.  I've converted this PR into a draft for now.

d-torrance:
I'm going to leave this PR as it is for now.  I think it's the correct thing to do for how we're currently handling the info files.

In particular, if we just run `info` and pull up the directory node, we see:

```
Macaulay2 and its packages
* AbstractToricVarieties: (AbstractToricVarieties).
                                links abstract simplicial (normal) toric 
                                  varieties to Schubert2
* AdjointIdeal: (AdjointIdeal). adjoint ideals of plane curves and related 
                                  computations
* AlgebraicSplines: (AlgebraicSplines).
                                splines on simplicial complexes, polytopal 
                                  complexes, and graphs

...
```

So currently, each package's Top node is a child of the directory node.  Entering one of them and then trying to go back up should send us back to the directory node, not some other Macaulay2 node.

Looking into the future, I think it totally makes sense to have a single Macaulay2 index node that can serve as the parent of all the package nodes, but that definitely would take some more work.  For example, refactoring `makePackageIndex` to take care of both the html and info cases, adding `~/.Macaulay2/local/share/info` to `INFOPATH` during `setup()`, etc.

d-torrance:
I think the fact that `info Macaulay2` works is just because `Macaulay2Doc` is the closest match that `info` can find.  For example, `info Mac` also pulls it up.

mahrud:
It doesn't:

[mahrud@noether ~]$ info Macaulay2
info: No menu item 'Macaulay2' in node '(dir)Top'
[mahrud@noether ~]$ info Mac
info: No menu item 'Mac' in node '(dir)Top'
[mahrud@noether ~]$ info Macaulay2Doc
[works]

d-torrance:
> It doesn't:
> 
> [mahrud@noether ~]$ info Macaulay2
> info: No menu item 'Macaulay2' in node '(dir)Top'
> [mahrud@noether ~]$ info Mac
> info: No menu item 'Mac' in node '(dir)Top'
> [mahrud@noether ~]$ info Macaulay2Doc
> [works]

Huh, strange.  It works for me in Ubuntu 20.10 and Debian unstable.  Maybe different versions of info, or some optional feature has been turned on?

Issue #1522, opened 2020-09-28
Lifting non-free modules


jchen419:
Currently lift(Module, Ring) gives an error if the module is not free:
```
i1 : R = QQ[x,y,z,w,u]/(u^4 - x*y*z*w)

o1 = R

o1 : QuotientRing

i2 : S = R/(x^3 - y^2*z)

o2 = S

o2 : QuotientRing

i3 : T = S/(y*w^2 - z*x^2)

o3 = T

o3 : QuotientRing

i4 : (N1, N2) = (coker gens ideal (T_0^4,T_1^4), comodule ideal (T_2^6,T_3^7))

o4 = (cokernel | xy2z y4 |, cokernel | z6 w7 |)

o4 : Sequence

i5 : M = N1 ++ N2

o5 = cokernel | xy2z y4 0  0  |
              | 0    0  z6 w7 |

                            2
o5 : T-module, quotient of T

i6 : lift(M, S)
stdio:6:1:(3): error: lift: expected module to be free
```
I suggest some code like the following be used to lift non-free modules:
```
i7 : nonFreeLift = method()

o7 = nonFreeLift

o7 : MethodFunction

i8 : nonFreeLift (Module, Ring) := Module => (M, S) -> (
          ringRel := presentation ring M ** S;
          liftRingRel := id_(lift(target relations M, S)) ** ringRel;
          subquotient(lift(gens M, S), lift(relations M, S) | liftRingRel)
     )

o8 = Module => -*Function[stdio:8:48-11:58]*-

o8 : Option

i9 : nonFreeLift (M, S)

o9 = subquotient (| 1 0 |, | xy2z y4 0  0  -x2z+yw2 0 0 0        0 0 |)
                  | 0 1 |  | 0    0  z6 w7 0        0 0 -x2z+yw2 0 0 |

                               2
o9 : S-module, subquotient of S
```
One may want to minimize the presentation before returning it (in the example above, the module is not minimally presented). Ideally, something like this would be the actual code that is run (cf. #1477) when `lift` is called for a non-free module.

DanGrayson:
Can you also produce the isomorphism from o9**T to M and verify it's an isomorphism?  That isomorphism would presumably be needed in applications, unless it's an identity map.

moorewf:
Forgive my ignorance, but isn't this covered by pushForward(phi,M) where
phi is the projection map from S to T?  I get the same answer in your
example (after pruning the result of pushForward).

[...]

Issue #1773, opened 2021-01-01
check_9 "InvariantRing" sometimes fails


d-torrance:
Hrm, the test failure is puzzling.  It definitely seems to be caused by these commits, and in particular the changes to `startup.m2.in` that get compiled into the binary.  I get different results depending on which binary I use, but using the same source directory for the `.m2` files:

### Normal M2 binary
*From PPA package based on release-1.17 branch*
```m2
i1 : needsPackage "InvariantRing";

i2 : R = QQ[x_1..x_4];

i3 : invariants diagonalAction(matrix {{0,1,-1,1},{1,0,-1,-1}}, R)

               2
o3 = {x x x , x x x }
       1 2 3   1 3 4

o3 : List
```

## M2 binary compiled from this branch
```m2
i1 : needsPackage "InvariantRing";

i2 : R = QQ[x_1..x_4];

i3 : invariants diagonalAction(matrix {{0,1,-1,1},{1,0,-1,-1}}, R)

       5   3 2   5 3 4     4 2 3     6 3 3   2
o3 = {x x x x , x x x x , x x x x , x x x , x x x , x x x }
       1 2 3 4   1 2 3 4   1 2 3 4   1 3 4   1 3 4   1 2 3

o3 : List
```

_Originally posted by @d-torrance in https://github.com/Macaulay2/M2/issues/1750#issuecomment-751742504_

d-torrance:
This failed again on a CI test of the Debian experimental package in armhf, so I figured I'd make an issue out of it:
https://ci.debian.net/data/autopkgtest/unstable/armhf/m/macaulay2/9176633/log.gz

https://github.com/Macaulay2/M2/pull/1750#issuecomment-751796540 is also relevant:
> Not sure what the underlying issue was, but it seemed like the new code in the binary was somehow changing how the keys of some hash tables used by invariants were sorted, which apparently messed up the algorithm

mahrud:
Yikes, if your guess is correct, I'm blaming all the hardcoded magic numbers in the interpreter.

mahrud:
Did it happen consistently? If so we can build from that commit and see what is going wrong.

d-torrance:
> Did it happen consistently? If so we can build from that commit and see what is going wrong.

I think so, yes.

DanGrayson:
Keys in hash tables are not "sorted", they're just deposited, along with corresponding values, in various "buckets".  A mathematical algorithm that changes its answer when hash codes change is incorrect.

[...]

PR #1929, opened 2021-02-10
Hookify and add top level caching for basis, pushForward

### LINKED ISSUE #909: 'basis' gives incorrect answer in presence of variables of degree 0
The following is incorrect:
```
A = ZZ/101[a..d, Degrees=>{4:0}]
basis(2, A^1)
```

The basis command returns ```image matrix {{1_A}}```, but the answer should be the zero module.

The problem seems to stem from the computation in findHeftAndVars in Macaulay2/m2/modules2.m2, which somehow returns a heft vector {0} in this case.  The internal engine code apparently doesn't check that the heft vector is correct.

mahrud:
(Naturally, this is on top of #1918, but I think only 26 files are changed in the commits specific to this PR)

- [x] fixes #909
- [x] `basis`:
  - [x] moved to `basis.m2`
  - [x] hookify (so a strategy can be added for associative algebras) and use top level computation caching
  - [x] fixed various related bugs and added more tests
  - [x] changed the behavior of `findHeft` when no variables are given
- [x] more features for the cached computations pattern
- [x] changed `new OptionTable from List` to just ignore null entries, so matches `HashTable`
- [x] `pushForward`:
  - [x] moved to `pushforward.m2`
  - [x] hookify (so pushFwd can be added as a strategy) and use top level computation
  - [x] added more tests
  - [x] unexport `PushforwardComputation`
  - [x] added strategy based on #1522
- [x] simplify a few things with resolutions and make sure FastNonminimal resolutions are cached.

Many more TODO items are added at the top of added files, however, @mikestillman, this bug likely requires a fix in the engine:
```m2
basis(0, ZZ/101[a, Degrees => {0}]) -- should give error
basis(0, ZZ/101[a, Degrees => {0}]/ideal (a^3)) -- should be {1 a a2}
```
Perhaps I'm misunderstanding the point of degree zero elements, but since `a` is not invertible I think it should be considered a generator, and the basis for the first one should be `1 a a2 a3 ...` which isn't finite. But even then, these two seem to be wrong:
```m2
basis(0, ZZ/101[a, Degrees => {0}], Variables => {0})
basis(0, ZZ/101[a, Degrees => {0}]/ideal (a^3), Variables => {0})
```
These two should still work though:
```m2
basis(0, frac(ZZ/101[a, Degrees => {0}]))
basis(0, frac(ZZ/101[a, Degrees => {0}]), Variables => {0})
```


mahrud:
@d-torrance @DanGrayson have either of you seen an InvariantRings error before? Also, shouldn't the artifacts file contain the error output?

DanGrayson:
> @d-torrance @DanGrayson have either of you seen an InvariantRings error before? Also, shouldn't the artifacts file contain the error output?

Sorry, I don't remember.  One would hope that the artifacts file would contain everything we need.

d-torrance:
> @d-torrance @DanGrayson have either of you seen an InvariantRings error before? Also, shouldn't the artifacts file contain the error output?

Yes -- #1773

mahrud:
Thanks, I restarted the builds.

mahrud:
I can't reproduce it, but here is the error:
```m2
2021-02-11T10:01:01.6128114Z /home/runner/work/M2/M2/M2/Macaulay2/packages/InvariantRing/Tests.m2:112:1: error:
2021-02-11T10:01:01.6165505Z  -- i3 : invariants2 = set {x_1*x_2*x_3,x_1^2*x_3*x_4}
2021-02-11T10:01:01.6166463Z  -- 
2021-02-11T10:01:01.6167210Z  --                    2
2021-02-11T10:01:01.6167836Z  -- o3 = set {x x x , x x x }
2021-02-11T10:01:01.6168468Z  --            1 2 3   1 3 4
2021-02-11T10:01:01.6169053Z  -- 
2021-02-11T10:01:01.6169638Z  -- o3 : Set
2021-02-11T10:01:01.6170198Z  -- 
2021-02-11T10:01:01.6170937Z  -- i4 : assert(set invariants T2 === invariants2)
2021-02-11T10:01:01.6172718Z  -- stdio:6:1:(3): error: assertion failed
2021-02-11T10:01:01.6175907Z  -- 
2021-02-11T10:01:01.6176823Z stdio:1:1:(3): error: test(s) #9 of package InvariantRing failed.
```
It's probably the same bug that you saw in #1773. @DanGrayson could you take a look at the code for equality of sets?

DanGrayson:
> @DanGrayson could you take a look at the code for equality of sets?

It looks okay to me:
```
equal(x:HashTable,y:HashTable):Expr := (
     if x == y then return True;
     if x.hash != y.hash then return False;
     if x.Class != y.Class || x.parent != y.parent then return False;
     if x.hash == 0 && x.Class == cacheTableClass then return True; -- cache tables have hash code 0
     if x.Mutable || y.Mutable then return False;
     if x.numEntries != y.numEntries || length(x.table) != length(y.table) then return False;
     foreach a at i in x.table do (
	  p := a;
	  q := y.table.i;
	  if p.next == q.next then (
	       -- p.next and q.next must both be bucketEnd
	       if p.hash != q.hash 
	       || p.key != q.key && False == equal(p.key,q.key) 
	       || p.value != q.value && False == equal(p.value,q.value) 
	       then return False;)
	  else (
	       plen := 0; pp := p; while pp != pp.next do (pp=pp.next; plen=plen+1);
	       qlen := 0; qq := q; while qq != qq.next do (qq=qq.next; qlen=qlen+1);
	       if plen != qlen then return False;
	       while true do (
		    if p == p.next then break;
		    z := q;
		    while true do (
			 if z.key == p.key
			 || z.hash == p.hash && equal(z.key,p.key)==True
			 then (
			      if z.value == p.value
			      || True == equal(z.value, p.value)
			      then break
			      else return False;
			      );
			 z = z.next;
			 if z == z.next then return False;
			 );
		    p = p.next;
		    )));
     True);
```

mahrud:
One thing mentioned by Mike and I agree with: Perhaps `basis` should do _less_.
- the option `SourceRing` should probably be eliminated and left to user to change the basis to another ring. At the very least, instead of assuming that there's always a canonical map from the source ring to the ring of the basis, the option should be replaced with an option that takes a `RingMap`, so the user can decide what the `DegreeLift` map should be. Better yet, we should eliminate this and add a new function `liftBasis` for lifting the source of a map to another module.
- the option `Truncate` should probably be deprecated, since we already have `truncate`

One possible thing that should be added instead of the above is to have the option `Limit` allow computing a finite number of basis elements of a module that is not finite over the base. Say, `basis(kk[x], Limit => 3)` should probably be `| 1 x x2 |` rather than an error.

PR #1930, opened 2021-02-10
Use findProgram/runProgram for fig2dev

d-torrance:
The `gfanInterface` package has two functions (`gfanRender` and `gfanRenderStaircase`) which call `fig2dev` to generate a .png image.  I currently need a [small patch in the Debian package](https://salsa.debian.org/science-team/macaulay2/-/blob/debian/master/debian/patches/use-debian-applications.patch) to get it to work properly.  But it works out of the box with `findProgram` and `runProgram`.

### DIFF
--- a/M2/Macaulay2/packages/gfanInterface.m2
+++ b/M2/Macaulay2/packages/gfanInterface.m2
@@ -2604,7 +2604,6 @@ doc ///
 			loadPackage("gfanInterface", Reload => true)
 
 		Text
-			The path to the executables should end in a slash.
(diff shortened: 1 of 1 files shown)


d-torrance:
I like how this package uses `show URL("file://" ...` to display the .png image after it's generated.  Since `show` calls `open`/`xdg-open`, this should end calling the user's preferred image viewer automatically.  We should probably migrate some other functions that do similar things to use `show` instead of having the user specify the viewer as a configuration option (e.g., `displayGraph` in `Graphs`).

Issue #1135, opened 2020-05-09
conservative garbage collector


DanGrayson:
Bill Furnish was working on a conservative garbage collector for M2 when he died.  It would be good to see whether the ideas in it are worth bringing to fruition.  It would replace libgc and be better adapted to working threads.

See https://github.com/garyfurnish/cgc1

DanGrayson:
I've forked it to Macaulay2, so we can keep track of it better.

mahrud:
Should we open a new "Project" for looking into pros & cons of switching to a different garbage collector? The other alternative is Julia's [gc.h](https://github.com/JuliaLang/julia/blob/master/src/gc.h).

DanGrayson:
If the project had multiple components that could be completed separately, that might be appropriate, but I can't discern any component tasks here.

mahrud:
Another alternative: https://github.com/v8/v8/tree/master/include/cppgc
This one comes with a paper: https://research.google/pubs/pub47359/
And a blog post: https://v8.dev/blog/high-performance-cpp-gc

PR #1934, opened 2021-02-11
Valgrind

jkyang92:
This pull request fixes two memory bugs, the first of which was required to get M2 to run under valgrind, and the second was found immediately after running under valgrind.

1. The `M2File` constructor previously didn't initialize `exclusiveChangeCondition`. While this tended to be fine in practice, valgrind is stricter about uninitialized values.
2. `mpz_get_str` and `mpfr_get_str` both use gmp/mpir's default allocator, which does not allocate in the GC heap, and so their return values must be manually freed.

I believe the second problem also affects the pull request #1465, since it also uses mpfr_get_str.

The main reason this pull request is a draft is that I didn't add a valgrind suppressions file to silence most of the valgrind warnings coming from libgc. I have one mostly written, but I'm not certain where in the tree it should go. If there is a place to put it, I can add that to this pull request.

### DIFF
--- a/M2/Macaulay2/d/gmp.d
+++ b/M2/Macaulay2/d/gmp.d
@@ -402,7 +402,11 @@ export (x:ZZ) ^^ (y:ZZ) : ZZ := (
 base := 10;
 toCstring(x:ZZ) ::= getstr(charstarOrNull(null()), base, x);
 
-export tostring(x:ZZ):string := tostring(toCstring(x));
+export tostring(x:ZZ):string := (
+     cstr := toCstring(x);
+     ret := tostring(cstr);
+     Ccode(void,"mp_free_str(", cstr, ")");
+     ret);
 
 export (x:int) + (y:ZZ) : ZZ := toInteger(x) + y;
 
--- a/M2/Macaulay2/d/gmp1.d
+++ b/M2/Macaulay2/d/gmp1.d
@@ -48,9 +48,13 @@ floor(x:double) ::= Ccode(double, "floor(", x, ")" );
 log2ten := log(10.) / log(2.);
 base := 10;
 
-getstr(returnexponent:long, base:int, sigdigs:int, x:RR) ::= tostring(
-     Ccode(charstarOrNull, "(M2_charstarOrNull) mpfr_get_str((char *)0,&", returnexponent, ",",
-	  base, ",(size_t)", sigdigs, ",", x, ",GMP_RNDN)"));
+getstr(returnexponent:long, base:int, sigdigs:int, x:RR):string := (
+     strptr := Ccode(charstarOrNull, "(M2_charstarOrNull) mpfr_get_str((char *)0,&", returnexponent, ",",
+	  base, ",(size_t)", sigdigs, ",", x, ",GMP_RNDN)");
+     ret := tostring(strptr);
+     Ccode(void, "mpfr_free_str(", strptr, ")");
+     ret);
+
 export format(
      s:int,			  -- number of significant digits (0 means all)
      ac:int,	    -- accuracy, how far to right of point to go (-1 means all)
--- a/M2/Macaulay2/d/gmp1.d
+++ b/M2/Macaulay2/d/gmp1.d
@@ -48,7 +48,7 @@ floor(x:double) ::= Ccode(double, "floor(", x, ")" );
 log2ten := log(10.) / log(2.);
 base := 10;
 
-getstr(returnexponent:long, base:int, sigdigs:int, x:RR):string := (
+getstr(returnexponent:long, base:int, sigdigs:int, x:RR) ::= (
      strptr := Ccode(charstarOrNull, "(M2_charstarOrNull) mpfr_get_str((char *)0,&", returnexponent, ",",
 	  base, ",(size_t)", sigdigs, ",", x, ",GMP_RNDN)");
      ret := tostring(strptr);
--- a/M2/Macaulay2/d/gmp_aux.c
+++ b/M2/Macaulay2/d/gmp_aux.c
@@ -21,6 +21,12 @@ int mpfr_hash(mpfr_srcptr x) {
   return 777 + h * 3737 + x->_mpfr_exp + 11 * x->_mpfr_sign;
 }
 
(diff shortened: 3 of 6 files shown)


DanGrayson:
Thanks, that looks great!

jkyang92:
Sorry about the churn, I had forgotten to run check after making the mpfr changes, and didn't realize why getstr had been a macro. Hopefully it should all be fixed now.

Also, do you have any thoughts on where to put a valgrind suppression file?

DanGrayson:
> Sorry about the churn, I had forgotten to run check after making the mpfr changes, and didn't realize why getstr had been a macro. Hopefully it should all be fixed now.
> 
> Also, do you have any thoughts on where to put a valgrind suppression file?

I've cancelled the redundant builds for you.  You may have been able to do that yourself.

You could put the file in `M2/files`.  There's one other file there.

mikestillman:
Hi Jay, this is great, I've wanted valgrind to work on M2 before, but the last time I tried, it appeared to not be compatible with gc.  Thanks!  I wonder if we can now use cachegrind too?

jkyang92:
I've not tried cachegrind, although I suspect it'll be easier to get to work, memcheck tends to be the troublesome one. 

There are limitations to what valgrind/memcheck can do. It's not aware of the layout of the GC heap, and so it doesn't really know what in the GC heap is valid vs invalid. But it can still catch some errors. The main thing for now is suppressing all of the warnings from libgc since libgc sometimes needs to read uninitialized memory to do it's mark-sweep process. In theory if we were to rewrite the garbage collector, we could add valgrind annotations to tell valgrind what memory is allocated.

DanGrayson:
> In theory if we were to rewrite the garbage collector, we could add valgrind annotations to tell valgrind what memory is allocated.

In connection with rewriting the garbage collector, see https://github.com/Macaulay2/M2/issues/1135

mahrud:
There are existing, field-tested garbage collectors with support for valgrind. The first step would be the ability to "hot-swap" the garbage collector, which is currently not possible. See #1734.

mahrud:
Would it be feasible at all to add valgrind tests to be run along github builds? Not necessarily for this PR, but it would be a nice way to automatically detect leaks in PRs before merging.

jkyang92:
I don't know how feasible it would be overall, valgrind tends to make everything much slower (10x is common) so I suspect we'll run out of time on github before we can run all the package checks. Even ignoring time constraints some work involved though, since check runs the tests in a subprocess, and so knowing that a subprocess had an error might take a bit of doing. 

Just generally though, it would be great to have a non-hacky way to run tests under valgrind though. For now I just wrote a bit of code that gets the test cases from the package and runs them in process.

Also, changed my mind on how to write the suppression file, and just suppressed all the valgrind warnings about uninitialized values in libgc.

mahrud:
Oh no not though `check`, I meant maybe just the engine tests, either the gtest ones or maybe a few quick tests under `tests/engine`.

mikestillman:
@jkyang92 What command line for valgrind have you been using to create your log files?  (I am having trouble on ubuntu 20.04 getting valgrind to display any warnings or leaks.  I'm probably doing something stupid...).  Here is an example line I am using, on a debug build of M2.

```
valgrind --tool=memcheck --leak-check=full --log-file=foo.log ./M2 -e "exit 0" <foo.m2 
```

@moorewf and I fixed a couple of the memory leaks you found, and have a couple more to fix (but your logs zero us in to exactly where the leaks are to be found).

jkyang92:
@mikestillman Here's the command line I was using:

```
valgrind --trace-children=yes --trace-children-skip=/bin/sh --suppressions=../../files/M2-suppressions.supp --leak-check=full --track-origins=yes ./M2 foo.m2 2> foo.log
```

As a note, memcheck is the default tool and the --trace-children=yes is needed since ./M2 is just a shell script. Otherwise you'll just be debugging the shell script. You might need to adjust the --trace-children-skip=/bin/sh, I think some systems might put it in /usr/bin/sh, but it's only needed to get rid of a bunch of traces that we don't need so it's not a big deal either way.

You can also put all of the common stuff in a .valgrindrc file so that you don't have to type it every time (See section 2.6.7 in [https://www.valgrind.org/docs/manual/manual-core.html](https://www.valgrind.org/docs/manual/manual-core.html))

mahrud:
@jkyang92 I made a pull request (https://github.com/jkyang92/M2/pull/1) to your branch that sets up CTest to run Valgrind. You can run any tests through valgrind by adding `-T memcheck`, for instance for engine tests:
```
ctest -T memcheck -R unit-tests:ARingQQFlint
...
Total Test time (real) =  23.51 sec

The following tests FAILED:
	 32 - unit-tests:ARingQQFlint.arithmetic (Failed)
-- Processing memory checking output:
1/3 MemCheck: #30: unit-tests:ARingQQFlint.create .......   Defects: 5
2/3 MemCheck: #31: unit-tests:ARingQQFlint.display ......   Defects: 5
3/3 MemCheck: #32: unit-tests:ARingQQFlint.arithmetic ...   Defects: 18
MemCheck log files can be found here: (<#> corresponds to test number)
/home/mahrud/Projects/M2/engine/M2/BUILD/build/Testing/Temporary/MemoryChecker.<#>.log
Memory checking results:
Potential Memory Leak - 28
```
Or m2 tests: (memleak5.m2 is the code from https://github.com/Macaulay2/M2/issues/1938)
```
ctest -T memcheck -R memleak5
...
Total Test time (real) = 236.61 sec

The following tests FAILED:
	658 - normal/memleak5.m2 (Failed)
-- Processing memory checking output:
1/1 MemCheck: #658: normal/memleak5.m2 ...............   Defects: 347
MemCheck log files can be found here: (<#> corresponds to test number)
/home/mahrud/Projects/M2/engine/M2/BUILD/build/Testing/Temporary/MemoryChecker.<#>.log
Memory checking results:
Potential Memory Leak - 347
```
Or even package tests:
```
ctest -T memcheck -R NormalToricVarieties-0 -V
```
That should help @mikestillman run the tests more easily, and once we clear out the bugs sufficiently we should be able to run this for the engine tests on github as well.

Issue #1904, opened 2021-01-31
Feature request: iterators


mahrud:
A cool feature of more developed languages like Julia or Python is that they have [iterator generators](https://www.python.org/dev/peps/pep-0255/). It would be a very cool feature if M2 supported something like the `yield` keyword in Python, so you could have something like `for p in nextPrime N`. This would require the interpreter to make a new call to that function on every cycle to get the next number.

Other than the memory advantage, another benefit of `yield` is a time speedup: in the `nextPrime` example, if a loop over primes ends at `17`, there's no need to pre-compute all primes under `N`. Therefore, [async generators](https://www.python.org/dev/peps/pep-0525/) are a good milestone towards [async functions](https://www.python.org/dev/peps/pep-0530/).

Another related problem that would be solved with iterators is that `accumulate` and `fold` use `reverse` and `drop`, which unnecessarily copy the data, hence an optimized copy-free implementation of `accumulate` and `fold` would be more efficient.

_Originally posted by @mahrud in https://github.com/Macaulay2/M2/issues/1897#issuecomment-769958455_

mahrud:
There's no reason this should take 10 seconds:
```m2
i7 : elapsedTime for i in 0..10000000 do ();
 -- 11.1732 seconds elapsed
```
Compare with python, using iterators:
```
$ time python -c 'for i in range(10000000): pass'

real	0m0.359s
user	0m0.347s
sys	0m0.010s
```
The problem is partially that `0..N` constructs a list very inefficiently, but also basic loops are unreasonably slow:
```m2
i8 : elapsedTime for i to 10000000 do ();
 -- 4.34505 seconds elapsed
```

DanGrayson:
Perhaps the difference is that python uses 32 bit integers for the values of i, whereas we use bignumbers.

mahrud:
> Perhaps the difference is that python uses 32 bit integers for the values of i, whereas we use bignumbers.

Nope:
```
$ time python -c 'for i in range(0,2**1024,2**1022): print(i)'
0
44942328371557897693232629769725618340449424473557664318357520289433168951375240783177119330601884005280028469967848339414697442203604155623211857659868531094441973356216371319075554900311523529863270738021251442209537670585615720368478277635206809290837627671146574559986811484619929076208839082406056034304
[...]

Issue #1878, opened 2021-01-25
Method with Dispatch => Thing, Option => true


mahrud:
```m2
func = method(Dispatch => Thing, Options => true)
func List     := true >> o -> L -> func(toSequence L, o)
func Sequence := true >> o -> S -> (print class S; S)

assert instance(func(1,2), Sequence)        -- prints Sequence
assert instance(func{1,2}, Sequence)        -- prints Sequence
assert instance(func{1}, Sequence)          -- prints ZZ, assertion fails
assert instance(func(1), Sequence)          -- error
assert instance(func sequence(1), Sequence) -- error
```
@DanGrayson can you explain the result of the third and fifth assertion above? Even the output of the third assection doesn't make sense to me. Why does `toSequence {1}` end up with class `ZZ`??

DanGrayson:
This code might display the behavior more perspicuously:
```m2
func = method(Dispatch => Thing, Options => true);
func Thing := true >> o -> S -> (o,S);
func(a,b,Foo=>bar)
func(a,Foo=>bar)
func(1:a,Foo=>bar)
func (1:1:a)
func (1:a)
func 1
```

A method defined both with options and with Dispatch=>Thing perhaps doesn't have the best behavior at the time the options are processsed.   The dilemma is well illustrated by this case:
```m2
i9 : func(a,Foo=>bar)

o9 = (OptionTable{Foo => bar}, a)

o9 : Sequence
```
After stripping the option(s) out, do you leave `a` in a sequence or not?  I chose not to.  One could imagine doing it the other way...

A strong case can be made for doing it the way we do by looking at these results, which show some pleasing consistency:
```m2

i5 : func = method(Dispatch => Thing, Options => true);

i6 : func Thing := true >> o -> S -> S;

i7 : func(a)

o7 = a

o7 : Symbol

i8 : func(a,Foo=>bar)

o8 = a

o8 : Symbol

i9 : func(a,Foo=>bar,Boo=>car)

o9 = a

o9 : Symbol

i10 : func(a,Foo=>bar,Boo=>car,Goo=>tar)

o10 = a

o10 : Symbol

```

A strong case can be made for doing it the other way by looking at this sequence of results, which shows a single exception:
```m2
Macaulay2, version 1.17.1.1
with packages: ConwayPolynomials, Elimination, IntegralClosure, InverseSystems, LLLBases,
               MinimalPrimes, PrimaryDecomposition, ReesAlgebra, Saturation, TangentCone

[...]

PR #1939, opened 2021-02-15
Implement binary (associative) methods with options

mahrud:
This is on top of #1929.

- Core:
  - methods.m2:
    - [x] rewrite single methods to be twice as efficient
    - [x] rewrite binary methods to be 10x as efficient
    - [x] implement support for binary methods with options
    - [x] minor organization improvements
  - [x] change intersect to be a binary associative method
  - [x] change tensor to be a binary associative method
  - [x] initialize randomSeed early so startup is reproducible
  - [x] some minor organization improvements in system.m2, etc.
- SimpleDoc:
  - [x] better debugging info
  - [x] new abbreviated usage notation

As a result of these changes, packages can now install methods on intersect or tensor for their object types, with arbitrary options.

Potential future changes:
- implement right associative methods
- make more things associative, like min/max/sum/etc.


mahrud:
@DanGrayson it might be good if you review individual files in these commits sooner rather than later, while the changes are fresh in my mind. methods.m2 would be my main review request.

DanGrayson:
What's the basic idea of "associative" methods?

mahrud:
> What's the basic idea of "associative" methods?

I'm not sure what you're asking me since they are not new, here is [the documentation](http://www2.macaulay2.com/Macaulay2/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/___Method__Function__Binary.html).

DanGrayson:
Gee, that was 15 years ago that we used to call them that.  It's a good thing we don't call them that any longer, since there's nothing intrinsically associative about them: they're just methods where `f(a,b,c)` is implemented as `f(f(a,b),c)`, and so on.

mahrud:
Well, they weren't thoroughly removed:
```m2
i1 : f = method(Binary => true, Options => true)
stdio:1:5:(3): error: associative methods with options not implemented yet
```
But I don't see what's wrong with having that. In fact, I think we should have both left associative and right associative methods, for when you specifically want `f(a, f(b, c))`

DanGrayson:
Okay.  But let me suggest better terminology, as associativity would refer to the identity `f(f(a,b),c) == f(a,f(b,c))`.  Let's call them *left-iterative* and *right-iterative*.

mahrud:
> Okay. But let me suggest better terminology, as associativity would refer to the identity `f(f(a,b),c) == f(a,f(b,c))`. Let's call them _left-iterative_ and _right-iterative_.

Left-associative and right-associative are the right [terminology](https://en.wikipedia.org/wiki/Operator_associativity):
> Operators may be associative (meaning the operations can be grouped arbitrarily), left-associative (meaning the operations are grouped from the left), right-associative (meaning the operations are grouped from the right) or non-associative (meaning operations cannot be chained, often because the output type is incompatible with the input types).

Regardless, this isn't the topic of this PR, but a matter for future changes.

DanGrayson:
> > Okay. But let me suggest better terminology, as associativity would refer to the identity `f(f(a,b),c) == f(a,f(b,c))`. Let's call them _left-iterative_ and _right-iterative_.
> 
> Left-associative and right-associative are the right [terminology](https://en.wikipedia.org/wiki/Operator_associativity):
> 

No, that's in the context of parsing an expression such as `a+b+c`, which is not happening here.

mahrud:
> No, that's in the context of parsing an expression such as `a+b+c`, which is not happening here.

I don't see the difference between a binary method `sum(a,b,c)` and `a+b+c`. They should be treated as synonyms, as is the case in languages like Julia. In particular, if `a+b+c` is taken as `Sum{1,2,3}`, then `expressionValue Sum{1,2,3}` and `sum(1,2,3)` should be literally the same.

DanGrayson:
Well, okay, I guess that is one way to look at it.

timduff35:
(This post is relevant to tomorrow's Internals meeting) I tried building from [refactor/methods](https://github.com/mahrud/M2/tree/refactor/methods) and got up to M2 `ninja M2-binary M2-core`, with the following output. Looks like it might be related to boost? My version is 1.71, and I am on Ubuntu 20.04.

[M2binarycoreOutput.txt](https://github.com/Macaulay2/M2/files/6239906/M2binarycoreOutput.txt)

mahrud:
> (This post is relevant to tomorrow's Internals meeting) I tried building from [refactor/methods](https://github.com/mahrud/M2/tree/refactor/methods) and got up to `ninja M2-binary M2-core`, with the following output. Looks like it might be related to boost? My version is 1.71, and I am on Ubuntu 20.04.
> 
> [M2binarycoreOutput.txt](https://github.com/Macaulay2/M2/files/6239906/M2binarycoreOutput.txt)

See https://github.com/Macaulay2/M2/issues/1982.

mahrud:
@DanGrayson this PR can be reviewed by you can.

M2/Macaulay2/m2/methods.m2
```diff
@@ -1,159 +1,262 @@
...
-MethodFunctionSingle.synonym = "method function with a single argument"
-MethodFunctionBinary.synonym = "binary method function"
-MethodFunctionWithOptions.synonym = "method function with options"
+-----------------------------------------------------------------------------
+-- Local utilities
+-----------------------------------------------------------------------------
 
-dispatcherFunctions = {}
+-- see lists.m2
+all' := (L, p) -> not any(L, x -> not p x)
```
DanGrayson, line 20: Why not just use `all` here?

M2/Macaulay2/m2/methods.m2
```diff
@@ -1,159 +1,262 @@
...
+    else {   printarg(" ",            args,   false) }) -- TODO: do better here, in what way?
 
-noMethodSingle = (meth,args,output) -> error toString stack( line0 meth, f(" ",args,output))
-noMethod = (meth,args,outputs) -> error toString stack join( {line0 meth},
-     if class args === Sequence and 0 < #args and #args <= 4
-     then apply(#args, i -> f(toString (i+1),args#i,if outputs#?i then outputs#i else false))
-     else {f(" ",args,
-	       false					    -- do better here
-	       )})
+-- TODO: what does this for exactly?
```
DanGrayson, line 34: "does this for"??

M2/Macaulay2/m2/ringmap.m2
```diff
@@ -435,17 +435,10 @@ RingMap ** Matrix := Matrix => (f,m) -> (
...
-     f ** M)
-
-tensor(Ring,RingMap,Matrix) := opts -> (S,f,m) -> (
-     if S =!= target f then error "tensor: expected ring and target of ring map to be the same";
-     f ** m)
-
-tensor(RingMap,Module) := Module => opts -> (f,M) -> f ** M
-
-tensor(RingMap,Matrix) := Matrix => opts -> (f,m) -> f ** m
+tensor(Ring, RingMap) := RingMap => {} >> opts -> (S, f) -> (
```
DanGrayson, line 438: This method doesn't seem mathematical, and it doesn't do anything, either.  What's it for?

M2/Macaulay2/m2/lists.m2
```diff
@@ -28,6 +28,7 @@ List + List  := List => (v,w) -> apply(v,w,plus)
      - List  := List => v -> apply(v,minus)
 List - List  := List => (v,w) -> apply(v,w,difference)
 Thing * List := List => (a,v) -> apply(v,x->a * x)
+List ** List := List => (X,Y) -> flatten for x in X list apply(Y, y -> {x, y})
```
DanGrayson, line 31: If this is implementing a Cartesian product of two lists somehow, then should the result be a list of pairs (of class Sequence), rather than a list of lists of length 2?  What is the intended application?

M2/Macaulay2/m2/fold.m2
```diff
@@ -1,4 +1,7 @@
 --		Copyright 1993-1999 by Daniel R. Grayson
+-- TODO: reverse is not copy-free, optimize this!
+-- TODO: drop is not copy-free, optimize this!
+-- TODO: both can be solved using iterators
```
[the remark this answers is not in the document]
mahrud: When reading the source for a project, it's much easier to see the tasks that need to be done right there, rather than being forces to search through ~600 issues for relevant tasks. There's also an issue for iterators: #1904.

M2/Macaulay2/m2/methods.m2
```diff
@@ -1,159 +1,262 @@
...
+-- newmethod1
+-- newmethod1234c
+
+notImplemented = x -> error concatenate between(" ", splice {x, "not implemented yet"})
+
+-----------------------------------------------------------------------------
+
+-- see fold.m2
+-- TODO: add Left associative vs Right associative methods
+foldL := (f, x, L) -> (scan(L, y -> x = f(x, y)); x)
```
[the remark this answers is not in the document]
mahrud: `fold` is a method that is defined in a later source file. I can't use a method to construct methods!! It would be good to put `foldL` and `foldR` in the interpreter, perhaps.

M2/Macaulay2/m2/intersect.m2
```diff
@@ -29,37 +29,34 @@ eliminationInfo Ring := (cacheValue symbol eliminationInfo) (R -> (
...
 -----------------------------------------------------------------------------
 
-intersect List     :=
-intersect Sequence := -* [same as input type] => *- true >> opts -> L -> (
-    if not #L > 0 then error "intersect: expected at least one object";
+-- intersect is now declared as a binary associative method in shared.m2,
+-- so the methods (intersect, Sequence) and (intersect, List) are pre-installed.
+-- Here, we override (intersect, List) with a specialized method for handling a uniform list.
+-- As a backup, at the end the List is converted to a Sequence.
+intersect List := true >> opts -> L -> (
```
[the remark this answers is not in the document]
mahrud: As the comment says, this is to allow for overriding the method `(intersect, List)`, specifically because there are sometimes faster algorithms for intersecting N objects at the same time rather than one by one. This is the case for modules, for instance.

M2/Macaulay2/m2/methods.m2
mahrud: As with `foldL`, it's because `all` is a method that isn't yet defined, and we can't use methods for defining methods.

M2/Macaulay2/m2/ringmap.m2
mahrud: This is to allow for the [method](https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_tensor_lp__Ring_cm__Ring__Map_cm__Matrix_rp.html) `(tensor, Ring, RingMap, Matrix)`.

I think it makes as much sense as `(tensor, RingMap, Matrix)`, don't you think?

M2/Macaulay2/m2/lists.m2
mahrud: This is used a few places. I'm ambivalent about returning lists or sequences, but it's current applications (for instance in PrimaryDecomposition package) might rely on returning a list of lists. @moorewf what do you prefer?

M2/Macaulay2/m2/methods.m2
mahrud: Should be "is for". What is `badClass` for?

M2/Macaulay2/m2/fold.m2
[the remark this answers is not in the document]
DanGrayson: No, this file really has nothing to do with iterators, and if you posted an issue, there could be a discussion about it, with differences of opinion presented.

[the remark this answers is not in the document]
mahrud: I'm not sure what you're asking me here exactly since I didn't say this is the one and only way to solve it, I just mentioned two problems and a possible solution, so anyone reviewing the code in the future doesn't have to spend as much time. The point of mentioning iterators is that it is a solution to the first two items listed. Does adding a few comments in the code hurt anything?

M2/Macaulay2/m2/methods.m2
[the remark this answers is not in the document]
DanGrayson: Let me be the one to "resolve" my conversations.

The method function `fold` can be used to make a method if the method isn't called before `fold` is defined.   It can probably work.  That would be good because it would eliminate duplication of code.

M2/Macaulay2/m2/intersect.m2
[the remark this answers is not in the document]
DanGrayson: I don't understand your answer, because N objects can be provided in a sequence and in a list equally well.

M2/Macaulay2/m2/methods.m2
DanGrayson: As with `fold`, if the method isn't called before "fold" is defined, it will work.

M2/Macaulay2/m2/methods.m2
DanGrayson: It's used further down in the file to print error messages.

M2/Macaulay2/m2/ringmap.m2
DanGrayson: > This is to allow for the [method](https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_tensor_lp__Ring_cm__Ring__Map_cm__Matrix_rp.html) `(tensor, Ring, RingMap, Matrix)`.
> 

In what sense does it "allow for" the [method](https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_tensor_lp__Ring_cm__Ring__Map_cm__Matrix_rp.html) `(tensor, Ring, RingMap, Matrix)`.

> I think it makes as much mathematical sense as `(tensor, RingMap, Matrix)`, don't you think?

No, since it never does any mathematical computation, and it doesn't correspond to any mathematical operation.

M2/Macaulay2/m2/fold.m2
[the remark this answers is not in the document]
DanGrayson: If it's a good idea, no one will ever see it.

By the way, I don't see how iterators allow "reverse" to be implemented.

M2/Macaulay2/m2/lists.m2
DanGrayson: It should be compatible with the way it works for sets:
```m2

i3 : set {a,b,c} ** set {1,2,3}

o3 = set {(a, 1), (a, 2), (a, 3), (b, 1), (b, 2), (b, 3), (c, 1), (c, 2), (c, 3)}

o3 : Set
```

M2/Macaulay2/m2/methods.m2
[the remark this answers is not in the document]
mahrud: I understand that, but a big point of this PR was to optimize methods as much as I could without changing the interpreter, and using a method call in the function that evaluates another method call slows things down quite a bit.

I'm surprised that you're concerned about duplication of literally one line of code when there are huge repetitions all over the place, but do you prefer if I used `foldL` in definition of `fold`?

Feel free to turn this conversation into a new issue for someone to implement a basic `foldL` function in the interpreter, which is then used in both `methods.m2` and `fold.m2`. This is my preferred solution, but I don't want to deal with the interpreter in this PR.

M2/Macaulay2/m2/intersect.m2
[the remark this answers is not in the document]
mahrud: No, there's a subtle difference between a list and a sequence in this context: #1878.

More to the point, when the input to an associative binary method is a Sequence it implies "compute by sequentially applying the binary method from left to right," but when a List is given, I wanted to keep the possibility of allowing specialized algorithms that consider the whole list in parallel rather than sequentially.

Note that I specifically don't want to always use the specialized algorithms, because in some cases it may be slower, so it's good to allow both so the user can decide which to use based on the type of input.

M2/Macaulay2/m2/methods.m2
mahrud: What? `all'` is used in the function `method`
https://github.com/Macaulay2/M2/blob/8950020fd3018ed094e38cedb8738be5fede37a4/M2/Macaulay2/m2/methods.m2#L208-L211
and `all` is defined by calling the function `method`:
https://github.com/Macaulay2/M2/blob/8950020fd3018ed094e38cedb8738be5fede37a4/M2/Macaulay2/m2/lists.m2#L101
The literal definition of the method `all` will call the function `method` which needs `all'`. How will this work?

This conversation is hypocritical because you wrote `all'`:
https://github.com/Macaulay2/M2/blob/80fdea3b784101577cfea93d428043805520a2e6/M2/Macaulay2/m2/methods.m2#L130
I just replaced it with a more efficient version than yours!

M2/Macaulay2/m2/methods.m2
mahrud: I can also use the search function! Can you show me a sample code that results in that error message? Specifically, this string:
```
expected an output class, but got:
```

M2/Macaulay2/m2/ringmap.m2
mahrud: > In what sense does it "allow for" the [method](https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_tensor_lp__Ring_cm__Ring__Map_cm__Matrix_rp.html) `(tensor, Ring, RingMap, Matrix)`.

I think this comment answers your question:
https://github.com/Macaulay2/M2/blob/8950020fd3018ed094e38cedb8738be5fede37a4/M2/Macaulay2/m2/intersect.m2#L75-L80

> No, since it never does any mathematical computation, and it doesn't correspond to any mathematical operation.

Okay, then let's get rid of this and `(tensor, Ring, RingMap, Matrix)` at the same time, because the latter has a redundant Ring input.

DanGrayson: Okay, sure.

M2/Macaulay2/m2/lists.m2
mahrud: Done.

M2/Macaulay2/m2/ringmap.m2
mahrud: Done.

M2/Macaulay2/m2/methods.m2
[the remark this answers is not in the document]
mahrud: I resolved duplicated code.

M2/Macaulay2/m2/fold.m2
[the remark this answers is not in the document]
mahrud: Changed the comment to point to the issue instead.

M2/Macaulay2/m2/methods.m2
mahrud: I'll resolve this conversation as I don't see a way to change anything here. You can of course undo it if you have another suggestion.

M2/Macaulay2/m2/methods.m2
mahrud: The comment is fixed, but I'll leave this conversation open as I'm still curious when that error can occur!

M2/Macaulay2/m2/intersect.m2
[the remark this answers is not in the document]
mahrud: Let me know if this is sufficient explanation. If you have any suggestions for how to allow a binary method on both lists and sequences and allow both sequential computation and all at once computation, let me know! This was the best solution I could think of.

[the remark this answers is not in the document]
DanGrayson: > No, there's a subtle difference between a list and a sequence in this context: #1878.
> 
> More to the point, when the input to an associative binary method is a Sequence it implies "compute by sequentially applying the binary method from left to right," but when a List is given, I wanted to keep the possibility of allowing specialized algorithms that consider the whole list in parallel rather than sequentially.

It seems strange to signal the preferred strategy by the having user pass a list instead of a sequence.  Since sequences are problematic(#1878), why not just insist that lists be used instead of sequences, and offer a "Strategy" option?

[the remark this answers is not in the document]
mahrud: I believe what you're saying is incompatible with `(tensor, List)` being a binary method. If something like #1979 was implemented, then I could make Lists of modules an exception and do it your way, but I don't know how to implement it another way so that tensoring a list of Modules happens in parallel but a list of rings happens associatively.

M2/Macaulay2/m2/methods.m2
mahrud: Is this dead code?

DanGrayson: Probably not.

M2/Macaulay2/m2/intersect.m2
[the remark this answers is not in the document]
DanGrayson: We could let all such associative methods be iterative by default, and let users specify the simultaneous strategy with the `Strategy` option.  (There doesn't seem to be a burning need for simultaneous tensor products of modules by default.)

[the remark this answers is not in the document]
mahrud: Ah sorry I meant intersect, not tensor (hence the section of code we're discussing). Using iterative intersection for modules is significantly slower, so I don't see what would be the improvement, and also I don't understand your suggestion with strategy. Where would you check for the strategy? Are you suggesting having all binary associative methods take options?

I still think having a meaningful and documented distinction between `(intersect, List)` and `(intersect, Sequence)` makes the most sense. You can apply the same logic to other binary functions if you have a better algorithm for simultaneous computation.

M2/Macaulay2/m2/methods.m2
mahrud: Then can you give me an example of code that results in that error?

DanGrayson: Not without a bit of study ...

M2/Macaulay2/m2/intersect.m2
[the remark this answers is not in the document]
DanGrayson: I don't think there can be a meaningful distinction between `(intersect,List)` and `(intersect,Sequence)`.  Lists and sequences are meant to be basically interchangeable as repositories of objects.

Not all associative methods would need to take options -- just those where you want to offer a way of applying the operation simultaneously to all the arguments.

M2/Macaulay2/m2/methods.m2
mahrud: Could you do a bit of study then? I spent quite a bit trying to figure this out and tracked it down to the `env.2` argument here:
https://github.com/Macaulay2/M2/blob/3c3e3f7ddbc1a0ff20bd6723e41251557dea2936/M2/Macaulay2/d/actors5.d#L800-L804
but this comment doesn't make sense and the code below it isn't much better either. Is it for making a method based on the output argument? Does M2 support such a thing? It would really help if you documented your code in the `d` directory.

M2/Macaulay2/m2/intersect.m2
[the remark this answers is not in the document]
mahrud: I think I offered a meaningful distinction between the two earlier:
> when the input to an associative binary method is a Sequence it implies "compute by sequentially applying the binary method from left to right," but when a List is given, I wanted to keep the possibility of allowing specialized algorithms that consider the whole list in parallel rather than sequentially.

Otherwise why have both lists and sequences? A lot of code would be much simpler if one of them simply inherited from the other.

M2/Macaulay2/m2/methods.m2
DanGrayson: Determining whether 1 line of code is dead doesn't seem urgent enough.

M2/Macaulay2/m2/intersect.m2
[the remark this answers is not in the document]
DanGrayson: The reason to have both sequences and lists is to provide something lightweight for use when calling functions with multiple arguments.  A sequence doesn't contain a pointer to its class (so you can't make a new type of sequence), and it also doesn't contain a precomputed hashcode.  Other than that, lists and sequences are meant to be basically interchangeable as repositories of objects, and trying to impose a semantic distinction, like the one you desire, would lead to confusion, even if documented.

M2/Macaulay2/m2/methods.m2
mahrud: I asked this 26 days ago. When do you think you would get around to it?

M2/Macaulay2/m2/intersect.m2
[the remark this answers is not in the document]
DanGrayson: I don't have an exact change to your code to propose.

M2/Macaulay2/m2/methods.m2
DanGrayson: Never, for even if one could prove that there is no way to trigger the error message from top level currently, it might be worthwhile to keep the error-detecting code present, as a way of preventing "internal" errors, caused by future changes to the code.  Such error-detection is especially valuable when chunks of code in widely separate areas must be closely coordinated.

mahrud: If you have an error that can't be triggered it gives the false belief that the code is safe, when in reality it may not be. Also keeping dead code around is bad etiquette for future readers and maintainers, so I'll delete it when I simplify the `d` code for methods in my next PR.

M2/Macaulay2/m2/intersect.m2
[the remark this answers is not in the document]
mahrud: A suggestion on where to check the value of the strategy option that allows both sequential and simultaneous computation strategies while keeping intersect a binary method would suffice. I've thought about this carefully and I can't think of any possibility that doesn't come with a significant performance hit.

M2/Macaulay2/m2/methods.m2
DanGrayson: No, let it stay.

mahrud: The way I'm simplifying `newmethod1234c` there's no point to passing this `badClass` function as an argument. If I'm misunderstanding something about `newmethod1234c`, please let me know, though it seems like you prefer that I don't wait.

DanGrayson: Is there any reason to change `newmethod1234c`?

mahrud: Yes. The methods it creates are slow when options are given. It's a major bottleneck for me.

DanGrayson: How many times do you have to call the methods to perform your computation?

Do you have some code that demonstrates the slowness?

mahrud: Sure, take a look at the benchmarks in `tests/normal/methods.m2`.

DanGrayson: What benchmarks?  (I don't see any timing commands or any use of the Benchmark package there.)

mahrud: If you don't see the tests I've added then I'm sorry but I really don't have the time or energy to keep debating you and convincing you that the code you haven't touched in more than a decade is in need of improvement, especially when you refuse to explain your code when I ask about it. Try writing any of your own benchmarks and compare with other symbolic algebra software.

mahrud:
Let me know if you have other comments or requests. I'm happy to make quick changes but since it's been months since I wrote this, I've moved on from it and don't anticipate having the time to make significant changes to it.

If the slight, behind-the-scenes difference of `(intersect, List)` and `(intersect, Sequence)` is absolutely unacceptable, you can comment out the redefinition of `(intersect, List)` in `intersect.m2`, but that would slow down intersection of modules.

If other major changes are needed, feel free to either accept the PR and make the changes yourself, or make an issue to do it later, or reject the PR. Though I've been working on other things on top of this branch and will be submitting changes based on it later.

DanGrayson:
> If the slight, behind-the-scenes difference of `(intersect, List)` and `(intersect, Sequence)` is absolutely unacceptable, you can comment out the redefinition of `(intersect, List)` in `intersect.m2`, but that would slow down intersection of modules.


It's probably better if both `(intersect, List)` and `(intersect, Sequence)` do it the same fast way, so the user doesn't unintentionally fall into the slow way of doing things.

mahrud:
> It's probably better if both `(intersect, List)` and `(intersect, Sequence)` do it the same fast way, so the user doesn't unintentionally fall into the slow way of doing things.

This would necessitate not making `intersect` a binary method. Is that what you intend?

DanGrayson:
> > It's probably better if both `(intersect, List)` and `(intersect, Sequence)` do it the same fast way, so the user doesn't unintentionally fall into the slow way of doing things.
> 
> This would necessitate making `intersect` not a binary method. Is that what you intend?

Well, it's your pull request, so if that's how you wish to implement it, that would be fine.

mahrud:
The PR is already implemented the way I think is best. I'm asking how you suggest I implement your suggestion.

DanGrayson:
> The PR is already implemented the way I think is best. I'm asking how you suggest I implement your suggestion.

I don't have a detailed change to your code to suggest.

mahrud:
@DanGrayson is this better?

DanGrayson:
I'll take a look ...

mahrud:
@DanGrayson fixed the merge conflicts here. I expect the tests to pass as before. Let me know when you take a look.

M2/Macaulay2/tests/normal/intersect.m2
```diff
@@ -0,0 +1,40 @@
...
+     subquotient(matrix {{a*b}},matrix {{d}})
+     )
+
+--
+R = ZZ/101[a..d]
+M = coker random(R^{0,-1}, R^{-2,-3})
+L = new SelfInitializingType of VisibleList
+t1 = first elapsedTiming intersect L(20:M)
+t2 = first elapsedTiming intersect  (20:M)
+assert(t1 > 4 * t2)
```
DanGrayson, line 40: What's the point of this assertion?

DanGrayson (changes requested):
Much improved, thank you.  I have just a few comments and suggestions.

M2/Macaulay2/tests/normal/intersect.m2
mahrud: It checks that taking intersection of 20 modules 2 at a time takes much longer than all at the same time. It'll prevent accidentally disabling the simultaneous intersection strategy.

Issue #1938, opened 2021-02-14
gmp memory leak


jkyang92:
The following code leaks gmp memory:

```
R = ZZ[x,y,z];
I = ideal(9*x^6*y^2*z^2-8*x^2*y^3*z^5,3*x^8*z^2-x^5*y^2*z^3,x^5*y^5-3*x^2*y^6*z^2);
mingens gb I;
I = null;
R = null;
collectGarbage();
collectGarbage();
collectGarbage();
collectGarbage();
```
valgrind gives the following stack trace

```
==32340== 672 bytes in 84 blocks are definitely lost in loss record 5 of 15
==32340==    at 0x48397EF: malloc (in /usr/lib64/valgrind/vgpreload_memcheck-amd64-linux.so)
==32340==    by 0x5768999: __gmp_default_allocate (in /usr/lib64/libgmp.so.10.4.1)
==32340==    by 0x577DEEB: __gmpz_realloc (in /usr/lib64/libgmp.so.10.4.1)
==32340==    by 0x5777231: __gmpz_gcdext (in /usr/lib64/libgmp.so.10.4.1)
==32340==    by 0x44454D: GBRing::gbvector_combine_lead_terms_ZZ(FreeModule const*, FreeModule const*, gbvector const*, gbvector const*, gbvector const*, gbvector const*, gbvector*&, gbvector*&) (gbring.cpp:1199)
==32340==    by 0x42E7E4: gbA::compute_s_pair(gbA::spair*) (gb-default.cpp:1303)
==32340==    by 0x431A8A: gbA::reduce_ZZ(gbA::spair*) (gb-default.cpp:1436)
==32340==    by 0x4321D8: reduceit (gb-default.cpp:1563)
==32340==    by 0x4321D8: gbA::process_spair(gbA::spair*) (gb-default.cpp:2281)
==32340==    by 0x43358F: gbA::do_computation() (gb-default.cpp:2500)
==32340==    by 0x433ADF: gbA::start_computation() (gb-default.cpp:2574)
==32340==    by 0x4016E4: GBProxy::start_computation() (comp-gb-proxy.hpp:55)
==32340==    by 0x50EF43: rawStartComputation (groebner.cpp:621)
```

I've attached the full valgrind output. You can also verify that it's not a false positive by running the code in a loop and seeing that it does consume an increasing amount of memory.

[gmpleak.log](https://github.com/Macaulay2/M2/files/5976827/gmpleak.log)

[...]

PR #1942, opened 2021-02-16
Valgrind in CTest

mahrud:
- [x] make tbb a requirement in CMake and remove `WITH_TBB` option.
- [x] update a Dockerfile
- [x] sets up CTest to run Valgrind. You can run any tests through valgrind by adding `-T memcheck`, for instance for engine tests:
```
ctest -T memcheck -R unit-tests:ARingQQFlint
...
Total Test time (real) =  23.51 sec

The following tests FAILED:
	 32 - unit-tests:ARingQQFlint.arithmetic (Failed)
-- Processing memory checking output:
1/3 MemCheck: #30: unit-tests:ARingQQFlint.create .......   Defects: 5
2/3 MemCheck: #31: unit-tests:ARingQQFlint.display ......   Defects: 5
3/3 MemCheck: #32: unit-tests:ARingQQFlint.arithmetic ...   Defects: 18
MemCheck log files can be found here: (<#> corresponds to test number)
/home/mahrud/Projects/M2/engine/M2/BUILD/build/Testing/Temporary/MemoryChecker.<#>.log
Memory checking results:
Potential Memory Leak - 28
```
Or m2 tests: (memleak5.m2 is the code from https://github.com/Macaulay2/M2/issues/1938)
```
ctest -T memcheck -R memleak5
...
Total Test time (real) = 236.61 sec

The following tests FAILED:
	658 - normal/memleak5.m2 (Failed)
-- Processing memory checking output:
1/1 MemCheck: #658: normal/memleak5.m2 ...............   Defects: 347
MemCheck log files can be found here: (<#> corresponds to test number)
/home/mahrud/Projects/M2/engine/M2/BUILD/build/Testing/Temporary/MemoryChecker.<#>.log
Memory checking results:
Potential Memory Leak - 347
```
Or even package tests:
```
ctest -T memcheck -R NormalToricVarieties-0 -V
```
That should help @mikestillman run the tests more easily, and once we clear out the bugs sufficiently we should be able to run this for the engine tests on github as well.

_Originally posted by @mahrud in https://github.com/Macaulay2/M2/issues/1934#issuecomment-779825651_

### DIFF
--- a/M2/Macaulay2/tests/normal/memleak5.m2
+++ b/M2/Macaulay2/tests/normal/memleak5.m2
@@ -1,0 +1,12 @@
+-- try with:
+--  ctest -T memcheck -R memleak5
+
+R = ZZ[x,y,z];
+I = ideal(9*x^6*y^2*z^2-8*x^2*y^3*z^5,3*x^8*z^2-x^5*y^2*z^3,x^5*y^5-3*x^2*y^6*z^2);
+mingens gb I;
+I = null;
+R = null;
+collectGarbage();
+collectGarbage();
+collectGarbage();
+collectGarbage();
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -133,11 +133,11 @@ jobs:
         if: matrix.build-system == 'cmake'
         run: |
           cmake -S../.. -B. -GNinja \
-            --debug-trycompile \
             -DBUILD_LIBRARIES="Givaro;FFLAS_FFPACK" \
-            -DCMAKE_BUILD_TYPE=MinSizeRel -DBUILD_NATIVE=OFF \
+            -DCMAKE_BUILD_TYPE=Release -DBUILD_NATIVE=OFF \
             -DCMAKE_PREFIX_PATH=`brew --prefix` \
-            -DCMAKE_INSTALL_PREFIX=/usr
+            -DCMAKE_INSTALL_PREFIX=/usr \
+            --debug-trycompile
 
       - name: Build libraries using Ninja
         if: matrix.build-system == 'cmake'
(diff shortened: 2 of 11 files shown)


M2/Macaulay2/tests/normal/memleak5.m2
```diff
@@ -0,0 +1,12 @@
...
+
+R = ZZ[x,y,z];
+I = ideal(9*x^6*y^2*z^2-8*x^2*y^3*z^5,3*x^8*z^2-x^5*y^2*z^3,x^5*y^5-3*x^2*y^6*z^2);
+mingens gb I;
+I = null;
+R = null;
+collectGarbage();
+collectGarbage();
+collectGarbage();
+collectGarbage();
```
DanGrayson, line 12: It seems that this file won't ever produce an error, so there is no reason to run it normally, and it could be put in a new directory called something like `tests/memory-leak-tests`, where people can actually find it.

mahrud: It does if you run it this way:
```
ctest -T memcheck -R memleak5
```
or if you manually run `M2 --check 2` through Valgrind, both of which are reasonable.
There are also other tests with `memleak` prefix in that directory, so I don't see your objection to this file in particular. Perhaps they can all be moved, but I don't want to change both CMake and autotools to handle another directory in this PR.

mahrud: If you decide to create such a directory, I propose naming it just `tests/memory`, and also collecting your various GC tests in there. Specifically `m2/basictests/testgc.c`, `util/gc_demo.c`, and `util/gc_test.c`.

mikestillman: I agree with this.  I think that having a separate test directory for such issues is good, but should not be part of this pull request.

PR #1948, opened 2021-02-19
Add ./configure --with-emacsdir=... option

d-torrance:
This lets the user select the installation directory for the Emacs Lisp files instead of being stuck with the default `${datadir}/emacs/site-lisp/macaulay2`.

For example, in the Debian package, they are installed (by `dh-elpa`) to `/usr/share/emacs/site-lisp/elpa/macaulay2-$VERSION`, and I currently need to [patch configure.ac](https://salsa.debian.org/science-team/macaulay2/-/blob/debian/master/debian/patches/use-dh-elpa.patch) so that `currentLayout#"emacs"` points to the right place.  It would be a lot simpler to do this in the build script instead.

We also switch the default value from `${datadir}/emacs/site-lisp/macaulay2` to `${datarootdir}/emacs/site-lisp/macaulay2`, as suggested by the [GNU Coding Standards](https://www.gnu.org/prep/standards/html_node/Directory-Variables.html):

> **‘lispdir’**
The directory for installing any Emacs Lisp files in this package. By default, it should be /usr/local/share/emacs/site-lisp, but it should be written as $(datarootdir)/emacs/site-lisp.

### DIFF
--- a/M2/configure.ac
+++ b/M2/configure.ac
@@ -1911,6 +1911,14 @@ test "$pdfdir"      || AC_SUBST(pdfdir,'${docdir}')
 test "$psdir"       || AC_SUBST(psdir,'${docdir}')
 test "$localedir"   || AC_SUBST(localedir,'${datarootdir}/locale')
 
+AC_ARG_WITH([emacsdir],
+    [AS_HELP_STRING([--with-emacsdir=...],
+        [set installation directory for installing Emacs Lisp files;
+         must begin with ${prefix} or ${exec_prefix}])],
+    [emacsdir=$withval],
+    [emacsdir=${datarootdir}/emacs/site-lisp/macaulay2])
+AC_SUBST(emacsdir)
+
 # Here we normalize all the configure variables so each one begins either with ${prefix} or with ${exec_prefix},
 # so they can be handled more simply in Macaulay2, by a single text replacement.
 save_prefix=$prefix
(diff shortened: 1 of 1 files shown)


DanGrayson:
Good idea.  But the other installation directory options look like this:
```
Fine tuning of the installation directories:
  --bindir=DIR            user executables [EPREFIX/bin]
  --sbindir=DIR           system admin executables [EPREFIX/sbin]
  --libexecdir=DIR        program executables [EPREFIX/libexec]
  --sysconfdir=DIR        read-only single-machine data [PREFIX/etc]
  --sharedstatedir=DIR    modifiable architecture-independent data [PREFIX/com]
  --localstatedir=DIR     modifiable single-machine data [PREFIX/var]
  --libdir=DIR            object code libraries [EPREFIX/lib]
  --includedir=DIR        C header files [PREFIX/include]
  --oldincludedir=DIR     C header files for non-gcc [/usr/include]
  --datarootdir=DIR       read-only arch.-independent data root [PREFIX/share]
  --datadir=DIR           read-only architecture-independent data [DATAROOTDIR]
  --infodir=DIR           info documentation [DATAROOTDIR/info]
  --localedir=DIR         locale-dependent data [DATAROOTDIR/locale]
  --mandir=DIR            man documentation [DATAROOTDIR/man]
  --docdir=DIR            documentation root [DATAROOTDIR/doc/Macaulay2]
  --htmldir=DIR           html documentation [DOCDIR]
  --dvidir=DIR            dvi documentation [DOCDIR]
  --pdfdir=DIR            pdf documentation [DOCDIR]
  --psdir=DIR             ps documentation [DOCDIR]
```

I wonder whether it would be easy to make this option be `--emacsdir` instead.

d-torrance:
> I wonder whether it would be easy to make this option be `--emacsdir` instead.

I don't think it's easy -- see [this Stack Overflow question](https://stackoverflow.com/questions/27416003/create-custom-configure-command-line-arguments).  Even automake uses [`--with-lispdir`](https://www.gnu.org/software/automake/manual/html_node/Public-Macros.html#index-AM_005fPATH_005fLISPDIR) for a similar purpose.

Issue #1931, opened 2021-02-11
Build error: multiple definition of `frobby_version'


haerski:
I'm having issues building the latest development branch. I'm using the CMake build system on Arch, and I'm getting the following error in the linking stage.

```
[350/351] Linking CXX executable usr-dist/x86_64-Linux-Arch-rolling/bin/M2-binary
FAILED: usr-dist/x86_64-Linux-Arch-rolling/bin/M2-binary 
[...]

PR #1951, opened 2021-02-21
update frobby hash

mahrud:


mahrud:
I cancelled the tests because none of the github builds actually build frobby, so nothing would be tested. I tested this locally on ubuntu, would be good to also test it on a mac as well.

DanGrayson:
Could you explain to me why this patch is needed?  If the problem is a possible collision with the symbol of the same name defined in d/version.dd, why don't we just change that instead?

mahrud:
As Jay explained in #1931, this is a mistake in frobby.h regardless.

DanGrayson:
Ah, it seems github doesn't display the correct diffs.  If I do it at home, I get this:
```diff
diff --git a/src/frobby.cpp b/src/frobby.cpp
index a9f2e39..05d0946 100644
--- a/src/frobby.cpp
+++ b/src/frobby.cpp
@@ -27,6 +27,8 @@
 #include "IdealFacade.h"
 #include "SliceParams.h"
 
+const char* const frobby_version = "0.9.5";
+
 class ConsumerWrapper {
 protected:
   ConsumerWrapper(size_t varCount):
diff --git a/src/frobby.h b/src/frobby.h
index c518f4d..6c4fe47 100644
--- a/src/frobby.h
+++ b/src/frobby.h
@@ -21,7 +21,7 @@
 
 #include "gmp.h"
 
-extern "C" const char* const frobby_version = "0.9.5";
+extern "C" const char* const frobby_version;
 
 /** The namespace FrobbyImpl is for internal use inside Frobby only.
  It is not part of the public interface of Frobby.
```

Is that right?

mahrud:
See https://github.com/Macaulay2/frobby/pull/10.

PR #1953, opened 2021-02-22
MonodromySolver: fix failing tests that were missed due to optional components

timduff35:
After my last PR #1621, some other changes were made in ```NumericalAlgebraicGeometry```. The result was that several tests for ```MonodromySolver``` fail in the latest release. Apparently packages with optional components are no longer checked, so I missed this until I tried to build (for the first time in quite awhile!) earlier today.

This pull request makes the following changes:
  1) Fixes failing tests
  2) To avoid future headaches involving optional components, mixed volumes are now computed with gfan instead of PHCPack.


DanGrayson:
External programs upon which cached example output depends should probably be installed as part of our github workflow, so we can test those examples regularly.  Those programs include phcpack, bertini, qepcad, bergman, and polymake.

d-torrance:
> External programs upon which cached example output depends should probably be installed as part of our github workflow, so we can test those examples regularly. Those programs include phcpack, bertini, qepcad, bergman, and polymake.

FYI, polymake is already in Ubuntu, so that would be easy to add.  phcpack, qepcad, and bergman are all on my to-do list.  (bertini has a nonfree license, so it's off the table)

PR #1961, opened 2021-02-24
CMake on gentoo

jkyang92:
This fixes two issues with building with cmake on gentoo.

1. Gentoo installs it's multilib headers in a different way than most distros. In particular for gmp, the real header goes in `/usr/include/<arch>/gmp.h`. Then `/usr/include/gmp.h` is a wrapper around it. In contrast most other systems put the real header in `/usr/include/gmp-64.h` (or `gmp-32.h`). This difference breaks the search method `FindGMP` uses to find the gmp version.
2. Gentoo installs boost such that the stacktrace library is header only. The autotools build correctly detects this, but the cmake build had required a linked stacktrace library.

@mahrud You might want to make sure both of these changes are reasonable.

Additionally, it is theoretically possible to use libbacktrace or addr2line with Boost.Stacktrace in header only mode by setting the right macros, see [https://www.boost.org/doc/libs/1_75_0/doc/html/stacktrace/configuration_and_build.html](https://www.boost.org/doc/libs/1_75_0/doc/html/stacktrace/configuration_and_build.html). But then for libbacktrace we'd have to link against libbacktrace ourselves, and it turns out `FindBacktrace` doesn't find `libbacktrace` but rather something else, and so I gave up on that for now.

### DIFF
--- a/M2/cmake/FindGMP.cmake
+++ b/M2/cmake/FindGMP.cmake
@@ -54,9 +54,26 @@ macro(_gmp_check_version)
         ${GMP_MAJOR_VERSION}.${GMP_MINOR_VERSION}.${GMP_PATCHLEVEL_VERSION})
     endif()
   endforeach()
-
-  set(GMP_VERSION
-    ${GMP_MAJOR_VERSION}.${GMP_MINOR_VERSION}.${GMP_PATCHLEVEL_VERSION})
+  #If a direct search fails, use try_compile.
+  #This is for systems that put multiarch headers in other places.
+  if(NOT GMP_VERSION)
+    try_compile(GMP_TEST_SUCCESS ${CMAKE_BINARY_DIR} ${CMAKE_CURRENT_LIST_DIR}/gmp_version.c
+      OUTPUT_VARIABLE _gmp_test_compile_output)
+    string(REGEX MATCH
+      "__GNU_MP_VERSION[ \t]+([0-9]+)" _gmp_major_version_match
+      "${_gmp_test_compile_output}")
+    if(_gmp_major_version_match)
+      set(GMP_MAJOR_VERSION "${CMAKE_MATCH_1}")
+      string(REGEX MATCH "__GNU_MP_VERSION_MINOR[ \t]+([0-9]+)"
+        _gmp_minor_version_match "${_gmp_test_compile_output}")
+      set(GMP_MINOR_VERSION "${CMAKE_MATCH_1}")
+      string(REGEX MATCH "__GNU_MP_VERSION_PATCHLEVEL[ \t]+([0-9]+)"
+        _gmp_patchlevel_version_match "${_gmp_test_compile_output}")
+      set(GMP_PATCHLEVEL_VERSION "${CMAKE_MATCH_1}")
+      set(GMP_VERSION
+        ${GMP_MAJOR_VERSION}.${GMP_MINOR_VERSION}.${GMP_PATCHLEVEL_VERSION})
+    endif()
+  endif()
 
   # Check whether found version exists and exceeds the minimum requirement
   if(NOT GMP_VERSION)
--- a/M2/cmake/check-libraries.cmake
+++ b/M2/cmake/check-libraries.cmake
@@ -40,7 +40,16 @@ endif()
 
 find_package(Threads	REQUIRED QUIET)
 find_package(LAPACK	REQUIRED QUIET)
-find_package(Boost	REQUIRED QUIET COMPONENTS regex ${Boost_stacktrace})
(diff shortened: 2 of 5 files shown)


M2/cmake/check-libraries.cmake
```diff
@@ -40,7 +40,16 @@ endif()
...
 find_package(LAPACK	REQUIRED QUIET)
-find_package(Boost	REQUIRED QUIET COMPONENTS regex ${Boost_stacktrace})
+find_package(Boost	REQUIRED QUIET COMPONENTS regex OPTIONAL_COMPONENTS stacktrace_backtrace stacktrace_addr2line)
+if(Boost_STACKTRACE_BACKTRACE_FOUND)
+  set(Boost_stacktrace_lib "Boost::stacktrace_backtrace")
+elseif(Boost_STACKTRACE_ADDR2LINE_FOUND)
+  set(Boost_stacktrace_lib "Boost::stacktrace_addr2line")
+else()
+  #fallback to header only mode
+  set(Boost_stacktrace_header_only YES)
```
mahrud, line 50: In this case you're assuming that the headers are found, but they may not be, right? Ideally we should get an error here if the headers aren't found rather than at the end when building M2-binary. You can do this with [`check_include_file`](https://cmake.org/cmake/help/latest/module/CheckIncludeFile.html).

mahrud: This is not relevant for macOS or distributions where Boost is one package, but for instance on Ubuntu boost-regex and boost-stacktrace are separate packages, and the latter may be missing.

M2/cmake/FindGMP.cmake
```diff
@@ -54,9 +54,26 @@ macro(_gmp_check_version)
         ${GMP_MAJOR_VERSION}.${GMP_MINOR_VERSION}.${GMP_PATCHLEVEL_VERSION})
     endif()
   endforeach()
-
-  set(GMP_VERSION
-    ${GMP_MAJOR_VERSION}.${GMP_MINOR_VERSION}.${GMP_PATCHLEVEL_VERSION})
+  #If a direct search fails, use try_compile.
+  #This is for systems that put multiarch headers in other places.
+  if(NOT GMP_VERSION)
+    try_compile(GMP_TEST_SUCCESS ${CMAKE_BINARY_DIR} ${CMAKE_CURRENT_LIST_DIR}/gmp_version.c
```
mahrud, line 60: I think the code below could be significantly simplified if you use the constant `gmp_version` instead of the macros.

Also, it might be preferable to keep this self-contained (e.g. there are no .c sources in `share/cmake/Modules`), and in particular I think using `gmp_version` would make the source short enough that you could just use `file(WRITE ...)` to make a temporary file, compile it, and test the output.

M2/cmake/check-libraries.cmake
jkyang92: I can certainly add that check, but I don't know if any distro installs only half the header files. In particular on ubuntu libboost-stacktrace-dev includes only the libraries themselves, and libboost-dev seems to include all the headers.

See [https://packages.ubuntu.com/groovy/amd64/libboost-stacktrace1.71-dev/filelist](https://packages.ubuntu.com/groovy/amd64/libboost-stacktrace1.71-dev/filelist) and [https://packages.ubuntu.com/groovy/amd64/libboost1.71-dev/filelist](https://packages.ubuntu.com/groovy/amd64/libboost1.71-dev/filelist)

Additionally, I believe the autotools build checks that boost is at least version 1.65, which is the first version boost-stacktrace is avaliable, but doesn't directly check for the header.

mahrud: I see, you're right, it doesn't matter then.

M2/cmake/FindGMP.cmake
mahrud: In particular, this source seems to work:
```c
#include "gmp.h"
#include "stdio.h"

void main() { printf("%s", gmp_version); }
```
and is short enough that it's reasonable to put it in `FindGMP.cmake` itself.

DanGrayson: The return type of `main` should be `int`, though.

jkyang92: I've made this change, I didn't exactly know where to put the temporary file though, so it's currently in `${CMAKE_BINARY_DIR}`.

mahrud (approved):

M2/cmake/FindGMP.cmake
mahrud: That's fine, it already has a bunch of temporary files anyway. I think `${CMAKE_CACHEFILE_DIR}/CMakeFiles` is the correct place, to be pedantic, but it doesn't matter.

PR #1963, opened 2021-02-25
Reconcile declaration & definition of regex_search

jamesjer:
Fedora is building packages with Link Time Optimization (LTO) enabled.  When building Macaulay2, a few warnings are emitted by the link time optimizer about mismatched declarations and definitions.  This patch fixes one of them.  The `start` and `range` parameters are declared as `int` in M2/Macaulay2/d/regex.dd, but defined as `size_t` in `M2/Macaulay2/d/boost-regex.cpp`.  Those types may have different bit widths (e.g., on 64-bit Linux systems, `int` is 32 bits and `size_t` is 64 bits).  Since the other regex functions define `start` and `range` as `int`, I chose to do likewise with `regex_search`.

### DIFF
--- a/M2/Macaulay2/d/boost-regex.cpp
+++ b/M2/Macaulay2/d/boost-regex.cpp
@@ -48,8 +48,8 @@ extern "C" {
 /* returns an array of pairs (s, r), indicating starting point
  * and length of the first match and its capture groups */
 M2_arrayint regex_search(const M2_string pattern,
-                         const size_t start,
-                         const size_t range,
+                         const int start,
+                         const int range,
                          const M2_string text,
                          int regex_flags,
                          int match_flags)


mahrud (approved):
I don't remember why I used `size_t` only for this one, but really the interpreter should be made to understand `size_t` instead.

DanGrayson:
If you mean the D language should include a size_t datatype, that's easy to arrange by adding it to this list in `arithmetic.d`:

```
export uchar := integerType "unsigned char";
export short := integerType "short";
export ushort := integerType "unsigned short";
export uint := integerType "unsigned int";
export long := integerType "long";
export ulong := integerType "unsigned long";
export longlong := integerType "long long";
export ulonglong := integerType "unsigned long long";
export int8_t := integerType "int8_t";
export uint8_t := integerType "uint8_t";
export int16_t := integerType "int16_t";
export uint16_t := integerType "uint16_t";
export int32_t := integerType "int32_t";
export uint32_t := integerType "uint32_t";
export int64_t := integerType "int64_t";
export uint64_t := integerType "uint64_t";
export float := arithmeticType "float";
```

PR #1971, opened 2021-02-28
Patch mpsolve configure.ac to work with autoconf 2.70

jkyang92:
See: https://github.com/robol/MPSolve/commit/3a890878239717e1d5d23f574e4c0073a7249f7a


jkyang92:
Seems like a test failure similar to #1742 (both segfault in MonomialIdeal::quotient). The changes in the pull request are build related, so I think should be unrelated.

PR #1973, opened 2021-03-01
fix gfanInterface's MarkedPolynomialList expression and texMath

pzinn:
@dmaclagan
Paolo Lella asked on the Macaulay2Web chat why the output of gfanInterface's MarkedPolynomialList looked different in WebApp mode.
Since the fix was quite simple, I did it myself. So if that's OK with the package authors here's the PR.

### DIFF
--- a/M2/Macaulay2/packages/gfanInterface.m2
+++ b/M2/Macaulay2/packages/gfanInterface.m2
@@ -247,21 +247,15 @@ markedPolynomialList List := L -> (
 	new MarkedPolynomialList from L
 )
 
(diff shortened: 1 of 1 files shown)


mahrud:
Is there a good documentation of what needs to be implemented by package authors for new types?

DanGrayson:
Dear Authors ( @mikestillman @dmaclagan ),

Is this pull request acceptable?

DanGrayson:
I guess the package authors are indifferent, so I'll pull it.

PR #1981, opened 2021-03-07
Update configure and lrslib

DanGrayson:


mahrud:
The CMake build has been building lrslib v071 with a simple and short patch and everything works fine, why add this huge patch?
https://github.com/DanGrayson/M2/blob/a4c772d0ef205d054882e4cc1794dec117cdebad/M2/libraries/lrslib/patch-071#L1-L14

I've mentioned it before that the way the configure build applies patches is not good (it should add `-p1` to the patch command), but still, the CMake build is smart enough to do that, so you can just edit the first two lines of the patch above to:
```
diff -ur /Users/dan/src/M2/M2.git/M2/BUILD/dan/builds.tmp/gallium-gcc10-development/libraries/lrslib/tmp/lrslib-071a/makefile lrslib-071a/makefile
--- /Users/dan/src/M2/M2.git/M2/BUILD/dan/builds.tmp/gallium-gcc10-development/libraries/lrslib/tmp/lrslib-071a/makefile	2020-10-17 01:18:42.000000000 -0500
+++ lrslib-071a/makefile	2021-02-24 12:23:19.000000000 -0600
```
and everything should work fine with both build systems.

DanGrayson:
> The CMake build has been building lrslib v071 with a simple and short patch and everything works fine, why add this huge patch?

So their makefile will use my setting of CPPFLAGS.

DanGrayson:
> ... the configure build applies patches is not good (it should add `-p1` to the patch command)

Why should it do it that way?

mahrud:
> So their makefile will use my setting of CPPFLAGS.

Why should it? lrslib only needs gmp, and the way you're supposed to provide that is with `INCLUDEDIR` and `LIBDIR`, not by passing a ton of unnecessary flags.

I thought you wanted the builds to be similar. If you don't care, then you can ignore this.

DanGrayson:
> > So their makefile will use my setting of CPPFLAGS.
> 
> Why should it? lrslib only needs gmp, and the way you're supposed to provide that is with `INCLUDEDIR` and `LIBDIR`, not by passing a ton of unnecessary flags.
> 

INCLUDEDIR allows you to include only one directory.  At a certain point I wanted two.

Issue #1982, opened 2021-03-07
CMake build, failure on linking:


kschwede:
This has been going on a few weeks (WSL2, Ubuntu 20.04), which is why I tried compiling in the classical way.
I'm following the instructions in INSTALL-Cmake.md

@mahrud what should I try?

```
schwede@2021AMD:~/M2CMake/M2/M2/BUILD/build$ ninja M2-binary M2-core
[1/2] Linking CXX executable usr-dist/x86_64-Linux-Ubuntu-20.04/bin/M2-binary
FAILED: usr-dist/x86_64-Linux-Ubuntu-20.04/bin/M2-binary
[...]

PR #1983, opened 2021-03-08
Use givaro/fflas-ffpack PPA packages for Ubuntu CMake builds

d-torrance:
As discussed in #1982, I've packaged git snapshots of givaro and fflas-ffpack for the PPA.  ~This PR is just to test the Github builds and isn't intended to be merged.~

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -134,7 +134,6 @@ jobs:
         if: matrix.build-system == 'cmake'
         run: |
           cmake -S../.. -B. -GNinja \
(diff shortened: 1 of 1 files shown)


mahrud:
Actually, instead of an empty commit, could you change the cmake section of the github action script to not build fflas_ffpack and givaro? I think there is also a default option set in cmake/configure.cmake (and documented in INSTALL-CMake.md) that could be changed.

d-torrance:
Everything worked!  I'm assuming that we would rather use the PPA packages and not build givaro/fflas-ffpack in the Ubuntu CMake builds going forward, and so this should be merged?

mahrud:
Yes

Issue #677, opened 2017-07-30
`factor` should be made to work in more cases


DanGrayson:
... such as for elements of `(frac (QQ[a])) [x,y,z]` or `(toField (QQ[a]/f))[x,y,z]`.  For the former case, clear denominators of the coefficients.  For the latter, "factory" provides the functionality.  In the former case, the factors should have coefficients without denominators.

pzinn:
Is it possible to have an example of the desired behavior in each case? I'm curious to see if my package `Factor` already resolves this.

DanGrayson:
```m2
Macaulay2, version 1.16.0.2
with packages: ConwayPolynomials, Elimination, IntegralClosure, InverseSystems, LLLBases, MinimalPrimes, PrimaryDecomposition, ReesAlgebra, Saturation, TangentCone, Truncations

i1 : (frac (QQ[a])) [x,y,z]

o1 = frac(QQ[a])[x, y, z]

o1 : PolynomialRing

i2 : factor x
stdio:2:1:(3): error: expected coefficient ring of the form ZZ/n, ZZ, QQ, or GF

i3 : (toField (QQ[a]/(a^2+1)))[x,y,z]

      QQ[a]
o3 = ------[][x, y, z]
      2
     a  + 1

o3 : PolynomialRing

i4 : factor x
stdio:4:1:(3): error: expected coefficient ring of the form ZZ/n, ZZ, QQ, or GF
```

pzinn:
that's the *un*desired behavior I'm assuming

DanGrayson:
Oh, right.  The desired behavior is to give the right answer.

d-torrance:
This was fixed in #1996

PR #1996, opened 2021-03-18
Update package RationalPoints2

8d1h:
Hi! This is an update of the package *RationalPoints2*. I fixed the error with the height function, and added some utilities related to number fields and some small improvements in styles and performance.

Also a fix for issue #677, since this is needed for the package. Actually for number fields I discovered that the C++ interface to factory is already there (!), so only some simple M2 code is needed for it to work. As for `frac(QQ[a])`, I suppose that the desired behavior is to factorize after removing all the denominators? Here are some sample outputs.
```macaulay2
i1 : toField(QQ[i]/(i^2+1))[x,y]; factor(x^2+y^2)

o2 = (x - i*y)(x + i*y)

o2 : Expression of class Product

i3 : frac(QQ[a])[x,y]; factor(x^2-1/a^2*y^2)

                       / 1\
o4 = (a*x - y)(a*x + y)|--|
                       | 2|
                       \a /

o4 : Expression of class Product
```


M2/Macaulay2/m2/orderedmonoidrings.m2
```diff
@@ -268,7 +268,21 @@ Ring OrderedMonoid := PolynomialRing => (			  -- no memoize
 		   f=f*RM_(-minexps); -- get rid of monomial in factor if f Laurent polynomial
 		   c=RM_minexps;
 		   );
-	       (facs,exps) := rawFactor raw f;	-- example value: ((11, x+1, x-1, 2x+3), (1, 1, 1, 1)); constant term is first, if there is one
+	       isNumberField := F -> isField F and instance(baseRing F, QuotientRing) and coefficientRing baseRing F === QQ and numgens baseRing F == 1 and numgens ideal baseRing F == 1;
```
mahrud, line 271: Would be nice if `NumberField` was a type.

8d1h: I totally agree! In fact I have some prototypical codes for a `NumberField` type. But I don't know if this should be a package or go into the core.

8d1h: Also most advanced things with number fields can be done with PARI/GP, and since the pari library is shipped with M2 (or used to be?), it would be nice to be able to use it.

DanGrayson, line 271: This function doesn't always work:
```m2
Macaulay2, version 1.17.2.1
               MinimalPrimes, PrimaryDecomposition, ReesAlgebra, Saturation, TangentCone

i1 : isNumberField := F -> isField F and instance(baseRing F, QuotientRing) and coefficientRing baseRing F === QQ and numgens baseRing F == 1 and numgens ideal baseRing F == 1;

i2 : K = toField ( QQ[x,y]/(x^2-3, y^2-5) )

o2 = K

o2 : PolynomialRing

i3 : isNumberField K

o3 = false
```

mikestillman:
I also agree that (better) engine support for number fields would be great!  I'd be interested in discussing this with you, to see what you would need.

-- Mike

M2/Macaulay2/m2/orderedmonoidrings.m2
8d1h: Yes... But to call `rawFactor` we need an extension with only one minimal polynomial.

8d1h: Also there is this function from my package that can help in this case.
```macaulay2
i1 : needsPackage "RationalPoints2"                                                                                                                                                            
                                                                                                                                                                                               
o1 = RationalPoints2                                                                                                                                                                           
                                                                                                                                                                                               
o1 : Package                                                                                                                                                                                   
                                                                                                                                                                                               
i2 : QQ[x]; extField{x^2-3, x^2-5}                                                                                                                                                             
                                                                                                                                                                                               
                QQ[a]                                                                                                                                                                          
o3 = --------------------------[]                                                                                                                                                              
      4     3      2                                                                                                                                                                           
     a  - 2a  - 10a  + 26a - 11                                                                                                                                                                
                                                                                                                                                                                               
o3 : PolynomialRing
```

mikestillman: Alas, we cannot use Pari/GP in Macaulay2.  The reason has to do with memory allocation, and the desire to use other packages without patching.  Perhaps we could try it again, but I don't think we can get it to be robust (pari changes the gmp memory allocation functions, which is not compatible with use of gmp in other libraries).  But still, perhaps we can think of some way to get this functionality.

M2/Macaulay2/m2/orderedmonoidrings.m2
```diff
@@ -268,7 +268,21 @@ Ring OrderedMonoid := PolynomialRing => (			  -- no memoize
 		   f=f*RM_(-minexps); -- get rid of monomial in factor if f Laurent polynomial
 		   c=RM_minexps;
 		   );
-	       (facs,exps) := rawFactor raw f;	-- example value: ((11, x+1, x-1, 2x+3), (1, 1, 1, 1)); constant term is first, if there is one
+	       isNumberField := F -> isField F and instance(baseRing F, QuotientRing) and coefficientRing baseRing F === QQ and numgens baseRing F == 1 and numgens ideal baseRing F == 1; 
+	       (facs,exps) := if isNumberField R then (
+        	   baseR := baseRing R;
+		   RM' := QQ(monoid(generators RM|generators baseR)); -- need to put everyone inside the same ring
```
DanGrayson, line 274: Isn't this more easily done with `flattenRing`?

8d1h:
> I also agree that (better) engine support for number fields would be great! I'd be interested in discussing this with you, to see what you would need.
> 
> -- Mike

That would be great, thanks!

M2/Macaulay2/m2/orderedmonoidrings.m2
8d1h: I'm afraid not... `flattenRing` treats the field as a field and does not expand it.
```macaulay2
i12 : first flattenRing(toField(QQ[i]/(i^2+1))[x,y])

       QQ[i]
o12 = ------[][x..y]
       2
      i  + 1

o12 : PolynomialRing
```
Also we need the generator of the field to be the last variable, so in this way we can be sure.

DanGrayson: Try it this way:
```m2
Macaulay2, version 1.17.2.1
               MinimalPrimes, PrimaryDecomposition, ReesAlgebra, Saturation, TangentCone

i1 : K = toField ( QQ[x]/(x^2-3) )

o1 = K

o1 : PolynomialRing

i2 : R = K[y]

o2 = R

o2 : PolynomialRing

i6 : flattenRing (R, CoefficientRing => QQ)

      QQ[y, x]       QQ[y, x]
o6 = (--------, map (--------, R, {y, x}))
        2              2
       x  - 3         x  - 3

o6 : Sequence
```

M2/Macaulay2/m2/orderedmonoidrings.m2
DanGrayson: > Yes... But to call `rawFactor` we need an extension with only one minimal polynomial. Maybe `isSimpleExtension` is a better name.

Or maybe `isSimpleNumberField`.

8d1h: Yes I ran into a lot of these problems with pari... So maybe we should just use the `gp` shell, then everything can stay in M2.

M2/Macaulay2/m2/orderedmonoidrings.m2
8d1h: Yes you are right! Thanks!

M2/Macaulay2/m2/orderedmonoidrings.m2
```diff
@@ -268,7 +268,19 @@ Ring OrderedMonoid := PolynomialRing => (			  -- no memoize
...
+		   minp := (ideal RM')_0;
+		   ((fs,es) -> (for f in fs list raw (map(RM, RM', generators RM|{R_0})) new RM' from f, es)) (
+		       rawFactor(raw toRM' f, raw minp)) -- apply rawFactor, but the factors need to be converted back to RM
+	       ) else if instance(R, FractionField) then (
+        	   denom := lcm \\ (t -> denominator t_1) \ listForm f;
+        	   baseRM := (baseRing R)(RM.monoid);
+		   f = (map(baseRM, RM, generators baseRM)) (denom * f);
+		   ((fs,es) -> (for i in (0..<#fs) list raw (((map(RM, baseRM, generators RM)) new baseRM from fs_i) * if i==0 then 1/denom else 1), es)) (
+		       rawFactor raw f) -- similar: convert back to RM, and put denom back into the leadCoefficient
+	       ) else rawFactor raw f;	-- example value: ((11, x+1, x-1, 2x+3), (1, 1, 1, 1)); constant term is first, if there is one
```
DanGrayson, line 283: Could you add some tests to ensure that the new facility for factoring that you have added to the core continues to work?  They don't have to be slow or large, and a good place for them would be the end of the file `M2/Macaulay2/tests/normal/factor.m2`.

DanGrayson:
It seems that your change to factoring in the core causes test 9 of the InvariantRing package to fail.

The code is this:
```m2
needsPackage "InvariantRing"
R2 = QQ[x_1..x_4]
T2 = diagonalAction(matrix{{0,1,-1,1},{1,0,-1,-1}}, R2)
invariants2 = set {x_1*x_2*x_3,x_1^2*x_3*x_4}
set invariants T2
assert(set invariants T2 === invariants2)
```

And running it gives this:
```m2

i1 : needsPackage "InvariantRing"

o1 = InvariantRing

o1 : Package

i2 : R2 = QQ[x_1..x_4]

o2 = R2

o2 : PolynomialRing

i3 : T2 = diagonalAction(matrix{{0,1,-1,1},{1,0,-1,-1}}, R2)

              * 2
o3 = R2 <- (QQ )  via 

     | 0 1 -1 1  |
     | 1 0 -1 -1 |

o3 : DiagonalAction

i4 : invariants2 = set {x_1*x_2*x_3,x_1^2*x_3*x_4}

                   2
o4 = set {x x x , x x x }
           1 2 3   1 3 4

o4 : Set

i5 : set invariants T2

                   2       4 2 3     5   3 2   6 3 3
o5 = set {x x x , x x x , x x x x , x x x x , x x x }
           1 2 3   1 3 4   1 2 3 4   1 2 3 4   1 3 4

o5 : Set

i6 : assert(set invariants T2 === invariants2)
stdio:6:1:(3): error: assertion failed
```

d-torrance:
> It seems that your change to factoring in the core causes test 9 of the InvariantRing package to fail.

I think it's unlikely that this test failure has anything to do with this PR specifically.  See #1773.

M2/Macaulay2/m2/orderedmonoidrings.m2
8d1h: Sure. I added the two examples above.

8d1h:
> It seems that your change to factoring in the core causes test 9 of the InvariantRing package to fail.

The test seems to work on my branch. Also I don't think it calls `factor` at all...

DanGrayson:
I think Doug is right, thanks, Doug.

I can cause the InvariantRing test to fail simply by creating some functions in the core.

PR #2000, opened 2021-03-20
LocalRings update

mahrud:
See this [project](https://github.com/orgs/Macaulay2/projects/5) for remaining tasks.


mahrud:
@DanGrayson this PR is ready by the way. The only build error is the InvariantRing test, which is unrelated.

I confirmed with @mikestillman that moving the content of `m2/localring.m2` into the `LocalRings` package is fine.

Issue #1924, opened 2021-02-09
Retiring Canned Examples?


pzinn:
I don't quite know what the history of Canned Examples is, but by now, we have two ways of generating examples: either by outputtting them to a .out file, or by using `capture`. I think that should be enough to cover all cases. I philosophically dislike Canned Examples because they're hard-coded and because they're not  mode-agnostic. About the latter point, what I mean is
(1) they look ugly in WebApp mode and
(2) they force us (well, me, really) to write ugly hacks in the handling of example output since `capture` produces examples in the current output mode (which I think is a good thing) whereas CannedExamples are always in standard mode. I will have to submit a PR anyway because https://github.com/Macaulay2/M2/commit/90960b09e1e841f8c643a91f1310b36edac90d46#commitcomment-46835042 broke `capture` in WebApp mode but I don't look forward to having to treat CannedExamples as well; originally when I rewrote `separateM2output` (the function that parses example output) I tried to make it as clean as possible and now it's kind of sloppy again.
So in conclusion I'm ready to go through the Core and check all CannedExamples and see if there's some way to replace them with normal examples. Suggestions of alternatives welcome.

d-torrance:
I think the main purpose of canned examples is for situations when generating the example results in an error, and consequently a build failure.  I can think of at least three reasons why this might happen:

* It makes sense to actually include an error in the example to actually document the feature we're documenting, e.g., `assert`:

```m2
     assert x prints an error message if x isn't true.

     +--------------------------------------------------+
     |  i1 : assert( (2+2) === 4)                       |
     +--------------------------------------------------+
     |  i2 : assert(rank matrix {{1, 2}, {2, 4}} == 2)  |
     |  stdio:2:1:(3): error: assertion failed          |
     +--------------------------------------------------+
```

* The example depends on some program that isn't a Macaulay2 build dependency, e.g., `(html, Digraph)`:

```m2
    Uses graphviz to create an .svg representation of G, which is returned as
     a string.

     +-------------------------------------------------------------------------------------------------------------------------+
     |  i2 : html completeGraph 2                                                                                              |
     |                                                                                                                         |
     |  o2 = <?xml version="1.0" encoding="UTF-8" standalone="no"?>                                                            |
     |       <!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN"                                                                    |
     |        "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">                                                              |
[...]

PR #2022, opened 2021-04-05
minor expression, texMath and html changes; fix separateM2output

pzinn:
- tiny changes to expressions and `texMath`
- a few missing `html` methods were added
- `separateM2output` fixed so it works in both WebApp and Standard mode (see also the discussion in https://github.com/Macaulay2/M2/issues/1924 -- I have not touched CannedExamples in this PR)


DanGrayson:
Thanks!

PR #2035, opened 2021-04-28
Correct Pappus matroid formulation

6x9equals42:
The matroids returned by `specficMatroid "pappus"` and `specificMatroid "nonpappus"` are incorrect.

On the diagrams on page 39 of Oxley (provided below), the Pappus matroid has 9 rank-3 circuits (nonbases) and the non-Pappus has 8. 

image

But the provided versions have 8 & 7 respectively. 
```
i2 : nonbases specificMatroid "pappus"

o2 = {set {0,1,2}, set {0,3,7}, set {0,4,8}, set {1,3,6}, set {1,5,8}, set {2,4,6}, set {2,5,7}, set {6,7,8}}

o2 : List

i3 : nonbases specificMatroid "nonpappus"

o3 = {set {0,1,2}, set {1,3,6}, set {2,4,6}, set {0,3,7}, set {2,5,7}, set {0,4,8}, set {1,5,8}}

o3 :  List
```

I have resolved this by adding {3,4,5} to the set of nonbases in the constructor here: https://github.com/Macaulay2/M2/blob/master/M2/Macaulay2/packages/Matroids.m2#L706.

Then the labels 0:8 on this version map to nodes 3,2,1,4,5,6,7,8,9 in the diagrams.

### DIFF
--- a/M2/Macaulay2/packages/Matroids.m2
+++ b/M2/Macaulay2/packages/Matroids.m2
@@ -703,7 +703,7 @@ specificMatroid String := Matroid => name -> (
 	) else if name == "vamos" then (
 		relaxation(specificMatroid "V8+", set{4,5,6,7})
 	) else if name == "pappus" then (
-		matroid(toList(0..8), {{0,1,2},{0,3,7},{0,4,8},{1,3,6},{1,5,8},{2,4,6},{2,5,7},{6,7,8}}/set, EntryMode => "nonbases")
+		matroid(toList(0..8), {{0,1,2},{0,4,6},{0,5,7},{1,3,6},{1,5,8},{2,3,7},{2,4,8},{3,4,5},{6,7,8}}/set, EntryMode => "nonbases")
 	) else if name == "nonpappus" then (
 		relaxation(specificMatroid "pappus", set{6,7,8})
 	) else if name == "AG32" then (


DanGrayson:
@jchen419 -- how does this look to you?

jchen419:
@DanGrayson: This is correct - in fact, I had already been made aware of this bug a couple months ago (cf. cc1f51e00ef0d857df02853980e069228ac6951b), and implemented fixes for an example as well. @6x9equals42: if you would like to also update the example, then I can approve this pull request - otherwise I plan to merge the commit referenced above.

6x9equals42:
I believe the construction of the non-Pappus matroid (and hence the relevant example) is still correct without any changes - all valid relaxations of the Pappus should be isomorphic, so it shouldn't matter which is used to define the non-Pappus matroid.

jchen419:
Although the change proposed in this pull request is mathematically correct, I think it would be good for the code to reflect e.g. the diagram in Oxley as closely as possible (i.e. if we're doing things correctly, then we don't just want an isomorphism to exist, it should also be "obvious"). Thus I propose that the nonbases be changed to match precisely those given in Oxley, under the "obvious" isomorphism of labels 0:8 -> 1:9. (This would only affect the line defining Pappus - neither non-Pappus nor the example would change from the current distributed version.) @6x9equals42: as before, if you would like to make this change, I would accept it, otherwise I will make this change (along with others) in another pull request.

jchen419:
@6x9equals42: thanks for the fix! @DanGrayson: everything looks good to me now

jchen419 (approved):

PR #2040, opened 2021-05-04
Generate grammar file for Pygments

d-torrance:
Pygments is a Python syntax highlighter.  See https://pygments.org/.  It's also the syntax highlighter used by [minted](https://ctan.org/pkg/minted?lang=en).

For example:

```
profzoom@peg-amy:~/src/macaulay2/M2/M2/Macaulay2/editors/pygments$ pygmentize -O full -x -l macaulay2.py:Macaulay2Lexer -o foo.html ../make-M2-symbols.m2 
```

Creates the following:
https://jsfiddle.net/mj8p67bw/show


DanGrayson:
Thanks!

PR #2046, opened 2021-05-06
COMPILE_FLAGS in C/CXXFLAGS instead of CPPFLAGS

jpoiret:
Put COMPILE_FLAGS in the corresponding C/CXXFLAGS variable, instead of CPPFLAGS which is for pre-processor flags.

This fixes a bug where when `-march=native` was in COMPILE_FLAGS (for example when BUILD_NATIVE is on), `./configure` would fail to pick up the proper architecture in the 4ti2 compilation as autoconf uses non-standard flags ordering and thus `-march=native` would shadow whatever `-march=` flag was being tested, falsifying the result, and leading to a failing compilation.

See https://savannah.gnu.org/support/?110484 for more.

### DIFF
--- a/M2/cmake/configure.cmake
+++ b/M2/cmake/configure.cmake
@@ -195,10 +195,11 @@ endif()
 
 # Flags based on options
 if(MEMDEBUG)
-  add_compile_options(-DMEMDEBUG)
+  add_compile_definitions(MEMDEBUG)
 endif()
 if(PROFILING)
-  add_compile_options(-pg -DPROFILING)
+  add_compile_definitions(PROFILING)
+  add_compile_options(-pg)
   add_link_options(-pg)
 endif()
 
@@ -206,9 +207,11 @@ endif()
 # Note: certain flags are initialized by CMake based on the compiler and build type.
 if(CMAKE_BUILD_TYPE MATCHES "Debug") # Debugging
   # INIT: -g
-  add_compile_options(-O0 -DGC_DEBUG -DMEMT_DEBUG -DMATHIC_DEBUG -DMATHICGB_DEBUG)
+  add_compile_definitions(GC_DEBUG MEMT_DEBUG MATHIC_DEBUG MATHICGB_DEBUG)
+  add_compile_options(-O0)
 else()
-  add_compile_options(-DNDEBUG -DOM_NDEBUG -DSING_NDEBUG -Wuninitialized)
+  add_compile_definitions(NDEBUG OM_NDEBUG SING_NDEBUG)
+  add_compile_options(-Wuninitialized)
 endif()
 if(CMAKE_BUILD_TYPE MATCHES "MinSizeRel")
   # INIT: -Os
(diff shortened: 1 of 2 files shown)


mahrud (changes requested):
If you go through `configure.cmake`, there are a number of calls like
```
  add_compile_options(-DNDEBUG -DOM_NDEBUG -DSING_NDEBUG -Wuninitialized)
```
For instance `-DSING_NDEBUG` is a no-debug flag for Factory. I'm worried that with this change Factory isn't getting this flag during preprocessing. I'm not sure what's the best solution though.

jpoiret:
You're right about that, we should modify them to use `add_compile_definitions` instead. I'll see if I can sort this out cleanly.

mahrud:
Yes, but then can we get a list of currently added definitions from cmake similar to how we can get all compile options with `COMPILE_OPTIONS`?

DanGrayson:
@jpoiret -- do you think you can get the changes completed in the next few days, in time for M2 1.18?

mahrud:
It's not a change that affects the released binaries, so I don't see a reason to rush it.

DanGrayson:
Oh, right, no rush.

jpoiret:
I've replaced `add_compile_options` with `add_compile_definitions` where applicable (I found the cppflags with a simple grep, turns out there weren't that many). I've also added the `COMPILE_DEFINITIONS` variable following what was being done with `COMPILE_OPTIONS` and `LINK_OPTIONS`. I've checked that the proper flags are passed to factory, for example.

This should be enough to properly separate pre-processor and processor flags!

M2/cmake/configure.cmake
```diff
@@ -243,6 +246,7 @@ add_compile_options(
   )
 
 # Querying the options so we can print them
+get_property(COMPILE_DEFINITIONS DIRECTORY PROPERTY COMPILE_DEFINITIONS)
```
mahrud, line 249: Ah, of course. Nice fix!

mahrud (approved):

jpoiret:
I should also note that in the long run, it should be better to add compile, pp and linker options per target rather than for the whole project as is possible with cmake, but this should suffice for now.

mahrud:
> I should also note that in the long run, it should be better to add compile, pp and linker options per target rather than for the whole project as is possible with cmake, but this should suffice for now.

Yes, I agree; only a few sources in the engine include headers for any given library and therefore need the preprocessor flags for that library only. Majority of sources shouldn't need more than a few generic flags.

jpoiret:
Actually, my bad, the flags are then passed to the main compilation processes as `-D-DNDEBUG` for example, which leads to the build failures highlighted above for macaulay2 itself. I'll try to fix this soon, and maybe also tie the flags to their targets.

Do you know if there's somewhere i could find which cppflag is used by which submodule?

mahrud:
> also tie the flags to their targets.

This isn't worth your time right now, and would probably make it harder to maintain the build system.

jpoiret:
Alright, this should now build properly, and pass the right CPPFLAGS to libraries and programs, sorry for the delay.

PR #2054, opened 2021-05-08
Update VirtualResolutions package to v1.4

mahrud:
Main improvement: new, much faster algorithm for `multigradedRegularity` based on #2047.


mahrud:
It's just on top of an already merged pull request. What's the problem?

PR #2056, opened 2021-05-09
Use a nonrecursive algorithms for KBasis

jkyang92:
This fixes a stack overflow I was getting running `basis(200000,QQ[x,y])` while testing other things. Additionally, it seems to provide a ~5-10% speedup for basis with large degrees (on a debug build, I didn't test an optimizing build).

Here's what I'm still planning on doing:

- [x] Reduce code duplication between the various `basis0_*` functions
- [x] Reimplement `basis0_multi_graded` in the same way
- [x] Remove the redundant parameter `firstvar`

I wanted to make this pull request to see if there are any thoughts on these changes. This shouldn't impact #1929, but if there's related changes to make I'd be happy to try to make them.

### DIFF
--- a/M2/Macaulay2/e/matrix-kbasis.cpp
+++ b/M2/Macaulay2/e/matrix-kbasis.cpp
@@ -70,9 +70,14 @@ class KBasis
                   // 0-diml module is asked for
 
   void insert();
-  void basis0_full(int firstvar);
-  void basis0_singly_graded(int firstvar);
-  void basis0_multi_graded(int firstvar);
+  bool try_insert_full();
+  bool try_insert_sg();
+  bool try_insert_mg();
+  bool backtrack(int &curr);
+  bool backtrack_mg(int &curr);
+  void basis0_full();
+  void basis0_singly_graded();
+  void basis0_multi_graded();
 
   KBasis(const Matrix *bottom,
          const int *lo_degree,
@@ -191,149 +196,151 @@ void KBasis::insert()
   if (limit > 0) limit--;
 }
 
-void KBasis::basis0_full(int firstvar)
+inline bool KBasis::backtrack(int &curr)
+{
+  // if we are the end, decrease the last entry to 0
+  if (curr == vars->len - 1)
+    {
+      kb_exp_weight -= var_wts[curr] * kb_exp[vars->array[curr]];
+      kb_exp[vars->array[curr]] = 0;
+      do {
+          curr--;
+      } while (curr >= 0 && kb_exp[vars->array[curr]] == 0);
+    }
+  if (curr < 0) return false;
+  kb_exp[vars->array[curr]]--;
+  kb_exp_weight -= var_wts[curr];
+  curr++;
+  return true;
+}
+
+inline bool KBasis::try_insert_full()
 {
   Bag *b;
-
-  if (kb_monideal->search_expvector(kb_exp, b)) return;
+  if (kb_monideal->search_expvector(kb_exp, b)) return false;
   insert();
-  if (limit == 0) return;
-  if (firstvar >= vars->len) return;
+  return true;
+}
 
-  int curr = firstvar;
+void KBasis::basis0_full()
+{
+  // insert the all zeros vector
+  if (!try_insert_full()) return;
+  if (vars->len == 0) return;
+
+  int curr = 0;
   do {
       int vcurr = vars->array[curr];
       // increase the curr index until we reach a limit
       do {
+          if (limit == 0 || system_interrupted()) return;
           kb_exp[vcurr]++;
           kb_exp_weight += var_wts[curr];
-          if (system_interrupted()) return;
-          if (kb_monideal->search_expvector(kb_exp, b)) break;
-          insert();
-          if (limit == 0) return;
-      } while (true);
-
-      // if we are the end, decrease the last entry to 0
-      if (curr == vars->len - 1)
-        {
-          kb_exp_weight -= var_wts[curr] * kb_exp[vars->array[curr]];
-          kb_exp[vars->array[curr]] = 0;
-          do {
-              curr--;
-          } while (curr >= firstvar && kb_exp[vars->array[curr]] == 0);
-        }
-      if (curr < firstvar) break;
-      kb_exp[vars->array[curr]]--;
-      kb_exp_weight -= var_wts[curr];
-      curr++;
-  } while (true);
+      } while (try_insert_full());
+  } while (backtrack(curr));
 }
 
-void KBasis::basis0_singly_graded(int firstvar)
+inline bool KBasis::try_insert_sg()
 {
   Bag *b;
-  // this bit deals with the all zeros vector
-  if (kb_monideal->search_expvector(kb_exp, b)) return;
+  if (kb_monideal->search_expvector(kb_exp, b)) return false;
   if (hi_degree && kb_exp_weight > kb_target_hi_weight)
     {
       if (do_truncation) insert();
-      return;
+      return false;
     }
-  if (lo_degree == 0 || kb_exp_weight >= kb_target_lo_weight)
-    {
-      insert();
-      if (limit == 0) return;
-    }
-  if (hi_degree && kb_exp_weight == kb_target_hi_weight) return;
+  if (!lo_degree || kb_exp_weight >= kb_target_lo_weight) insert();
+  if (hi_degree && kb_exp_weight == kb_target_hi_weight) return false;
+  return true;
+}
 
-  if (firstvar >= vars->len) return;
-
-  int curr = firstvar;
+void KBasis::basis0_singly_graded()
+{
+  // insert the all zeros vector
+  if (!try_insert_sg()) return;
+  if (vars->len == 0) return;
+  int curr = 0;
   do {
       int vcurr = vars->array[curr];
       // increase the curr index until we reach a limit
       do {
+          if (limit == 0 || system_interrupted()) return;
           kb_exp[vcurr]++;
           kb_exp_weight += var_wts[curr];
-          if (system_interrupted()) return;
-          if (kb_monideal->search_expvector(kb_exp, b)) break;
-          if (hi_degree && kb_exp_weight > kb_target_hi_weight)
-            {
-              if (do_truncation)
-                {
-                  insert();
-                  if (limit == 0) return;
-                }
-              break;
-            }
-          if (lo_degree == 0 || kb_exp_weight >= kb_target_lo_weight)
-            {
-              insert();
-              if (limit == 0) return;
-            }
-          if (hi_degree && kb_exp_weight == kb_target_hi_weight) break;
-      } while (true);
-
-      // if we are the end, decrease the last entry to 0
-      if (curr == vars->len - 1)
-        {
-          kb_exp_weight -= var_wts[curr] * kb_exp[vcurr];
-          kb_exp[vcurr] = 0;
-          do {
-              curr--;
-          } while (curr >= firstvar && kb_exp[vars->array[curr]] == 0);
-        }
-      if (curr < firstvar) break;
-      assert(kb_exp[vars->array[curr]] > 0);
-      kb_exp[vars->array[curr]]--;
-      kb_exp_weight -= var_wts[curr];
-      curr++;
-  } while (true);
+      } while (try_insert_sg());
+  } while (backtrack(curr));
 }
 
-void KBasis::basis0_multi_graded(int firstvar)
+inline bool KBasis::backtrack_mg(int &curr)
+{
+  int vcurr = -1;
+  // if we are the end, decrease the last entry to 0
+  if (curr == vars->len - 1)
+    {
+      vcurr = vars->array[curr];
+      kb_exp_weight -= var_wts[curr] * kb_exp[vcurr];
+      ntuple::multpower(heft_vector->len,
+                        kb_exp_multidegree,
+                        var_degs + (heft_vector->len * vcurr),
+                        -kb_exp[vcurr],
+                        kb_exp_multidegree);
+      kb_exp[vcurr] = 0;
+      do {
+          curr--;
+      } while (curr >= 0 && kb_exp[vars->array[curr]] == 0);
+    }
+  if (curr < 0) return false;
+  vcurr = vars->array[curr];
+  kb_exp[vcurr]--;
+  kb_exp_weight -= var_wts[curr];
+  ntuple::divide(heft_vector->len,
+                 kb_exp_multidegree,
+                 var_degs + (heft_vector->len * vcurr),
+                 kb_exp_multidegree);
+  curr++;
+  return true;
+}
+
+inline bool KBasis::try_insert_mg()
 {
   Bag *b;
-
-  if (system_interrupted()) return;
-  if (kb_monideal->search_expvector(kb_exp, b)) return;
-
+  if (kb_monideal->search_expvector(kb_exp, b)) return false;
   if (kb_exp_weight > kb_target_lo_weight)
     {
       if (do_truncation) insert();
-      return;
+      return false;
     }
-
   if (kb_exp_weight == kb_target_lo_weight)
     {
       if (EQ == ntuple::lex_compare(heft_vector->len,
                                     kb_target_multidegree,
                                     kb_exp_multidegree))
         insert();
-      return;
+      return false;
     }
+  return true;
+}
 
-  for (int i = firstvar; i < vars->len; i++)
-    {
-      int v = vars->array[i];
+void KBasis::basis0_multi_graded()
+{
+  // insert the all zeros vector
+  if (!try_insert_mg()) return;
+  if (vars->len == 0) return;
 
-      kb_exp[v]++;
-      kb_exp_weight += var_wts[i];
-      ntuple::mult(heft_vector->len,
-                   kb_exp_multidegree,
-                   var_degs + (heft_vector->len * v),
-                   kb_exp_multidegree);
-
-      basis0_multi_graded(i);
-
-      kb_exp[v]--;
-      kb_exp_weight -= var_wts[i];
-      ntuple::divide(heft_vector->len,
-                     kb_exp_multidegree,
-                     var_degs + (heft_vector->len * v),
-                     kb_exp_multidegree);
-      if (limit == 0) return;
-    }
+  int curr = 0;
+  do {
+      int vcurr = vars->array[curr];
+      // increase the curr index until we reach a limit
+      do {
+          if (limit == 0 || system_interrupted()) return;
+          kb_exp[vcurr]++;
+          kb_exp_weight += var_wts[curr];
+          ntuple::mult(heft_vector->len,
+                       kb_exp_multidegree,
+                       var_degs + (heft_vector->len * vcurr),
+                       kb_exp_multidegree);
+      } while (try_insert_mg());
+  } while (backtrack_mg(curr));
 }
 
 static bool all_have_pure_powers(const MonomialIdeal *M, M2_arrayint vars)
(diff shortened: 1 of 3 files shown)


mahrud:
There are three things I can think of:
- as I mention in that PR, I think dealing with degree zero variables should be dealt with in the engine:
```m2
basis(0, ZZ/101[a, Degrees => {0}]) -- should give error
basis(0, ZZ/101[a, Degrees => {0}]/ideal (a^3)) -- should be {1 a a2}
```

- I would welcome more clarity on what happens when `do_truncation` is true in this function. Is it assuming a standard grading or multigrading? Is it faster in either case than the algorithm in Truncations package? If not, then probably doesn't hurt to remove `do_truncation`, since I believe we don't ever use it.

- One thing that should be added instead is to have the option `Limit` or `BasisElementLimit` to allow computing a finite number of basis elements of a module that is not finite over the base. Say, `basis(kk[x], Limit => 3)` should probably be `| 1 x x2 |` rather than an error. This might be useful in power series computations.

mahrud:
> the algorithm is essentially doing a depth-first search. This may or may not be what we want.

Ohh is that what the algorithm is doing? Then yeah why not go one step beyond making it non-recursive and also make it a breath-first search?

> In tangentially related topics, I'm also somewhat curious if there's ever a use case for a "resumable" basis computation.

I think so. I tried thinking about this at the top level while working on my PR and have some ideas for how to do it.

jkyang92:
> > the algorithm is essentially doing a depth-first search. This may or may not be what we want.
> 
> Ohh is that what the algorithm is doing? Then yeah why not go one step beyond making it non-recursive and also make it a breath-first search?
> 
I'd rather not change the order of the output of basis if possible, I suspect there will be code that depends on it. Additionally breath-first search is harder to implement, in particular it's a convenient trick that in this case depth-first search can be implemented without a queue or stack. Most of the desired behavior can be obtained by using a call like `basis(0,3,kk[x,y])`

mahrud:
> I don't know why it does what it does, but it seems the Truncation package does eventually call basis with Truncate=>true on at least some codepaths.

I'm making some improvements in Truncations. This can be eliminated fairly easily if need be.

jkyang92:
So I tried implementing the zero degree stuff, but I broke a test in Core (I haven't pushed to code). In particular, the test runs the following
```
A = ZZ/101[a..d, Degrees=>{2:{1,2},2:{0,1}}]
basis({3},A^1)
```
I'm not completely certain what the right behavior here should be. But my current version of the code complains that the 3rd and 4th variables are degree 0 but the module isn't finite in it. 

In theory I think this complaint is correct in that the module is infinite in degrees of the form {3,*}. But doing this would make the partial degree code useless in the most common case of the fine grading on the polynomial ring.

One option would be to gate the new behavior behind some flag, but I don't really know if that's the right thing to do.

mahrud:
By the way, https://github.com/Macaulay2/M2/issues/909 is related to this.

mahrud:
> So I tried implementing the zero degree stuff, but I broke a test in Core (I haven't pushed the code). In particular, the test runs the following
> 
> ```
> A = ZZ/101[a..d, Degrees=>{2:{1,2},2:{0,1}}]
> basis({3},A^1)
> ```
> 
> I'm not completely certain what the right behavior here should be. But my current version of the code complains that the 3rd and 4th variables are degree 0 but the module isn't finite in it.

I think the correct behavior should be to give a basis over the subring `ZZ/101[c,d, Degrees => {2:{0,1}}]`, then it is finite. This is compatible with the partial multidegree behavior I think.

jkyang92:
> I think the correct behavior should be to give a basis over the subring `ZZ/101[c,d, Degrees => {2:{0,1}}]`, then it is finite. This is compatible with the partial multidegree behavior I think.

Then what should be the output of
```
B = ZZ/101[a..d, Degrees=>{2:{1,2},2:{0,1}}]/ideal(c^2,d^2)
basis({3},B^1)
```
In particular, In this case, it does make sense to give an answer over `ZZ/101`, but it seems weirdly inconsistent to do two different things.

mahrud:
I don't understand, what is inconsistent about it? It should ignore `c` and `d` when their degrees are 0 in the first component.

jkyang92:
> I don't understand, what is inconsistent about it? It should ignore `c` and `d` when their degrees are 0 in the first component.

But then why should we do something different if we have `basis({3},QQ[x,y,Degrees=>{1,0}]/ideal(y^2))` vs `basis({3},QQ[x,y,Degrees=>{{1,0},{0,1}}]/ideal(y^2))`. If I understand your proposal, in the first case we should get `|x3 x3y|` and in the second case we get `|x3|`

If we want the basis over the subring of degree zero variables, then what M2 does right now is basically correct.

mahrud:
Oh I see what you mean. I think I interpret the case of partial multidegree different from fixed multidegree with degree zero variables. In other words, I think I expect the following:
```m2
basis({3},   QQ[x,y,Degrees=>{1,0}]/ideal(y^2))         --> | x3 x3y |
basis({3},   QQ[x,y,Degrees=>{{1,0},{0,1}}]/ideal(y^2)) --> | x3 |
basis({3,0}, QQ[x,y,Degrees=>{{1,0},{0,1}}]/ideal(y^2)) --> | x3 x3y |
```
Does that make sense?

mahrud:
This is also the point of the option `[basis, Variables]`, so it would also make sense to change the current behavior to completely _disallow_ `{3}` for a bigraded ring, but still allow this for a partial degree basis computation:
```m2
basis({3, 0}, QQ[x,y,Degrees=>{{1,0},{0,1}}]/ideal(y^2), Variables => {0}) --> | x3 |
```
I think this is ideal, because then accidental errors involving incorrect degree length are caught and you can even do this:
```m2
basis({0, 3}, QQ[x,y,Degrees=>{{1,0},{0,1}}]/ideal(x^2), Variables => {1}) --> | y3 |
```
which you can't do using the current partial-multidegree inputs currently.

We could also allow something like `basis({3, *}, R)` (like, including the asterisk in the input), and remove the `Variables` option.

By the way, that last line currently crashes for me, does it crash for you?
```
terminate called after throwing an instance of 'exc::overflow_exception'
  what():  monomial exponent overflow: int32_t + int32_t
```

jkyang92:
> Oh I see what you mean. I think I interpret the case of partial multidegree different from fixed multidegree with degree zero variables. In other words, I think I expect the following:
> 
> ```m2
> basis({3},   QQ[x,y,Degrees=>{1,0}]/ideal(y^2))         --> | x3 x3y |
> basis({3},   QQ[x,y,Degrees=>{{1,0},{0,1}}]/ideal(y^2)) --> | x3 |
> basis({3,0}, QQ[x,y,Degrees=>{{1,0},{0,1}}]/ideal(y^2)) --> | x3 x3y |
> ```
> 
> In the line 2 above should give stuff in degree `{3, *}`, so in order to have any chance in general, the algorithm should ignore the other variables, but in line 3 it's always able to give everything of degree `{3, 0}`.
> 
> Does that make sense?

This makes sense, although a bit odd in my opinion, and requires a special case to get right.

> This is also the point of the option `[basis, Variables]`, so it would also make sense to change the current behavior to completely _disallow_ `{3}` for a bigraded ring, but still allow this for a partial degree basis computation:
> 
> ```m2
> basis({3, 0}, QQ[x,y,Degrees=>{{1,0},{0,1}}]/ideal(y^2), Variables => {0}) --> | x3 |
> ```
> 

This is useful, but the partial degree stuff can give an answer for `basis({3}, QQ[x,y,Degrees=>{{1,1},{1,2}}])` which you can't simulate with the current use of the `Variables` option. How useful this is is another question though.

> 
> We could also allow something like `basis({3, *}, R)` (like, including the asterisk in the input), and remove the `Variables` option.

This would need parser help I think, it's a syntax error right now. I guess you could do `symbol *`. Personally, I'm also partial to using `_`, but again it's a syntax error. We could also use `null` instead, and that wouldn't require a parser change. Either way it would also need engine changes to support the general case where the `*` can be anywhere.

> 
> By the way, that last line currently crashes for me, does it crash for you?
> 
> ```
> terminate called after throwing an instance of 'exc::overflow_exception'
>   what():  monomial exponent overflow: int32_t + int32_t
> ```

Crashes for me as well, no idea what causes it, I'll take a look.

mahrud:
> This makes sense, although a bit odd in my opinion, and requires a special case to get right.

I presumed handling partial multidegrees was already a special case, so I thought dealing with degree zero variables when a specific degree is giving would be the default case.

Good point about the difference between `Variables` and partial multidegree input. I'll think more about it.

Also good point about the parser error. We can enter `{3, }` instead of null, which is good enough. I've also thought about adding a `Multidegree` type several times, but this all can be dealt with separately.

mahrud:
I cherry-picked your commits on top of my `basis.m2` changes in #2113 and it looks like the tests are passing. The new branch is https://github.com/mahrud/M2/tree/refactor/basis.

There are a couple of things that don't seem to be working correctly:
* basis with `Truncate => true` is giving more generators than before
```m2
basis(2, QQ[a,b,c, Degrees => {1,2,3}], Truncate => true)
-- used to be: | a2 ab ac b c |
-- now giving: | a2 a3 a2b a2c ab ac b b2 bc c |
-- after trim: | b a2 c |
```
The answer gets trimmed anyway, but I think somewhere in the algorithm you're doing too much work. In particular, after `a2` is added, there's no reason to add `a3 a2b a2c`, and after `b` there's no reason to add `b2` and `bc`. I would say more efficient truncations is a solid reason to switch to BFS, even if there are breaking changes. I can potentially help with this, if you'd like.

* Partial multidegrees this was mentioned before. I don't think it's urgent, if you'd like to leave it as it is for now.
```m2
basis({3},   QQ[x,y,Degrees=>{1,0}]/ideal(y^2))         --> | x3 x3y |
basis({3},   QQ[x,y,Degrees=>{{1,0},{0,1}}]/ideal(y^2)) --> | x3 | (correct)
basis({3,0}, QQ[x,y,Degrees=>{{1,0},{0,1}}]/ideal(y^2)) --> | x3 x3y |
```

jkyang92:
> There are a couple of things that don't seem to be working correctly:
> 
>     * basis with `Truncate => true` is giving more generators than before
> 
> 
> ```m2
> basis(2, QQ[a,b,c, Degrees => {1,2,3}], Truncate => true)
> -- used to be: | a2 ab ac b c |
> -- now giving: | a2 a3 a2b a2c ab ac b b2 bc c |
> -- after trim: | b a2 c |
> ```
> 
> The answer gets trimmed anyway, but I think somewhere in the algorithm you're doing too much work. In particular, after `a2` is added, there's no reason to add `a3 a2b a2c`, and after `b` there's no reason to add `b2` and `bc`. I would say more efficient truncations is a solid reason to switch to BFS, even if there are breaking changes. I can potentially help with this, if you'd like.

Truncate giving more stuff is due to a change related to the degree zero stuff, as I said, I don't really understand why it does what it's supposed to do, but I should be able to simplify that at least somewhat.

 >     * Partial multidegrees this was mentioned before. I don't think it's urgent, if you'd like to leave it as it is for now.
> 
> 
> ```m2
> basis({3},   QQ[x,y,Degrees=>{1,0}]/ideal(y^2))         --> | x3 x3y |
> basis({3},   QQ[x,y,Degrees=>{{1,0},{0,1}}]/ideal(y^2)) --> | x3 | (correct)
> basis({3,0}, QQ[x,y,Degrees=>{{1,0},{0,1}}]/ideal(y^2)) --> | x3 x3y |
> ```

Isn't this correct as the current output?

Also, I though there were a couple test failures still or did they get indirectly fixed via your commits?

mahrud:
> Truncate giving more stuff is due to a change related to the degree zero stuff, as I said, I don't really understand why it does what it's supposed to do, but I should be able to simplify that at least somewhat.

I think I fully understand what the basis-free algorithm for non-standard graded rings in Truncations does now. It's a polyhedral algorithm, which we might be able to implement in the C++ only for the standard graded case. I believe it's fairly efficient.

> Isn't this correct as the current output?

Sorry, I had a typo in there. This is what I think should be correct:
```m2
basis({3},   QQ[x,y,Degrees=>{1,0}]/ideal(y^2))         --> | x3 x3y |
basis({3},   QQ[x,y,Degrees=>{{1,0},{0,1}}]/ideal(y^2)) --> | x3 |
basis({3,0}, QQ[x,y,Degrees=>{{1,0},{0,0}}]/ideal(y^2)) --> | x3 x3y |
```
Only the second one is correct right now. I believe the first and third should add `x3y` because in those cases the degree given is a full degree, not partial, so it should give a basis over `QQ` rather than `QQ[y]`.

> Also, I though there were a couple test failures still or did they get indirectly fixed via your commits?

All tests [pass](https://github.com/mahrud/M2/actions/runs/894376430) now. My commits fixed quite a few bugs in the top level code of basis, as well as point out several ambiguities or bugs in [this test file](https://github.com/Macaulay2/M2/blob/refactor/basis/M2/Macaulay2/tests/normal/basis3.m2), for instance:
https://github.com/Macaulay2/M2/blob/654564ed8769ad0e8c608e7d5153db9a7533e9be/M2/Macaulay2/tests/normal/basis3.m2#L40-L42

jkyang92:
Ah, at some point, you lost the commit https://github.com/Macaulay2/M2/pull/2056/commits/a0e48acdf5576ccf5d21c6abc85304d80384d844, Also you lost the changes in https://github.com/Macaulay2/M2/pull/2056/commits/eff87a63be4733ce1f3b9a014833ae7ee96d951c that affect modules2.m2 Those are the ones that makes partial multidegrees work correctly, and also the ones that break at least the NotherianOperators packages, since it expects the opposite behavior. Probably these got lost because they interact in non-trivial ways with some other changes you've made.

mahrud:
> Ah, at some point, you lost the commit [a0e48ac](https://github.com/Macaulay2/M2/commit/a0e48acdf5576ccf5d21c6abc85304d80384d844), Also you lost the changes in [eff87a6](https://github.com/Macaulay2/M2/commit/eff87a63be4733ce1f3b9a014833ae7ee96d951c) that affect modules2.m2 Those are the ones that makes partial multidegrees work correctly, and also the ones that break at least the NotherianOperators packages, since it expects the opposite behavior. Probably these got lost because they interact in non-trivial ways with some other changes you've made.

I thought the changes to `modules2.m2` would be covered by my changes to `findHeftandVars`. In particular, in the main strategy to basis I just call:
https://github.com/Macaulay2/M2/blob/654564ed8769ad0e8c608e7d5153db9a7533e9be/M2/Macaulay2/m2/basis.m2#L285

I'll take another look at that function. Could you give me examples of dealing with partial multidegrees that don't work on `refactor/basis`?

mahrud:
Never mind, my own tests can be used for this ... I'll fix it in my branch.

mahrud:
Okay, I think I figured out how to have cake and eat it to:
https://github.com/Macaulay2/M2/blob/c6f3692b31b9edbc24db11e96e1d3aa6402f2b03/M2/Macaulay2/m2/basis.m2#L61-L66

I don't think the codimension computation would take any significant time, but it allows us to tell when a finite answer over the degree zero variables is possible (so `varlist` should be passed entirely) and when it is not finite, in which case we should fallback to the old behavior and ignore the degree zero variables:
```m2
i23 : basis(0, A = ZZ/101[a, Degrees => {0}])

o23 = | 1 |

              1       1
o23 : Matrix A  <--- A

i24 : basis(0, A = ZZ/101[a, Degrees => {0}]/ideal(a^3))

o24 = | 1 a a2 |

              1       3
o24 : Matrix A  <--- A
```
The tests are running [here](https://github.com/mahrud/M2/actions/runs/897386355), but I'm pretty optimistic that they'll pass.

jkyang92:
I've been thinking about this, and I have a few thoughts.

Perhaps implicit partial multidegrees are an anti-feature. I guess they make some sense when doing something like `basis(3,QQ[x][y])` and perhaps we can support this particular case. If we remove implicit partial multidegrees, we should support general explicit partial multidegrees. This isn't too hard, but it does require some care.

I think we should either always work over `coefficientRing R` or `coefficientRing flattenRing R` and be globally consistent about this rather than trying to give some weird mix of cases. Given that the default grading is as a `coefficientRing flattenRing R`-algebra, I would say we should probably work over that rather than `coefficientRing R`. Of course, we need to give a user a way to override this behavior, and it seems to me that the `Variables` parameter we already have should be sufficient for this. although I'm open to some sort of `CoefficientRing` parameter if someone can figure out a consistent way to do it.

One thing I always found slightly suboptimal about the current `basis` command is that there's not an easy way to do total degrees on a multigraded setting. I have some ideas on how one might do this (supply a DegreeMap parameter or something), but I suspect this is overcomplicating the problem, at least for now.

mahrud:
> I think we should either always work over `coefficientRing R` or `coefficientRing flattenRing R` and be globally consistent about this rather than trying to give some weird mix of cases. Given that the default grading is as a `coefficientRing flattenRing R`-algebra, I would say we should probably work over that rather than `coefficientRing R`.

I partially agree here: I think by default the output should be a basis or generating set over `basis(0, first flattenRing R)`, which should by definition give `| 1 |`. In the case of positively graded variables, this will reduce to `coefficientRing flattenRing R`, but it might not in general.

> Of course, we need to give a user a way to override this behavior, and it seems to me that the `Variables` parameter we already have should be sufficient for this. although I'm open to some sort of `CoefficientRing` parameter if someone can figure out a consistent way to do it.

Perhaps a different method call, like `basis_K(0, M)` might be good for this, and would involve a pushforward. For instance, I understand why this doesn't work, but I think it should:
```m2
i1 : R = kk[x];

i2 : pushForward(map(R, kk), image basis(0, module R))
stdio:2:1:(3): error: module given is not finite over the base
```
Even using the `DegreeLimit` option it doesn't work:
```m2
i6 : pushForward(map(R, kk), module R, DegreeLimit => 0)
stdio:6:1:(3): error: module given is not finite over the base
```

> One thing I always found slightly suboptimal about the current `basis` command is that there's not an easy way to do total degrees on a multigraded setting. I have some ideas on how one might do this (supply a DegreeMap parameter or something), but I suspect this is overcomplicating the problem, at least for now.

I've wanted a `DegreeMap` option, too. All of this would be much simpler to develop if `basis` didn't belong in the Core, but extracting it was too difficult when I tried last. There are not too many Core scripts that depend on it, though, so it should be possible.

jkyang92:
> I partially agree here: I think by default the output should be a basis or generating set over `basis(0, first flattenRing R)`, which should by definition give `| 1 |`. In the case of positively graded variables, this will reduce to `coefficientRing flattenRing R`, but it might not in general.

I guess this is the point where I'm thinking about things differently. If it's a `coefficientRing first flattenRing R`-algebra, there's no reason that `basis(0, first flattenRing R)` must be `|1|`. On the otherhand, if you start with the grading, and say that R is an R_0-algebra (math R_0 here, not M2 R_0), then it certainly is the case that `basis(0, first flattenRing R)` is definitionally `|1|`. 

I prefer the first interpretation, but in the end either is acceptable. If we do the second one, the user needs a way talk about the degree 0 subring. I'm also uncertain about what the interaction between the second option and partial multidegrees is.

mahrud:
I think the R_0-module interpretation is the motivation behind how degree zero variables are treated, and how partial multidegrees treat some variables as having degree zero, so I think it would be consistent. For instance, with this interpretation, `basis(3, QQ[x][y])` means treat x as a degree zero variable and give a generating set over `QQ[x]`. In other words, these two would mean the same thing:
```m2
i1 : basis(3, QQ[x][y])

o1 = | y3 |

i2 : basis({3,0}, QQ[y,x, Degrees => {{1,0},{0,0}}])

o2 = | y3 |
```

This is definitely not in line with the issues I raised about `basis` in the beginning of this thread, because I think I understand what the reasoning was, and so I think we need fewer changes to the code for degree zero variables (a few edge cases maybe) and more changes to the documentation.

mahrud:
I believe if you undo the changes to `modules2.m2`, the tests will pass and this PR can be merged.

PR #2070, opened 2021-05-13
New version of `combine`

moorewf:
This pull request adds a new version of combine which allows for the resulting value of the product to include a factor which depends on the keys.  This may be used to implement (and test) divided powers and skew polynomial rings at top-level.


DanGrayson:
I've seen this cmake-macos failure before:
```
sh: line 1:  6713 Abort trap: 6           /usr/local/bin/gfan --help
```

It's unrelated to this pull request.

Issue #2071, opened 2021-05-13
Invalid html in "Working with Tor" docs


d-torrance:
I saw this on a local build of the Debian package:

```
validating: Complexes/html/___Working_spwith_sp__Tor.html
 *** invalid HTML: /build/macaulay2-1.17.2.1+git1187.b88ac7b+ds/M2/usr-dist/common/share/doc/Macaulay2/Complexes/html/___Working_spwith_sp__Tor.html
info: line 53: S separator in comment declaration
error: line 53: invalid comment declaration: found name start character outside comment but inside comment declaration
info: line 53: comment declaration started here
info: line 55: S separator in comment declaration
error: line 55: invalid comment declaration: found name start character outside comment but inside comment declaration
info: line 55: comment declaration started here
info: line 65: S separator in comment declaration
error: line 65: invalid comment declaration: found name start character outside comment but inside comment declaration
info: line 65: comment declaration started here
```

Here are the lines in question:
```html
root@peg-amy:/build/macaulay2-1.17.2.1+git1187.b88ac7b+ds# sed -n -e 53p -e 55p -e 65p M2/usr-dist/common/share/doc/Macaulay2/Complexes/html/___Working_spwith_sp__Tor.html 
<span><tt>(Tor, ZZ, Module, Matrix) -- not yet implemented</tt> (missing documentation<!-- tag: (Tor, ZZ, Module, Matrix) -- not yet implemented -->)</span>            </li>
<span><tt>(Tor, ZZ, Matrix, Module) -- not yet implemented</tt> (missing documentation<!-- tag: (Tor, ZZ, Matrix, Module) -- not yet implemented -->)</span>            </li>
<span><tt>symmetry of Tor -- not yet implemented</tt> (missing documentation<!-- tag: symmetry of Tor -- not yet implemented -->)</span>            </li>
```

It looks like the problems are the `--` inside the comments.  See https://stackoverflow.com/questions/3320350/why-does-the-html-validator-give-me-an-error-about-this-html-comment.

PR #2076, opened 2021-05-15
Small tweaks to html comments

d-torrance:
Two little tweaks to prevent issues like #2071 going forward:

* Use `(html, COMMENT)` for undocumented and missing tags instead of hard-coding `<!--` and `-->`.
* Raise an error in `(html, COMMENT)` if the comment contains a `--`. That way we'll know there's an issue right off the bat instead of getting html validation errors later down the road.  Note that `--` in a comment is invalid XML 1.0 -- see https://www.w3.org/TR/REC-xml/#sec-comments.

### DIFF
--- a/M2/Macaulay2/m2/html.m2
+++ b/M2/Macaulay2/m2/html.m2
@@ -139,7 +139,9 @@ html TT :=
 html CODE := (lookup(html, Hypertext)) @@ (x -> apply(x,fixNewLines))
 
 html CDATA   := x -> concatenate("<![CDATA[", x ,"]]>", newline)
-html COMMENT := x -> concatenate("<!--", x, "-->", newline)
+html COMMENT := x -> if match("--", concatenate x) then
+    error ///html comments cannot contain "--"/// else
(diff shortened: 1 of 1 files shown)


DanGrayson:
Why not just translate `--` to `&#45;&#45;` internally, and not give an error?

d-torrance:
> Why not just translate `--` to `&#45;&#45;` internally, and not give an error?

Comments aren't rendered by the browser, so no one would ever actually see those as hyphens.  I suppose we could do something like [tidy's `fix-bad-comments` option](https://api.html-tidy.org/tidy/quickref_next.html#fix-bad-comments) and convert the offending `-`'s to `=`'s, but personally, I'd rather get an error and choose the alternative myself.

Of course, in the long term, we should probably move away from XML 1.0 to HTML5 docs and this will be a non-issue.

mahrud:
> Of course, in the long term, we should probably move away from XML 1.0 to HTML5 docs and this will be a non-issue.

Highly in favor of this. The documentation is for humans, not machines.

PR #2083, opened 2021-05-19
Mention "assert" changes in changes.m2 for 1.18

d-torrance:
The new feature may be worth mentioning:

```m2
i1 : assert Equation(2 + 2, 5)
stdio:1:1:(3): error: assertion failed:
4 == 5 is false
```


d-torrance:
Oops, made a typo so I force-pushed a new commit.  ("a Expression" instead of "an Expression" -- I was fooled by the T in TO!)

d-torrance:
Somehow a stray comma ended up in that last commit so I force-pushed it again...  I probably shouldn't do these things so early in the morning lol.

DanGrayson:
Thanks!  I'll add a bit more clarification of the concept...

DanGrayson:
I had this possibility in mind:
```m2

i1 : d=3

o1 = 3

i2 : assert(Equation{symbol d,3})

i3 : d=4

o3 = 4

i4 : assert(Equation{symbol d,3})
stdio:4:1:(3): error: assertion failed:
d == 3 is false
```

d-torrance:
Ah, ok, that makes sense.

PR #2084, opened 2021-05-19
Create symlinks instead of copying Javascript files in Visualize

d-torrance:
The copyJS function was failing to copy any files that happened to be
symbolic links, which was the case in the Debian/Ubuntu package
because many of these files were symbolic links to the packaged
versions of the corresponding Javascript libraries.

Instead, we create symbolic links to the directories containing these
files.  This also saves disk space since we won't be duplicating these
files for every call to "visualize".

We also ensure that the destination directory ends in a trailing slash
and create it if it doesn't already exist.


d-torrance:
No big rush on this for 1.18, since it's not really a big issue upstream.  I can always just apply it as a patch to the 1.18 Debian package.

DanGrayson:
Authors ( @bstone ) : is this change okay with you?

d-torrance:
Just pushed a new version of this commit with a couple minor updates

PR #2093, opened 2021-05-23
CMake maintenance

mahrud:


mahrud:
@DanGrayson before you make the 1.18 release, please make a release candidate tag `release-1.18-rc1` so I can test building the brew bottle with it. Otherwise if any additional changes are needed to make the brew build work after `release-1.18` is tagged, you'll need to tag them with `release-1.18.1` for the brew formula.

mahrud:
@d-torrance do you know what is happening with `findProgram` and `runProgram`? Making a bottle for brew doesn't require all the tests to pass, but it does install all the packages, so this failure would probably prevent the brew build from finishing.

Have you tried replacing `gfan` in the example with another program? That might indicate whether the problem is with `gfan` or with `findProgram`.

d-torrance:
> @d-torrance do you know what is happening with `findProgram` and `runProgram`? Making a bottle for brew doesn't require all the tests to pass, but it does install all the packages, so this failure would probably prevent the brew build from finishing.
> 
> Have you tried replacing `gfan` in the example with another program? That might indicate whether the problem is with `gfan` or with `findProgram`.

I'll look into it.

d-torrance:
Something's wrong with the copy of gfan in the `cmake-macos-10.15-clang-11` build.  From the artifacts from https://github.com/Macaulay2/M2/actions/runs/869703819:

```m2
 -- /usr/local/bin/gfan exists and is executable. running ...
dyld: Symbol not found: __ZNKSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEE7compareEPKc
  Referenced from: /usr/local/bin/gfan
  Expected in: /usr/lib/libstdc++.6.dylib
 in /usr/local/bin/gfan
 -- return value: 6
```

PR #2095, opened 2021-05-24
a few typos in packages

fchapoton:
just a few simple typoes


DanGrayson:
Thanks!  Sorry for not catching that.

PR #2098, opened 2021-05-25
Restore M2-emacs commit

d-torrance:
It appears to have been accidentally reverted in #2095.


DanGrayson:
Thanks! Sorry for not catching that.

PR #2099, opened 2021-05-25
Load package documentation in "tests"

d-torrance:
*EDIT:*  This PR now only deals with the second issue mentioned in this original post, and doesn't touch `needsPackage` at all.

---

This fixes two small bugs when calling `needsPackage` twice on the same package but with different options:

### Before
```m2
i1 : needsPackage "Graphs"
--warning: symbol "diameter" in Core.Dictionary is shadowed by a symbol in Graphs.Dictionary
--  use the synonym Core$diameter

o1 = Graphs

o1 : Package

i2 : needsPackage("Graphs", FileName => "~/tmp/Graphs.m2")
tmp/Graphs.m2:27:1:(3):[10]: error: package Graphs not reloaded; try Reload => true
tmp/Graphs.m2:27:1:(3):[10]: --entering debugger (type help to see debugger commands)
tmp/Graphs.m2:27:1-50:23: --source code:
newPackage select((
    "Graphs",
        Version => "0.3.4",
        Date => "May 15, 2021",
        Authors => {
            {Name => "Jack Burkart", Email => "jburkar1@nd.edu"},
            {Name => "David Cook II", Email => "dwcook@eiu.edu", HomePage => "http://ux1.eiu.edu/~dwcook/"},
            {Name => "Caroline Jansen", Email => "cjansen@nd.edu"},
                {Name => "Amelia Taylor", Email => "originalbrickhouse@gmail.com"},
            {Name => "Augustine O'Keefe", Email => "aokeefe@tulane.edu"},
            {Name => "Contributors of note: Carlos Amendola, Alex Diaz, Luis David Garcia Puente, Roser Homs Pons, Olga Kuznetsova,  Shaowei Lin, Sonja Mapes, Harshit J Motwani, Mike Stillman, Doug Torrance"}
        },
        Headline => "graphs and directed graphs (digraphs)",
        Keywords => {"Graph Theory"},
        Configuration => {
            "DotBinary" => "dot",
            "JpgViewer" => ""
            },
        PackageImports => { "PrimaryDecomposition" },
        PackageExports => {
            "SimplicialComplexes"
            },
        DebuggingMode => false,
        ), x -> x =!= null)
```

### After
```m2
i1 : needsPackage "Graphs"
--warning: symbol "diameter" in Core.Dictionary is shadowed by a symbol in Graphs.Dictionary
--  use the synonym Core$diameter

o1 = Graphs

o1 : Package

i2 : needsPackage("Graphs", FileName => "~/tmp/Graphs.m2")
[...]

PR #2100, opened 2021-05-25
"diameter" in Core and Graphs

d-torrance:
This fixes a warning message I noticed while working on #2099.  Since #1465 was merged, there have been two exported `diameter` symbols -- one from `Core` (a compiled function that expects `RRi` objects) and one from `Graphs` (a method function with one installed method -- `(diameter, Graph)`) that shadows the one from `Core`:

```m2
i1 : needsPackage "Graphs"
--warning: symbol "diameter" in Core.Dictionary is shadowed by a symbol in Graphs.Dictionary
--  use the synonym Core$diameter
```

These have been united so that `diameter` is declared as a method function in `Core`, and two methods are installed -- `(diameter, RRi)` in `Core` and `(diameter, Graph)` in `Graphs`:

```m2
i1 : needsPackage "Graphs"

o1 = Graphs

o1 : Package

i2 : methods diameter

o2 = {0 => (diameter, Graph)}
     {1 => (diameter, RRi)  }

o2 : NumberedVerticalList
```

### DIFF
--- a/M2/Macaulay2/packages/Graphs.m2
+++ b/M2/Macaulay2/packages/Graphs.m2
@@ -144,7 +144,6 @@ export {
     "DFS",
     "descendants",
     "descendents",
-    "diameter",
     "distance",
     "distanceMatrix",
     "eccentricity",
@@ -897,7 +896,6 @@ descendants = method()
 descendants (Digraph, Thing) := Set => (D,v) -> set flatten breadthFirstSearch(D, v)
 descendents = descendants
 
-diameter = method()
 diameter Graph := ZZ => G -> (
     allEntries := flatten entries distanceMatrix G;
     if member(-1, allEntries) then infinity else max allEntries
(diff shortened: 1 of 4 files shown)


M2/Macaulay2/packages/Graphs.m2
```diff
@@ -3434,7 +3432,6 @@ doc ///
 --diameter
 doc ///
     Key
-        diameter
```
DanGrayson, line 3437: We probably have to also add some way for users of "Graphs" to discover this documentation node, by linking to it from another suitable one.

d-torrance: There were three other nodes that pointed to `diameter` under `SeeAlso`, so I've just pushed another commit updating those to `(diameter, Graph)`.

It's also accessible under the main `Graphs` page, plus the index and table of contents.

PR #2104, opened 2021-05-26
fix a minor bug in SpecialFanoFourfolds

giovannistagliano:


giovannistagliano:
Would it be possible to include this latest update in the v1.18?

DanGrayson:
> Would it be possible to include this latest update in the v1.18?

No, 1.18 is out the door, sorry!

giovannistagliano:
> > Would it be possible to include this latest update in the v1.18?
> 
> No, 1.18 is out the door, sorry!

Ok, no problem.

PR #2106, opened 2021-05-26
update emacs submodule hash

mahrud:


mahrud:
@DanGrayson this updates syntax highlighting for v1.18.

DanGrayson:
> @DanGrayson this updates syntax highlighting for v1.18.

I didn't know I should have waited for this, and have already made a distribution, for Ubuntu 21.04.  Okay, I'll redo it.

d-torrance:
> I didn't know I should have waited for this, and have already made a distribution, for Ubuntu 21.04.

Does this mean that the stable 1.18 release is coming very soon?  And will it be tagged in git?

I was thinking of making 1.18-rc1 packages for Debian experimental and the Ubuntu PPAs, but perhaps I'll just wait for 1.18.

DanGrayson:
Yes, I tagged 1.18-rc1 for Mahrud, updated the version number to 1.18 on a release branch, and have started building for various machines.

DanGrayson:
> > @DanGrayson this updates syntax highlighting for v1.18.
> 
> I didn't know I should have waited for this, and have already made a distribution, for Ubuntu 21.04. Okay, I'll redo it.

Probably emacs should not be a submodule, since it depends on the other files.  Submodules should be independent.

mahrud:
> > @DanGrayson this updates syntax highlighting for v1.18.
> 
> I didn't know I should have waited for this, and have already made a distribution, for Ubuntu 21.04.  Okay, I'll redo it.

Oh it's possible that your packaging scripts copy the updated emacs files into it already, so not necessarily. You should check the header of the files for the version.

DanGrayson:
> > > @DanGrayson this updates syntax highlighting for v1.18.
> > 
> > 
> > I didn't know I should have waited for this, and have already made a distribution, for Ubuntu 21.04.  Okay, I'll redo it.
> 
> Oh it's possible that your packaging scripts copy the updated emacs files into it already, so not necessarily. You should check the header of the files for the version.

Maybe not:

```
;; Auto-generated for Macaulay2-1.17. Do not modify this file manually.
```

DanGrayson:
> > > > @DanGrayson this updates syntax highlighting for v1.18.
> > > 
> > > 
> > > I didn't know I should have waited for this, and have already made a distribution, for Ubuntu 21.04.  Okay, I'll redo it.
> > 
> > 
> > Oh it's possible that your packaging scripts copy the updated emacs files into it already, so not necessarily. You should check the header of the files for the version.
> 
> Maybe not:
> 
> ```
> ;; Auto-generated for Macaulay2-1.17. Do not modify this file manually.
> ```

Oops, no, never mind -- wrong system.

d-torrance:
> Probably emacs should not be a submodule, since it depends on the other files. Submodules should be independent.

+1

The Debian packaging would be simpler if we were able to download a tarball from Github (using the `uscan` utility) and use that.  Right now, I need to use a script that merges the contents of the `M2` and `M2-emacs` repositories into a single tarball.

d-torrance:
> Yes, I tagged 1.18-rc1 for Mahrud, updated the version number to 1.18 on a release branch, and have started building for various machines.

Ok, great!  Once there's a 1.18 tag, I'll prepare a Debian package (for upload to experimental -- they're still in freeze for Debian 11 "bullseye") and PPA packages.

mahrud:
> Probably emacs should not be a submodule, since it depends on the other files. Submodules should be independent.

What does it depend on? That submodule is an independent emacs major mode. The scripts for generating it, of course, are in the editors directory. The point of making it a separate submodule was to create a separate emacs package.

> The Debian packaging would be simpler if we were able to download a tarball from Github (using the `uscan` utility) and use that.  Right now, I need to use a script that merges the contents of the `M2` and `M2-emacs` repositories into a single tarball.

Isn't the emacs package a separate package? Why do you need to merge their contents? I believe almost all other packaged languages have separate repositories for the compiler and the emacs mode, don't they?

d-torrance:
> > The Debian packaging would be simpler if we were able to download a tarball from Github (using the `uscan` utility) and use that.  Right now, I need to use a script that merges the contents of the `M2` and `M2-emacs` repositories into a single tarball.
> 
> Isn't the emacs package a separate package? Why do you need to merge their contents? I believe almost all other packaged languages have separate repositories for the compiler and the emacs mode, don't they?

Yes, `elpa-macaulay2` a separate *binary* package, but it's generated from the same *source* package as the other two binary package (`macaulay2` and `macaulay2-common`).

Currently, the two repositories are very tied together.  The files contained in `M2-emacs` are generated during the `M2` build process, and you can't do an `M2` build (at least for autotools) without first initializing the `M2-emacs` submodule.

An alternate solution, I think, would be to separate them further, so that there was a `Makefile` in `M2-emacs` that used an existing installation of Macaulay2 to generate its files and it could exist completely on its own.  It could still live as a submodule in `M2` I suppose, but make it optional.  Then I could split it off into its own source package in Debian.

DanGrayson:
> > > @DanGrayson this updates syntax highlighting for v1.18.
> > 
> > 
> > I didn't know I should have waited for this, and have already made a distribution, for Ubuntu 21.04.  Okay, I'll redo it.
> 
> Oh it's possible that your packaging scripts copy the updated emacs files into it already, so not necessarily. You should check the header of the files for the version.

You're right, thanks:
```
;; Auto-generated for Macaulay2-1.18. Do not modify this file manually.
```

mahrud:
You can make the package from the M2-emacs repository alone instead of generating it along with M2. And it should be easy to prevent autotools from needing emacs files.

d-torrance:
> You can make the package from the M2-emacs repository alone instead of generating it along with M2. And it should be easy to prevent autotools from needing emacs files.

Yeah, that's basically the "alternate solution" I mentioned above.  But a hypothetical `elpa-macaulay2` Debian source package would need to be able to re-generate all the generated files during build to be in line with Debian policy.  I could do all these things in the Debian packaging files, I suppose, but I'd rather there be an upstream build script that I could use.

PR #2124, opened 2021-05-31
Copy all Visualize assets into temporary directory (again)

d-torrance:
We replace the copyJS method (which originally copied the assets, but
more recently created symlinks to them) with a new (unexported)
method, fixAssetPaths, that replaces the relative paths in each
template html file with their real, non-symlinked, paths.  This avoids
a bug (feature?) in Firefox that was preventing the previous behavior
from working.  See https://bugzilla.mozilla.org/803999.

---

It turned out that my previous fix for Visualize not working in the Debian/Ubuntu package because of trying to copy symlinked Javascript/CSS/etc files (#2084) broke Visualize completely for Firefox users because of the bug mentioned above!

So now we just replace the links themselves in the generated html files so they point to the right place, no copying or symlinking required, e.g.,

```html
  <script src="/usr/share/javascript/jquery/jquery.min.js"></script>
  <script src="/usr/share/Macaulay2/Visualize/js/BootSideMenu.js"></script>
  <script src="/usr/share/javascript/d3/d3.min.js"></script>
  <script src="/usr/share/Macaulay2/Visualize/js/visSimplicialComplex2d.js"></script>
  <script src="/usr/share/javascript/bootstrap/js/bootstrap.min.js"></script>
  <script src="/usr/share/nodejs/clipboard/dist/clipboard.js"></script>
  <script src="/usr/share/Macaulay2/Visualize/js/nouislider.min.js"></script>
```

### DIFF
--- a/M2/Macaulay2/packages/Visualize.m2
+++ b/M2/Macaulay2/packages/Visualize.m2
@@ -546,8 +546,13 @@ copyJS(String) := opts -> dst -> (
 	    );
 	);
 
-    for dir in existingDirs do removeFile concatenate(dst, dir);
-    for dir in dirs do symlinkFile(basePath | "Visualize/" | dir, dst | dir);
+    for dir in dirs do (
+	makeDirectory(dst | dir);
+	for file in readDirectory(basePath | "Visualize/" | dir) do (
+	    if not member(file, {".", ".."}) then (
+		copyFile(realpath(basePath | "Visualize/" | dir | "/" | file),
+		    dst | dir | "/" | file)))
+    );
 
     return "Created directories at "|dst;
 )


DanGrayson:
One of the features of Macaulay2 distributions is that users without root account access can install Macaulay2 anywhere and expect it to work.  This will break that feature.  Can you use relative links instead?

d-torrance:
> One of the features of Macaulay2 distributions is that users without root account access can install Macaulay2 anywhere and expect it to work. This will break that feature. Can you use relative links instead?

I'm not sure why this would require root access.  All it's doing is replacing some strings in the generated html files.  If a user installed Macaulay2 in their home directory (and suppose the architecture-independent files are installed under `~/.local/share/Macaulay2`), then the links should look something like:

```html
 <script src="/home/m2user/.local/share/Macaulay2/Visualize/js/BootSideMenu.js"></script>
```

DanGrayson:
I was imagining that some of these strings are generated at build-time and find their way into distributions.  But even if that's not the case, one desirable feature of html files that users generate is that, after generating them, they can post them on a website for others to enjoy.  Won't this break that?

d-torrance:
> I was imagining that some of these strings are generated at build-time and find their way into distributions. 

No, the strings are generated at run-time (see the definition of the `assetPaths` hash table).

> But even if that's not the case, one desirable feature of html files that users generate is that, after generating them, they can post them on a website for others to enjoy. Won't this break that?

Hmm, this is a good point -- the original behavior copied everything into a temporary subdirectory that could be shared anywhere.  I think my previous PR (#2084) broke this already by using symlinks instead.

I think maybe going back to the original behavior of copying everything to the temporary directory, but being careful not to skip any symlinks, should work.

d-torrance:
I ended up restoring the original behavior, but using `copyFile` together with `realpath` instead of `copyDirectory`.  This way, all of the assets will be copied over, even the ones that are symlinks to their Debian/Ubuntu-packaged versions.

DanGrayson:
@bstone - is this change okay with you ?

bstone:
I am ok with this change. Thanks @d-torrance

PR #2129, opened 2021-06-03
MPI (message passing interface) --- run several M2 processes in parallel 

antonleykin:
We want to implement coarse parallelization via basic MPI routines --- collaborators are welcome. 
See the commentary in `packages/MPI.m2` 

To implement:
- [x] Basic message-passing functions in the kernel
- [ ] Additional message-passing functions in the kernel 
- [ ] A package facilitating the use of the functions above


M2/BUILD/MPI/Makefile
mahrud, line 1: If you're using MPICH, you can set `MPICH_CC=` and `MPICH_CXX=` environment variables to tell MPICH which base compiler to use.

mahrud:
- [Here](https://github.com/mahrud/M2/actions/runs/4845513942/jobs/8634476607#step:9:949)'s the result of running the three MPI examples on Github Actions, built using CMake on macOS, and linked with MPICH installed via Homebrew.

- example for `normalToricRing(Ideal,Thing)` fails, I'm not sure why.
- M2-binary compiles on Ubuntu, but crashed when building tvalues.m2
  - might have something to do with Ubuntu's MPICH vs Homebrew's

M2/BUILD/MPI/Makefile
antonleykin: Thanks! That's what I did, and it works.
With the recent update all builds fine on my version of Ubuntu.

antonleykin:
@DanGrayson and @mahrud, I pushed a commit that has a change that makes it possible to run `M2` both with `mpirun` and without. It builds for me with `autotools` and `cmake` on my Ubuntu --- see
M2/BUILD/MPI/Makefile
but I'm not sure how alter the "actions" to check everything

antonleykin:
> I think you should wrap things like `#include <mpi.h>` with a macro that checks if MPI is being used.

We probably should just require MPI at build time.
The basic functionality works and (unless it breaks anything anywhere in the non-MPI part) we should distribute the MPI-capable binaries in the next release.  

The only current "fail" I see looks unrelated to MPI:
https://github.com/Macaulay2/M2/actions/runs/4853189961/jobs/8649075617?pr=2129#step:9:3871

mahrud:
Even if you want the standard distribution to be MPI-capable (which is going to be difficult with brew and potentially also debian, since typically MPI-capable binaries are distributed in separate packages), there should be a way to disable MPI, similar to TBB, OpenMP, Python, etc.

antonleykin:
> Even if you want the standard distribution to be MPI-capable (which is going to be difficult with brew and potentially also debian, since typically MPI-capable binaries are distributed in separate packages), there should be a way to disable MPI, similar to TBB, OpenMP, Python, etc.

Fair enough. What would be the best way to omit MPI code with macros?

mahrud:
Inside `main()` there are several blocks wrapped in `#if WITH_PYTHON ... #endif` or `#if PROFILING ... #endif`, etc. I think a similar block like `WITH_MPI` in a few places would work.

Also, once MPI is a package and has its own tests, could you revert https://github.com/Macaulay2/M2/pull/2129/commits/f1aa4e4f6187f4f295417eb5815b3d97eb67f0f1?

antonleykin:
@mahrud, I've encapsulated all MPI-related code with macros, so ubuntu _non-MPI_ builds go through... but I don't understand why the _MPI_ build fails now.

@DanGrayson, could you glance at `M2/configure.ac`? The autotools _non-MPI_ ubuntu build works... but I'm not sure how to complete this so that _MPI_ works as well.

antonleykin:
> @mahrud, I've encapsulated all MPI-related code with macros, so ubuntu _non-MPI_ builds go through... but I don't understand why the _MPI_ build fails now.

I'll need some help with `cmake` here... since `cmake` fails both on ubuntu and macOS in a similar fashion: the object file coming from `d/M2lib.c`  is missing symbols "_MPI..."

Could it be that `WITH_MPI` is somehow undefined when `d/M2lib.c` is compiled?
 
> @DanGrayson, could you glance at `M2/configure.ac`? The autotools _non-MPI_ ubuntu build works... but I'm not sure how to complete this so that _MPI_ works as well.

All seems fine now with the MPI build via `autotools` --- the problem above doesn't emerge.

mahrud:
> Could it be that `WITH_MPI` is somehow undefined when `d/M2lib.c` is compiled?

Yes, sorry I should have added this when I added the `WITH_MPI` option.

I also cherry-picked a fix for Normaliz 3.10, which is on development (it might be helpful to rebase on top of development and force push).

antonleykin:
Everything is fine with `openmpi`, but there is an obstacle with `mpich`:
there is no analog ot variable `OMPI_COMM_WORLD_RANK` (used, e.g., in M2/Macaulay2/bin/main.cpp).
Is there any way to know if `mpirun` was used to execute the program?

I've replaced `openmpi` with `mpich` on my ubuntu and now M2 crashes (unless executed with `mpirun`).

mahrud:
Looking online, one suggestion seems to be to use the function `MPI_Comm_rank` instead, is that the same?

antonleykin:
> Looking online, one suggestion seems to be to use the function `MPI_Comm_rank` instead, is that the same?

Well, probing the environment variable was a way to determine if the execution started via `mpirun`. It looks like MPI builds successfully now on github... so perhaps I overlooked something on my local machine.

antonleykin:
> Even if you want the standard distribution to be MPI-capable (which is going to be difficult with brew and potentially also debian, since typically MPI-capable binaries are distributed in separate packages), there should be a way to disable MPI, similar to TBB, OpenMP, Python, etc.

Let me ask the opinion of @d-torrance (at least on debian). Do you think there need to distribute the MPI version of M2 separately? 
(So far, on ubuntu and macOS, there seems to be no need to have two binaries: the one we build can be executed with or without `mpirun`.)

mahrud:
Now that the tests pass, could you turn the MPI tests into either tests for the package or perhaps tests under `Macaulay2/tests/MPI`? (I think the former is better, unless you don't plan there to be an MPI package)

antonleykin:
I had success building this on Fedora 34.
It still segfaults on Ubuntu.

antonleykin:
The recent attempt to have "master" interrupt "workers" doesn't succeed yet:
M2/Macaulay2/packages/MPI/test-interruption.m2

The interruption should occur when `MPI_Probe` detects the necessary message from master:
M2/Macaulay2/e/interrupted.cpp
 
BTW, there may be a bug in the build process: `WITH_MPI` doesn't seem to be defined when `interrupted.cpp` is compiled (that's why `#ifdef` is commented out).

PR #2130, opened 2021-06-03
Python interface

d-torrance:
This is the current draft of the Python interface package (and a simple [NetworkX](https://networkx.org/) package as a proof of concept) that I talked about at this afternoon's M2internals meeting.  Note that in addition to the Python stuff, I've done a few other things:

* Make `toRR` a method function so that `(toRR, PythonObject)` will convert python objects to Macaulay2 reals.  I may do the same thing with other "convert to standard M2 class" functions like `toList`, `toSequence`, and `hashTable` that are currently all `CompiledFunction`'s.
* Add a `toZZ` method function that does the expected thing (using `lift`) in `Core` and is overloaded in `Python`
* Add `@@@` and `@@?` keywords.

Things to do:

- [x] documentation
- [x] tests
- [ ] support arbitrary precision integers? (currently all Python `int` objects are converted to `long` before they are converted to `ZZ`)
- [ ] garbage collection?  I don't know how Python's reference counting would work with GC, so I haven't dealt with it.
- [ ] get Sage working? 

Comments/contributions welcome!

Here's the example code from today:

```m2
Macaulay2, version 1.18.0.1
with packages: ConwayPolynomials, Elimination, IntegralClosure, InverseSystems, LLLBases, MinimalPrimes, PrimaryDecomposition,
               ReesAlgebra, Saturation, TangentCone

i1 : loadPackage "Python"
 -- success: python is present

o1 = Python

o1 : Package

i2 : rs "print('Hello, world!')"
Hello, world!

o2 = None

o2 : PythonObject of class NoneType

i3 : toPython {1, 2/3, "foo", (1,2,3), hashTable{"foo" => "bar"}}

o3 = [1, 0.6666666666666666, 'foo', (1, 2, 3), {'foo': 'bar'}]

o3 : PythonObject of class list

i4 : methods toPython

o4 = {0 => (toPython, Boolean)     }
     {1 => (toPython, CC)          }
     {2 => (toPython, Constant)    }
     {3 => (toPython, HashTable)   }
     {4 => (toPython, List)        }
     {5 => (toPython, Nothing)     }
     {6 => (toPython, PythonObject)}
     {7 => (toPython, QQ)          }
     {8 => (toPython, RR)          }
     {9 => (toPython, Sequence)    }
     {10 => (toPython, Set)        }
     {11 => (toPython, String)     }
     {12 => (toPython, ZZ)         }

o4 : NumberedVerticalList

[...]

Issue #1800, opened 2021-01-07
xor is not a keyword and does not take boolean


mahrud:
```m2
i1 : if true and true then true

o1 = true

i2 : if true xor true then true
stdio:2:9:(3): error: expected 2 arguments

i3 : if xor(true, true) then true
stdio:3:4:(3): error: expected argument 1 to be an integer
```

PR #2131, opened 2021-06-04
Updates to XOR

d-torrance:
While trying to convert Python binary operators into Macaulay2 binary operators and vice versa for #2130, I noticed that our XOR doesn't line up nicely with the others.

## Logical XOR

This addresses #1800.  In particular `xor` is now a keyword and `(symbol xor, Boolean, Boolean)` is a method:

```m2
i1 : 1 + 1 == 2 xor 2 + 2 == 4

o1 = false

i2 : 1 + 1 == 2 xor 2 + 2 == 5

o2 = true
```

## Bitwise XOR

The old `xor` method for bitwise XOR (which was only used twice) has been updated to use `^^` instead (since `^` is obviously taken!):

```m2
i3 : (1 + 2 + 8) ^^ (1 + 4 + 8)

o3 = 6
```
This is more consistent with the current use of  the keywords `&` and `|` for bitwise AND and OR, and it appears that this may have been considered at some point already considering `^^`'s precedence order, right where bitwise XOR would go:

https://github.com/Macaulay2/M2/blob/b5b9b18ccb0dbf4daeeefc69290f6f18331134ef/M2/Macaulay2/d/binding.d#L272-L277


d-torrance:
I should probably add some tests... I'll do that later.

d-torrance:
> I should probably add some tests... I'll do that later.

Done

mahrud:
> > I should probably add some tests... I'll do that later.
> 
> Done

Might that be too many basic tests? Also see 
https://github.com/Macaulay2/M2/blob/b5b9b18ccb0dbf4daeeefc69290f6f18331134ef/M2/Macaulay2/m2/basictests/A01.m2#L67-L72
https://github.com/Macaulay2/M2/blob/b5b9b18ccb0dbf4daeeefc69290f6f18331134ef/M2/Macaulay2/m2/basictests/A01.m2#L143-L147

d-torrance:
> Might that be too many basic tests?

I figured I might as well cover all the bases -- it doesn't take very long:

```m2
i14 : (tests "Core")#21

o14 = 
      usr-dist/common/share/doc/Macaulay2/Core/tests/boolean.m2:0:1-64:1:

o14 : TestInput

i15 : check_21 "Core"
 -- capturing check(21, "Core")                                              -- 0.050261 seconds elapsed
```

I did notice the existing tests, but I don't think the new XOR methods would be available yet when they're run, and I figured I might as well throw in AND and OR for good measure.

mahrud:
I'm not concerned about the time. It's redundant in that if most of the things you test are broken, I highly doubt the computation will even reach that step. That's why `runBasicTests` is defined in the interpreter, not in top level.

d-torrance:
> I'm not concerned about the time. It's redundant in that if most of the things you test are broken, I highly doubt the computation will even reach that step. That's why `runBasicTests` is defined in the interpreter, not in top level.

Fair enough -- I'll remove the AND/OR-only tests.

d-torrance:
Oops:

```m2
 -- running   check(335, "Core")                                             -- 0.313255 seconds elapsed
/home/runner/work/M2/M2/M2/BUILD/build/usr-dist/common/share/doc/Macaulay2/Core/tests/names.m2:0:1-1119:1: error:
 -- i1105 : assert( (toString weightRange) === "weightRange" )
 -- 
 -- i1106 : assert( (toString width) === "width" )
 -- 
 -- i1107 : assert( (toString wrap) === "wrap" )
 -- 
 -- i1108 : assert( (toString x) === "x" )
 -- 
 -- i1109 : assert( (toString xor) === "xor" )
 -- stdio:1110:22:(3): error: syntax error at ')'
 -- /home/runner/work/M2/M2/M2/Macaulay2/m2/startup.m2.in:671:16:(0): --backtrace: parse error--
 -- 
stdio:1:1:(3): error: test(s) #184 of package Core failed.
```

I'll fix this shortly.

PR #2138, opened 2021-06-10
Add support for entering integers in binary, octal, and hexadecimal

d-torrance:
For example:

```m2
i1 : 0b10101

o1 = 21

i2 : 0o755

o2 = 493

i3 : 0xdeadbeef

o3 = 3735928559
```


jchen419:
Hi Doug, do you know if base conversion between base 10 and an arbitrary (integer) base already exists in M2?

d-torrance:
> Hi Doug, do you know if base conversion between base 10 and an arbitrary (integer) base already exists in M2?

I don't believe so, no.

jchen419:
Hmm, OK, thanks - maybe a future feature request!

DanGrayson:
Or maybe a current feature request...

PR #2141, opened 2021-06-11
Add boolean return value to captureExampleOutput

d-torrance:
This accomplishes two things:
* We can use the return value to decide whether to run `possiblyCache` from inside `generateExampleResults`, which means we no longer need to pass it to `captureExampleOutput`.
* In verbose mode, the caching log no longer occurs before `elapsedTime` prints the timing results.

### Before
```m2
 -- making example result files in ../.Macaulay2/local/share/doc/Macaulay2/Foo/example-output/
 -- making example results for "foo"                                         -- caching example results for foo in /home/profzoom/tmp/Foo/examples/_foo.out
 -- copying: /home/profzoom/.Macaulay2/local/share/doc/Macaulay2/Foo/example-output/_foo.out -> /home/profzoom/tmp/Foo/examples/_foo.out
 -- 0.829066 seconds elapsed
 -- storing example results in ../.Macaulay2/local/share/doc/Macaulay2/Foo/example-output/_foo.out
```

### After
```m2
 -- making example result files in ../.Macaulay2/local/share/doc/Macaulay2/Foo/example-output/
 -- making example results for "foo"                                         -- 1.15328 seconds elapsed
 -- caching example results for foo in /home/profzoom/tmp/Foo/examples/_foo.out
 -- copying: /home/profzoom/.Macaulay2/local/share/doc/Macaulay2/Foo/example-output/_foo.out -> /home/profzoom/tmp/Foo/examples/_foo.out
 -- storing example results in ../.Macaulay2/local/share/doc/Macaulay2/Foo/example-output/_foo.out
```

### DIFF
--- a/M2/Macaulay2/m2/examples.m2
+++ b/M2/Macaulay2/m2/examples.m2
@@ -190,7 +190,7 @@ storeExampleOutput = (pkg, fkey, outf, verboseLog) -> (
 
 -- used in installPackage.m2
 -- TODO: reduce the inputs to this function
-captureExampleOutput = (desc, inputs, pkg, cacheFunc, inf, outf, errf, data, inputhash, changeFunc, usermode) -> (
+captureExampleOutput = (desc, inputs, pkg, inf, outf, errf, data, inputhash, changeFunc, usermode) -> (
     stdio << flush; -- just in case previous timing information hasn't been flushed yet
     -- try capturing in the same process
     if isCapturable(inputs, pkg, false) then (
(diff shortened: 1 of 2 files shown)


d-torrance:
Hmm, something's not working:

```m2
 -- capturing example results for "firstFunction(ZZ)"                        -- 0.633454 seconds elapsed
currentString:1:1:(3):[2]: error: expected true or false
```

d-torrance:
Ah, I didn't take into account the case where the example results are captured, which returns from a different point.

There was actually another bug here -- we weren't caching example results that were captured.  I'll push a fix shortly.

PR #2142, opened 2021-06-12
Grammar file updates

d-torrance:
This PR includes:

* A couple of changes to the Pygments lexer that I added for https://github.com/pygments/pygments/pull/1791 (copyright header and one more line of Python code)
* A README for the Pygments lexer.
* Removed some generated grammar files from git and added them to .gitignore so they won't muddy up `git status` after running `make-M2-symbols.m2`.


mahrud:
> * Removed some generated grammar files from git and added them to .gitignore so they won't muddy up `git status` after running `make-M2-symbols.m2`.

You shouldn't delete the files. They should be updated so users can download them.

d-torrance:
> > * Removed some generated grammar files from git and added them to .gitignore so they won't muddy up `git status` after running `make-M2-symbols.m2`.
> 
> You shouldn't delete the files. They should be updated so users can download them.

Ideally, generated files shouldn't be under version control.  Maybe there's a better way to get them to vim users?  Perhaps we could install them somewhere, or try to get them included in vim itself?

d-torrance:
Oops, noticed a typo in the README -- just pushed a new version of that commit.

mahrud:
> Ideally, generated files shouldn't be under version control. Maybe there's a better way to get them to vim users? Perhaps we could install them somewhere, or try to get them included in vim itself?

The point is that you shouldn't need to build, install, or clone M2 to get the syntax highlighting files for your editor of choice. It's the same with emacs files. These two didn't used to be generated at all, but two generated files (among many, many cached examples) won't hurt anyone.

d-torrance:
> > Ideally, generated files shouldn't be under version control. Maybe there's a better way to get them to vim users? Perhaps we could install them somewhere, or try to get them included in vim itself?
> 
> The point is that you shouldn't need to build, install, or clone M2 to get the syntax highlighting files for your editor of choice. It's the same with emacs files. These two didn't used to be generated at all, but two generated files (among many, many cached examples) won't hurt anyone.

Fair enough.  I've restored the vim files (and updated them for 1.18).

Issue #2158, opened 2021-06-25
cohomcalg under ubuntu


DanGrayson:
There's a problem with the cohomcalg package under Ubuntu 21.04:

```sh
ubuntu2104$ /usr/bin/cohomcalg 

    cohomCalg v0.32
    (compiled for Linux/Unix x86-64 / 64 bit)
    author: Benjamin Jurke (mail@benjaminjurke.com)
    Based on the algorithm detailed in arXiv:1003.5217

...
```

Here is the input file

[cohomcalg-input.txt](https://github.com/Macaulay2/M2/files/6718894/cohomcalg-input.txt)

Run the program with the command

```sh
/usr/bin/cohomcalg --integrated cohomcalg-input.txt
```

Then the output will be prefixed with 13770 identical lines of the form "M_full= ".  When the CohomCalg package calls the "value" function on the returned text, the parser crashes, presumably because of a stack overflow.

When we build cohomcalg 0.32 ourselves, this does not happen.

d-torrance:
I reproduced this, both in Ubuntu 21.04 and Debian unstable.  (At least the 13770 line part -- how do I run `value` on the returned text?)

One big difference with the Debian/Ubuntu package is that it's built against the Debian polylib package instead of using an embedded copy.  I wonder if that's related?

The Debian/Ubuntu package also gives much more verbose output in general.  For example, comparing the `CohomCalg` docs from the [website](https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.18/share/doc/Macaulay2/CohomCalg/html/index.html) and the Debian package:

### website
```m2
i9 : cohoms1 = for i from 0 to 6 list X_i => cohomCalg X_i

    cohomCalg v0.32
    (compiled for Linux/Unix x86-64 / 64 bit)
    author: Benjamin Jurke (mail@benjaminjurke.com)
    Based on the algorithm detailed in arXiv:1003.5217


Usage and generation of intermediate monomial files deactivated.

Starting computation of secondary sequences...
  0.00% completed (1 sec remaining)...            
                                                                           
Computation of secondary cohomologies and contributions complete.
Computing target cohomology 1 of 1 (0.0% done)...       
Computation of the target cohomology group dimensions complete.

    All done. Program run successfully completed.


    cohomCalg v0.32
    (compiled for Linux/Unix x86-64 / 64 bit)
    author: Benjamin Jurke (mail@benjaminjurke.com)
[...]

Issue #980, opened 2019-08-06
package CohomCalg is too noisy


DanGrayson:
Look at this noisy example output:
```
-- -*- M2-comint -*- hash: -1578154438

i1 : needsPackage "NormalToricVarieties"

o1 = NormalToricVarieties

o1 : Package

i2 : X = smoothFanoToricVariety(3,15)

o2 = X

o2 : NormalToricVariety

i3 : rays X

o3 = {{1, 0, 0}, {0, 1, 0}, {0, 0, 1}, {0, -1, -1}, {0, -1, 0}, {-1, 0, 0},
     ------------------------------------------------------------------------
     {-1, 1, 0}}

o3 : List

i4 : max X

o4 = {{0, 1, 2}, {0, 1, 3}, {0, 2, 4}, {0, 3, 4}, {1, 2, 6}, {1, 3, 6}, {2,
     ------------------------------------------------------------------------
     4, 5}, {2, 5, 6}, {3, 4, 5}, {3, 5, 6}}

o4 : List

i5 : S = ring X

o5 = S

o5 : PolynomialRing

i6 : SR = dual monomialIdeal X

o6 = monomialIdeal (x x , x x , x x , x x , x x , x x )
                     2 3   1 4   0 5   1 5   0 6   4 6

o6 : MonomialIdeal of S

i7 : KX = toricDivisor X

o7 = - X  - X  - X  - X  - X  - X  - X
        0    1    2    3    4    5    6

o7 : ToricDivisor on X

i8 : assert isVeryAmple (-KX)

i9 : cohoms1 = for i from 0 to 6 list X_i => cohomCalg X_i

    cohomCalg v0.31b
    (compiled on Aug  6 2019 @ 17:34:12 for Linux/Unix x86-64 / 64 bit)
    author: Benjamin Jurke (mail@benjaminjurke.net)
    Based on the algorithm presented in arXiv:1003.5217


Usage and generation of intermediate monomial files deactivated.

Starting computation of secondary sequences...
  0.00% completed (1 sec remaining)...            
                                                                           
Computation of secondary cohomologies and contributions complete.
Computing target cohomology 1 of 1 (0.0% done)...       
Computation of the target cohomology group dimensions complete.

    All done. Programm run successfully completed.


    cohomCalg v0.31b
    (compiled on Aug  6 2019 @ 17:34:12 for Linux/Unix x86-64 / 64 bit)
    author: Benjamin Jurke (mail@benjaminjurke.net)
[...]

PR #2165, opened 2021-06-28
runProgram/CohomCalg updates

d-torrance:
As reported in #2158, the Debian/Ubuntu cohomcalg package prints lots of extra polylib cruft we don't need, so we pipe to `tail` to extract just the final line of output.  This requires a small update to `runProgram` to support pipes.

We also default to silencing cohomcalg's output, closing #980.

### DIFF
--- a/M2/Macaulay2/m2/programs.m2
+++ b/M2/Macaulay2/m2/programs.m2
@@ -143,6 +143,7 @@ runProgram(Program, String, String) := opts -> (program, name, args) -> (
 	    makeDirectory opts.RunDirectory;
 	"cd " | opts.RunDirectory | " && " ) else "";
     cmd = cmd | program#"path" | addPrefix(name, program#"prefix") | " " | args;
+    if match("\\|", cmd) then cmd = "{ " | cmd | ";}";
     returnValue := run (cmd | " > " | outFile | " 2> " | errFile);
     message := "running: " | cmd | "\n";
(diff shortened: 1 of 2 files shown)


DanGrayson:
How is that polylib cruft not just an error in cohomcalg?

d-torrance:
It's apparently valid Macaulay2 code:

```m2
i1 : M_full=
     M_full=
     M_full=
     print "Hello, world!"
Hello, world!
```

DanGrayson:
Yes, it's valid M2 code, but just by accident.

Wouldn't the correct fix be to fix cohomcalg?

d-torrance:
> Wouldn't the correct fix be to fix cohomcalg?

Upstream cohomcalg patches out the cruft in an embedded copy of polylib, but in Debian, we can't use embedded copies of libraries, so we link against the Debian polylib package.

Even if we were to convince upstream polylib to remove the cruft (or at least print the matrix name to stderr along with the matrix itself) or patch the Debian package to do so, that would only fix this in cohomcalg for future releases.  All of the previously released Debian/Ubuntu packages (some of which will be around for awhile, e.g., Ubuntu 20.04's EOL is [April 2030](https://wiki.ubuntu.com/Releases)) will still have this problem

Grabbing just the last line of output in Macaulay2 itself is an easy fix that will work regardless of which cohomcalg package the user has installed.

PR #2166, opened 2021-06-28
Add source directories to path in the right order

mahrud:
Also changes the order in which directories are added, from
```
 X/m2 X/packages Y/m2 Y/packages
```
to
```
 X/m2 Y/m2 X/packages Y/packages.
```
This has occasionally caused issues on Mac systems where filenames aren't case sensitive. Note that after Core is loaded, the m2 directories are removed from the path.


mahrud:
`--srcdir` isn't tested anywhere, so I cancelled the tests.

PR #2160, opened 2021-06-26
Refactor mathic

mikestillman:
this pull request removes memtailor, mathic, mathicgb as libraries (or submodules), and places the code into subdirectories of the e directory.  The c++ code has been changed to give no warnings (or fewer at least, depending on the compiler).  Most unit tests are available, but in the autotools build, better support for the tests will be added in a later pull request.

PR #2173, opened 2021-07-01
Compile mathicgb submodule as a subdirectory

mahrud:
This is my alternative to #2160 for compiling mathicgb and friends as a part of engine while also keeping them submodules. This way those libraries can be developed and tested along with the engine, without rendering the libraries unusable. The unit tests for all three are also joined and run at the end with `ctest`.

This goes with:
- https://github.com/Macaulay2/mathicgb/pull/28
- https://github.com/Macaulay2/mathic/pull/8
- https://github.com/Macaulay2/memtailor/pull/7

I should add that Mike's changes to the three libraries are added to the libraries in the three PR's linked above.

If you use the autotools build, this PR essentially changes nothing, except moves the location of a few source files and ensures that the submodules are downloaded.

Keeping the engine modular makes it easier to understand, at least for me, but if there's an engine improvement planned that is impossible with these changes then this can be rejected and I'll try to maintain this fork on my own.


mahrud:
I should add that Mike's changes to the three libraries are added to the libraries in the three PR's linked above.

DanGrayson:
There doesn't seem to be any advantage to having those three subdirectories be submodules, because having them as subdirectories already achieves the modularity you say you want.  In addition, having them as submodules makes development of the engine harder, as more work is required to coordinate changes when 4 separate repositories are involved.

d-torrance:
What's the plan for mathicgb as a standalone library going forward?  There's the `mgb` command line utility and also the option to compile Singular with it (and maybe other projects?).   If these are still to be supported, then I think going back to using submodules makes a lot of sense.

mahrud:
> There doesn't seem to be any advantage to having those three subdirectories be submodules

The advantage of this way is that those libraries can be developed and tested along with the engine, without leaving them unmaintained. The point is to prevent diverging of the sources.

> having them as subdirectories already achieves the modularity you say you want

It doesn't. An example of modularity is being able to use `mgb` to compute a single gb without running all of M2 so that I can compare the timing with something like msolve. The new subdirectories don't have the necessary sources to build mgb, and the directory structure has diverged so much that it's not easy to transfer the changes forth and back.

> In addition, having them as submodules makes development of the engine harder, as more work is required to coordinate changes when 4 separate repositories are involved.

There are many solutions online that don't involve diverging sources. The simplest solution, which is still better than the new setup, is how factory is an independent subdirectory of singular. Is that too hard?

DanGrayson:
> > having them as subdirectories already achieves the modularity you say you want
> 
> It doesn't. An example of modularity is being able to use `mgb` to compute a single gb without running all of M2 to allow comparison with something like msolve. The new subdirectories don't have the necessary sources to build mgb, and the directory structure has diverged so much that it's not easy to transfer the changes forth and back.
> 

Probably no one uses `mgb`.  Even if someone does, it would be possible for Mike to add the relevant source files to the Macaulay2 repository and build `mgb` using a Makefile there.  It could even be distributed along with `M2`.

d-torrance:
> Probably no one uses `mgb`.

FWIW, the `mathicgb` Debian package (which contains only `mgb` and its manpage) has been installed [at least 90 times](https://qa.debian.org/popcon.php?package=mathicgb).

---

Since embedded libraries are discouraged in Debian packages, my current draft of the Macaulay2 package strips out the memtailor/mathic/mathicgb subdirectories and links against the existing shared libraries (see [this patch](https://salsa.debian.org/science-team/macaulay2/-/blob/debian/development/debian/patches/use-shared-mathicgb-and-friends.patch)).  This currently works just fine (better, actually, on s390x -- fewer #2162 issues), but I suppose this may change if mathicgb and friends diverge too much from the original repositories.

mikestillman:
I was never completely happy with the way mathicgb became a library, where it exposes almost none of the code I want to use.  Also, it is a complete work in progress, that got stalled after Bjarke left for google, and Cristian Eder and I were reluctant to change what he had done, e.g. it used the very new (at the time) c++11 move semantics, etc (I might be putting words into Christian's mouth here though!) and we thought we might mess up the code.  But now I don't feel that way any longer...

In any case, I think I'll put the modules back in.  But I want to make sure I have access to the entire set of header files for these libraries, not just the one exposed by mathicgb.  That doesn't seem to be compatible with using what is there?  How should I do that?  @d-torrance ?

d-torrance:
> But I want to make sure I have access to the entire set of header files for these libraries, not just the one exposed by mathicgb. That doesn't seem to be compatible with using what is there? How should I do that? @d-torrance ?

Currently, the Debian packages contain lots of the headers already:

```m2
profzoom@peg-amy:~$ apt-file search -x "/usr/include/m(emtailor|athic)"
libmathic-dev: /usr/include/mathic.h      
libmathic-dev: /usr/include/mathic/Action.h
libmathic-dev: /usr/include/mathic/BinaryKDTree.h
libmathic-dev: /usr/include/mathic/BitTriangle.h
libmathic-dev: /usr/include/mathic/BoolParameter.h
libmathic-dev: /usr/include/mathic/CliParameter.h
libmathic-dev: /usr/include/mathic/CliParser.h
libmathic-dev: /usr/include/mathic/ColumnPrinter.h
libmathic-dev: /usr/include/mathic/ComTree.h
libmathic-dev: /usr/include/mathic/Comparer.h
libmathic-dev: /usr/include/mathic/DivFinder.h
libmathic-dev: /usr/include/mathic/DivList.h
libmathic-dev: /usr/include/mathic/DivMask.h
libmathic-dev: /usr/include/mathic/GeoFront.h
libmathic-dev: /usr/include/mathic/Geobucket.h
libmathic-dev: /usr/include/mathic/HashTable.h
libmathic-dev: /usr/include/mathic/Heap.h
libmathic-dev: /usr/include/mathic/HelpAction.h
libmathic-dev: /usr/include/mathic/IntegerParameter.h
libmathic-dev: /usr/include/mathic/KDEntryArray.h
libmathic-dev: /usr/include/mathic/KDTree.h
libmathic-dev: /usr/include/mathic/NameFactory.h
libmathic-dev: /usr/include/mathic/PackedKDTree.h
libmathic-dev: /usr/include/mathic/PairQueue.h
libmathic-dev: /usr/include/mathic/StlSet.h
libmathic-dev: /usr/include/mathic/StringParameter.h
libmathic-dev: /usr/include/mathic/Timer.h
libmathic-dev: /usr/include/mathic/TourTree.h
libmathic-dev: /usr/include/mathic/display.h
libmathic-dev: /usr/include/mathic/error.h
libmathic-dev: /usr/include/mathic/stdinc.h
libmathicgb-dev: /usr/include/mathicgb.h
libmemtailor-dev: /usr/include/memtailor.h
libmemtailor-dev: /usr/include/memtailor/Arena.h
libmemtailor-dev: /usr/include/memtailor/ArenaVector.h
libmemtailor-dev: /usr/include/memtailor/BufferPool.h
libmemtailor-dev: /usr/include/memtailor/MemoryBlocks.h
libmemtailor-dev: /usr/include/memtailor/stdinc.h
```
It looks like maybe some of the mathicgb ones might be missing, but that should be an easy fix by adding them here:
https://github.com/Macaulay2/mathicgb/blob/4d23c24302c2395efbbeef9ea1fd52e4f0f56f65/Makefile.am#L64-L66

mahrud:
> Actually maybe @mahrud 's changes in this pull request do what I want. I have been slow in the last few weeks, but i'm planning on trying it out.

With changes in this PR, all sources in mathicgb, etc. are available within the `e/` directory.

I don't care if mathicgb, etc. are necessarily submodules. If you want to simplify things you can merge mathic and memtailor into mathicgb and have that be a subdirectory of `e/` with all the necessary files to create a single, standalone library plus `mgb`. You can then archive the memtailor/mathic/mathicgb repositories and tell people to use this subdirectory for building `mgb`, for instance. Compare this with how `factory` is a subdirectory of `singular` with all necessary files to make a standalone library.

Doug's concern about is slightly different than mine, but since the singular/factory setup works for them, something like this should work here as well.

mikestillman:
@d-torrance I'm planning on accepting this pull request.  But how does that affect M2 under ubuntu and debian?  In particular, if mathicgb is a debian package, and we include it as a submodule, to compile in, is this a problem?  The problem currently (or, before I removed mathicgb as a submodule) is that there are a number of c++ include files that I want to use, which are not in the current versions (I believe) of the mathicgb package.  Maybe we just need to use the suggestion you made above.

> It looks like maybe some of the mathicgb ones might be missing, but that should be an easy fix by adding them here: https://github.com/Macaulay2/mathicgb/blob/4d23c24302c2395efbbeef9ea1fd52e4f0f56f65/Makefile.am#L64-L66

I expect the answer is: we must include it as a library if it is available with a specific version.  We probably need to put this check back into M2 (@mahrud or is it already there?) before accepting this?

d-torrance:
I think adding a check back in could wait for another PR.  There's currently nothing causing too many problems for the Debian package -- I've just patched things to use the existing library packages instead of compiling the new e subdirectories.  The patch would probably require a few tweaks after this PR is merged, but I don't expect it should be too complicated.

mahrud:
> I expect the answer is: we must include it as a library if it is available with a specific version. We probably need to put this check back into M2 (@mahrud or is it already there?) before accepting this?

I think it's easier to not do any version checks and always build the libraries when compiling from source. For ubuntu/debian, Doug can make sure the right version is available using Debian's package requirements and use his patch to make M2 link with those instead.

> I think adding a check back in could wait for another PR. There's currently nothing causing too many problems for the Debian package -- I've just patched things to use the existing library packages instead of compiling the new e subdirectories. The patch would probably require a few tweaks after this PR is merged, but I don't expect it should be too complicated.

Since the debian packages use the autotools build, I suspect after this PR your patches would need to change only superficially (maybe some location changes, etc.)

FYI, with this PR the CMake build compiles the submodules into standalone libraries and links with them, as opposed to the autotools build which just builds the object files and links them all together at the end.

mikestillman:
@DanGrayson I'm going to accept this pull request and the 3 corresponding ones for mathic, mathicgb, memtailor.  Seem ok?

DanGrayson:
> @DanGrayson I'm going to accept this pull request and the 3 corresponding ones for mathic, mathicgb, memtailor. Seem ok?

Yes, seems ok.

mikestillman:
@mahrud I'm about to merge this pull request.  Thanks! This is better than the solution I took when I ripped these out, I think.

d-torrance:
I opened a PR (https://github.com/Macaulay2/mathicgb/pull/29) to install all the mathicgb headers for the autotools build as discussed above.  (It's already happening in the CMake build.)

Issue #2179, opened 2021-07-06
Missing header in e/mathicgb/RawVector.hpp ?


haerski:
I was building from the latest commit (3cc9185) in `development` with gcc version 11.1.0 when I got an error that I hadn't seen before.

```
[3/59] Building CXX object Macaulay2/e/CMakeFiles/M2-engine.dir/mathicgb/F4MatrixReducer.cpp.o
FAILED: Macaulay2/e/CMakeFiles/M2-engine.dir/mathicgb/F4MatrixReducer.cpp.o 
/usr/bin/c++ -DNDEBUG -DOM_NDEBUG -DSING_NDEBUG -I/home/haerski/work/M2/M2/Macaulay2/e -I/home/haerski/work/M2/M2/Macaulay2/d -I/home/haerski/work/M2/M2/Macaulay2/c -IMacaulay2/d -I/usr/include/libxml2 -isystem usr-host/include -isystem /usr/include/eigen3 -O3 -DNDEBUG -march=native -Wuninitialized -g3 -I/home/haerski/work/M2-cmake/usr-host/include -I/home/haerski/work/M2/M2/include -I/home/haerski/work/M2-cmake/include -Wall -Wextra -Wfatal-errors -Wcast-qual -Wno-unused-parameter -Wno-attributes -Wno-cast-qual -Wno-sign-compare -Wno-unused-local-typedefs -Wno-deprecated-copy -Wno-unknown-pragmas -Wno-ignored-qualifiers -Wno-implicit-fallthrough -fopenmp -lgomp -lpthread -std=gnu++17 -MD -MT Macaulay2/e/CMakeFiles/M2-engine.dir/mathicgb/F4MatrixReducer.cpp.o -MF Macaulay2/e/CMakeFiles/M2-engine.dir/mathicgb/F4MatrixReducer.cpp.o.d -o Macaulay2/e/CMakeFiles/M2-engine.dir/mathicgb/F4MatrixReducer.cpp.o -c /home/haerski/work/M2/M2/Macaulay2/e/mathicgb/F4MatrixReducer.cpp
In file included from /home/haerski/work/M2/M2/Macaulay2/e/mathicgb/SparseMatrix.hpp:6,
                 from /home/haerski/work/M2/M2/Macaulay2/e/mathicgb/F4MatrixReducer.hpp:6,
                 from /home/haerski/work/M2/M2/Macaulay2/e/mathicgb/F4MatrixReducer.cpp:4:
[...]

PR #2180, opened 2021-07-06
Include header for std::numeric_limits

d-torrance:
Otherwise, mathicgb fails to compile using gcc 11.

See: #2179, Macaulay2/mathicgb#25, Macaulay2/mathicgb#26,
Macaulay2/mathicgb#27


haerski:
Fixes #2179

PR #2189, opened 2021-07-17
renamed FastLinAlg->FastMinors, renamed RandomRationalPoints->RandomP…

kschwede:
This renames:
FastLinAlg -> FastMinors
RandomRationalPoints -> RandomPoints
and fixes some incorrect references due to the naming.  

This also has numerous changes and improvements to RandomPoints.  There are large performance improvements (in many examples, several order of magnitude).  Additionally, there is new functionality where we estimate dimension by intersecting with random hyperplanes (over field extensions).


kschwede:
By the way, I'm not sure if there are other places where I should rename RandomRationalPoints.m2 -> RandomPoints.m2 and FastLinAlg.m2 -> FastMinors.m2

In particular, these exist in m2.vim.dict and m2.vim.syntax.

Also, should I change "changes.m2".  Either renaming the {TO "FastLinAlg::FastLinAlg",...} etc. entries or removing the TO commands?

Also, did not making these changes contribute to the build failures?

DanGrayson:
The one build failure is unrelated.

DanGrayson:
The files m2.vim.dict and m2.vim.syntax are automatically generated, so don't worry about them.

PR #2190, opened 2021-07-18
Pygments lexer updates

d-torrance:
Two small updates to the Pygments lexer based on comments on https://github.com/pygments/pygments/pull/1791.


d-torrance:
I think I was too hasty pushing this -- I noticed a mistake in one of the commits.  I'll push again after the Pygments team has finished reviewing my PR.

PR #2191, opened 2021-07-20
added BettiCharacters package

galettof:
Introducing the BettiCharacters package to compute finite group characters on free resolutions and graded modules.


DanGrayson:
Thank you for the submission.  Could you convert the canned examples to uncanned examples?  Running example code regularly is one good way to make sure your package continues to work.

galettof:
Thank you for the feedback. The example was slow to compile so I had replaced it with a canned one to speed things during development. I reverted back to an uncanned example.

DanGrayson:
If examples are slow, it should be easy to switch to simpler examples that don't take so long.

galettof:
This particular example is one of the smallest I am aware of in the literature with the features I want to showcase, and I don't know of any simpler ones. Still it only takes a little over a minute to compile on my aging machine, so it think it is fine to include it as is.

PR #2200, opened 2021-07-27
Fix cddlib includes

DanGrayson:


d-torrance:
This isn't detecting cddlib on slightly older Debian/Ubuntu systems any more.  From https://github.com/Macaulay2/M2/pull/2200/checks?check_run_id=3173328809 (running Ubuntu 20.04):

```
checking for cddlib/setoper.h... no
cddlib not found, will build
```

The headers were moved from `/usr/include/cdd` to `/usr/include/cddlib` with version 094l-1, uploaded to Debian unstable last December -- see https://sources.debian.org/src/cddlib/094l-2/debian/changelog/#L14.

DanGrayson:
What prompted me to do this was that on the mac here, they are found in /usr/local/include/cddlib, and we were looking only in /usr/include, /usr/include/cdd, and /usr/include/cddlib.  It was just too much kludgery.

Doug, you can't have a different debian build script for each version of ubuntu, can you?

d-torrance:
> Doug, you can't have a different debian build script for each version of ubuntu, can you?

Sure -- I actually have two (one for Debian unstable and Ubuntu >= 21.04 and another for Ubuntu 18.04 and 20.04) in different branches.

It would be nice to keep supporting `/usr/include/cdd` for now for the sake of the Ubuntu Github actions, though.

DanGrayson:
Okay, this latest version ought to do it.

PR #2207, opened 2021-08-05
Some typos

fchapoton:
suggested by codespell


DanGrayson:
Thanks!

But could you avoid updating the emacs submodule?

fchapoton:
I am afraid I still do not master how to handle "submodules". Sorry..

d-torrance:
One way to fix it:

```
git checkout origin/development -- M2/Macaulay2/editors/emacs/
```

fchapoton:
hopefully better now. This  business of submodules is source of troubles..

DanGrayson:
Thanks!

Issue #321, opened 2015-07-22
GCD in iterated polynomial ring?


pzinn:
Speaking of iterated polynomial rings, a problem I've encountered recently.
I'm not sure why one can't compute say
gcd(x^2-t^2,x^3-t^3)
in R2=R[t], R=QQ[x]. (it returns: error: not implemented yet)

Apparently the condition is that R2 should be "factoryAlmostGood" but I think this is too restrictive. 
In fact
rawGCD(raw (x^2-t^2), raw (x^3-t^3))
produces the correct answer.
Any thoughts?

DanGrayson:
Iterated polynomial rings are represented internally by "flattened" polynomial rings, so we should be able to make this work.

pzinn:
... 6 years later, I want to close this issue. but first, a basic problem: shouldn't `gcd` test that the ring is a UFD?
```
i1 : R=ZZ[x]/(x^2+3)

o1 = R

o1 : QuotientRing

i2 : gcd(4,2*(1+x)) -- that's one answer:

o2 = 2

o2 : R

i3 : (1+x)*(1-x) -- but (1+x) would work too

o3 = 4

o3 : R
```

DanGrayson:
Or better, just give an error if the ring is a quotient ring.

PR #2209, opened 2021-08-10
fix of #321

pzinn:
Rings are flattened if necessary to compute gcd.
A test is added.
An error is produced is ring is a quotient ring since we can't guarantee it's a UFD.

### DIFF
--- a/M2/Macaulay2/m2/factor.m2
+++ b/M2/Macaulay2/m2/factor.m2
@@ -17,10 +17,14 @@ gcd(RingElement,ZZ) := (r,s) -> gcd(promote(s,ring r),r)
 gcd(RingElement,RingElement) := RingElement => (r,s) -> (
      R := ring r;
      if ring s =!= R then error "gcd: expected elements in the same ring";
+     if instance(R,QuotientRing) then error "gcd: unimplemented for this ring";
      if isField R then if r == 0 and s == 0 then 0_R else 1_R
      else if factoryAlmostGood R then (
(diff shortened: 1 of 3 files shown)


pzinn:
checks have failed. interesting. could it be  some packages rely on gcd working even with quotient rings?

DanGrayson:
It looks like an example in the Graphs package uses "gcd" on a quotient ring, which is why at least one, and probably all, of the builds failed.

https://github.com/Macaulay2/M2/pull/2209/checks?check_run_id=3286586524#step:11:38934

pzinn:
`noetherianOperators` of the package `NoetherianOperators` also uses `gcd` in quotient rings.
At this stage, several options:
- remove the test / error altogether
- replace it with a warning -- is there a standardized way of doing that?
- fix packages / examples...

DanGrayson:
No, better to have an error than a warning.  Why not consult with the authors of the examples to see what they had in mind -- why they thought the quotient ring would support gcd?

Maybe look at the quotient rings to see whether it's obvious.

pzinn:
> It looks like an example in the Graphs package uses "gcd" on a quotient ring, which is why at least one, and probably all, of the builds failed.
> 
> https://github.com/Macaulay2/M2/pull/2209/checks?check_run_id=3286586524#step:11:38934

that one is easy to fix. I'll check with the authors if they're OK with my fix.

DanGrayson:
Rerunning all the jobs, so we can get at least one of them to go through all the tests.

DanGrayson:
Thanks!

PR #2214, opened 2021-08-24
Fix workspace sizes for CC LAPACK routines

jchen419:
This PR fixes some segfaults when calling LAPACK routines (e.g. eigenvalues, eigenvectors) with large matrices (>= 200x200, double precision RR/CC).

Continuation of Macaulay2/M2@151b5be0a25fcd7c4756cf445c4b4a625c5aea08.

### DIFF
--- a/M2/Macaulay2/e/lapack.cpp
+++ b/M2/Macaulay2/e/lapack.cpp
@@ -2046,7 +2046,7 @@ bool Lapack::eigenvalues(const LMatrixRR *A, LMatrixCC *eigvals)
   bool ret = true;
   char dont = 'N';
   int wsize = 3 * size;
-  double *workspace = newarray_atomic(double, wsize);
+  double *workspace = newarray_atomic(double, 2*wsize);
   int info;
 
   double *copyA = make_lapack_array(*A);
@@ -2115,7 +2115,7 @@ bool Lapack::eigenvectors(const LMatrixRR *A,
   char doit = 'V';
   int wsize = 4 * size;
 
-  double *workspace = newarray_atomic(double, wsize);
+  double *workspace = newarray_atomic(double, 2*wsize);
   int info;
 
   double *copyA = make_lapack_array(*A);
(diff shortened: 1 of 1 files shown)


antonleykin:
Could you post the updated test code here?

DanGrayson:
Thanks, Justin.  Could you tell us how the changes prevent segmentation faults?  Were the previous numbers simply wrong, for some reason?

jchen419:
@antonleykin: Here is some test code which (on my machine) segfaults without the fix, and no longer segfaults with the fix:
```
testRoutine = method(Options => true)
testRoutine (ZZ, ZZ, Ring, Function) := List => true >> opts -> (m, n, k, f) -> (
	apply(m, i -> (
		A := random(k^n, k^n);
		if opts.?Hermitian and opts.Hermitian then A = A + matrix table(n, n, (i,j) -> conjugate A_(j,i));
		elapsedTime (A, f(A, opts))
	))
)

setRandomSeed 1
L = testRoutine(10, 200, CC_53, eigenvectors);
apply(L, p -> (A = p#0; (E, V) = p#1; elapsedTime norm(A*V - V*diagonalMatrix E)))
```

jchen419:
@DanGrayson: Anton and I looked through the LAPACK documentation, and found that the specified sizes in `lapack.cpp` sometimes differed from the recommended sizes (mostly for CC routines - I should note that we did not modify the RRR/CCC routines, since we expect those to be handled by Eigen in the future). E.g. for eigenvectors one can see LWORK specified [here](http://www.netlib.org/lapack/explore-html/db/d55/group__complex16_g_eeigen_ga0eb4e3d75621a1ce1685064db1ac58f0.html#ga0eb4e3d75621a1ce1685064db1ac58f0). The edits here should now make `lapack.cpp` match the LAPACK documentation.

DanGrayson:
What is the "4" in the line `int wsize = 4 * size` ?

jchen419:
Hi Dan, are you referring to line 2116? From [here](http://www.netlib.org/lapack/explore-html/d9/d8e/group__double_g_eeigen_ga66e19253344358f5dee1e60502b9e96f.html#ga66e19253344358f5dee1e60502b9e96f): "if JOBVL = 'V' or JOBVR = 'V', LWORK >= 4*N".

DanGrayson:
Thanks.  

It will be good to get rid of lapack, if it has bugs that cause segmentation faults.

antonleykin:
Just a note: I ran the piece of code Justin attached (thanks!) with `development` built on Fedora 34 and, indeed, there is no crash.
I'm wondering about the other function: e.g., has `SVD` crashed (or returned nonsense) before?

Issue #2227, opened 2021-09-07
misbehavior of `needsPackage`


DanGrayson:
It shouldn't try to load it a second time:

```m2
Macaulay2, version 1.18.0.1

i1 : needsPackage ("Foo",FileName => "./Foo.m2")

o1 = Foo

o1 : Package

i2 : needsPackage ("Foo",FileName => "./Foo.m2")
Foo.m2:1:1:(3):[11]: error: package Foo not reloaded; try Reload => true
Foo.m2:1:1:(3):[11]: --entering debugger (type help to see debugger commands)
Foo.m2:1:1-1:1: --source code:
newPackage "Foo"
```

d-torrance:
I think the problem is that it's expecting the `FileName` option and `pkg#"source path"` to agree as strings:
https://github.com/Macaulay2/M2/blob/6742a499d5a9e428b0438926b5c5f495fe4bf554/M2/Macaulay2/m2/packages.m2#L181-L186

But they may not, e.g., `./Foo.m2` v. `/full/path/to/current/dir/Foo.m2`.  One possible fix:
```diff
diff --git a/M2/Macaulay2/m2/packages.m2 b/M2/Macaulay2/m2/packages.m2
index 07745fd1c..a63df7379 100644
--- a/M2/Macaulay2/m2/packages.m2
+++ b/M2/Macaulay2/m2/packages.m2
@@ -181,7 +181,8 @@ needsPackage = method(TypicalValue => Package, Options => options loadPackage)
 needsPackage String  := opts -> pkgname -> (
     if PackageDictionary#?pkgname
     and instance(pkg := value PackageDictionary#pkgname, Package)
-    and (opts.FileName === null or opts.FileName == pkg#"source file")
+    and (opts.FileName === null or
+       realpath opts.FileName == realpath pkg#"source file")
     then use value PackageDictionary#pkgname
     else loadPackage(pkgname, opts))

```

DanGrayson:
Thanks!

Would you like to submit a pull request?  Or shall I just make the change?

d-torrance:
I'd already made the changes in a local branch, so I went ahead and pushed it and opened #2229.

PR #2229, opened 2021-09-07
Use realpath when resolving filenames for needsPackage

d-torrance:
Otherwise, we may not recognize that the two paths actually refer to
the same file and so there's not need to reload the package.

Closes: #2227


DanGrayson:
Thanks!

Issue #2230, opened 2021-09-07
assertion failing in check(4, "NumericalImplicitization") on arm64


d-torrance:
Today, the continuous integration tests on the arm64 Debian package started failing (full logs [here](https://ci.debian.net/data/autopkgtest/unstable/arm64/m/macaulay2/15080634/log.gz) and [here](https://ci.debian.net/data/autopkgtest/unstable/arm64/m/macaulay2/15093457/log.gz)):

```m2
 -- capturing check(4, "NumericalImplicitization")                           -- capture failed; retrying ...
 -- running   check(4, "NumericalImplicitization")                          
 ulimit -c unlimited; ulimit -t 700; ulimit -m 850000; ulimit -s 8192; ulimit -n 512;  cd /tmp/M2-23113-0/2-rundir/; GC_MAXIMUM_HEAP_SIZE=400M "/usr/bin/M2-binary" -q --int --no-randomize --no-readline --silent --stop --print-width 77 -e 'needsPackage("NumericalImplicitization",Reload=>true,FileName=>"/usr/share/Macaulay2/NumericalImplicitization.m2")' <"/tmp/M2-23113-0/1.m2" >>"/tmp/M2-23113-0/1.tmp" 2>&1
/tmp/M2-23113-0/1.tmp:0:1: (output file) error: Macaulay2 exited with status code 1
/tmp/M2-23113-0/1.m2:0:1: (input file)
M2: *** Error 1
 -- 3.62783 seconds elapsed
 -- capturing check(5, "NumericalImplicitization")                           -- 2.09619 seconds elapsed
 -- capturing check(6, "NumericalImplicitization")                           -- 2.58701 seconds elapsed
 -- capturing check(7, "NumericalImplicitization")                           -- 15.7876 seconds elapsed
 -- skipping  check(8, "NumericalImplicitization")                           -- 0.000084171 seconds elapsed
 -- capturing check(9, "NumericalImplicitization")                           -- 0.606022 seconds elapsed
 -- capturing check(10, "NumericalImplicitization")                          -- 0.070274 seconds elapsed
 -- capturing check(11, "NumericalImplicitization")                          -- 0.759777 seconds elapsed
/usr/share/Macaulay2/NumericalImplicitization/tests.m2:59:1-67:1: error:
 -- Points found: 12
 -- Points found: 15
 -- Points found: 15
 -- Points found: 15
 -- Points found: 16
 -- Points found: 16
 -- Points found: 16
 -- Points found: 16
[...]

Issue #2223, opened 2021-09-03
Errors compiling eigen.cpp


d-torrance:
After #2221 was merged, there have been some errors compiling eigen.cpp for various PPA builds:

### Ubuntu 20.04 arm64

```
[...]

PR #2236, opened 2021-09-17
minor changes to linear algebra; compilation time/RAM improved for eigen.cpp

antonleykin:
corrected NO_LAPACK flag placement; 
made DivideConquer=>true default;
mentioned eigen in docs

### DIFF
--- a/M2/Macaulay2/e/eigen.cpp
+++ b/M2/Macaulay2/e/eigen.cpp
@@ -1,3 +1,10 @@
+//#define NO_LAPACK
+/*
+Uncommenting the above has two consequences:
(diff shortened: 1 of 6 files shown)


DanGrayson:
The failure seems to be related to changing `SVD = method(Options=>{DivideConquer=>false})` to `SVD = method(Options=>{DivideConquer=>true})`.

DanGrayson:
See also #2230

d-torrance:
I applied this as a patch to the current draft of the Debian package and it fixed the build errors in #2223.

DanGrayson:
That's good news, thanks, Doug.

PR #2252, opened 2021-10-08
Debug monideal crashes

mikestillman:
This pull request **should not be merged!**. It is only here to be able to run github actions to hopefully finally figure out the monideals sat bug that is sporadic and crashes M2.

### DIFF
--- a/M2/Macaulay2/e/monideal.cpp
+++ b/M2/Macaulay2/e/monideal.cpp
@@ -801,6 +801,7 @@ MonomialIdeal *MonomialIdeal::quotient(const MonomialIdeal &J) const
       result = next_result;
     }
   debug_in_sat = false;
+  std::cout << "  leaving quotient" << std::endl;
   return result;
 }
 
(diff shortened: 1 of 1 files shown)


mahrud:
(Just a note: I only turned this into a draft to prevent it from being accidentally merged, following Mike's comment, but github checks also run on draft PRs)

mahrud:
Looking at the logs, the last few lines are:
```m2
about to delete result
called: remove_MonomialIdeal
deleting node 0x45b44c0
deleting node 0x45b4500
  leaving remove_MonomialIdeal
entering next, p = 0x45b48c0
  about to go left, p = 0x45b48c0
  p = 0
-- SIGSEGV
```
It looks like here:
https://github.com/Macaulay2/M2/blob/6bf9dbcd6386ac504ac97a21848216ad3126ce7c/M2/Macaulay2/e/monideal.cpp#L271-L281
`p = 0x45b48c0` is not null, but `p = p->left` sets it to null, so then `if(p->tag == ...)` crashes. Shouldn't it just return null if `p->left` is null? The same for `prev()`, I presume:
https://github.com/Macaulay2/M2/blob/6bf9dbcd6386ac504ac97a21848216ad3126ce7c/M2/Macaulay2/e/monideal.cpp#L300-L308

Issue #2257, opened 2021-10-11
"intersect" and "intersection" do the same thing


DanGrayson:
Get rid of "intersection", as it is the more recent one.

```m2

i1 : apropos "intersect"

o1 = {intersect, intersectInP, intersection}

o1 : List

i2 : help intersect

o2 = intersect -- compute an intersection
     ************************************

     Description
     ===========

     This function calculates the intersection of a list or sequence of compatible objects.

     ...

i3 : help intersection

o3 = intersection -- compute an intersection
     ***************************************

     Description
     ===========

     This function calculates the intersection of a list or sequence of compatible objects.

     ...
```

d-torrance:
Fixed by #3742

Issue #1668, opened 2020-12-08
Missing documentation nodes in Macaulay2Doc


mahrud:
After #1666 is merged, there will be 214 warnings left, and 1370 node are not subnodes, which would be good to improve.

- [x] missing nodes:
- [x] CC'
- [x] GCstats
- [x] RR'
- [x] SymbolBody
- [x] columnRankProfile
- [x] isWeylAlgebra
- [x] monomialCurveIdeal
- [x] nullSpace
- [x] rowRankProfile
- [x] symbolBody
- [x] utf8check
- [ ] missing methods:
- [ ] (ambient,AffineVariety)
- [ ] (ambient,CoherentSheaf)
- [ ] (ambient,GradedModule)
- [ ] (ambient,ProjectiveVariety)
- [ ] (binomial,RingElement,ZZ)
- [ ] (columnRankProfile,MutableMatrix)
- [ ] (debug,ZZ)
- [ ] (degree,ChainComplexMap)
- [ ] (degree,GradedModuleMap)
- [ ] (degree,Matrix)
- [ ] (indices,HashTable)
- [ ] (isHomogeneous,Number)
- [ ] (isHomogeneous,PolynomialRing)
- [ ] (isHomogeneous,QuotientRing)
- [ ] (isWeylAlgebra,Ring)
- [ ] (jacobian,RingElement)
- [ ] (lift,Matrix,CC',CC')
- [ ] (lift,Matrix,RR',RR')
- [ ] (lift,Module,InexactNumber')
- [ ] (lift,Module,InexactNumber)
- [ ] (lift,Module,Number)
- [ ] (lift,Module,RingElement)
- [ ] (lift,MutableMatrix,InexactNumber')
- [ ] (lift,MutableMatrix,InexactNumber)
- [ ] (lift,MutableMatrix,Number)
- [ ] (lift,MutableMatrix,RingElement)
- [ ] (liftable,CC,RR')
- [ ] (map,ChainComplex,ChainComplex,ChainComplexMap)
- [ ] (map,GaloisField,GaloisField)
- [ ] (module,Ideal)
- [ ] (nullSpace,MutableMatrix)
- [ ] (part,InfiniteNumber,InfiniteNumber,Number)
- [ ] (part,InfiniteNumber,ZZ,Number)
- [ ] (part,ZZ,InfiniteNumber,Number)
- [ ] (part,ZZ,Number)
- [ ] (part,ZZ,ZZ,Number)
[...]

Issue #1203, opened 2020-05-27
packages sharing immutable symbols without values


DanGrayson:
Many packages introduce immutable symbols with capitalized names to use as names of optional arguments, and there is no reason such things need to be private to the package (they don't get values assigned).  Perhaps we should add another dictionary to the path for symbols in common use, shared by packages, and ensure that the user gets a new warning when exporting such symbols that shadow another.  There would be a variant of "export" called "exportShared" that puts such symbols into that new dictionary for all to share, or gets it from there, if it's already there.

DanGrayson:
There should also be something similar for methods.

DanGrayson:
We could also allow packages to create method functions, since typically packages install methods for different combinations of types.

PR #2263, opened 2021-10-23
fix documentation of shared symbols

mahrud:
Fixes https://github.com/Macaulay2/M2/issues/2257#issuecomment-950042791.

### DIFF
--- a/M2/Macaulay2/m2/document.m2
+++ b/M2/Macaulay2/m2/document.m2
@@ -196,8 +196,8 @@ makeDocumentTag String      := opts -> key -> (
 -- correct its package, if it is incorrect (e.g. truncate, quotient)
 -- TODO: can this be modified to fix the tag in-place? then we would only need to
 -- fix the tag in (validate, TO), rather than also in (info, TO) and (html, TO).
-fixup DocumentTag := DocumentTag => tag -> makeDocumentTag(
-    if instance(key := tag.Key, String) then return tag else key, Package => package key)
+fixup DocumentTag := DocumentTag => tag -> (
+    if (rawdoc := fetchAnyRawDocumentation tag) =!= null then rawdoc.DocumentTag else tag)
 
 -----------------------------------------------------------------------------
 -- formatting document tags
--- a/M2/Macaulay2/m2/help.m2
+++ b/M2/Macaulay2/m2/help.m2
@@ -387,37 +387,18 @@ getBody := (key, tag, rawdoc) -> (
 
 -- TODO: help symbol% before Macaulay2Doc is installed doesn't work
 help = method(Dispatch => Thing)
--- overview nodes and formatted documentation keys
-help String := key -> (
-    rawdoc := fetchAnyRawDocumentation makeDocumentTag key;
-    tag := getOption(rawdoc, symbol DocumentTag);
-    if tag.?Key and tag.Key =!= key then help tag.Key
-    else if      isGlobalSymbol key then help getGlobalSymbol key
-    else getBody(key, tag, rawdoc))
+help DocumentTag := tag -> (
+    rawdoc := fetchAnyRawDocumentation tag;
+    tag = if rawdoc =!= null then rawdoc.DocumentTag else tag;
+    getBody(tag.Key, tag, rawdoc))
 
--- Methods
 help Sequence := key -> (
-    if key === () then return if inDebugger then debuggerUsageMessage else help "initial help";
-    -- TODO: make this work with hook strategies; e.g. (foo, ZZ, Strategy => Default)
-    if lookup key === null then error("expected ", toString key, " to be a method");
-    rawdoc := fetchAnyRawDocumentation makeDocumentTag key;
-    tag := getOption(rawdoc, symbol DocumentTag);
-    getBody(key, tag, rawdoc))
+    if key =!= () then help makeDocumentTag key else
+    if inDebugger then debuggerUsageMessage else help "initial help")
 
--- Options
-help Array := key -> (
-    verifyKey key;
-    rawdoc := fetchAnyRawDocumentation makeDocumentTag key;
-    tag := getOption(rawdoc, symbol DocumentTag);
-    getBody(key, tag, rawdoc))
-
--- everything else: Symbols, Types, ScriptedFunctors, Functions, Keywords, and Packages
-help Symbol := key -> (
-    rawdoc := fetchAnyRawDocumentation makeDocumentTag key;
-    tag := getOption(rawdoc, symbol DocumentTag);
-    getBody(key, tag, rawdoc))
-
-help DocumentTag := tag -> help tag.Key
+help String :=
+help Symbol :=
+help Array :=
 help Thing := x -> help makeDocumentTag x
 help List  := l -> DIV between(HR{}, help \ l)
 help ZZ    := i -> seeAbout(help, i)


DanGrayson:
I'm not sure this pull request addresses the real issue, which seems to be that this documentation node refers requests for documentation of `intersection` to the page for `intersect`:

```m2
document {
    Key => { intersect, (intersect, List), (intersect, Sequence), intersection },
    Headline => "compute an intersection",
    PARA {
	"This function calculates the intersection of a list or sequence of compatible objects."
	},
    SeeAlso => {
	-- add references to intersect methods installed in packages _other than Core_ here
	"M0nbar::M0nbar",
	"NAGtypes::NAGtypes",
	"Polyhedra::Polyhedra"
	}
    }
```

mahrud:
No, I'm pretty sure it's addressing the problem. In particular, this isn't a problem only with intersection.

You can try separating intersection into it's own node in Macaulay2Doc and you'll get the same problem.

mahrud:
Try links to `isEmpty` and `isEmpty(Polyhedron)` in the documentation of Polyhedra. Here is the source:
```html
<li><span><a href="../../Macaulay2Doc/html/_is__Empty.html">isEmpty</a> -- checks if a Polyhedron is empty</span>      </li>
<li><span><a title="checks if a Polyhedron is empty" href="_is__Empty.html">isEmpty(Polyhedron)</a> -- checks if a Polyhedron is empty</span>      </li>
```
The real issue, which this PR fixes, is that for instance this html file in the Polyhedra package is created incorrectly: https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2/share/doc/Macaulay2/Polyhedra/html/_is__Empty.html

In this case, the node in `shared.m2` is simply
```m2
document { Key => isEmpty,      methodstr, SeeAlso => { "Polyhedra::Polyhedra",(isEmpty, RRi)} }
```
which isn't making any misdirections, as you suggest.

DanGrayson:
There should be no documentation node for "isEmpty" in the Polyhedra package, because "isEmpty" is in the coere, and an error should be signaled when it tries to make one here:
```m2
document {
     Key => {isEmpty, (isEmpty,Polyhedron)},
     Headline => "checks if a Polyhedron is empty",
     Usage => " b = isEmpty P",
     Inputs => {
...
```

mahrud:
> There should be no documentation node for "isEmpty" in the Polyhedra package

Why? There's no reason to disallow this, when with the changes in this PR both can exist at the same time. This was a necessary change a while back.

For instance, a common word like "quotient" or "euler" can mean different things in the context of different packages.

DanGrayson:
> > There should be no documentation node for "isEmpty" in the Polyhedra package
> 
> Why? There's no reason to disallow this, when with the changes in this PR both can exist at the same time. This was a necessary change a while back.
> 
> For instance, a common word like "quotient" or "euler" can mean different things in the context of different packages.

Because the symbol "isEmpty" is exported by the Core.  Each package is allowed to document only its own symbols.  it has nothing to do with the "meaning" of the symbol in the context of the package -- the package can document the methods it installs on the shared symbol, and that's sufficient.

Actually, it would be nice if we had a way to tell, after the fact, which package installed a given method function, so we don't allow the wrong package to document a method function.

mahrud:
> Each package is allowed to document only its own symbols.

This has not been the case for a few versions, and the reason is that it made the documentation system more annoying and less effective.

mahrud:
> Actually, it would be nice if we had a way to tell, after the fact, which package installed a given method function, so we don't allow the wrong package to document a method function.

Why? These are shared symbols. The whole point is to allow packages to share them, this include sharing the documentation!

mahrud:
> the package can document the methods it installs on the shared symbol, and that's sufficient.

Here's the reason that this is insufficient:
Let's say package A exports a function (not even a method!) 'fancyFunc', and package B also exports 'fancyFunc', but these two packages don't depend on each other, so they both need to export it. Then at some point the dependencies change and one is loaded before the other, and suddenly there's an error, when 'fancyFunc' could do completely different things in package A and B and they may not even be methods to be distinguished by their input types. What are package writers supposed to do then?

This is just an unnecessary requirement that was removed for good reason, and there's no point in bringing it back.

Meanwhile, this PR is a simplification of the documentation system that happens to also fix this problem, so there's also no point here in discussing whether to bring back that requirement.

DanGrayson:
> > Each package is allowed to document only its own symbols.
> 
> This has not been the case for a few versions, and the reason is that it made the documentation system more annoying and less effective.

We'll have to put that back to the way it was before.

DanGrayson:
> > Actually, it would be nice if we had a way to tell, after the fact, which package installed a given method function, so we don't allow the wrong package to document a method function.
> 
> Why? These are shared symbols. The whole point is to allow packages to share them, this include sharing the documentation!

Yes, but the methods are not shared.  If foo is a shared method and two packages install methods for `foo(X,Y,Z)`, then that's a conflict that should result in an error.

DanGrayson:
> > the package can document the methods it installs on the shared symbol, and that's sufficient.
> 
> Here's the reason that this is insufficient: Let's say package A exports a function (not even a method!) 'fancyFunc', and package B also exports 'fancyFunc', but these two packages don't depend on each other, so they both need to export it. Then at some point the dependencies change and one is loaded before the other, and suddenly there's an error, when 'fancyFunc' could do completely different things in package A and B and they may not even be methods to be distinguished by their input types. What are package writers supposed to do then?
> 
> This is just an unnecessary requirement that was removed for good reason, and there's no point in bringing it back.
> 
> Meanwhile, this PR is a simplification of the documentation system that happens to also fix this problem, so there's also no point here in discussing whether to bring back that requirement.

If two packages export symbols with the same name, there should be no problem.  Each package should be able to document its own symbols.

mahrud:
> Yes, but the methods are not shared. If foo is a shared method and two packages install methods for `foo(X,Y,Z)`, then that's a conflict that should result in an error.

Plenty of packages install the same exact method on types with the same name. This requirement limits what you can document using Macaulay2.

> If two packages export symbols with the same name, there should be no problem. Each package should be able to document its own symbols.

With your arbitrary requirement this would result in an error.

DanGrayson:
> > Yes, but the methods are not shared. If foo is a shared method and two packages install methods for `foo(X,Y,Z)`, then that's a conflict that should result in an error.
> 
> Plenty of packages install the same exact method on types with the same name. This requirement limits what you can document using Macaulay2.

I intended X, Y, and Z to be the same types, not types with the same names.

> 
> > If two packages export symbols with the same name, there should be no problem. Each package should be able to document its own symbols.
> 
> With your arbitrary requirement this would result in an error.

No, my requirement was that each package should document only its own symbols.  That's not determined by looking at the name of the symbol, but by looking at the symbol itself.

mahrud:
> I intended X, Y, and Z to be the same types, not types with the same names.

And what happens if a package intends to overwrite a method installed by another package in order to improve it or cover more possibilities?

> No, my requirement was that each package should document only its own symbols. That's not determined by looking at the name of the symbol, but by looking at the symbol itself.

What if one of those packages is Core or in one of the pre-loaded packages?

---

These are not hypothetical issues. They are all issues I've had to deal with repeatedly in multiple packages I've contributed to. Your resistance to see that this requirement is pointless is counterproductive because instead of trying to make documentation more difficult you could be making it easier by fixing any of #666, #1203, #1649, #1668, or just properly documenting how you designed symbols, symbol bodies, and frames so that someone else can fix those issues.

DanGrayson:
> > I intended X, Y, and Z to be the same types, not types with the same names.
> 
> And what happens if a package intends to overwrite a method installed by another package in order to improve it or cover more possibilities?

That shouldn't be done.

> 
> > No, my requirement was that each package should document only its own symbols. That's not determined by looking at the name of the symbol, but by looking at the symbol itself.
> 
> What if one of those packages is Core or in one of the pre-loaded packages?

Okay, what if?  Is there a problem?

mahrud:
@DanGrayson regardless of your disagreement on what nodes should be documentable by a package, this pull request fixes a bug in the internal `fixup` method and simplies the `help` method, so it should be merged.

Issue #1945, opened 2021-02-17
List << List


mahrud:
This is a suggestion to implement the following method:
```m2
i1 : List << List := (A, B) -> #A == #B and all(#A, i -> A#i <= B#i);

i2 : {1,2,3} << {1,2,3}

o2 = true

i3 : {1,2,3} << {1,5,3}

o3 = true

i4 : {1,2,-1} << {1,5,3}

o4 = true

i5 : {1,2,3} << {1,5,-1}

o5 = false
```
This would be a partial order, which is why `A#i <= B#i` makes more sense, because with `<` we wouldn't know if `A << B and B << A` implies `A==B` or not.

Of course, there are many ways to impose a total order on lists, one for each monomial order probably, but I think this is a natural order to have because `A << B` always implies that the multinomial `x^B` is in the ideal of `x^A`.

I'd like to have this as a simplification in order to generalize `DegreeLimit` options, for instance:
- `res(..., DegreeLimit => {1,2,3})` should compute up to and including degree `<< {1,2,3}` (hence `<=` above)
- `basis({0,0,0},{1,2,3}, R)` currently is not accepted by the basis command, but it could just print the basis for all multidegrees `{0,0,0} << D << {1,2,3}`
- this would match how `submatrixByDegrees` interprets the results.

Tangentially related: I think `CC < CC` should give an error. It's not mathematically correct!

mahrud:
@mikestillman @DanGrayson any objections to this?

DanGrayson:
I can't think of a better use for List << List.  In the documentation, please compare it to the behavior of List ? List and friends (such as List <= List).

What's wrong with CC < CC?   Yes, CC is not an ordered field, but it's convenient, for printing, to put things in order.

mahrud:
> What's wrong with CC < CC? Yes, CC is not an ordered field, but it's convenient, for printing, to put things in order.

Plenty of things that are wrong are convenient, that's not a good reason. If user wants to sort some complex numbers before printing them, then they should provide such a function. This result is meaningless:
```m2
i1 : sort{1+ii, 1-ii}

o1 = {1-ii, 1+ii}

o1 : List

i2 : sort{2_(ZZ/11), 3_(ZZ/11)}

o2 = {3, 2}

o2 : List
```
It should give an error in both cases.

mahrud:
At the very least, a one time warning should be given that this order is made up, like this one:
```m2
-- warning: experimental computation over inexact field begun
[...]

Issue #1358, opened 2020-06-30
HH on AffineVariety


lcrantonh:
HH seems to be assuming that the variety is projective.
```m2
R = QQ[x,y,z]
I = ideal(z^2-y,y*z-x,y^2-x*z)
M = coker (res I).dd_2
F = sheaf(Spec R,M) --says sheaf is a quotient of OO_{Proj R}^3(-2)
HH^0(F) --returns QQ^2 but should be M
HH^1(F) --returns QQ but should be 0
```

DanGrayson:
It's definitely confused:
```
i4 : F = sheaf(Spec R,M) --says sheaf is a quotient of OO_{Proj R}^3(-2)

o4 = cokernel {2} | -z 1  |
              {2} | y  -z |
              {2} | -x y  |

                                                   3
o4 : coherent sheaf on Spec R, quotient of OO       (-2)
                                             Proj R

i4 : F = sheaf(Spec R,M) --says sheaf is a quotient of OO_{Proj R}^3(-2)

o4 = cokernel {2} | -z 1  |
              {2} | y  -z |
              {2} | -x y  |

                                                   3
o4 : coherent sheaf on Spec R, quotient of OO       (-2)
                                             Proj R
```

Thanks for the bug report.

PR #2289, opened 2021-11-08
Truncations v1.0

mahrud:
This is a significant rewrite of the Truncations package. The main improvements are:

- reorganizing and clarifying the algorithms with comments
- adding support for Cox rings of simplicial normal toric varieties
- fixing truncation of modules with relations
- making `truncate` accept arbitrary options (this involves minor changes to some other packages)
- significantly optimizing a whole bunch of algorithms
- adding a bunch of new tests for things that were wrong before
- fixing the documentation to correctly describe what truncation does
- also closing #1945.

Small note: commit aec5e02 is a short-term patch for a much larger problem that should be resolved in https://github.com/Macaulay2/M2/issues/1358.

cc: @mikestillman, @eisenbud


mahrud:
The only failing test is from SpecialFanoFourfolds. @DanGrayson you might want to take a look at this also, while Mike reviews it, as there are some changes in the Core included here (mainly to fix a few documentation issues).

DanGrayson:
> @DanGrayson you might want to take a look at this also

will do ...

M2/Macaulay2/m2/lists.m2
```diff
@@ -175,6 +175,8 @@ rotate(ZZ,VisibleList) := (n,s) -> (
 sort List :=  opts -> internalsort
 rsort List := opts -> internalrsort
 
+List << List := (A, B) -> all(min(#A, #B), i -> A#i <= B#i)
```
DanGrayson, line 178: Let's not add this.  We already have comparison for lists with the operator `<=`, and `<<` doesn't look like a comparison operator.

DanGrayson:
> > @DanGrayson you might want to take a look at this also
> 
> will do ...

I made one comment.

M2/Macaulay2/m2/lists.m2
mahrud: This was discussed in https://github.com/Macaulay2/M2/issues/1945. Quoting your comment there:
> I can't think of a better use for List << List. In the documentation, please compare it to the behavior of List ? List and friends (such as List <= List).

DanGrayson: Oh, okay, never mind.

mikestillman:
@DanGrayson can you resolve the conflict in changes.m2?  Then accept this pull request?

PR #2298, opened 2021-11-10
Remove Python from list of "improved packages" in 1.19 changes.

d-torrance:
It's also listed in "new packages", which is more appropriate because it hasn't been distributed before.


DanGrayson:
Well, this is strange -- I've merged your pull request, but running "git pull" on the development branch doesn't get it.

Issue #2039, opened 2021-05-04
minimalBetti doesn't accept monomial ideals


eisenbud:
The doc for MonomialIdeal claims that routines that take an ideal as arg will also take a monomial ideal, but this is not true of minimalBetti

PR #2301, opened 2021-11-11
add in new packages QuartenaryQuartics and GroebnerStrata, added chan…

mikestillman:
2 new packages: QuartenaryQuartics, GroebnerStrata


mikestillman:
@mahrud There is an interesting bug, that I didn't notice on the caching pull request.  I still need to check, but here is the problem: compute a resolution of an ideal I (e.g. compute regularity).  Then try minimalBetti I.  This fails, as the resolution computed is the wrong type.  Probably minimalBetti should use the cached resolution that is there, if it is done.  The previous behavior though was to simply recompute the resolution, using FastNonminimal.  Either way, it gave an answer, now we get an error.

I can fix this example or test so it doesn't fail.  But maybe we should fix this behavior.  It might not be so easy though...  Thoughts?

Here is the bug (I should make this into a github issue):
```
R = ZZ/101[a..d]
I = monomialCurveIdeal(R, {1,3,4})
regularity I -- behind the scenes, computes a free resolution of I
minimalBetti I -- fails, as it requires a FastNonmimal resolution type.
```

mahrud:
I've seen that error in other issues related to minimalBetti, like #2039, so I think it's a bug in minimalBetti that should be fixed.

mahrud:
I think minimalBetti should just call `betti` if a minimal resolution is already cached, right? There's no point in using FastNonminimal in that case.

mahrud:
@mikestillman #2305 fixes this issue.

mahrud:
I'm not sure if rerunning the checks picks up the changes in that PR. You can try merging with development.

mikestillman:
@mahrud Thanks!  @DanGrayson: I would really like this pull request in version 1.19.  Hopefully the latest change in #2305 will fix this problem.  We will see in a few hours.  There is one more type in an example that I will fix, but first I'll wait for these tests to finish.

mikestillman:
@DanGrayson Should I merge in development into my this branch?  I think the changes in #2305 (which were merged into development) fix the errors that are happening.

PR #2307, opened 2021-11-15
Add new package K3s

giovannistagliano:
A package for the constructions of K3 surfaces as described in the paper https://arxiv.org/abs/2110.15819.


DanGrayson:
"K3s" is too cryptic.  What about naming the package "K3Surfaces" instead?

giovannistagliano:
Okay, I'm going to rename the package.

PR #2311, opened 2021-11-16
Document how to build with Python support

d-torrance:


d-torrance:
@DanGrayson -- when you build the 1.19 documentation for the website, could you build using the `--with-python` option so that the `Python` documentation is included?  I noticed that it was missing in your [post to the Google group](https://groups.google.com/g/macaulay2/c/z84AKp3hvlE/m/lmqvf_wcAgAJ).

DanGrayson:
> @DanGrayson -- when you build the 1.19 documentation for the website, could you build using the `--with-python` option so that the `Python` documentation is included? I noticed that it was missing in your [post to the Google group](https://groups.google.com/g/macaulay2/c/z84AKp3hvlE/m/lmqvf_wcAgAJ).

I tried that, but I couldn't get it to work.

PR #2312, opened 2021-11-16
Include Python documentation even when Macaulay2 is built w/o Python support

d-torrance:
When Macaulay2 is built without Python support, we load dummy versions of all of the methods, e.g.,:

```m2
i1 : loadPackage "Python"

o1 = Python

o1 : Package

i2 : rs "2 + 2"
stdio:2:1:(3): error: Macaulay2 was built without Python support
```

But this still allows us to build the documentation using cached examples.


DanGrayson:
Thanks!

DanGrayson:
How did you succeed without setting CacheExampleOutput to true?

d-torrance:
> How did you succeed without setting CacheExampleOutput to true?

I didn't.  :)  I just didn't commit that change to git.  Should I have?

DanGrayson:
Ah, yes.  The point is that if it is set to true, then when the optional components are present it will rerun the examples, with possibly updated output.  And when you add examples, the output from them will be cached.

I'll do it for you.

PR #2314, opened 2021-11-18
A few string features

d-torrance:
In particular:

* Merge `characters` and `toList` for strings, since they do the exact same thing.  In fact, `characters` had a faster implementation (using a pre-computed array of one-character strings instead of building them on the fly), so we adopt it for all of the various string features from [#1897](https://github.com/Macaulay2/M2/pull/1897).  `characters` is now defined only on the top level.
* Export `reverse` for strings.  It already existed in the interpreter, but wasn't available from top level.

```m2
i1 : reverse "Hello, world!"

o1 = !dlrow ,olleH
```

* Implement `pack` for strings.
```m2
i2 : pack("The quick brown fox jumps over the lazy dog", 5)

o2 = {The q, uick , brown,  fox , jumps,  over,  the , lazy , dog}

o2 : List
```

* Fix some inconsistent behavior when 0 is an argument to `pack`:

### Before
```m2
i1 : pack(0, {})

o1 = {}

o1 : List

i2 : pack({}, 0)
stdio:2:1:(3): error: expected argument 2 to be a positive integer
```

### After
```m2
i1 : pack(0, {})

o1 = {}

o1 : List

i2 : pack({}, 0)

o2 = {}

o2 : List
```

Thanks to @pzinn for the suggestions!


d-torrance:
I also added `tally(String)`:

```m2
i1 : tally "Hello, world!"

o1 = Tally{  => 1}
           ! => 1
           , => 1
           d => 1
           e => 1
           H => 1
           l => 3
           o => 2
           r => 1
           w => 1

o1 : Tally
```

Issue #2315, opened 2021-11-18
TypeError: unhashable type: 'list'


d-torrance:
Found some strange behavior in the `Python` package:

```m2
1 : loadPackage "Python"

o1 = Python

o1 : Package

i2 : rand = import "random"

o2 = <module 'random' from '/usr/lib/python3.9/random.py'>

o2 : PythonObject of class module

i3 : rand@@choice toPython {1,2,3}
Traceback (most recent call last):
  File "/usr/lib/python3.9/random.py", line 346, in choice
    return seq[self._randbelow(len(seq))]
TypeError: unhashable type: 'list'
stdio:3:5:(3): error: python error
```
This shouldn't be an issue.  Running the equivalent code in Python itself:

```python
>>> import random
>>> random.choice([1,2,3])
1
```

Also, it would be nice if we didn't need `toPython` in the above example, but right now it interpets the list as separate arguments without `toPython`:

```m2
i4 : rand@@choice {1, 2, 3}
TypeError: choice() takes 2 positional arguments but 4 were given
stdio:4:5:(3): error: python error
```

d-torrance:
Another example:

```m2
i1 : loadPackage "Python"

o1 = Python

o1 : Package

i2 : L = toPython {1,2,3}

o2 = [1, 2, 3]

o2 : PythonObject of class list

i3 : L + L
TypeError: unhashable type: 'list'

The above exception was the direct cause of the following exception:

SystemError: <method-wrapper '__add__' of list object at 0x7f6d8a4302c0> returned a result with an error set
stdio:3:3:(3): error: python error
```

PR #2316, opened 2021-11-18
A few Python updates

d-torrance:
* Interpret lists (and other non-sequence visible lists) as arguments themselves and not as lists of arguments for `toFunction` (addressing part of [#2315 (comment)](https://github.com/Macaulay2/M2/issues/2315#issue-1057501584)).

### Before
```m2
i1 : loadPackage "Python"

o1 = Python

o1 : Package

i2 : (rs "len") {1, 2, 3}
TypeError: len() takes exactly one argument (3 given)
stdio:2:1:(3): error: python error
```

### After
```m2
i2 : (rs "len") {1, 2, 3}

o2 = 3

o2 : PythonObject of class int
```

* Convert all visible lists to Python lists with `toPython` unless another method function is installed (like `toPython(Sequence)`, which converts sequences to tuples).

```m2
i3 : toPython [1, 2, 3]

o3 = [1, 2, 3]

o3 : PythonObject of class list

i4 : toPython <|1, 2, 3|>

o4 = [1, 2, 3]

o4 : PythonObject of class list
```

* Add a couple more `value(PythonObject)` hooks for useful classes from the `collections` module.

```m2
i7 : c = import "collections"

o7 = <module 'collections' from '/usr/lib/python3.9/collections/__init__.py'>

o7 : PythonObject of class module

i8 : d = c@@defaultdict rs "int"

o8 = defaultdict(<class 'int'>, {})

o8 : PythonObject of class collections.defaultdict

i9 : d_"foo"

o9 = 0

o9 : PythonObject of class int

i10 : value d

o10 = HashTable{foo => 0}

o10 : HashTable

i11 : c@@Counter "foo"

o11 = Counter({'o': 2, 'f': 1})

o11 : PythonObject of class collections.Counter

i12 : value oo

o12 = Tally{f => 1}
            o => 2

o12 : Tally
```

### DIFF
--- a/M2/Macaulay2/packages/Python.m2
+++ b/M2/Macaulay2/packages/Python.m2
@@ -190,6 +190,8 @@ addHook((value, PythonObject),
     Strategy => "unknown -> PythonObject")
 addPyToM2Function({"function", "builtin_function_or_method", "method-wrapper"},
     toFunction, "function -> FunctionClosure")
+addPyToM2Function("Counter", x -> new Tally from dictToHashTable x,
+    "Counter -> Tally")
 addPyToM2Function({"dict", "defaultdict"}, dictToHashTable, "dict -> HashTable")
 addPyToM2Function({"set", "frozenset"}, set @@ iterableToList, "set -> Set")
 addPyToM2Function("list", iterableToList, "list -> List")
--- a/M2/Macaulay2/packages/Python.m2
+++ b/M2/Macaulay2/packages/Python.m2
@@ -434,6 +434,12 @@ assert Equation(5 ^^ y, 7)
 assert Equation(x xor y, rs "7")
 assert Equation(x xor 2, 7)
 assert Equation(5 xor y, 7)
+
+----------------------
+-- unary operations --
+----------------------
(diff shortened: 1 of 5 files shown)


d-torrance:
Just added a few more commits adding unary operations:

```m2
i2 : x = rs "5"

o2 = 5

o2 : PythonObject of class int

i3 : -x

o3 = -5

o3 : PythonObject of class int

i4 : +x

o4 = 5

o4 : PythonObject of class int
```

DanGrayson:
The function name "rs" is pretty cryptic.  Can you improve that (eventually)?

d-torrance:
`rs` (which already existed in the package before my updates) is a wrapper around `runPythonString`. It concatenates the arguments and prints some debugging info if `debugLevel` is positive. Maybe we should make `runPythonString` do this instead and just make `rs` an alias?

DanGrayson:
Maybe I'm responsible for creating "rs".  But still, it's too cryptic, even as a synonym, and will make code unreadable.  `runPythonString` sounds good.  Or `pythonValue`, which is like `value`.

PR #2321, opened 2021-11-21
Update vim files for 1.19

d-torrance:


DanGrayson:
It's too late for this to make it into 1.19, which has already been tagged.

d-torrance:
I don't think these files are distributed (not in the Debian package at least), so it shouldn't matter that it's past the tag. Any vim users could grab them from the git repository.  (Not a vim user myself, but figured I push the updates since they were generated when I worked on the Pygments PR.)

PR #2322, opened 2021-11-22
"Ways to use" improvements

mahrud:
- add "Ways to use" for method Commands (e.g. help) and Manipulators (e.g. flush)
- add "Ways to use this method" for method keys

### DIFF
--- a/M2/Macaulay2/m2/help.m2
+++ b/M2/Macaulay2/m2/help.m2
@@ -49,10 +49,11 @@ counter := 0
 next := () -> counter = counter + 1
 optTO := key -> (
     tag := makeDocumentTag(key, Package => package key);
+    ptag := getPrimaryTag tag;
     fkey := format tag;
+    if currentHelpTag.?Key and instance(currentHelpTag.Key, Sequence) and currentHelpTag =!= ptag then return;
     if isUndocumented tag then return;
     if isSecondaryTag tag then (
-	ptag := getPrimaryTag tag;
 	-- this is to avoid doubling "\" in documentation for symbol \ and symbol \\
 	ref := if match("\\\\", fkey) then concatenate("/// ", fkey, " ///") else format fkey;
 	-- TODO: figure out how to align the lists using padding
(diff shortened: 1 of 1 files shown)


pzinn:
this PR causes trouble in WebApp mode. specifically, `LATER` (is this a new feature? what is it for?) does not produce the correct output.

pzinn:
I think the content of `LATER` should always be a `Hypertext` element, not just a string or sequence of strings.

pzinn:
anyway I can't pinpoint when the problem appeared but my comment stands:
I think the content of `LATER` should always be a `Hypertext` element, not just a string or sequence of strings. e.g. 
```
text = if text =!= null and #text > 0 then text else if tag =!= opttag then LATER {() -> SPAN headline opttag};
```

mahrud:
I'm pretty sure anything that can be rendered by html (or other output method) is fine. Do you know what document tag is causing an issue? I wonder if `headline` was returning null or something due to #2323?

mahrud:
Here is a run on WebApp mode:
```m2
[mahrud@noether m2]$ M2 --webapp --silent
i1 : debug Core;
debug Core;
i2 : a = LATER {() -> "abcd"};
a = LATER {() -> "abcd"};
i3 : html a
html a

o3 = <pre style="display:inline">
&lt;pre style=&quot;display:inline&quot;>
abcd&lt;/pre></pre>
i4 : a
a

o4 = <pre style="display:inline">
abcd</pre>

o4 : <pre style="display:inline">
LATER</pre>
```
The main difference I notice here is the last output; where it's printing `LATER` instead of evaluating it.

pzinn:
sorry about the successive edits making my point unclear. a screenshot is probably easier:
- before the suggested fix above:
before
- after:
after
notice the change of font in the LATER part

pzinn:
ah, oops. `headline` does return `null` occasionally. `nonnull SPAN headline opptag` then?

mahrud:
Sorry, I'm still not following the issue is related to the use of strings in `LATER`, or how the output you get is different.

That's a silly text to appear after `Degree`, by the way. Someone should combine the documentation of `matrix` and `(matrix, List)` and `[matrix, Degree]` I think.

pzinn:
I'm not judging the content, just the form. in the first version the `LATER` text is output in `PRE` rather than the normal font.

mahrud:
Or maybe that's not feasible the way `LATER` is designed, in which case yea your solution sounds good.

Issue #732, opened 2018-02-15
Odd behavior of gb / poincare


pzinn:
Let's run the following program:

```
i1 : R=QQ[a..k]

o1 = R

o1 : PolynomialRing

i2 : I=trim ideal(a*b+c*d,a*e+b*f+c*g+d*h);

o2 : Ideal of R

i3 : M=cokernel gens I

o3 = cokernel | ae+bf+cg+dh ab+cd |
                           
o3 : R-module, quotient of R^1

i4 : poincare M

o4 = 1 - 2T^2  + T^4

o4 : ZZ[T]

i5 : M.cache.poincare

o5 = 1 - 2T^2  + T^4

o5 : ZZ[T]

i6 : dim I

o6 = 9
```

Basically, I compute the poincare polynomial of some ideal. really internally that's the poincare polynomial of the module M which is the quotient module. and I check that M has correctly cached that poincare polynomial for future use. So far, so good.
Now I restart M2 and do the following (slightly weird) thing:

```
i1 : R=QQ[a..k]

o1 = R

o1 : PolynomialRing

i2 : I=trim ideal(a*b+c*d,a*e+b*f+c*g+d*h);

o2 : Ideal of R

i3 : M=cokernel gens I

o3 = cokernel | ae+bf+cg+dh ab+cd |
                            
o3 : R-module, quotient of R^1

i4 : use degreesRing R

o4 = ZZ[T]

o4 : PolynomialRing

i5 : M.cache.poincare=1-2*T^2+T^4;

i6 : dim I

o6 = 10
```

I just tried to be helpful and gave the poincare polynomial to M so M2 doesn't have to compute it. surprise: the dim of I (or M) comes out wrong.
what happened? to see the problem let's check what the Groebner basis was.
in the first case one gets

```
i8 : leadTerm gb presentation M

o8 = | ae ab cde |
```

which is fine, whereas in the second one gets

```
i6 : leadTerm gb presentation M

o6 = | ae ab |
```

which is incorrect. how did it get there? I'd have to go through the gb code and figure out how it uses the information on the poincare polynomial. but one thing is sure, something is fishy...

pzinn:
I'd be happy to help, but it seems the computation of g.b. is done at a fairly low level, and I'd need some pointers on how the code is organized...

mikestillman:
Sigh, the `trim` command seems to be double computing a minimal generating set, at least in the case when the argument is an Ideal.  I wonder when this behavior started.

[...]

Issue #2338, opened 2021-11-27
poincare in presence of degree zero variables


mahrud:
Perhaps it should give an error instead:
```m2
i1 : R = kk[x,y, Degrees => {0,1}]

o1 = R

o1 : PolynomialRing

i2 : poincare ideal (x,y)

o2 = 0

o2 : ZZ[T]

i8 : poincare ideal (x+y^2)

          2
o8 = 1 - T

o8 : ZZ[T]
```
c.f. https://github.com/Macaulay2/M2/issues/909

PR #2339, opened 2021-11-27
TLC for hilbert*/poincare/etc.

### LINKED ISSUE #50: comparing a Hilbert polynomial to zero
Hi,

I was trying to compare a Hilbert polynomial to zero and discovered that M2 doesn't let me do this.  I would like to be able to do this, and kludge that lets me do this is below.

I don't think this is a bug per-say and maybe I should be posting this to google groups instead of here.

Nonetheless, here is the code.

```m2
Macaulay2, version 1.6
with packages: ConwayPolynomials, Elimination, IntegralClosure, LLLBases, PrimaryDecomposition, ReesAlgebra, TangentCone

i1 : A = QQ[x,y,z,w]

o1 = A

o1 : PolynomialRing

i2 : I = coker gens monomialCurveIdeal(A,{1,2,3})

o2 = cokernel | z2-yw yz-xw y2-xz |

                            1
o2 : A-module, quotient of A

i3 : hilbertPolynomial(I)

o3 = - 2_P  + 3_P
          0      1

o3 : ProjectiveHilbertPolynomial

i4 : hilbertPolynomial(I) != 0
stdio:4:22:(3): error: no method for binary operator == applied to objects:
--            - 2_P  + 3_P  (of class ProjectiveHilbertPolynomial)
--                 0      1
--     ==     0 (of class ZZ)

On the other hand,

i5 : hilbertPolynomial(A^0) 

o5 = 0

o5 : ProjectiveHilbertPolynomial

i6 : hilbertPolynomial(A^0) == hilbertPolynomial(I)

o6 = false
```

I want i4 to return false like i6 does.
Here is a very bad kludge to get around this.

```m2
i7 : 
     -- here is a VERY bad kludge to check if a hilbert polynomial is zero.
     hilbertPolynomial ZZ := ProjectiveHilbertPolynomial => o -> (M) -> ( 
         new ProjectiveHilbertPolynomial from {}
         )

o7 = ProjectiveHilbertPolynomial => {_Function[stdio:9:58-10:9]_}

o7 : Option

i8 : ProjectiveHilbertPolynomial == ZZ := (M,N) -> (M == hilbertPolynomial N)

o8 = {_Function[stdio:12:44-12:71]_}

o8 : FunctionClosure

i9 : 
     hilbertPolynomial(A^0) == 0

o9 = true

i10 : hilbertPolynomial(I) != 0

o10 = true

i11 : hilbertPolynomial(A^0) != 0

o11 = false
```

Any suggestions?

mahrud:
- moved code related to Hilbert functions to hilbert.m2
- added (heft, Monoid)
- adds arithmetic ops for ProjectiveHilbertPolynomial, closes #50
- move tests related to hilbert.m2
- hookify poincare, hilbertPolynomial, and hilbertFunction
- hookify and cache degree and multidegree
- hookified and improved caching for hilbertSeries
- minor change in matrix1.m2
- minor changes in modules2.m2
- add 'poincare M = hf' for storing the poincare polynomial
- update documentation of poincare


mahrud:
@mikestillman a few questions about `poincare` for you:
- I updated the documentation, since I thought the previous explanation of the mathematical meaning of the output was wrong. Could you check that the new documentation is correct?
- The documentation often treats the result of `poincare` as the Hilbert function, but that's not accurate, right? Is it just that modules with the same Hilbert function have the same numerator of the reduced Hilbert series?
- I wanted to make some changes to `poincareN`, but I didn't to avoid conflicting with Complexes stuff. Here is my suggestion, which I'm happy to make for the version in the Complexes package if you'd like:
https://github.com/Macaulay2/M2/blob/6fb09f7037e82ed753be6358a03d694001b99106/M2/Macaulay2/m2/chaincomplexes.m2#L485-L486

mahrud:
Also, I removed `installHilbertFunction` (since "Hilbert function" here is inaccurate, I believe) in favor of `poincare M = p`, which should be easier to remember. `installHilbertFunction` still works, but by calling this routine.

If you have any suggestions for how to fix #732 or #2338, I'm happy to add them to this PR. You seemed to be close to figuring out the issue back then.

Issue #1515, opened 2020-09-26
registerFinalizer not working


mahrud:
The [examples in the documentation](https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.16/share/doc/Macaulay2/Macaulay2Doc/html/_register__Finalizer.html) haven't printed the string since at least v1.12:
```m2
      +-------------------------------------------------------------------------------------+
      |i1 : R = QQ[a..d];                                                                   |
      +-------------------------------------------------------------------------------------+
      |i2 : makeGB = (n) -> (g := gb((ideal vars R)^4); registerFinalizer(g, "gb("|n|")"););|
      +-------------------------------------------------------------------------------------+
      |i3 : for i from 1 to 10 do (makeGB i);                                               |
      +-------------------------------------------------------------------------------------+
      |i4 : collectGarbage()                                                                |
      +-------------------------------------------------------------------------------------+
```

DanGrayson:
Good catch.

DanGrayson:
Compiling a debug version gives some informative output:
```m2
Macaulay2, version 1.17.1.1
with packages: ConwayPolynomials, Elimination, IntegralClosure, InverseSystems, LLLBases,
               MinimalPrimes, PrimaryDecomposition, ReesAlgebra, Saturation, TangentCone

i1 : 	  R = QQ[a..d];

i2 : 	  makeGB = (n) -> (g := gb((ideal vars R)^4); registerFinalizer(g, "gb("|n|")"););

i3 : 	  for i from 1 to 10 do (makeGB i);

i4 : 	  collectGarbage()
GC Warning: Finalization cycle involving 0x1132e1060
GC Warning: Finalization cycle involving 0x1135779c0
GC Warning: Finalization cycle involving 0x11305a6c0
GC Warning: Finalization cycle involving 0x1130bf540
GC Warning: Finalization cycle involving 0x11322e960
GC Warning: Finalization cycle involving 0x11311dd80
GC Warning: Finalization cycle involving 0x1134d5d80
GC Warning: Finalization cycle involving 0x113290900
GC Warning: Finalization cycle involving 0x113178de0
GC Warning: Finalization cycle involving 0x1135e5480
```

DanGrayson:
Here's the way it looks in 1.6:
```m2
Macaulay2, version 1.6
with packages: ConwayPolynomials, Elimination, IntegralClosure, LLLBases, PrimaryDecomposition,
               ReesAlgebra, TangentCone

[...]

PR #2344, opened 2021-12-02
Null out values in the dictionaries created by capture before returning

### LINKED ISSUE #2330: Capture memory usage
`capture` seems to have a slow memory leak. Running the following loop on my system increases M2's memory usage by ~0.5MB/s seemingly indefinitely. (I ran it till M2 was using 300MB).

```
while true do (scan(apply(100,i-> "x = 1;y = 2;"),capture);collectGarbage())
```
The following loop seems to leak much faster (1GB within 15min), but it's harder to figure out what's leaking:
```
needsPackage "SpecialFanoFourfolds"
while true do (scan(values tests SpecialFanoFourfolds,capture @@ code);collectGarbage())
```
In contrast the following loop does not cause an increase in memory usage:
```
while true do (scan(apply(100,i-> "x = 1;y = 2;"),v -> (x = 1;y = 2;));collectGarbage())
```

Valgrind does not seem to detect a leak.

jkyang92:
This will partially fix #2330. Each time `new Dictionary` is called at the M2 level, and then values stored to symbols from that dictionary, the values actually go into `globalFrame` and thus require an entry in that frame. However these entries are not reclaimable in any way.

This change at least prevents those values from leaking by nulling out those entries before the exit from capture. This does not prevent the `globalFrame` itself from getting larger, but that is a much smaller issue.

This is not the optimal fix, a better fix would be to allow the explicit creation of a Dictionary at the M2 level with a temporary frame, but that's a much larger change. This also does not fix that the Dictionary and the associated symbols leak, but that's a much smaller leak, and also somewhat harder to fix. 

@d-torrance This probably fixes one of the culprits of the out of memory test failures with capture, so it's worth testing to see which of those still occur with this. For that it's probably necessary to disable some of the checks in `isCapturable`.

### DIFF
--- a/M2/Macaulay2/m2/examples.m2
+++ b/M2/Macaulay2/m2/examples.m2
@@ -89,6 +89,8 @@ capture String := opts -> s -> if opts.UserMode then capture' s else (
     if not hasmode ArgNoPreload then
     scan(Core#"pre-installed packages", needsPackage);
     needsPackage \ toString \ flatten { if opts.PackageExports === null then currentPackage else opts.PackageExports };
+    --shallow copy, but we only need to remember which things started with attributes
+    oldAttributes := copy Attributes;
     -- TODO: is this still necessary? If so, add a test in tests/normal/capture.m2
     -- dictionaryPath = prepend(oldPrivateDictionary,      dictionaryPath); -- this is necessary mainly due to T from degreesMonoid
     dictionaryPath = prepend(User#"private dictionary", dictionaryPath); -- this is necessary mainly due to indeterminates.m2
@@ -97,14 +99,16 @@ capture String := opts -> s -> if opts.UserMode then capture' s else (
     ret := capture' s;
     collectGarbage();
 
+    --Without the toSequence {v}, if v is a Sequence, hasAnAttribute breaks
     scan(value \ values User#"private dictionary", v ->
-	if hasAttribute(v, ReverseDictionary) then removeAttribute(v, ReverseDictionary));
+        if hasAnAttribute toSequence {v} and not oldAttributes#?v
+        then remove(Attributes,v));
     -- null out all symbols in the private dictionary, otherwise those values leak
     -- See bug #2330 for details.
-    scan(values User#"private dictionary", s -> s <- null);
+    scan(values User#"private dictionary", s -> (if mutable s then s <- null));
     User#"private dictionary" = oldPrivateDictionary;
     -- null out the symbols in the OutputDictionary as well
-    scan(values OutputDictionary, s -> s <- null);
+    scan(values OutputDictionary, s -> (if mutable s then s <- null));
     popvar symbol OutputDictionary;
     -- TODO: this should eventually be unnecessary
     scan(keys oldMutableVars, symb -> symb <- oldMutableVars#symb);
@@ -136,8 +140,7 @@ isCapturable = (inputs, pkg, isTest) -> (
     and not match("(addHook|export|newPackage)",              inputs) -- these commands have permanent effects
     and not match("(installMethod|installAssignmentMethod)",  inputs) -- same as above
     and not match("(Global.*Hook|add.*Function|Echo|Print)",  inputs) -- same as above
-    and not match("(method)",                                 inputs) -- same as above
-    and not match("(protect)",                                inputs) -- currently capture tries to clear all symbols, so protect breaks it
+    and not match("(importFrom|exportFrom)",                  inputs) -- currently capture tries to clear all symbols created, these break it
     )
 
 -----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


mahrud:
Not directly related, but this may have been much easier to detect if `registerFinalizer` worked (see #1515), since we would have been about to tell whether the memory was being reclaimed after `capture`.

jkyang92:
Well I broke something in the example generation by doing this. I was worried about this, but I didn't have issues with most of the tests, so I thought it was fine. I'll take a look

mahrud:
Ah, I think your fix attempted to clear a protected variable in an example:
```m2
 -- capturing example results for "protect"                                 ../../../../../Macaulay2/m2/examples.m2:104:51:(1):[15]: error: assignment to protected variable 's'
```
You can add `-* no-capture-flag *-` for that one.

mahrud:
It might be easier to just add `protect` to the list in this function:
https://github.com/Macaulay2/M2/blob/729dcedb9437682f631c99e1473e47732dc06d1f/M2/Macaulay2/m2/examples.m2#L112

mahrud:
On that note, a great test would be removing some of the packages from this list and seeing how the tests do now:
https://github.com/Macaulay2/M2/blob/729dcedb9437682f631c99e1473e47732dc06d1f/M2/Macaulay2/m2/examples.m2#L122

jkyang92:
Okay, I finally figured out the build failures. Unfortunately I don't have a fix yet. The problem is `capture` can break the `package` command (among other things)

```m2
capture("R=QQ",UserMode=>false);
print package QQ;
```
This prints `null` instead of `Core`. This happens independent of my changes, but my changes made it so that this breaks the documentation.

The cause of this is that capture tries to remove all newly defined things from the Attributes table (examples.m2:100). However, it can't actually tell if the code being captured added it to the attributes table, or if it was in the attributes table to start, so it just removes everything that's in the temporary dictionary.

DanGrayson:
Instead of checking for the keyword "protect" in the code, why not alter the loop, changing it from
```m2
 scan(values User#"private dictionary", s -> s <- null);
```
to
```m2
 scan(values User#"private dictionary", s -> if mutable s then s <- null);
```
?

mahrud:
> Instead of checking for the keyword "protect" in the code, why not alter the loop, changing it from

The example that failed called `protect` on a new variable. We don't want `capture` to leak protected variables in the dictionary after it's done.

DanGrayson:
> > Instead of checking for the keyword "protect" in the code, why not alter the loop, changing it from
> 
> The example that failed called `protect` on a new variable. We don't want `capture` to leak protected variables in the dictionary after it's done.

Presumably the variable had no value, so little space was wasted.

jkyang92:
First, I'm going with what @DanGrayson suggested, using `mutable` to detect which ones we can actually clear.

Second, `importFrom` and `exportFrom` cause serious issues with the code to clear the dictionaries, because they create new `SymbolClosure`s in `User#"private dictionary"` that point to the same entries as other `SymbolClosure`s. This as far as I can tell is undetectable, and so I've had to simply forbid them. Luckily, their use is rare.

Third, I had to mark examples that add implementations of methods to existing types with `no-capture-flag`, otherwise these methods leak out of the example. Methods on types that did not exist previously, are probably safe.

Finally, it turns out that resetting the whole `Attributes` table breaks `needsPackage`, so I'm going with a modified version of what was in capture in the first place, but instead of removing everything in `User#"private dictionary"` from `Attributes`, I remove only those that weren't there before. The important point here is that we don't want to delete anything that was already in the table, which caused me a lot of pain.

I'm going to let this run, and hopefully I don't get more failures, it all works on my system now. If these tests pass I'm going to push my changes that re-enable capture on `DiffAlg` and `MultiprojectiveVarieties`. I did test memory usage of the packages, and on my computer I could remove all of the package special cases, but I suspect that `EngineTests` and `SpecialFanoFourfolds` at least will still fail on some systems. They use a lot of memory, just not enough GC memory to cause the limit to complain. Also probably we should just keep `ThreadedGB` and `RunExternalM2` there for now, since those don't seem like they are going to use capture anyways. I would really appreciate if someone with Macualay2 setup on a non amd64 system could test some of these packages as well. 

What I have now is a partial fix, and we will still leak some values, but I have some thoughts on a better fix. But these would take more effort to implement, and I'm not going to be able to implement it right now.

1. It would be better to have "weak hash tables" for Attributes, which don't keep the keys live. This will need support from scc though. If we had this, we wouldn't need to worry about clearing the Attributes table since the entries would naturally disappear when garbage collection occurred.
2. As I've stated before, it would be better if the `Dictionary` objects created by capture had their own frame, this way manual deletion and the associated pains could be avoided. Additionally, it would be much harder for values from capture to leak. However this is very involved, because a lot of code assumes that there is only one global frame.

jkyang92:
So I was too optimistic about `MultiprojectiveVarieties`. It still fails at least on i686. But Core is fine now on 32 bit systems (See also #1834) . I don't really have the right setup to test `DiffAlg` but it works on all systems I'm able to build M2 on (See also #1728).

DanGrayson:
Thanks, good fix!

PR #2354, opened 2021-12-09
update FindTBB.cmake and mathicgb hash

mahrud:
@mikestillman @moorewf I think this (and the connected PR https://github.com/Macaulay2/mathicgb/pull/31) should resolve your issue. I'm setting this to a draft. Let me know if the issue is fixed and I'll mark it as ready.


mahrud:
@DanGrayson there's an accompanying PR that you missed.

PR #2362, opened 2021-12-21
Update Givaro and fflas-ffpack

### LINKED ISSUE #2347: givaro_bool_constant does not name a type
The following line 20 in `aring-zzp-ffpack.hpp` is causing compilation issues in `aring-zzp-ffpack.cpp` for me on Gentoo, due to `givaro_bool_constant` not being defined
```c++
#define bool_constant givaro_bool_constant
```
Looking at the logs, it seems this is to fix some c++17/14 compatability issues with givaro and ffpack? But since givaro is the one defining the potentially conflicting synonym, it seems to me there's no reason not to just delete this line (and the associated undef). In particular, both givaro and c++17 agree that `bool_constant<B>` is the same as `integral_constant<bool,B>`, so I don't see why we need to force ffpack to use givaro_bool_constant. In particular deleting the line works for me.

Also what systems were having this issue to start with, and do they still have this issue? Deleting all references to `givaro_bool_constant` still lets me compile on both Gentoo and Arch.

mahrud:
- update to givaro 4.2.0 and fflas_ffpack 2.5.0
- removed the old GCC diagnostic pragmas
- correct their entries in the copyright string
- set rpath when linking with libomp
- add Eigen and TBB to copyright

This fixes #2347. I don't think the diagnostic pragmas are still needed either, so I removed them for now.

@DanGrayson: you should update https://github.com/Macaulay2/givaro and https://github.com/Macaulay2/fflas-ffpack.

@mikestillman: incidentally, I've been experimenting with removing both givaro and fflas-ffpack and using flint for rank computation in the F4 algorithms, and haven't noticed a significant speed difference. Do you have benchmark examples where fflas-ffpack is faster?

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -132,7 +132,7 @@ jobs:
         run: |
           cmake -S../.. -B. -GNinja \
             -DCMAKE_BUILD_TYPE=Release -DBUILD_NATIVE=OFF \
-            -DCMAKE_PREFIX_PATH="`brew --prefix`;`brew --prefix tbb@2020_u3`" \
+            -DCMAKE_PREFIX_PATH="`brew --prefix`;`brew --prefix tbb@2020`" \
             -DCMAKE_INSTALL_PREFIX=/usr \
             -DWITH_PYTHON=ON \
             --debug-trycompile
@@ -163,8 +163,8 @@ jobs:
           make -C ../.. all
           export PYVERSION=`python3 -c "from sys import version_info; \
             print(f'{version_info.major}.{version_info.minor}')"`
-          export CPPFLAGS="-I`brew --prefix`/include -I`brew --prefix tbb@2020_u3`/include"
-          export  LDFLAGS="-L`brew --prefix`/lib     -L`brew --prefix tbb@2020_u3`/lib \
+          export CPPFLAGS="-I`brew --prefix`/include -I`brew --prefix tbb@2020`/include"
+          export  LDFLAGS="-L`brew --prefix`/lib     -L`brew --prefix tbb@2020`/lib \
             -L`brew --prefix python`/Frameworks/Python.framework/Versions/${PYVERSION}/lib"
           ../../configure --enable-download --with-python
 
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -39,7 +39,7 @@ jobs:
           - autotools
         os:
           - ubuntu-20.04
-          - macos-10.15
+          - macos-11
         compiler:
           - default
         include:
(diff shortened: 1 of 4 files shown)


mikestillman:
> @mikestillman: incidentally, I've been experimenting with removing both givaro and fflas-ffpack and using flint for rank computation in the F4 algorithms, and haven't noticed a significant speed difference. Do you have benchmark examples where fflas-ffpack is faster?

@mahrud Yes, in minimal betti computations, the non-minimal resolution is computed, then the ranks of the degree 0 parts of each map are created, and we need the ranks of these matrices.  Once the number of rows and columns of these matrices over a finite field are above perhaps 5000-10000 or so (I'll need to redo computations to see), and the matrices are dense, then ffpack rank computation was consistently 3 times faster than flint rank computation.  However, I don't have any tests in place that confirm this.  But perhaps after the latest version of flint, these values are not so different.  I/we should rerun these computations.  The examples I was using were the minimal betti numbers of inverse systems of one homogeneous polynomial (random, or sum of powers of random linear forms), and resolutions of Prym curves (examples from Frank Schreyer).  After about genus 16 to 20, these start getting quite big and difficult.  For genus 24, I am interested in whether one betti number jumps from the generic value (by 1).  This requires the rank of a square 3.25 million by 3.25 million matrix, which is fairly dense, I believe.  Out of range currently, without some simplification.

DanGrayson:
There's a givaro-related compilation error: https://github.com/Macaulay2/M2/runs/4595017962?check_suite_focus=true#step:11:2384

mahrud:
I'm aware. Feel free to push a fix if you're aware of any.

DanGrayson:
Okay, I've reproduced it at home ...

jkyang92:
The issue here is that givaro redefines `std::bool_constant`. On systems using libstdc++ this is fine, because duplicate using statements are safe. But for systems using libc++ for their c++ standard library, it defines `std::bool_constant` is a inline namespace, so it's real name is something like `std::_1::bool_constant`, but since it's an inline namespace, the name `std::bool_constant` also refers to it. But now when givaro redefines `std::bool_constant` there are two things both called `std::bool_constant`, and even though they are equivalent, this triggers an error.

Probably what givaro should do is put their definition behind an `#ifdef`. As I mentioned in #2347, there is a closed givaro bug about the error, but it seems to me that they misidentified the issue.

DanGrayson:
That's right.  I expect to have a work-around soon.

DanGrayson:
Okay, let's see if that works ...

DanGrayson:
Strange: `configure: error: tbb include file (tbb.h) not found`

DanGrayson:
Strange, because we also have this: 
```
relevant environment variable values, if any:
   CPPFLAGS = -I/usr/local/include -I/usr/local/opt/tbb@2020_u3/include
```

DanGrayson:
I see no reason for "make" to return an exit code of 1 here: https://github.com/Macaulay2/M2/runs/4662074665?check_suite_focus=true#step:10:903

PR #2365, opened 2021-12-27
Updates to Functors

mahrud:
- add type `Functor` as ancestor of `ScriptedFunctor`
- reorganize gateway.m2 into functors.m2
- added RingMap_* notation for pushForward
- switch RingMap to global assignment function
- fix printing of pushforward


DanGrayson:
It would be good to merge the development branch into this one, now that the other pull request has been granted.

M2/Macaulay2/tests/ComputationsBook/completeIntersections/test.out.expected
```diff
@@ -731,9 +731,9 @@ o108 : T'
 
 i109 : psi = map(K,B)
 
-o109 = map (K, B, {0, 0, 0})
+o109 = psi
 
-o109 : RingMap K <--- B
+o109 : RingMap K <--- B, given by {0, 0, 0}
```
DanGrayson, line 736: The idea of this output line is to simply give the type of the output.  The output line with an equal sign is where the output itself goes.

mahrud: Notice that the output two lines above has changed to just `psi`. That's because I changed `RingMap` to use global assignments instead.

DanGrayson: Either you want to see the value of psi in the output or you don't.  If you do, change RingMap back.

mahrud: No, this is how I wanted it, so that this would work nicely:
```m2
i1 : S = kk[x]; R = S[y];

i3 : f = map(R, S)

o3 = f

o3 : RingMap R <--- S, given by {x}

i4 : f_*

o4 = pushforward functor f_*

o4 : Functor
```
I want ring maps to have a global name so the pushforward functor is printed nicely while the information about the ring map is still visible in the `o3 : ` line.

mahrud:
I didn't make this change in this PR, but I'd like to change `<---` to something else so ring maps aren't half commented:
```m2
o3 = f

o3 : RingMap R <--- S, given by {x}
```

M2/Macaulay2/tests/ComputationsBook/completeIntersections/test.out.expected
DanGrayson: The output line with a colon on it is for the type only, not the value.  You can use `describe` to manually get more information about something that prints out as its name, as we do for rings.

mahrud: Okay, I'll remove the generators.

DanGrayson:
> I didn't make this change in this PR, but I'd like to change `<---` to something else so ring maps aren't half commented:
> 

That does currently look sort of bad:

Screen Shot 2022-01-30 at 8 10 30 AM

DanGrayson:
We could just reduce it from 3 hyphens to 1, as here:

Screen Shot 2022-01-30 at 8 12 03 AM

DanGrayson:
A functor is not a type, so this code doesn't make sense:
```m2

i6 : code lookup (describe, ScriptedFunctor)

o6 = ../../../../Macaulay2/m2/functors.m2:62:25-62:95: --source code:
     describe   Functor := F -> if F.?describe   then F.describe   else (lookup(describe,   Type)) F

i7 : ancestors Functor

o7 = {Functor, MutableHashTable, HashTable, Thing}

o7 : List
```

DanGrayson:
If you rename a file and then edit it before committing, it is very difficult to discern what editing you did, as git notices no similarity between the files.  In future, it would help me a lot if after renaming a file, it would be committed immediately, with no changes.

```sh
commit 7496be1bf35bb216f49c244089b6f700e1b94ee5
Author: Mahrud Sayrafi <mahrud@berkeley.edu>
Date:   Fri Nov 26 06:15:10 2021 -0600

    refactor gateway.m2 into functors.m2

 M2/Macaulay2/m2/chaincomplexes.m2                            |   1 +
 M2/Macaulay2/m2/code.m2                                      |   3 ---
 M2/Macaulay2/m2/document.m2                                  |  16 +++++++++-------
 M2/Macaulay2/m2/exports.m2                                   |   1 +
 M2/Macaulay2/m2/ext.m2                                       |   2 +-
 M2/Macaulay2/m2/functors.m2                                  | 107 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 M2/Macaulay2/m2/gateway.m2                                   |  87 ---------------------------------------------------------------------------------------
 M2/Macaulay2/m2/help.m2                                      |  14 +++++++-------
 M2/Macaulay2/m2/loadsequence                                 |  13 +++++++------
 M2/Macaulay2/m2/mathml.m2                                    |   2 +-
 M2/Macaulay2/m2/matrix1.m2                                   |   1 +
 M2/Macaulay2/m2/methods.m2                                   |   5 -----
 M2/Macaulay2/m2/tor.m2                                       |   2 +-
 M2/Macaulay2/m2/varieties.m2                                 |  13 ++++---------
 M2/Macaulay2/packages/Macaulay2Doc/debugging.m2              |   2 +-
 M2/Macaulay2/packages/Macaulay2Doc/doc12.m2                  |  15 ++++++---------
 M2/Macaulay2/packages/Macaulay2Doc/functions/help-doc.m2     |   2 +-
 M2/Macaulay2/packages/Macaulay2Doc/functions/homology-doc.m2 |   2 ++
 M2/Macaulay2/packages/Macaulay2Doc/methods.m2                |   2 +-
 19 files changed, 151 insertions(+), 139 deletions(-)
```

DanGrayson:
A functor is not a type, so this code also is misdirected:
```m2
     ../src/M2/M2.git/M2/Macaulay2/m2/functors.m2:61:25-61:95: --source code:
     expression Functor := F -> if F.?expression then F.expression else (lookup(expression, Type)) F
```

DanGrayson:
There don't seem to be any objects of class Functor, and there's only one type that inherits from it (ScriptedFunctor), so what's the reason for introducing it?

mahrud:
Not sure why you keep repeating that "a functor isn't a type." I've defined and documented it:
https://github.com/Macaulay2/M2/blob/6f017538a34f54c4aafee0aea9dd19aa09365e89/M2/Macaulay2/m2/functors.m2#L42

> There don't seem to be any objects of class Functor, and there's only one type that inherits from it (ScriptedFunctor), so what's the reason for introducing it?

Here's one:
https://github.com/Macaulay2/M2/blob/6f017538a34f54c4aafee0aea9dd19aa09365e89/M2/Macaulay2/m2/pushforward.m2#L249-L256

There will be more like this in the future.

Re: AfterPrint of ring maps, you said:
> We could just reduce it from 3 hyphens to 1

I'm fine with that as well. Feel free to do that either here or after you accept this.

DanGrayson:
> Not sure why you keep repeating that "a functor isn't a type." 

Oops, you're right.  Sorry about that.

mahrud:
No worries. Why is this PR labeled "under discussion"? I'm still waiting for your second review after implementing changes from the first one.

DanGrayson:
Okay, I'll take a look.

M2/Macaulay2/packages/Macaulay2Doc/doc12.m2
```diff
@@ -629,37 +629,34 @@ document {
      Key => {ScriptedFunctor,
 	  (symbol^, ScriptedFunctor, Thing),
 	  (symbol_, ScriptedFunctor, Thing),
-	  (symbol SPACE, ScriptedFunctor, Thing)},
+	  Functor, (symbol SPACE, Functor, Thing)},
```
DanGrayson, line 632: You've documented "Functor" by simply pointing to the documentation for "ScriptedFunctor", but that documentation node doesn't mention "Functor", nor does it explain the difference between the two.  Why not make a new node for "Functor" and explain it, really?  Include code showing an example of a functor that's not a scripted functor.

M2/Macaulay2/m2/document.m2
```diff
@@ -244,21 +245,22 @@ fSeq := new HashTable from {
...
+    -- TODO: make sure hh.argument is also covered
+    (4, class, Functor, ZZ) => s -> (
 	hh := s#0;
 	if hh.?subscript
 	then (toString hh, "_", toString s#1, "(", toString s#2, ",", toString s#3, ")")
 	else (toString hh, "^", toString s#1, "(", toString s#2, ",", toString s#3, ")")),
-    (3, class, ScriptedFunctor, ZZ) => s -> (
+    (3, class, Functor, ZZ) => s -> (
 	hh := s#0;
 	if hh.?subscript
```
DanGrayson, line 257: Isn't it just scripted functors that take subscripts, and not functors?  So this code is strange.

M2/Macaulay2/packages/Macaulay2Doc/doc12.m2
```diff
@@ -629,37 +629,34 @@ document {
...
      }
 
 document {
      Key => argument,
-     Headline => "specify the function in a scripted functor for an argument",
-     TT "argument", " -- a key used in scripted functors under which is
+     Headline => "specify the function in a functor for an argument",
+     TT "argument", " -- a key used in functors under which is
      stored the function that accepts the arguments.",
-     SeeAlso => "ScriptedFunctor"
```
DanGrayson, line 644: If you remove this line, then the reader of this documentation page has no quick way to get to the documentation for scripted functors.

M2/Macaulay2/packages/Macaulay2Doc/doc12.m2
```diff
@@ -629,37 +629,34 @@ document {
...
      }
 
 document {
      Key => subscript,
      Headline => "specify the function in a scripted functor for a subscript",
      TT "subscript", " -- a key used in scripted functors under which is
      stored the function of one variable that accepts the subscript and
-     returns a scripted functor that accepts the arguments.",
-     SeeAlso => "ScriptedFunctor"
```
DanGrayson, line 653: If you remove this line, then the reader of this documentation page has no quick way to get to the documentation for scripted functors.

M2/Macaulay2/packages/Macaulay2Doc/doc12.m2
```diff
@@ -629,37 +629,34 @@ document {
...
+     returns a functor that accepts the arguments.",
      }
 
 document {
      Key => superscript,
      Headline => "specify the function in a scripted functor for a superscript",
      TT "superscript", " -- a key used in scripted functors under which is
      stored the function of one variable that accepts the superscript and
-     returns a scripted functor that accepts the arguments.",
-     SeeAlso => "ScriptedFunctor"
```
DanGrayson, line 662: If you remove this line, then the reader of this documentation page has no quick way to get to the documentation for scripted functors.

DanGrayson (changes requested):

M2/Macaulay2/packages/Macaulay2Doc/doc12.m2
mahrud: These pages are subnodes, so there is a link displayed at the top.

M2/Macaulay2/packages/Macaulay2Doc/doc12.m2
mahrud: Same as above.

M2/Macaulay2/packages/Macaulay2Doc/doc12.m2
mahrud: Same as above.

M2/Macaulay2/packages/Macaulay2Doc/doc12.m2
mahrud: Is there really a point for separating Functor and ScriptedFunctor? Given that not all functors are scripted, I can just rename ScriptedFunctor to Functor if you prefer.

M2/Macaulay2/m2/document.m2
mahrud: This is a good point. I think just renaming ScriptedFunctor to Functor would be simpler and also resolve this.

mahrud: I fixed this for now.

M2/Macaulay2/packages/Macaulay2Doc/doc12.m2
mahrud: I documented Functor for now, but long term, it might be simpler to eliminate ScriptedFunctor.

DanGrayson: > Is there really a point for separating Functor and ScriptedFunctor? 

It's your pull request that introduces Functor, so you must have thought there was a point -- probably because `f_*` has neither a subscript nor a superscript.

M2/Macaulay2/m2/document.m2
mahrud: I think until now the entire point of `ScriptedFunctor` has been a technical thing, namely storing different functions for subscript and superscript, but there are many "functions" in Macaulay2 that are really functors, as in they take objects to objects and morphisms to morphisms, like `basis` or `truncate`. I think making that more explicit is good. This also helps with the general goal of keeping Macaulay2 syntax close to the mathematical notation.

M2/Macaulay2/packages/Macaulay2Doc/doc12.m2
mahrud: Here you go:
https://github.com/Macaulay2/M2/blob/0596fa84d9d169f5d6ec2a57072c25933a0941ab/M2/Macaulay2/packages/Macaulay2Doc/doc12.m2#L635
https://github.com/Macaulay2/M2/blob/0596fa84d9d169f5d6ec2a57072c25933a0941ab/M2/Macaulay2/packages/Macaulay2Doc/doc12.m2#L645
Did you try rebuilding the documentation to see this?
image

M2/Macaulay2/packages/Macaulay2Doc/doc12.m2
mahrud: Yes, I introduced `Functor` for mathematical reasons rather than syntactic reasons, and made `ScriptedFunctor` inherit from it for semantic reasons. But I do not care for the term `ScriptedFunctor` and would rather they be removed in a future PR.

DanGrayson:
I still don't see what f_* has in common with the various "scripted functors".  (Actually, a better name might be "scripted function".)

The reason for the existence of the class "ScriptedFunctor" is so that something like HH can appear in expressions such as `HH M` , ` HH_i M `, ` HH^i M `, and ` HH^i_Z M `.  The push forward map `f_*`, by contrast, appears only in expressions such as ` f_* M `, so ` f_* ` might as well just be a function.

mahrud:
Like I said before, there are many "functions" in Macaulay2 that are really functors, and just as there's a distinction between functions and methods of various kind, functors are special in certain ways and have mathematical properties that one might want to symbolically manipulate. Eventually we can have syntax for composition of functors or derived functors, but the first step is to have a proper type `Functor` to start building all of that on top of, and `f_*` is just a first example.

This PR is only short because you don't like long PR's, but it's certainly a prerequisite for other changes involving varieties.

DanGrayson:
Oh, in that case, `HH` is not a proper functor, because it doesn't come equipped with a source category and a target category.  Rather, it's a catch-all method function for lots of different functors.

Also, `hh` is not a functor, for the same reason as above, but also because its values are natural numbers, not objects in a category.

Maybe it's best to divorce Functor and ScriptedFunctor.  We could also change ScriptedFunctor to ScriptedFunction to avoid confusion.

M2/Macaulay2/packages/Macaulay2Doc/doc12.m2
DanGrayson: That's only visible in the html and info versions of the documentation.  If you type `help superscript`, you won't see it.

mahrud: That sounds like an issue that should be fixed in `help`, since `viewInfo` also shows the right thing. I made an issue for it.

mahrud:
> Oh, in that case, `HH` is not a proper functor, because it doesn't come equipped with a source category and a target category. Rather, it's a catch-all method function for lots of different functors.

Sure, `HH` is a symbol for different functors, but it's a proper functor in all cases, isn't it?

> Also, `hh` is not a functor, for the same reason as above, but also because its values are natural numbers, not objects in a category.

Sure, but `hh` is just the dimension of certain sheaf cohomology groups, and sheaf cohomology is a functor, so it's just a small abuse of notation.

> Maybe it's best to divorce Functor and ScriptedFunctor. We could also change ScriptedFunctor to ScriptedFunction to avoid confusion.

I'm not opposed to introducing `ScriptedFunction` in a future pull request, but most current objects of that type would have to change to being functors. I think the changes in this PR are useful regardless of future plans.

DanGrayson:
> > Oh, in that case, `HH` is not a proper functor, because it doesn't come equipped with a source category and a target category. Rather, it's a catch-all method function for lots of different functors.
> 
> Sure, `HH` is a symbol for different functors, but it's a proper functor in all cases, isn't it?

But, you have there would be notation for composing functors.  How are you going to check two functors are composable if you don't know the source and target of either of them?

DanGrayson:
> > Also, `hh` is not a functor, for the same reason as above, but also because its values are natural numbers, not objects in a category.
> 
> Sure, but `hh` is just the dimension of certain sheaf cohomology groups, and sheaf cohomology is a functor, so it's just a small abuse of notation.

If it's not a functor, it simply doesn't belong to the class of functors.

mahrud:
> But, you have there would be notation for composing functors. How are you going to check two functors are composable if you don't know the source and target of either of them?

This is not a change from this PR.

> If it's not a functor, it simply doesn't belong to the class of functors.

This is not a change from this PR. Both `hh` and `HH` are still `ScriptedFunctor`. Feel free to correct the mistakes dating back to 1997 after this PR.

DanGrayson:
> > But, you have there would be notation for composing functors. How are you going to check two functors are composable if you don't know the source and target of either of them?
> 
> This is not a change from this PR.

Right, but it's the basis of your justification for introducing the new class called "Functor".

> > If it's not a functor, it simply doesn't belong to the class of functors.
> 
> This is not a change from this PR. Both `hh` and `HH` are still `ScriptedFunctor`. Feel free to correct the mistakes dating back to 1997 after this PR.

But the ancestor of ScriptedFunctor is Functor, so hh and HH are functors, too, in your setup.

mahrud:
> Right, but it's the basis of your justification for introducing the new class called "Functor".

No, it's mostly cleaning up ancient code that was slowing down my computations, and introducing new notation could not be done in a package.

> But the ancestor of ScriptedFunctor is Functor, so hh and HH are functors, too, in your setup.

They are still functors in your setup, aren't they?
> `ScriptedFunctor` -- the class of all scripted functors

I don't know why we're having this back and forth. Given that it's been two months, I can't make major changes to this PR. You can reject and close it if you'd like, but perhaps it would be more productive to accept the changes, then if you want introduce `ScriptedFunction` and change whatever you think is not really a functor to be that instead.

DanGrayson:
Okay, I'll fix up the pull request.

mahrud:
You can push commits to my branch connected to this PR, if you'd like.

DanGrayson:
Okay, I've separate ScriptedFunctor from Functor and renamed it to ScriptedFunction.

mahrud:
> Okay, I've separated ScriptedFunctor from Functor and renamed it to ScriptedFunction.

You did more than that. You also switched back to the inefficient code for HH and Ext ...

DanGrayson:
> > Okay, I've separated ScriptedFunctor from Functor and renamed it to ScriptedFunction.
> 
> You did more than that. You also switched back to the inefficient code for HH and Ext ...

In what sense is it inefficient?

mikestillman:
@mahrud I'm also interested in seeing an example where it is inefficient, to see what the issue really is.

mahrud:
It starts to really slow down when you add up the time it takes to make a few thousand calls to even the identity functor. It's not just a little slower than calling a function or method, it's a lot slower!

I have code for computing certain derived functors, and I need the ability to cache them in a way that behaves well with composing the functors. I don't know why was "gateway.m2" written the way it is (my guess is just so in the documentation things like `HH_i` look good?), but it's not amenable to either caching or composition, and even with caching the slowdown is noticeable. It's just an old idea waiting to be replaced by something better, so I find it unproductive to revert all the changes over something pedantic like that the name includes "Functor" ...

mahrud:
Here is a concrete reason why the code in gateway is inefficient:
```m2
i1 : HH_2 === HH_2

o1 = false
```
Calling this thousands of times liters the memory with new hash tables and functions.
My current PR wouldn't have fixed this yet, but it was the first step towards simplifying things.

DanGrayson:
> Here is a concrete reason why the code in gateway is inefficient:
> 
> ```m2
> i1 : HH_2 === HH_2
> 
> o1 = false
> ```
> 
> Calling this thousands of times liters the memory with new hash tables and functions. My current PR wouldn't have fixed this yet, but it was the first step towards simplifying things.

Yes, but the interpreter allocates lots of memory all the time.  For example, when a function is called, a vector is allocated to hold pointers to the values of all the local variables.

DanGrayson:
What's your ultimate plan for `HH_2`?

mahrud:
> What's your ultimate plan for `HH_2`?

I don't understand your question. My goal right now is to improve a ~25 year old piece of code.

Do you have any improvements planned?

DanGrayson:
> > What's your ultimate plan for `HH_2`?
> 
> I don't understand your question. 

How will you modify the code for `HH` so `HH_2` doesn't allocate a hash table?

mahrud:
> How will you modify the code for `HH` so `HH_2` doesn't allocate a hash table?

I'm not changing that in this PR.

PR #2367, opened 2021-12-31
TBB 2021 support

jkyang92:
This adds support for TBB >=2021. To do this it adds a flag to config.h for TBB >=2021. The only file that really uses TBB is `e/NCAlgebras/NCF4.cpp`, and that only uses `parallel_do` which can be directly replaced by `parallel_for_each` in this case. So instead of implementing a whole shim header, I just added `#ifdef`s where needed.

I'm not entirely happy with the config.h flag solution, but since the file that we need to include depends on the tbb version, I didn't see a better solution.

This needs the mathicgb pull request https://github.com/Macaulay2/mathicgb/pull/30, and has the same caveat that I believe there was other work on TBB support in progress.


M2/Macaulay2/d/version.dd
```diff
@@ -62,14 +62,15 @@ header "
    Givaro::GivaroMain givaroMain;
 
    /* TODO - fix conditionals once TBB 2021 support is added (#2105) */
```
d-torrance, line 64: You can remove this `TODO` comment since you did it.  :)

mahrud:
> This needs the mathicgb pull request [Macaulay2/mathicgb#30](https://github.com/Macaulay2/mathicgb/pull/30), and has the same caveat that I believe there was other work on TBB support in progress.

This would conflict with the work in https://github.com/moorewf/M2-frank/tree/ncEngine.

PR #2380, opened 2022-02-05
Update M2-emacs submodule

d-torrance:
Includes two PRs that replace obsolete functions (see https://github.com/Macaulay2/M2-emacs/pull/27 and https://github.com/Macaulay2/M2-emacs/pull/28).


d-torrance:
I just pushed a new commit to also include the menus from Macaulay2/M2-emacs#29.

PR #2388, opened 2022-02-18
TBB 2021 Support for M2

mikestillman:
TBB 2021 introduced some breaking changes. We use TBB in a few places, with plans to use it more. mathicgb uses it, and that code has already been merged into mathicgb, in the branch fix-tbb. This pull request is done by Frank Moore and Mike Stillman, with useful suggestions from Mahrud Sayrafi, Jay Yang, Doug Torrance, and allows M2 to use this version of TBB, older versions, or not include it at all.

This pull request does the following:

- removes all uses of task_scheduler_init, which is not present in later versions of TBB (e.g. from >= 2021)
- allows use of both tbb2021, and tbb2020 (not sure about older versions).
- allows to not include tbb (via --disable-tbb, -DWITH_TBB=off). I see this mostly as a way to get M2 running on machines that don't have tbb, and as a way to test serial vs parallel code. The shim code is in mathicgb's mtbb.hpp header file.
- fixes many (most) compiler warnings from mathic, mathicgb code, as well as some in the e directory too.
- renames some functions in `e/VectorArithmetic.hpp` (not relevant for TBB2021, as such).
- tests for mathicgb should now use tbb, and they all pass too.

What isn't done yet:

- cmake build: if WITH_TBB=off is given, but tbb is present on the system, then this will attempt to link with it anyway. This apparently then fails sometimes.
- On Apple M1, I am still getting MonomialIdeal::sat crashes sometimes, as with github builds... Got to fix that! (But that is probably not related to this pull request). (Other than this, with BUILD_NATIVE=off, together with homebrew, the build works fine.)


DanGrayson:
A link error here: https://github.com/Macaulay2/M2/runs/5253933843?check_suite_focus=true#step:13:249

mikestillman:
@DanGrayson Could you look at this, especially the autotools build, to make sure we didn't screw anything up?  It appears to pass all tests at least...

mikestillman:
@d-torrance does this work on your ubuntu builds?  @mahrud There is one potential issue remaining.  If a cmake build detects that tbb is present, even with WITH_TBB=off, I think it tries to link with it, but I believe it can fail.

DanGrayson:
I'll take a look.

d-torrance:
> @d-torrance does this work on your ubuntu builds?

I haven't tried building them yet, but I'll let you know!

d-torrance:
@mikestillman - A couple things I've noticed so far when trying to build this branch using a system copy of mathicgb instead of checking out the submodule:
* Like you've mentioned you've been planning on doing, now some of the additional headers are being included (in particular, `mathicgb/mtbb.hpp`).  They're not currently distributed with mathicgb (at least in the autotools build), but I opened https://github.com/Macaulay2/mathicgb/pull/29 a few months ago that does that.  Could that be merged?
* I think it would be good if `#include "mathicgb/mathicgb/mtbb.hpp"` and others like it were changed to `#include "mathicgb/mtbb.hpp"`.  We already have `-I@srcdir@/mathicgb` in `CPPFLAGS`, so the leading directory isn't necessary.  (At least in the autotools build -- maybe that's not the case for the cmake build?  I'm sure it would be an easy fix if it's not.)  And if mathicgb is installed on the system and the headers are in `/usr/include/mathicgb`, then it would also find them there.

(There are a few other fixes that are necessary for building using a system mathicgb, like deleting a bunch of lines in the `e` directory Makefile.  I'm planning to eventually submit a PR to make this configurable.)

mikestillman:
> @mikestillman - A couple things I've noticed so far when trying to build this branch using a system copy of mathicgb instead of checking out the submodule:
> 
> * Like you've mentioned you've been planning on doing, now some of the additional headers are being included (in particular, `mathicgb/mtbb.hpp`).  They're not currently distributed with mathicgb (at least in the autotools build), but I opened [Install all header files (autotools build) mathicgb#29](https://github.com/Macaulay2/mathicgb/pull/29) a few months ago that does that.  Could that be merged?

Done.

> * I think it would be good if `#include "mathicgb/mathicgb/mtbb.hpp"` and others like it were changed to `#include "mathicgb/mtbb.hpp"`.  We already have `-I@srcdir@/mathicgb` in `CPPFLAGS`, so the leading directory isn't necessary.  (At least in the autotools build -- maybe that's not the case for the cmake build?  I'm sure it would be an easy fix if it's not.)  And if mathicgb is installed on the system and the headers are in `/usr/include/mathicgb`, then it would also find them there.
> 

I think there is only one place this occurs (in e/m2tbb.hpp), and I will push this to mikestillman:ncEngine.

> (There are a few other fixes that are necessary for building using a system mathicgb, like deleting a bunch of lines in the `e` directory Makefile. ~I'm planning to eventually submit a PR to make this configurable.~ _EDIT_: see #2394)

In what order should we do these merges?

d-torrance:
> > @mikestillman - A couple things I've noticed so far when trying to build this branch using a system copy of mathicgb instead of checking out the submodule:
> > 
> > * Like you've mentioned you've been planning on doing, now some of the additional headers are being included (in particular, `mathicgb/mtbb.hpp`).  They're not currently distributed with mathicgb (at least in the autotools build), but I opened [Install all header files (autotools build) mathicgb#29](https://github.com/Macaulay2/mathicgb/pull/29) a few months ago that does that.  Could that be merged?
> 
> Done.

Thanks!

> > * I think it would be good if `#include "mathicgb/mathicgb/mtbb.hpp"` and others like it were changed to `#include "mathicgb/mtbb.hpp"`.  We already have `-I@srcdir@/mathicgb` in `CPPFLAGS`, so the leading directory isn't necessary.  (At least in the autotools build -- maybe that's not the case for the cmake build?  I'm sure it would be an easy fix if it's not.)  And if mathicgb is installed on the system and the headers are in `/usr/include/mathicgb`, then it would also find them there.
> 
> I think there is only one place this occurs (in e/m2tbb.hpp), and I will push this to mikestillman:ncEngine.

Great!

> > (There are a few other fixes that are necessary for building using a system mathicgb, like deleting a bunch of lines in the `e` directory Makefile. ~I'm planning to eventually submit a PR to make this configurable.~ _EDIT_: see #2394)
> 
> In what order should we do these merges?

I think this one first.  I've marked the other as a draft and will fix any conflicts after this is merged.

DanGrayson:
What's this error about?  https://github.com/Macaulay2/M2/runs/5335137996?check_suite_focus=true#step:9:14

mikestillman:
> What's this error about? https://github.com/Macaulay2/M2/runs/5335137996?check_suite_focus=true#step:9:14

Hmm, this is a new one for me.  I have not seen that behavior, and I don't think I changed anything related to that, after it ran correctly again a few days ago...  I'll look into it some more...

d-torrance:
> What's this error about? https://github.com/Macaulay2/M2/runs/5335137996?check_suite_focus=true#step:9:14

Looks like the same error as this: https://github.com/Macaulay2/homebrew-tap/pull/131

mahrud:
Try adding `` `brew --prefix boost@1.76` `` here:
https://github.com/Macaulay2/M2/blob/448a45ab372e6942254f52f511a9cb6f7de03518/.github/workflows/test_build.yml#L135

> There is one potential issue remaining. If a cmake build detects that tbb is present, even with WITH_TBB=off, I think it tries to link with it, but I believe it can fail.

Try changing this part:
https://github.com/Macaulay2/M2/blob/db7630570307ec4a19f4ec54cef1420298587312/M2/cmake/check-libraries.cmake#L77-L81

PR #2389, opened 2022-02-20
Add "--with-documentation" option to configure (autotools build)

d-torrance:
This gives some finer control over which types of documentation (html, info, and/or pdf) are built when first calling the `configure` script.  For example `./configure --without-documentation` would build no documentation, and `./configure --with-documentation=pdf` would only build the pdf documentation.

Note that `./configure --with-documentation` or just `./configure` will build the html and info documentation (the current default).

### DIFF
--- a/M2/Macaulay2/packages/Makefile.in
+++ b/M2/Macaulay2/packages/Makefile.in
@@ -12,7 +12,10 @@ PACKAGES = $(ALL_PACKAGES)
 # packages under development that should be checked but not installed:
 DEVEL = EngineTests
 
-MakeDocumentation ?= true
+MakeDocumentation ?= $(if @DOCUMENTATION@,true,false)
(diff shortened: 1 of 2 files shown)


DanGrayson:
It seems that nothing actually looks at the values of MakeInfo, etc., to control whether the documentation is made.

d-torrance:
> It seems that nothing actually looks at the values of MakeInfo, etc., to control whether the documentation is made.

They're included in the `ARGS` Makefile variable:
https://github.com/Macaulay2/M2/blob/71d1aec3f29095f8e35e8ea144b2af64166d1245/M2/Macaulay2/packages/Makefile.in#L27-L30

This is later passed to `installPackage`, which build the documentation:
https://github.com/Macaulay2/M2/blob/71d1aec3f29095f8e35e8ea144b2af64166d1245/M2/Macaulay2/packages/Makefile.in#L77-L83

PR #2390, opened 2022-02-21
CotangentSchubert update

pzinn:
it's all in the title. see also the updated accompanying report
https://www.unimelb-macaulay2.cloud.edu.au/#tutorial-CotangentSchubert


pzinn:
OK, so one test is failing. weird. I'll investigate.

PR #2387, opened 2022-02-18
Add grammar for highlight.js

d-torrance:
[highlight.js](https://highlightjs.org/) is a JavaScript syntax highlighter with language auto-detection and zero dependencies.

With this PR, `make-M2-symbols.m2` generates a `macaulay2.js` file that can be used with highlight.js to syntax highlight Macaulay2 code, such as [this example](https://webwork.piedmont.edu/~dtorrance/highlightjs-macaulay2).

PR #2391, opened 2022-02-21
Add syntax highlighting to html documentation

d-torrance:
This draft is on top of #2387.

We add syntax highlighting to the html documentation using [highlight.js](https://highlightjs.org/).  For example:

image

In particular,

* A new method `M2CODE` is added, which returns a `CODE` object with its `class` attribute set to `language-macaulay2`, which is used to signal highlight.js to syntax highlight its contents.  This is used by default for `Example`, `CannedExample`, `SourceCode`, and `Usage` nodes in the documentation.
* A new `SCRIPT` `Hypertext` object is added for including highlight.js in each html page.
* `npm` is called during build (both cmake and autotools) and uses [webpack](https://webpack.js.org/) to generate a bundled copy of highlight.js including only the Macaulay2 language and the default syntax highlighting theme (minus the background color).  If it's not available, then highlight.js will be an empty file and no syntax highlighting will be performed.  (Caveat: `npm` will try to download various node packages, so internet access would be required during build for this to work.  This probably isn't any issue for most folks except possibly for downstream package maintainers like myself, but there's already a standard workaround for this in Debian.)

Any thoughts?  Note that there are lots of [other themes](https://highlightjs.org/static/demo/) besides the default that's currently used.  Maybe this could even be a build option?


d-torrance:
Since #2387 has been merged, I'll go ahead and remove the draft status for this PR.

If anyone wants to see it in action, I've pushed the docs (`Macaulay2Doc` only) to the [Github Pages branch of my fork](https://www.friedcheese.org/M2/doc/Macaulay2/Macaulay2Doc/html/).

d-torrance:
> any reason to use highlightjs rather than prism, for which we had Macaulay2 syntax highlighting?

Not really, other than I just implemented Macaulay2 syntax highlighting in highlightjs right before this PR (in #2387), initially for a completely different reason than using it in the Macaulay2 documentation.  The idea of using it for the docs came later, but it was fresh in my mind.

pzinn:
okay. it's just a bit silly that we now have two competing types of syntax highlighting for the documentation (see e.g. https://www.unimelb-macaulay2.cloud.edu.au/usr/share/doc/Macaulay2/Macaulay2Doc/html/_substring.html for prism.js)

d-torrance:
> okay. it's just a bit silly that we now have two competing types of syntax highlighting for the documentation (see e.g. https://www.unimelb-macaulay2.cloud.edu.au/usr/share/doc/Macaulay2/Macaulay2Doc/html/_substring.html for prism.js)

Ah, indeed.  I didn't realize that.  :\

PR #2392, opened 2022-02-23
A few html-check-links fixes

d-torrance:
So it turns out that I broke `make -C html-check-links` last year in #1875...  By switching from using `xargs` to `find`'s `-exec` option, we ended up not checking *any* of the html documentation for broken links.  Oops!

An alternate solution to the original problem (when `find` doesn't find anything, e.g., when not building the documentation,`ls` gets empty input, and so we end up checking all the non-html files in the directory for broken links) is to go back to using `xargs` but pass its `-r` option (or `--no-run-if-empty`), which means that we own't even bother running `ls` if `find` doesn't find anything.  (Note that this is a GNU extension to `xargs`, but this is already the default behavior for the BSD `find` and `-r` exists as a noop for compatibility with the GNU version.  So I believe this change should be fine on macOS.)

After getting the check working again, I found two failures in the documentation (author homepages that were missing `https://`'s), which I fixed.  Note that one of these (David Eisenbud's) gives a 404 error: https://www.msri.org/~de.  It appears 31 times in various package files, so it should probably be updated.

### DIFF
--- a/M2/Macaulay2/Makefile.in
+++ b/M2/Macaulay2/Makefile.in
@@ -3,7 +3,8 @@
 VPATH = @srcdir@
 include ../include/config.Makefile
 all: @pre_docdir@/COPYING-GPL-2 @pre_docdir@/COPYING-GPL-3 @pre_docdir@/README @pre_docdir@/LAYOUT
-SUBDIRS = util c d e system bin m2 man packages editors html-check-links tests
+SUBDIRS = util c d e system bin m2 man packages editors \
+	$(if $(filter html, @DOCUMENTATION@), html-check-links) tests
 
 define do-in-subdirs
 $(foreach d,$(SUBDIRS),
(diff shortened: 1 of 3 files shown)


d-torrance:
Hmm, the Github builds are failing...

```
cd /home/runner/work/M2/M2/M2/BUILD/build/usr-dist/common/share/doc/Macaulay2 && find . -name \*.html -o -name \*.xhtml | xargs -r ls -rt | LD_LIBRARY_PATH="/home/runner/work/M2/M2/M2/BUILD/build/usr-host/lib:$LD_LIBRARY_PATH" xargs -r "/home/runner/work/M2/M2/M2/BUILD/build/Macaulay2/html-check-links"/html-check-links --no-absolute-links
xargs: /home/runner/work/M2/M2/M2/BUILD/build/Macaulay2/html-check-links/html-check-links: terminated by signal 11
ls: write error: Broken pipe
ls: write error: Broken pipe
make[2]: *** [Makefile:62: run-html-check-links] Error 125
```

Not sure what's going on, as it worked locally.  Turning into a draft while I investigate.

d-torrance:
I can't figure out why the GitHub builds are failing, so I opened #2398.  There are still a couple useful things in this PR, so I'll go ahead and mark it ready for review.  In particular:

* Fix the two broken links.
* Skip `html-check-links` entirely when not building the html docs.

PR #2394, opened 2022-02-25
Autotools build update for mathicgb and friends

d-torrance:
This is a draft pending #2388, as there may be some conflicts.

I've adapted the [patch](https://salsa.debian.org/math-team/macaulay2/-/blob/debian/master/debian/patches/use-shared-mathicgb-and-friends.patch) I'm currently using for the Debian package to work upstream.  In particular:

* By default, instead of compiling mathicgb and friends directly in the `e` directory, we build static libraries and link against those.  This is consistent with what the cmake build does.
* There are options to `configure` (`--disable-build-memtailor`, `--disable-build-mathic`, and `--disable-build-mathicgb`) to instead use the libraries already available on the system.  My plan is to use these in the Debian package.
* To preserve the current behavior of always checking out the mathicgb and friends submodules, we don't require `--enable-download` to be passed to `configure` for them like we do for other submodules that might be listed in `BUILDSUBLIST`.


d-torrance:
Oops:

```
In file included from /Users/runner/work/M2/M2/M2/BUILD/build/../../submodules/mathic/src/mathic.cpp:1:
In file included from /Users/runner/work/M2/M2/M2/BUILD/build/../../submodules/mathic/src/mathic.h:10:
/Users/runner/work/M2/M2/M2/BUILD/build/../../submodules/mathic/src/mathic/HashTable.h:19:10: fatal error: 'memtailor.h' file not found
#include <memtailor.h>
         ^~~~~~~~~~~~~
1 error generated.
make[1]: *** [Makefile:1072: src/libmathic_la-mathic.lo] Error 1
```
I'll figure out a fix soon.

d-torrance:
This is strange.  From https://github.com/Macaulay2/M2/runs/5338062998?check_suite_focus=true:
```
clang: error: unsupported option '-fopenmp'
```
Why are we getting a clang error during the gcc build?

DanGrayson:
> Why are we getting a clang error during the gcc build?

On the mac, gcc is clang:

```sh
gallium$ gcc -v
Configured with: --prefix=/Library/Developer/CommandLineTools/usr --with-gxx-include-dir=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/4.2.1
Apple clang version 13.0.0 (clang-1300.0.29.30)
Target: x86_64-apple-darwin21.2.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
gallium$ type gcc
gcc is hashed (/usr/bin/gcc)
```

d-torrance:
> On the mac, gcc is clang:

Oh, ok!

d-torrance:
Not sure why we were getting that error, but it was only happening when linking the `mgb` command-line tool, so I just stopped building that.

d-torrance:
I've rebased this onto `development` now that #2388 has been merged.

d-torrance:
Note also https://github.com/Macaulay2/mathicgb/pull/32, which would simplify things a bit.

d-torrance:
Oops:

```
In file included from ../../../../Macaulay2/e/NCAlgebras/NCF4.hpp:4,
                 from ../../../../Macaulay2/e/interface/groebner.cpp:24:
../../../../Macaulay2/e/m2tbb.hpp:7:10: fatal error: mathicgb/mtbb.hpp: No such file or directory
    7 | #include "mathicgb/mtbb.hpp"
      |          ^~~~~~~~~~~~~~~~~~~
compilation terminated.
```

d-torrance:
I made a mistake in https://github.com/Macaulay2/mathicgb/pull/29 where in out-of-source builds, the headers wouldn't be found.  It's corrected in https://github.com/Macaulay2/mathicgb/pull/33.

d-torrance:
> The 2nd build failed with "monideal.cpp:787".

That's #1742.  I don't think it's related to this pull request.

Issue #2399, opened 2022-02-28
highlightjs directory missing


DanGrayson:
```sh
gmake -C ../editors highlightjs/highlight.js
gmake[4]: Entering directory '/Users/dan/src/M2/M2.git/M2/BUILD/dan/builds.tmp/gallium-development/Macaulay2/editors'
warning: html docs will not have syntax highlighting
touch highlightjs/highlight.js
touch: highlightjs/highlight.js: No such file or directory
gmake[4]: *** [Makefile:25: highlightjs/highlight.js] Error 1
gmake[4]: Leaving directory '/Users/dan/src/M2/M2.git/M2/BUILD/dan/builds.tmp/gallium-development/Macaulay2/editors'
```

@d-torrance

d-torrance:
Oh, whoops.  I must have just tested the no-npm case with that directory already existing.  Should be an easy fix -- I'll push a PR shortly.

PR #2400, opened 2022-02-28
Create highlightjs directory before empty highlight.js file

d-torrance:
This prevents a "No such file or directory" error in the autotools build when building without npm available.

Closes: #2399


d-torrance:
I also took the opportunity to rewrite the highlight.js README a bit.  In particular, the "Usage" section was moved up a bit, and a few things were rewritten for clarification.

PR #2401, opened 2022-03-02
New package OnlineLookup

pzinn:
- the new package OnlineLookup for now only supports OEIS lookups:
```
i1 : needsPackage "OnlineLookup"

o1 = OnlineLookup

o1 : Package

i2 : oeis {1,7,11,17,31}

o2 = {0 => (A068674 (see https://oeis.org/A068674 ) Numbers which yield primes when a 3 is prefixed or appended.)               }
     {1 => (A158594 (see https://oeis.org/A158594 ) Numbers which yield a prime whenever a 3 is prefixed, appended or inserted.)}

o2 : NumberedVerticalList
```
--> This package is not preloaded by default, because I'm not sure how to do this, but I would like it to be!
- various minor improvements to `tex` output.
In particular, noone uses "\special" in the 21st century (it doesn't work with pdf(la)tex), so I replaced it with "\href" from the standard hyperref package.

### DIFF
--- a/M2/Macaulay2/m2/expressions.m2
+++ b/M2/Macaulay2/m2/expressions.m2
@@ -335,6 +335,7 @@ RowExpression = new HeaderType of Expression
 RowExpression.synonym = "row expression"
 net RowExpression := w -> horizontalJoin apply(toList w,net)
 html RowExpression := w -> concatenate apply(w,html)
+tex RowExpression := w -> concatenate apply(w,tex)
 texMath RowExpression := w -> concatenate apply(w,texMath)
 toString'(Function, RowExpression) := (fmt,w) -> concatenate apply(w,fmt)
 -----------------------------------------------------------------------------
--- a/M2/Macaulay2/m2/installedpackages
+++ b/M2/Macaulay2/m2/installedpackages
@@ -10,3 +10,4 @@ ReesAlgebra
 ConwayPolynomials
 InverseSystems
 SimpleDoc
+OnlineLookup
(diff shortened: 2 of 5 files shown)


DanGrayson:
It looks like you didn't add it to the list of pre-loaded packages in the file `m2/installedpackages`.  Based on our previous discussion, I think you intended to do that.

mahrud:
Why not put it in `SimpleDoc` along with `arXiv`, `wikipedia`, and `stacksProject`? That package is preloaded as well.

DanGrayson:
> Why not put it in `SimpleDoc` along with `arXiv`, `wikipedia`, and `stacksProject`? That package is preloaded as well.

It's unrelated to documentation.

pzinn:
> Why not put it in `SimpleDoc` along with `arXiv`, `wikipedia`, and `stacksProject`? That package is preloaded as well.

I had made the same comment in M2internals. Inversely, why exactly are `arXiv`, `wikipedia`, and `stacksProject` in `SimpleDoc`? they're not related to documentation either.

pzinn:
> It looks like you didn't add it to the list of pre-loaded packages in the file `m2/installedpackages`. Based on our previous discussion, I think you intended to do that.

done

DanGrayson:
> > Why not put it in `SimpleDoc` along with `arXiv`, `wikipedia`, and `stacksProject`? That package is preloaded as well.
> 
> I had made the same comment in M2internals. Inversely, why exactly are `arXiv`, `wikipedia`, and `stacksProject` in `SimpleDoc`? they're not related to documentation either.

They are related to documentation -- they make hypertext HREF items that can be used as links inside docuementation.

Issue #2405, opened 2022-03-07
equality of modules over inexact rings


mikestillman:
Equality of modules `M == N` should perhaps first check if `M === N`?  Currently, the code immediately goes to Grobner type computations, which gives different answers over inexact fields due to rounding errors.

```
R = RR_53
M = coker matrix{{1.3, 1.4}, {1.1, 1.5}, {.3, .6}}
assert(M === M) -- this is good
assert(not(M == M)) -- M===M but M != M...
```

DanGrayson:
Good idea.  And also fix this, if possible:
```m2
Macaulay2, version 1.19.1.1
with packages: ConwayPolynomials, Elimination, IntegralClosure, InverseSystems, LLLBases, MinimalPrimes, PrimaryDecomposition, ReesAlgebra, Saturation, TangentCone

i1 : f = matrix{{1.3, 1.4}, {1.1, 1.5}, {.3, .6}};

                3          2
o1 : Matrix RR    <--- RR
              53         53

i2 : isSubset (image f, image f)
-- warning: experimental computation over inexact field begun
--          results not reliable (one warning given per session)

o2 = false
```

mahrud:
This is also tangentially related to #1726. In case anyone fixes this, fixing that (and therefore also #2373) would also be nice.

PR #2406, opened 2022-03-08
fix module == comparison

DanGrayson:
resolves issue #2405


mahrud:
Could you revert this and simply add `M === N or ` on the first line? Like this:
```diff
<<<
Module == Module := (M,N) -> (
===
>>>
```

mahrud:
I believe there is potentially a mistake in this PR that may have gone unnoticed because of the excessive diffs. I made a new PR to fix it.

PR #2407, opened 2022-03-09
better fix for module == comparison

mahrud:
Reverts Macaulay2/M2#2406 and applies simpler fix for it.

Also I removed the `N.relations == 0` that was added here:
https://github.com/Macaulay2/M2/blob/0b0159476fbb752b5c0bfd9a50a50b41c0cf95a5/M2/Macaulay2/m2/modules.m2#L251

This is either superfluous or wrong, and here is why: suppose it is necessary to check the case when M doesn't have relations but N does and they are just 0, then the result of `M == N` and `N == M` would be different, because then this part would return false:
https://github.com/Macaulay2/M2/blob/0b0159476fbb752b5c0bfd9a50a50b41c0cf95a5/M2/Macaulay2/m2/modules.m2#L239-L240

If it is necessary then there should be a check for both M and N, and there should be a test for it.

### DIFF
--- a/M2/Macaulay2/m2/modules.m2
+++ b/M2/Macaulay2/m2/modules.m2
@@ -230,7 +230,7 @@ describe Module := M -> Describe (
      )
 toExternalString Module := M -> toString describe M
 
-Module == Module := (M,N) -> (
+Module == Module := (M,N) -> M === N or (
      -- this code might not be the quickest - Mike should check it
      ring M === ring N
      and degrees ambient M === degrees ambient N
@@ -242,6 +242,8 @@ Module == Module := (M,N) -> (
 	       -- else 
 		    (
 		    -- temporary
+		    M.relations === N.relations
+		    or
 		    isSubset(image M.relations, image N.relations)
 		    and
 		    isSubset(image N.relations, image M.relations)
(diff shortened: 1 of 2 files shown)


mahrud:
@DanGrayson the diff looks like a lot, but most of it is in reverting your PR. This is the simplified diff:
```diff
--- a/M2/Macaulay2/m2/modules.m2
+++ b/M2/Macaulay2/m2/modules.m2
@@ -230,7 +230,7 @@ describe Module := M -> Describe (
      )
 toExternalString Module := M -> toString describe M
 
-Module == Module := (M,N) -> (
+Module == Module := (M,N) -> M === N or (
      -- this code might not be the quickest - Mike should check it
      ring M === ring N
      and degrees ambient M === degrees ambient N
@@ -242,6 +242,8 @@ Module == Module := (M,N) -> (
               -- else 
                    (
                    -- temporary
+                   M.relations === N.relations
+                   or
                    isSubset(image M.relations, image N.relations)
                    and
                    isSubset(image N.relations, image M.relations)
```

mahrud:
Actually, it would be great if you could rebase and yank out both your commit b832d5a28d6da49057540d59c175e0bf8ea5603b and my revert commit f3100fd so that `git blame` looks nice. This would require a force push to development.

Issue #2408, opened 2022-03-09
OnlineLookup example requires network access


d-torrance:
Network access may not be available during build.  For example, from [this PPA build](https://launchpadlibrarian.net/589863537/buildlog_ubuntu-impish-amd64.macaulay2_1.19.1.1+git202203090010-0ppa202203090110~ubuntu21.10.1_BUILDING.txt.gz):

```m2
 -- making example results for "oeis"                                       
 ulimit -c unlimited; ulimit -t 700; ulimit -m 850000; ulimit -s 8192; ulimit -n 512;  cd /tmp/M2-124904-0/1-rundir/; GC_MAXIMUM_HEAP_SIZE=400M "/<<PKGBUILDDIR>>/M2/usr-dist/x86_64-Linux-Ubuntu-21.10/bin/M2-binary" -q --int --no-randomize --no-readline --silent --stop --print-width 77 -e 'needsPackage("OnlineLookup",Reload=>true,FileName=>"/<<PKGBUILDDIR>>/M2/Macaulay2/packages/OnlineLookup.m2")' <"/tmp/M2-124904-0/0_oeis.m2" >>"/<<PKGBUILDDIR>>/M2/usr-dist/common/share/doc/Macaulay2/OnlineLookup/example-output/_oeis.errors" 2>&1
/<<PKGBUILDDIR>>/M2/usr-dist/common/share/doc/Macaulay2/OnlineLookup/example-output/_oeis.errors:0:1: (output file) error: Macaulay2 exited with status code 1
/tmp/M2-124904-0/0_oeis.m2:0:1: (input file)
M2: *** Error 1
 -- 0.255548 seconds elapsed
 -- storing example results in ../../../usr-dist/common/share/doc/Macaulay2/OnlineLookup/example-output/_oeis.out
 -- warning: missing file /<<PKGBUILDDIR>>/M2/usr-dist/common/share/doc/Macaulay2/OnlineLookup/example-output/_oeis.out
../../m2/debugging.m2:23:6:(1):[9]: error: installPackage: 1 error(s) occurred running examples for package OnlineLookup:

_oeis.errors
************
-- -*- M2-comint -*- hash: 1103500785

i1 : oeis {1,3,31,1145}
stdio:1:1:(3): error: can't open socket : Name or service not known
```

Perhaps this example should be cached?

pzinn:
.. but https://github.com/Macaulay2/M2/issues/1924 ...

DanGrayson:
[...]

PR #2415, opened 2022-03-15
OnlineLookup update: isc, fix #2414 and #2408

### LINKED ISSUE #2414: `getWWW` http request deemed invalid by some servers
some servers require the proper \r\n for CRLF, cf
https://stackoverflow.com/questions/5757290/http-header-line-break-style
M2 `endl`, on the other hand, only sends \n.
the result is that `getWWW` (in http.m2) requests are invalid and will be rejected by some servers.
random example: try
```
getWWW "http://wayback.cecm.sfu.ca"
```

pzinn:
- `OnlineLookup` now allows to query the Inverse Symbolic Calculator. This required fixing #2414.
- `OnlineLookup` now gracefully warns user when no internet is available, which should solve #2408.
also, examples are now cached.

### DIFF
--- a/M2/Macaulay2/m2/http.m2
+++ b/M2/Macaulay2/m2/http.m2
@@ -21,14 +21,14 @@ GET := (host,url,connection) -> (
 POST := (host,url,body,connection) -> (
      connection = 
      openInOut connection
-     << "POST " << url << " HTTP/1.1" << endl
-     << "User-Agent: " << httpProduct << endl
-     << "Connection: close" << endl
-     << "Host: " << host << endl
-     << "Content-type: application/x-www-form-urlencoded" << endl
-     << "Content-length: " << # body << endl << endl
-     << body << endl
-     << endl << flush;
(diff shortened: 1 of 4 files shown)


M2/Macaulay2/m2/http.m2
```diff
@@ -5,14 +5,16 @@ needs "methods.m2"
...
-     << "Connection: close" << endl
-     << "Host: " << host << endl
-     << endl << flush;
+     << "GET " << url << " HTTP/1.1" << crlf
+     << "User-Agent: " << httpProduct << crlf
+     << "Connection: close" << crlf
+     << "Host: " << host << crlf
+     << crlf << flush;
      first( get connection, close connection )
      )
```
DanGrayson, line 20: Shouldn't the uses of endl below also be replaced by crlf?

pzinn: I'm assuming you mean in POST. fixed.

Issue #2417, opened 2022-03-16
"class template argument deduction failed" compiling NCF4.cpp in Ubuntu 18.04


d-torrance:
I'm getting the following compiling the `development` branch in Ubuntu 18.04 (default GCC version is 7.5.0):

```
g++ -pipe -Wall -Wno-shadow -Wcast-qual -Wno-sign-conversion -Wno-sign-compare -Wno-parentheses -Wno-sign-compare  -Wuninitialized -I/usr/include  -Wno-error=pragmas -std=gnu++14 -g3 -O2  -Wno-unused-local-typedefs -Wno-unknown-pragmas -std=gnu++17 -Wno-mismatched-tags -I../../../../Macaulay2/e -DNDEBUG -I/root/M2/M2/BUILD/doug/../../include -I/root/M2/M2/BUILD/doug/include -I/root/M2/M2/BUILD/doug/usr-host/include -isystem /usr/include/libxml2   -DSING_NDEBUG -DOM_NDEBUG -DSING_NDEBUG -DOM_NDEBUG -I/usr/include/x86_64-linux-gnu/singular -I/usr/include/singular -I/usr/include/x86_64-linux-gnu/singular -I/usr/include/singular -I/usr/include/x86_64-linux-gnu/singular -I/usr/include/singular   -I/usr/src/gtest   -DNDEBUG  -I/usr/include/cdd -I/usr/include/eigen3 -I/usr/include -DBOOST_STACKTRACE_LINK -I../d -I../../../../Macaulay2/e/../c -I../../../../Macaulay2/e/../d -Wno-fatal-errors -Wno-unused-variable  -c -Wno-cast-qual ../../../../Macaulay2/e/NCAlgebras/NCF4.cpp -o NCAlgebras/NCF4.o
In file included from /usr/include/mathicgb/mtbb.hpp:21:0,
                 from ../../../../Macaulay2/e/m2tbb.hpp:7,
                 from ../../../../Macaulay2/e/NCAlgebras/NCF4.hpp:4,
                 from ../../../../Macaulay2/e/NCAlgebras/NCF4.cpp:1:
/usr/include/tbb/tbb.h:21:154: note: #pragma message: TBB Warning: tbb.h contains deprecated functionality. For details, please see Deprecated Features appendix in the TBB reference manual.
 details, please see Deprecated Features appendix in the TBB reference manual.")
                                                                               ^
In file included from ../../../../Macaulay2/e/aring-translate.hpp:18:0,
[...]

PR #2074, opened 2021-05-13
Use list initialization for class template argument deduction

d-torrance:
Workaround for GCC bug 85883, which wasn't fixed until GCC 9.
However, older versions of GCC are still commonplace (e.g., the
default in Ubuntu 18.04 is GCC 7.)

Closes: #2073

Issue #2073, opened 2021-05-13
error: no matching function for call to ‘ConcreteVectorArithmetic()’


d-torrance:
I've seen the following on a number of PPA builds:

* [Ubuntu 18.04, arm64](https://launchpadlibrarian.net/538549516/buildlog_ubuntu-bionic-arm64.macaulay2_1.17.2.1+git202105131229-0ppa202105121825~ubuntu18.04.1_BUILDING.txt.gz)
* [Ubuntu 18.04, i386](https://launchpadlibrarian.net/538549092/buildlog_ubuntu-bionic-i386.macaulay2_1.17.2.1+git202105131229-0ppa202105121825~ubuntu18.04.1_BUILDING.txt.gz)
* [Ubuntu 18.04, ppc64el](https://launchpadlibrarian.net/538549331/buildlog_ubuntu-bionic-ppc64el.macaulay2_1.17.2.1+git202105131229-0ppa202105121825~ubuntu18.04.1_BUILDING.txt.gz)

```
[...]

PR #2418, opened 2022-03-17
Use list initialization for class template argument deduction

d-torrance:
For GCC 7 support.  Closes: #2417

---

I realized that this was the same issue as #2073, which had a very trivial fix (`()` -> `{}` -- see #2074).


mikestillman:
This is better "modern c++" than what was there, in any case.  I'll accept this now.

Issue #2375, opened 2022-01-19
Resurrect FilePosition


pzinn:
There's an unused m2 type `FilePosition` which I think would be useful for debugging purposes. It seems similar to the d type Position except for the fact that it's missing the `loadDepth` (but that can be changed). It would be great if it could be used -- but for that one'd need a new m2 function that gives the current Position in the file (stdio or actual file) being inputted. it could clean up a lot of the formatting code in code.m2, debugging.m2 and nets.m2.
more specifically: 
- one would need first a d function `currentColumnNumber` since there is only `currentLineNumber` currently implemented (digression: it's very easy to confuse `currrentLineNumber()` and `lineNumber` even though they're quite distinct. maybe `currentLineNumber` could be renamed `currentRowNumber`?)
- one should decide whether to add `loadDepth` or not to `FilePosition`
- finally, one can define (purely at the m2 level) a function `currentPosition` that combines the info above.

pzinn:
If there are no comments, I will implement this (including the renaming to `currentRowNumber`, which will be easy since it's only used in a couple of places).

DanGrayson:
What would it be used for?

pzinn:
well, the real motivation for me to introduce `currentColumnNumber` is I need it for proper error highlighting in the WebApp interface (and I'd rather code this in m2 rather than d). Adding `currentPosition` is only an extra line of code then, combining the various existing `current*`. Longer term, I think `FilePosition` can definitely be used in code.m2, debugging.m2 and nets.m2, but I haven't started looking into it yet.

DanGrayson:
Okay, sounds good.

PR #2421, opened 2022-03-21
implement #2375 plus various minor fixes

pzinn:
this PR
- implements #2375 
- uses it immediately in webapp.m2 for proper error highlighting
- webapp.m2 got broken by the recent changes to documentation, so that is fixed too
- various minor `texMath` and `html` fixes that I've been accumulating over the past few months


mahrud:
This is nice! I wonder if it'll make profiling (e.g. with flamegraphs) easier.

PR #2422, opened 2022-03-24
Updating RationalMaps

kschwede:
This is a rather substantial update to RationalMaps.m2.
This also impacts SpecialFanoFourfolds.m2 and Divisor.m2 which relied on RationalMaps.

There is also a fix of a rare bug in FastMinors.m2


kschwede:
Hi all, 
I'm not sure why that check failed.  It seemed to fail on MultiGradedRationalMap which conceivably could be related.  However, that doesn't import any package I changed or any package that does, as far as I can see.  Am I missing something?
Karl

d-torrance:
That segfault is a well-known bug (#1742) and likely completely unrelated to your changes.

mikestillman:
This is not your bug.  It is my bug which I am still trying to track down!  On the plus side, on my Apple M1, I have obtained the crash locally.

PR #2424, opened 2022-03-28
Fix MonomialIdeal::sat crash, see issue #1742

mikestillman:
This pull request (hopefully) fixes the MonomialIdeal::sat crashes we have been seeing for quite some time now.  See issue #1742.

The bug is somewhat subtle.  In the function `MonomialIdeal::sat`, a `MonomialIdeal` is passed in as a pointer (reference).  An iterator traverses the data structure.  While the iteration was going on, every now and then, the garbage collector would run, and would seemingly erroneously free the passed in MonomialIdeal.  This left the iterators in a bad state, causing the crash.

The problem is, I believe, that there were no pointers to this MonomialIdeal object anywhere on the stack: at the front end, in `Saturation.m2`, this function was being called via `rawSaturate(raw monomialIdeal I, raw monomialIdeal J)`, which does not keep a pointer.  It seems the optimizer was removing all of the uses of this object on the stack as well.

When it was collected (during the loop in `MonomialIdeal::sat`), it sometimes freed the entire data structure, so the iterators don't point anywhere good anymore.  How could this be?  Well, the iterator itself doesn't need the container (the MonomialIdeal) to traverse the data structure.  Even though the iterator is a struct, with one member being the pointer to the MonomialIdeal, the optimizer is eliding this variable (i.e. destructing the struct)!  And then the object was being collected prematurely.  And. even though it seemed to be on the call stack, it probably was in a register, and then getting reused.

The fix we decided on (@DanGrayson  and I discussed this): call a new function (in `finalize.{hpp,cpp}`) called `keepAlive` at the end of the function `MonomialIdeal::sat` as in `keepAlive(&J)`.  This prevents the optimizer from removing this function early.

To see the bug in action (every time, not just intermittently), place `GC_gcollect()` inside the loop as 
the last line, and remove the `keepAlive` call.  Then run the following lines of code.  The last line will (as far as I can tell) always crash.  Placing the keepAlive call back in, even with the `GC_gcollect`, keeps it from crashing.
```
    R = ZZ/101[x,y,u,v]
    I = ideal(x*u, y*u, y*v)
    NN = ideal (x*u, x*v, y*u, y*v)
    saturate(I, NN)
```

### DIFF
--- a/M2/Macaulay2/e/finalize.cpp
+++ b/M2/Macaulay2/e/finalize.cpp
@@ -3,6 +3,12 @@
 #include "finalize.hpp"
 #include "engine-includes.hpp"
 
+void keepAlive(const void*)
+{
+  // do nothing.  This is just present to make sure the argument is
+  // not garbage collected earlier than it should be.
+}
+
 #include <atomic_ops.h>
 // AO_fetch_and_add1 is not available on some architectures (e.g., hppa)
 #ifndef AO_HAVE_fetch_and_add1
--- a/M2/Macaulay2/e/finalize.hpp
+++ b/M2/Macaulay2/e/finalize.hpp
@@ -7,6 +7,31 @@ class GBComputation;
 class ResolutionComputation;
 class SchreyerOrder;
 
+//* keepAlive: call at the end of a function when using a
+//* finialized object.  More specifically: if you are using an
+//* iterator into a data structure, or any pointers into such, which
+//* are pointed to by the finalized object, call this at the end of
+//* the function in question.
+//*
+//* The reason for this is that the optimizer can potentially elide
+//* the finalized object and the GC finalizer might get called
+//* otherwise on this object.  The finalizer function might then free
+//* the data you are using from underneath you.
+//* 
+//* This happened for instance in the MonomialIdeal class, in
+//* e.g. MonomialIdeal::sat.  If a pointer to the MonomialIdeal is not
+//* being kept in the frontend (as a RawMonomialIdeal) then it was
+//* happening that the optimizer was optimizing out the local
+//* existence of the pointer to this object.  This in turn caused the
+//* destructor of the object to be called, which then left the
+//* iterator traversing the data structure in a bad state (in fact,
+//* all nodes of the data structure were deleted by the destructor).
+//* 
+//* Adding in e.g. `nopAndKeepAlive(&J)` at the end of
+//* MonomialIdeal::sat allows this to not occur.
+
+void keepAlive(const void*);
+
 // These functions should be called if G will not be freed by its owner
 void intern_polyring(const PolynomialRing *G);
 void intern_monideal(MonomialIdeal *G);
--- a/M2/Macaulay2/e/monideal.cpp
+++ b/M2/Macaulay2/e/monideal.cpp
@@ -3,6 +3,7 @@
 #include "monideal.hpp"
 #include "monoid.hpp"
 #include "text-io.hpp"
+#include "finalize.hpp" // for keepAlive
 
 unsigned int MonomialIdeal::computeHashValue() const
 {
@@ -174,6 +175,7 @@ bool MonomialIdeal::is_equal(const MonomialIdeal &mi0) const
       i++;
       j++;
     }
+  keepAlive(&mi0);
   return true;
 }
 
@@ -619,6 +621,7 @@ MonomialIdeal *MonomialIdeal::intersect(const MonomialIdeal &J) const
           }
     }
   MonomialIdeal *result = new MonomialIdeal(get_ring(), new_elems);
+  keepAlive(&J);
   return result;
 }
 
@@ -649,6 +652,7 @@ MonomialIdeal *MonomialIdeal::operator*(const MonomialIdeal &J) const
         new_elems.insert(b);
       }
   MonomialIdeal *result = new MonomialIdeal(get_ring(), new_elems);
+  keepAlive(&J);
   return result;
 }
 
@@ -666,6 +670,7 @@ MonomialIdeal *MonomialIdeal::operator+(const MonomialIdeal &J) const
       new_elems.insert(b);
     }
   MonomialIdeal *result = new MonomialIdeal(get_ring(), new_elems);
+  keepAlive(&J);
   return result;
 }
 
@@ -683,6 +688,7 @@ MonomialIdeal *MonomialIdeal::operator-(const MonomialIdeal &J) const
           result->insert_minimal(b);
         }
     }
+  keepAlive(&J);
   return result;
 }
 
@@ -714,6 +720,7 @@ MonomialIdeal *MonomialIdeal::quotient(const MonomialIdeal &J) const
       delete result;
       result = next_result;
     }
+  keepAlive(&J);
   return result;
 }
 
@@ -792,6 +799,7 @@ MonomialIdeal *MonomialIdeal::sat(const MonomialIdeal &J) const
       delete result;
       result = next_result;
     }
+  keepAlive(&J);
   return result;
 }
 


DanGrayson:
This will be very welcome!

DanGrayson:
Congratulations!!

mikestillman:
@DanGrayson It would be an interesting experiment if you could try making the two changes I suggest to MonomialIdeal::sat on your machine (add in `GC_gcollect()`, remove the new `keepAlive(&J)`), and see if the 4 lines of M2 code above crash for you.  Since it is only on some machines that we see it, perhaps this optimization isn't done on some architectures.  On the other hand, it was crashing on github architectures...

mikestillman:
@DanGrayson Well, at least all the tests passed!

d-torrance:
I still got one segfault in last night's PPA builds after applying this patch, on [amd64 in Ubuntu 21.10](https://launchpadlibrarian.net/593647519/buildlog_ubuntu-impish-amd64.macaulay2_1.19.1.1+git202203241707-0ppa202203282012~ubuntu21.10.1_BUILDING.txt.gz):

```m2
../../m2/debugging.m2:23:6:(1):[9]: error: installPackage: 1 error(s) occurred running examples for package SimplicialDecomposability:

_is__Shedding__Vertex.errors
****************************
-- -*- M2-comint -*- hash: -2081083635

i1 : R = QQ[a..f];

i2 : S = simplicialComplex {a*b*c, c*d, d*e, e*f, d*f};

i3 : isSheddingVertex(a, S)
-- SIGSEGV
-* stack trace, pid: 58009
 0# stack_trace(std::ostream&, bool) at ../d/main.cpp:127
 1# segv_handler at ../d/main.cpp:244
 2# 0x00007F857846C520 in /lib/x86_64-linux-gnu/libc.so.6
 3# interface_rawColon at ../d/interface.dd:331
 4# evaluate_evalraw at ../d/evaluate.d:1303
```

mikestillman:
@d-torrance @DanGrayson The crash on amd64 appears to be from interface.dd.  If the stack trace is correct, it isn't even going into the engine routine?  For the pull request, I had fixed the internal engine routine MonomialIdeal::quotient(MonomialIdeal& J) in the same way as MonomialIdeal::sat.

d-torrance:
I've seen this example segfault before and the stack trace did go up to the engine code.  So maybe it's still related somehow? See https://github.com/Macaulay2/M2/issues/1742#issuecomment-876079065.

mikestillman:
Hmm, I fixed what I thought was the issue in that function.  But maybe this is somehow related.  Please keep recording all crashes you find!

DanGrayson:
I don't see a way for this crash in interface.dd to be related.

d-torrance:
I tried running the `isSheddingVertex` example code over and over again in a while loop and eventually (after 291740 iterations!) got a segfault with the stack trace going up to the engine code:

```m2
-- SIGSEGV
-* stack trace, pid: 4784
 0# stack_trace(std::ostream&, bool) at ../d/main.cpp:127
 1# segv_handler at ../d/main.cpp:244
 2# 0x00007FC84E84E520 in /usr/lib/x86_64-linux-gnu/libc.so.6
 3# varpower::quotient(int const*, int const*, intarray&) at ../e/varpower.cpp:258
 4# MonomialIdeal::quotient(int const*) const at ../e/monideal.cpp:697
 5# interface_rawColon at ../d/interface.dd:331
 6# evaluate_evalraw at ../d/evaluate.d:1303
```

mikestillman:
@d-torrance What code did you use to obtain a segfault?  I am trying:
```
needsPackage "SimplicialDecomposability"
R = QQ[a..f];
count = 0;
while true do (
  count = count + 1;
  if count % 1000 == 0 then << "." << flush;
  if count % 80000 == 0 then << endl << flush;
  S = simplicialComplex {a*b*c, c*d, d*e, e*f, d*f};
  isSheddingVertex(a, S);
  )
```
but after a million or so calls, no problem yet! (On Apple M1).

d-torrance:
I had pretty similar code:

```m2
i = 0
while true do (
    print i;
    i = i + 1;
    R = QQ[a..f];
    S = simplicialComplex {a*b*c, c*d, d*e, e*f, d*f};
    isSheddingVertex(a, S))
```

It was on an amd64 machine running Ubuntu 21.10.

d-torrance:
I realized after the fact that I ran that while loop on an unpatched copy of Macaulay2 without this change!  I ran it again on a patched copy and it was still chugging along when I left my office last night...

---

Last night's [amd64 Ubuntu 21.10 PPA build](https://launchpadlibrarian.net/593841212/buildlog_ubuntu-impish-amd64.macaulay2_1.19.1.1+git202203291520-0ppa202203291524~ubuntu21.10.1_BUILDING.txt.gz) had another segfault, also apparently in `interface.dd`:

```m2
 -- running   check(3, "SpecialFanoFourfolds")                              
 ulimit -c unlimited; ulimit -t 700; ulimit -m 850000; ulimit -s 8192; ulimit -n 512;  cd /tmp/M2-139214-0/7-rundir/; GC_MAXIMUM_HEAP_SIZE=400M "/<<PKGBUILDDIR>>/M2/usr-dist/x86_64-Linux-Ubuntu-21.10/bin/M2-binary" -q --int --no-randomize --no-readline --silent --stop --print-width 77 -e 'needsPackage("SpecialFanoFourfolds",Reload=>true,FileName=>"/<<PKGBUILDDIR>>/M2/Macaulay2/packages/SpecialFanoFourfolds.m2")' <"/tmp/M2-139214-0/6.m2" >>"/tmp/M2-139214-0/6.tmp" 2>&1
/tmp/M2-139214-0/6.tmp:0:1: (output file) error: Macaulay2 exited with status code 1
/tmp/M2-139214-0/6.m2:0:1: (input file)
M2: *** Error 1
 -- 8.24283 seconds elapsed

...

SpecialFanoFourfolds.m2:2921:1-2932:1: error:
 -- -- -*- M2-comint -*- hash: 1433144546
 -- 
 -- i1 : --/<<PKGBUILDDIR>>/M2/Macaulay2/packages/SpecialFanoFourfolds.m2:2932: location of test code
 --       -- Test 3 -- 21 examples from GMtables
 --      X = for i from 1 to 21 list (
 --         A = GMtables(i,ZZ/65521);
 --         time specialGushelMukaiFourfold((rationalMap(ideal A_0,Dominant=>2)) ideal A_1,InputCheck=>0)
 --      ); 
 -- -- SIGSEGV
 -- -* stack trace, pid: 139281
 --  0# stack_trace(std::ostream&, bool) at ../d/main.cpp:127
 --  1# segv_handler at ../d/main.cpp:244
 --  2# 0x00007FFB08FF1520 in /lib/x86_64-linux-gnu/libc.so.6
 --  3# interface_rawSaturate at ../d/interface.dd:292
 --  4# evaluate_evalraw at ../d/evaluate.d:1303
```

d-torrance:
Possibly related, the same `SpecialFanoFourfolds` test ran out of memory on a [few](https://launchpadlibrarian.net/593858296/buildlog_ubuntu-bionic-armhf.macaulay2_1.19.1.1+git202203291520-0ppa202203291524~ubuntu18.04.1_BUILDING.txt.gz) [armhf](https://launchpadlibrarian.net/593864608/buildlog_ubuntu-focal-armhf.macaulay2_1.19.1.1+git202203291520-0ppa202203291524~ubuntu20.04.1_BUILDING.txt.gz) and [s390x](https://launchpadlibrarian.net/593851919/buildlog_ubuntu-impish-s390x.macaulay2_1.19.1.1+git202203291520-0ppa202203291524~ubuntu21.10.1_BUILDING.txt.gz) builds:

```m2
SpecialFanoFourfolds.m2:2921:1-2932:1: error:
 -- 
 -- i1 : --/<<PKGBUILDDIR>>/M2/Macaulay2/packages/SpecialFanoFourfolds.m2:2932: location of test code
 --       -- Test 3 -- 21 examples from GMtables
 --      X = for i from 1 to 21 list (
 --         A = GMtables(i,ZZ/65521);
 --         time specialGushelMukaiFourfold((rationalMap(ideal A_0,Dominant=>2)) ideal A_1,InputCheck=>0)
 --      ); 
 -- 
 -- 
 --  *** out of memory trying to allocate 131108 bytes, exiting ***
 ```

PR #2428, opened 2022-04-07
fix a bug with multi-graded rings

kschwede:
This fixes a bug with multi-graded rings that David Eisenbud pointed out.


DanGrayson:
Could you also add a test for the thing you fixed?

kschwede:
> Could you also add a test for the thing you fixed?

Done.

DanGrayson:
Thanks!

PR #2431, opened 2022-04-13
Remove unnecessary steps in autotools submodule builds

d-torrance:
We check that we really need to do the following things before we do them:

* Run `make install`.  (Otherwise, we reinstall everything, so any reverse dependencies end up getting rebuilt.  Since memtailor was being reinstalled, mathic was getting rebuilt.  And since it was getting reinstalled, mathicgb was getting rebuilt, causing these to be rebuilt even we if had only made changes to Macaulay2 itself.)
* Copy the license files.

### DIFF
--- a/M2/GNUmakefile.in
+++ b/M2/GNUmakefile.in
@@ -140,7 +140,7 @@ ifeq ($(no-check-$1),)
 	ulimit -v 1000000 && $(MAKE) -C submodules/$1 check
 endif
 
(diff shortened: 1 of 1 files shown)


d-torrance:
Oops, the builds are failing.  I'll investigate.

d-torrance:
Found a [silly typo](https://github.com/Macaulay2/M2/compare/759cc6754eb4645fc4e155fc9700a735cca858df..4630a508e189d4ccfd1cf8f9e762e5264320f594)... this should work!

PR #2433, opened 2022-04-13
Core documentation

DanGrayson:
making progress on issue #1668


pzinn:
what's this `toList1` vs `toList`? just curious

DanGrayson:
> what's this `toList1` vs `toList`? just curious

I found a package that defined a method `tolist`, which they had to do simply because `toList` is not a method function.  So I renamed the compiled function to `toList1` and made a method function called `toList`, which can be overloaded.

Issue #1856, opened 2021-01-22
Powers of 0 over a finite field


8d1h:
Hi, over a finite field, certain powers of 0 are given as 1.
```macaulay2
F = ZZ/5;
(0..10) / (i->0_F^i) -- returns (1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0)
```
If `GF` is used, the result is different.
```macaulay2
F = GF 5;
(0..10) / (i->0_F^i) -- returns (0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0)
```
Is this the expected behavior?

mikestillman:
Hmmm, I would hope it would be consistent.  I believe that we did decide at one point that a^0 == 1, for all a. I'll look into this.  Flint is being used for both, so hopefully it will be an easy fix.

@DanGrayson (and everyone), what should our default answer be for this?

DanGrayson:
I think it's fine if x^0 is always 1, no matter the value of x, because, at least for natural numbers, the natural way to think of m^n is as the number functions from an n element set to an m element set, so 0^0==1 is fine.

But it looks like a short is made in the results shown, based on the exponent of the multiplicative group of the field, so that x^4=1 for all x.  That's certainly not right.

mikestillman:
@DanGrayson I have fixed this (I believe), will issue a pull request soon.

mikestillman:
Fixed in PR #2440

Issue #638, opened 2017-06-16
isHomogeneous doesn't seem to work for Weyl algebras


foxjwill:
Consider the ring defined in M2 by 

```D = QQ[x, dx, WeylAlgebra=>{x=>dx}, Degrees=>{1,-1}]```
 
This is a quotient of the free algebra QQ<X,dX> by the two-sided ideal J generated by dX*X-X*dX-1. Assigning the free algebra the grading induced by deg(X)=1=-deg(dX), one checks that the ideal J is homogeneous. Hence, D, with the given grading, is (ZZ-)graded.

### Problem
Each of the following return false, despite them all being true:

- `isHomogeneous D`
- `isHomogeneous x`
- `isHomogeneous ideal x`

and so on.

**Comment by antonleykin** [comment]
A `WeylAlgebra`  W is not graded, so `isHomogeneous W` returns false.
The general behaviour of `isHomogeneous RingElement` is to return false when `isHomogeneous Ring` is `false`. An alternative behaviour would be to give an error.

**Comment by DanGrayson** [comment]
General Weyl algebras are not homogeneous, but this one is, because of the negative degree assigned to dx.

**Comment by antonleykin** [comment]
@mahrud: Let's fix `rawIsHomogeneous` in the kernel.

**Comment by ggsmith** [comment]
In version 1.19.1, I get slightly different behaviour than reported above.  However, the bug has not been fixed.
```
i1 : D = QQ[x,dx, WeylAlgebra => {x => dx}, Degrees => {1,-1}]

o1 = D

o1 : PolynomialRing, 1 differential variables

i2 : isHomogeneous D

o2 = true

i3 : isHomogeneous x

o3 = false

i4 : degree x

o4 = {1}

o4 : List

i5 : isHomogeneous dx

o5 = false

i6 : degree dx

o6 = {-1}

o6 : List

i7 : isHomogeneous (x*dx)

o7 = false

i8 : degree (x*dx)

o8 = {0}

o8 : List
```

mikestillman:
Fixed in PR #2440.

PR #2440, opened 2022-04-15
Fix git issues #1856, #638

mikestillman:
This is the first of several pull requests (I hope) for fixing bugs in the engine.  This one fixes the two bugs mentioned.  I might still need to add a test file for #638, but I have added one for #1856.

0^0 should consistently be 1.  0^n, for n>0 should consistently be 0 (this was a bug in some GF rings).

Elements in Weyl algebras were always coming out inhomogeneous.  Now, if the x and dx variables have degrees which are negative of each other, homogeneous elements should be recognized as such (fixed with @antonleykin )

### DIFF
--- a/M2/Macaulay2/e/GF.cpp
+++ b/M2/Macaulay2/e/GF.cpp
@@ -35,7 +35,6 @@ private:
 #include "interrupted.hpp"
 
 #include "aring-m2-gf.hpp"
-#include <iostream>
 
 bool GF::initialize_GF(const RingElement *prim)
 {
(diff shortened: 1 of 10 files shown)


DanGrayson:
That's a thorough test file!

mikestillman:
@DanGrayson Looks like I screwed something up (in ffpack exponentiation, in characteristic 2).  I'll fix it, add in the other test, and we can see how it goes.

DanGrayson:
@mikestillman -- is there a reason for the test "ARingZZpFFPACK.arithmetic2" to have failed here?

DanGrayson:
OOps, just saw your message above, so maybe you're on it already.

mikestillman:
@DanGrayson You might want to look at this, but it is ready to be merged, once you agree.

PR #2451, opened 2022-04-19
Generalize binomial to allow noninteger arguments

d-torrance:
Previously, we got errors using `binomial` with integers that weren't `ZZ` objects:

```m2
i1 : binomial(5.0, 6/2)
stdio:1:1:(3): error: no method found for applying binomial to:
     argument 1 :  5 (of class RR)
     argument 2 :  3 (of class QQ)
```
It also makes sense to allow noninteger arguments using the gamma function, which is what we already do for `!`.  In fact, we can do this without changing the existing code!

For example:
```m2
i1 : binomial(5.0, 6/2)

o1 = 10

i2 : binomial(5.1, pi)

o2 = 10.2930676778848

o2 : RR (of precision 53)
```
Note that the arguments are cast to `ZZ` if possible so the output is a `ZZ` object whenever both arguments are integers regardless of their type.


d-torrance:
> I'm not sure I understand the logic of this change. in particular why `i=floor i` and `n=floor n`? either this is a function that now accepts real arguments, and it should treat them as such, or it should only accept integer arguments. (and the user can use `floor` appropriately when calling it).

Yeah, I wasn't confident that this behavior made the most sense.  You've convinced me that it doesn't.  :)

I'll push an updated PR soon.

PR #2452, opened 2022-04-19
Add --with-system-gc configure option to use existing gc on system

d-torrance:
This option restores the pre-#2446 behavior of using the system gc when available.  If merged, I plan to use it for the Debian package so I don't need to [patch `configure.ac` directly](https://salsa.debian.org/math-team/macaulay2/-/blob/a9c27f627ac2ac9fb8977b68a9fdbc34d6c68ad8/debian/patches/use-debian-gc.patch).  I also propose that we use it for the GitHub builds so we're not rebuilding gc every time.


DanGrayson:
Good idea.

PR #2456, opened 2022-04-20
Use AC_ARG_WITH instead of AC_ARG_ENABLE for mathicgb/etc. checks

d-torrance:
It's a subtle difference, but `AC_ARG_ENABLE` is intended for use with internal features and `AC_ARG_WITH` for external packages, so the latter is more appropriate in this case.

So to use the mathicgb library already available on the system, we now use `--with-system-mathicgb` instead of `--disable-build-mathicgb`.

This also adds consistency with the [new `--with-system-gc` option](https://github.com/Macaulay2/M2/pull/2452).

### DIFF
--- a/M2/INSTALL
+++ b/M2/INSTALL
@@ -64,12 +64,18 @@ Macaulay2.  If your operating system distinguishes between development
 versions and run-time versions of the libraries, you will need the development
 version.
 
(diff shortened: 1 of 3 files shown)


d-torrance:
Just added a new commit updating `INSTALL` so it documents the various `--with-system-X` options as well as Debian/Ubuntu package info.

PR #2463, opened 2022-04-25
Support for unicode synonyms (#1186)

pzinn:
This implements #1186. It adds a bunch of unicode synonyms; the workaround for the problem in the discussion, namely that `R ⊗ S` works but not `R⊗S`, is to consider all unicode starting with 0xE2 (which includes all mathematical operators) as special (non alphanumeric) symbols that cannot be part of the name of a symbol. The advantage is that it doesn't force to rewrite the parser to fully support unicode, which would be a serious rewrite; of course, it doesn't solve other unicode-related issues (cf #1069).

### DIFF
--- a/M2/Macaulay2/d/ctype.d
+++ b/M2/Macaulay2/d/ctype.d
@@ -25,7 +25,7 @@ foreach c in " \t\r"	                   do setchartype(c,WHITE);
 foreach c in "\n"                          do setchartype(c,NEWLINE);
 foreach c in "$"                           do setchartype(c,DOLLAR);
 
-for c from 128 to 225	       	    	   do setchartype(char(c),ALPHA);
+for c from 128 to 225	       	    	   do setchartype(char(c),ALPHA);  -- 226 is unicode math symbol
 for c from 227 to 255	       	    	   do setchartype(char(c),ALPHA);
 					      setchartype('\'',ALPHA);
 					      setchartype('\"',QUOTE);
@@ -54,6 +54,7 @@ export isnewline  (c:char):bool := (chartype(c) & NEWLINE  ) != 0;
 export isquote    (c:char):bool := (chartype(c) & QUOTE    ) != 0;
 
 export isalnum  (s:string):bool := (
+     if int(uchar(s.0)) == 226 && length(s) == 3 then return true; -- unicode math symbol
      foreach c in s do if !isalnum(c) then return false;
      true);
 
--- a/M2/Macaulay2/d/ctype.d
+++ b/M2/Macaulay2/d/ctype.d
@@ -25,7 +25,8 @@ foreach c in " \t\r"	                   do setchartype(c,WHITE);
 foreach c in "\n"                          do setchartype(c,NEWLINE);
 foreach c in "$"                           do setchartype(c,DOLLAR);
 
-for c from 128 to 255	       	    	   do setchartype(char(c),ALPHA);
+for c from 128 to 225	       	    	   do setchartype(char(c),ALPHA);
+for c from 227 to 255	       	    	   do setchartype(char(c),ALPHA);
 					      setchartype('\'',ALPHA);
 					      setchartype('\"',QUOTE);
 
--- a/M2/Macaulay2/d/lex.d
+++ b/M2/Macaulay2/d/lex.d
@@ -410,6 +410,11 @@ gettoken1(file:PosFile,sawNewline:bool):Token := (
 		    return errorToken
 		    )
 	       is word:Word do return Token(word,file.filename, line, column, loadDepth,globalDictionary,dummySymbol,sawNewline))
+	  else if ch == 226 then (
+	       tokenbuf << char(getc(file));
+	       tokenbuf << char(getc(file));
+	       tokenbuf << char(getc(file));
+	       return Token(makeUniqueWord(takestring(tokenbuf),parseWORD),file.filename, line, column, loadDepth,globalDictionary,dummySymbol,sawNewline))
 	  else (
 	       when recognize(file)
 	       is null do (
--- a/M2/Macaulay2/m2/exports.m2
+++ b/M2/Macaulay2/m2/exports.m2
@@ -1244,7 +1244,33 @@ export {
         "numcols" => "numColumns",
         "numrows" => "numRows",
         "res" => "resolution",
-        "sub" => "substitute"
+        "sub" => "substitute",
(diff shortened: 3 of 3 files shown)


pzinn:
There seems to be a conflict with the package MultiprojectiveVarieties which exports symbols  "∏","⋂","⋃","PP".
I'll have to investigate why it does so.

M2/Macaulay2/d/ctype.d
```diff
@@ -25,7 +25,8 @@ foreach c in " \t\r"	                   do setchartype(c,WHITE);
 foreach c in "\n"                          do setchartype(c,NEWLINE);
 foreach c in "$"                           do setchartype(c,DOLLAR);
 
-for c from 128 to 255	       	    	   do setchartype(char(c),ALPHA);
+for c from 128 to 225	       	    	   do setchartype(char(c),ALPHA);
+for c from 227 to 255	       	    	   do setchartype(char(c),ALPHA);
```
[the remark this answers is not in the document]
pzinn: 0xE2 (226) is where mathematical symbols are in unicode.

pzinn:
specifically the issue is with the product symbol ∏, which I made synonym of `product`, but which is used by MultiprojectiveVarieties for products of varieties. unsure how to reconcile.

DanGrayson:
Packages should be able to export symbols that are the same as core symbols, with only a warning message.

DanGrayson:
That package defines "∏","⋂","⋃".  Those should be available for use by everyone.  But the package defines ∏ to be just for product of varieties.  Ideally, the package would define a binary method for MultiprojectiveVariety * MultiprojectiveVariety and "product" would then figure out what to do.  Currently the package defines a binary method for MultiprojectiveVariety ** MultiprojectiveVariety, which is sort of odd, because ** is for tensor product, not product.

@giovannistagliano

M2/Macaulay2/d/ctype.d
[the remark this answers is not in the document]
DanGrayson: Could add a comment to clarify the code?

pzinn:
> That package defines "∏","⋂","⋃". Those should be available for use by everyone. But the package defines ∏ to be just for product of varieties. Ideally, the package would define a binary method for MultiprojectiveVariety * MultiprojectiveVariety and "product" would then figure out what to do. Currently the package defines a binary method for MultiprojectiveVariety ** MultiprojectiveVariety, which is sort of odd, because ** is for tensor product, not product.
> 
> @giovannistagliano

yes ideally there'd be a way of reconciling the use of ∏ ... otherwise I can just remove ∏ and ∑ it's not a big deal, mildly unsatisting.

DanGrayson:
> yes ideally there'd be a way of reconciling the use of ∏ ... otherwise I can just remove ∏ and ∑ it's not a big deal, mildly unsatisfying.

But what's the problem?  Why doesn't the symbol ∏ in the package just "shadow" the symbol ∏ in the core?

giovannistagliano:
> That package defines "∏","⋂","⋃". Those should be available for use by everyone. But the package defines ∏ to be just for product of varieties. Ideally, the package would define a binary method for MultiprojectiveVariety * MultiprojectiveVariety and "product" would then figure out what to do. Currently the package defines a binary method for MultiprojectiveVariety ** MultiprojectiveVariety, which is sort of odd, because ** is for tensor product, not product.
> 
> @giovannistagliano

I use ** for product because * is used for the intersection (as in in the class Set).

pzinn:
> But what's the problem? Why doesn't the symbol ∏ in the package just "shadow" the symbol ∏ in the core?

oh, I see. I misunderstood your comment. there is a problem indeed. Fixing.

pzinn:
OK I found the problem, it's  in binding.d makeSymbol where the string of the symbol is tested (once more!) for alphanumericity:
```
     if dictionary.frameID == 0 && isalnum(word.name) && !thread
                                   ------------------
```
I can easily fix this but first, why such a test? removing it seems to be harmless...

M2/Macaulay2/d/ctype.d
```diff
@@ -54,6 +54,7 @@ export isnewline  (c:char):bool := (chartype(c) & NEWLINE  ) != 0;
 export isquote    (c:char):bool := (chartype(c) & QUOTE    ) != 0;
 
 export isalnum  (s:string):bool := (
+     if int(uchar(s.0)) == 226 && length(s) == 3 then return true; -- unicode math symbol
```
DanGrayson, line 57: I'm not sure it's a good idea to change the definition of "isalnum", which has a specific meaning.  Also, we already allow unicode characters in symbols, as here:
```m2

i11 : 你好 = 1

o11 = 1

i12 : 你好

o12 = 1
```

So it doesn't seem necessary.

pzinn: unicode symbols are still allowed. the only extra rule is, unicodes starting with 0xe2 (which means mathematical symbols and a few others) can only be used stand-alone as symbols. your code above will run just fine.

pzinn: the alternative is to revert ctype.d to how it was before, and have an extra test when parsing in lex.d.
I thought it was uglier, but if you prefer that can be done.

DanGrayson:
> I use ** for product because * is used for the intersection (as in in the class Set).

We should probably stop using * for intersection of sets and start using the appropriate unicode symbol.

M2/Macaulay2/d/ctype.d
DanGrayson: Well, "isalnum" should continue to check whether ascii strings are alphanumeric - that's what its name means.  Maybe you could do something different higher up.

I assume the goal to make `x≠y` work the way `x*y` works, which is a legitimate goal.

pzinn: that's right. right now `x≠y` is not considered alphanumeric. Is that the part you disagree with?

DanGrayson: No, I think we shouldn't call `≠` alphanumeric.  The adjective "alphanumeric" has a standard meaning and applies to strings of ascii characters -- it means each character is an alphabetic letter or a numeric digit.

pzinn: I see. before this PR, everything from 128 to 255 was considered ALPHA -- and a fortiori, ALNUM. what you're suggesting is a serious change in `ctype.d`. This PR had a much more modest goal.

PR #2465, opened 2022-04-25
Autotools build updates for syntax highlighting

d-torrance:
A few changes prompted by the discussion in #2408:

* Previously, we created the JavaScript file necessary for syntax highlighting the html documentation whenever `npm` was found during `configure`.  We now make it opt-in with the `--enable-syntax-highlighting` option.
* Previously, we downloaded the necessary JavaScript prereqs (using `npm install`) regardless of whether or not the user passed `--enable-download` to `configure`.  It's now a requirement.
* Running `npm install` has been moved earlier in the build, alongside fetching the other libraries as part of `make -C libraries fetch`.

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -166,7 +166,7 @@ jobs:
           export CPPFLAGS="-I`brew --prefix`/include"
           export  LDFLAGS="-L`brew --prefix`/lib \
             -L`brew --prefix python`/Frameworks/Python.framework/Versions/${PYVERSION}/lib"
-          ../../configure --enable-download --with-python --with-system-gc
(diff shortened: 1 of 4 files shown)


DanGrayson:
In the 5th build, the npm command has failed.

d-torrance:
> In the 5th build, the npm command has failed.

Hmm, interesting.  It's from one of the cmake builds, so I'm not sure how these autotools-only changes would have caused it.

DanGrayson:
> > In the 5th build, the npm command has failed.
> 
> Hmm, interesting. It's from one of the cmake builds, so I'm not sure how these autotools-only changes would have caused it.

Yes, that's strange.  I've restarted the job.

PR #2469, opened 2022-05-03
update some references in doc

giovannistagliano:


pzinn:
Is that a good idea? I don't think packages should mess with the output mode of Macaulay2...

giovannistagliano:
> Is that a good idea? I don't think packages should mess with the output mode of Macaulay2...

I think it'd be better to have output printed as tex only if tex is available. Unfortunately I didn't have time to implement tex for for the types in my packages.

d-torrance:
> Unfortunately I didn't have time to implement tex for for the types in my packages.

One quick and dirty temporary solution is to use the following:

```m2
texMath X := texMath @@ net
```
Then the webapp output will look just like the standard output without needing to change `topLevelMode`.

DanGrayson:
Yes, packages should not set the top level mode -- that's only for top level interfaces.

giovannistagliano:
> > Unfortunately I didn't have time to implement tex for for the types in my packages.
> 
> One quick and dirty temporary solution is to use the following:
> 
> ```m2
> texMath X := texMath @@ net
> ```
> 
> Then the webapp output will look just like the standard output without needing to change `topLevelMode`.

Many thanks! This however does not fix the second line of output.

giovannistagliano:
> Yes, packages should not set the top level mode -- that's only for top level interfaces.

Okay, now just a warning is printed.

This however is a problem for the types in my packages.

Before:
`
i1 : needsPackage "Cremona";

i2 : f = rationalMap vars(QQ[vars(0..2)])

o2 = -- rational map --
     source: Proj(QQ[a, b, c])
     target: Proj(QQ[a, b, c])
     defining forms: {
                      a,
                      
                      b,
                      
                      c
                     }

o2 : RationalMap (linear rational map from PP^2 to PP^2)

i3 : f;

o3 : RationalMap (linear rational map from PP^2 to PP^2)
`

Now:
`
i1 : needsPackage "Cremona";

i2 : f = rationalMap vars(QQ[vars(0..2)])

o2 = RationalMap {...13...}

o2 : RationalMap

i4 : f;
`

d-torrance:
> This however does not fix the second line of output.

Setting `X#{WebApp,AfterPrint}` to be the same thing as `X#{Standard,AfterPrint}` seems to do the job:

image

giovannistagliano:
> > This however does not fix the second line of output.
> 
> Setting `X#{WebApp,AfterPrint}` to be the same thing as `X#{Standard,AfterPrint}` seems to do the job:
> 
> image

OK, thanks, this seems to work.  I think this should be done automatically for all types where texMath is not available.

pzinn:
yes, we discussed this with @antonleykin a while ago (though I can't find the discussion anywhere). The point is, one can do something like
```
html BasicList := x -> (if lookup(texMath,class x) === BasicList#texMath and lookup(net,class x) =!= BasicList#net then html @@ net else lookup(html,Thing)) x;
```
(I redefine `html` rather than `texMath`, BTW -- if output is just plain text, might as well render in HTML, no need to go through LaTeX).
and similarly for say HashTables. but this breaks inheritance,  and breaking inheritance has all kinds of annoying side effects.
So I don't think this is the correct approach.

giovannistagliano:
>

DanGrayson:
Thanks!

PR #2474, opened 2022-05-05
Matroids v1.5

jchen419:


jchen419:
Hi Dan, I'm not sure what's causing this error: the file `Matroids/SmallMatroids.txt` should still be there (as can be seen from [https://github.com/Macaulay2/M2/pull/2474/files](https://github.com/Macaulay2/M2/pull/2474/files)), and I can run `check_0 "GKMVarieties"` just fine on my machine. The size did significantly increase from before though (to include matroids on 9 elements): is it possible that the size increase is the cause?

DanGrayson:
Jay, your former way of getting the file was this: `select(apply(path, p -> p | "Matroids/SmallMatroids.txt"), p -> fileExists p)`.  But now you are doing just this: `get db`.  So now the path to the file must start in the current directory, instead of in one of the directories on the path.

So just revert to the old way.  Better yet, use `searchPath` to do the work for you.

DanGrayson:
@ggsmith -- in HyperplaneArrangements you have the following code:
```m2
multiplierIdeal(QQ, CentralArrangement, List) := Ideal => (s,A,m) -> (
     if (#hyperplanes A != #m) then error "expected one weight for each hyperplane";
     R := ring A;
     if not A.cache.?irreds then
	  A.cache.irreds = select(flatten drop(flats(A),1), F->(0 != euler F));
     exps := A.cache.irreds/(F->max(0,floor(s*weight(F,m))-rank(F)+1));
     if not A.cache.?multipliers then A.cache.multipliers = new MutableHashTable;
     if not A.cache.multipliers#?exps then (
	  ideals := A.cache.irreds/(F-> trim ideal toList (A_F));
	  A.cache.multipliers#exps = intersect apply(#exps, i->(ideals_i)^(exps_i)))
     else
     	  A.cache.multipliers#exps
     )
```

Sometimes (as in the failing runs on this pull request) the result of `toList(A_F)` can be an empty list, so when the ideal of that is formed, it's in ZZ instead of in the ring of the arrangement.

I tried changing the code so it would make the zero ideal in the ring of the arrangement, but then there is still an error:
```m2
i1 : loadPackage "MultiplierIdeals"

o1 = MultiplierIdeals

o1 : Package

i2 : R = QQ[x,y,z]

o2 = R

o2 : PolynomialRing

i3 : A = arrangement ((x^2 - y^2)*(x^2 - z^2)*(y^2 - z^2)*z)

o3 = {z, y - z, y + z, x - z, x + z, x - y, x + y}

o3 : Hyperplane Arrangement 

i4 : logCanonicalThreshold A
stdio:4:1:(3): error: division by zero
```

ggsmith:
I confused.  I am unable to reproduce this error on my machine.  Both ```HyperplaneArrangements``` and ```MultiplierIdeals``` install and check without errors.

```
i1 : needsPackage "HyperplaneArrangements";

i2 : R = QQ[x,y,z];

i3 : A = arrangement((x^2-y^2)*(x^2-z^2)*(y^2-z^2)*z)

o3 = {z, y - z, y + z, x - z, x + z, x - y, x + y}

o3 : Hyperplane Arrangement 

i4 : logCanonicalThreshold A

     3
o4 = -
     7

o4 : QQ
```

mikestillman:
@ggsmith Your comments might be related to my last pull request that was accepted yesterday: I disallow division by zero now, whereas one day ago, that was not the case all the time...  Did you rebuild since then?

DanGrayson:
> I am unable to reproduce this error on my machine. Both `HyperplaneArrangements` and `MultiplierIdeals` install and check without errors.

This pull request modifies the Matroids package, which HyperplaneArrangements uses, so to debug this, one has to check out the code of this pull request.  As a result of the changes, for the example I showed, the length of the list `A.cache.irreds` goes from 14 to 34, and the lengths of some of the lists `toList (A_F)` are 0.  Maybe that's the result of something incorrect in the pull request.

But, just theoretically, would you expect any of the lists `toList (A_F)` to be empty?

jchen419:
One change which might be relevant is that in this PR, I changed the behavior of `flats(Matroid, ZZ)`: now `flats(M, s)` returns all intersections of s hyperplanes in M (as opposed to all flats of rank s in M). On the other hand, I still get an error with the example above when I revert to the previous behavior. @ggsmith: do you recall using the instance `flats(Matroid, ZZ)` in `HyperplaneArrangements`?

ggsmith:
@jchen419: The ```HyperplaneArrangements``` package definitely does use ```flats(Matroid,ZZ)```; for example in ```flats(ZZ, CentralArrangement)```.

jchen419:
@ggsmith: I see. I am trying to debug this, and am running into the following:
```
i1 : needsPackage "HyperplaneArrangements"

o1 = HyperplaneArrangements

o1 : Package

i2 : R = QQ[x,y,z]; A = arrangement((x^2-y^2)*(x^2-z^2)*(y^2-z^2)*z)

o3 = {z, y - z, y + z, x - z, x + z, x - y, x + y}

o3 : Hyperplane Arrangement

i4 : flat(A, {})
stdio:3:2:(3): error: expected list, sequence, or set

i5 : flat(A, {})
stdio:4:1:(3): error: no method for prefix operator not applied to object: not displayed
```
For some reason the first error is different from subsequent errors (probably due to some caching mechanisms). Do you know what could be causing this?

ggsmith:
@jchen419: As far as I can tell, the ```flat``` method does not use anything connected to the ```Matroids``` package, nor is anything cached.  I don't what is going on with your example.

jchen419:
@ggsmith: I agree that the code for `flat` is completely unrelated to `Matroids`, which is why I'm confused as to why I'm seeing this bug. Hopefully just a weird quirk on my machine...

jchen419:
@ggsmith: Actually, I take that last comment back: `flat` is using the `Matroids` package, at least with `Validate => true`, by way of calling `closure` (which calls `hyperplanes`). With that, I think I found one bug: changing `!=` to `=!=` in line 670 of `HyperplaneArrangements`:
```
             if newF != closure(A, F) then error "not a flat";
```
 fixes the bug in my example above.

DanGrayson:
A test for GKMVarieties is still failing.

jchen419:
Thanks @DanGrayson. I have to say, this has greatly undermined my confidence in the current state of tests: what with (non-engine!) code working on some architectures but failing on others, and tests in Core(!) failing on a top-level user package (and of course none of the errors being reproducible on my personal machine, so each revision attempt is a stab in the dark...)

mikestillman:
@jchen419 What was the problem?  (i.e. what undermined your confidence?)

DanGrayson:
The most recent failure was this timing test in the core:

https://github.com/Macaulay2/M2/runs/6344709618?check_suite_focus=true#step:12:498

It's not reliable, so I've just commented it out.

jchen419:
@mikestillman: the test in Core that failed in the very last run was mentioned by @DanGrayson. But in general, there were all sorts of funky things happening: I really don't understand how top-level code will work on some architectures/compilers but not on others. In this case, the first error was caused by a file not being found, only on certain builds (and apparently `searchPath` did nothing to remedy the issue on those builds), but on my 2nd try those builds were apparently finding the file, but not reading it.

In another direction, there were even some cases where some tests were passing that I later realized definitely should not have passed (of course the tests have now been corrected, but the fact that incorrect tests were passing is quite suspicious).

PR #2476, opened 2022-05-05
Update mathic and mathicgb submodules

d-torrance:
I just uploaded the latest mathic package to Debian and noticed that the submodule in Macaulay2 is a couple commits behind (missing https://github.com/Macaulay2/mathic/pull/7).


d-torrance:
> I believe the commit for mathicgb needs to be updated too. I have done that on a branch that I will be submitting a pull request on, but it isn't related to my pull request, so perhaps we should do that here.

Done!

mikestillman:
@d-torrance By the way, in the cmake build, I still get many warnings from mathic and/or mathicgb.  They might all be the same one (they are deprecated behavior warnings.  [-Wdeprecated-declarations]...

On a similar note, on my apple M1, I now get many warnings (all the same, I think) about macros being redefined.  This is not related to mathic, but to flint and givaro, regarding redefined macros  [-Wmacro-redefined].

d-torrance:
> @d-torrance By the way, in the cmake build, I still get many warnings from mathic and/or mathicgb. They might all be the same one (they are deprecated behavior warnings. [-Wdeprecated-declarations]...

I get these too in the autotools build.  I think they're from including `tbb.h` directly rather than the specific header we need.

mikestillman:
@d-torrance Actually, these are new (bidirectional iterator tags, or something similar).  I'm using tbb2021, so I don't get those deprecation warnings from tbb.

d-torrance:
> @d-torrance Actually, these are new (bidirectional iterator tags, or something similar). I'm using tbb2021, so I don't get those deprecation warnings from tbb.

Ah, ok.  Debian's still using TBB 2020, so I haven't seen those yet.

mikestillman:
@DanGrayson I'm going to accept this request.

Issue #1851, opened 2021-01-21
gcd(0,0,...) leads M2 1.17 to crash


giovannistagliano:
This is a simple example where i get a crash of M2: 
`O = 0_(ZZ/41[x]); gcd {O,O}`

This is another example:
`O = 0_(ZZ/41[x]); gcd {O,O,x}`

But strangely the following work:
`O = 0_(ZZ/41[x]); gcd {O,x,O}; gcd {O,x}; gcd {O}; `

Before my terminal closes I get this error:

`Flint exception (Impossible inverse):
    Cannot invert modulo 41*1
Aborted (core dumped)
`

mikestillman:
I missed this one, apparently.  I'll work on fixing it.  Sorry for the late response on it!

mikestillman:
@DanGrayson This one is easy to fix, I think.  Should I fix it for 1.20?  Or have you already started release builds...?

DanGrayson:
Sure, I can wait a little while.

PR #2495, opened 2022-05-11
fix bug #1851 in rawGCD

mikestillman:
The fix was to return the zero element before trying to divide by the lead coefficient (the attempt to divide was crashing, or returning an uncaught exception).


mikestillman:
@DanGrayson I just added a fix to #2496 to this pull request.

DanGrayson:
That's frustrating -- cancelling the third job won't work.  Could you submit a new pull request?

mikestillman:
Yes, I'll add in a test also.

Issue #2496, opened 2022-05-11
0 / 0


pzinn:
this is weird:
```
i1 : R=QQ[x];

i2 : 0_R/0_R

     0
o2 = -
     0

o2 : frac R

i3 : oo+1
stdio:3:3:(3): error: a non unit was found in a ring declared to be a field
```

mikestillman:
aarggh!  @DanGrayson I need to fix this one too this morning...

mikestillman:
I was sure I had handled all divide by zero issues that were left...  I guess not...!  thanks for catching this @pzinn !

pzinn:
> aarggh! @DanGrayson I need to fix this one too this morning...

Sorry :-P

PR #2497, opened 2022-05-11
Fix bugs #1851 and #2496

mikestillman:
fix bugs #1851 and #2496


DanGrayson:
Are we good to go for 1.20 now?

mikestillman:
I am!

DanGrayson:
Okay.

PR #2502, opened 2022-05-13
Update Factory source tarball download url

slel:
Use the ftp mirror at `www.singular.uni-kl.de/ftp/pub/Math/`
after it moved there from `service.mathematik.uni-kl.de`.


DanGrayson:
Already addressed in commit a156a3cd47a27d481167b85eb3846975ebb43986, thanks!!

PR #2515, opened 2022-05-28
Fix typo in alphaCertified documentation

d-torrance:
`Configuration` key to `needsPackage` should be a list of options.  This would ordinarily result in an error, but it didn't since the package was already loaded when generating the example.

---

I noticed this after playing around with the package after Kisun's talk earlier today in Cleveland.


klee669:
Thanks a lot!

PR #2410, opened 2022-03-10
second attempt at implementing #2404

pzinn:
compared to my first attempt, following @mahrud suggestion:
- no primed functions, fewer changes overall
- no strong dependence on tvalues.m2
cons:
- list of redefined functions is hard-coded (currently in typicalvalues.m2)
- this redefinition needs to be done at the very end (had to move it out of reals.m2, though might eventually want to reorganize a bit)
~Before I continue with this I'd like some feedback on which option is better.~

PR #2517, opened 2022-06-03
A few special function updates

d-torrance:
Some of the error messages for special functions that can't handle `RRi` or `CC` arguments were confusing, e.g.,:

```m2
i1 : log1p ii
stdio:1:1:(3): error: expected a number

i2 : erf interval(0, 1)
stdio:2:1:(3): error: expected a number
```
Both `CC` and `RRi` are subclasses of `Number`, so this doesn't really make sense.

So we clarify these messages, adding a `WrongArgRRorRRi` function to the d language to help out:

```m2
i1 : log1p ii
stdio:1:1:(3): error: expected a real number or interval

i2 : erf interval(0, 1)
stdio:2:1:(3): error: expected a real number
```
In the process, I discovered that `BesselJ` and `BesselY` were broken, I think since #2410:

```m2
i3 : BesselJ(2, 3)
stdio:3:1:(3): error: no method found for applying numeric to:
     argument 1 :  infinity (of class InfiniteNumber)
     argument 2 :  3 (of class ZZ)
```

So that's been fixed and unit tests added.
```m2
i3 : BesselJ(2, 3)

o3 = .486091260585891

o3 : RR (of precision 53)
```


d-torrance:
Hmm, a bunch of build failures running `npm`.  I'm guessing this may be because of webpack 5.73.0, which was just released yesterday.  I'll investigate.

d-torrance:
Oops - not related to webpack at all, just a documentation bug on my part.  I'll push a fix in a sec.

PR #2518, opened 2022-06-03
Add `toExternalString(RRi)`

d-torrance:
### Before

```m2
i1 : toExternalString interval(1, 2)

o1 = [1,2]

i2 : value oo === interval(1, 2)

o2 = false
```

### After

```m2
i1 : toExternalString interval(1, 2)

o1 = interval(.1p53e1,.2p53e1)

i2 : value oo === interval(1, 2)

o2 = true
```


DanGrayson:
Thanks.

This is strange:
```m2
i11 : toExternalString 1.

o11 = .1p53e1
```

The answer could have been "1p53".

d-torrance:
It looks like it's written so that the number before the `p` is always between 0 and 1:

https://github.com/Macaulay2/M2/blob/940e8f48f768c114c0ecec771fdce6d5cf479578/M2/Macaulay2/d/gmp1.d#L173-L194

PR #2521, opened 2022-06-04
Use multiple precision for Boost math functions

d-torrance:
This is a followup to #2436, using [Boost's MPFR support](https://www.boost.org/doc/libs/1_79_0/libs/multiprecision/doc/html/boost_multiprecision/tut/floats/mpfr_float.html) so that the functions we use from Boost behave like the ones we use from MPFR.

Note that Boost makes heavy use of templates for this feature, so it takes a while to compile `boostmath-tmp.cc` after these changes.

### Before
```m2
i1 : inverseErf 0.5p100

o1 = .47693627620447

o1 : RR (of precision 53)
```

### After
```m2
i1 : inverseErf 0.5p100

o1 = .476936276204469873381191844309

o1 : RR (of precision 100)
```

### DIFF
--- a/M2/Macaulay2/d/boostmath.dd
+++ b/M2/Macaulay2/d/boostmath.dd
@@ -2,68 +2,78 @@ use util;
 use common;
 
 header "#include <iostream>
+  #include <boost/multiprecision/mpfr.hpp>
   #include <boost/math/special_functions/beta.hpp>
   #include <boost/math/special_functions/erf.hpp>
   #include <boost/math/special_functions/gamma.hpp>
 
-  #define BOOST_MATH_CALL(f, ...)       \\
-  try {                                 \\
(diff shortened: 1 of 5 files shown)


d-torrance:
Converting to a draft for now, as I don't think this is correct, as it's currently just using the default precision set by Boost and not the precision it should be using.

d-torrance:
It's fixed now!

```m2
i1 : inverseErf 0.5p100

o1 = .476936276204469873381418353643

o1 : RR (of precision 100)

i2 : erf oo

o2 = .5

o2 : RR (of precision 100)

i3 : erf .476936276204469873381191844309p100 -- inverseErf 0.5p100 from above

o3 = .499999999999999999999796411626

o3 : RR (of precision 100)
```
Boost actually uses decimal digits for multiprecision, but there's a nice helper function we can use to convert from binary.

Issue #1310, opened 2020-06-20
Feature request: C calls in top level (foreign function interface)


mahrud:
[FFI](https://en.wikipedia.org/wiki/Foreign_function_interface) would allow top level interaction with system calls and dynamic libraries without the need to add bindings in the d language and recompile.

Here's are some examples:
- haskell [demo](https://wiki.haskell.org/FFI_Introduction) and [documentation](https://wiki.haskell.org/Foreign_Function_Interface).
- Julia [manual](https://docs.julialang.org/en/v1/manual/calling-c-and-fortran-code/).

DanGrayson:
This would be a valuable feature, especially if the interface would parse library include files and figure out what to do from them (who wants to manually type 1000 function declarations?).  Implementing it would be a big challenge and take a lot of time.

mahrud:
Thanks, @d-torrance!

PR #2527, opened 2022-06-10
Add ForeignFunctions package

d-torrance:
I've been working on a package that uses [libffi](https://sourceware.org/libffi/) to allow calling C functions from top level (see #1310).  It's still pretty rough, but basically works, so I figured I'd open up a draft PR for comments.  I still need to implement support for some libffi features like structs and variadic functions.

There's one new class, `Pointer`, in the interpreter that is exported by the package, and lots of D language functions for converting Macaulay2 objects into corresponding C objects (e.g., `ZZ` -> `int`) to get a `Pointer` object with their address, and then going back again.

No clue if this works in macOS, and I haven't messed with the CMake build yet, either.

```m2
i1 : loadPackage "ForeignFunctions"

o1 = ForeignFunctions

o1 : Package

i2 : libm = openSharedLibrary "libm.so.6"

o2 = libm.so.6

o2 : SharedLibrary

i3 : cCos = foreignFunction(libm, "cos", "double", "double")

o3 = libm.so.6::cos

o3 : ForeignFunction

i4 : cCos numeric pi

o4 = -1

o4 : RR (of precision 53)

i5 : libc = openSharedLibrary "libc.so.6"

o5 = libc.so.6

o5 : SharedLibrary

i6 : puts = foreignFunction(libc, "puts", "void", "pointer")

o6 = libc.so.6::puts

o6 : ForeignFunction

i7 : puts "Hello, world!"
Hello, world!
```

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -145,7 +145,7 @@ jobs:
       - name: Build Macaulay2 using Ninja
         if: matrix.build-system == 'cmake'
         run: |
-          cmake --build . --target M2-core M2-emacs install-packages
+          cmake --build . --target M2-core M2-emacs M2-highlightjs install-packages
           if [[ "${{ runner.os }}" == "Linux" ]] && [[ "${{ matrix.compiler }}" == "clang-11" ]]; then
               sudo apt-get install -y -q --no-install-recommends dpkg-dev
               echo "GIT_COMMIT=`git describe --dirty --always --match HEAD`" >> $GITHUB_ENV
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -73,7 +73,7 @@ jobs:
         run: |
           brew config
           brew tap macaulay2/tap
-          brew install autoconf automake bison boost tbb ccache cmake ctags gnu-tar libtool make ninja pkg-config yasm
+          brew install autoconf automake bison boost tbb ccache cmake ctags gnu-tar libtool make ninja pkg-config yasm libffi
           brew install --only-dependencies macaulay2/tap/M2
 
 # ----------------------
@@ -132,7 +132,7 @@ jobs:
         run: |
           cmake -S../.. -B. -GNinja \
             -DCMAKE_BUILD_TYPE=Release -DBUILD_NATIVE=OFF \
-            -DCMAKE_PREFIX_PATH="`brew --prefix`" \
+            -DCMAKE_PREFIX_PATH="`brew --prefix`;`brew --prefix libffi`" \
             -DCMAKE_INSTALL_PREFIX=/usr \
             -DWITH_PYTHON=ON \
             --debug-trycompile
--- a/M2/INSTALL-CMake.md
+++ b/M2/INSTALL-CMake.md
@@ -280,6 +280,7 @@ The main targets for building Macaulay2 are:
 - `M2-binary`: build the Macaulay2 executable
 - `M2-core`: generate and copy the Core package
 - `M2-emacs`: generate the M2-mode package for Emacs
+- `M2-highlightjs`: generate syntax highlighting library for javascript
 
 In addition, the following targets are available:
 - `scc1`: build the Safe C Compiler
--- a/M2/INSTALL-CMake.md
+++ b/M2/INSTALL-CMake.md
@@ -23,9 +23,9 @@ There are various tools needed to compile Macaulay2 dependencies.
 - On Mac OS X, using Homebrew, install `autoconf automake bison libtool pkg-config yasm`.
 
 There are 10 libraries that must be found on the system.
-- On Debian/Ubuntu, install `libopenblas-dev libgmp3-dev libxml2-dev libreadline-dev libgdbm-dev libboost-regex-dev libboost-stacktrace-dev libatomic-ops-dev libomp-dev libtbb-dev`.
-- On Fedora/CentOS, install `openblas-devel gmp-devel libxml2-devel readline-devel gdbm-devel boost-devel libatomic_ops-devel libomp-devel tbb-devel`.
-- On Mac OS X, using Homebrew, install `gmp libxml2 readline gdbm boost libatomic_ops libomp tbb`.
+- On Debian/Ubuntu, install `libopenblas-dev libgmp3-dev libxml2-dev libreadline-dev libgdbm-dev libboost-regex-dev libboost-stacktrace-dev libatomic-ops-dev libomp-dev libtbb-dev libffi-dev`.
+- On Fedora/CentOS, install `openblas-devel gmp-devel libxml2-devel readline-devel gdbm-devel boost-devel libatomic_ops-devel libomp-devel tbb-devel libffi-devel`.
+- On Mac OS X, using Homebrew, install `gmp libxml2 readline gdbm boost libatomic_ops libomp tbb libffi`.
 
 **TIP**: x86_64 binary packages for all dependencies on Mac OS X 10.15+ and Linux distributions are available through the [Macaulay2 tap](https://github.com/Macaulay2/homebrew-tap/) for Homebrew. To download the dependencies this way run:
 ```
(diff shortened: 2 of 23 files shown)


mahrud:
Is it possible to declare new types without changing and recompiling `ffi.d`?

d-torrance:
> Is it possible to declare new types without changing and recompiling `ffi.d`?

Not yet, but that's the goal.  :)

mahrud:
> Not yet, but that's the goal. :)

If it can be done without significant function call overhead, that would be very cool! :)

d-torrance:
Variadic functions are working now.  The types need to be specified for the variadic arguments:

```m2
i3 : printf = foreignFunction(libc, "printf", "void", {"pointer", "..."})

o3 = libc.so.6::printf

o3 : ForeignFunction

i4 : printf("%d %s %.1f\n", ("uint", 1), ("pointer", "2"), ("double", 3.0))
1 2 3.0
```

jkyang92:
Maybe not directly in scope, but it would be great to add the ability to associate destructors to pointers (see [https://hackage.haskell.org/package/base-4.16.0.0/docs/Foreign-ForeignPtr.html](https://hackage.haskell.org/package/base-4.16.0.0/docs/Foreign-ForeignPtr.html) for inspiration.) In particular, it will be important to consider how memory management works if we start returning pointers from foreign code.

d-torrance:
Got structs working!

```m2
i2 : tm = foreignStructType("tm", {
	"tm_sec" => int,
	"tm_min" => int,
	"tm_hour" => int,
	"tm_mday" => int,
	"tm_mon" => int,
	"tm_year" => int,
	"tm_wday" => int,
	"tm_yday" => int,
	"tm_isdst" => int,
	"tm_gmtoff" => long,
	"tm_zone" => charstar})

                                                       
o2 = tm

o2 : ForeignStructType

i3 : gmtime = foreignFunction("gmtime", voidstar, voidstar);

i4 : asctime = foreignFunction("asctime", charstar, voidstar);

i5 : epoch = tm value gmtime address long 0

o5 = HashTable{tm_gmtoff => 0}
               tm_hour => 0
               tm_isdst => 0
               tm_mday => 1
               tm_min => 0
               tm_mon => 0
               tm_sec => 0
               tm_wday => 4
               tm_yday => 0
               tm_year => 70
               tm_zone => GMT

o5 : ForeignObject of type tm

i6 : asctime address epoch

o6 = Thu Jan  1 00:00:00 1970


o6 : ForeignObject of type charstar
```

mahrud:
Cool, I'll try it out soon!

I agree with Jay that making sure memory management is done correctly (ideally without needing the garbage collector) is important.

d-torrance:
Currently, I have it set up to use the garbage collector for allocating memory for any C objects we create from some Macaulay2 object.  For example:

```m2
i2 : peek int 0

o2 = ForeignObject{address => 0x7f97edaf7e70}
                   type => int32
```
The address is a `Pointer` object that was allocated using `getmem_atomic(sizeof(int32_t))`.

The only time we might get a `Pointer` object with the address of something not allocated by us would be if a foreign function returns a pointer.  For example, `gmtime` in the example above returns a pointer to a `tm` struct that is statically allocated.

Sometimes, functions like this will let us first allocate the memory ourselves and pass the pointer as an argument, but not always.  I guess maybe providing some wrapper around `GC_REGISTER_FINALIZER` might work?

jkyang92:
> The only time we might get a `Pointer` object with the address of something not allocated by us would be if a foreign function returns a pointer. For example, `gmtime` in the example above returns a pointer to a `tm` struct that is statically allocated.

This is mostly correct, but with C libraries, even if you pass it memory, it often still has a "cleanup" function that needs to be called to deallocate related data, for example, if you use `gmp` without doing the reallocation magic that we do, then you would need to call `mpz_clear` at the end.

> Sometimes, functions like this will let us first allocate the memory ourselves and pass the pointer as an argument, but not always. I guess maybe providing some wrapper around `GC_REGISTER_FINALIZER` might work?

This will probably be enough, but a few wrappers to do reasonable things automatically would be nice. A potentially idea is an optional argument to the method that calls foreign functions that tells it how to interpret returned pointers. Also some care needs to be taken so if the object is manually deallocated, the finalizer is not called, but it might be reasonable to leave this to the person using the ffi interface and just provide them with a mechanism to deregister the finalizer.

Incidentally, some of the more friendly functionality can be provided in the M2 level once the basic primitive operations are created. So at some level, the first step is to make it possible to register finalizers.

d-torrance:
There's now a wrapper around `GC_REGISTER_FINALIZER` for memory management:

```m2
i2 : malloc = foreignFunction("malloc", voidstar, ulong)

o2 = malloc

o2 : ForeignFunction

i3 : free = foreignFunction("free", void, voidstar)

o3 = free

o3 : ForeignFunction

i4 : finalizer = x -> (print("finalizing " | net x); free x)

o4 = finalizer

o4 : FunctionClosure

i5 : for i to 9 do (x := malloc 8; registerFinalizer(x, finalizer))

i6 : collectGarbage()
finalizing 0x7f9084073d40
finalizing 0x7f9084073ce0
finalizing 0x7f908404b790
finalizing 0x7f908404b7b0
finalizing 0x7f9084073cc0
finalizing 0x7f9084073d60
finalizing 0x7f908404b7d0
finalizing 0x7f9084073d00
finalizing 0x7f9084073d20
```

DanGrayson:
Why would memory allocated with malloc get collected and finalized by libgc?

DanGrayson:
> Why would memory allocated with malloc get collected and finalized by libgc?

Ah, probably because the resulting pointer is stored in a new type of Expr, which itself is allocated by libgc.

d-torrance:
Had some time over the holiday weekend to finish documenting this package.  I think it's in pretty good shape now.

d-torrance:
The examples are failing on macOS at calls to `dlsym` without a library handle:

```m2
_register__Finalizer_lp__Foreign__Object_cm__Function_rp.errors
***************************************************************
-- -*- M2-comint -*- hash: -641194708

i1 : malloc = foreignFunction("malloc", voidstar, ulong)
stdio:1:10:(3): error: invalid handle passed to dlsym()
_foreign__Function.errors
*************************
o2 : ForeignFunction

i3 : mpfrVersion()

o3 = 4.1.0

o3 : ForeignObject of type char*

i4 : mpfrVersion = foreignFunction("mpfr_get_version", charstar, void)
stdio:4:15:(3): error: invalid handle passed to dlsym()
___Foreign__Functions.errors
****************************
-- -*- M2-comint -*- hash: -394951237

i1 : mycos = foreignFunction("cos", double, double)
stdio:1:9:(3): error: invalid handle passed to dlsym()
```
For some reason, I'd hardcoded `RTLD_DEFAULT` as `(void *)0` (which is what it is in Linux), but in macOS, it's `(void *)-2`.  I'll push a fix soon.

d-torrance:
Haven't fixed the `dlsym` issue yet, but just rebased on top of `development` to resolve the conflicts with the recently merged pull requests.

d-torrance:
Hopefully the latest commit fixes the issue on macOS.

We use the `RTLD_DEFAULT` macro and make sure that `use system` is at the top of the file so the `_GNU_SOURCE` macro is defined before we include `dlfcn.h` (otherwise `RTLD_DEFAULT` will be undefined).  Previously, `dlfcn.h` was getting included too early by one of the GC headers, which is apparently why I hardcoded `(void *)0` instead of using `RTLD_DEFAULT` when I first wrote that code a couple months ago.

d-torrance:
Just found a little bug with pointer arithmetic and pushed a fix.  (We were converting `ZZ` to `int`, which is 32 bits, instead of `long`, which will be the same size as a pointer, so 64 bits for most users.)

### Before 

```m2
i1 : needsPackage "ForeignFunctions"

o1 = ForeignFunctions

o1 : Package

i2 : address int 0

o2 = 0x7f50e5d1f570

o2 : Pointer

i3 : oo - 0x7f50e5d1f570

o3 = 0x7f5100000000

o3 : Pointer
```

### After
```m2
i1 : needsPackage "ForeignFunctions"

o1 = ForeignFunctions

o1 : Package

i2 : address int 0

o2 = 0x7fa596b03660

o2 : Pointer

i3 : oo - 0x7fa596b03660

o3 = (nil)

o3 : Pointer
```

d-torrance:
Converting back to a draft for the time being, as I've decided to make some pretty significant top-level changes (in particular, it would make sense if each instance of a `ForeignType` would actually be a class that a `ForeignObject` can be an instance of, much like `Ring` and `RingElement`).

d-torrance:
Oops, the builds are all failing.  I thought I'd fixed that issue before I pushed -- I'll take a look.

d-torrance:
Okay, should be good now.

PR #2528, opened 2022-06-14
Add -latomic flag for autotools TBB check

d-torrance:
TBB 2021 finally was uploaded to Debian unstable, and the [armel rebuild of Macaulay2 against it failed](https://buildd.debian.org/status/fetch.php?pkg=macaulay2&arch=armel&ver=1.20%2Bds-4%2Bb1&stamp=1655196258&raw=0):

```
checking for library containing TBB_runtime_interface_version... no
configure: error: tbb library not found
```
Here's the relevant lines in `config.log`:

```
configure:7795: g++ -o conftest -std=gnu++14 -g -ffile-prefix-map=/<<PKGBUILDDIR>>=. -fstack-protector-strong -Wformat -Werror=format-security -g3 -O2 -Wdate-time -D_FORTIFY_SOURCE=2 -DNDEBUG -Wl,-z,relro -Wl,-z,now -g3 conftest.cpp -ltbb  -ldl  >&5
/usr/bin/ld: /usr/lib/gcc/arm-linux-gnueabi/11/../../../arm-linux-gnueabi/libtbb.so: undefined reference to `__atomic_fetch_sub_8'
/usr/bin/ld: /usr/lib/gcc/arm-linux-gnueabi/11/../../../arm-linux-gnueabi/libtbb.so: undefined reference to `__atomic_load_8'
/usr/bin/ld: /usr/lib/gcc/arm-linux-gnueabi/11/../../../arm-linux-gnueabi/libtbb.so: undefined reference to `__atomic_fetch_add_8'
```
I think this is due to [GCC bug #104248](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=104248).  An easy fix is to add `-latomic` to `LIBS`.  See also https://github.com/Macaulay2/mathicgb/pull/41.

I also cleaned up a couple other things I noticed in the TBB check (removing an unused variable and moving the check later so `-ltbb` doesn't end up in the Fortran flags).

### DIFF
--- a/M2/configure.ac
+++ b/M2/configure.ac
@@ -398,26 +398,6 @@ AC_SEARCH_LIBS(hstrerror,resolv)
 AC_SEARCH_LIBS(dlopen,dl)
 AC_SEARCH_LIBS(gethostbyname,nsl)
 
-AC_SUBST(ENABLE_TBB,yes)  AC_ARG_ENABLE(tbb, AS_HELP_STRING(--disable-tbb,disable use of the tbb (threaded building blocks) library), ENABLE_TBB=$enableval)
-if test $ENABLE_TBB = yes
-then AC_DEFINE(WITH_TBB,1,[whether the tbb library has been enabled by the builder])
-     AC_LANG(C++)
-     AC_CHECK_HEADER(tbb/tbb.h,
-	[AC_SEARCH_LIBS(TBB_runtime_interface_version,"tbb",
-	   ,
-	   [AC_MSG_ERROR([tbb library not found])])],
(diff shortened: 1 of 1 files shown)


d-torrance:
Hmm, the Mac build is failing:

```
checking for library containing TBB_runtime_interface_version... -ltbb
checking for gdbm.h... yes
checking for library containing gdbm_close... no
configure: error: libgdbm not found
Error: Process completed with exit code 1.
```

d-torrance:
I'm going to close this, since it's really a TBB bug.  There's a [proposed fix ](https://salsa.debian.org/science-team/tbb/-/merge_requests/1) in the Debian package, which hopefully should solve the issue once it goes in.

PR #2531, opened 2022-06-16
utf8 width for strings, nets

pzinn:
This is an attempt to better handle strings with utf8 content; cf the discussion in #1069.
A priori, there are 3 distinct concepts associated to a `String`:
(1) its length -- its number of bytes
(2) its utf8 length -- its number of utf8 characters
(3) its (utf8) width -- its width in a monospace font
Most characters have width 1, so in the first approximation one can assume (2)=(3).
Right now M2 defines two methods for strings, `length` and `width`. Both point to `#` which computes its length (1).
- This PR implements (2) and assigns it to `width`. It *does not* distinguish between (2) and (3) (e.g. for wide chinese characters), this is left for future work.
- It also fixes various methods for nets, most notably `horizontalJoin`.
- It introduces (and exports) a new method `utf8substring`.
- The other thing it does not do is fix `wrap` since its current implementation is very much byte-based.
- Since redefining `width` is a potentially breaking change, one should carefully check its repercussions.

Example:
```
i1 : R=QQ[α,β]

o1 = R

o1 : PolynomialRing

i2 : 1/(α^2+1)

        1
o2 = ------
      2
     α  + 1

o2 : frac R

i3 : I=ideal(α^2+2*α*β,β^2+1)

             2          2
o3 = ideal (α  + 2α*β, β  + 1)

o3 : Ideal of R

i4 : netList I_*

     +---------+
     | 2       |
o4 = |α  + 2α*β|
     +---------+
     | 2       |
     |β  + 1   |
     +---------+

i5 : wrap(5,"abcdeαβγδϵ12345") -- still not right

o5 = abcde
     αβγδ
     ϵ123
     45
```

### DIFF
--- a/M2/Macaulay2/d/actors5.d
+++ b/M2/Macaulay2/d/actors5.d
@@ -482,6 +482,26 @@ numparms(e:Expr):Expr := (
      else WrongArg("a function"));
 setupfun("numparms",numparms);
 
+utf8substrfun(e:Expr):Expr := (
+     when e is args:Sequence do
+	  if length(args) != 3 then WrongNumArgs(3) else
+	       when args.0 is s:stringCell do
+		   when args.1 is start:ZZcell do
+		       when args.2 is width:ZZcell do
+		       toExpr(utf8substr(s.v,toInt(start),toInt(width)))
+		       else WrongArgSmallInteger(3)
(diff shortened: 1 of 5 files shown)


pzinn:
Documentation will come next, I wanted to run the tests first.

Issue #2533, opened 2022-06-27
timing-quotient test often fails on amd64


ginggs:
Similar to #1804, the timing-quotient test often fails in [Ubuntu's amd64 autopkgtests](https://autopkgtest.ubuntu.com/packages/macaulay2/kinetic/amd64).  I don't see this on the other architectures.  Ubuntu's autopkgtests are run on VMs with 1 vCPU and 1592MiB RAM by default.

Below are results of recent failures:
```
 --      assert BinaryOperation {symbol <, tim#0, standardSecond}
 -- stdio:225:1:(3): error: assertion failed:
 -- .39568<.348378 is false
 -- .37496<.345403 is false
 -- .491385<.39548 is false
 -- .365218<.344969 is false
 -- 2.74432<2.25892 is false
 -- .361029<.341223 is false
 -- .384533<.346314 is false
 -- .37798<.347436 is false
 -- .371521<.34619 is false
 -- .389059<.347442 is false
 ```
I'm happy to run tests to help find a value that works.

PR #2535, opened 2022-06-28
Increase upper bound of timing-quotient test

ginggs:
Closes: #2533


d-torrance:
Would it help if I upload a new package to Debian with this patch so we don't have to wait for the next release to get this fix for the Ubuntu autopkgtests?

ginggs:
@d-torrance that would be great, thanks!

d-torrance:
> @d-torrance that would be great, thanks!

Done! Just uploaded macaulay2 1.20+ds-7, which includes this patch, to unstable.

Issue #2513, opened 2022-05-23
Invalid vector access when building CotangentSchubert examples


jamesjer:
I did a test build of Macaulay 1.20 for Fedora.  The build failed here:
```
 -- making example result files in ../../../usr-dist/common/share/doc/Macaulay2/CotangentSchubert/example-output/
 -- making example results for "basisCoeffs"                                
 ulimit -c unlimited; ulimit -t 700; ulimit -m 850000; ulimit -s 8192; ulimit -n 512;  cd /tmp/M2-196276-0/1-rundir/; GC_MAXIMUM_HEAP_SIZE=400M "/builddir/build/BUILD/M2-release-1.20/M2/BUILD/x86_64-redhat-linux-gnu/usr-dist/x86_64-Linux-Fedora-37/bin/M2-binary" -q --int --no-randomize --no-readline --silent --stop --print-width 77 -e 'needsPackage("CotangentSchubert",Reload=>true,FileName=>"/builddir/build/BUILD/M2-release-1.20/M2/Macaulay2/packages/CotangentSchubert.m2")' <"/tmp/M2-196276-0/0_basis__Coeffs.m2" >>"/builddir/build/BUILD/M2-release-1.20/M2/BUILD/x86_64-redhat-linux-gnu/usr-dist/common/share/doc/Macaulay2/CotangentSchubert/example-output/_basis__Coeffs.errors" 2>&1
/builddir/build/BUILD/M2-release-1.20/M2/BUILD/x86_64-redhat-linux-gnu/usr-dist/common/share/doc/Macaulay2/CotangentSchubert/example-output/_basis__Coeffs.errors:0:1: (output file) error: Macaulay2 killed by signal 6
/tmp/M2-196276-0/0_basis__Coeffs.m2:0:1: (input file)
M2: *** Error 6
 -- 1.12171 seconds elapsed
 -- storing example results in ../../../usr-dist/common/share/doc/Macaulay2/CotangentSchubert/example-output/_basis__Coeffs.out
 -- warning: missing file /builddir/build/BUILD/M2-release-1.20/M2/BUILD/x86_64-redhat-linux-gnu/usr-dist/common/share/doc/Macaulay2/CotangentSchubert/example-output/_basis__Coeffs.out
 -- making example results for "segreClass"                                  -- 0.470983 seconds elapsed
 -- storing example results in ../../../usr-dist/common/share/doc/Macaulay2/CotangentSchubert/example-output/_segre__Class.out
[...]

PR #2536, opened 2022-07-01
Fix a crash in basis for degrees of length 0

jkyang92:
I changed Monoid::set_degrees to add the right number of entries to the degree_of_var_ list even if degvars is zero. This should fix the crash in #2513.


mikestillman:
@jkyang92 Thanks for this fix!  Can you also add in a test that shows this working correctly now?

PR #2541, opened 2022-07-11
utf8 characters, utf8 wrap, utf8 doc

pzinn:
This is a continuation of #2531 -- the goal is to improve handling of strings with utf8 characters.
- `characters` has been rewritten so it now returns the list of utf8 characters, not the list of bytes as characters. The old method can still be accessed with `toList`, though it's hard to imagine any circumstance when one would want to do so (it's better, if the actual bytes are needed, to use `ascii`; also note that the output of `toList` on a string with utf8 characters will fail `validate` / `utf8check`). In particular one may wonder if iterating over strings (#1897) should be over the bytes or the utf8 characters @d-torrance... just a thought.
- `wrap` has been rewritten so it correctly treats utf8 characters.
- documentation has been added.

### DIFF
--- a/M2/Macaulay2/d/actors5.d
+++ b/M2/Macaulay2/d/actors5.d
@@ -496,6 +496,14 @@ utf8substrfun(e:Expr):Expr := (
 );
 setupfun("utf8substring",utf8substrfun);
 
+utf8charactersfun(e:Expr):Expr := (
+     when e
+     is s:stringCell do toExpr(utf8characters(s.v))
+     else WrongArg("a string"));
+setupfun("characters",utf8charactersfun);
+
+
+
 stringWidth(e:Expr):Expr := (
      when e
      is s:stringCell do toExpr(utf8width(s.v))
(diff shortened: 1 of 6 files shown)


d-torrance:
> In particular one may wonder if iterating over strings ([Iterating over strings #1897](https://github.com/Macaulay2/M2/pull/1897)) should be over the bytes or the utf8 characters @d-torrance... just a thought.

I suppose there's a chance that either case might be desirable depending on the circumstances.  With this PR, would the following work?

```m2
i1 : for c in "🎉🎉🎉" do print c
�
�
�
�
�
�
�
�
�
�
�
�

i2 : for c in characters "🎉🎉🎉" do print c
🎉
🎉
🎉
```

d-torrance:
Ok, cool!  That seems like reasonable behavior.

DanGrayson:
> * `wrap` has been rewritten so it correctly treats utf8 characters.

Thanks for the contribution.  How does `wrap` determine the font width of a utf8 character?

Screen Shot 2022-07-12 at 2 03 29 PM

pzinn:
We've had this debate before, cf #1069. As mentioned in #2531, for now it is assumed that every unicode character has (in a monospace font) width 1; this is the case of most characters. Of course this will fail if you don't use a monospace font, or if your renderer (emacs, browser) substitutes missing characters from another font. But then there's not much one can do anyway.

PR #2553, opened 2022-08-10
Build and deploy engine documentation to GitHub Pages

d-torrance:
After bringing this up in M2internals this afternoon, I figured I'd mess around with it.  This workflow builds the doxygen documentation for the engine every Sunday at midnight UTC (or at least it should -- I haven't tested the cron part...), and then publishes it to GitHub pages.  See for example [this copy](https://www.friedcheese.org/M2) that was built and deployed from my fork.

There would need to be some setup from someone with the proper privileges to set up `gh-pages` as the branch that gets published (see https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-from-a-branch).  The URL should end up being something like "https://Macaulay2.github.io/M2".

Note that I used `make tinydoc` since `fulldoc` and `smalldoc` take a lot longer (graphviz is run thousands of times).

### DIFF
--- a/.github/workflows/deploy-doxygen.yml
+++ b/.github/workflows/deploy-doxygen.yml
@@ -1,0 +1,34 @@
+name: Build and deploy engine documentation
+on:
+  schedule:
+    - cron: '0 0 * * 0'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install dependencies
+        run: |
+          sudo add-apt-repository -y -n ppa:macaulay2/macaulay2
+          sudo apt-get update
+          sudo apt-get install -y doxygen libtool-bin libtbb-dev \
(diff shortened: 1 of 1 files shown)


d-torrance:
If all goes well, we should see a `gh-pages` branch appear the first Sunday morning after this is merged into `master`.

DanGrayson:
Okay, I've just merged it into `pre-master`.  If the tests succeed, I'll merge it into `master` later today.

d-torrance:
The build [worked](https://github.com/Macaulay2/M2/runs/7823071864?check_suite_focus=true), and now the [gh-pages](https://github.com/Macaulay2/M2/tree/gh-pages) branch exists with the engine documentation!  It should be ready to set up GitHub Pages now (https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-from-a-branch).

PR #2557, opened 2022-08-15
fix typos in package SchurVeronese

fchapoton:


mahrud:
Since all the changes are in comments or strings, if you add `[skip ci]` somewhere in the commit message, Github skips running all the checks, which saves resources.

fchapoton:
Thanks for the suggestion. I will try to remember about that. Right now, it appears that the package `VersalDeformations` is broken (and not by any of my changes).

PR #2565, opened 2022-08-16
Many typos again

fchapoton:
this time, they were found using command like
```
egrep -R -w --no-filename -o '[a-z]+ism' . | sort | uniq -c | sort -rn
```


DanGrayson:
Thanks!  So many misspelled words!

PR #2570, opened 2022-08-18
fix some typos for words s* [skip ci]

fchapoton:
once more ; please tell me if I should rather stop..

:signal_strength:


DanGrayson:
No, don't stop -- fixing these misprints is totally worthwhile!

Issue #2262, opened 2021-10-21
Use of == in Graphs


pzinn:
Example of a problem with the `Graphs` package:
```
i1 : needsPackage "Graphs"

o1 = Graphs

o1 : Package

i2 : G=graph{{a,1}};

i3 : H=graph{{b,2}};

i4 : cartesianProduct(G,H)
stdio:4:1:(3): error: no method for binary operator == applied to objects:
--            1 (of class ZZ)
--     ==     a (of class Symbol)
```
The issue is, `cartesianProduct` tries to compare vertex labels with `==`. This is problematic since vertex labels can be anything (in the help, both symbols like a,b,c and numbers 1,2,3 are used).
I suspect `===` should be used. There are I believe quite a few other functions defined in `Graphs` that suffer from the same problem.

fchapoton:
This may be because the vertex set is build using
{{{
    V := toList(set vertexSet G ** set vertexSet H);
}}}
Could someone propose another way to build the cartesian product of two `vertexSet` ?

PR #2576, opened 2022-08-22
trying to enhance graph products

fchapoton:
so that they allow more general labels, as required by #2262


fchapoton:
Well, this seems to (sometimes) trigger a failure in Chordal Graphs.

DanGrayson:
The failure is here:
```m2
G = cartesianProduct(cycleGraph 3, pathGraph 5)
I = edgeIdeal G
N = chordalNet(toLex I)
assert(treewidth elimTree N == 6)
```

Could it be that treewidth depends on the order the vertices of a graph are listed in?  (We get 10 instead of 6.)

fchapoton:
This in turn is using
```
treewidth(ChordalGraph):= G -> 
    max(for v in G.vertexSet list #children(G,v))
```

DanGrayson:
Well, that looks independent of the order of vertices.  Hmm.

fchapoton:
It seems to me that the result could depend on the ordering of the vertices of the graph. So it could be normal that the tree-width is changed. Indeed the doctest just below find another tree-width.

Maybe one should ask the authors of the package if it is ok to change the doctest ?

DanGrayson:
The definition of treewidth at https://en.wikipedia.org/wiki/Chordal_graph makes it clear that the ordering of the vertices is not relevant.

fchapoton:
Hmm. Trying inside sagemathcell (so M2 version 1.18) to change the labels in the example, I got

```
loadPackage "Graphs";loadPackage "Chordal";
C = graph({{0,22},{22,1},{1,0}},EntryMode=>"edges");
G = cartesianProduct(C, pathGraph 5)
I = edgeIdeal G
N = chordalNet(toLex I)
treewidth elimTree N
```
and this gives 10.

fchapoton:
Either there is a bug, or we do not understand something in the math. I do not claim to understand if treewidth depends on the order of vertices or not. Shall we ask the package authors ?

DanGrayson:
Yes, let's.

@jburkar1 , @dcookii , @carolinejansen , @abokeefe ?

fchapoton:
ping @jburkar1 , @dcookii , @carolinejansen , @abokeefe

We need your help in understanding a failing doctest, please.

DanGrayson:
Maybe we should try emailing them directly.

DanGrayson:
Okay, I've sent email to the authors of Graphs and of Chordal.

DanGrayson:
Frédéric, if we could demonstrate that there is a bug in `treewidth` outside of this pull request, it would make a good issue and would give the authors something to focus on.

fchapoton:
Still not clear to me what goes wrong. Here is another tentative

```
loadPackage "Graphs";loadPackage "Chordal";loadPackage "NautyGraphs";
C1 = graph({{0,22},{22,1},{1,0}},EntryMode=>"edges");
G1 = cartesianProduct(C1, pathGraph 5);
I1 = edgeIdeal G1;
N1 = chordalNet(toLex I1);
treewidth elimTree N1
C2 = graph({{0,2},{2,1},{1,0}},EntryMode=>"edges");
G2 = cartesianProduct(C2, pathGraph 5);
I2 = edgeIdeal G2;
N2 = chordalNet(toLex I2);
treewidth elimTree N2
areIsomorphic(G1, G2)
N3 = chordalNet(toLex I2,"SuggestOrder");
chordalTria N1;
chordalTria N2;
chordalTria N3;
codimCount(N1)
codimCount(N2)
codimCount(N3)
```

which gives three different results.

fchapoton:
Here is a smaller example, where we can see explicitly that both edge ideals are isomorphic

```
loadPackage "Graphs";loadPackage "Chordal";loadPackage "NautyGraphs";
C1 = graph({{0,22},{22,1},{1,0}},EntryMode=>"edges");
G1 = cartesianProduct(C1, pathGraph 2);
I1 = edgeIdeal G1
N1 = chordalNet(toLex I1);
treewidth elimTree N1
C2 = graph({{0,2},{2,1},{1,0}},EntryMode=>"edges");
G2 = cartesianProduct(C2, pathGraph 2);
I2 = edgeIdeal G2
N2 = chordalNet(toLex I2);
treewidth elimTree N2
```

fchapoton:
in the previous simplified example, the elim trees look really different.

fchapoton:
Here is an even more simplified case
```
loadPackage "Graphs";loadPackage "Chordal";loadPackage "NautyGraphs";
C1 = graph({{0,22},{1,0}},EntryMode=>"edges");
pt = graph({{0,{1}}});
G1 = cartesianProduct(C1, pt);
I1 = edgeIdeal G1
N1 = chordalNet(toLex I1);
treewidth elimTree N1
C2 = graph({{0,2},{1,0}},EntryMode=>"edges");
G2 = cartesianProduct(C2, pt);
I2 = edgeIdeal G2
N2 = chordalNet(toLex I2);
treewidth elimTree N2
```

fchapoton:
Very strangely, the failure depends on the value of the third vertex in C1 ; it fails for 22 and 12, works for -6,  10 and 11 and fails differently for -1..

DanGrayson:
Why do we see no test results?

fchapoton:
Do you mean "please show us the results" ?

```
loadPackage "Graphs";loadPackage "Chordal";loadPackage "NautyGraphs";

testtreewidth := n -> (
    C1 = graph({{0,n},{1,0}},EntryMode=>"edges");
    pt = graph({{0,{1}}});
    G1 = cartesianProduct(C1, pt);
    I1 = edgeIdeal G1;
    N1 = chordalNet(toLex I1);
    treewidth elimTree N1
    )

for i from -24 to 24 list testtreewidth i
```

giving

```
        {3, 3,
    4, 4, 3, 4, 3, 2, 3, 3, 
    4, 4, 3, 4, 3, 2, 3, 3,
    4, 4, 3, 4, 3, 2, 2, 2,
    4, 4, 3, 4, 3, 2, 3, 3, 
    4, 4, 3, 4, 3, 2, 3, 3, 
    4, 4, 3, 4, 3, 2, 3}
```

fchapoton:
This strange behaviour depends on the label in the second factor graph:

loadPackage "Graphs";loadPackage "Chordal";loadPackage "NautyGraphs";

```
testtreewidth := (n, m) -> (
    C1 = graph({{0,n},{1,0}},EntryMode=>"edges");
    pt = graph({{0,{m}}});
    G1 = cartesianProduct(C1, pt);
    I1 = edgeIdeal G1;
    N1 = chordalNet(toLex I1);
    treewidth elimTree N1
    )

for i from -24 to 24 list testtreewidth (i, 1)

for i from -24 to 24 list testtreewidth (i, 2)
```
giving

```
{3, 3, 4, 4, 3, 4, 3, 2, 3, 3, 4, 4, 3, 4, 3, 2, 3, 3, 4, 4, 3, 4, 3, 2, 2, 2, 4, 4, 3, 4, 3, 2, 3, 3, 4, 4, 3, 4, 3, 2, 3, 3, 4, 4, 3, 4, 3, 2, 3}

{3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 2, 2, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3}
```

d-torrance:
@fchapoton - What's the status of this pull request?

fchapoton:
well, still not resolved, despite I made some progress about what is happening.

Issue #2586, opened 2022-08-31
Serve engine docs (doxygen) on macaulay2.com


mahrud:
The gh-pages branch added after #2553 has added more than 200MB to the size of the repository, and every week nearly every file in the branch is updated, so I think it should be removed from this repository.

Having the engine documentation online sounds useful, but there's no need for tracking the changes across thousands of auto-generated html files via git, so I think it should be moved out of git and onto the server hosting macaulay2.com.

cc: @d-torrance

d-torrance:
Yikes -- I didn't realize how big that branch was!  I've opened #2588 to remove the workflow.  Someone with the proper permissions will need to delete the branch itself.

mahrud:
It still remains to:
- [ ] remove the existing `gh-pages` branch
- [ ] generate and serve the doxygen documentation on the macaulay2.com server

d-torrance:
FYI, `gh-pages` is back since the workflow was run again on Sunday.  This should stop once #2588 is merged into `master`.

mahrud:
Can we still have this set up somewhere?

d-torrance:
One option would be to create a new repository under the Macaulay2 organization that consists of a single workflow that clones the main Macaulay2 repository every week or month or whatever, generates the engine documentation, and publishes it to GitHub pages.  Then it wouldn't muddy up the main repository.

d-torrance:
Here's a draft: https://github.com/d-torrance/M2-engine-docs

mahrud:
My preference would be to not use git at all, especially given that doxygen is not git friendly (updates every single file every time) and that Dan already has a server with storage and nightly tests that this can easily be incorporated into. But I'm not the one doing the work so it's not up to me.

d-torrance:
> **Written by Claude** (Claude Opus 5, via Claude Code), not by @d-torrance, whose account posted it -- please weigh it accordingly.

Both halves of this are in place, so closing.

The branch is gone — `git ls-remote origin refs/heads/gh-pages` comes back empty, so the 200 MB of
weekly-rewritten HTML is no longer in this repository. That was the actual complaint here.

And the engine documentation is online and maintained: the draft linked above,
[`d-torrance/M2-engine-docs`](https://github.com/d-torrance/M2-engine-docs), publishes to
<https://d-torrance.github.io/M2-engine-docs/>, which is live and was last built three days ago.

Being straight about what that is and isn't: it is GitHub Pages under a personal account, not
[...]

PR #2588, opened 2022-08-31
Remove engine documentation workflow

d-torrance:
It was creating a massive branch inside the repository.  Partially addresses #2586.


DanGrayson:
I get variable results when I try to measure the size of the repository now: 806MB on Mac OS, 180MB on Debian.

DanGrayson:
> I get variable results when I try to measure the size of the repository now: 806MB on Mac OS, 180MB on Debian.

Never mind -- I think it had to do with multiple remotes.   Now I get 139MB on both.

If there really is bloat in the repository, how will we get rid of it?

d-torrance:
`git push origin :gh-pages` should do the trick

DanGrayson:
According to the list at https://github.com/Macaulay2/M2/branches/all the branch named gh-pages doesn't exist.

PR #2592, opened 2022-09-02
Make "abs" overloadable

d-torrance:
We make `abs` overloadable and take the opportunity to implement `abs(Constant)`.

## Before
```m2
i1 : abs Constant := abs @@ numeric

o1 = -*Function[/usr/share/Macaulay2/Core/classes.m2:90:48-90:55]*-

o1 : FunctionClosure

i2 : abs pi
stdio:2:1:(3): error: expected a number, real or complex
```

## After

```m2
i1 : abs pi

o1 = 3.14159265358979
```

### DIFF
--- a/M2/Macaulay2/d/actors4.d
+++ b/M2/Macaulay2/d/actors4.d
@@ -96,13 +96,13 @@ setupfun("setGroupID",setpgidfun);
 
 absfun(e:Expr):Expr := (
      when e
-     is i:ZZcell do toExpr(abs(i.v))				    -- # typical value: abs, ZZ, ZZ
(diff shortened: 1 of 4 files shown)


d-torrance:
Oops, I missed this test:

```m2
testing: ../../../../../Macaulay2/tests/normal/000-core.m2
ii313 : 
        
        --
        assert ( class (x->x) === FunctionClosure )

ii314 : assert ( class abs === CompiledFunction )
../../../../../Macaulay2/tests/normal/000-core.m2:452:1:(3):[5]: error: assertion failed
```
Just pushed a new commit swapping `abs` out with `any`, which is a `CompiledFunction`.

PR #2593, opened 2022-09-02
Add bitwise not operator

d-torrance:
I noticed that despite having the other big bitwise operators (`&`, `|`, and `^^`), we were missing bitwise not.  Most other languages use `~` for this, which is a postfix operator in Macaulay2.  It may look a little strange, but it works.

So for example:

## Python

```python
>>> ~1234
-1235
>>> ~-1234
1233
```
## Macaulay2

```m2
i1 : 1234~

o1 = -1235

i2 : (-1234)~

o2 = 1233
```

### DIFF
--- a/M2/Macaulay2/d/actors3.d
+++ b/M2/Macaulay2/d/actors3.d
@@ -698,6 +698,13 @@ bitxorfun(e:Expr):Expr := (
      else WrongNumArgs(2)
      else WrongNumArgs(2));
 setupfun("bitxorfun",bitxorfun);
+
+bitnotfun(e:Expr):Expr := (
+    when e
+    is x:ZZcell do toExpr(not(x.v))
+    else WrongArgZZ());
+setupfun("bitnotfun", bitnotfun);
+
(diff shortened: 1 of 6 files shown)


mahrud:
I don't understand. Doesn't bit-wise not depend on the bit-size of the argument? How does this make sense for arbitrary precision integers?

d-torrance:
> I don't understand. Doesn't bit-wise not depend on the bit-size of the argument? How does this make sense for arbitrary precision integers?

It's a wrapper around GMP's `mpz_com`.  From the [documentation](https://gmplib.org/manual/Integer-Logic-and-Bit-Fiddling): "These functions behave as if twos complement arithmetic were used (although sign-magnitude is the actual implementation). The least significant bit is number 0."

Regardless of the number of bits, computing the ones' complement of an integer `n` encoded using two's complement will always give us `-(n + 1)`, so that's what happens.

Python has the same behavior for its `~` operator, and it also uses arbitrary precision integers.

PR #2594, opened 2022-09-03
Add truncate method for real numbers

d-torrance:
This rounds a real number toward 0:

```m2
i1 : truncate 3.5

o1 = 3

i2 : truncate(-3.5)

o2 = -3
```


d-torrance:
> It would be good if the documentation for floor, ceiling, and these new methods for truncate would refer to each other in the "see also" section.

Done!

Doing this also made me realize that it would make sense for `truncate` to work more like `floor` and `ceiling`, i.e., be defined for all `Number` objects.  So now it supports `CC` and `RRi` as well:

```m2
i1 : truncate (2.5 + ii)

o1 = 2

i2 : truncate interval (-3.7, -1.5)

o2 = -1
```

It raises an error for intervals where truncation doesn't make sense, i.e., it can't decide whether to use `floor` or `ceiling` because 0 is an interior point:

```m2
i3 : truncate interval (-2, 5)
stdio:3:1:(3): error: cannot truncate
```
(Or maybe this should just return `0` in this case)?

DanGrayson:
Probably it should just return 0 in that case.

d-torrance:
> Probably it should just return 0 in that case.

Yeah, that makes sense. I've just force-pushed this change.

PR #2597, opened 2022-09-05
A few typos

fchapoton:
and that's it


DanGrayson:
Thanks!

When you say "and that's it", does that mean there are no more typos?

PR #2600, opened 2022-09-06
Simplify monoid constructor and fix various bugs

mahrud:
- renamed and reorganizd ofcm.m2 -> monoids.m2
- extended several methods to Monoid type
- improved printing for Monoids, added an AfterPrint
- simplified findHeft slightly
- simplified argument processing for monoid
- simplified the monoid constructor
- fixed degreeLength degreesMonoid 0
- modified the polynomial ring constructor slightly
- rewrote documentation of monoid
- updated documentation for Monoid, degrees*


mahrud:
It's probably easiest to review the very first commit independently via this link:
https://github.com/Macaulay2/M2/pull/2600/commits/ee135ebf24d50b154a71fda3ea71cbb5039d68fe#diff-0cd9e58800181c6fbd348538e174c7fe0aedc2467ea9afba3f2091627deeba9c
Which shows that the changes are not so scary:
image

M2/Macaulay2/m2/document.m2
```diff
@@ -418,7 +418,8 @@ processSignature := (tag, fn) -> item -> (
     else if fn =!= null then fn else key;
 
     -- checking for various pieces of the synopsis item
-    isVariable   := y -> match(///\`[[:alnum:]']+\'///, y);
+    -- allow either a variable name or a visible list for rare exceptions
+    isVariable   := y -> match(///\`[[:alnum:]']+\'///, y) or match(///(\".+\")|(\(.+\))|(\[.+\])|(\{.+\})|(<\|.+\|>)///, y);
```
DanGrayson, line 422: What's the purpose of recognizing those other strings as variables?

M2/Macaulay2/m2/indeterminates.m2
```diff
@@ -90,6 +90,12 @@ runLengthEncode0 = x -> (
 		    (dupin = null; oi = i; m = 1))));
      x)
 
+rle = method(Dispatch => Thing)
```
DanGrayson, line 93: What is this function supposed to do, and why is it needed?

mahrud: It's not something I've added, it used to be there before, I just moved it. I think it's subtly different from `runLengthEncode`.

M2/Macaulay2/m2/document.m2
mahrud: It's so that `SimpleDoc` can process something like `[a,b,c,...]` under `Inputs` here:
https://github.com/Macaulay2/M2/blob/3a7b5bfa27cc0f6631b7868128042d2950e87e2a/M2/Macaulay2/packages/Macaulay2Doc/functions/monoid-doc.m2#L23-L27

PR #2601, opened 2022-09-06
more typos [skip ci]

fchapoton:
I am afraid typos are very much infinite in number.


mahrud:
If each contribution by someone introduces a few new typos then definitely!

PR #2613, opened 2022-09-15
Introduce iterators

d-torrance:
We introduce [iterators](https://en.wikipedia.org/wiki/Iterator_pattern), very heavily inspired by Python's implementation.  In particular, if a class has an installed method function for `iterator`, then it's iterable, and if it has method functions for both `iterator` and `next`, then it's an iterator.  If `next` eventually returns the new `StopIteration` symbol, then we can pass an iterable to `scan`, `apply`, and `toList`.  For example:

```m2
i1 : BoundedPrimeIterator = new SelfInitializingType of MutableHashTable;

i2 : iterator BoundedPrimeIterator := identity;

i3 : next BoundedPrimeIterator := i -> (
         q := nextPrime(i#"val" + 1);
         if q > i#"max" then StopIteration else i#"val" = q);

i4 : BoundedPrimeIterable = new SelfInitializingType of HashTable;

i5 : iterator BoundedPrimeIterable := x -> BoundedPrimeIterator {
         "val" => 1, "max" => x#"max"};

i6 : P = BoundedPrimeIterable {"max" => 20};

i7 : scan(P, print)
2
3
5
7
11
13
17
19

i8 : apply(P, x -> x + 1)

o8 = {3, 4, 6, 8, 12, 14, 18, 20}

o8 : List

i9 : toList P

o9 = {2, 3, 5, 7, 11, 13, 17, 19}

o9 : List
```
As a proof of concept, the `Python` package now uses this feature.

```m2
i10 : needsPackage "Python";

i11 : apply(toPython "Hello, world!", identity)

o11 = {H, e, l, l, o, ,,  , w, o, r, l, d, !}

o11 : List
```
This doesn't completely address #1904, since it's almost entirely done at top level, and so there's no support for `for` loops or anything like `yield` to create generators, but it's a start!


d-torrance:
A happy accident: adding `toList(Thing)`, intended to be used on iterables, also makes the following work, which is kind of nice:

```m2
i1 : toList 5

o1 = {0, 1, 2, 3, 4}
```

M2/Macaulay2/d/actors3.d
```diff
@@ -1647,7 +1647,7 @@ map(e:Expr):Expr := (
 	  then map(a.0,a.1,a.2)
 	  else WrongNumArgs(2,3))
      else WrongNumArgs(2,3));
-setupfun("apply",map);
+setupfun("apply",map).Protected = false;
```
mahrud, line 1650: Given how inefficient methods are implemented in the top level, I would be very hesitant about this change. I'd rather `scan` and `apply` remained low level functions until methods are made more efficient.

mahrud: Perhaps a new `iterate` method can play that role for now.

d-torrance: Good point.  I'll play around with this some more.  Perhaps I can move the `scan` and `apply` code for iterators to the interpreter for the best of both worlds.

d-torrance:
I've moved most of the code to the interpreter so than `scan` and `apply` remain compiled functions.  As a consequence, it was pretty easy to add support for `for` loops after that!

```m2
i1 : RandomIterator = new Type of HashTable;

i2 : iterator RandomIterator := identity;

i3 : next RandomIterator := x -> (
         y := random 20;
         if y == 0 then StopIteration else y);

i4 : rand = new RandomIterator;

i5 : for x in rand list x

o5 = {18, 12, 2, 8, 3, 15, 6, 11, 4, 10, 19, 12, 7, 12, 5, 16, 8, 14, 5, 1, 17,
     4}

o6 : List
```

d-torrance:
Oops, I broke `for` loops:

```m2
i1 : for i when i < 15 list i^2
stdio:1:14:(3): error: expected a function
```

Converting to a draft while I investigate.

d-torrance:
Figured it out -- any `for` loops that didn't use the `in` or `to` keywords thought they were dealing with an iterator.  Should be fixed now.

d-torrance:
Question:

Currently, iterators don't work very well when they inherit from some kind of list, e.g.,:

```m2
i1 : BoundedPrimeIterator = new SelfInitializingType of MutableList;

i2 : iterator BoundedPrimeIterator := identity;

i3 : next BoundedPrimeIterator := x -> (
         x#0 = nextPrime(x#0 + 1);
         if x#0 > x#1 then StopIteration
         else x#0);

i4 : P = BoundedPrimeIterator {1, 6};

i5 : next P

o5 = 2

i6 : next P

o6 = 3

i7 : next P

o7 = 5

i8 : next P

o8 = StopIteration

o8 : Symbol
```

So far, so good.  However:
```m2
i9 : P = BoundedPrimeIterator {1, 6};

i10 : scan(P, print)
1
6
```

Of course, this makes total sense, since `P` already is a list!

Here's my question:  Would it be desirable to allow subclasses of `BasicList` be iterable and compatible with `scan`, `apply`, and `for`?  This would require moving the `getIterator` check up to the top of this sequence of conditionals:
https://github.com/Macaulay2/M2/blob/55164055c923bb7702566d395c2fff1a9367198b/M2/Macaulay2/d/actors3.d#L2099-L2111

So we'd be checking the class of every object and seeing if it has an `iterator` key, even if it is a traditional list, sequence, or string.  That might be too expensive to make it worthwhile.

Thoughts?

mahrud:
Couldn't there be a use-case when someone actually needs to loop through the entries in an iterator inherited from a basic list though? For instance, I might want to scale the range of an iterator:
```m2
i1 : X = new SelfInitializingType of List;

i2 : apply(X{1, 3}, times_10)

o2 = {10, 30}

o2 : X
```
(Note that `apply` returns an object of type `X` again)

mahrud:
I haven't looked at your implementation carefully yet, but I guess a basic question I have is why did you go with "an iterator type is a type that has a `next` method" route rather than "an iterator is a type that inherits from an `Iterator` type" route?

d-torrance:
> Couldn't there be a use-case when someone actually needs to loop through the entries in an iterator inherited from a basic list though?

Under the alternate proposal, if we *don't* define an `iterator` method, then `apply` and friends would fall through to the `is List` case and run as normal.  But if we do, then we could make it behave however we want.

I'm thinking of classes that are subclasses of `BasicList` just because we needed to hold one or two objects for later reference, not because we think of them as some kind of list.  One example I have in mind is in the new `ForeignFunctions` package, there's a `ForeignArray` class that inherits from `BasicList` because it's a list containing one item -- a `Pointer` object.  It would make a lot of sense to iterate over a `ForeignArray` to get the various foreign objects it contains, but with the current behavior, we'd just get that single `Pointer` object and be done with it.

> I haven't looked at your implementation carefully yet, but I guess a basic question I have is why did you go with "an iterator type is a type that has a `next` method" route rather than "an iterator is a type that inherits from an `Iterator` type" route? (in particular, `Iterator` being a new basic type defined in the interpreter)

This was the "very heavily inspired by Python" part.  :)  (In Python, any class with an `__iter__` method is iterable, and any class with `__iter__` (usually just `return self`) and `__next__` methods is an iterator.)

This way, we can make any existing (currently non-list) class iterable by defining these methods and iteration will just work.  Case in point, this PR makes `PythonObject` iterable by just changing a few lines of code.  `PythonObject` is already a class defined in the interpreter, so trying to make it a subclass of some other new class would have been a lot more work.

d-torrance:
After @jkyang92's comments from M2internals yesterday, I've dropped support for `apply` for iterators from this PR.  Instead, it would make a lot of sense if `apply` (and `select`) returned iterators instead of lists.  But first, we'd need to have some interpreter-level class to which these things could belong.  Perhaps the same class as the `yield`-statement generators proposed in #1904?

I've also dropped `toList(ZZ)`, since it was just a consequence of defining `toList(Thing)` using `apply` (now it uses `for`).  I may re-propose it in a separate PR.

mahrud:
> Instead, it would make a lot of sense if `apply` (and `select`) returned iterators instead of lists.

This would be great, and I suspect it's easier for `select` because the return type and input type are more or less the same. Probably a generic iterator class to be used as output makes sense.

Along a different direction: I'm always subtly confused about how `SelfInitializingType` methods are applied to types of that kind, but can something like that be used for iterators? So for instance iterable objects would be defined like
```m2
T = new IterableType of Type
```
in such a way that `apply` returns a generic iterator object if the input has `IterableType` as parent.

mahrud:
I'm confused. Was this pull request ready to be merged? I thought we were still discussing the behavior.

DanGrayson:
Well, the tests passed, and it wasn't a draft.  I was under the impression that the discussion concerns additional features that could be added, either here, or in a future pull request.

d-torrance:
The big question I had was whether `scan` and `for` should check for the existence of an `iterator` method before or after it checks to see if the `Expr` we're iterating over is some sort of list.  I'm currently thinking that the current behavior (check if it's a list first) is the best to avoid the additional overhead.  There's a small price of not being able to use the new features for classes that inherit from `BasicList`, but that's easy to overcome (e.g., I switched `ForeignObject` over to a subclass of `HashTable` in 21d0159e99234c82fe4695f74c43a476f4f48c91 so we could iterate over foreign arrays).

The other concerns that I've seen shared so far:

* *Why not a new `Iterator` class?*  I think the flexibility of being able to use any (non-list) class is quite nice.  Probably most of the time they will be hash tables to keep track of the things that the `next` method needs to do its thing, but it's nice not be locked into that.  In the `Python` package, `PythonObject`'s can be iterators.  And here's an example of a `FunctionClosure` that works:

```m2
i1 : FunctionIterator = new Type of FunctionClosure;

i2 : iterator FunctionIterator := identity;

i3 : next FunctionIterator := f -> f();

i4 : f = new FunctionIterator from (
         x := 0;
         () -> (
     	if x >= 10 then StopIteration
     	else x = x + 1));

i5 : for x in f list x

o5 = {1, 2, 3, 4, 5, 6, 7, 8, 9, 10}

o5 : List
```
* *The distinction between "iterator" and "iterable" is confusing*.  I agree, and am totally open to switching over to something else ("aggegrate"? "collection"?) to refer to the thing that is being iterated.

DanGrayson:
By the way, if you ever wanted `for` to return an iterator, you could introduce the syntax `for ... iterator ...`, and thus `for ... list ...` would always return a list.

PR #2614, opened 2022-09-17
Python updates (remove redundant code & integer conversion improvements)

d-torrance:
`python-c.c` was full of lots of one-line wrappers around Python C API functions, which just as easily could have been put in `Ccode` statements in `python.d`.  We make this change to ease maintenance.

### DIFF
--- a/M2/Macaulay2/d/basic.d
+++ b/M2/Macaulay2/d/basic.d
@@ -3,13 +3,6 @@
 use expr;
 
 header "#include <engine.h>"; -- required for raw hash functions
-header "
-#ifdef WITH_PYTHON
-#  include <Python.h>
-#else
-#  define PyObject_Hash(o) 0
-#  define PyErr_Clear()    0
-#endif";
 
 export hash(e:Expr):int := (
      when e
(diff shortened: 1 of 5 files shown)


d-torrance:
I added a few more commits improving conversion between GMP and Python integers using routines from [gmpy2](https://github.com/aleaxit/gmpy).  This previously only worked for integers that fit inside a `long`.

### Before
```m2
i2 : toPython 10^100

o2 = 0

o2 : PythonObject of class int

i3 : value pythonValue "10**100"
OverflowError: Python int too large to convert to C long
stdio:3:1:(3): error: python error
```

### After
```m2
i2 : toPython 10^100

o2 = 100000000000000000000000000000000000000000000000000000000000000000000000000
     00000000000000000000000000

o2 : PythonObject of class int

i3 : value pythonValue "10**100"

o3 = 100000000000000000000000000000000000000000000000000000000000000000000000000
     00000000000000000000000000
```

d-torrance:
After this evening's M2internals, I added a `hash` member to the `pythonObject` struct in the interpreter and made it so that mutable objects (like Python lists) use `nextHash` instead of some hardcoded value.

### Before
```m2
i1 : needsPackage "Python"; debug Core

i3 : x = set apply(100, i -> toPython {1, 2, 3});

i4 : tally \\ length \ buckets x

o4 = Tally{0 => 255}
           100 => 1

o4 : Tally
```

### After
```m2
i1 : needsPackage "Python"; debug Core

i3 : x = set apply(100, i -> toPython {1, 2, 3});

i4 : tally \\ length \ buckets x

o4 = Tally{0 => 156}
           1 => 100

o4 : Tally
```

DanGrayson:
Could you fix the conflicts?

d-torrance:
> Could you fix the conflicts?

Done

Issue #2617, opened 2022-09-19
ForeignFunctions on big-endian architectures


d-torrance:
The new `ForeignFunctions` package isn't working on big-endian architectures.  For example, from one of[ today's PPA builds on s390x](https://launchpadlibrarian.net/624433041/buildlog_ubuntu-jammy-s390x.macaulay2_1.20.0.1+git202209171831-0ppa202209180032~ubuntu22.04.1_BUILDING.txt.gz):

```m2
ForeignFunctions.m2:1234:1-1313:1: error:
 -- 
 -- i1 : --/<<PKGBUILDDIR>>/M2/Macaulay2/packages/ForeignFunctions.m2:1313: location of test code
 --      
 --      -----------
 --      -- value --
 --      -----------
 --      
 --      -- integer types
 --      assert Equation(value uint8(2^8 - 1), 2^8 - 1)
 -- stdio:8:1:(3): error: assertion failed:
 -- 64 == 255 is false
 -- 
ForeignFunctions.m2:1314:1-1325:1: error:
 -- -- -*- M2-comint -*- hash: 795122903
 -- 
 -- i1 : --/<<PKGBUILDDIR>>/M2/Macaulay2/packages/ForeignFunctions.m2:1325: location of test code
 --      
 --      -------------------
 --      -- foreignObject --
 --      -------------------
 --      assert Equation(value foreignObject 3, 3)
 -- stdio:6:1:(3): error: assertion failed:
 -- 2042745120 == 3 is false
```

d-torrance:
Figured out the issue. For some reason, I had decided to initially dereference pointers to integer types to `long` and then cast them to the appropriate type.  This is undefined behavior that just happens to work on little-endian machines, but not so much on big-endian.

PR #2618, opened 2022-09-20
Fix ForeignIntegerType bug on big-endian systems

d-torrance:
I only tested #2527 on little-endian architectures, and after it was merged, the first round of PPA builds on the big-endian s390x architecture all failed miserably.

For some reason, I thought it was a good idea to dereference the pointers to the various integer types as if they were `long` and then cast to the appropriate type after that.  That happens to work just fine on little-endian systems, but doesn't work on big-endian systems.

For example, let's say we initially allocated an 8-bit integer, say `0x01`.  Right after it in memory, we'd have 56 bits of garbage, say `0x23456789abcdef`.  When we later dereference this as a `long` (on a 64-bit system), we think we have `0x0123456789abcdef`, and then when we cast this back to an `int8_t` on a big-endian system, we'd get `0xef` instead of the `0x01` we were expecting.

So instead, we dereference each pointer as the correct type right off the bat.  I added corresponding `toExpr` and `toInteger` functions to deal with each of the fixed-width integer types to simplify this.

Closes: #2617


d-torrance:
Just pushed a few more commits to improve support for 64-bit foreign integer types on 32-bit systems.

PR #2619, opened 2022-09-22
Fix double "the" typos

fchapoton:
found using

`grep --no-filename -o -P -R '\b([a-z]{2,}) \1\b' M2/Macaulay2 | sort -u`

and then `git grep`


fchapoton:
oops, again forgot to add [skip-ci] when commiting, sorry

DanGrayson:
So many typos!!

PR #2620, opened 2022-09-22
some more typos, again [skip ci]

fchapoton:
Once again, do not hesitate to tell me if I should stop.


DanGrayson:
Thanks!   Fixing all these misprints and typos is well worthwhile!

PR #2624, opened 2022-09-24
Method degree polynomial posets

fchapoton:
The main aim here is to add a method `degreePolynomial` to the Posets package.

Warning : I did my best to write correct code, but I could not test it.

Also tweaking the doc of `tuttePolynomial` and adding some accents to `Poincaré`

### DIFF
--- a/M2/Macaulay2/packages/Posets.m2
+++ b/M2/Macaulay2/packages/Posets.m2
@@ -6407,6 +6407,7 @@ assert(moebiusFunction B === new HashTable from {("010","010") => 1, ("010","011
 assert(toString rankGeneratingFunction B === "q^3+3*q^2+3*q+1")
 assert(toString zetaPolynomial B == "q^3")
 assert(toString coxeterPolynomial B == "t^8+t^7+t^6-2*t^5-2*t^4-2*t^3+t^2+t+1")
+assert(toString degreePolynomial B == "x^3+3*x^2*y+3*x*y^2+y^3")
 assert(dilworthNumber B === 3)
 assert(isAtomic B == true)
 assert(isBounded B == true)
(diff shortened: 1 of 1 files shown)


DanGrayson:
All the tests have failed.

What prevents you from testing at home?

fchapoton:
I have difficulties in compiling M2. I have now tried and fixed the proposed code using sagemathcell.

fchapoton:
This seems to be working fine now. I hope I have not messed with the commits, which looks a bit strange here in github.

fchapoton:
ooch, I tried to merge, and it got even worse in terms of modified files

fchapoton:
ok, I have rebase and it is now showing the only modified file

gwynwhieldon:
Also, that was just a suggestion. I'd also be fine with it being merged as
is.

DanGrayson:
@fchapoton - what do you think of Gwyn's suggestion.  Could you do it easily?

fchapoton:
I can do that, except that I do not know where it should go.. somewhere in the same file?

DanGrayson:
Maybe Gwyn had a test in mind, in which case you could put it in `TEST /// ... ///` toward the end of the file.

gwynwhieldon:
Yes, that's what I was thinking.

The tests start at line 6164 in the file, and the first two are (small,
easy to read) examples of tests for (respectively) a new constructor and
the equality operator, and a function on a poset.

In theory, the tests would be nicely in order matching the order of methods
in the package. In practice, I don't think that's true at all here, so just
throwing it down at the bottom of the tests would probably be best (but
above line 6695, where the package functionality and tests ends at "end;").

Best,

Gwyn

fchapoton:
I have added a doctest on the boolean poset.

gwynwhieldon:
Alright, that LGTM too.

PR #2631, opened 2022-10-07
Iterator updates

d-torrance:
A few updates:

* There is a new `Iterator` class to simplify creating iterators.  It will likely work most of the time for writing `iterator` methods for classes.  Each `Iterator` object is a nullary `FunctionClosure` that serves as its own `next` method, so `next iter` is equivalent to `iter()`.
* `iterator` methods have been installed for several of the basic Macaulay2 types for which it makes sense (`VisibleList`, `Sequence`, and `String`), using the above `Iterator` class.
* A few features I added in [#1897](https://github.com/Macaulay2/M2/issues/1897) were removed since we can just fall back on iterators now.  In particular, `apply`, `scan`, and `for` used to create a copy of each string (as a sequence) and iterated over it, but now we can just use the new `iterator` method and traverse through the original string without copying.
* Support is added for `apply` (just the 2-argument version for now) and `select` for iterators.  In both instances, we return a new `Iterator` object.
* The two iterator classes for arrays in `ForeignFunctions` have been removed and replaced with `Iterator`.  (I didn't touch iteration in `Python` -- that's one case where it's easier to use a `PythonObject` as an iterator than an `Iterator`.)
* The documentation has been rewritten from scratch and hopefully is less confusing.  In particular, the focus is much less on writing new classes with `iterator` and `next` methods and much more on just using them.  The term "iterable" has been mostly removed (it's still in some error messages in the `d` directory for brevity, but I'm open to removing it there as well).

One thing I'm not 100% happy with is my code for `(net, Iterator)`.  Right now it uses `frames` to guess a representation of the thing we're iterating over, which works great for some common cases:

```m2
i1 : code (net, Iterator)

o1 = -- code for method: net(Iterator)
     M2/Macaulay2/m2/iterators.m2:10:22-13:55: --source code:
     net Iterator := iter -> (
         x := first first frames iter;
         FunctionApplication("iterator",
             (if instance(x, String) then format else net) x))

i2 : iterator "foo"

o2 = iterator "foo"

o2 : Iterator

i3 : iterator {1, 2, 3}

o3 = iterator {1, 2, 3}

o3 : Iterator
```
But sometimes it fails miserably.  For example, this simple iterator that just keeps returning random numbers:

```m2
i4 : Iterator(() -> random 20)

o4 = -*Function[stdio:4:14-4:24]*-

o4 : Iterator
```
Any suggestions?

### DIFF
--- a/M2/Macaulay2/packages/Macaulay2Doc/doc_iterators.m2
+++ b/M2/Macaulay2/packages/Macaulay2Doc/doc_iterators.m2
@@ -1,82 +1,102 @@
 doc ///
   Key
-    "iterators"
+    Iterator
   Headline
-    an overview of iterators in Macaulay2
+    class for iterators
   Description
     Text
-      A class in Macaulay2 is @EM "iterable"@ if there is a method
-      function installed for @TO iterator@ that takes as input instances
-      of that class and returns instances of an @EM "iterator"@ class, which
-      has method functions installed for @TO iterator@ (typically @TO identity@)
-      and @TO next@.  Note that iterators are themselves iterable.
+      This is a class designed to simplify writing @TO iterator@ methods.
+      Each instance is a nullary @TO FunctionClosure@ that serves as the
+      @TO next@ method for the iterator.
     Example
-      PrimeIterator = new SelfInitializingType of MutableHashTable;
-      iterator PrimeIterator := identity;
-      next PrimeIterator := p -> p#"val" = nextPrime(p#"val" + 1);
-      i = PrimeIterator {"val" => 1};
-      next i
-      next i
-      next i
+      iter = iterator {1, 2, 3}
+      code iter
     Text
-      If we wish to iterate over an object more than once, then it should have
-      a separate corresponding iterator class.
+      Each call of @TT "next iter"@ is equivalent to @TT "iter()"@.
     Example
-      PrimeIterable = new Type of HashTable;
-      iterator PrimeIterable := x -> PrimeIterator {"val" => 1};
-      P = new PrimeIterable;
-      i = iterator P;
-      next i
-      next i
-      j = iterator P;
-      next j
-      next j
+      next iter
+      iter()
+      next iter
+      iter()
     Text
-      If we would like iteration to stop at a certain point, then the @TO next@
-      method should return the symbol @TO StopIteration@.
+      For example, let us create a class that we may use to iterate over the
+      Fibonacci numbers.
     Example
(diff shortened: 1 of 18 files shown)


d-torrance:
I fixed the `net(Iterator)` issue:

```m2
i1 : Iterator(() -> random 20)

o1 = iterator ()

o1 : Iterator
```

M2/Macaulay2/packages/Macaulay2Doc/doc_iterators.m2
```diff
@@ -1,92 +1,128 @@
...
-      for x in P list x + 2
+      FibonacciNumbers = new Type of HashTable;
+      iterator FibonacciNumbers := fib -> Iterator(
+	  a := 0;
+	  b := 1;
+	  () -> (
+	      r := a;
+	      (a, b) = (b, a + b);
+	      r));
+      fibonacci = new FibonacciNumbers;
```
mahrud, line 33: What would `new Iterator of FibonacciNumbers` do here?

d-torrance:
```m2
i3 : new Iterator of FibonacciNumbers
stdio:3:1:(3): error: basic type for 'new' method should have been BasicList, HashTable, or Dictionary
```

What behavior were you expecting?

---

I'm thinking that I chose a bad example again...  There's no effective difference between different instances of the `FibonacciNumbers` class, so it's not really a good example of a class that a package author might write and want to install an `iterator` method for.

Maybe something like the following instead?

```m2
i1 : LucasSequence = new Type of HashTable;

i2 : iterator LucasSequence := seq -> Iterator(
         a := 0;
         b := 1;
         () -> (
     	r := a;
     	(a, b) = (b, seq#"P" * a - seq#"Q" * b);
     	r));

i3 : fibonacci = new LucasSequence from {"P" => 1, "Q" => -1};

i4 : for i in fibonacci list if i > 100 then break else i

o4 = {0, 1, 1, 2, 3, 5, 8, 13, 21, 34, 55, 89}

o4 : List
```
Or I could just drop the new class altogether from the documentation -- there's no reason that an `Iterator` object has to be returned from some installed `iterator` method.  Perhaps just simplify it down to the following?

```m2
i1 : fibonacci = Iterator(
         a := 0;
         b := 1;
         () -> (
     	r := a;
     	(a, b) = (b, a + b);
     	r));

i2 : for i in fibonacci list if i > 100 then break else i

o2 = {0, 1, 1, 2, 3, 5, 8, 13, 21, 34, 55, 89}
```

PR #2633, opened 2022-10-08
More iterator updates

d-torrance:
We remove the `FibonacciNumbers` class example and add some explanation about how iteration over an `Iterator` differs from other classes.

### DIFF
--- a/M2/Macaulay2/d/actors3.d
+++ b/M2/Macaulay2/d/actors3.d
@@ -1594,7 +1594,7 @@ map(e:Expr,f:Expr):Expr := (
 	 iter := getIterator(e);
 	 if iter != nullE
 	 then applyEEE(getGlobalVariable(applyIteratorS), iter, f)
-	 else WrongArg(1,"a list, sequence, integer, or iterable")));
+	 else WrongArg(1,"a list, sequence, integer, or iterable object")));
 map(e1:Expr,e2:Expr,f:Expr):Expr := (
      when e1
      is a1:Sequence do (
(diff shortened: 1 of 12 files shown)


d-torrance:
I added a few more features to this pull request, adding `toSequence`, `accumulate`, and `fold` for instances of classes with the `iterator` method installed.

For example:

```m2
i1 : toSequence iterator {1, 2, 3}

o1 = (1, 2, 3)

o1 : Sequence

i2 : fold(plus, iterator {1, 2, 3})

o2 = 6

i3 : iter = accumulate(plus, iterator {1, 2, 3})

o3 = iterator iterator {1, 2, 3}

o3 : Iterator

i4 : next iter

o4 = 3

i5 : next iter

o5 = 6

i6 : next iter

o6 = StopIteration

o6 : Symbol
```

M2/Macaulay2/d/actors3.d
```diff
@@ -2124,7 +2124,36 @@ toSequence(e:Expr):Expr := (
...
+			     while true do provide nullE));
+		     r.j = y;
+		     j = j + 1);
+		 Expr(
+		     if j == 0 then emptySequence
+		     else if j == length(r) then r
+		     else new Sequence len j do (
+			 foreach x in r do provide x)))
+	     else buildErrorPacket("no method for applying next to iterator"))
+	 else WrongArg("a list, sequence, string, or iterable")));
```
DanGrayson, line 2156: "iterable" is an adjective, so it doesn't match the other entries in this list.

d-torrance: I borrowed the term "iterable" (as a noun) from Python here.  It's already in several other error messages,e.g.:

https://github.com/Macaulay2/M2/blob/fe53c0b3cd9ccb7ebac5786ecec466293a8f8502/M2/Macaulay2/d/evaluate.d#L327-L328

It's shorthand for "instance of a class with the `iterator` method installed".  It can a confusing term, though, so I'm totally fine with changing it.  Any suggestions?

DanGrayson: Let's fix all of those.  How about "iterable object"?

d-torrance: Sounds good!  I'll push a new commit here in a bit.

d-torrance: Done

PR #2636, opened 2022-10-11
Add support for older versions of libffi

d-torrance:
In the `ForeignFunctions` package, we use the relatively new `ffi_get_struct_offsets` function (introduced in libffi 3.3 in 2019) to find the addresses of the various elements of a C `struct`.

This causes problems building on older systems, e.g., I had to backport libffi for the Ubuntu 18.04 PPA and @pzinn reported a build failure on Fedora 35 in Slack.

We provide an alternate implementation of `ffi_get_struct_offsets` for these older systems.

This currently only works for the autotools build.  @mahrud: How would one add a check to the cmake build to determine if `ffi_get_struct_offsets` exists and if so, define the macro `HAVE_FFI_GET_STRUCT_OFFSETS`? Maybe `check_function_exists`?

### DIFF
--- a/M2/Macaulay2/d/ffi.d
+++ b/M2/Macaulay2/d/ffi.d
@@ -11,6 +11,39 @@ header "#include <dlfcn.h>
 	#endif
 	#include <M2mem.h>";
 
+-- ffi_get_struct_offsets not introduced until libffi 3.3 in 2019
+-- so we provide an alternate implementation
+header "
+#ifndef HAVE_FFI_GET_STRUCT_OFFSETS
+#define FFI_ALIGN(v, a)  (((((size_t) (v))-1) | ((a)-1))+1)
+
+ffi_status
+ffi_get_struct_offsets(ffi_abi abi, ffi_type *struct_type, size_t *offsets)
+{
+  ffi_cif cif;
+  ffi_status status;
+  ffi_type **ptr;
+  size_t size;
+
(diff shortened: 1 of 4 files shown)


mahrud:
> This currently only works for the autotools build. @mahrud: How would one add a check to the cmake build to determine if `ffi_get_struct_offsets` exists and if so, define the macro `HAVE_FFI_GET_STRUCT_OFFSETS`? Maybe `check_function_exists`?

Yes, see https://github.com/Macaulay2/M2/blob/master/M2/cmake/configure.cmake#L322 and https://github.com/Macaulay2/M2/blob/master/M2/cmake/check-libraries.cmake#L329. Since this depends on a library, it should probably go in the latter file.

d-torrance:
> > This currently only works for the autotools build. @mahrud: How would one add a check to the cmake build to determine if `ffi_get_struct_offsets` exists and if so, define the macro `HAVE_FFI_GET_STRUCT_OFFSETS`? Maybe `check_function_exists`?
> 
> Yes, see https://github.com/Macaulay2/M2/blob/master/M2/cmake/configure.cmake#L322 and https://github.com/Macaulay2/M2/blob/master/M2/cmake/check-libraries.cmake#L329. Since this depends on a library, it should probably go in the latter file.

Great, thanks!  I've just pushed a commit adding the check for the cmake build.

PR #2637, opened 2022-10-12
Set randomSeed to 0 for basic test

d-torrance:
Although the probability of a given integer not appearing after calling `rawRandomZZ 10` 100 times is extraordinarily rare, it is possible, and has caused[ at least one PPA build](https://launchpadlibrarian.net/628254443/buildlog_ubuntu-jammy-arm64.macaulay2_1.20.0.1+git202210111525-0ppa202210071724~ubuntu22.04.1_BUILDING.txt.gz) to fail.

We set `randomSeed` to 0 so that we get deterministic results.


mahrud:
Do we win the lottery?

PR #2641, opened 2022-10-13
fix #2503

### LINKED ISSUE #2503: Substitution / promotion issues
There are still a few issues with substitution/promotion that I want to take care of. Here's one (well, two, actually):
```
i1 : R=QQ[x]
   
o1 = R

o1 : PolynomialRing

i2 : sub(x/(x+1),x=>x+1) -- rejects

stdio:3:4:(3): error: expected x to be a generator of frac R

i3 :  sub(x/(x+1),x_(frac R)=>x+1) -- accepts but wrong target therefore wrong result

o3 = 1

o3 : R
```
I might add some more in this thread.

pzinn:
- fixes the various promotion / substitution issues mentioned in #2503 
- fixes a related bug with `Number * Vector` ( `1/2 * vector {1}` would return the wrong type of vector)
- unrelated simplification of `texMath Product` that I threw in

### DIFF
--- a/M2/Macaulay2/m2/expressions.m2
+++ b/M2/Macaulay2/m2/expressions.m2
@@ -1073,9 +1073,9 @@ texMath Product := v -> (
     else (
 	v = apply(v, x -> if class x === Power and (x#1 === 1 or x#1 === ONE) then x#0 else x);
 	p := precedence v;
-	nums := apply(v, x -> isNumber x);
+	nums := apply(v, x -> isNumber x or (class x === Power and isNumber x#0));
 	precs := apply(v, x -> precedence x <= p);
-	seps := apply (n-1, i-> if nums#i and (nums#(i+1) or class v#(i+1) === Power and isNumber v#(i+1)#0) then "\\cdot "
+	seps := apply (n-1, i-> if nums#(i+1) then "\\times "
 	    else if class v#i =!= Power and class v#i =!= Subscript and not precs#i and not precs#(i+1) then
 	    if nums#i or class v#i === Symbol then "\\," else "\\ "
 	    else "");
--- a/M2/Macaulay2/m2/matrix.m2
+++ b/M2/Macaulay2/m2/matrix.m2
@@ -75,19 +75,14 @@ reduce = (tar,rawF) -> (
 
 addHook(ReduceHooks, Strategy => Default, (tar, rawF) -> rawF % raw gb presentation tar)
 
-Matrix * Number := Matrix * ZZ := (m,i) -> i * m
-Number * Matrix := (r,m) -> (
-     S := ring m;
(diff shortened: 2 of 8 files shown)


pzinn:
interesting problem. The fix to `Number*Vector`, as necessary as it may be (cf the example above, has generated a new problem. it seems `WeylGroups` defines a descendent of `Vector`. This in itself is the source of an infinite number of potential problems of the same kind as the one that makes the example fail. after all, independently of my PR, we have very similar definitions, e.g., 
`Vector + Vector := Vector => (v,w) -> vector(v#0+w#0)` which would also fail for descendants of `Vector`.

pzinn:
incidentally I had the same issue with my package `CotangentSchubert`. I define a new type of `Vector` as part of a `DiagonalAlgebra`. In https://github.com/Macaulay2/M2/blob/b754a3d591812474926f8f0f4c6ca4b3a76a01d0/M2/Macaulay2/packages/CotangentSchubert/cotangent.m2#L165 I'm forced to redefine `D+D` (where `D` is the descendent of `Vector`).
So it seems to me that the package `WeylGroups` should do something similar.
Also, this issue was discussed in https://groups.google.com/g/macaulay2/c/dOLI8BqdE7I/m/57ecnAh1AgAJ?utm_medium=email&utm_source=footer&pli=1

pzinn:
let's tenatively make the change I suggested. We can ask the 2 package owners later. (one is me)

pzinn:
finally got everything to work. 3 packages were affected:
- `CotangentSchubert`. @pzinn are you OK with the change: YES
- `WeylGroups`. @baptistecalmes are you OK with the one-line change redefining a `Number * Weight` to be a `Weight`?
- `NoetherianOperators`: @kroner @jchen419 @antonleykin exact same thing with `Number * DiffOp`

baptistecalmes:
Dear Paul,

I totally missed this bug report.

I’m ok with the spirit of this change.
It’s been a while, but I guess it was probably an omission/bug, it should return a Weight as you say.

However:

If p is a weight, the syntax
-p
also needs to return a Weight.

It doesn’t work with your fix.
(which doesn’t pass the tests, by the way, because of that issue)

One needs to explicitly define -p.
It is not automatically computed as (-1)*p by M2.

Here is a version of the file updated accordingly.
https://github.com/baptistecalmes/WeylGroups/blob/main/WeylGroups.m2

(I’ve also updated the documentation structure with subnodes while I was at it).

Should I submit another PR with this updated file of WeylGroups?

jchen419:
Hi Paul, the change to `NoetherianOperators` looks fine, as it should not break any functionality (looping @haerski in as well).

pzinn:
@baptistecalmes if you want you can directly push your file to my main branch. or you can let me do it, as you wish.

PR #2642, opened 2022-10-15
Some capitals

fchapoton:
to some famous names


fchapoton:
forgot to fix "schubert" too

PR #2643, opened 2022-10-18
Add foreign function pointers for callbacks to Macaulay2 functions in ForeignFunctions package

d-torrance:
This allows us to pass Macaulay2 functions to C code.  For example:

```m2
i2 : compar = foreignFunctionPointerType("compar", int, {voidstar, voidstar})

o2 = compar

o2 : ForeignFunctionPointerType

i3 : f = (a, b) -> value int a - value int b

o3 = f

o3 : FunctionClosure

i4 : intcmp = compar f

o4 = int32(*f)(void*,void*)

o4 : ForeignObject of type compar

i5 : qsort = foreignFunction("qsort", void, {voidstar, ulong, ulong, compar})

o5 = qsort

o5 : ForeignFunction

i6 : x = (4 * int) {4, 2, 3, 1}

o6 = {4, 2, 3, 1}

o6 : ForeignObject of type int32[4]

i7 : qsort(x, 4, size int, intcmp)

i8 : x

o8 = {1, 2, 3, 4}

o8 : ForeignObject of type int32[4]
```

### DIFF
--- a/M2/Macaulay2/d/ffi.d
+++ b/M2/Macaulay2/d/ffi.d
@@ -561,3 +561,49 @@ ffiUnionType(e:Expr):Expr := (
 	toExpr(x))
     else WrongArg("a list"));
 setupfun("ffiUnionType", ffiUnionType);
+
+----------------------------
+-- function pointer types --
+----------------------------
+
+ffiClosureFunction(cif:Pointer "ffi_cif *", ret:voidPointer,
+    args:Pointer "void **", userData:voidPointer):void := (
+    nargs := Ccode(int, cif, "->nargs");
+    f := Expr(Ccode(FunctionClosure, userData));
+    x := Expr(
+	if nargs == 1 then Expr(pointerCell(Ccode(voidPointer, "*", args)))
+	else Expr(new Sequence len nargs at i do provide pointerCell(
+		Ccode(voidPointer, args, "[", i, "]"))));
+    when applyEE(f, x)
+    is ptr:pointerCell
+    do Ccode(void, "memcpy(", ret, ", ", ptr.v, ", ", cif, "->rtype->size)")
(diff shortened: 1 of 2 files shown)


mahrud:
I'm not sure if this is quite the goal, but here is a suggestion: perf is a widely used profiler with various ways to visualize cpu, memory, etc, for instance using [flamegraphs](http://www.brendangregg.com/flamegraphs.html). Here's an example of profiling M2 running a sample script:

> perf record -F 99 -a -g -- M2 --script sample.m2
> perf script | stackcollapse-perf.pl > out.perf-folded
> flamegraph.pl out.perf-folded > perf-kernel.svg
firefox perf-kernel.svg

Various profiling tools (perf, dtrace, etc.) require a [symbol](http://www.brendangregg.com/perf.html#Symbols) map to translate memory addresses to functions and structs. gcc automatically keeps this information in debug sections, but top-level functions are not mapped.

Languages that have a virtual machine, like Java or Node have the same issue, which is why the profiler [perf](http://www.brendangregg.com/perf.html) has a way to inform it about the symbol translations. Quoting from [here](http://www.brendangregg.com/blog/2014-09-17/node-flame-graphs-on-linux.html):

> In 2009, Linux perf_events added JIT symbol support, so that symbols from language virtual machines like the JVM could be inspected. It works in the following amazingly simple way:
> - Your JIT application must be modified to create a /tmp/perf-PID.map file, which is a simple text database containing symbol addresses (in hex), sizes, and symbol names.
> - That's it.

For instance [perf-map-agent](https://github.com/jvm-profiling-tools/perf-map-agent) is a java agent for building this file. If the overhead isn't too much, I suspect we could use something like this feature of yours to have proper profiling in Macaulay2.

PR #2646, opened 2022-10-25
JIT compilation of SLPs

antonleykin:
This adds an ability to use `g++` to compile `SLProgram`s on the fly and use the resulting function in a dynamically linked library for SLP evaluation.


d-torrance:
Perhaps the JIT code could be generalized and moved to its own package?

antonleykin:
> Perhaps the JIT code could be generalized and moved to its own package?

At this point this is a very local solution --- this effects only SLPrograms defined in the SLPexpressions package.
Hypothetically it would be nice to at least think about generalization: e.g., JIT compiler for all code in M2 language.

antonleykin:
@timduff35, I've committed changes related to MonodromySolver   

@d-torrance, there were conflicts with ForeignFunctions... I think because I started practicing "rebase"... but am not good at it :)

timduff35:
> @timduff35, I've committed changes related to MonodromySolver
> 
> @d-torrance, there were conflicts with ForeignFunctions... I think because I started practicing "rebase"... but am not good at it :)

Changes are ok by me!

PR #2647, opened 2022-10-26
Fix #2640 -- double enter no longer interrupts

### LINKED ISSUE #2640: Inputting strings with empty lines
I'm not sure if this is a bug or a feature. Try the following: (can be done in emacs, terminal or web interface, result is the same)
type `a="[enter]"[enter]` where [enter] is the new line key.
M2 correctly interprets this as assigning to `a` the string containg the single new line character. so far, so good.
now type `a="[enter][enter]"[enter]`. you'd assume M2 would interpret it as a assigning to `a` the strings with two newlines. but no, after the second [enter] it seems to give up on the input and returns the `i1:` prompt again.
note however that `a="[enter]blah[enter]"[enter]` will work just fine.
any reason for this strange behaviour?

pzinn:
It's all in the title. tests:
```
i1 : f=i->(
     
     i+1
     
     )

o1 = f

o1 : FunctionClosure

i2 : a="
     
     "

o2 = 



i3 : b=///
     
     ///

o3 = 



i4 : c= -* long
     
     comment
     *- 1

o4 = 1
```

### DIFF
--- a/M2/Macaulay2/d/lex.d
+++ b/M2/Macaulay2/d/lex.d
@@ -112,7 +112,6 @@ getstringslashes(o:PosFile):(null or Word) := (		    -- /// ... ///
      getc(o);		  -- pass '/'
      getc(o);		  -- pass '/'
      pos := position(o);
-     hadnewline := false;
      tokenbuf << '\"';					    -- "
      while true do (
 	  ch := getc(o);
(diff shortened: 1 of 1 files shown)


pzinn:
not sure what's going on on mac os

DanGrayson:
I've had trouble with this before.  Our configure script tries to detect which command line options the compiler accepts for enabling openmp.  There are 10 possibilities.  It works on my laptop, which is an up-to-date version 12.6 of Mac OS.  I tried, in a pull request, to have the github runner run Mac OS 12 instead of 11, but the same problem occurred.  I wish I could log in to the github running machines!

@mikestillman

PR #2649, opened 2022-10-28
Try to fix openmp

DanGrayson:


DanGrayson:
Thanks for trying this, Mahrud.  It might work!  Indeed, the clang on my machine is version 14, not 11.

mahrud:
I'm not sure how autotools looks for openmp, but `llvm-config --libdir` should point to where the headers and libraries are located.

DanGrayson:
The error results from the configure script trying various command line options for enabling openmp with the C compiler, to see which one works.  On my system, this works:
```sh
gallium$ clang -Xpreprocessor -fopenmp -lomp foo.c
gallium$ gcc -Xpreprocessor -fopenmp -lomp foo.c
gallium$ clang -v
Apple clang version 14.0.0 (clang-1400.0.29.102)
Target: x86_64-apple-darwin21.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
gallium$ gcc -v
Apple clang version 14.0.0 (clang-1400.0.29.102)
Target: x86_64-apple-darwin21.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

Perhaps the compiler on the github runner machine is still slightly different, and requires different options.

We used to be able to download artifacts after a build failed, so we could try to analyze things.  I'd like to see config.log.

mahrud:
> Perhaps the compiler on the github runner machine is still slightly different, and requires different options.

Nothing is special about the github machine. You've probably set up some non-standard environment variables on your system to make compilers find things more easily.

> We used to be able to download artifacts after a build failed, so we could try to analyze things. I'd like to see config.log.

You can still do that [here](https://github.com/Macaulay2/M2/actions/runs/3353374215). Autotools doesn't seem to be telling the compiler where to find the openmp headers or libraries at all:
```
configure:10216: g++ -o conftest -std=gnu++14 -g3 -O2 -Xpreprocessor -fopenmp -lomp -I/usr/local/include -DNDEBUG -L/usr/local/lib   -L/usr/local/opt/python@3.10/Frameworks/Python.framework/Versions/3.10/lib -g3 conftest.cpp -ltbb  >&5
conftest.cpp:103:10: fatal error: 'omp.h' file not found
#include <omp.h>
         ^~~~~~~
1 error generated.
```
For comparison, this is what cmake uses:
```
-isystem /usr/local/opt/libomp/include -Xpreprocessor -fopenmp /usr/local/opt/libomp/lib/libomp.dylib -fopenmp=libomp
```

DanGrayson:
I think `/usr/local/opt/libomp/include` isn't needed, because homebrew puts a symbolic link to omp.h into /usr/local/include.

But, we aren't installing it with brew!  Here's our install line:
```
          brew install autoconf automake bison boost tbb ccache cmake ctags gnu-tar libtool make ninja pkg-config yasm libffi
```

Maybe we should try adding libomp to that line.

mahrud:
> I think `/usr/local/opt/libomp/include` isn't needed, because homebrew puts a symbolic link to omp.h into /usr/local/include.

That's not true about [libomp](https://github.com/Homebrew/homebrew-core/blob/89ac56e4c8b0cee300629283fa04ddba27e96d24/Formula/libomp.rb#L23-L24):
```rb
  keg_only "it can override GCC headers and result in broken builds"
```

> But, we aren't installing it with brew!

This is also not true:
```
2022-10-29T23:42:00.4098910Z [33mWarning:[0m libomp 15.0.2 is already installed and up-to-date.
```
I even added it in this PR just to be sure:
https://github.com/Macaulay2/M2/blob/1de727f62c7f7106d10052eb2d580a1e74514f4d/.github/workflows/test_build.yml#L76
But it isn't needed so we can remove it.

Really, it is autotools that needs to find the right directory or receive it as an argument and pass it to the compiler.

DanGrayson:
Yes, brew claims libomp is keg-only, but nevertheless, it links it into /usr/local:
```sh
$ ls -l /usr/local/include/omp.h
lrwxrwxr-x  1 brew  admin  37 Oct 30 14:20 /usr/local/include/omp.h -> ../Cellar/libomp/15.0.3/include/omp.h
```

DanGrayson:
Congratulations, this time it succeeded!  I'll merge.

mahrud:
> Yes, brew claims libomp is keg-only, but nevertheless, it links it into /usr/local:
> 
> ```shell
> $ ls -l /usr/local/include/omp.h
> lrwxrwxr-x  1 brew  admin  37 Oct 30 14:20 /usr/local/include/omp.h -> ../Cellar/libomp/15.0.3/include/omp.h
> ```

Are you sure you didn't manually run `brew link libomp` at some point?

DanGrayson:
Hmm, that's a good question.  Would it remember?  The date on that link is yesterday, and what I did yesterday is update and upgrade.

DanGrayson:
Anyway, libomp is nowhere else on my computer.  Where do you think the github runners are finding it?

mahrud:
I instructed autotools explicitly where to find it:
https://github.com/Macaulay2/M2/blob/02f6a264764afc1836d8f74ddfa8ec58b6b3e8f4/.github/workflows/test_build.yml#L165-L166

Whereas for cmake I just need to put the path to llvm libraries in the environment:
https://github.com/Macaulay2/M2/blob/02f6a264764afc1836d8f74ddfa8ec58b6b3e8f4/.github/workflows/test_build.yml#L112-L113

DanGrayson:
Ah, so you did add it to the brew install line.  Thanks!

mahrud:
> Ah, so you did add it to the brew install line. Thanks!

No! It was preinstalled!

DanGrayson:
Does github preinstall some brew packages for us??

mahrud:
You can see everything about the images here: https://github.com/actions/runner-images#available-images
Even on ubuntu there are many brew packages preinstalled.

DanGrayson:
Thanks.  I clicked through, but I see nothing about libomp here: https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md

PR #2650, opened 2022-10-29
Fix urlEncode now that apply(String, Function) returns an Iterator

d-torrance:
We need a list to pass to `concatenate`.

I broke this in #2631 -- sorry, @pzinn!

### Before
```m2

i1 : oeis apply(30, n -> binomial(n + 2, 2))
stdio:1:1:(3): error: expected a sequence or list of strings, integers, or symbols
```

### After
```m2
i1 : oeis apply(30, n -> binomial(n + 2, 2))

o1 =   1. A000217 (see https://oeis.org/A000217 ) Triangular numbers: a(n) =
          binomial(n+1,2) = n*(n+1)/2 = 0 + 1 + 2 + ... + n.
       2. A161680 (see https://oeis.org/A161680 ) a(n) = binomial(n,2): number
          of size-2 subsets of {0,1,...,n} that contain no consecutive integers.
       3. A179865 (see https://oeis.org/A179865 ) Number of n-bit binary numbers
          containing one run of 0's.
       4. A105340 (see https://oeis.org/A105340 ) a(n) = n*(n+1)/2 mod 2048.
       5. A105338 (see https://oeis.org/A105338 ) a(n) = n*(n+1)/2 mod 512.
       6. A105339 (see https://oeis.org/A105339 ) a(n) = n*(n+1)/2 mod 1024.
       7. A089594 (see https://oeis.org/A089594 ) Alternating sum of squares to
          n.

o1 : OL
```

### DIFF
--- a/M2/Macaulay2/packages/OnlineLookup.m2
+++ b/M2/Macaulay2/packages/OnlineLookup.m2
@@ -24,7 +24,7 @@ percentEncoding = hashTable transpose {
 -- TODO: use in/move to html.m2
 urlEncode = method()
 urlEncode Nothing := identity
-urlEncode String := s -> concatenate apply(s, c -> if percentEncoding#?c then percentEncoding#c else c)
+urlEncode String := s -> concatenate toList apply(s, c -> if percentEncoding#?c then percentEncoding#c else c)
 
 oeisHTTP := "http://oeis.org";
 oeisHTTPS := "https://oeis.org";


pzinn:
thanks!

DanGrayson:
Re: "now that apply(String, Function) returns an Iterator"

That seems counterintuitive -- for `apply` to return an iterator.  It should just do it.

d-torrance:
> Re: "now that apply(String, Function) returns an Iterator"
> 
> That seems counterintuitive -- for `apply` to return an iterator. It should just do it.

For the most part, `apply` should return the same type of object that it was given, e.g.:

```m2
i1 : apply({1, 2, 3}, x -> x^2)

o1 = {1, 4, 9}

o1 : List

i2 : apply([1, 2, 3], x -> x^2)

o2 = [1, 4, 9]

o2 : Array

i3 : apply((1, 2, 3), x -> x^2)

o3 = (1, 4, 9)

o3 : Sequence
```
For strings, this never really made sense because there's no guarantee that the given function would return characters that could then be concatenated back into a string.  So the old behavior for `apply(String, Function)` was to create a brand-new sequence that contained the characters of the original string and run `apply` on it, returning a `Sequence` object.  But this wasted memory, as now we had two copies of the characters of the string floating around -- the original string itself and the sequence we created for `apply`.  The new behavior avoids this copy.

And for any non-list object, `apply` always returns an `Iterator`, since there's no guarantee that the iteration will ever end.  For example:

```m2
i1 : primeiter = Iterator(n := 0; () -> n = nextPrime(n + 1))

o1 = iterator 0

o1 : Iterator

i2 : squareprimeiter = apply(primeiter, x -> x^2)

o2 = iterator iterator 0

o2 : Iterator

i3 : next squareprimeiter

o3 = 4

i4 : next squareprimeiter

o4 = 9

i5 : next squareprimeiter

o5 = 25

i6 : next squareprimeiter

o6 = 49

i7 : next squareprimeiter

o7 = 121

i8 : next squareprimeiter

o8 = 169
```

PR #2656, opened 2022-11-02
Restore apply(String, Function) behavior (returns a Sequence)

d-torrance:
Dan's comment (https://github.com/Macaulay2/M2/pull/2650#issuecomment-1297272479) got me thinking a little more about the changes I made in #2631 in the return value of `apply(String, Function)` (returning an `Iterator` instead of a `Sequence`). For example:

* While in general there's no guarantee that iterating over an `Iterator` object will terminate, there definitely is when iterating over a `String`, so returning a `Sequence` makes about as much sense as anything else.
* The change in behavior broke at least five lines of code -- four in `Core` and one in `OnlineLookup`.  It may break even more code out there, so it's probably a good idea to restore the earlier behavior before the 1.21 release.

So we, on the surface, restore the original behavior.  It now returns a `Sequence` again, but still uses iterators under the hood so that it doesn't make a copy of each character of the string into a new sequence like it did pre-#2631.

### 1.20 behavior
```m2
i1 : apply("foo", identity)

o1 = (f, o, o)

o1 : Sequence
```

### Current development branch behavior
```m2
i1 : apply("foo", identity)

o1 = iterator iterator "foo"

o1 : Iterator
```

### Behavior after this pull request
```m2
i1 : apply("foo", identity)

o1 = (f, o, o)

o1 : Sequence
```

### DIFF
--- a/M2/Macaulay2/d/actors3.d
+++ b/M2/Macaulay2/d/actors3.d
@@ -1632,6 +1632,9 @@ map(e:Expr,f:Expr):Expr := (
 	  if !isInt(i)
 	  then WrongArgSmallInteger()
 	  else map(toInt(i),f))
+     is s:stringCell do (
+	 toSequence(applyEEE(
+		 getGlobalVariable(applyIteratorS), getIterator(e), f)))
      else (
 	 iter := getIterator(e);
 	 if iter != nullE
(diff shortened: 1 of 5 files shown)


pzinn:
agreed. my package `MergeTeX` no longer works, for example...

PR #2663, opened 2022-11-04
New version of package MixedMultiplicity

goelKriti:
The package is revised to consider a broader class of rings as compared to its previous version.

### DIFF
--- a/M2/Macaulay2/packages/=distributed-packages
+++ b/M2/Macaulay2/packages/=distributed-packages
@@ -247,3 +247,4 @@ CodingTheory
 WhitneyStratifications
 JSON
 ForeignFunctions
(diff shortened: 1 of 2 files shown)


goelKriti:
I guess the checks fail because there are two commits, each containing the package file with minor changes. Is there a way to recitfy the situation?

d-torrance:
Having multiple commits is fine.  The issue is that `MixedMultiplicity` was added to `=distributedpackages` a second time, but it's already there:

https://github.com/Macaulay2/M2/blob/b754a3d591812474926f8f0f4c6ca4b3a76a01d0/M2/Macaulay2/packages/%3Ddistributed-packages#L204-L208

PR #2665, opened 2022-11-07
Begin changelog for 1.21

d-torrance:
Add my changes to Core since 1.20:

* #2613, #2631
* #2594
* #2595
* #2593


d-torrance:
@DanGrayson -- when you merged, the "functionality added" section ended up before the "improved packages" section, but the other changelog entries have that order reversed.  Does that matter?

DanGrayson:
Yes, I thought that would happen.  I'll fix it later on, when I add my bits to the changelog.

PR #2668, opened 2022-11-07
Matroids v1.6.0

jchen419:
- Add methods: isNonCrossing, isPositivelyOriented, positiveOrientation, isPositivelyOrientable, kruskalSpanningForest, coordinatingPath, rescalingRepresentative, searchRepresentation, writeToString, saveMatroid, readFromFile
- Add documentation for (co)extensions from Matt Mastroeni

### DIFF
--- a/M2/Macaulay2/packages/Matroids.m2
+++ b/M2/Macaulay2/packages/Matroids.m2
@@ -1,11 +1,10 @@
 newPackage("Matroids",
 	AuxiliaryFiles => true,
-	Version => "1.5.0",
-	Date => "May 4, 2022",
+	Version => "1.6.0",
+	Date => "November 7, 2022",
 	Authors => {{
 		Name => "Justin Chen",
-		Email => "jchen@math.berkeley.edu",
-		HomePage => "https://math.berkeley.edu/~jchen"}},
+		Email => "jchen@math.berkeley.edu"}},
 	Headline => "a package for computations with matroids",
 	Keywords => {"Matroids"},
 	HomePage => "https://github.com/jchen419/Matroids-M2",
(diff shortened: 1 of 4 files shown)


jchen419:
So I think (one current) issue is that after #2625, `delete` no longer works on a `MutableList` (since these are `BasicList`s, but not `VisibleList`s). @d-torrance, @DanGrayson: any opinions on keeping `delete` functional on `MutableList`s?

d-torrance:
> So I think (one current) issue is that after #2625, `delete` no longer works on a `MutableList` (since these are `BasicList`s, but not `VisibleList`s). @d-torrance, @DanGrayson: any opinions on keeping `delete` functional on `MutableList`s?

I think it totally makes sense for `delete` to work in this case.  I just didn't think about it for #2625.  The following should work:

```diff
--- a/M2/Macaulay2/m2/structure.m2
+++ b/M2/Macaulay2/m2/structure.m2
@@ -20,7 +20,8 @@ position(VisibleList,VisibleList,Function) := o -> (v,w,f) -> (
      )
 
 delete = method()
-delete(Thing, VisibleList) := (x,v) -> select(v, i -> i =!= x)
+delete(Thing, VisibleList)  :=
+delete(Think, MustableList) := (x,v) -> select(v, i -> i =!= x)
 
 -- Local Variables:
 -- compile-command: "make -C $M2BUILDDIR/Macaulay2/m2 "
```
This would create a brand-new `MutableList` object without the deleted element instead of modifying the original, which may or not be the desired behavior.

jchen419:
@d-torrance: That's a good point - I guess `delete` has always returned a new copy, which makes sense for immutable objects. Should it be smarter for mutable objects? What does `remove` do for a `MutableHashTable`? (Btw: I was thinking if `delete` should just be defined on `BasicList`s.)

d-torrance:
> @d-torrance: That's a good point - I guess `delete` has always returned a new copy, which makes sense for immutable objects. Should it be smarter for mutable objects? What does `remove` do for a `MutableHashTable`? (Btw: I was thinking if `delete` should just be defined on `BasicList`s, as anyways this is the level of generality that `select` is defined for.)

`remove` modifies the original:

```m2
i1 : x = new MutableHashTable from {1 => 2, 3 => 4}

o1 = MutableHashTable{...2...}

o1 : MutableHashTable

i2 : remove(x, 1)

i3 : peek x

o3 = MutableHashTable{3 => 4}
```
Maybe `delete` could keep its current behavior, but for `BasicList` objects, too, and `remove(MutableList, Thing)` could be added?

jchen419:
> Maybe `delete` could keep its current behavior, but for `BasicList` objects, too, and `remove(MutableList, Thing)` could be added?

That sounds good to me - though I imagine the documentation should have a caveat, stating that `remove` is more sensible than `delete` for `MutableList`s.

PR #2672, opened 2022-11-09
Add ability to opt out of compiling with libffi

d-torrance:
@mikestillman has reported issues compiling on Apple silicon machines due to the `ForeignFunctions` package and libffi.  We add the ability to opt out of compiling with libffi using either `--without-libffi` in autotools or `-DWITH_FFI=OFF` in cmake.

As a consequence, we cache the `ForeignFunctions` examples and provide dummy values for the interpreter-level functions that would have otherwise been imported from `Core`.

The last commit sets up the GitHub builds to try building without libffi to make sure that this all works, which is why this is currently a draft.  I'll restore the GitHub builds to their current behavior afterward.


d-torrance:
The builds all worked (with the exception of the unrelated `autotools-macos` failure) with libffi disabled.  See https://github.com/Macaulay2/M2/actions/runs/3425214448.

PR #2673, opened 2022-11-09
add random(ZZ, Ideal) and random(List, Ideal)

mahrud:
Suggested by @naddington and @eisenbud, with the slight caveat that this implementation returns a matrix. Here are some examples:

```m2
i1 : S = ZZ/101[x, y]

o1 = S

o1 : PolynomialRing

i2 : I = ideal"x2, y2"

             2   2
o2 = ideal (x , y )

o2 : Ideal of S

i3 : random(2, I)

o3 = | 28x2+38y2 |

             1       1
o3 : Matrix S  <--- S

i4 : random({3, 3}, I)

o4 = | -21x3-47x2y+49xy2-33y3 13x3-13x2y+14xy2-2y3 |

             1       2
o4 : Matrix S  <--- S

i5 : R = ZZ/101[x, y, z, Degrees => {{1,0}, {-1,1}, {0,1}}]

o5 = R

o5 : PolynomialRing

i6 : J = ideal"x2, y2, z2"

             2   2   2
o6 = ideal (x , y , z )

o6 : Ideal of R

i7 : random({{2, 2}}, J)

o7 = | 16x4y2+30x3yz-11x2z2 |

             1       1
o7 : Matrix R  <--- R

i8 : random(toList(3:{1, 1}), J)

o8 = | 15x2y -20x2y -24x2y |

             1       3
o8 : Matrix R  <--- R
```

### DIFF
--- a/M2/Macaulay2/m2/genmat.m2
+++ b/M2/Macaulay2/m2/genmat.m2
@@ -156,6 +156,9 @@ random(Module, Module) := Matrix => opts -> (F,G) -> (
 				   r)))));
 	  map(F, G, applyTable(degreesTable, k -> (randomElement k)()))))
 
+random(ZZ,   Ideal) := Matrix => opts -> (d, I) -> random({d}, I, opts)
+random(List, Ideal) := Matrix => opts -> (L, I) -> generators I * random(source generators I, (ring I)^(-L), opts)
+
 -- Local Variables:
 -- compile-command: "make -C $M2BUILDDIR/Macaulay2/m2 "
 -- End:
--- a/M2/Macaulay2/packages/Macaulay2Doc/functions/random-doc.m2
+++ b/M2/Macaulay2/packages/Macaulay2Doc/functions/random-doc.m2
@@ -1,101 +1,162 @@
-document {
-     Key => random,
-     Headline => "get a random element",
-     "This function can be used to get random elements of various sorts.",
-     SeeAlso => {"setRandomSeed"}
-     }
-document {
-     Key => (random, ZZ),
-     Headline => "random integer",
-     Usage => "random n",
-     Inputs => {"n"=> {}},
-     Outputs => {ZZ => {"a random integer in the range ", TT "0 .. n-1"}},
-     EXAMPLE lines ///
-     random 57
-     random 10^50
-     tally apply(100, i -> random 7)
-     ///,
-     SeeAlso => {setRandomSeed, tally}
-     }
+undocumented {
+    (random, GaloisField),
+    (random, QuotientRing),
+    (random, RingFamily),
+    }
 
-document {
-     Key => (random, ZZ,ZZ),
-     Headline => "random integer in a range",
-     Usage => "random(min,max)",
-     Inputs => {"min","max"},
-     Outputs => {ZZ => {"a random integer in the range ", TT "min .. max"}},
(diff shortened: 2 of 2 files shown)


naddington:
Thanks for this guys. I hate to quibble, but would it make more sense to have random(ZZ, Ideal) return a polynomial, and random(List, Ideal) return a matrix? Otherwise we’re going to be typing “det random(3, I)” a lot.

Best,
Nick

mikestillman:
I agree, `random(3, I)` should return a polynomial.

mahrud:
I agree that it would make more sense, so I should have explained why I wrote it with this caveat. The main reason is that in M2 degrees are always returned as lists (e.g. `degree x^3 = {3}`), so I think we would want `random(3, I)` and `random({3}, I)` to return the same thing, and moreover for bigraded rings we would want to call `random({3, 3}, I)` to get a polynomial also, but then `random(List, Ideal)` would sometimes return a polynomial and sometimes return a matrix! Of course, it's not impossible to get passed the complications above, but I figured it was simpler and avoid bugs to always return a matrix and have the user either do one of the following:
```m2
f = det random(3, I);
--
f = (random(3, I))_(0,0);
--
I = ideal random(3, I);
```

This is not an issue for `random(ZZ, Ring)` and `random(List, Ring)` because that one doesn't support returning more than one element. I thought about adding a similar feature of returning multiple elements to that function, but decided against changing the behavior.

Having explained that, if you think returning a polynomial is worth the potential confusion, let me know and I'm happy to implement that instead.

mahrud:
We could also have it always return a list, so you can always use `first random(3, I)`. Is that better?

mahrud:
> Another possibility would be to return a polynomial every time and just call `random` multiple times if more than one random polynomial is desired. So, in the examples above, `i4` could be replaced by `apply(2, i -> random(3, I))`, and `i7` could be replaced by `random({2, 2}, J)`. I think this would be more consistent with `random(List, Ring)`.

Having the option to get an ideal for r forms of degree d was David's suggestion:

> I use Nick's method too, and have written it too many times; so I agree that it should be in the core. However, I usually want an ideal, and not just generated by one random element, so I do `ideal (gens I *random(source gens I, R^L)` where L is a list of degrees. Of course if you are making a general facility, then there should be a method for r forms of degree d, too, so you don't always have to do toList (r:-d) or some such. There are probably other useful variations, too.

mahrud:
Okay I just made a new pull request (#2679) which returns as follows:
1. `random(3, I)`, polynomial
2. `random({3}, I)`, **polynomial**
3. `random({3, 4}, I)`, list
4. `random({{3}, {4}}, I)`, list
5. `random({{3}}, I)`, list
6. `random({}, I), empty list
7. `random({3, 3}, J)`, polynomial
8. `random({{3, 3}}, J)`, list
9. `random({}, J), empty list

Case 2 is the main counter-intuitive return type here, because the braces around the degree are not treated as a list of degrees, but rather as a multidegree of size 1. Behind the scenes, case 1 is converted to case 2 and case 3 is converted to case 4 automatically.

But there's another counter intuitive case: say `I = ideal 2_ZZ`, then should `random({}, I)` be a single even integer like case 1 and 7 above or an empty list like case 6 and 9 above? In the new PR when degree length is zero I return an error.

Issue #2670, opened 2022-11-08
Undocumented behavior of "remove"


d-torrance:
According to the [documentation](https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.20/share/doc/Macaulay2/Macaulay2Doc/html/_remove.html), `remove` is for removing keys from mutable hash tables.  But according to the source code, it also supports lists, sequences, and databases:

https://github.com/Macaulay2/M2/blob/b754a3d591812474926f8f0f4c6ca4b3a76a01d0/M2/Macaulay2/d/actors5.d#L571-L598

For example:

```m2
i1 : remove({1, 2, 3}, 0)

o1 = {2, 3}

o1 : List
```

Since the same functionality can be achieved using `drop`, perhaps `remove` for lists should be analogous to hash tables, i.e., it could be used to modify *mutable* lists, e.g., something like:

```m2
i1 : x = new MutableList from {1, 2, 3}

o1 = MutableList{...3...}

o1 : MutableList

i2 : remove(x, 0)

i3 : peek x

o3 = MutableList{2, 3}
```

See also https://github.com/Macaulay2/M2/pull/2668#issuecomment-1306565188.

PR #2680, opened 2022-11-13
Update and document "remove" for lists

d-torrance:
Previously, `remove(L, i)` did the same thing as `drop(L, {i, i})` for a list (or sequence) `L` and an integer `i`, but this was undocumented ~~and not used anywhere~~.  (*Edit*: It's actually used in several packages.)

We change the behavior so that it behaves like `remove` for hash tables and dictionaries.  In particular, it now only works for *mutable* lists, modifies the existing list, and returns `null`:

```m2
i1 : x = new MutableList from {1, 2, 3, 4}

o1 = MutableList{...4...}

o1 : MutableList

i2 : remove(x, 0)

i3 : peek x

o3 = MutableList{2, 3, 4}
```

Closes: #2670

One other small change: we update the error message when subscripting a list or sequence with a `ZZ` that's too large to fit inside an `int`:

### Before
```m2
i1 : L = {}

o1 = {}

o1 : List

i2 : L_{2^31}
stdio:2:2:(3): error: array index 2147483648 out of bounds 0 .. -1
```
### After
```m2
i1 : L = {}

o1 = {}

o1 : List

i2 : L_(2^31)
stdio:2:2:(3): error: expected a small integer
```

### DIFF
--- a/M2/Macaulay2/d/actors5.d
+++ b/M2/Macaulay2/d/actors5.d
@@ -552,21 +552,15 @@ endlfun(e:Expr):Expr := (
 setupfun("endl",endlfun).Protected = false;
 setupconst("newline", toExpr(newline));
 
-remove(x:Sequence,i:int):Sequence := (
-     n := length(x);
-     if i < 0 then i = i + n;
-     if 0 <= i && i < n then (
-	  new Sequence len n-1 do foreach y at j in x do if i != j then provide y
-	  )
-     else x
-     );
-
-remove(x:List,i:int):List := (
-     v := remove(x.v,i);
-     if v == x.v then return x;
-     r := List(x.Class, v, 0, x.Mutable);
-     r.hash = hash(r);
-     r);
+remove(x:List,i:int):Expr:= (
+     n := length(x.v);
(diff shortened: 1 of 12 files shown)


d-torrance:
So, despite being undocumented, `remove` for lists was actually used in 5 packages and one of the basic tests.  I've switched the packages to use `drop` instead and removed the basic tests.  I also added a changelog entry mentioning the breaking change, if this is accepted for 1.21.

One question:  In the current proposal, an error is raised if specifying an out of range index, e.g.,:

```m2
i1 : x = new MutableList from {1, 2, 3}

o1 = MutableList{...3...}

o1 : MutableList

i2 : remove(x, 3)
stdio:2:1:(3): error: array index 3 out of bounds 0 .. 2
```
Is this the desired behavior, or should it just do nothing in this case?  That's what happens with `remove` for hash tables:

```m2
i3 : y = new MutableHashTable from {"a" => 1, "b" => 2, "c" => 3}

o3 = MutableHashTable{...3...}

o3 : MutableHashTable

i4 : remove(y, "d")

i5:
```

DanGrayson:
I think giving such an error message is fine.

PR #2681, opened 2022-11-13
Introduce rings with torsion class group

mahrud:
- introduced monoids with torsion degree groups
- wip: improved reduceDegree
- fixed newRing for rings with torsion degree group
- fixed basis at the top-level
- updated NormalToricVariety to support torsion class groups
- added support for passing degrees via a map or the degree group


mahrud:
This package adds support for torsion in the degree group, but since many aspects of these rings are not verified yet (e.g. truncations, sheaf cohomology, etc.) and the changes are largely invisible to someone not intending to use rings with torsion degrees, I didn't announce it in `changes.m2`. The goal, as discussed with @mikestillman, is to allow people that are aware of these changes (e.g. internal people and several others) to be able to easily test using these rings in the new M2 release and let us know if they find any bugs or what features are missing before we add the remaining changes (e.g. moving some of the changes in the engine) and document & announce it publicly.

mahrud:
@ggsmith this is currently the changes to the pre-conditions of `ring NormalToricVariety`:
```diff
        if isDegenerate X then 
            error "-- not yet implemented for degenerate varieties";
-       if not isFreeModule classGroup X then 
-           error "-- gradings by torsion groups not yet implemented";
        -- constructing ring
```
Do you have a preference about what should be added here, other than a one-time warning from Core when a ring with torsion degree group is created? For instance, is this what you had in mind?
```diff
+       if not isSimplicial X then 
+           printerr "Warning: computations over the Cox ring of a non-simplicial toric variety are experimental";
```

ggsmith:
Personally, I would prefer something like:
``` 
if not isFreeModule classGroup X then
    printerr "Warning: computations over Cox rings with torsion grading groups are experimental."
```

mahrud:
Looks like all the checks have passed. @ggsmith do all the changes in NormalToricVarieties look acceptable to you? Here is an example of what happens the first time:
```m2
i6 : S = ring X
 -- Warning: computations over Cox rings with torsion grading groups are experimental
 -- Warning: computations over rings with torsion grading groups are experimental
```
The second warning is once per session, so after the first time only the first warning appears.

ggsmith:
I have no objections to the changes made in ```NormalToricVarieties```.

PR #2685, opened 2022-11-16
Polyhedra: Fix bug in regularSubdivision(Matrix, Matrix)

lkastner:
In this method a lower envelope is computed whose (projected) facets in
turn are a regular subdivision of the input point set. Previously
the return values probably were actual polyhedra, but now these are just
sets of indices indicating which points together form a cell. However
these indices use the vertices of the intermediate polyhedron used for
computing the lower envelope. Thus they need to be remapped onto the
input points.

### DIFF
--- a/M2/Macaulay2/packages/Polyhedra/extended/polyhedron/methods.m2
+++ b/M2/Macaulay2/packages/Polyhedra/extended/polyhedron/methods.m2
@@ -134,13 +134,21 @@ regularSubdivision (Polyhedron,Matrix) := (P,w) -> (
    apply (S, s -> convexHull LP_s)
 )
 
-regularSubdivision (Matrix,Matrix) := (M,w) -> (
+regularSubdivision (Matrix,Matrix) := (MM, w) -> (
+   M := promote(MM, QQ);
    n := numColumns M;
    -- Checking for input errors
    if numColumns w != numColumns M or numRows w != 1 then error("The weight must be a one row matrix with number of points many entries");
    P := convexHull(M||w,matrix (toList(numRows M:{0})|{{1}}));
    F := select(faces (1,P), f -> #(f#1) ==0);
-   apply (F, f -> f#0)
+   pointIndices := new MutableHashTable;
+   for i from 0 to numcols M-1 do
+      pointIndices#(M_{i}) = i;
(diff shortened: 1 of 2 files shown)


lkastner:
Should the PR be against `master` or `development`?

mikestillman:
@lkastner Your code failed because you need to make some of your variables local variables by using :=, including (I think) M, vertP, pointIndices, and permutation.  Also, you might want to use ```M_{i}``` instead of ```M_i```, although it might not make a difference...  I can create some examples for the function too.

mikestillman:
@lkastner Here is your function with some :='s added in.  I'm testing it on some other examples now (I didnt' change the use of vectors (```M_i```) to matrices (```M_{i}```)).

```
regularSubdivision (Matrix,Matrix) := (MM, w) -> (
   M := promote(MM, QQ);
   n := numColumns M;
   -- Checking for input errors
   if numColumns w != numColumns M or numRows w != 1 then error("The weight must be a one row matrix with number of points many entries");
   P := convexHull(M||w,matrix (toList(numRows M:{0})|{{1}}));
   F := select(faces (1,P), f -> #(f#1) ==0);
   pointIndices := new MutableHashTable;
   for i from 0 to numcols M-1 do
      pointIndices#(M_i) = i;
   permutation := new MutableHashTable;
   vertP := (vertices P)^{0..numrows M - 1};
   for i from 0 to numcols vertP - 1 do
      permutation#i = pointIndices#(vertP_i);
   sort apply (F, f -> sort apply(f#0, i->permutation#i))
  )
```

lkastner:
> @lkastner Your code failed because you need to make some of your variables local variables by using :=, including (I think) M, vertP, pointIndices, and permutation. Also, you might want to use `M_{i}` instead of `M_i`, although it might not make a difference... I can create some examples for the function too.

Yes, sorry about that, my M2 didn't finish building on time yesterday, so I only tried it out in the M2 shell so far. Will fix.

mikestillman:
@DanGrayson I'm good with this pull request.  @lkastner Thanks for fixing that so fast!

mikestillman:
@DanGrayson I was going to merge this, but it is based off of master.  Can I just change the base to development? (I get warnings from github when I try that)

PR #2686, opened 2022-11-18
Primary decomposition of pseudomonomial ideals

alanavc:
Here we present the package PseudomonomialPrimaryDecomposition for the computation of the primary decomposition of squarefree pseudomonomial ideals. Internally, the package performs computation using bitwise logical operations on integers. Accepted in Journal of Software for Algebra and Geometry.


mikestillman:
@DanGrayson this was accepted by JSAG, so should get a star too...

PR #2690, opened 2022-11-23
Remove unused StartDictionary from ParseTree union

d-torrance:
We never actually create any `StartDictionary` objects.


mahrud:
What was it even for?

d-torrance:
> What was it even for?

No idea!

Issue #2683, opened 2022-11-15
FFI example error on Ubuntu 22.04


mahrud:
One [example](https://github.com/Macaulay2/homebrew-tap/actions/runs/3465824882/jobs/5789005813#step:9:15916) seems to be failing when bottling M2 from the main branch for Ubuntu 22.04 when using brew's libffi:
```m2
 -- making example results for "ForeignFunctionPointerType Function"        
 ulimit -c unlimited; ulimit -t 700; ulimit -m 850000; ulimit -s 8192; ulimit -n 512;  cd /tmp/M2-106561-0/25-rundir/; GC_MAXIMUM_HEAP_SIZE=400M "/tmp/macaulay2-20221114-21055-dzcob8/M2-master/M2/BUILD/build-brew/usr-dist/x86_64-Linux-Ubuntu-22.04/bin/M2-binary" -q --no-randomize --no-readline --silent --stop --print-width 77 -e 'needsPackage("ForeignFunctions",Reload=>true,FileName=>"/tmp/macaulay2-20221114-21055-dzcob8/M2-master/M2/Macaulay2/packages/ForeignFunctions.m2")' <"/tmp/M2-106561-0/0___Foreign__Function__Pointer__Type_sp__Function.m2" >>"/tmp/macaulay2-20221114-21055-dzcob8/M2-master/M2/BUILD/build-brew/usr-dist/common/share/doc/Macaulay2/ForeignFunctions/example-output/___Foreign__Function__Pointer__Type_sp__Function.errors" 2>&1
/tmp/macaulay2-20221114-21055-dzcob8/M2-master/M2/BUILD/build-brew/usr-dist/common/share/doc/Macaulay2/ForeignFunctions/example-output/___Foreign__Function__Pointer__Type_sp__Function.errors:0:1: (output file) error: Macaulay2 exited with status code 1
/tmp/M2-106561-0/0___Foreign__Function__Pointer__Type_sp__Function.m2:0:1: (input file)
M2: *** Error 1
 -- 0.908139 seconds elapsed
```
I can also reproduce this locally on Fedora using systems libffi.
```m2
i3 : capture examples "ForeignFunctionPointerType Function"
-- SIGSEGV
-* stack trace, pid: 1050644
 0# stack_trace(std::ostream&, bool) at /home/mahrud/Projects/M2/M2/M2/Macaulay2/bin/main.cpp:127
[...]

PR #2701, opened 2022-11-29
Disable valgrind helpers when valgrind is disabled.

jkyang92:
@mikestillman @MichaelABurr Hopefully fixes the issue noted in https://github.com/Macaulay2/M2/issues/2683#issuecomment-1330734590. I don't have the ability to verify that this actually fixes the issue, but it should.

### DIFF
--- a/M2/Macaulay2/d/M2mem.c
+++ b/M2/Macaulay2/d/M2mem.c
@@ -204,6 +204,9 @@ char *getmoremem_atomic (char *s, size_t old, size_t new) {
      return p;
      }
 
+/* Valgrind helper functions */
+#ifndef NVALGRIND
+
 void *I_WRAP_SONAME_FNNAME_ZU(libgcZdsoZd1,GC_malloc)(size_t s){
     long result;
     OrigFn fn;
@@ -249,6 +252,7 @@ void *I_WRAP_SONAME_FNNAME_ZU(libgcZdsoZd1,GC_realloc)(void *p, size_t s){
     return (void*)result;
 }
 
+#endif /* NVALGRIND */
 
 /*
  Local Variables:
--- a/M2/Macaulay2/d/M2mem.h
+++ b/M2/Macaulay2/d/M2mem.h
@@ -22,11 +22,20 @@ extern char *getmoremem(char *, size_t oldsize, size_t newsize);
 extern char *getmoremem1(char *, size_t newsize);
 extern char *getmoremem_atomic(char *, size_t oldsize, size_t newsize);
 
+/* Valgrind helper functions,
+ * NVALGRIND can be defined to explicitly disable this
+ * But also valgrind.h (included by memcheck.h) will define NVALGRIND
+ * if we are on a platform it doesn't understand
+ */
+#ifndef NVALGRIND
 void *I_WRAP_SONAME_FNNAME_ZU(libgcZdsoZd1,GC_malloc)(size_t);
 void *I_WRAP_SONAME_FNNAME_ZU(libgcZdsoZd1,GC_malloc_atomic)(size_t);
 void *I_WRAP_SONAME_FNNAME_ZU(libgcZdsoZd1,GC_malloc_ignore_off_page)(size_t);
 void *I_WRAP_SONAME_FNNAME_ZU(libgcZdsoZd1,GC_malloc_atomic_ignore_off_page)(size_t);
 void *I_WRAP_SONAME_FNNAME_ZU(libgcZdsoZd1,GC_realloc)(void*, size_t);
+#endif /* NVALGRIND */
+
+
 
 #define sizeofarray(s,len) (sizeof(*(s)) - sizeof((s)->array) + (len)*sizeof((s)->array[0]))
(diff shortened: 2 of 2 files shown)


d-torrance:
I can confirm that this works.  Currently building Macaulay2 on an M1 laptop (@mikestillman -- finally got it!), and adding `-DNVALGRIND` to `CPPFLAGS` got me past the `OrigFn` issue.

jkyang92:
@d-torrance Can you also try it without the -DNVALGRIND flag? in theory the valgrind.h file should define it if it's on a platform it doesn't recognize, but I don't know if that actually works

DanGrayson:
I prefer the solution in https://github.com/Macaulay2/M2/pull/2702 , because it's an option for the configure script, hence easier for users to implement.

jkyang92:
@DanGrayson This should not require users to add anything to the CFLAGS, although we can certainly add a configure option to add a -DNVALGRIND to the CFLAGS to disable it even on platforms where valgrind works. Also none of the code here involved requires valgrind to be installed to compile.

mahrud:
I also prefer this solution. Seems simpler and doesn't involve even more configure options.

MichaelABurr:
Should the default for valgrind be on or off?  I lean towards disabling valgrind by default and only enabling if by compiler flag (perhaps combined with the solution here).

d-torrance:
> Can you also try it without the -DNVALGRIND flag? in theory the valgrind.h file should define it if it's on a platform it doesn't recognize, but I don't know if that actually works

@jkyang92 - Oh nice -- I didn't realize it was already defined for me!  Yes, it works great without explicitly adding anything to `CPPFLAGS`.

mahrud:
There's already a `MEMDEBUG` option in both autotools and cmake:
https://github.com/Macaulay2/M2/blob/ddacbc2601d51f9e88194462d103909fcda044d4/M2/cmake/configure.cmake#L21
https://github.com/Macaulay2/M2/blob/ddacbc2601d51f9e88194462d103909fcda044d4/M2/configure.ac#L447

Would be nice if this could replace that or be combined with it (e.g. define `NVALGRIND` if `MEMDEBUG` is not defined), but that's the main reason I want to avoid having multiple similar sounding options.

DanGrayson:
The default for inclusion of valgrind code should definitely be *not*, for a user who runs into the problem Mike encountered, where OrigFn was an undefined type, will be mystified and have no idea about what to do.

So, instead of using a variable called "NVALGRIND", which means "no valgrind", use a variable that means "yes valgrind".

jkyang92:
I don't mind hiding it behind a flag, but the issue Mike encountered is a bit of a special case, related to the fact that M1 is an architecture unsupported by the valgrind headers.

Of note, none of this code requires the installation of valgrind on the system, since the idea and the implementation is we include the headers needed for this in our own repository. And in fact, there's no case where installing valgrind would have fixed Mike's issue or any similar issue.

As for NVALGRIND macro name, the name is somewhat "standardized" in that for the valgrind.h and memcheck.h headers use this to disable themselves if defined. The name is also in analogy to the NDEBUG macro that nearly every project uses.

For now, on this pull request, I will add some code to tie NVALGRIND to NDEBUG. Although I'm conflicted as to whether to just add it to the M2mem.* files or whether to add NVALGRIND at the configure/cmake step

DanGrayson:
> I don't mind hiding it behind a flag, but the issue Mike encountered is a bit of a special case, related to the fact that M1 is an architecture unsupported by the valgrind headers.


We're going to have more and more users with M1 machines, and some of them will try to compile from source code.  It is a special case, but our code should deal with all special cases.

mikestillman:
I think that the apple M1 is an important case, and we should support it "out of the box"!

jkyang92:
> I think that the apple M1 is an important case, and we should support it "out of the box"!

Sure, my point was mostly that we aren't likely to encounter this issue again. And this pull request would certainly fix the M1 case sufficiently. We can't fix the fact that they can't use valgrind on those platforms of course.

As a note again, the valgrind.h header included via the memcheck.h header defines NVALGRIND if it doesn't find a platform it knows how to deal with.

mahrud:
Until earlier this year gfan would not work on M1 because of an architecture problem (it would compile fine, and show its version and such fine, but no computations). That didn't force us to exclude gfan by default on all systems. I'm sure Valgrind support on M1 isn't far behind either.

DanGrayson:
> > I think that the apple M1 is an important case, and we should support it "out of the box"!
> 
> Sure, my point was mostly that we aren't likely to encounter this issue again. And this pull request would certainly fix the M1 case sufficiently. We can't fix the fact that they can't use valgrind on those platforms of course.
> 
> As a note again, the valgrind.h header included via the memcheck.h header defines NVALGRIND if it doesn't find a platform it knows how to deal with.

Interesting!  @mikestillman -- could you test this pull request on your M1 mac to see if it works?

mikestillman:
Hi all, this works on my apple M1 mac.  Well, it fails while running ForeignFunctions examples, but that fix is I think in Doug's pull request #2705, but no build errors from my standard build for this (on cmake and autotools builds).  I recommend we accept it!

PR #2702, opened 2022-11-29
make use of valgrind code optional

DanGrayson:


DanGrayson:
@jkyang92 -- take a look at this

(on some machines, reported by Mike, the valgrind type "OrigFn" is not defined.)

DanGrayson:
@mikestillman

mahrud:
If you add a new configure option, could you please also add it to `cmake/configure.cmake` and `INSTALL-CMake.md`?

DanGrayson:
> If you add a new configure option, could you please also add it to `cmake/configure.cmake` and `INSTALL-CMake.md`?

I would be happy to, if I knew anything about cmake.  But I don't, too bad!

mahrud:
> I would be happy to, if I knew anything about cmake. But I don't, too bad!

It's easier to learn by example:
https://github.com/Macaulay2/M2/blob/ddacbc2601d51f9e88194462d103909fcda044d4/M2/cmake/configure.cmake#L30

DanGrayson:
Closing, in deference to #2701

PR #2703, opened 2022-11-29
Add the Homebrew library directory to rpath in macOS

d-torrance:
This will enable `dlopen` to find various shared libraries used by the `ForeignFunctions` package. 

This fixes 4 out of the 5 failing examples from the `ForeignFunctions` package reported by @mikestillman on Apple M1 machines, at least for the autotools build.

@mahrud -- Any suggestions on how to make the corresponding fix for the cmake build?  I'm assuming somewhere in `Macaulay2/bin/CMakeLists.txt`, but this seems like it'd part of the `TODO: plenty of Darwin specific linker flags` comment.  :)

### Before
```m2
i2 : openSharedLibrary "mpfr"
[...]

PR #2705, opened 2022-11-30
Check brew library directory when using dlopen on Apple M1 systems

d-torrance:
This is an alternate to #2703 to fix 4 out of the 5 example errors in `ForeignFunctions` on Apple M1 systems.

The issue is that Homebrew installs its M1 libraries in `/opt/homebrew/lib`, but `dlopen` doesn't look there when it tries to open a shared library.  So we get around this by adding a wrapper around `dlopen` at top level:

* If we're on any other type of system, just run `dlopen` as usual.
* If the filename of the library has a leading `/` (i.e., the user specified an absolute path), then just run `dlopen` as usual.
* Try running `dlopen` as usual.
* If that fails, check if the desired library exists in `$(brew --prefix)/lib`, and if it does, run `dlopen` on it.

### DIFF
--- a/M2/Macaulay2/packages/ForeignFunctions.m2
+++ b/M2/Macaulay2/packages/ForeignFunctions.m2
@@ -521,7 +521,7 @@ net SharedLibrary := lib -> lib#1
 -- dlopen fails
 importFrom_Core {"isAbsolutePath"}
 if (version#"operating system" == "Darwin" and
-    version#"architecture" == "aarch64")
+    member(version#"architecture", {"aarch64", "arm64"}))
 then (
     brewPrefix := replace("\\s+$", "", get "!brew --prefix");
     dlopen' = filename -> (
--- a/M2/Macaulay2/packages/ForeignFunctions.m2
+++ b/M2/Macaulay2/packages/ForeignFunctions.m2
@@ -516,6 +516,24 @@ SharedLibrary = new SelfInitializingType of BasicList
 SharedLibrary.synonym = "shared library"
 net SharedLibrary := lib -> lib#1
 
+-- on apple silicon machines, shared libraries are often in /opt/homebrew/lib,
+-- but this is not in DYLD_LIBRARY_PATH, so we try there if the first call to
+-- dlopen fails
+importFrom_Core {"isAbsolutePath"}
+if (version#"operating system" == "Darwin" and
+    version#"architecture" == "aarch64")
+then (
+    brewPrefix := replace("\\s+$", "", get "!brew --prefix");
+    dlopen' = filename -> (
+	if isAbsolutePath filename then dlopen filename
+	else (
+	    try dlopen filename
+	    else (
+		brewLibrary := brewPrefix | "/lib/" | filename;
+		if fileExists brewLibrary then dlopen brewLibrary
+		else error("could not find " | filename))))
+    ) else dlopen' = dlopen
+
 openSharedLibrary = method(TypicalValue => SharedLibrary,
     Options => {FileName => null})
 openSharedLibrary String := o -> name -> (
@@ -523,7 +541,7 @@ openSharedLibrary String := o -> name -> (
 	"lib" | name |
 	if version#"operating system" == "Darwin" then ".dylib"
 	else ".so");
-    SharedLibrary {dlopen filename, name})
+    SharedLibrary {dlopen' filename, name})
 
 ----------------------
 -- foreign function --


M2/Macaulay2/packages/ForeignFunctions.m2
```diff
@@ -516,14 +516,31 @@ SharedLibrary = new SelfInitializingType of BasicList
...
 
+-- on apple silicon machines, shared libraries are often in /opt/homebrew/lib,
+-- but this is not in DYLD_LIBRARY_PATH, so we try there if the first call to
+-- dlopen fails
+if (version#"operating system" == "Darwin" and
+    version#"architecture" == "aarch64")
+then (
+    brewPrefix := replace("\\s+$", "", get "!brew --prefix");
+    dlopen' = filename -> (
+	if first filename == "/" then dlopen filename
```
mahrud, line 527: You could potentially use one of these here:
https://github.com/Macaulay2/M2/blob/master/M2/Macaulay2/m2/startup.m2.in#L152-L154

d-torrance: Oh nice -- thanks for pointing these out!

mikestillman:
@d-torrance I ran this pull request on my apple M2 mac (running MacOS 12.6), and I still get 5 errors during example building using cmake.  I am pretty sure I am using your ffi branch for this run...

I am currently running it on autotools.

mikestillman:
@d-torrance, For instance, 
```
i1 : mps = openSharedLibrary "mps"
stdio:1:7:(3): error: dlopen(libmps.dylib, 0x0001): tried: '/Users/mike/src/M2-current-branches/M2-development/M2/BUILD/mike/builds.tmp/arm64-appleclang/usr-dist/arm-Darwin-macOS-12.6/bin/../lib/Macaulay2/lib/libmps.dylib' (no such file), '/Users/mike/src/M2-current-branches/M2-development/M2/BUILD/mike/builds.tmp/arm64-appleclang/usr-dist/arm-Darwin-macOS-12.6/bin/../lib/libmps.dylib' (no such file), '/Users/mike/src/M2-current-branches/M2-development/M2/BUILD/mike/builds.tmp/arm64-appleclang/usr-dist/arm-Darwin-macOS-12.6/bin/../lib/Macaulay2/lib/libmps.dylib' (no such file), '/Users/mike/src/M2-current-branches/M2-development/M2/BUILD/mike/builds.tmp/arm64-appleclang/usr-dist/arm-Darwin-macOS-12.6/bin/../lib/libmps.dylib' (no such file), 'libmps.dylib' (no such file), '/usr/local/lib/libmps.dylib' (no such file), '/usr/lib/libmps.dylib' (no such file), '/private/var/folders/wb/8v4mm0j52pq9pf5gkr8f23z40000gr/T/M2-83569-0/11-rundir/libmps.dylib' (no such file)
```

```
bash-3.2$ brew --prefix mps
Error: No available formula with the name "mps". Did you mean mpw, mpv, mpc, mpd or mas?
```

d-torrance:
Hmm, interesting... That error should have occurred inside a `try` clause, and then it should be looking for the Homebrew library in the corresponding `else` clause.

What are `version#"operating system"` and `version#"architecture"`?  If they're not `"Darwin"` and `"aarch64"`, respectively, then there's no wrapper around `dlopen`.

mikestillman:
@d-torrance : Some version keys/values:
architecture => arm64   
operating system release => 21.6.0
operating system => Darwin

d-torrance:
Ah, that's the problem -- your system has `arm64` instead of `aarch64` (which is what it is on my university's M1 laptop).  That's an easy fix -- I'll push a commit here in moment.

DanGrayson:
@mikestillman -- I really like that you called your machine an "apple M2 mac".

DanGrayson:
@mikestillman -- does this now work on the M1?

mikestillman:
@d-torrance @DanGrayson All ForeignFunction examples run, except one: the qsort example fails (on my Apple M1)

d-torrance:
The `qsort` example shouldn't be part of the build after #2707.

mikestillman:
good!  @DanGrayson Let's accept this pull request now, and then I think we are ready to start doing 1.21 finally!

mikestillman:
@DanGrayson Once this is merged into development, I'll try building that directly too.  I think it will work, as I have been doing that as we go (but not in my most recent test for this pull request).

PR #2707, opened 2022-12-02
Temporarily stop exporting foreign function pointer types

d-torrance:
This is a band-aid for #2683 so that it doesn't hold up the release of 1.21 any longer than it already has.  :)

The vast majority of the features of the `ForeignFunctions` package work on a wide variety of systems.  The only exception that has been causing problems are `ForeignFunctionPointerType` objects, which wrap Macaulay2 functions as pointers to C functions, e.g., for callbacks.  See #2683.

To get around the crashes that occur with this feature on a number of systems, it is no longer exported, and the documentation and tests are now disabled.  Anyone interested in experimenting with the foreign function pointer types can still access them via `debug ForeignFunctions`.

### DIFF
--- a/M2/Macaulay2/packages/ForeignFunctions.m2
+++ b/M2/Macaulay2/packages/ForeignFunctions.m2
@@ -30,7 +30,7 @@ export {
     "ForeignPointerArrayType",
     "ForeignUnionType",
     "ForeignStructType",
-    "ForeignFunctionPointerType",
+--    "ForeignFunctionPointerType",
     "ForeignObject",
 
 -- built-in foreign types
@@ -67,7 +67,7 @@ export {
     "foreignPointerArrayType",
     "foreignStructType",
     "foreignUnionType",
-    "foreignFunctionPointerType",
+--    "foreignFunctionPointerType",
     "foreignSymbol",
     "getMemory",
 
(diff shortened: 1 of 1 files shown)


d-torrance:
I just added a commit reverting (almost all of) #2672, which was an earlier "solution" to the problem by allowing builds without any `libffi` support.  It was too complicated, though, e.g., by requiring the caching of all the example outputs.

mikestillman:
I'll test this on my apple M1 after I teach today.

d-torrance:
> I'm happy with removing the cached examples, but can we keep the `WITH_FFI` build option? I imagine it might come in handy on weird setups!

Yeah, that would be nice.  I guess another possibility would be to keep `WITH_FFI` (and `--with-libffi` in autotools), and when it's disabled, just remove `ForeignFunctions` from `=distributed-packages`.  Then we wouldn't need to worry about caching examples.  Or maybe there's a better solution?

I'm going to remove the last commit from this PR, and I'll work on an alternate implementation of `WITH_FFI`/`--with-libffi` in another PR.

mahrud:
What does the Python package do when libpython isn't linked? Maybe the package can check a variable in `version` and if libffi isn't linked just call `endPackage`?

d-torrance:
> What does the Python package do when libpython isn't linked?

It does this:

https://github.com/Macaulay2/M2/blob/ddacbc2601d51f9e88194462d103909fcda044d4/M2/Macaulay2/packages/Python.m2#L25-L30

It's even uglier than the current behavior of the `ForeignFunctions` package -- `no-python.m2` contains a bunch of dummy definitions of methods so we can still load the documentation even when Macaulay2 was built with Python support.

> Maybe the package can check a variable in `version` and if libffi isn't linked just call `endPackage`?

Yeah, that's probably the simplest solution.  The package would have very minimal documentation on these builds, but I suppose that's okay if the package itself isn't functional!

mahrud:
If you do `endPackage` right away then there won't be any documentation for the package on those systems, which I imagine is fine.

I guess you might want to consider other packages depending on it in the future also.

PR #2708, opened 2022-12-02
Support non-libffi builds without caching ForeignFunctions examples

d-torrance:
This partially reverts #2672.  Rather than still shipping a non-working copy of the `ForeignFunctions` package with cached examples when Macaulay2 is built with `--without-libffi` (autotools) or `-DWITH_FFI=OFF)` (cmake), we remove `ForeignFunctions` from the list of distributed packages.

This is a draft PR for now.  The last commit sets the GitHub builds to build without libffi.  If that goes well, then I'll remove that commit.


mikestillman:
@d-torrance Which pull request is planning on going into 1.21?  I'll try that one on my mac.  I tried one earlier, and was down to one example error, rather than 5.

d-torrance:
> @d-torrance Which pull request is planning on going into 1.21? I'll try that one on my mac. I tried one earlier, and was down to one example error, rather than 5.

Both #2705 and #2707.  #2705 fixes 4 of the example errors, and #2707 "fixes" the 5th one by removing that feature for now.

Would it be easier if it was just one pull request?  The problems weren't closely related so I kept them separate.

mikestillman:
Seems to work now on my apple M1 !  Thanks @d-torrance!  @DanGrayson I think I'm done with changes for 1.21...

d-torrance:
The macOS autotools build for this PR was failing because of a subtle difference between GNU sed and BSD sed.  Just force-pushed a fix.

d-torrance:
I redid this PR based on @mahrud's [comment](https://github.com/Macaulay2/M2/pull/2707#issuecomment-1336040786).   If Macaulay2 is built w/o libffi support, then just a tiny stub package is loaded with a single documentation page that says that the package doesn't work.

.github/workflows/test_build.yml
```diff
@@ -134,7 +134,8 @@ jobs:
             -DCMAKE_PREFIX_PATH="`brew --prefix`;`brew --prefix libffi`" \
             -DCMAKE_INSTALL_PREFIX=/usr \
             -DWITH_PYTHON=ON \
-            --debug-trycompile
+            --debug-trycompile \
+            -DWITH_FFI=OFF
```
mahrud, line 138: Why does CMake build without FFI?

d-torrance: This was from a temporary commit to check and make sure that the various builds all worked when disabling libffi.  They did, so I'll remove it here in a moment.

mahrud: Ah I see.

PR #2709, opened 2022-12-06
remove use of sprintf and get d and e directories to compile without warnings

mikestillman:


DanGrayson:
Can you make any sense of the failing build?

mikestillman:
@DanGrayson I had stupidly missed a semicolon on a line in a test that I changed to remove sprintf.  But I never even compiled that one locally.  I'm kind of surprised only one test failed...

Hopefully this last one line change will take care of this!

mikestillman:
@DanGrayson I think we can take this now?

PR #2710, opened 2022-12-09
Raise an error in "factor" if the characteristic is too large

d-torrance:
If we try to call factory's `setCharacteristic` with a characteristic greater than 2^29, then factory aborts, crashing Macaulay2.  So we raise an error instead before calling factory.

### Before
```m2
i1 : R = ZZ/(nextPrime(2^29 + 1))[x]

o1 = R

o1 : PolynomialRing

i2 : factor x
characteristic is too large(max is 2^29)Aborted (core dumped)
```

### After
```m2
i1 : R = ZZ/(nextPrime(2^29 + 1))[x]

o1 = R

o1 : PolynomialRing

i2 : factor x
stdio:3:1:(3): error: characteristic is too large (max is 2^29)

i3 : 
```
Is there anything else other than `factor` that uses factory and might have the same issue?


d-torrance:
I moved the check to the engine so we wouldn't need to duplicate it for any other top-level routines that call factory -- `irreducibleCharacteristicSeries` is another one.

PR #2719, opened 2022-12-25
Corrected the misspelled words in PolyominoIdeals package

j-rizwan:
Thank you for already merging the package we have written.
We have corrected the misspelled words and updated the file.


DanGrayson:
Thanks!

PR #2724, opened 2022-12-29
fixes a (typo) bug introduced in commit 2eb9c32c9a33374

mikestillman:
I should not have merged that branch into development directly.  It did have an interesting bug: a typo in MonomialIdeal::is_equal.  It only gets to this code if the hash codes are the same, which rarely happens.  However it did happen with one of the MonomialOrbit tests.  I have also added in a test that the two ideals in question (which have the same hash code) are in fact different.


mikestillman:
@DanGrayson Is the lint/codespell test expected to fail here?

DanGrayson:
> @DanGrayson Is the lint/codespell test expected to fail here?

Yes, and I've just merged a pull request fixing that.

mikestillman:
@DanGrayson The tests all pass, I'll merge this into development.

PR #2726, opened 2022-12-30
Deduplicate ring generators of R ** R and R ^** n

mahrud:
This commit introduces variable name deduplication, which was demonstrated at the last internals meeting. Along the way, the polynomial ring constructor is simplified and a few bugs in there involving WeylAlgebra variables are resolved, Ring ^** ZZ is added, and the file orderedmonoidrings.m2 is renamed to polyrings.m2, which better reflects what is actually in there.

- simplified the polynomial ring constructor
- separated isPrime and factor
- added variable deduplication for tensor(kk[x], kk[x])
- fixed WeylAlgebra indices in Eliminate
- reverted to storing Weyl algebra variables as symbols
- renamed R.basering to R.BaseRing internally
- improved underscore ring element access
- fixed an issue involving kk[x] followed by kk[x_1]
- simplified Ring ^ List
- introduced Ring^**ZZ and Monoid^**ZZ
- renamed: orderedmonoidrings.m2 -> polyrings.m2

Note that I intentionally moved the last commit to the end to make it easier to review everything else separately.

Here's a demo:
```m2
i1 : A = kk[x_0..x_1];

i2 : A ** A

o2 = kk[x   , x   , x   , x   ]
         0,0   1,0   0,1   1,1

o2 : PolynomialRing

i3 : A ** A ** A

o3 = kk[x   , x   , x   , x   , x ..x ]
         0,0   1,0   0,1   1,1   0   1

o3 : PolynomialRing

i4 : A ** A ** A ** A

o4 = kk[x     , x     , x     , x     , x     , x     , x     , x     ]
         0,0,0   1,0,0   0,1,0   1,1,0   0,0,1   1,0,1   0,1,1   1,1,1

o4 : PolynomialRing
```
Note the `A ^** 3` above is somewhat unexpected. I'm not sure if that is ideal, but it is an improvement.


pzinn:
Can you explain changes made to `promote` and `lift`? (if any)

mahrud:
> Can you explain changes made to `promote` and `lift`? (if any)

I don't think there are any. Nothing comes to mind at least. Looking at the code, there's one new method:
```m2
o2 = -- code for method: lift(RingElement,type of MonoidElement)
     ../../Macaulay2/m2/monoids.m2:91:61-97:10: --source code:
     lift(RingElement, MonoidElement) := MonoidElement => (m, M) -> (
         k := coefficientRing first flattenRing(R := ring m);
         if instance(m, monoid k)
         or instance(m, monoid R)
         or instance(m, R.FlatMonoid)
         then new M from m.RawMonomial
         else error "expected monomial from same monoid")
```
Is there a specific usage you're wondering about?

pzinn:
Probably misread. Saw promote and lift mentioned in changed files, but it
must be from earlier patches

DanGrayson:
There's a conflict that needs to be resolved.

pzinn:
still catching up... watching last M2internals and wondering why the ordering of the multi-indices in variable deduplication is the way it is, since it breaks `runLengthEncoding`, so to speak. i.e., I think the extra index should be the first one, not the last one?

mahrud:
I agree, it would be good if the multi-indices worked with `runLengthEncoding`.

pzinn:
question (not directly about this PR, but related): when one takes the tensor product of two rings A**B, there are natural maps from A to A**B and B to A**B. When the names of variables are disjoint, I guess one can use `sub`, though it's not very clean. But what about if variable names overlap? is there a clean way to define these maps?

mahrud:
I agree that there should be a clean way to do this. I like the way NormalToricVarieties produces [maps to and from products](http://www2.macaulay2.com/Macaulay2/doc/Macaulay2-1.18/share/doc/Macaulay2/NormalToricVarieties/html/___Normal__Toric__Variety_sp_us_sp__Array.html). I've made several changes and rewritten parts of `ringmaps.m2` which I'm planning on making into a pull request soon. Let me know if you have ideas about this or other improvements.

pzinn:
OK let me expand my two suggestions:
* about deduplication about variables, I think it should ignore indices.
at the simplest level that would mean `QQ[x_1,x_2,x_3] ** Q[x]` should be `QQ[x_(0,1),x_(0,2),x_(0,3),x_1]` and similarly `QQ[x_1,x_2] ** QQ[x_1,x_2,x_3]` should be `QQ[x_(0,1),x_(0,2),x_(1,1),x_(1,2),x_(1,3)]` (not sure about starting the numbering at 0, but that's a detail). that naive strategy would still fail on triple tensor products because it would give something like (I forget the order) `QQ[x] ** QQ[x] ** QQ[x] = QQ[x_(0,0),x_(0,1),x_1]`.
so even better would be `QQ[x_1,x_2,x_3] ** Q[x] = QQ[x_1,x_2,x_3,x_4]` (i.e., it would try to match the number of indices), and then  `QQ[x] ** QQ[x] ** QQ[x] = QQ[x_0,x_1,x_2]`.
* maybe the natural maps `A->A**B` and `B->A**B` can be in the cache of either the source or the target ring(s)?

PR #2729, opened 2023-01-01
Mode agnostic AfterPrint

pzinn:
Right now, there's a lot of repeats in the code involving `AfterPrint` and `AfterNoPrint`. One problem is, the output of `After(No)Print` is the same no matter the mode, yet the interpreter calls different methods (say `{Standard,AfterPrint}` vs `{WebApp,AfterPrint}`). The other is that each AfterPrint method has the same `concatenate(interpreterDepth:"o") << lineNumber` type code at the start.
I've decided to reorganize this. There is now a `Type#AfterPrint` method (and `Type#AfterNoPrint` similarly) which is by default called by both `{Standard,AfterPrint}` and `{WebApp,AfterPrint}` (the two currently working modes of output). One can still define the latter functions directly, so backwards compatibility is preserved. `Type#AfterPrint` does not output directly; rather, it returns an object or sequence of objects to be printed by AfterPrint. (if it returns `null`, then no AfterPrint is produced)
As mentioned above, advantages:
- standardized template for the output (the `o42: `)
- much less redundancy in the code
- mode agnosticity (e.g., recently, some new AfterPrints were defined for monoids and rings, and with the old system I'd have to manually redo it for the WebApp mode).

### DIFF
--- a/M2/Macaulay2/m2/expressions.m2
+++ b/M2/Macaulay2/m2/expressions.m2
@@ -74,6 +74,8 @@ assert Expression := v -> (
     if not val then error toString("assertion failed:" || net v | " is false")
 )
 
+nullf = x -> null
+
 --Holder2 = new WrapperType of Expression			    -- Holder{ printable form, value form }
 --Holder2.synonym = "holder"
 --Holder = new WrapperType of Holder2			    -- Holder{ printable form, value form }, with printable form === value form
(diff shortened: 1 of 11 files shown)


pzinn:
Intresting. It seems I modified (improved?) Matrix's AfterPrint at some point. It used to provide source/target only if they were free. My new code always provides source/target, hence the mismatch in an example. Thoughts on this? I see no reason to hide nonfree source/targets.

mahrud:
Non-free modules can be _huge_ to print, so I don't think it's a good idea to always print them if someone wanted to look at the entries of the matrix representing a map between them.

pzinn:
fine. reverted.

PR #2732, opened 2023-01-03
Only use __has_builtin if it's defined

d-torrance:
It wasn't introduced until GCC 10, but older versions are still prevalent, e.g., GCC 7 in Ubuntu 18.04 and GCC 9 in Ubuntu 20.04.

### DIFF
--- a/M2/Macaulay2/e/overflow.hpp
+++ b/M2/Macaulay2/e/overflow.hpp
@@ -22,6 +22,12 @@
 #error integer type definitions not available
 #endif
 
+#if defined(__has_builtin)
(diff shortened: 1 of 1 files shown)


mahrud:
GCC has `__builtin_expect` anyway, so could we do
```cpp
#ifdef __GNUC__ OR __has_builtin(__builtin_expect)
```
instead?

d-torrance:
Sounds good!  I just pushed a new version.

To avoid the nested `#if` statements, I defined a `HAS_BUILTIN` macro that potentially could be used elsewhere.  I'm not sure if `overflow.hpp` is the best place for it, though.

Issue #2682, opened 2022-11-14
Error linking with libquadmath


mahrud:
I kept getting linking errors for a while, and I just noticed that it's also causing issues building brew bottles on ubuntu:
```
/home/linuxbrew/.linuxbrew/opt/binutils/bin/ld: Macaulay2/d/CMakeFiles/M2-interpreter.dir/boostmath-tmp.cc.o: undefined reference to symbol 'ldexpq@@QUADMATH_1.0'
/home/linuxbrew/.linuxbrew/opt/binutils/bin/ld: /home/linuxbrew/.linuxbrew/opt/gcc/lib/gcc/current/libquadmath.so.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
ninja: build stopped: subcommand failed.
```
On my machine linking with `quadmath` resolves this, but apparently this fails on MacOS. @d-torrance any ideas what might be behind this?

d-torrance:
Does defining the `BOOST_MATH_DISABLE_FLOAT128` macro help at all?  We use MFPR for all the Boost special functions, so there's no reason to link against libquadmath for 128-bit floats.

mahrud:
It doesn't seem like it. Can you try seeing if your executables are linked with `libquadmath`, and if so does passing that actually remove it?

d-torrance:
I was able to reproduce this using brew's Boost, which is version 1.80.  The Debian/Ubuntu packages are still at 1.74, which is probably why I never saw this (plus the autotools build was already adding `-lquadmath` to the linker flags -- I'm not 100% sure why...)

This is https://github.com/boostorg/math/issues/733, and linking with `-lquadmath` appears to be the correct solution.

mahrud:
Not the ideal fix, and I'm not sure if it is universal, but closing this for now.

PR #2736, opened 2023-01-05
Link against libquadmath for Boost Math Toolkit (autotools)

d-torrance:
This is an autotools fix for #2682, which I finally ran into during a Debian build (they're starting to transition from Boost 1.74 to 1.81).

### DIFF
--- a/M2/configure.ac
+++ b/M2/configure.ac
@@ -1395,6 +1395,7 @@ LIBS="$LIBS $BOOST_REGEX_LIB"
 
 AC_CHECK_HEADERS([boost/math/special_functions.hpp], [],
     [AC_MSG_ERROR([Boost Math Toolkit not available])])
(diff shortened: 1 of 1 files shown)


mahrud:
Is the boost math library worth it? I feel like it might be more worthwhile to go towards using Arb, which has many [hypergeometric functions](https://arblib.org/acb_hypgeom.html) and I presume is more geared towards math software usage than boost (and anecdotally I think it is faster).

mahrud:
In particular, I would love to see FFI level integration with something like Arb, so that people can more easily call whatever hypergeometric functions they need rather than having to recompile all of M2!

d-torrance:
I choose Boost initially because it had some of the special functions I wanted that weren't already included in MPFR and we were already using it for a few other things like regexes and stacktraces.

I haven't dealt with Arb at all, but it looks very cool!

PR #2739, opened 2023-01-08
LieTypes major update / fix #2730

### LINKED ISSUE #2730: LieTypes tensor product bug
Note the error last line. Seems the `**` routine assumes the modules are irreducible. @dswinarski 

```
i2 :  needsPackage "LieTypes"

o2 = LieTypes

o2 : Package

i4 : g=simpleLieAlgebra("A",2)

o4 = g

o4 : LieAlgebra

i5 : U=irreducibleLieAlgebraModule({4,2},g)

o5 = U

o5 : g module

i6 : V=irreducibleLieAlgebraModule({3,1},g)

o6 = V

o6 : g module

i7 : U**V

o7 = {LieAlgebra => g                                        }
      isIrreducible => false
      DecompositionIntoIrreducibles => HashTable{{0, 5} => 1}
                                                 {1, 3} => 1
                                                 {1, 6} => 1
                                                 {2, 1} => 1
                                                 {2, 4} => 2
                                                 {3, 2} => 2
                                                 {3, 5} => 1
                                                 {4, 0} => 1
                                                 {4, 3} => 2
                                                 {5, 1} => 2
                                                 {5, 4} => 1
                                                 {6, 2} => 2
                                                 {7, 0} => 1
                                                 {7, 3} => 1
                                                 {8, 1} => 1

o7 : g module

i8 : oo**V
stdio:9:3:(3): error: key not found in hash table
```

pzinn:
I've rewritten quite a bit of the package `LieTypes`:
* Improved output methods
* Introduced the shorthand `LL_(...) (g)` for `irreducibleLieAlgebraModule({...},g)`
* Fixed and exported `LieAlgebraModuleFromWeights`
* Fixed and optimized tensor product of modules (cf #2730)
There's a lot to say here, but to make a long story short, I improved the search for a Weyl group conjugate in the dominant chamber, and also I improved `weightDiagram`, see `character` below.
* Added `^**`, `symmetricPower`, `exteriorPower`, `adams` and `^` for various operations on Lie algebra modules
* Added `trivialModule`
* Additional sanity checks
* Allow inputting weights as vectors
* `isIrreducible` is now a method (no reason to have a field for it when it can be recomputed on the fly)
* use of `VirtualTally` rather than `HashTable` to take advantage of its methods (for the decomposition into irreducibles, and for weights)
* Added `character` method which computes the character of a module: this is strictly equivalent to `weightDiagram` and in fact now `weightDiagram` just applies `listForm` to `character`. I've played around with various stategies to compute characters depending on the type. I kept intact the original code for Freudenthal's method, but added 3 more strategies, all of which are coded via hooks and accessible with `Strategy => ...` Surprisingly Freudenthal's method is almost never the fastest except maybe for small rank algebras and small highest weights, for which it doesn't really matter anyway.

As usual, the code is already available on the cloud, cf https://www.unimelb-macaulay2.cloud.edu.au/?user=LieTypes

what I will provide in the next commit:
* `fusionProduct` suffers from the same problems as tensor product, will fix.
* I will remove the method `isIsomorphic` as it is not really useful: in fact it's strictly equivalent to `===` (there's no fundamental distinction between equality and isomorphism of `LieAlgebraModule`s...)
* provide Dynkin diagrams with the numbering of the vertices (either in the doc or via a method).

### DIFF
--- a/M2/Macaulay2/packages/LieTypes.m2
+++ b/M2/Macaulay2/packages/LieTypes.m2
@@ -10,7 +10,6 @@ newPackage(
 	  },
     Keywords => {"Lie Groups and Lie Algebras"},
     PackageImports => {"ReesAlgebra"},
-    DebuggingMode => true,
     Certification => {
 	 -- same article as for package ConformalBlocks
 	  "journal name" => "The Journal of Software for Algebra and Geometry",
@@ -170,9 +169,12 @@ simpleLieAlgebra(String,ZZ) := (type,m) -> (
     if type=="E" and not member(m,{6,7,8}) then error "The rank for type E must be 6, 7, or 8.";
     if type=="F" and m!=4 then error "The rank for type F must be 4.";
     if type=="G" and m!=2 then error "The rank for type G must be 2.";
+    Q:=sum entries quadraticFormMatrix(type,m);
+    l:=lcm(denominator\Q);
+    Q=apply(Q,q->lift(q*l,ZZ));
+    x:=getSymbol "x";
     new LieAlgebra from {"LieAlgebraRank"=>m,"RootSystemType"=>type,"isSimple"=>true,cache=>new CacheTable from {
-	    x:=getSymbol "x";
-	    "Ring" => ZZ(monoid [x_1..x_m,Inverses=>true,MonomialOrder=>Lex])}}
+	    "Ring" => ZZ(monoid [x_1..x_m,Inverses=>true,MonomialOrder=>{Weights=>Q,Lex}])}}
     )
 -*simpleLieAlgebra(IndexedVariable) := (v) -> (
     if #v > 2 or not member(v#0,{symbol sl, symbol so, symbol sp}) or not instance(v#1,ZZ) then error "Input not understood; enter sl_k, sp_k, or so_k, or use the syntax simpleLieAlgebra(\"A\",1) instead";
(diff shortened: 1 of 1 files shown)


pzinn:
actually that test failing is related to the `isIsomorphic` comment: I added at the last minute a test comparing two `LieAlgebraModule`s with `==` I think it would make sense to define `==` as `===` for them.

mahrud:
Your pull request #2729 reverted the emacs submodule from 1.21 to 1.19, and now this pull request is moving it to 1.20. Could you fix the hash to be 
[86150290208bd9e468511b4b69a187ce870f303e](https://github.com/Macaulay2/M2-emacs/tree/86150290208bd9e468511b4b69a187ce870f303e)?

pzinn:
this emacs thing is a pain. fixing.

pzinn:
I've thrown in a few more methods

DanGrayson:
The diff still shows that you're modifying the emacs submodule.

pzinn:
concerning emacs, I followed Mahrud's instructions.
BTW please don't merge this PR yet, still talking to Dave @dswinarski about this update.

pzinn:
I'm waiting for #2750 to be merged before updating this PR, since `LieTypes` suffers from the same issue mentioned in that other PR description.

PR #2745, opened 2023-01-11
Update PolyominoIdeals.m2

fchapoton:
fixing a misspelled word that breaks the codespell linter


fchapoton:
@DanGrayson : this should be an easy instant merge, please

PR #2747, opened 2023-01-13
Add capitals

fchapoton:
to some names of people inside packages/

### DIFF
--- a/M2/Macaulay2/packages/Book3264Examples.m2
+++ b/M2/Macaulay2/packages/Book3264Examples.m2
@@ -651,7 +651,7 @@ doc ///
       equivalently calculate the degree of $\alpha \cap [X]$.  In this way we ``push the problem
       forward'' to ${\mathbb P}^n$.
       
-      As an example, if we want to calculate the degree of the top chern class of
+      As an example, if we want to calculate the degree of the top Chern class of
       the tangent bundle to a hypersurface $X$ of degree $4$ in ${\mathbb P}^3$, we can compute:
     Example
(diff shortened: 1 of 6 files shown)


fchapoton:
why is there no codespell linter check ?

DanGrayson:
> why is there no codespell linter check ?

That's strange, there should be.  I'll restart all the jobs.

DanGrayson:
> > why is there no codespell linter check ?
> 
> That's strange, there should be. I'll restart all the jobs.

No, it only offers to rerun the jobs that were already run.

DanGrayson:
> why is there no codespell linter check ?

Is your code based on an old branch, before that lint test was added?

fchapoton:
no, my code is based on the most recent development branch

but I made a mistake and first asked for "master" as the target branch, before changing it ; maybe that's the reason

DanGrayson:
> no, my code is based on the most recent development branch
> 
> but I made a mistake and first asked for "master" as the target branch, before changing it ; maybe that's the reason

That's probably not the reason -- a pull request could change the workflow, and that should be tested.

I've restarted all the jobs, anyway.  We'll see.

d-torrance:
It only runs on pushes or pull requests to `development`, so I think that's indeed the reason:

https://github.com/Macaulay2/M2/blob/048c777dc703365e44c3865b2c1faed698b82b3a/.github/workflows/lint.yml#L2-L4

DanGrayson:
Thanks again for all the work on improving such things!

Issue #2757, opened 2023-02-01
non-Hypertext inside Hypertext, validate discussion


pzinn:
This is not so much an issue as a call for discussion about the following.
Ever since #1288 (2.5 years ago), the role of `html` has changed quite a bit. It used to be that `html` would only accept instances of `Hypertext` as argument, and would produce the corresponding HTML code. After a series of PRs following #1288, `html` now accepts any type and converts it to HTML code according to the following recipe: by default (in particular for all mathematical types), it does something like `htmlLiteral @@ tex` where `htmlLiteral` makes sure characters that have a special meaning in HTML are encoded (the implicit assumption is that something like KaTeX will be running and concert it); for some other types (including, but not limited to, `Hypertext`) it just produces HTML code directly.
There's one problem with this new paradigm which I uncovered as part of my current improvements to `locate`, `code`, `listSymbols` (in relation to   #2655 and  #2375, among other things): it can lead to `Hypertext` which is considered illegal by `validate`. Let me give an example (which is relevant to my new, improved, `listSymbols`): suppose you want to create an HTML `TABLE` whose contents can have some mathematical stuff, say a ring `QQ[x]`. Currently the only way to produce this is with something roughly like
```
TABLE {{QQ[x]}}
```
 (note that you *can't* write `TABLE {{ html(QQ[x]) }}` -- that would doubly `htmlLiteral` the stuff inside `html(...)` which in general will not produce the desired result).
Now this will work perfectly fine -- except if you tried to `validate` this (e.g., to include it in documentation), it will tell you `QQ[x]` has an invalid type inside a `Hypertext` element.
Which is a valid point -- with the "old" logic that is still enforced by `validate`, one can't have non `Hypertext` inside `Hypertext` (well, except strings and options, obviously).
So at some point in the near future we will have to make a choice:
- either embrace the new paradigm and accept non-`Hypertext` inside `Hypertext`. This is the easier option, and the one I would probably favor, except it means getting rid or changing drastically how `validate` works. There's a whole bunch of related nasty problems (what if stuff inside a `Hypertext` is say local variables? can't `toExtternalString` it...) which will need to be resolved.
[...]

Issue #2754, opened 2023-01-30
<br/> not respected within certain HTML elements in conversion to Net


pzinn:
`html SPAN {"a", BR{}, "b"}` produces (as should be) `<span> a <br/> b </span>`, which results in a & b on separate lines. however, `net SPAN {"a", BR{}, "b"}` results in `ab` (no line break).
In general, it seems the conversion of HTML to Net ignores `BR` within tags which are not of the type `HypertextContainer` or `HypertextParagraph` (as opposed to just `Hypertext`, not that these distinctions of type make much sense in the first place). This does not match the intended effect (e.g. in a browser) and should therefore be considered a bug.
The relevant code is the rather confusing `parser` stuff in `format.m2`.

pzinn:
among the many related issues with the `parser` part of `format.m2`: in a browser, `DIV { DIV 1, DIV 2 }` and `DIV { DIV 1, DIV PRE 2 }` will look exactly the same up to font choice. However, once converted to `Net`, the first one is 1 and then 2 below, as should be, but the second one inserts an extra empty line in between. I have not found a way to prevent this extra line break yet.

d-torrance:
Fixed in #4272

Issue #2655, opened 2022-11-01
Proper formatting of source code


pzinn:
This is a minor issue, but the code in the documentation is not quite correctly formatted.
consider `help exponents` (or https://www.unimelb-macaulay2.cloud.edu.au/usr/share/doc/Macaulay2/Macaulay2Doc/html/_exponents.html)
The code section is produced with `M2CODE`, so in HTML is translated into a `<pre><code language="Macaulay2">...</code></pre>`.
so far, so good. but the code is invalid: it's
```
../../../../Macaulay2/m2/monoids.m2:120:30-120:50: --source code:
exponents RingElement   := f -> first \ listForm f
```
Note that the first line is not actually M2 code.
There's a quick and easy way to fix this: move `--source code:` to the start of the line.
But I think a better way would be proper formatting of the output of the command `code`, which should neatly separate the source location (possibly using the resurrected `FilePosition`) and the actual code.
In particular I'd love to have both `error` and `code` tag separately source location to make it easier to jump to the location in question in your editor.

DanGrayson:
If you modify the format of that line, you'll also have to modify the code for the emacs function `M2-send-to-program-or-jump-to-source-code`, which will jump to the source code when RETURN is pressed on that line.

By the way, the source code position is wrong at my end -- it points to a non-existent file.  I wonder how that got screwed up.  It hasn't worked since the feature appeared in version 1.17.

pzinn:
> If you modify the format of that line, you'll also have to modify the code for the emacs function `M2-send-to-program-or-jump-to-source-code`, which will jump to the source code when RETURN is pressed on that line.

that would be a small price to pay.

Issue #2569, opened 2022-08-17
locate and code for functions involving options, caching, or @@ are useless


mahrud:
Say I want to see where `net Variety` is defined, so I use `locate`:
```m2
i5 : locate(net, Variety)

o5 = (../../Macaulay2/m2/classes.m2, 90, 47, 90, 54, 90, 50)
```
But that doesn't make sense, so instead I look up its code:
```m2
i8 : code(net, Variety)

o8 = -- code for method: net(Variety)
     ../../Macaulay2/m2/classes.m2:90:48-90:55: --source code:
     Function @@ Function := Function => (f,g) -> x -> f g x
     | symbol   class                   value      location of symbol
     | ------   -----                   -----      ------------------                       
     | f      : MethodFunctionSingle -- net        ../../Macaulay2/m2/classes.m2:90:38-90:39
     | g      : MethodFunctionSingle -- expression ../../Macaulay2/m2/classes.m2:90:40-90:41
     | -- function f:
     | function net: source code not available
     | -- function g:
     | function expression: source code not available
```
But this is useless: nowhere in this output does it say that `net Variety` is defined on line 24 of `varieties.m2`, or even what `net` and `expression` do or even where they are defined -- it just says the are also in `classes.m2`! Even the most useful part of this output, which says that `net Variety` is the composition of `net` and `expression`, is written in too much generality and would be much easier to understand if it just said something like `x -> net expression x`.

To a smaller extent, the same is true for functions involving options:
```m2
i17 : locate(codim, ProjectiveVariety)

o17 = (../../Macaulay2/m2/option.m2, 15, 19, 17, 33, 17, 7)

o17 : Sequence

i18 : code(codim, ProjectiveVariety)

o18 = -- code for method: codim(ProjectiveVariety)
      ../../Macaulay2/m2/option.m2:15:20-19:34: --source code:
        (opts,f) -> args -> (
             -- Common code for functions created with >> to process options and arguments.
             uncurry(f, override (opts,args))
             )
        )
      | symbol   class              value                         location of symbol
      | ------   -----              -----                         ------------------                     
      | f      : FunctionClosure -- ...                           ../../Macaulay2/m2/option.m2:15:9-15:10
      | opts   : OptionTable     -- OptionTable{Generic => false} ../../Macaulay2/m2/option.m2:15:4-15:8 
      | -- function f:
      | ../../Macaulay2/m2/varieties.m2:299:66-299:87: --source code:
[...]

PR #2760, opened 2023-02-03
locate, code, html updates

pzinn:
This is a continuation of #2421. Besides #2375, it also addresses #2655, #2777, and is tangentially related to #2754 and #2757.
phew. overall, the goal of this PR is to make several methods related to debugging more "semantic" by producing more meaningful output than simply a `Net` or `String`.
- `locate` now outputs a `FilePosition`, and this in turn helps remove the huge redundancy in formatting output of the form `filename:0:0-42:42` which is all over the place in the code. not entirely finished (also, I'm not entirely happy with the way `locate` internally finds the position of code, but that's for another PR -- see comments below and related issue #2569 )
- `code` and `showSymbols` now output HTML rather than a `Net`.
- the `AfterPrint` of `Hypertext` types is suppressed (following @mahrud 's suggestion)
- in relation to #2757, I've decided to go ahead with my idea to hijack the `hypertext` method as an intemediate step to producing html. It's strictly speaking not necessary but I think the code is much cleaner this way; there's no obligation for anyone to use this and the overall effect is at this stage invisible, but since `hypertext` already existed, there's no reason not to use it.
- still, in relation to #2757, I have *not* touched documentation at all. So these changes should not have any impact on it even though I'm already experimenting with this on the web app, but I think it's too early to PR.
- there's a couple of assorted minor fixes, like missing `validContent` and some `short` changes (still not ready for export).
- one thing I am not happy about at this stage is various hacks I had to use because of the oddities of converting `Hypertext` to `net` in `format.m2` (which is relevant since several methods now output `Hypertext` and in standard mode, it gets converted to `net`). Eventually, I want to rewrite some of `format.m2` so that it behaves more reasonably, but at this stage, I had to create another of these `IntermediateMarkupType`s (that we'll need to get rid of at some point) just to have `DIV` with an extra vertical bar on the left. Not great but only way to circumvent the non-recursivity of the `parser` of `format.m2`. I also had to hack the spacing of `PRE`, see #2754.
- #2655 is now moot because now the output of code is properly semantic, there's a clear separation between the code itself and what precedes it, so it's irrelevant if the latter is not valid code itself.

### DIFF
--- a/M2/Macaulay2/m2/code.m2
+++ b/M2/Macaulay2/m2/code.m2
@@ -98,9 +98,8 @@ editMethod String := filename -> (
      chkrun concatenate(
 	  if getenv "DISPLAY" != "" and editor != "emacs" then "xterm -e ",
 	  editor, " ", filename))
-EDIT = method(Dispatch => Thing)
-EDIT Nothing := arg -> (stderr << "--warning: source code not available" << endl;)
-EDIT FilePosition := x -> (
+editMethod Nothing := arg -> (stderr << "--warning: source code not available" << endl;)
+editMethod FilePosition := x -> (
      filename := x#0; start := x#1;
      editor := getViewer("EDITOR", "emacs");
      if 0 != chkrun concatenate(
@@ -111,7 +110,7 @@ EDIT FilePosition := x -> (
 	  filename
 	  ) then error "command returned error code")
 editMethod Command := c -> editMethod c#0
-editMethod Function := args -> EDIT locate args
+editMethod Function := args -> editMethod locate args
 editMethod Sequence := args -> (
     if args === () then (
 	editor := getViewer("EDITOR", "emacs");
@@ -119,7 +118,7 @@ editMethod Sequence := args -> (
 	    if getenv "DISPLAY" != "" and editor != "emacs" then "xterm -e ",
 	    editor)
 	)
-    else EDIT locate args
+    else editMethod locate args
      )
 editMethod ZZ := i -> editMethod previousMethodsFound#i
 edit = Command editMethod
--- a/M2/Macaulay2/m2/code.m2
+++ b/M2/Macaulay2/m2/code.m2
@@ -1,6 +1,6 @@
 --		Copyright 1993-1999, 2008 by Daniel R. Grayson
 
--- TODO: needs "document.m2" for formatDocumentTag, but this casues a loop
+-- TODO: needs "document.m2" for formatDocumentTag, but this causes a loop
 needs "debugging.m2" -- for FilePosition
 needs "gateway.m2"
 needs "lists.m2"
@@ -11,9 +11,35 @@ needs "nets.m2"
 -- code
 -----------------------------------------------------------------------------
 
-getSourceLines = method(Dispatch => Thing) 
-getSourceLines Nothing := null -> null
-getSourceLines FilePosition := x -> (
+limit := 4
+
+codeFunction := (f,depth) -> (
+    if depth <= limit then (
+	l := locate f;
+	if l === null then DIV{"function ", f, ": source code not available"}
+	else (
+	    syms := flatten \\ sortByHash \ values \ drop(localDictionaries f,-1);
+	    DIV flatten {
+		code l,
+		if #syms > 0 then INDENT listSymbols syms,
+		if codeHelper#?(functionBody f)
+		then apply(
+		    codeHelper#(functionBody f) f,
+		    (comment,val) -> INDENT {
+			comment, BR{},
+			if instance(val, Function) then codeFunction(val,depth+1) else hold val -- hold for OptionTable or Option
+			})
+	      	}
+	    )
+      	)
+    )
+
+-- stores previously listed methods, hooks, or tests to be used by (code, ZZ)
+previousMethodsFound = null
+
+code = method(Dispatch => Thing)
+code Nothing    := identity
+code FilePosition := x -> (
     filename := x#0; start := x#1; stop := x#3;
      if filename =!= "stdio" then (
 	  wp := set characters " \t\r);";
@@ -40,37 +66,8 @@ getSourceLines FilePosition := x -> (
 	      PRE M2CODE concatenate between_"\n" toList apply(start-1 .. stop-1, i -> file#i)
 	      }
 	  ))
-
-limit := 4
-
-codeFunction := (f,depth) -> (
-    if depth <= limit then (
-	l := locate f;
-	if l === null then DIV{"function ", f, ": source code not available"}
-	else (
-	    syms := flatten \\ sortByHash \ values \ drop(localDictionaries f,-1);
-	    DIV flatten {
-		getSourceLines l,
-		if #syms > 0 then INDENT listSymbols syms,
-		if codeHelper#?(functionBody f)
-		then apply(
-		    codeHelper#(functionBody f) f,
-		    (comment,val) -> INDENT {
-			comment, BR{},
-			if instance(val, Function) then codeFunction(val,depth+1) else hold val -- hold for OptionTable or Option
-			})
-	      	}
-	    )
-      	)
-    )
-
--- stores previously listed methods, hooks, or tests to be used by (code, ZZ)
-previousMethodsFound = null
-
-code = method(Dispatch => Thing)
-code Nothing    := identity
 code Symbol     :=
-code Pseudocode := s -> getSourceLines locate s
+code Pseudocode := s -> code locate s
 code Sequence   := s -> (
     key := select(s, x -> not instance(x, Option));
     -- handle strategies
--- a/M2/Macaulay2/m2/code.m2
+++ b/M2/Macaulay2/m2/code.m2
@@ -12,8 +12,9 @@ needs "nets.m2"
 
 getSourceLines = method(Dispatch => Thing) 
 getSourceLines Nothing := null -> null
-getSourceLines Sequence := x -> (
-     (filename,start,startcol,stop,stopcol,pos,poscol) -> if filename =!= "stdio" then (
+getSourceLines FilePosition := x -> (
+    filename := x#0; start := x#1; stop := x#3;
+     if filename =!= "stdio" then (
 	  wp := set characters " \t\r);";
 	  file := (
 	       if match("startup.m2.in$", filename) then startupString
@@ -33,33 +34,34 @@ getSourceLines Sequence := x -> (
 	       ) do stop = stop + 1;
 	  if #file < stop then error("line number ",toString stop, " not found in file ", filename);
 	  while stop >= start and file#(stop-1) === "" do stop = stop-1;
-	  stack prepend(
-	       concatenate(filename, ":", 
-		    toString start, ":", toString (startcol+1),
-		    "-",
-		    toString stop, ":", toString (stopcol+1),
-		    ": --source code:"),
-	       apply(start-1 .. stop-1, i -> file#i)
-	       )
-	  )) x
+	  DIV {
+	      x, ": --source code",
+	      PRE M2CODE concatenate between_"\n" toList apply(start-1 .. stop-1, i -> file#i)
+	      }
+	  ))
 
 limit := 4
-indent := n -> "| "^(height n, depth n) | n
 
 codeFunction := (f,depth) -> (
      if depth <= limit then (
-	  if locate f === null then concatenate("function ", toString f, ": source code not available")
-	  else stack(
-	       syms := flatten \\ sortByHash \ values \ drop(localDictionaries f,-1);
-	       getSourceLines locate f,
-	       if #syms > 0 then indent listSymbols syms,
-	       if codeHelper#?(functionBody f) 
-	       then toSequence apply(
-		    codeHelper#(functionBody f) f, 
-		    (comment,val) -> indent stack (
-			      comment, 
-			      if instance(val, Function) then codeFunction(val,depth+1) else net val
-			      )))))
+	 l := locate f;
+	  if l === null then DIV{"function ", f, ": source code not available"}
+	  else (
+	      syms := flatten \\ sortByHash \ values \ drop(localDictionaries f,-1);
+	      DIV flatten {
+		  getSourceLines l,
+	       	  if #syms > 0 then INDENT listSymbols syms,
+	       	  if codeHelper#?(functionBody f)
+	       	  then apply(
+		      codeHelper#(functionBody f) f,
+		      (comment,val) -> INDENT {
+			  comment, BR{},
+			  if instance(val, Function) then codeFunction(val,depth+1) else hold val -- hold for OptionTable or Option
+			  })
+	      }
+	  )
+      )
+  )
 
 -- stores previously listed methods, hooks, or tests to be used by (code, ZZ)
 previousMethodsFound = null
@@ -80,11 +82,11 @@ code Sequence   := s -> (
 	    and store#key.HookAlgorithms#?strategy
 	    then store#key.HookAlgorithms#strategy));
     if func =!= null or (func = lookup key) =!= null
-    then "-- code for method: "          | formatDocumentTag key || code func
+    then DIV {"-- code for method: " | formatDocumentTag key, code func }
     else "-- no method function found: " | formatDocumentTag key)
 code Function   := f -> codeFunction(f, 0)
 code Command    := C -> code C#0
-code List       := L -> stack between_"---------------------------------" apply(L, code)
+code List       := L -> DIV between_(HR{}) apply(L, code)
 code ZZ         := i -> code previousMethodsFound#i
 
 -----------------------------------------------------------------------------
@@ -100,7 +102,8 @@ editMethod String := filename -> (
 	  editor, " ", filename))
 EDIT = method(Dispatch => Thing)
 EDIT Nothing := arg -> (stderr << "--warning: source code not available" << endl;)
-EDIT Sequence := x -> ((filename,start,startcol,stop,stopcol,pos,poscol) -> (
+EDIT FilePosition := x -> (
+     filename := x#0; start := x#1;
      editor := getViewer("EDITOR", "emacs");
      if 0 != chkrun concatenate(
 	  if getenv "DISPLAY" != "" and editor != "emacs" then "xterm -e ",
@@ -108,16 +111,17 @@ EDIT Sequence := x -> ((filename,start,startcol,stop,stopcol,pos,poscol) -> (
 	  " +",toString start,
 	  " ",
 	  filename
-	  ) then error "command returned error code")) x
+	  ) then error "command returned error code")
 editMethod Command := c -> editMethod c#0
 editMethod Function := args -> EDIT locate args
 editMethod Sequence := args -> (
-     editor := getViewer("EDITOR", "emacs");
-     if args === () 
-     then chkrun concatenate(
-	  if getenv "DISPLAY" != "" and editor != "emacs" then "xterm -e ",
-	  editor)
-     else EDIT locate args
+    if args === () then (
+	editor := getViewer("EDITOR", "emacs");
+	chkrun concatenate(
+	    if getenv "DISPLAY" != "" and editor != "emacs" then "xterm -e ",
+	    editor)
+	)
+    else EDIT locate args
      )
 editMethod ZZ := i -> editMethod previousMethodsFound#i
 edit = Command editMethod
@@ -252,7 +256,7 @@ debuggerHook = entering -> (
      if entering then (
 	  pushvar(symbol inDebugger, true);
 	  c := code current;
-	  if c =!= null then << c << endl;
+	  if c =!= null then print c;
 	  )
      else (
 	  popvar symbol inDebugger;
--- a/M2/Macaulay2/m2/code.m2
+++ b/M2/Macaulay2/m2/code.m2
@@ -35,7 +35,7 @@ getSourceLines FilePosition := x -> (
 	  if #file < stop then error("line number ",toString stop, " not found in file ", filename);
 	  while stop >= start and file#(stop-1) === "" do stop = stop-1;
 	  DIV {
-	      x, ": --source code",
+	      x, ": --source code:",
 	      PRE M2CODE concatenate between_"\n" toList apply(start-1 .. stop-1, i -> file#i)
 	      }
 	  ))
@@ -43,25 +43,25 @@ getSourceLines FilePosition := x -> (
 limit := 4
 
 codeFunction := (f,depth) -> (
-     if depth <= limit then (
-	 l := locate f;
-	  if l === null then DIV{"function ", f, ": source code not available"}
-	  else (
-	      syms := flatten \\ sortByHash \ values \ drop(localDictionaries f,-1);
-	      DIV flatten {
-		  getSourceLines l,
-	       	  if #syms > 0 then INDENT listSymbols syms,
-	       	  if codeHelper#?(functionBody f)
-	       	  then apply(
-		      codeHelper#(functionBody f) f,
-		      (comment,val) -> INDENT {
-			  comment, BR{},
-			  if instance(val, Function) then codeFunction(val,depth+1) else hold val -- hold for OptionTable or Option
-			  })
-	      }
-	  )
-      )
-  )
+    if depth <= limit then (
+	l := locate f;
+	if l === null then DIV{"function ", f, ": source code not available"}
+	else (
+	    syms := flatten \\ sortByHash \ values \ drop(localDictionaries f,-1);
+	    DIV flatten {
+		getSourceLines l,
+		if #syms > 0 then INDENT listSymbols syms,
+		if codeHelper#?(functionBody f)
+		then apply(
+		    codeHelper#(functionBody f) f,
+		    (comment,val) -> INDENT {
+			comment, BR{},
+			if instance(val, Function) then codeFunction(val,depth+1) else hold val -- hold for OptionTable or Option
+			})
+	      	}
+	    )
+      	)
+    )
 
 -- stores previously listed methods, hooks, or tests to be used by (code, ZZ)
 previousMethodsFound = null
--- a/M2/Macaulay2/m2/debugging.m2
+++ b/M2/Macaulay2/m2/debugging.m2
@@ -3,11 +3,6 @@
 needs "nets.m2"
 needs "methods.m2"
 
-symbolLocation = s -> (
-     t := locate s;
-     if t =!= null then t#0 | ":" | toString t#1| ":" | toString (t#2+1) | "-" | toString t#3| ":" | toString (t#4+1)
-     else "")
-
 processArgs := args -> concatenate (
      args = sequence args;
      apply(args, x -> 
@@ -15,7 +10,7 @@ processArgs := args -> concatenate (
 	  else if class x === Symbol then ("'", toString x, "'")
 	  else silentRobustString(40,3,x)
 	  ),
-     apply(args, x -> if class x === Symbol then ("\n", symbolLocation x, ": here is the first use of '",toString x, "'") else "")
+     apply(args, x -> if class x === Symbol then ("\n", locate x, ": here is the first use of '",toString x, "'") else "")
      )
 olderror := error
 error = args -> (
@@ -139,7 +134,7 @@ show1(Sequence,Function) := show1(List,Function) := (types,pfun) -> (
 	       else install pfun v);
 --	  if hasAttribute(v,PrintNet) then v = getAttribute(v,PrintNet) else
 --	  if hasAttribute(v,PrintNames) then v = getAttribute(v,PrintNames) else
-	  if hasAttribute(v,ReverseDictionary) then v = getAttribute(v,ReverseDictionary);
+--	  if hasAttribute(v,ReverseDictionary) then v = getAttribute(v,ReverseDictionary);
 	  if w#?v then w#v else w#v = new Descent
 	  );
      scan(types, install);
@@ -187,16 +182,13 @@ localSymbols(Type,Pseudocode) := (X,f) -> select2(X,localSymbols f)
 
 localSymbols Type := X -> select2(X,localSymbols ())
 
-robust := y -> silentRobustNet(55,4,3,y)
-abbreviate := x -> (
-     if instance(x, Function) and match("^-\\*Function.*\\*-$", toString x) then "..."
-     else robust x)
 listSymbols = method()
 listSymbols Dictionary := d -> listSymbols values d
-listSymbols List := x -> (
-     netList(Boxes=>false, HorizontalSpace => 1, prepend(
-	  {"symbol" || "------","", "class" || "-----", "", "value" || "-----", "location of symbol" || "------------------"},
-	  apply (x, s -> {toString s,":", robust class value s, "--", abbreviate value s, symbolLocation s}))))
+listSymbols List := x -> TABLE prepend(
+    apply({"symbol", "class", "value", "location of symbol"},s->TH {s}),
+    apply(x, y -> apply({y,Abbreviate class value y,Abbreviate value y,locate y},s->TD {s}))
+    );
+
 
 listLocalSymbols = Command(f -> listSymbols localSymbols f)
 
--- a/M2/Macaulay2/m2/debugging.m2
+++ b/M2/Macaulay2/m2/debugging.m2
@@ -10,7 +10,7 @@ processArgs := args -> concatenate (
 	  else if class x === Symbol then ("'", toString x, "'")
 	  else silentRobustString(40,3,x)
 	  ),
-     apply(args, x -> if class x === Symbol then ("\n", locate x, ": here is the first use of '",toString x, "'") else "")
+     apply(args, x -> if class x === Symbol then ("\n", toString locate x, ": here is the first use of '",toString x, "'") else "")
      )
 olderror := error
 error = args -> (
--- a/M2/Macaulay2/m2/debugging.m2
+++ b/M2/Macaulay2/m2/debugging.m2
@@ -186,7 +186,7 @@ listSymbols = method()
 listSymbols Dictionary := d -> listSymbols values d
 listSymbols List := x -> TABLE prepend(
     apply({"symbol", "class", "value", "location of symbol"},s->TH {s}),
-    apply(x, y -> apply({y,Abbreviate class value y,Abbreviate value y,locate y},s->TD {s}))
+    apply(x, y -> apply({y,Abbreviate {class value y},Abbreviate {value y},locate y},s->TD {s}))
     );
 
 
@@ -234,6 +234,28 @@ generateAssertions List := y -> (
 	       )))^-1
 
 currentPosition = () -> new FilePosition from { currentFileName, currentRowNumber(), currentColumnNumber() }
+
+FilePosition = new Type of BasicList
+FilePosition.synonym = "file position"
+toString FilePosition :=
+net FilePosition := p -> concatenate(
+    p#0,":",toString p#1,":",toString p#2,
+    if #p>3 then ("-",toString p#3,":",toString p#4),
+--    if #p>5 then (" (",toString p#5,":",toString p#6,")")
+    )
+
+locate' = locate -- defined in d/actors4.d
+locate = method(Dispatch => Thing, TypicalValue => FilePosition)
+locate Nothing     :=
+locate FunctionBody:=
+locate Function    :=
+locate Pseudocode  :=
+locate Sequence    :=
+locate Symbol      := FilePosition => x -> if (x':=locate' x) =!= null then new FilePosition from x'
+locate List        := List     => x -> apply(x, locate)
+protect symbol locate
+
+
 
 -- Local Variables:
 -- compile-command: "make -C $M2BUILDDIR/Macaulay2/m2 "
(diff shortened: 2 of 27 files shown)


pzinn:
BTW let me record the various problems with how `locate` internally works:
- `locate` always gets the end point of functions wrong. that's because it works recursively by taking [min,max] of the "positions" of the pieces of the code of the function, but position = right before the first character (except, for strings, oddly, for which it's not defined at all). so the end point will be right before the first character of the last non-string part of the code of the function.
- `locate` is not UTF8-compliant. it will get the location of functions/symbols containing utf8 extended characters wrong.
- `locate` is not thread-safe (OK, minor point at this stage)

pzinn:
I'm stumped by the Normaliz error. works fine locally.

mahrud:
> I'm stumped by the Normaliz error. works fine locally.

It's not your fault, Normaliz 3.10 is running on the macOS actions.

mahrud:
There seem to be a few, perhaps unintended consequences: missing colon, extra blank lines, and the dashes seem to be too long:
```diff
 o1 = -- code for method: determinant(Matrix)
-     Macaulay2/m2/multilin.m2:166:43-174:54: --source code:
+     Macaulay2/m2/multilin.m2:166:42-174:53: --source code
      determinant Matrix := RingElement => opts -> f -> (
           if rank source f != rank target f
           or not isFreeModule source f
@@ -11,9 +11,11 @@ o1 = -- code for method: determinant(Matrix)
             basicDet f
           else 
             (exteriorPower(numgens source f, f, opts))_(0,0))
-     ---------------------------------
+
+     ------------------------------------------------------------------------------------------------------------------------
+
      -- code for method: determinant(MutableMatrix)
-     Macaulay2/m2/mutablemat.m2:307:35-313:26: --source code:
+     Macaulay2/m2/mutablemat.m2:307:34-312:25: --source code
      determinant MutableMatrix := opts -> (M) -> (
          if numRows M =!= numColumns M then error "expected a square matrix";
          if isField ring M then
@@ -21,3 +23,5 @@ o1 = -- code for method: determinant(Matrix)
          else
            determinant matrix M
          )
+
+o1 : DIV
```

Also, I think I would be in favor of hiding the AfterPrint for all hypertext objects, at least for `DIV`: it's kind of annoying that `help` always prints a `DIV` at the end.

pzinn:
The dashes being too long is just the output of `HR`, I haven't chosen it. not sure why it's so long.
same for the extra blank lines, though I think they're fine.
The colon is a typo I guess.
Hiding all hypertext afterprint is an interesting idea; why not?

M2/Macaulay2/packages/Macaulay2Doc/functions/installAssignmentMethod-doc.m2
```diff
@@ -56,7 +56,7 @@ document {
 	  x_c
      ///,
      "The source code explains how it works, behind the scenes.",
-     PRE ("    " | code {(symbol _,Symbol,Thing),((symbol _,symbol =),Symbol,Thing)}),
+     PRE ("    " | net code {(symbol _,Symbol,Thing),((symbol _,symbol =),Symbol,Thing)}),
```
mahrud, line 59: Why do you `net` this here?

pzinn: because at this stage, documentation is not ready to have non-Hypertext inside Hypertext, this is #2757.
I've managed to make it work since but I'd like to break this down into multiple PRs considering the amount of changes.

M2/Macaulay2/packages/Macaulay2Doc/overview4.m2
```diff
@@ -5,7 +5,8 @@
 undocumented (hypertext, Hypertext)
 
 document {
-     Key => {hypertext,(hypertext, List),(hypertext, Sequence)},
+--     Key => {hypertext,(hypertext, List),(hypertext, Sequence)},
```
mahrud, line 8: Instead of commenting this line, could you remove it and add `undocumented methods html` at the top? This prevents lots of undocumented method warnings.

M2/Macaulay2/packages/Macaulay2Doc/functions/installAssignmentMethod-doc.m2
mahrud: I don't understand; isn't `code` returning hypertext now?

M2/Macaulay2/packages/Macaulay2Doc/overview4.m2
pzinn: do you mean `undocumented methods hypertext`? or `html`, or both?

M2/Macaulay2/packages/Macaulay2Doc/functions/installAssignmentMethod-doc.m2
pzinn: precisely, it used to return `Net` which would just be copy/pasted into documentation. now it returns a `Hypertext` with potentially non-`Hypertext` in it, in the present case a `FilePosition`.

mahrud:
This seems to be a new error:
```m2
i1 : toExternalString code det
stdio:1:1:(3): error: can't convert local symbol or invisible global symbol 'f' to external string
Macaulay2/m2/option.m2                                                                      : here is the first use of 'f'
```

M2/Macaulay2/m2/document.m2
```diff
@@ -550,7 +550,7 @@ getSourceCode :=  val         -> DIV {"class" => "waystouse",
     fixup DIV {SUBSECTION "Code", PRE M2CODE demark_newline unstack stack apply(enlist val, m -> (
 		f := lookup m; if f === null then error("SourceCode: ", toString m, ": not a method");
 		c := code f;   if c === null then error("SourceCode: ", toString m, ": code for method not found");
-		reproduciblePaths toString c))}}
+		reproduciblePaths toString net c))}}
```
mahrud, line 553: Might it be easier to change `toString Hypertext` to do `toString net x`, and use `toExternalString Hypertext` to get what `toString Hypertext` currently does?

pzinn:
> This seems to be a new error:
> 
> ```m2
> i1 : toExternalString code det
> stdio:1:1:(3): error: can't convert local symbol or invisible global symbol 'f' to external string
> Macaulay2/m2/option.m2                                                                      : here is the first use of 'f'
> ```

yes, this is also mentioned in #2757: the problem is that the current output of `code det` (which, according to discussion of #2758, is not quite right anyway -- but I digress) contains a local variable "f", which can't be `toExternalString`ed. So there is still some need to process non-Hypertext inside Hypertext before say putting it in documentation; I think I've got it, but it's not part of this PR.

M2/Macaulay2/packages/VectorGraphics.m2
```diff
@@ -719,7 +719,8 @@ new SVG from GraphicsObject := (S,g) -> (
     ss
     )
 
-html GraphicsObject := g -> html SVG g;
+hypertext GraphicsObject := g -> SVG g
```
mahrud, line 722: Which objects should have a `hypertext` method now?

M2/Macaulay2/m2/document.m2
pzinn: I'm not sure that's a good idea. right now `toString` and `net` have very different outputs. here I do mean, take the usual `net` output, but turn it into a string (among other things, to be fed into `reproduciblePaths`, which itself is a bit of a hack and I will rewrite in a future PR).

M2/Macaulay2/packages/VectorGraphics.m2
pzinn: this is completely optional -- basically, for types whose `html` output is going through some Hypertext stuff, instead of writing
```
html MyType := x -> html ( DIV x)
```
or whatever, you can write
```
hypertext MyType := x -> DIV x
html MyType := html @@ hypertext
``
cleaner and potentially useful long term.

File: M2/Macaulay2/m2/methods.m2 [hunk 5]
```diff
@@ -700,15 +700,24 @@ Function Thing = (f,x,e) -> (
...
--- defined in d/actors4.d
-locate' = locate -- TODO: why does (net, FunctionBody) in nets.m2 need locate'?
-locate = method(Dispatch => Thing, TypicalValue => Sequence)
-locate Nothing    := Sequence => x -> locate' x
-locate Function   := Sequence => x -> locate' x
-locate Pseudocode := Sequence => x -> locate' x
-locate Sequence   := Sequence => x -> locate' x
-locate Symbol     := Sequence => x -> locate' x
-locate List       := List     => x -> apply(x, locate)
+FilePosition = new Type of BasicList
```
mahrud, line 703: Would it make sense to move both `FilePosition` and `locate` to `debugging.m2`?  They're not used until there, and also various other locating functions are defined there.

M2/Macaulay2/packages/VectorGraphics.m2
mahrud: Not sure if I follow. Where would `hypertext MyType` be called from?

M2/Macaulay2/m2/document.m2
mahrud: `toString` and `net` are far closer than `toExternalString`. For instance, what's the point of `toString PRE "hi"` returning `PRE{hi}`? That output can't be used anywhere else, can it?

M2/Macaulay2/packages/Macaulay2Doc/overview4.m2
mahrud: Sorry, I meant `hypertext`.

M2/Macaulay2/m2/hypertext.m2
```diff
@@ -363,6 +364,14 @@ style Hypertext := true >> o -> x -> (
     append(x,"style"=>str)
     )
 
+-- hijacked "hypertext"
```
mahrud, line 367: If you're going to repurpose hypertext for this, I don't think there's a need for comments like this. A mysteriously commented `fixup` won't help anyone. Also, is the typical value `Hypertext`?

pzinn: the typical value was always "Hypertext". I'll remove the offending comment...

M2/Macaulay2/packages/VectorGraphics.m2
pzinn: at this stage, only by `html`.

M2/Macaulay2/m2/document.m2
pzinn: I don't disagree with you, but maybe this is outside the scope of this PR. The standard behavior of `toString` for pretty much any type of `BasicList` is to write the name of the type followed by the contents. I'm not ready to change this at this stage, because it may have unintended consequences.

M2/Macaulay2/m2/methods.m2
pzinn: I can try, but I'm a little confused with all the complicated dependencies between files.

M2/Macaulay2/m2/debugging.m2
```diff
@@ -139,7 +134,7 @@ show1(Sequence,Function) := show1(List,Function) := (types,pfun) -> (
 	       else install pfun v);
 --	  if hasAttribute(v,PrintNet) then v = getAttribute(v,PrintNet) else
 --	  if hasAttribute(v,PrintNames) then v = getAttribute(v,PrintNames) else
-	  if hasAttribute(v,ReverseDictionary) then v = getAttribute(v,ReverseDictionary);
+--	  if hasAttribute(v,ReverseDictionary) then v = getAttribute(v,ReverseDictionary);
```
mahrud, line 137: @DanGrayson What was the purpose of these three lines? Seems kind of silly now that all three are commented.

pzinn: yeah, none of them made sense...

M2/Macaulay2/m2/format.m2
```diff
@@ -154,6 +156,9 @@ scan({net, info},
 		BK -> ());
 	    -- Stack the pieces vertically
 	    stack x);
+	parser INDENT := x -> ( -- INDENT is like DIV but with extra |s on the left. sadly, can't be absorbed into DIV because of non-recursivity of this parser
```
mahrud, line 159: Can we use nested `DIV`s for this purpose?

pzinn: I don't think so. the way `parser` is pretty sneaky (`parser` is only called by the outermost `DIV`, after that it's all `parser'`). anyway I already spent way too much time rewriting this `INDENT` thing, this is the best I could come up with.

M2/Macaulay2/m2/html.m2
```diff
@@ -162,40 +164,59 @@ html TO2  := x -> (
...
+hypertext FilePosition :=
+hypertext Dictionary := TTc "class-name"
+--hypertext VerticalList         := x -> UL apply(x, y -> new LI from hold y)
+--hypertext NumberedVerticalList := x -> OL apply(x, y -> new LI from hold y)
+scan(methods hypertext, (h,t) -> html t := html @@ hypertext);
 -- except not these descendants
 html Monoid :=
 html RingFamily :=
-html Ring := lookup(html,Thing)
+html Ring := texHtml
```
mahrud, line 217: Maybe it's just me, but `lookup(html, Thing)` is more informative to me than `texHtml`.

Alternatively, why not change it into this?
```m2
html Monoid :=
html RingFamily :=
html Ring :=
html Thing := x -> if lookup(hypertext, class x) then html hypertext x else "$" | htmlLiteral texMath x | "$"
```

M2/Macaulay2/m2/html.m2
```diff
@@ -162,40 +164,59 @@ html TO2  := x -> (
...
+	  , BR{})))
+hypertext Time := x -> DIV { x#1, DIV ("-- ", toString x#0, " seconds", "class" => "token comment") }
+TTc = c -> x -> TT {toString x,"class"=>"token "|c}
+hypertext Pseudocode :=
+hypertext CompiledFunctionBody := TTc "function"
+hypertext Command :=
+hypertext FunctionBody :=
+hypertext Function := f -> TT deepSplice {
+    if hasAttribute(f,ReverseDictionary) then toString getAttribute(f,ReverseDictionary) else (
+	t := locate if instance(f,Command) then f#0 else f;
```
mahrud, line 195: See https://github.com/Macaulay2/M2/issues/1531.

M2/Macaulay2/m2/hypertext.m2
mahrud: Is there a reason why you can't bring all the `hypertext` methods defined in `html.m2` here? Would be nice to have a section in `hypertext.m2` for the ways internal M2 types are converted to hypertext objects defined above.

M2/Macaulay2/m2/code.m2
```diff
@@ -33,33 +34,34 @@ getSourceLines Sequence := x -> (
...
-	       if #syms > 0 then indent listSymbols syms,
-	       if codeHelper#?(functionBody f) 
-	       then toSequence apply(
-		    codeHelper#(functionBody f) f, 
-		    (comment,val) -> indent stack (
-			      comment, 
-			      if instance(val, Function) then codeFunction(val,depth+1) else net val
-			      )))))
+	 l := locate f;
+	  if l === null then DIV{"function ", f, ": source code not available"}
```
mahrud, line 48: Is the indentation off here? Might as well also change the `if depth ...` line and fix indentation on this whole block.

M2/Macaulay2/m2/html.m2
pzinn: sure

M2/Macaulay2/m2/code.m2
```diff
@@ -100,24 +102,26 @@ editMethod String := filename -> (
...
 editMethod Function := args -> EDIT locate args
 editMethod Sequence := args -> (
-     editor := getViewer("EDITOR", "emacs");
-     if args === () 
-     then chkrun concatenate(
-	  if getenv "DISPLAY" != "" and editor != "emacs" then "xterm -e ",
-	  editor)
-     else EDIT locate args
+    if args === () then (
+	editor := getViewer("EDITOR", "emacs");
```
mahrud, line 119: You could save one layer of indentation with something like this:
```m2
editMethod Sequence := args -> if args =!= () then EDIT locate args else (
    editor := ...
```

M2/Macaulay2/m2/code.m2
pzinn: I wish we used a tool to automatically fix indentation, what I do for my other projects. will try to fix manually.

mahrud: Emacs can auto-fix indentation for M2, but I also wish there was a linter for M2.

M2/Macaulay2/m2/html.m2
mahrud: (This would eliminate the need for the `scan` a few lines higher also.)

M2/Macaulay2/m2/methods.m2
mahrud: I checked before commenting. Here's a little hack I use for searching the first usage of a symbol roughly by order of files being loaded:
```bash
reset ; for src in startup.m2 Core.m2 `cat loadsequence`; do grep -iHn "\bFilePosition\b" $src; done
```

M2/Macaulay2/packages/VectorGraphics.m2
mahrud: Still confused, but with the fix I suggested below (`html Thing := x -> if lookup(hypertext, class x) =!= null then html hypertext x ...`) would eliminate the need for all `html Blah := html @@ hypertext`.

M2/Macaulay2/m2/document.m2
mahrud: Is there a descendant of `BasicList` where `net` and `toString` print different things? I think in all the types you're referring to `net` also writes the name of the type followed by the contents, so if `net Hypertext` knows how to properly format the object, I definitely think `toString Hypertext` should also do that.

M2/Macaulay2/m2/code.m2
pzinn: this code needs rewriting anyway so I'm not gonna bother at this point.

mahrud:
For what it's worth, I reviewed all of this PR and I think it's great. The suggestions above are mostly the things I would do if I was writing this or the things I would like to change in the future.

M2/Macaulay2/m2/html.m2
pzinn: this is perverted:
```
help = new Command from help
-- This ensures that "methods help" and "?help" work as expected
setAttribute(help#0, ReverseDictionary, symbol help)
```
I don't think I'm gonna touch for now.

mahrud: Definitely, lol! I didn't mean you should fix this now, just pointing out that issue.

M2/Macaulay2/m2/document.m2
pzinn: again, not disagreeing. I'll make a note of it and try at some point.

M2/Macaulay2/m2/html.m2
pzinn: wait no that doesn't work.  for say a `Ring` the lookup will be nonnull and yet we do want to `texMath` it, not go through `hypertext`.

mahrud: What does `hypertext Ring` do? Maybe that should also behave differently then?

pzinn: well, it's inherited from `hypertext Type`...

M2/Macaulay2/m2/hypertext.m2
```diff
@@ -364,13 +368,43 @@ style Hypertext := true >> o -> x -> (
...
+hypertext IndeterminateNumber :=
+hypertext Manipulator :=
+hypertext Boolean := TTc "constant"
+hypertext Type :=
+hypertext FilePosition :=
+hypertext Dictionary := TTc "class-name"
+--hypertext VerticalList         := x -> UL apply(x, y -> new LI from hold y)
+--hypertext NumberedVerticalList := x -> OL apply(x, y -> new LI from hold y)
+
+scan(methods hypertext, (h,t) -> html t := html @@ hypertext);
```
mahrud, line 406: This line still bothers me. I can't think of a solution right now, but I'll think about it.

M2/Macaulay2/m2/hypertext.m2
```diff
@@ -4,6 +4,8 @@
 needs "regex.m2" -- for toLower
 needs "lists.m2" -- for all
 needs "methods.m2"
+needs "packages.m2" -- for Package
+needs "monoids.m2" -- for Monoid etc
```
mahrud, line 8: AHHH no this is very bad! `hypertext.m2` is near the top of `loadsequence`, `packages.m2` and `monoids.m2` are far, far below. The few hypertext definitions should move there instead.

mahrud: This was fine in `html.m2` because it's near the bottom, so admittedly this is a consequence of my own suggestion, but fortunately this files doesn't seem to use Monoid at all and `hypertext Package` can safely move to `packages.m2`.

pzinn: hehe I told you I had a bad feeling about moving things around...

mahrud:
Ahh, one more necessary change: in order to have the emacs M2-mode understand how to jump to the source code, for instance this one:
```m2
i1 : locate(symbol^, Matrix, List)

o1 = Macaulay2/m2/matrix.m2:399:33-399:61

o1 : FilePosition
```
the output needs to have one of [several formats](https://github.com/Macaulay2/M2-emacs/blob/86150290208bd9e468511b4b69a187ce870f303e/M2.el#L346-L409):
```m2
-*FunctionBody[../../m2/startup.m2.in:123:19-123:21]*-
-*Function[../../m2/res.m2:191:40-202:36]*-
     ../../m2/res.m2:210:45-214:6: --source code:
o1 = ../../m2/res.m2:210:45-214:6:
   | ../../m2/res.m2:210:45-214:6:
     ../../m2/res.m2:210:45-214:6: --source code:
     ../../m2/res.m2:210:45: --source code:
     ./packages/Posets.m2:1329:1:(3):[7]: error: type mismatch: ...
     ./packages/Posets.m2:1358: warning: documentation already provided for 'Posets :: moebiusFunction'
```
I can think of three ways to resolve this:
1. simply add a colon at the end of the address:
```m2
o1 = Macaulay2/m2/matrix.m2:399:33-399:61:
```
2. adjust the emacs mode to not require the colon
3. wrap it and adjust the emacs mode to understand this
```m2
o1 = -*FilePosition[Macaulay2/m2/matrix.m2:399:33-399:61]*-
```
cc: @d-torrance

mahrud:
Also, I thought this would have worked:
```m2
i17 : locate(symbol^, Matrix, List)

o17 = Macaulay2/m2/matrix.m2:399:33-399:61

o17 : FilePosition

i18 : code oo
stdio:23:1:(3): error: no method found for applying code to:
     argument   :  Macaulay2/m2/matrix.m2:399:33-399:61 (of class FilePosition)
```

pzinn:
> Ahh, one more necessary change: in order to have the emacs M2-mode understand how to jump to the source code, for instance this one:
> 
> ```m2
> i1 : locate(symbol^, Matrix, List)
> 
> o1 = Macaulay2/m2/matrix.m2:399:33-399:61
> 
> o1 : FilePosition
> ```
> 
> the output needs to have one of [several formats](https://github.com/Macaulay2/M2-emacs/blob/86150290208bd9e468511b4b69a187ce870f303e/M2.el#L346-L409):
> 
> ```m2
> -*FunctionBody[../../m2/startup.m2.in:123:19-123:21]*-
> -*Function[../../m2/res.m2:191:40-202:36]*-
>      ../../m2/res.m2:210:45-214:6: --source code:
> o1 = ../../m2/res.m2:210:45-214:6:
>    | ../../m2/res.m2:210:45-214:6:
>      ../../m2/res.m2:210:45-214:6: --source code:
>      ../../m2/res.m2:210:45: --source code:
>      ./packages/Posets.m2:1329:1:(3):[7]: error: type mismatch: ...
>      ./packages/Posets.m2:1358: warning: documentation already provided for 'Posets :: moebiusFunction'
> ```
> 
> I can think of three ways to resolve this:
> 
> 1. simply add a colon at the end of the address: (this is the least work)
> 
> ```m2
> o1 = Macaulay2/m2/matrix.m2:399:33-399:61:
> ```
> 
> 2. adjust the emacs mode to not require the colon
> 3. wrap it and adjust the emacs mode to understand this
> 
> ```m2
> i1 : locate(symbol^, Matrix, List)
> 
> o1 = -*FilePosition[Macaulay2/m2/matrix.m2:399:33-399:61]*-
> 
> o1 : FilePosition
> ```
> 
> This last way is more work, but it matches the output of `lookup`:
> 
> ```m2
> i25 : lookup(symbol^, Matrix, List)
> 
> o25 = -*Function[Macaulay2/m2/matrix.m2:399:33-399:61]*-
> 
> o25 : FunctionClosure
> ```
> 
> cc: @d-torrance

ah yes, I had a related conversation with @d-torrance -- my preference would be 2 because I think this colon is useless (historically, AFAICT, it came from the error messages which have one more `:(n)`); but I'm not sure how to implement 2. Alternatively, 3 would work too.

pzinn:
not directly related to this PR: @mahrud I'm struggling to fix a test, but I'm running blind because locally I can't get the ComputationsBook tests to run; when I do `ctest`, I get error messages about the `.trim` files not existing. I vaguely remember having had this problem, but don't remember how to fix it.

mahrud:
> not directly related to this PR: @mahrud I'm struggling to fix a test, but I'm running blind because locally I can't get the ComputationsBook tests to run; when I do `ctest`, I get error messages about the `.trim` files not existing. I vaguely remember having had this problem, but don't remember how to fix it.

You have to run `ninja M2-tests` or something like that first. I'm not proud of it but couldn't figure out how to make it work with ctest alone ...

d-torrance:
Currently, M2-mode doesn't recognize a file location when there's anything in front of it except whitespace.  For example, trying to call `M2-send-to-program-or-jump-to-source-code` with the point in the `o1` line below results in the following:

```m2
i1 : "Macaulay2/m2/matrix.m2:399:33-399:61"

o1 = Macaulay2/m2/matrix.m2:399:33-399:61

i2 : Macaulay2/m2/matrix.m2:399:33-399:61
stdio:2:10:(3): error: no method for binary operator / applied to objects:
--            Macaulay2 (of class Symbol)
--      /     m2 (of class Symbol)
```
This is why I added an extra newline for `(net, TestInput)`, e.g.,:

```m2
i3 : (tests "JSON")#0

o3 = 
     /usr/share/Macaulay2/JSON/test-parse.m2:0:1-295:1:

o3 : TestInput
```
I think I like option 3 (wrapping it in `-*FilePosition[...]*-` the most.

pzinn:
The problem with solution 3 is that *everywhere* there is a file location, the extra `-*FilePosition*-` will appear. That includes the output of `code`, for example. So I don't think that's a good idea, actually.

d-torrance:
Option 2 is really easy:

```diff
diff --git a/M2.el b/M2.el
index 2837cd0..7065816 100644
--- a/M2.el
+++ b/M2.el
@@ -369,7 +369,7 @@ can be executed with \\[M2-send-to-program]."
           ;; * a pipe, for paths enclosed in a box, e.g.,:
           ;;   | ../../m2/res.m2:210:45-214:6:
           ;;                (1                     1)   (2       2)   (3      3)   (4      4)   (5      5)   (6      6))
-          (looking-at "^ *\\(o+[1-9][0-9]* = \\|| \\)?\\([^:\n]+\\):\\([0-9]+\\):\\([0-9]+\\)-\\([0-9]+\\):\\([0-9]+\\):"))
+          (looking-at "^ *\\(o+[1-9][0-9]* = \\|| \\)?\\([^:\n]+\\):\\([0-9]+\\):\\([0-9]+\\)-\\([0-9]+\\):\\([0-9]+\\):?"))
         (let ((filename (buffer-substring (match-beginning 2) (match-end 2)))
               (linenum (string-to-number (buffer-substring (match-beginning 3) (match-end 3))))
               (colnum (if (match-beginning 4) (string-to-number (buffer-substring (match-beginning 4) (match-end 4))) 1))
```

pzinn:
> Option 2 is really easy:
> 
> ```diff
> diff --git a/M2.el b/M2.el
> index 2837cd0..7065816 100644
> --- a/M2.el
> +++ b/M2.el
> @@ -369,7 +369,7 @@ can be executed with \\[M2-send-to-program]."
>            ;; * a pipe, for paths enclosed in a box, e.g.,:
>            ;;   | ../../m2/res.m2:210:45-214:6:
>            ;;                (1                     1)   (2       2)   (3      3)   (4      4)   (5      5)   (6      6))
> -          (looking-at "^ *\\(o+[1-9][0-9]* = \\|| \\)?\\([^:\n]+\\):\\([0-9]+\\):\\([0-9]+\\)-\\([0-9]+\\):\\([0-9]+\\):"))
> +          (looking-at "^ *\\(o+[1-9][0-9]* = \\|| \\)?\\([^:\n]+\\):\\([0-9]+\\):\\([0-9]+\\)-\\([0-9]+\\):\\([0-9]+\\):?"))
>          (let ((filename (buffer-substring (match-beginning 2) (match-end 2)))
>                (linenum (string-to-number (buffer-substring (match-beginning 3) (match-end 3))))
>                (colnum (if (match-beginning 4) (string-to-number (buffer-substring (match-beginning 4) (match-end 4))) 1))
> ```

that's d be great. can you submit a PR to my `locate` branch?
also, can we now remove the `:` in your `TestInput` output? :-P

mahrud:
> also, can we now remove the `:` in your `TestInput` output? :-P

And the preceding new line!

d-torrance:
> > also, can we now remove the `:` in your `TestInput` output? :-P
> 
> And the preceding new line!

That's still necessary for Emacs support until some more substantial changes are made to `M2-send-to-program-or-jump-to-source-code`.  If we remove it:

```m2
i1 : tests "JSON"

o1 = HashTable{0 => src/macaulay2/M2/M2/Macaulay2/packages/JSON/test-parse.m2:0:1-295:1}
               1 => src/macaulay2/M2/M2/Macaulay2/packages/JSON/test-encode.m2:0:1-62:1

o1 : HashTable
```
With the point on either of these two lines, `M2-send-to-program-or-jump-to-source-code` gives us the following:

```
M2-jump-to-source-code: File not found: HashTable{0 => src/macaulay2/M2/M2/Macaulay2/packages/JSON/test-parse.m2
M2-jump-to-source-code: File not found: 1 => src/macaulay2/M2/M2/Macaulay2/packages/JSON/test-encode.m2
Mark set 
```

mahrud:
I'd still prefer to remove it now and fix emacs later. You could go the `-*TestInput[...]*-` route, for instance.

mahrud:
> The problem with solution 3 is that _everywhere_ there is a file location, the extra `-*FilePosition*-` will appear. That includes the output of `code`, for example. So I don't think that's a good idea, actually.

There is a balanced solution: have the default appearance of FilePosition look like `-*FilePosition[...]*-` but inside output of `code` or other common places drop the `-*FilePosition*-` (e.g. by having them print `fp#0` if `fp` is a FilePosition`).

d-torrance:
> I'd still prefer to remove it now and fix emacs later. You could go the `-*TestInput[...]*-` route, for instance.

Sounds good to me!  See https://github.com/pzinn/M2/pull/4 and https://github.com/Macaulay2/M2-emacs/pull/36.

Now it looks like this, and `M2-send-to-program-or-jump-to-source-code` works as expected:

```m2
i1 : tests "JSON"

o1 = HashTable{0 => -*TestInput[src/macaulay2/M2/M2/Macaulay2/packages/JSON/test-parse.m2:0:1-295:1]*-}
               1 => -*TestInput[src/macaulay2/M2/M2/Macaulay2/packages/JSON/test-encode.m2:0:1-62:1]*-

o1 : HashTable
```

pzinn:
@d-torrance I think your PR broke a test, any chance you could fix it?

d-torrance:
> @d-torrance I think your PR broke a test, any chance you could fix it?

Oops!  https://github.com/pzinn/M2/pull/5 should fix it.

pzinn:
thanks everyone, I think this PR is ready to ship!

pzinn:
argh! as I was fixing #2777 I looked up the list of void hypertext elements and decided to implement `INPUT`, which was missing. Sadly it conflicts with a symbol in `NumericalAlgebraicGeometry` @antonleykin. I still don't understand how these conflicts arise and why there isn't a clean way to resolve them.

pzinn:
a simple fix seems to be to change this line:
```
--- a/M2/Macaulay2/packages/NumericalAlgebraicGeometry/SLP.m2
+++ b/M2/Macaulay2/packages/NumericalAlgebraicGeometry/SLP.m2
@@ -28,7 +28,7 @@ slpDET = 4; --"DET";
 
 
 CONST = symbol CONST;
-INPUT = symbol INPUT;
+INPUT := symbol INPUT;
 protect CONST
 protect INPUT
```
but it might have some consequences I'm not aware of...

antonleykin:
> argh! as I was fixing #2777 I looked up the list of void hypertext elements and decided to implement `INPUT`, which was missing. Sadly it conflicts with a symbol in `NumericalAlgebraicGeometry` @antonleykin. I still don't understand how these conflicts arise and why there isn't a clean way to resolve them.

What are the current implications of this conflict?

`INPUT` is SLP.m2 may be OK as a local (to this file) symbol, but... conceptually we may be in a situation when a global protected symbol has the name we want in, say, Core. I don't know a good way to take care of this either.

pzinn:
well, as it stands NAG will not run since it will cause an error about trying to alter a protected symbol.
The problem, as far as I understand, is not that you're trying to use `INPUT` in your file, but the fact that you're trying redefine it with the line `INPUT = symbol INPUT`.
I'll give my fix a try. The part I never understood is why "your" `INPUT` will then have package null and not package NAG as it used to. Maybe not so important but still confusing.

antonleykin:
> The problem, as far as I understand, is not that you're trying to use the symbol `INPUT` in your file, but the fact that you're trying to redefine it with the line `INPUT = symbol INPUT`.

In this case with what `INPUT` is for both `Core` and `NAG` there is no need to redefine.

Hypothetically, if my package is split up in many files and I (really) need an already protected symbol `INPUT` to be local (to the package) and usable anywhere in the package, what can I do?

pzinn:
> > The problem, as far as I understand, is not that you're trying to use the symbol `INPUT` in your file, but the fact that you're trying to redefine it with the line `INPUT = symbol INPUT`.
> 
> In this case with what `INPUT` is for both `Core` and `NAG` there is no need to redefine.
> 
> Hypothetically, if my package is split up in many files and I (really) need an already protected symbol `INPUT` to be local (to the package) and usable anywhere in the package, what can I do?

I understand. I dont' know of a good solution, maybe someone else does...
All I can think of is, you can write `I = symbol INPUT`, or you can write `INPUT := symbol INPUT` in every file.

M2/Macaulay2/m2/hypertext.m2
```diff
@@ -346,6 +355,14 @@ addAttribute(TD,     htmlAttr | {"colspan", "headers", "rowspan"})
 addAttribute(TH,     htmlAttr | {"colspan", "headers", "rowspan"})
 addAttribute(IMG,    htmlAttr | {"alt", "src", "srcset", "width", "height",
 	"sizes", "crossorigin", "longdesc", "referrerpolicy", "ismap", "usemap"})
+addAttribute(OL, htmlAttr | {"start"=>"0", "reversed", "type"})
```
d-torrance, line 358: @pzinn - What was the rationale behind giving `"start"` an initial value of 0?  This gives inconsistent results between `net` and `html`.  For example:

```m2
i1 : beginDocumentation()

i2 : OL{"foo", "bar", "baz"}

o2 =   1. foo
       2. bar
       3. baz

i3 : html oo

o3 = <ol start="0">
       <li>
     foo  </li>
       <li>
     bar  </li>
       <li>
     baz  </li>
     </ol>
```
And in my browser, this gives:

image

Also, `make validate-html` now fails on any doc files with an `<ol>` tag, e.g.,:

```
validating: Macaulay2Doc/html/_writing_spdocumentation.html
 *** invalid HTML: /<<PKGBUILDDIR>>/M2/usr-dist/common/share/doc/Macaulay2/Macaulay2Doc/html/_writing_spdocumentation.html
error: line 72: there is no attribute "start"
```

pzinn: The rationale is that all counters start at 0 in Macaulay2, including `NumberedVerticalList`.
the `Net` output is a bug which I will fix. 
I don't understand the validate issue -- `OL` does have the start attribute...

d-torrance: Ok, that makes sense.

`start` is an attribute for `<ol>` tags in HTML5, but not in XHTML 1.1, which is the spec that's used for validation.

pzinn: I don't understand why `validate` doesn't refer to the attributes that are defined in `hypertext.m2`...

d-torrance: The issue isn't with `validate` but with `make validate-html`, which is a different thing.  It runs `util/validator-runner.py`, which runs the check from https://validator.w3.org/ locally.

pzinn: got it. this is outside my area of expertise then :-P

d-torrance: No worries!  I'll see if I can figure something out.

PR #2764, opened 2023-02-04
Add "packages" button to top of html documentation pages

d-torrance:
This allows users to easily switch from viewing a documentation page in one package to viewing a documentation page for another package.

We replace some commented out code referring to a `directoryButton`, which may have been intended to point to the directory page that is installed in `applicationDirectory`.  However, the URL to this page would be impossible to know when the documentation is built for a distribution because it depends on the value of the user's `HOME` environment variable and the `applicationDirectorySuffix` variable.  The "packages provided with Macaulay2" page in Macaulay2Doc is the next best thing.

image

PR #2765, opened 2023-02-05
Lint workflow updates

d-torrance:
We now use the Ubuntu package for `codespell` rather than getting it from Conda, which fixes the errors seen in recent pull requests (#2760, #2763, and #2764).

A few other small changes:

* We now run on `ubuntu-latest` rather than `ubuntu-20.04`.  This should simplify maintenance, since presumably GitHub will get rid of its `ubuntu-20.04` runners in a few years.
* We no longer check out the submodules, since `codespell` is only run inside the `packages` directory.
* We add two new words ("ba" and "sur") to the ignored word list.  I'm assuming this was needed because of a difference in versions between the `codespell` package in Ubuntu v. Conda.

Here's an example of a successful run of the workflow with these updates: https://github.com/d-torrance/M2/actions/runs/4096896422

Cc: @fchapoton

### DIFF
--- a/.codespell_ignore
+++ b/.codespell_ignore
@@ -7,6 +7,7 @@ ans
 anumber
 archiv
 assasin
+ba
 bloks
 braket
 bu
@@ -62,6 +63,7 @@ singl
 skelton
 sring
 strat
+sur
 te
 thets
 toom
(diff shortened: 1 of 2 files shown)


fchapoton:
looks good to me

fchapoton (approved):

fchapoton:
I would suggest to merge that soon, in order to repair the CI for other pull requests.

PR #2768, opened 2023-02-07
ARing ElementType exception safety and refactor

jkyang92:
This is very much work in progress, and more a proposal. This is part of a bigger proposal I have, which I've written up here https://github.com/Macaulay2/M2/wiki/Ring-and-RingElement-Refactor-Proposal. But this branch currently only has the changes for exception safety with RT::ElementType

Most of the code that creates ElementType values directly is not exception safe, since there's no guarantee that the appropriate destructors will be called. as is, this pull request fixes this by creating for RT::ElementType a wrapper named RT::Element which has a destructor that cleans up the value, and thus is exception safe. 

For almost all ARings, Element is a zero cost wrapper around ElementType. There are three ARing types that cannot support this, all because they need additional data for their destructors, these types are

- `ARingTower`
- `ARingGFFlint`
- `ARingGFFlintBig`

For both `ARingGFFlint` and `ARingGFFlintBig`, if we circumvent the public flint api, we can implement them with zero overhead. `ARingTower` would need a redesign of the `ARingTower::ElementType` data structure to support this.

In theory, one might want to replace all uses of ElementType with Element, but there are a couple of things that prevent this:

- The types mentioned above that don't support a zero cost wrapper (this is important if there is performance sensitive code)
- arrays of `Element` are harder to construct (need placement new)
- Not every `Ring` is a `ConcreteRing` so not every ring has a corresponding `ARing`

At this point, on the exception safety front, a lot of the code that uses ElementType has not been changed to use Element, of which the majority is in the matrix code.


jkyang92:
@mahrud Not as planned. The main goal is to make the code more exception safe (fewer explicit init/clear pairs), easier to understand (hopefully), and remove the ring_elem union. Except needed to make the interface workable, it shouldn't involve any real changes to the implementation of the actual rings, just repackaging what exists into a more usable and consistent interface.

jkyang92:
I've marked this pull request as ready, There are a couple minor things that could be done, but it's not clear if they're useful.

1. Add copy/move constructors to Element, This is not too hard, it does require some care for certain types where it's not that easy to create a copy constructor. But right now, it's not needed, since Element currently is only ever used as a local variable.
2. Rewrite functions to work with Element directly instead of ElementType. While possible to do, there doesn't really seem to be a point. It would require changing most of the ARing code for very little gain.

jkyang92:
I had forgotten that flint does weird things with pointers, so the previous implementation for `from_ring_elem_const` for flint integers and rationals was wrong.

Issue #2762, opened 2023-02-04
runProgram/findProgram wishlist


mahrud:
This is somehow the first time I've ever had to deal with these two methods, so here are some thoughts about what I expected before reading the documentation:
1. `normaliz = findProgram "normaliz"` should work (e.g. if I'm just prototyping)
2. `runProgram("normaliz", "--version")` should also work (e.g. if I'm just lazy)
3. `status runProgram(normaliz, "--version")` should give the return code
4. `toString runProgram(normaliz, "--version")` should give the string output
5. `normaliz << input` should do the obvious thing, which is to run with that input
6. `MinimumVersion` would be useless if I didn't know bash-fu. Why not take the minimum version and then a regex rule for getting the version from the output:
```m2
normaliz = findProgram("normaliz", "--version", "2.8", "Normaliz ([0-9.]+)")
```
Otherwise it would be simpler to just use `get "!normaliz --version"` (which is _impossible_ to find in the documentation unless you already know about it!). Alternatively, if the version has to be extracted from a file, I would expect something like:
```m2
ver = select("NMZ_VERSION *([0-9.]+)", "$1", get "libnormaliz/version.h")
normaliz = findProgram("normaliz", "--version", ver)
```
That is, if there are only three arguments, the third argument is the stated version, but if there are four arguments, the 3rd one is the minimum version and the 4th is the regex for extracting it.

cc: @d-torrance

d-torrance:
Great suggestions!

d-torrance:
Just to clarify -- is this the behavior you were suggesting for `Program << Thing`?

```m2
i1 : M2 = findProgram "M2"

o1 = M2

o1 : Program

i2 : M2 << "2 + 2"

o2 = 0

o2 : ProgramRun

i3 : toString oo

o3 = 
     i1 : 2 + 2
          
     o1 = 4
```

mahrud:
Exactly! It may be too much syntactic sugar, but easy to remember syntax gets used more :) We can think about this longer also, because perhaps something like this would also be nice:
```m2
runProgram(M2, "-q") << "2 + 2"
```
but that might require non-superficial changes.

mahrud:
Updates:
1. `msolve = findProgram "msolve"` doesn't work because running `msolve` alone prints help and returns code 1, which is not an uncommon choice. As long as the program is found and it is executable, I think this should return a `Program` object (i.e. the user chose not to provide a test command, so no need to test!)
[...]

PR #2776, opened 2023-02-13
findProgram/runProgram updates

d-torrance:
This is a partial response to #2762 (wishlist items 1 - 5).   For example:

```m2
i1 : M2 = findProgram "M2"

o1 = M2

o1 : Program

i2 : M2 << "2 + 2"

o2 = 0

o2 : ProgramRun

i3 : status oo

o3 = 0

i4 : toString ooo

o4 = 
     i1 : 2 + 2
          
     o1 = 4

i5 : peek runProgram("M2", "--version")

o5 = ProgramRun{command => /usr/bin/M2 --version}
                error => 
                output => 1.21

                return value => 0
```

The related documentation has also been converted to `SimpleDoc` format.

I plan on working on the `MinimumVersion` suggestions (wishlist item 6) soon, but in another PR.


d-torrance:
All the builds were failing because I tried documenting `(toString, ProgramRun)`, but that conflicts with the following:

https://github.com/Macaulay2/M2/blob/048c777dc703365e44c3865b2c1faed698b82b3a/M2/Macaulay2/packages/Macaulay2Doc/functions/toString-doc.m2#L5

Just force-pushed a fix.

mahrud:
> All the builds were failing because I tried documenting `(toString, ProgramRun)`, but that conflicts with the following:
> 
> https://github.com/Macaulay2/M2/blob/048c777dc703365e44c3865b2c1faed698b82b3a/M2/Macaulay2/packages/Macaulay2Doc/functions/toString-doc.m2#L5
> 
> Just force-pushed a fix.

This shouldn't cause a conflict; a document tag that is marked as undocumented can be later documented without an error. This allows for marking every method undocumented by default and then documenting the exceptions.

mahrud:
Deleted the previous comment. I'm not sure what is going on, but the `undocumented methods toString` line should not have caused a conflict ...

d-torrance:
I'm not sure if this is what we want or not, but currently `undocumented methods toString` set up dummy documentation for each of the various `toString` method functions.

Here's a toy example:

```m2
i1 : newPackage("Foo", Headline => "foo");

i2 : foo = method();

i3 : foo ZZ := foo RR := x -> null;

i4 : beginDocumentation()

i5 : undocumented methods foo

i6 : peek Foo#"raw documentation"

o6 = MutableHashTable{foo(RR) => HashTable{DocumentTag => Foo :: foo(RR)}}
                                           filename => stdio
                                           linenum => 6
                                           undocumented => true
                      foo(ZZ) => HashTable{DocumentTag => Foo :: foo(ZZ)}
                                           filename => stdio
                                           linenum => 6
                                           undocumented => true
```

d-torrance:
The failing `cmake-macos-12-clang-15` build is puzzling.  It was from the "Example: Translation and rotation sub-actions of the adjoint action of SE(3)" example from the `SubalgebraBases` package.

From the artifacts for this build:
```m2
i5 : R = QQ[x_1..x_9, MonomialOrder => Lex];
stdio:5:5:(3): error: expected a list, hash table, database, or dictionary
```

PR #2788, opened 2023-03-03
Add texMath(ForeignObject)

d-torrance:
Improves display of foreign objects in web app:

image

### DIFF
--- a/M2/Macaulay2/packages/ForeignFunctions.m2
+++ b/M2/Macaulay2/packages/ForeignFunctions.m2
@@ -124,7 +124,7 @@ ForeignObject = new SelfInitializingType of HashTable
 ForeignObject.synonym = "foreign object"
 net ForeignObject := x -> net value x
 texMath ForeignObject := texMath @@ value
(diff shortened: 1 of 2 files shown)


d-torrance:
> Wouldn't that give an error if I was using a ForeignObject that doesn't have an equivalent value in M2?

Sure, but that shouldn't happen.  `value` is installed for all of the various `ForeignObject` types.  This is also how `(net, ForeignObject)` is implemented.

pzinn:
nice. I didn't realize ForeignFunctions worked on the cloud, thought there was some compilation problems.
Incidentally:
Screenshot from 2023-03-04 14-19-06

d-torrance:
> nice. I didn't realize ForeignFunctions worked on the cloud, thought there was some compilation problems. 

Yup!  I think this was fixed in #2636.

> Incidentally:

Already taken care of in #2731.  :)  The new `AfterPrint` interface is so much nicer!

d-torrance:
@pzinn - I just added a couple commits with the small `AfterPrint` changes you suggested.

PR #2790, opened 2023-03-09
Use prism.js for syntax highlighting in html docs

d-torrance:
After discussion with @pzinn, we decided that it would make the most sense if Macaulay2 and Macaulay2Web both used the same syntax highlighting library -- currently it's [highlight.js](https://highlightjs.org/) and [Prism](https://prismjs.com/), resp.

This PR adds support for using either library for syntax highlighting the html documentation, defaulting to Prism.  The Prism build is very similar to the highlight.js one -- the necessary libraries are downloaded during the build using [npm](https://www.npmjs.com/) and the JavaScript and CSS are all packed together using [webpack](https://webpack.js.org/).  The CSS file is just the one Paul put together for Macaulay2Web, a modification of the default Prism theme with a few color changes.

A new key-value pair has been added to the `Layout` hash tables -- `"syntax highlighting"`, which will either be `"no"`, `"prism"`, or `"highlightjs"`.  This is set during build and is used to determine which, if any, `<script>` tag to include in the html documentation files.  (Previously, we always had a highlight.js `<script>` tag, but it pointed to an empty file if syntax highlighting was disabled.)

This is a draft for now as I haven't updated the cmake build yet.  @mahrud -- my thought is to add an `M2-prism` target similar to the current `M2-highlightjs` one, and then some sort of configuration variable, like `cmake -DSYNTAX_HIGHLIGHTING=...`.  It seems like there should be two though -- a bool for yes/no, and a string for prism/highlightjs if the bool was yes.  Does that seem okay?


d-torrance:
Is this unnecessarily complicated? Would it make more sense to just drop highlightjs and have Prism be the only option?

d-torrance:
A couple related questions:

* Should enabling syntax highlighting be the default?  Previously, it wasn't in the autotools build but was in the cmake build (by virtue of there not be an option *not* to enable it).
* Should we remove all the other vendored copies of JavaScript libraries (KaTeX and all the ones used by the Visualize package) and use a similar npm/webpack build for those too?

mahrud:
I'm opposed to the change in the binary (via startup.m2) and a build option for something that should really be just an option to `installPackage` or something like that.

d-torrance:
> I'm opposed to the change in the binary (via startup.m2)

This is mostly an artifact of my first draft of this PR where there was an option to choose between Prism and highlight.js.  Now that's it's just Prism, it would be easy to go back to the current behavior of having the `<script>` tag regardless, but just having it point to an empty file if syntax highlighting is disabled.

> and a build option for something that should really be just an option to `installPackage` or something like that.

This is a build-time decision, though.  Should we generate `prism.js` and ship it with the `Style` package or not?

I suppose another option would be to bring the autotools build in line with the cmake build's current behavior.  If `npm` is present, then go ahead and enable syntax highlighting.  If not, give a warning and just use an empty file.

mahrud:
> This is a build-time decision, though. Should we generate `prism.js` and ship it with the `Style` package or not?

I don't see the point of this decision. First, you can generate prism.js with every release and ship it with M2, similar to emacs or vim syntax files, and maybe also a build target to force regenerating all highlighters. Second, who doesn't want syntax highlighting? There could be build options for countless things, but at some point we have to just ask people with very specific needs to edit the code or build files themselves.

d-torrance:
Good points!  Converting to a draft again and work on this some more later.

d-torrance:
I've simplified everything after @mahrud's suggestions.  In particular, `prism.js` is now committed to the git repository in the `Style` package, just like KaTeX, and so all html documentation will now be syntax highlighted, and we don't need to worry about re-building it during every build.

There are still build targets (`make -C Macaulay2/editors prism/prism.js` for autotools and `ninja M2-prism` for cmake) to regenerate the file when needed, and even one for the autotools build to copy it to the `Style` directory and commit it to git (`make -C Macaulay2/editors update-syntax-highlighting`) to make things nice and simple.

DanGrayson:
@mahrud - what do you think of it now?

mahrud (approved):

d-torrance:
I think I messed up something in the `editors` directory `Makefile` for in-source builds with this PR:

```
make[3]: Circular emacs/M2-init.el <- emacs/M2-init.el dependency dropped.
cp emacs emacs/M2-init.el
cp: -r not specified; omitting directory 'emacs'
make[3]: *** [Makefile:27: emacs/M2-init.el] Error 1
```

Issue #497, opened 2016-07-14
Visualize Documentation Links


bstone:
It seems that some links in the documentation are not working properly on the website. In particular here:

http://www.math.uiuc.edu/Macaulay2/doc/Macaulay2-1.9.1/share/Macaulay2/Visualize/graph-example.html

One thought I have is that I wrote this using html5. Does the server only know html4.1? If so, would it be easy to upgrade the server? Or do I need to revisit my code?

I will say the documentation seems to work well if you run installPackage and view it locally.

DanGrayson:
The best way to fix HTML validity problems is with the validator at https://validator.w3.org/.  Give it a try.

Servers don't care whether a web page is HTML 4 or HTML 5 -- each page says what it is and each browser adapts to that.

I don't know why browsers are more particular about errors in web pages from the internet than about errors in web pages from the local file system.  Perhaps it's a newly heightened security concern.  We did notice that, too, and as a result, all the web pages we generate that are deposited in `share/doc/Macaulay2` are validated by the same validator.

mahrud:
@bstone I'd be interested in getting this to work again, if you have time! What is the status now?

bstone:
I plan on working on this package during the workshop next week. I am not sure how to fix this, but if you have ideas and want to help that would be wonderful.

mahrud:
I just realized this works here:
http://www2.macaulay2.com/Macaulay2/doc/Macaulay2-1.16/share/Macaulay2/Visualize/graph-example.html
but not here:
https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.16/share/Macaulay2/Visualize/graph-example.html

DanGrayson:
Probably because the apache http server on www2 is ignoring the directive `AddType 'application/xhtml+xml' .html` in the file `doc/.htaccess`.  Maybe that subjects the file to more rigorous checking.

mahrud:
Then remove that directive to solve this issue! It's not an `xhtml` document. What's wrong with `html`? See that the www2 page only has a few minor validation issues: https://validator.w3.org/nu/?doc=http%3A%2F%2Fwww2.macaulay2.com%2FMacaulay2%2Fdoc%2FMacaulay2-1.16%2Fshare%2FMacaulay2%2FVisualize%2Fgraph-example.html

No reason to wholesale block the webpage from being usable.

[...]

PR #2795, opened 2023-03-12
Use CSS counters instead of OL "start" element for zero-based numbering

d-torrance:
A small change from #2760 adding `start=0` to each `<ol>` tag for zero-based numbering of ordered lists is causing `make validate-html` to fail due to 3 html documentation pages with `<ol>` tags. From a [recent PPA build](https://launchpadlibrarian.net/655577774/buildlog_ubuntu-jammy-amd64.macaulay2_1.20.0.1+git202303111459-0ppa202302030934~ubuntu22.04.1_BUILDING.txt.gz):

```
 *** invalid HTML: /<<PKGBUILDDIR>>/M2/usr-dist/common/share/doc/Macaulay2/Macaulay2Doc/html/_writing_spdocumentation.html
error: line 72: there is no attribute "start"
...
*** invalid HTML: /<<PKGBUILDDIR>>/M2/usr-dist/common/share/doc/Macaulay2/Macaulay2Doc/html/_document.html
error: line 109: there is no attribute "start"
...
*** invalid HTML: /<<PKGBUILDDIR>>/M2/usr-dist/common/share/doc/Macaulay2/Text/html/___O__L.html
error: line 74: there is no attribute "start"
```
An alternate solution is to use CSS counters to achieve the same result but still have valid XHTML code.

Cc: @pzinn

### DIFF
--- a/M2/Macaulay2/m2/html.m2
+++ b/M2/Macaulay2/m2/html.m2
@@ -96,7 +96,15 @@ html Hypertext := x -> (
     attr := "";
     cont := if T.?Options then (
 	(op, ct) := override(options T, toSequence x);
-	scanPairs(op, (key, val) -> if val =!= null then attr = " " | key | "=" | format val | attr);
+	scanPairs(op, (key, val) -> (
+		-- workaround for "start" attribute not being valid XHTML 1.1
+		if T === OL and key == "start"
+		then (
+		    element := toString(changeBase(val, 10) - 1);
+		    attr = concatenate(" style=\"counter-reset: element ",
+			element, ";\"", attr))
+		else if val =!= null
+		then attr = " " | key | "=" | format val | attr));
 	sequence ct) else x;
     pushIndentLevel 1;
     (head, prefix, suffix, tail) := (
--- a/M2/Macaulay2/packages/Style/doc.css
+++ b/M2/Macaulay2/packages/Style/doc.css
@@ -78,6 +78,14 @@ dl.element {
   margin: 0px;
 }
 
+ol li::marker {
+  content: counter(element) ". ";
+}
+
+ol li {
+  counter-increment: element;
+}
+
 /*
 
  Local Variables:


pzinn:
I'm not sure about this. You're replacing a simple, built-in feature, with a complicated, custom one.
also, I think this will supersede the `start` option which means
```
OL {a,b,c,"start"=>"1"}
```
(which one could imagine wanting to do occasionally) will no longer work...

d-torrance:
I totally agree -- "start" is much nicer. The better long term solution would be to switch specs from xhtml to html5.

mahrud:
> I totally agree -- "start" is much nicer. The better long term solution would be to switch specs from xhtml to html5.

c.f. https://github.com/Macaulay2/M2/issues/497

d-torrance:
A big hurdle for moving to HTML5 is that `<tt>` has been removed.  And we use that a *lot*!

DanGrayson:
Is this ready to be merged, or is it still under discussion?

d-torrance:
I think it's still under discussion.  I guess there are several options:

1. Merge this, have zero-based ordered lists, and valid XHTML.
2. Don't merge this, go back to having one-based ordered lists (so remove the `start` attribute), and have valid XHTML.
3. Don't merge this, keep the current behavior, but have invalid XHTML.  Most browsers should still render the docs just fine, though.
4. Don't merge this and start porting the documentation to HTML5.

My preference would be for option 4, but it's a nontrivial task due to the prevalence of `<tt>` tags in the docs.  (Most of these should probably end up being changed to `<var>` tags.)

DanGrayson:
Well, then, what should we do, Doug?

d-torrance:
How important is it to you that `make validate-html` pass, Dan?  If it doesn't matter, then I think we can close this, and possibly work towards transitioning the docs to HTML5 for a future release.

DanGrayson:
Is there a good reason to transition to html 5?

d-torrance:
> Is there a good reason to transition to html 5?

Several, I think:

* Specific to this PR, `<ol>` doesn't have a `start` attribute in XHTML 1.1 but it does in HTML 5.  So our current code isn't XHTML 1.1-compliant and this is why `make validate-html` fails.
* As Mahrud pointed out, this likely would fix #497.
* See also this [StackOverflow question](https://stackoverflow.com/questions/66898396/xhtml-whats-the-advantage-of-using-xhtml-in-modern-application-development).  A couple key quotes:  *"True XHTML ... is not compatible with many frameworks because they use non-standard elements and attributes that don't validate as XHTML"* and *"Most developers consider the original XHTML 1 old tech these days"*.

DanGrayson:
I think the validator at https://validator.w3.org/ doesn't handle html5.  Validation of our html is really important, for then it will work with all browsers.

d-torrance:
Yes, but https://validator.w3.org/nu/ handles html5.

d-torrance:
We could also set up a GitHub action to validate the documentation regularly using https://github.com/anishathalye/proof-html.

DanGrayson:
It's not as good at checking.  The documentation contains this warning: "The Nu Html Checker should not be used as a means to attempt to unilaterally enforce pass/fail conformance of documents to any particular specifications; it is intended solely as a checker, not as a pass/fail certification mechanism."

d-torrance:
@DanGrayson -- It seems like you're firmly in the camp of:

* sticking with XHTML 1.1
* making sure that our html documentation validates

Is that correct?

Then this means we have to do something, as our documentation doesn't validate right now because of `<ol>` tags with `start` attributes.

I've updated this PR so that it more accurately matches the desired behavior and doesn't monkey around with the underlying `OL` object -- it just hacks `html(Hypertext)` to switch `start` tags to `style` tags that accomplish the same thing using CSS.

I think this is pretty much the only way we can support the `start` attribute for `OL` objects while maintaining valid XHTML 1.1.

Consider for example:

```m2
document {
    Key => foo,
    OL {"foo", "bar"},
    OL {"baz", "qux", "start" => "5"}}
```

This produces the following html:

```m2
i3 : html help foo

o3 = <div>
       <h1>foo</h1>
       <div>
         <h2>Description</h2>
         <ol style="counter-reset: element -1;">
           <li>
     foo      </li>
           <li>
     bar      </li>
         </ol>
         <ol style="counter-reset: element 4;">
           <li>
     baz      </li>
           <li>
     qux      </li>
         </ol>
       </div>
       <div class="waystouse">
         <h2>For the programmer</h2>
         <p>The object <a href="/home/profzoom/.Macaulay2/local/share/doc/Macaulay2/User/html/_foo.html">foo</a> is <span>a <a title="the class of all symbols" href="/usr/share/doc/Macaulay2/Macaulay2Doc/html/___Symbol.html">symbol</a></span>.</p>
       </div>
     </div>
```

In the browser, we get:

image

pzinn:
I still disagree with this for the reasons mentioned above.

d-torrance:
> I still disagree with this for the reasons mentioned above.

I believe I've fixed one of your original reasons at least -- the `start` value is used even when it's not 0.  It's still an ugly hack, though.  :)

Bascially, we can't have all three of the following things:

1.  `<ol>` tags with `start` attributes
2. valid (X)HTML
3.  XHTML 1.1

Currently, we have 1 & 3.  This PR is proposing we switch to 2 & 3 while mimicking the behavior of 1.  (I would prefer 1 & 2, but that would take a lot of work and Dan seems intent on sticking with 3.)

pzinn:
XHTML was abandoned a long time ago, and XHTML1.1 has been deprecated for at least a decade.
It makes no sense to stick to it in the age of HTML5.

mikestillman:
I agree with @pzinn

DanGrayson:
Okay, I concede.

d-torrance:
If you concede, the this should have been closed instead of merged.

d-torrance:
This should definitely be reverted at some point.  The question is when:

* Now, shipping 1.22 with exactly 3 pages of docs that don't use valid XHTML 1.1?
* After 1.22 is released and as we start transitioning the docs to HTML 5?

I guess I don't really have a strong opinion either way.

DanGrayson:
> If you concede, the this should have been closed instead of merged.

Oops, I'll revert it.

DanGrayson:
Reverted in commit fe6dafec120baf5b361aef8990d667d531c326b9

PR #2799, opened 2023-03-18
Add support for 0th powers of ring maps

### LINKED ISSUE #2792: Raising a ring self-map to the 0 power should be legal
```
R = ZZ[x, y, z];
f = map(R, R, {y, x*z, x^2 + 1});

f^1
f^5
f^0
```

The commands `f^1` and `f^5` give the expected result, based on iterating `f`, but `f^0` gives an error "missing unit element," although the operation has a well-defined answer, namely the identity map. Hoping this isn't too hard to fix. I am running Macaulay2 version 1.21.

d-torrance:
This fixes #2792:

```m2
i1 : R = ZZ[x, y, z];

i2 : f = map(R, R, {y, x*z, x^2 + 1});

o2 : RingMap R <--- R

i3 : f^0

o3 = map (R, R, {x, y, z})

o3 : RingMap R <--- R
```

### DIFF
--- a/M2/Macaulay2/tests/normal/hashtables.m2
+++ b/M2/Macaulay2/tests/normal/hashtables.m2
@@ -1,13 +1,15 @@
+importFrom_Core {"BinaryPowerMethod"}
 plus' = (x,y) -> ( z := x+y ; if z == 0 then continue ; z )
 Poly = new Type of HashTable
 assert( (one = new Poly from {1 => 1}) === new Poly from {1 => 1} )
+Poly#1 = x -> one
 assert( (T = new Poly from {2 => 1}) === new Poly from {2 => 1} )
 - Poly := (f) -> applyValues(f,minus);
 Poly + Poly := (f,g) -> merge(f,g,plus');
 Poly - Poly := (f,g) -> f + -g;
 Poly * Poly := (f,g) -> combine(f,g,times,times,plus');
-Poly ^ ZZ := lookup(symbol ^,CC,ZZ)
-assert( (f = one+T+T^2+T^3+T^4+T^5) === new Poly from {32 => 1, 16 => 1, 8 => 1, 1 => 1, 2 => 1, 4 => 1} )
+Poly ^ ZZ := BinaryPowerMethod
+assert( (f = T^0+T^1+T^2+T^3+T^4+T^5) === new Poly from {32 => 1, 16 => 1, 8 => 1, 1 => 1, 2 => 1, 4 => 1} )
 assert( ((T-one)*f) === new Poly from {64 => 1, 1 => -1} )
 assert( (f - (T^3 + T^5)) === new Poly from {16 => 1, 1 => 1, 2 => 1, 4 => 1} )
 assert( (tally {1,1,1,2,4} - tally {1,2}) === new Tally from {4 => 1, 1 => 2} )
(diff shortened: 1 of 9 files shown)


pzinn:
this may already have been mentioned, but I don't think this definition makes sense if `source f != target f`

d-torrance:
> this may already have been mentioned, but I don't think this definition makes sense if `source f != target f`

I agree, but I kept it consistent with `Matrix ^ ZZ`.  I'd be very happy to change both!

pzinn:
Ah, I see. yes, I would probably change both. The only risk is, this is a breaking change...

mahrud:
You might as well also remove the condition in `actors.d` that checks if the exponent is zero.

d-torrance:
> You might as well also remove the condition in `actors.d` that checks if the exponent is zero.

I'm a little hesitant to do that just in case someone were to use `BinaryPowerMethod` to implement exponentiation for some other class at some point where using the `T#1` feature for 0th powers would work.  It's actually a pretty cool feature -- if you write code for `*`, you can get `^` for free!

Here's a dumb example:

```m2
i2 : String#1 = "";

i3 : String * String := (s, t) -> s | t;

i4 : String^ZZ := BinaryPowerMethod;

i5 : "foo"^0

o5 = 

i6 : "foo"^3

o6 = foofoofoo
```

mahrud:
The problem is that because it's deep inside the interpreter, there's practically zero chance that someone will find out how that works and actually use it! If we had gone with my plan of turning `T#1` into a function, then I would have agreed with you, because then we could use it for numerous functions, and people would have a chance of noticing it. See for instance how this is re-implemented in the top-level:
https://github.com/Macaulay2/M2/blob/048c777dc703365e44c3865b2c1faed698b82b3a/M2/Macaulay2/m2/varieties.m2#L356-L366

d-torrance:
I like the `T#1` as a function idea a lot so I went ahead and made that change.  We also check that `source f` and `target f` agree for both matrices and ring maps.  And as suggested in https://github.com/Macaulay2/M2/issues/2792#issuecomment-1474181863, we use `BinaryPowerMethod` for matrix and monomial ideal exponentiation.  This results in a huge speed increase!

### Before
```m2
i1 : elapsedTime (random(ZZ^50, ZZ^50))^500;
 -- 58.8298 seconds elapsed
```


### After

```m2
i1 : elapsedTime (random(ZZ^50, ZZ^50))^500;
 -- 2.11871 seconds elapsed
```

d-torrance:
There's a failing test due to the change for 0th powers of matrices.  The previous behavior is really inconsistent, though:

```m2
i1 : R=ZZ/101[a,b];

i2 : f=matrix(R,{{1,a},{0,1}});

             2       2
o2 : Matrix R  <--- R

i3 : g=matrix(R,{{1,0},{b,1}});

             2       2
o3 : Matrix R  <--- R

i4 : h=f*g*f*g;

             2       2
o4 : Matrix R  <--- R

i5 : apply({(-1, 3, 0, 2), (-1, 3, 2, 0), (3, -1, 0, 2), (3, -1, 2, 0)},
    (i, j, k, l) -> h^i * h^j == h^k * h^l)

     
o5 = {false, false, false, true}
```

d-torrance:
I went ahead and reverted that change for now -- maybe whether sources and targets should agree is a discussion for another PR.

DanGrayson:
So, if the source and target are not the same, what does the 0-th power now yield?

DanGrayson:
Here's my question: if the source and target are not the same, why should we define f^n for just some values of n?

d-torrance:
> Here's my question: if the source and target are not the same, why should we define f^n for just some values of n?

This is a great question!

For the purposes of this PR, I'm proposing that integer powers of ring maps should behave the same as integer powers of matrices.  And so in particular, when the source and target disagree, the only values of n that work will be 0 and 1.

I'd be happy to change it, but as Paul mentioned above, it would be a breaking change if we change the matrix behavior, and at least one test will have to be removed/modified.

DanGrayson:
Okay, then, why should we do that for integer powers of matrices?

d-torrance:
> Okay, then, why should we do that for integer powers of matrices?

I don't know that we should -- that's just how it's been.  This is the test that fails if we require the source and target to agree:

https://github.com/Macaulay2/M2/blob/048c777dc703365e44c3865b2c1faed698b82b3a/M2/Macaulay2/tests/normal/000-core.m2#L617C1-L620

DanGrayson:
The test should be modified to make the source and target the same.

d-torrance:
I've made some changes so that `Matrix^ZZ` and `RingMap^ZZ` both raise an error when taking a 0th power if the source and target disagree:

```m2
i1 : R = QQ[x]; A = matrix {{x}};

             1       1
o2 : Matrix R  <--- R

i3 : A^0
stdio:2:2:(3): error: expected source and target to agree

i4 : S = QQ[y]; f = map(R, S, {x});

o5 : RingMap R <--- S

i6 : f^0
stdio:4:2:(3): error: expected source and target to agree
```

M2/Macaulay2/tests/normal/hashtables.m2
```diff
@@ -24,32 +26,6 @@ assert( class(tally{1,2,3}+set{3,4,5}) === Tally )
 assert( mutable( merge(new Type of List, new Type of Ring, first) ) )
 assert( class((new VirtualTally from {1 => -1, 2 => -2, 3 => -3}) + tally {1,2,3}) === VirtualTally )
 
-end
```
mahrud, line 27: Just curious, are we going towards not using `generateAssertions` anymore? Removing the bottom part makes it so you have to generate the assertions above by hand.

pzinn: didn't know `generateAssertions` even existed. that being said, I thought we switched to assertions being of the form `Equation { lhs, rhs }` ?

mahrud: The point of `generateAssertions` is to not have to write the `rhs` manually.

pzinn: I understand that. mine was that it doesn't generate assertions of the form `Equation{...}`.

mahrud: Ah, yes. `generateAssertions` is way older, definitely before `Equation { lhs, rhs }` was a thing.

d-torrance: I don't see the point of keeping the `generateAssertions` code in the file.  Once you've generated the assertions and have pasted the desired calls to `assert` for your tests, then it's just cruft.

PR #2815, opened 2023-04-16
Add "member" as alias for "isMember" for RRi's

d-torrance:
As suggested by Luke Oeding at MAAG.


d-torrance:
Converting to a draft for now, as after some discussion at MAAG, there are better solutions.  One possibility that came up is to convert `member` to `isMember` everywhere, and maybe add `member` as a synonym for now, possibility to be removed at some point.

mahrud:
fwiw I prefer `isMember` and `isMutable` over `member` and `mutable` as a convention.

d-torrance:
I've updated this so that there's only one method, `isMember`, with `member` as a synonym.  I only bothered changing instances in `Core`.

d-torrance:
I went ahead and also switched `mutable` -> `isMutable` (retaining `mutable` as a synonym) and also fixed a display issue with an error I ran into while working on this:

### Before
```m2
i1 : foo = 2

o1 = 2

i2 : export{"foo" => "bar"}
stdio:2:1:(3): error: symbol intended as exported synonym already used internally: "foo"
new FilePosition from {"stdio",1,0,1,3,$: it was used here
```

### After
```m2
i1 : foo = 2

o1 = 2

i2 : export {"foo" => "bar"}
stdio:2:1:(3): error: symbol intended as exported synonym already used internally: "foo"
stdio:1:0-1:3: it was used here
```

DanGrayson:
Let's deprecate "member".  If code out there runs "member", there should be an error message stating that it has been replaced by "isMember".

d-torrance:
> Let's deprecate "member". If code out there runs "member", there should be an error message stating that it has been replaced by "isMember".

What about `mutable`?  Should using it also raise an error saying to use `isMutable`?

mahrud:
Isn't it a bit rushed to deprecate them without warning, 2 weeks before a release?

DanGrayson:
> Isn't it a bit rushed to deprecate them without warning, 2 weeks before a release?

No, the error message is the warning.

DanGrayson:
> What about `mutable`? Should using it also raise an error saying to use `isMutable`?

Good idea.

d-torrance:
Why not a warning instead of an error?  Maybe we could even introduce a function like Elips's `make-obsolete`:

```m2
makeObsolete = method()
makeObsolete(Symbol, Function, String) := (oldfunc, newfunc, ver) -> (
    oldfunc <- x -> (
	printerr("warning: the function ", format toString oldfunc,
	    " is obsolete since ", ver, "; use ", format toString newfunc,
	    " instead.");
	newfunc x);)
```

Which gives:

```m2
i3 : foo = () -> "Hello, world!"

o3 = foo

o3 : FunctionClosure

i4 : makeObsolete(bar, foo, "1.22")

i5 : bar()
 -- warning: the function "bar" is obsolete since 1.22; use "foo" instead.

o5 = Hello, world!
```

Then we could just put `makeObsolete(member, isMember, "1.22")` and `makeObsolete(mutable, isMember, "1.22")` somewhere to warn folks without breaking a bunch of code.

DanGrayson:
But we want to break the code -- that's the only reliable way to replace all  the uses of "member" in our own code.

d-torrance:
I think we should break the code eventually -- warn users for a year or so to give them time to switch and then remove `member` and `mutable` after that.

PR #2817, opened 2023-04-23
Switch to llvm@14 for Github Actions

mahrud:


DanGrayson:
Mahrud, does this fix something?

mahrud:
yes, nightly test have been failing for the past few weeks: https://github.com/Macaulay2/M2/actions/runs/4933802672

PR #2840, opened 2023-05-16
SagbiGbDetection

timduff35:
This PR introduces the package "SagbiGbDetection", coauthored with [Elima Shehu](https://github.com/elimashehu) and [Viktoriia Borovik](https://github.com/V-Borovik). This package can be used to decide whether or not a list of polynomials forms a Groebner basis for the ideal it generates. Similar functionality is provided for SAGBI bases of finitely-generated polynomial algebras.

A few minor changes to the package "MonodromySolver" are also included.


mahrud:
If you install the package with `debugLevel=1` you will get two warnings:
1. from SubalgebraBases:
```m2
 -- warning: newPackage: use the Contributors or Acknowledgement keywords to acknowledge contributors in the package documentation
```
2. from SagbiGbDetection:
```m2
 -- warning: document: documentation headlines must be less than 100 characters long:
```

timduff35:
Thanks, should be taken care of now. I can't tell, is that what caused actions to fail before?

mahrud:
I think something else was failing before:
```m2
 -- warning: element of type PARA may not contain an element of type UL
currentString:1:1:(3):[2]: error: validation failed: PARA{TEX{"See documentation nodes for t$
```

Also, when using SimpleDoc, there is a special keyword `Contributors` (and `Acknowledgement` and `References`) that you can use in place of `Text`; e.g.:
https://github.com/Macaulay2/M2/blob/048c777dc703365e44c3865b2c1faed698b82b3a/M2/Macaulay2/packages/VirtualResolutions/doc.m2#L68-L69

timduff35:
Great, I fixed some things in documentation and added those keywords, and now all checks seem to pass.

PR #2841, opened 2023-05-17
Latest version of Complexes package

mikestillman:
This package is now almost fully documented.

@DanGrayson If it passes tests, it can be merged in.


mahrud:
One small wishlist item that I meant to tell you soon: is there a reason that complexes are never cached? For instance:
```m2
i5 : hash complex R

o5 = 1543866

i6 : hash complex R

o6 = 1543894

i7 : complex R === complex R

o7 = false
```
This is kind of annoying because it fills up hash tables:
```m2
i8 : tally apply(10, i -> complex R)

            1
o8 = Tally{R  => 1}
            
           0
            1
           R  => 1
            
           0
            1
           R  => 1
            
           0
            1
           R  => 1
            
           0
            1
           R  => 1
            
           0
            1
           R  => 1
            
           0
            1
           R  => 1
            
           0
            1
           R  => 1
            
           0
            1
           R  => 1
            
           0
            1
           R  => 1
            
           0

o8 : Tally
```

DanGrayson:
@mikestillman -- a test from the book is failing...

DanGrayson:
A timing test in the Complexes package is failing.

mikestillman:
@DanGrayson I will fix the two failing tests in Complexes, and then hopefully I am done and we can finalize M2 1.22. (I just got back to Ithaca from my travels, and so have more time now, so it should go faster).

mikestillman:
@DanGrayson @mahrud @d-torrance @ggsmith .  The plan is to do a final commit fixing the failing test(s) later today.  If that in fact goes through with no problems, then the release is ready to go.

mahrud:
https://github.com/Macaulay2/homebrew-tap/issues/181#issue-1714706144 is still preventing brew builds to finish on macOS also, and it's possible a fix in the source might be needed. Could you take a look?

DanGrayson:
Yes, the tests have passed, merging ...

mahrud:
Since the workshop is getting very close, could you make a tag and begin the process of a new release? (I'll try to fix the brew compile issue as a patch in the formula)

DanGrayson:
Will do...

PR #2850, opened 2023-05-22
Port documentation to HTML5

d-torrance:
We begin porting the documentation from XHTML 1.1 -> HTML5.  For the most part, it's pretty simple -- changing the `<!DOCTYPE>` and `<html>` tags at the top of each page does most of the work already.

The biggest issue is we have a *ton* of `<tt>` tags scattered all over the place, and it's been removed.  Right now, I'm proposing a hacky solution:

```m2
i2 : html TT "foo"

o2 = <span style="font-family: monospace">foo</span>
```

Ideally, it would be good to change all of these to `<code>`, `<kbd>`, `<samp>`, or `<var>`, but that would be a monumental task (`grep TT -r | wc -l` in the `packages` directory returns 7938, and that's not even counting the docs that use `{\tt ... }`!)  So this `<span>` solution is probably the most realistic.

Still to do:

- [x] Add a GitHub action that validates the documentation every few days or so, and maybe even on certain pull requests (e.g., if a PR modifies a package, validate that package's documentation, or if it modifies `Core` stuff dealing with documentation like `html.m2`, go ahead and validate everything).
- [x] Fix any HTML5 validation issues that pop up after building all the documentation -- so far, I've only checked `viewHelp foo`.


d-torrance:
> is there any way to have the style transferred to CSS, like `<span class='tt'>`?

Yes, definitely -- good idea!

d-torrance:
> Keep in mind that the documentation also should be easily convertible to LaTeX (see `book.tex`), where there are converted to `{\tt ...}` or `\texttt{...}`, so I don't think you should get rid of them.

Great point! And a good excuse not to make massive changes. :)

pzinn:
comparing with my branch:
- you probably want to add the standard HTML attributes to the new `SAMP`, `KBD`, etc?
- `TTc` in `hypertext.m2` probably needs to be replaced with something else. I'd chosen `SAMP` but not sure which one is best...

d-torrance:
@pzinn - I cherry-picked the commit from your branch that does these things.  There were some conflicts that I tried to resolve as best as I could.  Do you see any issues?

pzinn:
yeah sorry that branch was based on my current fork which has strayed pretty far from main... let me look

pzinn:
looks good

d-torrance:
I think this PR is ready.

@DanGrayson - what's the status of the 1.22 release?  I wasn't originally planning on this being included, but if it will be, then I should add a changelog entry, too.

DanGrayson:
Mike wants the decompose bug fix included in 1.22, so I'm waiting for that.

d-torrance:
Ok, I'll open a new PR here shortly with a couple HTML-related blurbs added to the changelog.

pzinn:
oops, I missed a problem with `html TT` in this PR (it should run `fixNewLines`). I can submit a PR for that.

d-torrance:
@DanGrayson - I'm getting an error viewing the 1.22 documentation on the webpage (e.g., https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.22/share/doc/Macaulay2/Macaulay2Doc/html/_changes_cm_sp1.22.html), almost certainly due to the changes from this PR.  Is the web server still set to serve the docs as XHTML?

d-torrance:
According to my browser, the server is running NGINX, not Apache, and so `.htaccess` won't work (see https://www.nginx.com/resources/wiki/start/topics/examples/likeapache-htaccess/).  It looks like NGINX uses `/etc/nginx/mime.types`.

M2/Macaulay2/m2/html.m2
pzinn, line 1: I missed this commit.
Unfortunately that breaks my `VectorGraphics` package (svg objects seem to care whether they have `/>` or `>`)

d-torrance: Apparently svg objects are "[foreign elements](https://www.w3.org/TR/2010/WD-html5-20101019/syntax.html#foreign-elements)" instead of void.  Would switching them to one of the other children of `Hypertext` so that they get a closing `</svg>` tag help?

pzinn: yeah that works. I'm still not sure I see the point of this commit -- I find it more satisfying to see `<br/>` than `<br>`!

d-torrance: If I remember, I think the HTML5 validator gave warnings for self-closing tags.

pzinn: I see. that's sad but OK.

d-torrance: Just confirmed it -- this is the warning:

> Trailing slash on void elements [has no effect](https://github.com/validator/validator/wiki/Markup-%C2%BB-Void-elements#trailing-slashes-in-void-element-start-tags-do-not-mark-the-start-tags-as-self-closing) and [interacts badly with unquoted attribute values](https://github.com/validator/validator/wiki/Markup-%C2%BB-Void-elements#trailing-slashes-directly-preceded-by-unquoted-attribute-values).

PR #2852, opened 2023-05-26
Dg algebras update

moorewf:
This update to DGAlgebras allows for one to see the differentials of the underlying complex as block matrices.  The idea for this addition, as well as the documentation for these new features, were provided by @eisenbud, with the features themselves added by @moorewf.


pzinn:
you know block matrices are natively displayed?
```
i1 : blockMatrixForm=true; random(ZZ^3++ZZ^2,ZZ^4++ZZ^1)

o2 = | 9 4 7 2 | 4 |
     | 2 0 9 8 | 0 |
     | 4 7 0 2 | 1 |
     +---------+---+
     | 9 0 0 3 | 8 |
     | 9 5 9 8 | 6 |

              5        5
o2 : Matrix ZZ  <--- ZZ
```

moorewf:
Nice!  Is there a way to have the indices of the direct summands be printed as well?

pzinn:
what are indices of direct summands? maybe I need an example to understand.

moorewf:
You can supply labels for the direct summands, and they are stored in the cache of the module returned by directSum, and returned by the function "indices".

```
i1 : V = directSum {"foo" => ZZ^2, "bar" => ZZ^2}

       4
o1 = ZZ

o1 : ZZ-module, free

i2 : blockMatrixForm = true;

i3 : random(V,V)

o3 = | 8 5 | 4 6 |
     | 1 9 | 9 8 |
     +-----+-----+
     | 7 4 | 6 6 |
     | 2 0 | 1 4 |

              4        4
o3 : Matrix ZZ  <--- ZZ

i4 : indices V

o4 = {foo, bar}

o4 : List
```

It would be nice if there was an option for something like:

```
          | foo | bar |
     -----+-----+------
o3 =  foo | 8 5 | 4 6 |
          | 1 9 | 9 8 |
     -----+-----+-----+
      bar | 7 4 | 6 6 |
          | 2 0 | 1 4 |
```

pzinn:
I see. TODO!

M2/Macaulay2/packages/DGAlgebras.m2
```diff
@@ -1,26 +1,64 @@
...
      Headline => "Data type for DG algebras",
-     Version => "1.1.0",
+     Version => "1.1.1--with fix of killCycles and new code and docs for displayBlockDiff and blockDiff",
      Date => "August 20, 2020",
      Authors => {
 	  {Name => "Frank Moore",
 	   HomePage => "http://www.math.wfu.edu/Faculty/Moore.html",
-	   Email => "moorewf@wfu.edu"}},
+	   Email => "moorewf@wfu.edu"},
+       {Name => "Some documentation added by Daniel Rostamloo and David Eisenbud"}},
```
mahrud, line 10: We'd rather you use the [Contributors](https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/___Contributors.html) key in the documentation. This "name" is going to mess up the list of package [authors](https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_authors_spof_sp__Macaulay2_sppackages.html).

pzinn:
I remember now why I didn't do it. To see, the problem, let's continue your example:
```
i6 : W=V++ZZ^1

       5
o6 = ZZ

o6 : ZZ-module, free

i7 : random(V,W)

o7 = | 5 0 | 0 9 | 1 |
     | 1 7 | 2 5 | 5 |
     +-----+-----+---+
     | 2 4 | 5 5 | 2 |
     | 6 9 | 5 1 | 6 |

              4        5
o7 : Matrix ZZ  <--- ZZ
```
all's well, there are 3 horizontal blocks. but the problem is, the labels have been lost... inversely, components of a summand don't know their own label, so there's no way to recurse to get the label names.

M2/Macaulay2/packages/DGAlgebras.m2
```diff
@@ -1,26 +1,64 @@
--- -*- coding: utf-8 -*-
+-*- coding: utf-8 -*-
 newPackage("DGAlgebras",
      Headline => "Data type for DG algebras",
-     Version => "1.1.0",
+     Version => "1.1.1--with fix of killCycles and new code and docs for displayBlockDiff and blockDiff",
```
mahrud, line 4: FYI, this has the following effect:
image
I suspect this wasn't intended?

pzinn:
To continue the previous message, in order to get the output routines to work in a reasonable manner while keeping track of indices, I think indices have to be reprogrammed better. A minimal change would be to move the field `indices` in the cache to the components while keeping `indexComponents` in the full object. But this is probably opening a can of worms since indices seems to be used in a bunch of places in Core...

PR #2853, opened 2023-05-27
Quote string hash table keys

d-torrance:
Otherwise, it is difficult to distinguish which keys are symbols and which keys are strings.

This is something we talked about at an M2internals a few months ago.

### Before:

```m2
i1 : hashTable{"foo" => null, foo => null}

o1 = HashTable{foo => null}
               foo => null

o1 : HashTable
```

### After

```m2
i1 : hashTable{"foo" => null, foo => null}

o1 = HashTable{foo => null  }
               "foo" => null

o1 : HashTable
```

### DIFF
--- a/M2/Macaulay2/m2/nets.m2
+++ b/M2/Macaulay2/m2/nets.m2
@@ -190,7 +190,9 @@ net HashTable := x -> (
      	  net class x,
 	  "{", 
 	  -- the first line prints the parts vertically, second: horizontally
- 	  stack (horizontalJoin \ apply(sortByName pairs x,(k,v) -> (net k, " => ", net v))),
+ 	  stack (horizontalJoin \ apply(sortByName pairs x,(k,v) -> (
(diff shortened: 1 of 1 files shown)


mahrud:
I don't feel like this is a very pressing issue, but if you're changing the keys then you could say the same about the values:
```m2
i2 : new HashTable from { "a" => "a", a => a }

o2 = HashTable{a => a}
               a => a

o2 : HashTable
```

d-torrance:
At least for now, I wanted to focus on keys since we need to know the types so we can get the corresponding values (do we use `foo#bar` or `foo#"bar"`?)

I think quoting values is probably a worthwhile thing to do at some point, but it should probably also be done with strings inside other containers like lists, which also have the same issue:

```m2
i1 : {"foo", foo}

o1 = {foo, foo}

o1 : List
```

I seem to remember playing around with this once and having undesired consequences.  It's been a while, so I can remember what they were...

PR #2854, opened 2023-05-27
minor HTML/CSS, doc

pzinn:
- fix for #2853 as mentioned in comments. I also renamed `fixNewLines` to something more meaningful.
- minor doc improvements
- when I introduced `INDENT` in #2760, I forgot to export it and to PR the corresponding css. fixed.

### DIFF
--- a/M2/Macaulay2/m2/html.m2
+++ b/M2/Macaulay2/m2/html.m2
@@ -120,20 +120,21 @@ html HTML := x -> demark(newline, {
 treatImgSrc := x -> apply(x, y -> if class y === Option and y#0 === "src" then "src" => toURL y#1 else y)
 html IMG := (lookup(html, IMG)) @@ treatImgSrc
 
-fixNewLines := method()
-fixNewLines Hypertext :=
-fixNewLines Nothing :=
-fixNewLines Option := identity
-fixNewLines Thing := x -> replace("\r\n","\n",toString x) -- toString prevents LaTeX being inserted...
+toStringMaybe := method()
+toStringMaybe Hypertext :=
+toStringMaybe Nothing :=
+toStringMaybe OptionTable :=
+toStringMaybe Option := identity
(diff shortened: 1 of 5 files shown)


M2/Macaulay2/m2/html.m2
```diff
@@ -120,20 +120,21 @@ html HTML := x -> demark(newline, {
 treatImgSrc := x -> apply(x, y -> if class y === Option and y#0 === "src" then "src" => toURL y#1 else y)
 html IMG := (lookup(html, IMG)) @@ treatImgSrc
 
-fixNewLines := method()
-fixNewLines Hypertext :=
-fixNewLines Nothing :=
-fixNewLines Option := identity
-fixNewLines Thing := x -> replace("\r\n","\n",toString x) -- toString prevents LaTeX being inserted...
+toStringMaybe := method()
```
mahrud, line 123: Personally I feel like `fixNewLines` was much more descriptive of what the method does. `toStringMaybe` doesn't say anything about literally the only change that this function makes!

pzinn:
The point is, the change it does to strings is I suspect useless. The real purpose of this method (in my opinion) is to prevent mathematical expressions, say `1/2`, of being turned into LaTeX expressions, in this case, `$\frac{1}{2}$`, which would then be printed verbatim. It does so by applying `toString` hence the new name...

mahrud:
Why would there be mathematical expressions inside `CODE` or `TT`? Those should have been converted to string before this point.

pzinn:
I don't want to reopen the discussion in #2757. Let me just point out that right now, nothing prevents you from inputting `TT {1/2}` and it's better for the output to be `1/2` than `$\frac{1}{2}$`.

pzinn:
I don't want to reopen the discussion in #2757. Let me just point out that right now, nothing prevents you from inputting `TT {1/2}` and it's better for the output to be `1/2` than `$\frac{1}{2}$`.

Issue #2858, opened 2023-06-02
WITH_OMP=off fails in cmake build


mikestillman:
Adding -DWITH_OMP=off to the cmake line appears to have no effect.  I still get the `-Xclang` bug even though it should not be using openmp any longer? @mahrud This is in a newly created build directory.

mahrud:
The reason is that `WITH_OMP=OFF` means CMake doesn't _require_ OpenMP, but it still uses it opportunistically:
https://github.com/Macaulay2/M2/blob/6cba5590cebc8914ba82dbe3d6bc0bedf4403424/M2/cmake/check-libraries.cmake#L61-L65
You could comment out the second `find_package` here, but I can make a patch soon.

PR #2859, opened 2023-06-02
keep OpenMP flags together

mahrud:
Fixes https://github.com/Macaulay2/homebrew-tap/issues/181 and https://github.com/Macaulay2/M2/issues/2858 and part of https://github.com/Macaulay2/M2/issues/2809

Not sure if this is having the intended effect, but I also don't have time to fix it currently. If anyone wants to finish this feel free to.

### DIFF
--- a/M2/cmake/check-libraries.cmake
+++ b/M2/cmake/check-libraries.cmake
@@ -73,6 +73,8 @@ foreach(lang IN ITEMS C CXX)
     string(REGEX REPLACE "^lib" "" _lib "${_lib}")
     set(OpenMP_${lang}_LDLIBS "${OpenMP_${lang}_LDLIBS} -L${_libdir} -Wl,-rpath,${_libdir} -l${_lib}")
   endforeach()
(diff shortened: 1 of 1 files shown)


mikestillman:
This problem is only relevant on MacOS using the system clang, which has a braindamaged openmp: it requires to have two parameters next to each other to announce usage of openmp: `-Xclang -fopenmp`, and one must link in libomp on one's own.  cmake version 3.26.4 (or some software in that release) sees two -fopenmp's on one command line and removes the second one, which is an error.

Version 3.26.4 of cmake has this problem, but if you use cmake 3.24.2 (e.g. which currently comes with CLion), the problem does not appear, and the M2-binary executable is correctly built.

The fix above did not work on my Mac, running on Apple M1.

mahrud:
I understand. I filed an issue with CMake and they already merged a patch in v3.27.0.

> The fix above did not work on my Mac, running on Apple M1.

Yes, I realized this wasn't working, so I made it a draft and cancelled the tests.

Issue #2857, opened 2023-06-01
Duplicate large block deallocation in AssociativeAlgebras::oppositeRing example


d-torrance:
Here's the output from the failing build (`AssociativeAlgebras::oppositeRing` example):

```m2
-- -*- M2-comint -*- hash: -1318199081

i1 : R = QQ[q]/ideal{q^4+q^3+q^2+q+1}

o1 = R

o1 : QuotientRing

i2 : A = skewPolynomialRing(R,q,{x,y,z,w})
Using numthreads = 0
Using GC ring in VectorArithmetic.

o2 = A

o2 : FreeAlgebraQuotient

i3 : x*y == q*y*x

o3 = true

i4 : Aop = oppositeRing A
Using numthreads = 0
Using GC ring in VectorArithmetic.
Duplicate large block deallocation
Aborted (core dumped)
```
I haven't seen this particular error before, but there are few known issues that have given similar output and might be related (#35, #151, #679, #714, #2119).

_Originally posted by @d-torrance in https://github.com/Macaulay2/M2/issues/2855#issuecomment-1570964773_

d-torrance:
This just happened in an [Ubuntu 22.04 PPA build](https://launchpadlibrarian.net/669616315/buildlog_ubuntu-jammy-amd64.macaulay2_1.22~rc1+ds-1~ubuntu22.04.1_BUILDING.txt.gz) as well, so I made it an issue.

d-torrance:
It [happened again](https://launchpadlibrarian.net/669656020/buildlog_ubuntu-jammy-amd64.macaulay2_1.22~rc1+ds-1~ubuntu22.04.1_BUILDING.txt.gz)!  Only Ubuntu 20.04 so far, which is the same release that GitHub uses.  This time we got a stack trace:

```m2
_opposite__Ring.errors
**********************
o2 : FreeAlgebraQuotient

i3 : x*y == q*y*x

o3 = true

i4 : Aop = oppositeRing A
Using numthreads = 0
Using GC ring in VectorArithmetic.
-- SIGSEGV
-* stack trace, pid: 114773
 0# stack_trace(std::ostream&, bool) at ./M2/Macaulay2/d/main.cpp:123
 1# segv_handler at ./M2/Macaulay2/d/main.cpp:240
 2# 0x00007FAB7A8BC520 in /lib/x86_64-linux-gnu/libc.so.6
 3# GC_free in /lib/x86_64-linux-gnu/libgc.so.1
[...]

PR #2862, opened 2023-06-06
Add some more escape sequences for strings

d-torrance:
I noticed that Macaulay2 was missing a few string escape sequences that are common in other languages.  The following have been added:

* `\a` (audible bell)
* `\e` & `\E` (escape)
* `\v` (vertical tab)
* `\xnn` (character w/ hexadecimal ASCII encoding `nn`)

All of these are available in C, for example (`\e` and `\E` are GCC extensions).

Fun fact - you can get colors in Macaulay2 using [ANSI escape codes](https://en.wikipedia.org/wiki/ANSI_escape_code).  For example, `"\e[31mfoo\e[m"` is `"foo"`, but red.  (It might be cool to use this for syntax highlighting inside the REPL, but right now `net` thinks all these extra characters are visible and the spacing gets messed up.)

### DIFF
--- a/M2/Macaulay2/d/lex.d
+++ b/M2/Macaulay2/d/lex.d
@@ -187,7 +187,8 @@ getstring(o:PosFile):(null or Word) := (
 		    hexcoming = hexcoming - 1;
 		    )
 	       else (
-		    printErrorMessage(o.filename,line,column,"expected hex digit in unicode sequence here");
+		    printErrorMessage(o.filename,line,column,"expected " +
+			tostring(hexcoming) + " more hex digit(s)");
 		    empty(tokenbuf);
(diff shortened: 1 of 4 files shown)


d-torrance:
Ubuntu builds are all failing with #2857.

DanGrayson:
Doug, do you think you can fix the failing builds?

d-torrance:
> Doug, do you think you can fix the failing builds?

They're unrelated to this pull request -- both failing builds are instances of #2857.

mikestillman:
@d-torrance Could you merge development into this pull request, as well as your other outstanding pull requests? This hopefully will make the errors during github builds go away

d-torrance:
> @d-torrance Could you merge development into this pull request, as well as your other outstanding pull requests? This hopefully will make the errors during github builds go away

Done!

PR #2871, opened 2023-06-09
resubmit package

mperlm3:
there was a test which was failing. We believe it is now working.


d-torrance:
The builds are failing because during the build, Macaulay2 is run using the `--no-preload` option, so the `Classic` package that defines the `(ideal, String)` method isn't available.

A simple fix would be to add the `PackageImports => {"Classic"}` option to `newPackage`.

mikestillman:
@mperlm3 Can you merge development into your branch, and fix your =distributed file?  Then we can accept this pull request (if tests pass!)

DanGrayson:
Mike, what's wrong with the =distributed-packages file changes?

mikestillman:
@DanGrayson @mperlm3 I just changed the `=distributed` file.  I'm not sure why it gave a conflict.

M2/Macaulay2/packages/QuadraticIdealExamplesByRoos.m2
```diff
@@ -0,0 +1,417 @@
...
+ "depthZeroTable",
+ "almostKoszul",
+ "roosIsotopes",
+ "onedimToricIrrationalPoincare",
+ "twodimToricIrrationalPoincare"
+}
+
+
+---TO DO: make this a method to allow user to pick coefficient field
+roosTable = (S = QQ[x,y,z,u];
```
mahrud, line 28: If someone knows better correct me if I'm wrong, but my understanding is that defining global rings when the package loads can be problematic.

mikestillman:
I thought I fixed the `=distributed` conflicts...

M2/Macaulay2/packages/QuadraticIdealExamplesByRoos.m2
mikestillman: I need to look, but that is probably why the package installation is failing now?

mperlm3:
I just pushed what I have. Tell me what to do to get this taken care of.

Thanks,
Mike P

mikestillman:
Hi @mperlm3, I'll look at what is going on with your package.  But the stuff you just added in doesn't seem like what you probably meant to add?

d-torrance:
See [my comment above](https://github.com/Macaulay2/M2/pull/2871#issuecomment-1584999212):

> The builds are failing because during the build, Macaulay2 is run using the `--no-preload` option, so the `Classic` package that defines the `(ideal, String)` method isn't available.
> 
> A simple fix would be to add the `PackageImports => {"Classic"}` option to `newPackage`.

mperlm3:
Since my icloud autosyncs, I think I just added some files on accident. Sorry about that!

Can someone please do me a favor and make the change @d-torrance suggested, so that I do not make another mistake? Or do I need to do it?

-Mike P

d-torrance:
I just opened the pull request https://github.com/mperlm3/M2/pull/1 at your fork.  If you merge it, it should fix everything.  In particular it:

* removes the various autosync files.
* merges in the `development` branch, fixing the `=distributed-packages` conflcit.
* adds the `Classic` import I mentioned above

mperlm3:
Ok, I think I just did it. Thanks for your help!

mikestillman:
One test (codespell) is failing.  Codespell wants the following change. Should I just make that change?  (I also need to look at @mahrud's question before merging it.)
 M2/Macaulay2/packages/QuadraticIdealExamplesByRoos.m2:339: alpha-numeric ==> alphanumeric

PR #2878, opened 2023-06-14
Update prism.js symbols for version 1.22.

d-torrance:
This commit was created using `make -C Macaulay2/editors update-syntax-highlighting` as outlined in `VERSION-README`.


mahrud:
I presume M2-symbols.el also needs to be updated, no?

Issue #2877, opened 2023-06-10
foreign functions


DanGrayson:
I get this error when compiling on Mac OS 13.4:

../../../../../../../Macaulay2/packages/ForeignFunctions.m2:566:9:(2):[69]: error: dlsym(RTLD_DEFAULT, GC_malloc): symbol not found
../../../../../../../Macaulay2/m2/methods.m2:154:80:(1):[67]: --back trace--
../../../../../../../Macaulay2/m2/option.m2:17:8:(1):[66]: --back trace--
../../../../../../../Macaulay2/packages/ForeignFunctions.m2:563:30:(2):[65]: --back trace--
../../../../../../../Macaulay2/m2/methods.m2:154:80:(1):[64]: --back trace--
../../../../../../../Macaulay2/m2/option.m2:17:8:(1):[63]: --back trace--
../../../../../../../Macaulay2/packages/ForeignFunctions.m2:617:12:(2):[62]: --back trace--
../../../../../../../Macaulay2/m2/startup.m2.in:66:17:(0):[62]: --back trace--

@d-torrance --- any clues for me?

d-torrance:
Are you linking against libgc statically or dynamically?  If it's the former, then I don't think `dlsym` would be able to find any of its symbols.  If it's the latter, then the only thing I can think of is maybe they `#define`'d `GC_MALLOC` to something other than `GC_malloc` in macOS 13.4 for some reason.

mikestillman:
When compiling on autotools on MacOS, apple M1, I get 4 example errors in ForeignFunctions, basically it can't find mpfr or mps, e.g. the code snippet below.  In fact, the library is in the `/opt/homebrew` tree (and that location isn't being checked here).
```
i1 : openSharedLibrary "mpfr"
[...]

PR #2881, opened 2023-06-15
Also check for "arm" for Apple silicon dlopen hack

d-torrance:
We also switch member -> isMember

---

This should fix https://github.com/Macaulay2/M2/issues/2877#issuecomment-1593283953


mikestillman:
@d-torrance Several of your pull requests are still failing, but now the failure is the html validation problem, which I think was one of the pull requests that I just accepted.  Could you try merging development in again on these, and if they pass, I'll accept them

d-torrance:
> @d-torrance Several of your pull requests are still failing, but now the failure is the html validation problem, which I think was one of the pull requests that I just accepted. Could you try merging development in again on these, and if they pass, I'll accept them

Done!

PR #2886, opened 2023-06-21
Use clang-tidy to use nullptr instead of 0 or NULL in e directory.

moorewf:


mahrud:
This is going to heavily conflict with a ton of changes I worked on with Mike!

mahrud:
There are scripts (like git-clang-format) that automatically clang-format the code but **only** the parts that you are changing in a given commit. Can we set up something like that, instead of changing everything in the engine all at once?

moorewf:
Sorry about that!  Mike and I were experimenting with some clang-tidy functionality and thought this would be a good step toward some cleaning of the code.

mikestillman:
@mahrud This doesn't change formatting, it is only using clang-tidy to change 0, NULL to nullptr's.  But perhaps we should incrementally add in your changes on what we were working on (and others too) that might conflict with this, then we can rerun this.

mahrud:
I also meant to suggest setting up something _like_ git-clang-format but with clang-tidy.

mikestillman:
The script is in e/util.

@mahrud What branch are these changes on?

mahrud:
Here is my alternative solution using [clang-tidy-diff.py](https://github.com/llvm/llvm-project/blob/main/clang-tools-extra/clang-tidy/tool/clang-tidy-diff.py):
```
git diff -U0 --staged | clang-tidy-diff.py -p1 -fix -path=M2/BUILD/build -checks="-*,modernize-use-nullptr"
```
You need to run this from the top of the git repository. This will feed the diff of staged files only to clang-tidy and correct those instances of null only. Those changes will _not_ be staged automatically, so you get a chance to review them first and add them.

We can add it as a [pre-commit hook](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) so that it's applied on every commit automatically, but I strongly prefer that we close this pull request.

mahrud:
> @mahrud What branch are these changes on?

[feature/monoids](https://github.com/mahrud/M2/tree/feature/monoids)

mahrud:
@mikestillman was this meant to be merged?

mikestillman:
@DanGrayson Oops, this was not meant to be merged.  I thought it was a draft pull request.

mikestillman:
It will cause a lot of conflicts in another branch we are working on, confusing that project.

mahrud:
> The script is in e/util.

I tried to use the script to rerun this, now that the other large PR is merged, but it doesn't correct nearly as many cases as were fixed in this PR. Maybe something isn't working on my machine?

PR #2892, opened 2023-06-27
Converted mutable matrices to matrices in instances of `sub()`

bstone:
At some point M2 no longer allowed for mutable matrices to be used in `sub()`. Instances of this was fixed by converting the mutable matrices to a matrix.


bstone:
I forgot to mention, this is an update for `QuillenSuslin.m2`. Updated author info as well.

mahrud:
@mikestillman @DanGrayson this PR seems to have been merged into master only, hence the commit isn't present in development, which may be contributing to merging issues in other pull requests.

bstone:
Was this a mistake on my part? Should I be making the PRs to a development branch instead?

mahrud:
You should make PRs to development branch, but before merging maintainers can also switch the base branch if it's not correct, but that wasn't noticed in this case. I don't think you have to do anything else now.

Issue #2889, opened 2023-06-25
write a lark parser for the M2 script language


fchapoton:
This would allow:
- to have a formal definition of the language
- to be able to automatically check all `.m2` files for correctness
- maybe open the way for better syntax highlighting in text editors

See https://github.com/lark-parser/lark

In attachment, a first tentative, just a sketch.

[macaulay2.lark.txt](https://github.com/Macaulay2/M2/files/11859072/macaulay2.lark.txt)

how to use: inside an ipython or python session,
```
from lark import Lark
with open("macaulay2.lark") as f:
    L = Lark(f.read(), start="expr")
parsed = L.parse("4+4")
print(parsed.pretty())
```

fchapoton:
What would be a short and simple script file to serve as good first testing ground ?

Maybe adding a "boolean_expr" would be a good idea ?

mikestillman:
This sounds interesting!  Perhaps having the syntax for making methods would be good to place in the grammar too?  Although there is no real new syntax there, perhaps, it seems an important piece.

I have not used lark.  I'll try it out.

fchapoton:
my new tentative so far. This manages to parse one-line things, but not serious scripts. Sigh. Not clear to me how to handle the ";".

file:///home/chapoton/macaulay2.lark

d-torrance:
I've done some similar work for [Tree-sitter](https://tree-sitter.github.io/tree-sitter/) (see https://github.com/d-torrance/tree-sitter-macaulay2).  I've tried to base it on how the Macaulay2 interpreter does its parsing.  The code is scattered between several different files in the d directory and it's pretty confusing!  It would be really cool if we could put together an official description of  the Macaulay2 grammar using EBNF to help with writing tools like this.

watsonkm:
Ooh, I'd be interested in helping with this!

I've been working on digging through the parser internals for a few blog posts I've been writing--would it be useful to anyone if I helped draft out a tentative EBNF grammar from what I understand of the parser?

d-torrance:
> would it be useful to anyone if I helped draft out a tentative EBNF grammar from what I understand of the parser?

Yes, absolutely!

watsonkm:
Perfect! I'll toss the base of one together over the next few evenings and then add a MR to share :)

watsonkm:
(apologies for delay! things got a bit busier than intended, but I definitely will be putting together a PR for this in the near future.)

d-torrance:
[...]

PR #2893, opened 2023-06-27
adding a prototype of lark file

fchapoton:
in relation with #2889 ; here as a pull request

works a little bit, but needs works by people used to the language

### DIFF
--- a/M2/Macaulay2/macaulay2.lark
+++ b/M2/Macaulay2/macaulay2.lark
@@ -13,8 +13,6 @@ FUNCTION: "sin" | "cos" | "tan" | "abs" | "acos"
         | "zeta" | "tanh" | "sqrt" | "sec" | "sech"
         | "erf" | "erfc" | "coth" | "csc" | "eint"
 
-BOOLEAN: "true" | "false"
-
 expr: SYMBOL
     | string
     | SIGNED_NUMBER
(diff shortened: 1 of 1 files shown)


fchapoton:
how to avoid triggering the Github CI at each push ?

fchapoton:
thanks.
I feel that I am not familiar enough with the M2 script  language to push this tentative much further.

mahrud:
If it's just a single file, not needed or used by anything, perhaps putting it on a wiki page would be more appropriate?

fchapoton:
yes, indeed. is there a wiki for M2 somewhere ?

d-torrance:
> yes, indeed. is there a wiki for M2 somewhere ?

Yes, look up at the top of this page for the "Wiki" tab.  :)  https://github.com/Macaulay2/M2/wiki

fchapoton:
I was not able to upload the file on the wiki. I have just created

https://github.com/Macaulay2/M2/wiki/Lark-Grammar-for-M2-language

Should I copy paste the lark file in this page ?

mahrud:
Since it's currently incomplete, pasting it in a code block on the wiki where others can add to it is one way to go.

However, can I ask what's the intended purpose of this file? If it can be eventually used as code validator or linter for instance, it would be good to include scripts and instructions for how to test it on a small example in a new subdirectory of the [editors](https://github.com/Macaulay2/M2/tree/master/M2/Macaulay2/editors) directory.

Is the goal that this file would be populated and maintained by hand (which would be a daunting task, or automatically by M2?

fchapoton:
I have copy-pasted the macaulay2.lark file in the wiki.

The goal would be to use that in a colde validator, yes. I have added instructions on how to use the grammar to create a parser, see the wiki page.

If done coprrectly, and as long as the language itself does not change, there is not maintenance. But the language may be evolving ?

DanGrayson:
There was a failing test run -- maybe you could fix it by merging your branch with the master branch.

mahrud:
I think this should be closed, since it is now a wiki page: https://github.com/Macaulay2/M2/wiki/Lark-Grammar-for-M2-language

fchapoton:
yes, you can close. Sorry for not being able to push this further.

PR #2912, opened 2023-08-07
`LeftIdeal` replaces the old `Ideal`; new `Ideal` is two-sided

antonleykin:
`Ideal` is a type of `LeftIdeal` now.
* Most functions are defined for `LeftIdeal`; 
* the ones that make sense only for a two-sided ideal are defined for Ideal.
* The method `ideal` constructs an ideal typical for the given ring: e.g., 
  * `LeftIdeal` for a Weyl algebra and 
  * `Ideal` for all other algebras: e.g., polynomial ring, skew-commutative ring, `AssociativeAlgebra`, etc.

Current state:
* there seem to be _no_ errors in the build process
* there still may be methods that can migrate from `Ideal` to `LeftIdeal`; currently 
  * (clean session of M2)  `#methods Ideal ` is 223 and `# methods LeftIdeal` is 66 
  * (after `needsPackage "BernsteinSato"`) `#methods Ideal ` is 271 and `# methods LeftIdeal` is 151 

**Whoever is interested in noncommutative things, please look at this.**


mahrud:
Since the majority of Core code related to ideals is being touched, I think this is a good opportunity to move all of them into `ideals.m2`, with separation of those defined for LeftIdeal and those defined for Ideal.

antonleykin:
> Since the majority of Core code related to ideals is being touched, I think this is a good opportunity to move all of them into `ideals.m2`, with separation of those defined for LeftIdeal and those defined for Ideal.

I agree, but it may be best to keep the edits in-place for now, so that people reviewing them can do so easily and are convinced that these are non-drastic (and hold water).

mahrud:
Totally valid point. My only suggestion is that we can add a _second commit_ that does the moving. This way the first commit can be reviewed first, and the second will be only a movement so should require minimal reviewing. No strong preference.

mikestillman:
@DanGrayson This pull request (currently a draft request) consists of one commit.  However, it shows many commits by others that we have just recently merged into development, making it hard to see what has been added in this pull request.  It also contains a conflicting file now, even though the request (I think) has nothing to do with that.  Any idea how to fix this problem?

mikestillman:
@moorewf and @ggsmith This pull request will be changed to non-draft soon.  What are your thoughts on it?

d-torrance:
It looks like commits from `development` were cherry-picked onto this branch (there are lots of messages like "d-torrance authored and Anton Leykin committed 2 days ago" under the "Commits" tab) rather than merging `development` or rebasing on top of it.  Assuming that the `development` branch is up-to-date locally, `git rebase development` and then pushing this branch with the `--force` option should fix it.

antonleykin:
> It looks like commits from `development` were cherry-picked onto this branch (there are lots of messages like "d-torrance authored and Anton Leykin committed 2 days ago" under the "Commits" tab) rather than merging `development` or rebasing on top of it. Assuming that the `development` branch is up-to-date locally, `git rebase development` and then pushing this branch with the `--force` option should fix it.

Thanks, @d-torrance !
Is there an easy way to avoid these kind of cherrypicks? 
(I believe these particular ones were obtained by a combination of pulls and pushes done by `magit`.)

What is the simplest workflow to avoid this?
Should we
* `git rebase main/development`  (here: `main` is `Macaulay2` as apposed to `origin` which is `antonleykin`) and then
* `git push --force` (forcing a push every time?)

d-torrance:
> Thanks, @d-torrance ! Is there an easy way to avoid these kind of cherrypicks? (I believe these particular ones were obtained by a combination of pulls and pushes done by `magit`.)

I think just be careful when rebasing.  I'm not entirely sure what happened, but it looks like something that might have happened if you were on `development` and did `git rebase LeftModule` instead of the other way around.
 
> What is the simplest workflow to avoid this? Should we
> 
> * `git rebase main/development`  (here: `main` is `Macaulay2` as apposed to `origin` which is `antonleykin`) and then
> * `git push --force` (forcing a push every time?)

Yes, that should work!  I generally like rebasing like this instead of merging on my pull requests so we don't end up with even more merge commits than we already have.

mahrud:
Just getting around to looking into this PR. I'm wondering, wouldn't the following alternative be simpler?
* Let `Ideal` be an abstract type, meaning not necessarily two-sided
* Define methods that work for any type of ideal for `Ideal`
* Add `LeftIdeal` as a **subtype** of `Ideal` (i.e. opposite of this PR)
* Whenever a method should do something different for left ideals, define a method which either implements the specialized routine or returns an error.

Is there a problem with this approach?

pzinn:
I know this has been discussed at length (including today!), but for what it's worth, here are my 2 cents:
why do we care about left ideals anyway? I understand why we care about *two-sided ideals*: that's because they're needed to make quotient rings. But left ideals are really just left modules which happen to be sitting in the ring. There's not much more about them that's interesting. So what we should be focussing is on defining left/right/bimodules...

mikestillman:
@pzinn I agree, an interesting point.

@moorewf What do you think about not having left/right ideals, instead just having left/right modules?

moorewf:
I'm fine with that as well.  There doesn't seem to be a need to distinguish between the two.

antonleykin:
A sensible solution could be to have `LeftIdeal` to support the math habits of the $D$-modules community (a left ideal generated by a system of differential operators is as traditional as an ideal generated by polynomials), but I can see how (for instance in general associative algebras) we can get by without introducing `RightIdeal`... unless someone demands it.

pzinn:
But couldn't you define your D-modules as left *modules* rather than left ideals? after all, they are called D-*modules*...

mahrud:
You could alternatively define `LeftIdeal` to inherit from left modules, no?

mahrud:
I realized #3003 probably caused all the conflicts here, but since all of them are just renames from `Dmodules/X` to `WeylAlgebras/X`, a rebase should be able to fix it.

antonleykin:
This sequence doesn't seem to help
```
git rebase main/development 
git push --force
```
Experts on rebasing, what do you suggest?

PR #2913, opened 2023-08-11
Use gcd in is_weak_member (e/montableZZ.cpp).

0-t-t-o:
It looks like the variable "g" was not used as intended.


mikestillman:
@0-t-t-o Thanks for seeing this bug.

PR #2916, opened 2023-08-18
Update URL to Macaulay2 webpage in docs

d-torrance:
Now that https://faculty.math.illinois.edu/ doesn't exist anymore, we fix various links in the documentation, either using local links (for packages referring to other packages) or replacing with https://macaulay2.com/.

@DanGrayson -- do you have an updated URL to replace https://faculty.math.illinois.edu/~dan/?  Or should we just remove the `HomePage` from your packages?

### DIFF
--- a/M2/Macaulay2/editors/highlightjs/macaulay2.js.in
+++ b/M2/Macaulay2/editors/highlightjs/macaulay2.js.in
@@ -2,7 +2,7 @@
 Language: Macaulay2
 Author: Doug Torrance <dtorrance@piedmont.edu>
 Description: Macaulay2 is a software system devoted to supporting research in algebraic geometry and commutative algebra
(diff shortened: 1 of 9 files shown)


d-torrance:
The HTML5 validation check I wrote is failing on this and some other PR's. I'll see if I can figure out what's happening.

d-torrance:
I noticed a bunch more links that used `math.uiuc.edu` (which had apparently been forwarded to `faculty.math.illinois.edu`), so now they've been updated as well.

mikestillman:
@d-torrance This seems to fail still?  Do you see the problem?

d-torrance:
> @d-torrance This seems to fail still? Do you see the problem?

Weird! It worked locally... The error is a bit different now. I'll play around with it some more.

PR #2926, opened 2023-09-08
Add version switcher to web documentation

d-torrance:
As discussed at a recent M2internals meeting, we add a dropdown menu to the web documentation so users can navigate easily to the documentation for different versions of Macaulay2.  To check it out in action, see https://webwork.piedmont.edu/M2/Macaulay2/share/doc/Macaulay2/FirstPackage/html/index.html (`FirstPackage` and `Style` documentation only).

Note that the dropdown menu will only appear when served over a webserver (so not when using `viewHelp`, for example).  It assumes the directory structure currently used by the website (i.e., the entire `share` directory is inside a directory of the form `Macaulay2-X.Y`), and relies on the existence of a file `versions.json` in the same directory that contains the various `Macaulay2-X.Y` directories that contains the various version numbers in a JSON array, e.g., `["1.22", "1.21", "1.20"]`.  This is documented in `VERSION-README`.

### DIFF
--- a/M2/Macaulay2/m2/html.m2
+++ b/M2/Macaulay2/m2/html.m2
@@ -55,6 +55,8 @@ defaultCharset := () -> META { "http-equiv" => "Content-Type", "content" => "tex
 
 defaultHEAD = title -> HEAD splice { TITLE title, defaultCharset(), defaultStylesheet(), KaTeX(),
     SCRIPT {"src" => getStyleFile "prism.js", ""},
+    SCRIPT {"var current_version = '", version#"VERSION", "';"},
+    SCRIPT {"src" => getStyleFile "version-select.js"},
     LINK {
 	"rel" => "icon", "type" => "image/x-icon",
 	"href" => getStyleFile "icon.gif"}}
--- a/M2/Macaulay2/m2/installPackage.m2
+++ b/M2/Macaulay2/m2/installPackage.m2
@@ -269,8 +269,12 @@ buttonBar := tag -> DIV {
 	    else if tag === "toc"
 	    then HREF {htmlDirectory | tocFileName, "Table of Contents"}
 	    else TO tag)},
-    prepend("class" => "right",
-	DIV splice between_" | " {
+    DIV join({"class" => "right",
+	    LITERAL ///<form method="get" action="https://www.google.com/search">
+  <input placeholder="Search" type="text" name="q" value="">
+  <input type="hidden" name="q" value="site:macaulay2.com/doc">
+</form>///},
+	splice between_" | " {
 	    nextButton tag,
 	    prevButton tag,
 	    forwardButton tag,
--- a/M2/Macaulay2/m2/installPackage.m2
+++ b/M2/Macaulay2/m2/installPackage.m2
@@ -240,8 +240,8 @@ BACKWARD  := tag -> if PREV#?tag then BACKWARD0 PREV#tag else if UP#?tag then UP
 topNodeButton  := (htmlDirectory, topFileName)   -> HREF {htmlDirectory | topFileName,   "top" };
 indexButton    := (htmlDirectory, indexFileName) -> HREF {htmlDirectory | indexFileName, "index"};
 tocButton      := (htmlDirectory, tocFileName)   -> HREF {htmlDirectory | tocFileName,   "toc"};
-pkgButton      := TO2 {"packages provided with Macaulay2", "packages"};
-homeButton     := HREF {"http://macaulay2.com/", "Macaulay2 website"};
+pkgButton      := TO2 {"packages provided with Macaulay2", "Packages"};
+homeButton     := HREF {"https://macaulay2.com/", "Macaulay2"};
 
 nextButton     := tag -> if NEXT#?tag then HREF { htmlFilename NEXT#tag, "next" }     else "next"
 prevButton     := tag -> if PREV#?tag then HREF { htmlFilename PREV#tag, "previous" } else "previous"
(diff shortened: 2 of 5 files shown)


mahrud:
A few suggestions based on an example that has been burned into my brain:
image
1. have a "latest release" that removes the version number from the url (eventually adding git versions would be cool!)
2. is "top" basically a duplicate of the "package name"?
3. while changing that, perhaps also change from:
```
[version] | next | prev | forward | backward | up | top | index | toc | packages | Macaulay2 website
pkgname :: docnode
```
Into:
```
Macaulay2 >> Documentation [version]                   prev | next | up | index | toc
Packages >> pkgname :: docnode
```
I understand this may mismatch the format of "info" but I don't see why the website should match a decades old format.

d-torrance:
> 1. have a "latest release" that removes the version number from the url (eventually adding git versions would be cool!)

This is already happening with the `Macaulay2` directory pointing to `Macaulay2-1.22`.  (I think Dan's using a symlink on the main website?  My little demo site uses an Apache alias.)  It would be cool to mention that it's the latest release in the dropdown, though!  That should be pretty easy to add.

> 2. is "top" basically a duplicate of the "package name"?

It is.

> 3. while changing that, perhaps also change from:
> 
> ```
> [version] | next | prev | forward | backward | up | top | index | toc | packages | Macaulay2 website
> pkgname :: docnode
> ```
> 
> Into:
> 
> ```
> Macaulay2 >> Documentation [version]                   prev | next | up | index | toc
> Packages >> pkgname :: docnode
> ```
> 
> I understand this may mismatch the format of "info" but I don't see why the website should match a decades old format.

I like this.  I guess the question is whether we would change the header just for 1.23 and beyond, or retroactively change the header for the old docs, too, for a consistent look.  We'd need to rebuild the older docs for the version switcher to work across all versions, so how much would we cherry-pick from this PR?

mikestillman:
@DanGrayson,  @d-torrance has 4 remaining pull requests that pass all tests.  @d-torrance these are all ready to merge?  @DanGrayson do you want to merge them, or should I?

d-torrance:
My other pull requests are all ready.  @mahrud had some suggestions above for this one that I'd like to implement first, and I'd also welcome any other feedback about this version-switching feature for the website documentation.  I'll go ahead and mark this as a draft for now.

mahrud:
> I guess the question is whether we would change the header just for 1.23 and beyond, or retroactively change the header for the old docs, too, for a consistent look. We'd need to rebuild the older docs for the version switcher to work across all versions, so how much would we cherry-pick from this PR?

One possibility is to write it as a standalone script that parses the database files from old versions and outputs html docs; another is to do a massive regex replace!

d-torrance:
I've implemented @mahrud's suggestions for the navigation bar:
 https://webwork.piedmont.edu/M2/Macaulay2/share/doc/Macaulay2/FirstPackage/html/index.html

mahrud:
Looks pretty good! I was thinking "Documentation" should link to [Macaulay2Doc](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/). Also, which of the following do you think makes the most sense?
1. Macaulay2 >> Documentation [version picker] (current)
2. Macaulay2 [version picker] >> Documentation
3. Macaulay2 >> [version picker] Documentation (cmake)

I now think 2 might be better, but don't feel very strongly.

(ps: maybe one day we can remove forward/backward and stick to prev/next like [every](https://flintlib.org/doc/examples.html) [one](https://cmake.org/cmake/help/latest/manual/cmake.1.html) [else](https://magma.maths.usyd.edu.au/magma/overview/2/27/10/).)

d-torrance:
> Looks pretty good! I was thinking "Documentation" should link to [Macaulay2Doc](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/). Also, which of the following do you think makes the most sense?
> 
> 1. Macaulay2 >> Documentation [version picker] (current)
> 2. Macaulay2 [version picker] >> Documentation
> 3. Macaulay2 >> [version picker] Documentation (cmake)
> 
> I now think 2 might be better, but don't feel very strongly.

I like 2, too.  I'll try to change it before M2internals on Monday.
 
> (ps: maybe one day we can remove forward/backward and stick to prev/next like [every](https://flintlib.org/doc/examples.html) [one](https://cmake.org/cmake/help/latest/manual/cmake.1.html) [else](https://magma.maths.usyd.edu.au/magma/overview/2/27/10/).)

I only vaguely understand the difference between forward/backward and prev/next from reading the `Info` manual a few years ago.  Do we have any packages that actually take advantage of it?  I suppose Sphinx (which most of these other sites seem to use) may not have this capability.  Honestly, I don't really see much point to them.  Macaulay2 documentation nodes (or at least in my packages lol) aren't really written linearly.  Maybe some of the tutorial-style nodes in `Macaulay2Doc`, but for the most part it's just a bunch of pages for the various methods a package exports.

mahrud:
There are packages which take advantage of subnodes, but I agree that nobody reads the documentation linearly.

d-torrance:
A search bar was brought up at M2internals as well.  One potential issue is that there's been some discussion of telling Google (using `robots.txt`) not to index anything outside the unversioned `Macaulay2` directory.  Maybe there's a way to allow it to still index them so that our search feature would work, but prioritize the unversioned pages for folks searching from Google itself?

mahrud:
Maybe the search bar shouldn't limit the version then, just to keep things simple and leave that option open.

mahrud:
This looks great! However, the search doesn't seem to work for me! For instance, searching "hilbert series" brings me here:
https://www.google.com/search?q=hilbert+series&q=site%3Awww.macaulay2.com%2Fdoc
It's possible that this is caused by recent shift from `macaulay2.com` to `www.macaulay2.com` and google hasn't caught on yet? I'm not sure what is going on ...

d-torrance:
> This looks great! However, the search doesn't seem to work for me! For instance, searching "hilbert series" brings me here: https://www.google.com/search?q=hilbert+series&q=site%3Awww.macaulay2.com%2Fdoc It's possible that this is caused by recent shift from `macaulay2.com` to `www.macaulay2.com` and google hasn't caught on yet? I'm not sure what is going on ...

Oops!  I was wondering why I was getting so few results lol.  I dropped the `www` and it works much better now. :)

d-torrance:
The documentation for 1.21 and 1.22 with the latest changes is now live at https://webwork.piedmont.edu/M2/Macaulay2-1.22/share/doc/Macaulay2/Macaulay2Doc/html/index.html.  (`Macaulay2Doc`, `Style`, and `FirstPackage` only).

mahrud:
Looks great and works great! Thank you :)

By the way, here's a concrete approach towards these two long-term suggestions:
> 3. using a css or client-side js framework
> 4. reducing the size of the docs (currently ~140MB uncompressed, 10MB gzipped)

Using a javascript template system like [mustache.js](https://github.com/janl/mustache.js), we could have every request redirected to a single html template which uses javascript to read the URL and load an appropriate json file containing a serialized version of the documentation node and populate the template with it.

For instance, the raw documentation node for `FirstPackage` is less than 520 bytes, but the html file is 5226 bytes. Multiply that by total number of html files, which is about 10k and it becomes significant enough that with this optimization we could easily store all documentation for multiple versions on GitHub Pages for instance.

mahrud:
I just noticed the left and right don't quite seem aligned, do you see the same thing?
image
It's very minor, but more noticeable locally:
image

PR #2930, opened 2023-09-14
Store document tag filename relative to package source directory & fix locate(DocumentTag)

d-torrance:
Currently, the `filename` key for each tag in the documentation database stores the filename of the documentation source relative to whichever directory we were in when calling `installPackage`.  This filename is then used by `locate(DocumentTag)` to tell us where the documentation source is located.  But the probability that we are in the same directory as we were when the package was installed is small.

Also, in the autotools build, we run `installPackage` inside a temporary directory, but in the cmake build, we just run it in the `packages` directory, so the filenames stored in the documentation databases are different depending on the build system.  (I think based on reading the code -- I haven't actually tested a cmake build to verify this.)

So we standardize the filename stored in the documentation database by making it relative to the package source directory.  We then prepend the package source directory to this when calling `locate`.

### Before

For example, let's say I run the following in the `packages` directory:

```m2
i1 : installPackage "JSON"
 -- making example results for "JSON"                                        -- 0.806262 seconds elapsed
 -- making example results for "fromJSON"                                    -- 0.823951 seconds elapsed
 -- making example results for "toJSON"                                      -- 0.837141 seconds elapsed
 -- warning: found 3 documentation node(s) not listed as a subnode
 -- warning: missing node: nil cited by nil
 -- warning: 1 warning(s) occurred in documentation for package JSON

o1 = JSON

o1 : Package

i2 : locate makeDocumentTag "JSON"

o2 = JSON.m2:198:0

o2 : FilePosition
```

So far, so good.  But now let's exit Macaulay2, `cd` to the home directory, and then fire `M2` back up:

```m2
i1 : needsPackage "JSON"

o1 = JSON

o1 : Package

i2 : locate makeDocumentTag "JSON"

o2 = JSON.m2:198:0

o2 : FilePosition
```

We get the same output, even though `JSON.m2` is not in the home directory.  Also, we won't be able to jump to the file in Emacs.

### After

Now, after doing the same thing, we get the correct output after `cd`'ing to the home directory:

```m2
i2 : locate makeDocumentTag JSON

o2 = .Macaulay2/local/share/Macaulay2/JSON.m2:198:0

o2 : FilePosition
```

### DIFF
--- a/M2/Macaulay2/m2/document.m2
+++ b/M2/Macaulay2/m2/document.m2
@@ -654,7 +654,8 @@ document List := opts -> args -> (
 	    storeRawDocumentation(tag2, new HashTable from {
 		    PrimaryTag => tag, -- tag must be primary
 		    symbol DocumentTag => tag2,
-		    "filename" => currentFileName,
+		    "filename" => relativizeFilename(
+			currentPackage#"source directory", currentFileName),
 		    "linenum" => currentRowNumber()
 		    })));
     -- Check BaseFunction
@@ -673,7 +674,8 @@ document List := opts -> args -> (
     if #out > 0 then o.Outputs = out else remove(o, Outputs);
     if #ino > 0 then o.Options = ino else remove(o, Options);
     -- Set the location of the documentation
-    o#"filename" = currentFileName;
+    o#"filename" = relativizeFilename(
+	currentPackage#"source directory", currentFileName);
     o#"linenum"  = currentRowNumber();
     currentDocumentTag = null;
     storeRawDocumentation(tag, new HashTable from o))
@@ -689,7 +691,8 @@ undocumented Thing := key -> if key =!= null then (
     storeRawDocumentation(tag, new HashTable from {
 	    symbol DocumentTag => tag,
 	    "undocumented"     => true,
-	    "filename"         => currentFileName,
+	    "filename"         => relativizeFilename(
+		currentPackage#"source directory", currentFileName),
 	    "linenum"          => currentRowNumber()
 	    }))
 
--- a/M2/Macaulay2/m2/document.m2
+++ b/M2/Macaulay2/m2/document.m2
@@ -316,7 +316,9 @@ storeRawDocumentation := (tag, rawdoc) -> (
     fkey := format tag;
     if currentPackage#rawKey#?fkey and signalDocumentationError tag then (
 	newloc := toString new FilePosition from (
-	    minimizeFilename rawdoc#"filename", rawdoc#"linenum", 0);
+	    minimizeFilename(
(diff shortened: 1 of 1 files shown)


mahrud:
This is nice, but there's one downside to be aware of: I initially added `(locate, DocumentTag)` as a hack for quickly finding the source of a documentation node, especially in Macaulay2Doc since it's so disorganized, so that I could edit that file. I haven't tested your PR, but I think it might be more useful for `locate` to give you the location of the source, not where the documentation has been installed.

For instance in your example, editing `.Macaulay2/local/share/Macaulay2/JSON.m2` will have no effect on the documentation, even if you reinstall the package.

I understand that the source file may not be present, for instance if M2 is installed from a distributed package, but personally I don't think `(locate, DocumentTag)` is going to be used by typical uses much.

d-torrance:
As long as `pkg#"source directory` points in the right place, then it should still work for package development.  For example, if I'm messing around with `Macaulay2Doc`, I generally run `M2` with the `--srcdir` option to load everything from my clone of the git repository instead of from `/usr/share`.  Then we get the following, so the proposed change should work as expected:

```m2
i1 : (needsPackage "Macaulay2Doc")#"source directory"

o1 = /home/profzoom/src/macaulay2/M2/M2/Macaulay2/packages/
```

mahrud:
Excellent, thanks!

mahrud:
`locate makeDocumentTag` has stopped working for me. When I press enter on the output, it asks:
```
Find this error in (default (../../Macaulay2/packages/Macaulay2Doc/functions/net-doc.m2): ~/Projects/M2/M2/M2/BUILD/build
```
And regardless of what I enter the file doesn't open.

d-torrance:
Strange!

One question: Have your documentation databases been rebuilt since this commit went in?  If not, then this will indeed be broken.

If the databases are up to date, then what output do you get with the following?

```m2
debug Core
currentDirectory()
(package makeDocumentTag "net")#"source directory"
(fetchAnyRawDocumentation makeDocumentTag "net")#"filename"
locate makeDocumentTag "net"
```

mahrud:
Try opening this:
```m2
i9 : net => locate tag

o9 = net => (../../Macaulay2/packages/Macaulay2Doc/functions/net-doc.m2:73:0)
```

d-torrance:
Ah, gotcha.  Those are regex issues in `M2-comint-mode`.  Right now, it's thinking `(`, `{`, and `|` are part of the filename.  It's not specific to document tags, either.  For example, this gives me the same thing:

```m2
i1 : { locate(rank, Matrix) }

o1 = {/usr/share/Macaulay2/Core/matrix1.m2:663:19-666:20}
```

PR #2933, opened 2023-09-17
Mperlman new package: QuadraticIdealExamplesByRoos

mikestillman:
I could not figure out why PR #2871 was messed up, or how to fix it.  So I created a new branch off of development, and merged Michael Perlman's branch (which just has 2 files changed) in.  Now I am doing a new pull request here.  @DanGrayson @d-torrance @mahrud if there is a better way for me to handle this, please let me know!


mikestillman:
@mperlm3 There is an issue, mentioned by @mahrud that the rings you create in this package do not behave as expected, for example:
```
i22 : I = roosTable#10

              2
o22 = ideal (x , x*y, x*z)

o22 : Ideal of QQ[x, y, z, u]

i23 : use ring I

o23 = QQ[x, y, z, u]

o23 : PolynomialRing

i24 : x

o24 = x

o24 : Symbol
```

The problem is the symbols x,y,... are local to the package and not the global ones you expect them to be.

One way around this would be make roosTable a method that takes a ring with 4 variables as input.  Perhaps easiest is if we discuss this at AIM next week?

d-torrance:
One other option would have been a pull request to Michael's fork with the typo fix (like I did at https://github.com/mperlm3/M2/pull/1).  When merged, it would update the original pull request.

mperlm3:
I am not sure that we want to make roosTable a method that takes a ring
with four variables as input. I will think about it at AIM next week and
hopefully settle it then.

-Mike P

PR #2960, opened 2023-10-11
Overload "isMember" for ideal membership problem

d-torrance:
This lets us quickly check if a given ring element is in an ideal:

```m2
i1 : R = QQ[x,y,z]

o1 = R

o1 : PolynomialRing

i2 : I = ideal(-x^3 + y, x^2*y - z)

               3       2
o2 = ideal (- x  + y, x y - z)

o2 : Ideal of R

i3 : isMember(x*y^3 - z^2 + y^5 - z^3, I)

o3 = true

i4 : J = ideal(x*z - y, x*y + 2*z^2, y - z)

                             2
o4 = ideal (x*z - y, x*y + 2z , y - z)

o4 : Ideal of R

i5 : isMember(x^3*z - 2*y^2, J)

o5 = false
```


mahrud:
Would be nice to have a similar thing for modules and rings:
```m2
i16 : F = R^2;

i17 : isMember(F_1, F)
stdio:17:1:(3): error: no method found for applying isMember to:
                                    2
     argument 1 :  | 0 | (of class R )
                   | 1 |
                    2
     argument 2 :  R  (of class Module)

i18 : isMember(x, R)
stdio:18:1:(3): error: no method found for applying isMember to:
     argument 1 :  x (of class R)
     argument 2 :  R (of class PolynomialRing)
```
Not quote sure whether ring membership should just check if the ring of x is R or whether it should run `liftable`.

d-torrance:
Good idea!  The easy implementation would be to make `isMember(RingElement, Ring)` and `isMember(Vector, Module)` wrappers around `instance`.

> Not quote sure whether ring membership should just check if the ring of x is R or whether it should run `liftable`.

Or maybe even whether we could go the other direction with `promote`?  For example, if `R = ZZ[x]`, what should `isMember(2_R, ZZ)` and `isMember(2, R)` return?

Issue #2935, opened 2023-09-21
Incorrect use of Keywords


mahrud:
Looking at the documentation node "[packages provided with Macaulay2](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_packages_spprovided_spwith_sp__Macaulay2.html)," I believe the use of `Keywords` inside calls to `newPackage` is somewhat misunderstood in some packages:
https://github.com/Macaulay2/M2/blob/379c9634000dd87221dc353e1aaaca413f75c81c/M2/Macaulay2/packages/PolyominoIdeals.m2#L22
https://github.com/Macaulay2/M2/blob/379c9634000dd87221dc353e1aaaca413f75c81c/M2/Macaulay2/packages/MatchingFields.m2#L12

There are also a few dozen packages that are still "Uncategorized" and would be good to give some keywords.

cc: @j-rizwan @ollieclarke8787

d-torrance:
I think this issue is fixed by #2966, #3010, and e4d4c5f137915da3174d0a2f02f63ce644f555d8.

PR #2966, opened 2023-10-18
Updated keywords as per the issue #2935

j-rizwan:
The keywords are changed to list the package under the correct category.


d-torrance:
@DanGrayson  - this was merged into `master` and `development` is missing this commit.

Issue #2965, opened 2023-10-18
cmake build should get location of ffi.h from pkg-config


dimpase:
On my system I have
```
# pkg-config -cflags libffi 
-I/usr/lib64/libffi/include
```
but the build breaks with 
```
M2/M2/Macaulay2/d/ffi.d:7:18: fatal error: ffi.h: No such file or directory
    7 |         #include <ffi.h>
      |                  ^~~~~~~
```

mahrud:
My understanding is that importing dependencies via `find_package` and a `FindBlah.cmake` script is preferred to using `pkg-config`, which is very rigid and in many cases (c.f. #1227 and #1498) the root cause of hard to debug compilation errors, which is why I'd like to use `find_package` for fflas and givaro as the only two libraries currently added via `pkg-config`:
https://github.com/Macaulay2/M2/blob/379c9634000dd87221dc353e1aaaca413f75c81c/M2/cmake/check-libraries.cmake#L166-L168

If you could provide more info, I could try to help figure out why cmake isn't finding ffi correctly on your system.

dimpase:
> If you could provide more info, I could try to help figure out why cmake isn't finding ffi correctly on your system.

On my system, and on Gentoo in general, the FFI headers are not installed in a standard location known by the compiler. 
The only way to find it is by calling pkg-config, as far as I can see. In fact I had a branch with an explicit call to pkg-config (in the same manner you did it for the version of FFI - but actually cmake has a macro for exactly this, `pkg_get_variable`), but then I replaced it by the call to `pkg_` macro.

I am not at all convinced by your argument that it's better not to use this macro. While it might be harmful to blindly get the CFLAGS pkg-config provides (FFI_CFLAGS) added to your CFLAGS/CPPFLAGS (as I understand was the reason for bugs you described), I don't do this in my branch.

mahrud:
There are certainly ways to give cmake hints about non-standard locations, and that's perfectly fine.

I'll try the build on Gentoo soon and let you know. In the meantime, I don't think the gc fix needs to wait for this, but it's up to you.

dimpase:
> There are certainly ways to give cmake hints about non-standard locations, and that's perfectly fine.

Hardcoded hints aren't really a solution, IMHO.
The only reliable way is to find the headers location is via pkg-config. Indeed, libffi allows its headers to go to a user-configured directory, via `--includedir=` configure switch.

PR #2968, opened 2023-10-18
check that GC has C++ support enabled (cmake build)

### LINKED ISSUE #2964: cmake build should check for gc/gc_cpp.h
I had GMP installed system-wide without C++ interface enabled, and this ended up breaking the build. So there is a missing check.

dimpase:
This will fix #2964 (#2965 will go to another PR)

### DIFF
--- a/M2/cmake/FindBDWGC.cmake
+++ b/M2/cmake/FindBDWGC.cmake
@@ -80,8 +80,12 @@ if(NOT BDWGC_VERSION_OK)
       )
   endif()
 
+  # check that bdwg-gc has C++ support installed (it must be configured with --enable-cplusplus)
+  set(CMAKE_REQUIRED_INCLUDES "${BDWGC_INCLUDE_DIR}")
+  check_include_files(gc/gc_cpp.h BDWGC_CXX_SUPPORT LANGUAGE CXX)
+
   include(FindPackageHandleStandardArgs)
-  find_package_handle_standard_args(BDWGC DEFAULT_MSG BDWGC_INCLUDE_DIR BDWGC_LIBRARIES BDWGC_VERSION_OK)
+  find_package_handle_standard_args(BDWGC DEFAULT_MSG BDWGC_INCLUDE_DIR BDWGC_LIBRARIES BDWGC_VERSION_OK BDWGC_CXX_SUPPORT)
 
   mark_as_advanced(BDWGC_INCLUDE_DIR BDWGC_LIBRARIES)
 


mahrud:
Any ideas why a build on macOS failed?
```
FAILED: Macaulay2/c/CMakeFiles/scc1.dir/chk.c.o 
/usr/local/bin/ccache /usr/local/opt/llvm@14/bin/clang -DNDEBUG -DOM_NDEBUG -DSING_NDEBUG -I/Users/runner/work/M2/M2/M2/Macaulay2/c -O3 -DNDEBUG -std=gnu11 -isysroot /Applications/Xcode_14.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.1.sdk -mmacosx-version-min=12.7 -Wuninitialized -g -I/Users/runner/work/M2/M2/M2/BUILD/build/usr-host/include -I/Users/runner/work/M2/M2/M2/include -I/Users/runner/work/M2/M2/M2/BUILD/build/include -Wall -Wextra -Wfatal-errors -Wcast-qual -Wno-unused-parameter -Wno-attributes -Wno-unused-command-line-argument -MD -MT Macaulay2/c/CMakeFiles/scc1.dir/chk.c.o -MF Macaulay2/c/CMakeFiles/scc1.dir/chk.c.o.d -o Macaulay2/c/CMakeFiles/scc1.dir/chk.c.o -c /Users/runner/work/M2/M2/M2/Macaulay2/c/chk.c
/Users/runner/work/M2/M2/M2/Macaulay2/c/chk.c:268:10: fatal error: 'gdbm.h' file not found
#include <gdbm.h>
         ^~~~~~~~
1 error generated.
```

dimpase:
by an oversight, I also pushed a fix for #2965 here. Oops - but let's see if it flies, then we can upgrade this PR to close #2965 as well.

Anyway, I noticed on CI that [aea4484](https://github.com/Macaulay2/M2/pull/2968/commits/aea4484c19ae54ed3f1c1d93eec3dd7790d2dabe) fixed the macOS build (I don't know why, though).

dimpase:
OK, CI passes, I have to remove commented out lines in `M2/cmake/check-libraries.cmake`

M2/cmake/FindFFI.cmake
```diff
@@ -1,81 +0,0 @@
-# Attempts to discover ffi library with a linkable ffi_call function.
```
mahrud, line 1: I'd rather you didn't entirely delete this file, in case it may be useful in the future.

M2/cmake/check-libraries.cmake
```diff
@@ -80,9 +80,12 @@ if(WITH_TBB)
 endif()
 
 if(WITH_FFI)
-  find_package(FFI REQUIRED QUIET)
-  execute_process(COMMAND pkg-config --modversion libffi
-    OUTPUT_VARIABLE LIBFFI_VERSION OUTPUT_STRIP_TRAILING_WHITESPACE)
+	# find_package(FFI REQUIRED QUIET)
+	# execute_process(COMMAND pkg-config --modversion libffi
+	# OUTPUT_VARIABLE LIBFFI_VERSION OUTPUT_STRIP_TRAILING_WHITESPACE)
```
mahrud, line 85: It's okay to leave these commented, but use two spaces for indenting instead of tabs.

M2/Macaulay2/d/CMakeLists.txt
```diff
@@ -144,8 +144,9 @@ add_library(M2-interpreter OBJECT ${CLIST} ${CXXLIST} ${TAGS}
   $<$<BOOL:${WITH_XML}>:xml-c.c xml-c.h>
   $<$<BOOL:${WITH_PYTHON}>:python-c.c>)
 
+target_include_directories(M2-interpreter PUBLIC ${FFI_INCLUDE_DIRS})
```
mahrud, line 147: Why should this be public?

M2/cmake/FindFFI.cmake
dimpase: How about putting the hash of the commit, which removed it, next to the commented out `find_package(FFI...` ?

M2/Macaulay2/d/CMakeLists.txt
dimpase: I had trouble understanding cmake docs on that, which say

The INTERFACE, PUBLIC and PRIVATE keywords are required to specify the [scope](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#target-usage-requirements) of the following arguments. 
PRIVATE and PUBLIC items will populate the [INCLUDE_DIRECTORIES](https://cmake.org/cmake/help/latest/prop_tgt/INCLUDE_DIRECTORIES.html#prop_tgt:INCLUDE_DIRECTORIES) property of `target`. PUBLIC and INTERFACE items will populate the [INTERFACE_INCLUDE_DIRECTORIES](https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_INCLUDE_DIRECTORIES.html#prop_tgt:INTERFACE_INCLUDE_DIRECTORIES) property of `target`. 

Should it rather be PRIVATE?

M2/cmake/FindFFI.cmake
dimpase: I'd be surprised if it ever would need resurrection. I can imagine `libffi` getting a `cmake` interface,
then it could be used instead of `pkg_` interface.

mahrud: If your solution doesn't work on some machine, I want to be able to simply uncomment the previous lines and comment out the pkgconfig line without needing to bring back a file with git.

dimpase: OK, done

M2/cmake/check-libraries.cmake
dimpase: OK, done

M2/Macaulay2/d/CMakeLists.txt
dimpase: done

dimpase:
OK, all addressed. Had a minor git snafu due to submodules getting "dirty", but all good, I think.

M2/Macaulay2/d/CMakeLists.txt
mahrud: `PUBLIC` means if another component depends on the current component (`M2-interpreter` in this case), they will also have these compiler flags passed onto them, `PRIVATE` means they don't.

mahrud:
For reasons I explained [here](https://github.com/Macaulay2/M2/issues/2965#issuecomment-1772105011), I think this PR should go back to only fixing the gc issue. I don't think increasing dependency on pkg-config is an improvement.

M2/cmake/FindFFI.cmake
```diff
@@ -1,3 +1,4 @@
+################### This file is made obsolete by the PR 2968 #######################
```
mahrud, line 1: This is an open source script used by [other](https://github.com/llvm/llvm-project/blob/main/llvm/cmake/modules/FindFFI.cmake) projects. I don't see a point in marking it obsolete ...

dimpase: This file is made obsolete for M2, it does not mean it's obsolete in general (although in fact it probably is :-))

dimpase:
> For reasons I explained [here](https://github.com/Macaulay2/M2/issues/2965#issuecomment-1772105011), I think this PR should go back to only fixing the gc issue. I don't think increasing dependency on pkg-config is an improvement.

I think your fear of pkg-config is unfounded. M2 was shooting itself in the foot by blindly adding CFLAGS emitted by pkg-config to its CFLAGS, and this branch does not do this.

M2/cmake/FindFFI.cmake
dimpase: By the way, LLVM has updated that file few days ago, so M2 copy is obsolete, regardless of this PR :-)

mahrud:
> M2 was shooting itself in the foot by blindly adding CFLAGS emitted by pkg-config to its CFLAGS, and this branch does not do this.

How so? The problem with fflas was that when we called `target_link_libraries(M2-interpreter PUBLIC PkgConfig::FFLAS_FFPACK)`, it would also cause the appropriate `cflags`, including `-fabi-version=6`, to be appended to the compiler options. This is exactly what you do with `PkgConfig::FFI`, no?

mahrud:
There are many differences between autotools and cmake patterns, and if it's a choice between a cached `find_package` call vs multiple lines extracting flags by external calls to `pkg-config`, the former is strictly preferred. In particular, I think https://github.com/Macaulay2/M2/commit/d00d4b03b7126812dd13c49c5e8460d49cf343c9 should have extracted the version from the header file so that it is cached rather than a call `pkg-config`. In fact, this is strictly an anti-pattern since `execute_process` is run in _every call_ to cmake, but `find_package` only runs if the package hasn't been found.

dimpase:
> There are many differences between autotools and cmake patterns, and if it's a choice between a cached `find_package` call vs multiple lines extracting flags by external calls to `pkg-config`, the former is strictly preferred. In particular, I think [d00d4b0](https://github.com/Macaulay2/M2/commit/d00d4b03b7126812dd13c49c5e8460d49cf343c9) should have extracted the version from the header file so that it is cached rather than a call `pkg-config`. 

You still can't avoid a call to pkg-config (via a `pkg_` macro, or not), as you don't know whether the headers are on the default compiler path.


> In fact, this is strictly an anti-pattern since `execute_process` is run in _every call_ to cmake, but `find_package` only runs if the package hasn't been found.

that's why calling a `pkg_` macro might be preferable to explicitly calling pkg-config.

(ditto for autotools build - there are autoconf macros such as `PKG_CHECK_MODULES` to do the job right)

dimpase:
> > M2 was shooting itself in the foot by blindly adding CFLAGS emitted by pkg-config to its CFLAGS, and this branch does not do this.
> 
> How so? 
I was referring to #1227, where that was exactly the case.

> The problem with fflas was that when we called `target_link_libraries(M2-interpreter PUBLIC PkgConfig::FFLAS_FFPACK)`, it would also cause the appropriate `cflags`, including `-fabi-version=6`, to be appended to the compiler options. This is exactly what you do with `PkgConfig::FFI`, no?

I don't use `PUBLIC` there, but, anyhow, rather unexpected by me, `target_link_libraries` does do stuff to compiler options, and you are right, it does take CFLAGS from pkg-config.

Although then `target_include_directories(M2-interpreter PRIVATE ${FFI_INCLUDE_DIRS})` in my branch is not needed,
as its job is done by `target_link_libraries`.

So this call can be removed from this branch.
I still don't worry about CFLAGS in this case, as libffi is a pure C library, it doesn't need to care about C++ ABIs etc;
and its only CFLAGS are include dirs, cf. https://github.com/libffi/libffi/blob/master/libffi.pc.in

mahrud:
> that's why calling a `pkg_` macro might be preferable to explicitly calling pkg-config.

I dislike calls to the `pkg_` macros because it's harder to unset the cached variables set by them when debugging. With `find_package` I can simply run `cmake -U*FFI* .`, but the `pkg_` macros have hidden variables that are harder to erase.

Anyway, this is my opinion based on experience debugging cmake issues. I understand that you disagree, and I'll help you find a solution to your issue on Gentoo, but I'll say no to this PR for now so I can get back to my job applications.

mahrud (changes requested):
The gc changes are all good, but the ffi changes are under discussion.

dimpase:
> The gc changes are all good, but the ffi changes are under discussion.

OK, I just took out the FFI stuff, will put it on another PR

mahrud (approved):

PR #2970, opened 2023-10-20
set the path for ffi headers

### LINKED ISSUE #2965: cmake build should get location of ffi.h from pkg-config

dimpase:
Fixes #2965 

Split off from #2968 (see it for the discussion)

### DIFF
--- a/M2/cmake/FindFFI.cmake
+++ b/M2/cmake/FindFFI.cmake
@@ -1,3 +1,4 @@
+################### This file is made obsolete by the PR 2968 #######################
 # Attempts to discover ffi library with a linkable ffi_call function.
 #
 # Example usage:
--- a/M2/cmake/FindFFI.cmake
+++ b/M2/cmake/FindFFI.cmake
@@ -1,4 +1,6 @@
 ################### This file is made obsolete by the PR 2968 #######################
+# This is public domain code
+#
 # Attempts to discover ffi library with a linkable ffi_call function.
 #
 # Example usage:
--- a/M2/cmake/check-libraries.cmake
+++ b/M2/cmake/check-libraries.cmake
@@ -80,9 +80,11 @@ if(WITH_TBB)
 endif()
 
 if(WITH_FFI)
-  find_package(FFI REQUIRED QUIET)
-  execute_process(COMMAND pkg-config --modversion libffi
-    OUTPUT_VARIABLE LIBFFI_VERSION OUTPUT_STRIP_TRAILING_WHITESPACE)
+  # find_package(FFI REQUIRED QUIET) # obsoleted by PR 2689
+  # execute_process(COMMAND pkg-config --modversion libffi
+  #         OUTPUT_VARIABLE LIBFFI_VERSION OUTPUT_STRIP_TRAILING_WHITESPACE)
+  pkg_check_modules(FFI REQUIRED IMPORTED_TARGET libffi)
+  set(LIBFFI_VERSION ${FFI_VERSION})
 else()
   set(LIBFFI_VERSION "not present")
 endif()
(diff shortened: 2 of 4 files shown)


dimpase:
@mahrud @d-torrance

M2/cmake/FindFFI.cmake
```diff
@@ -1,3 +1,6 @@
+################### This file is made obsolete by the PR 2968 #######################
+# This is public domain code
```
d-torrance, line 2: Are we sure this is true?  My guess is that it has an Apache license since it's included in the LLVM source (see https://github.com/llvm/llvm-project/blob/main/llvm/cmake/modules/FindFFI.cmake)

dimpase: not %100 sure, no. It's good to dig it up

mahrud: I forgot where I saw this, but CMake FindModules are always under the lightest license (after all, it's barely code ... and who would want to copyright code for _findin_ something open source)

dimpase:
OK, so here is a minimal change, just one line (calling `pkg_get_variable` - I suppose this macro has no side effects/hidden variables)

mahrud:
Sure. I think using it as a backup (as long as it doesn't give an error if pkg-config doesn't find ffi) makes a lot of sense.

Two requests:
1. There seem to be too many unrelated commits, including a revert commit. Could you prune those?

mahrud:
Sure. I think using it as a backup (as long as it doesn't give an error if pkg-config doesn't find ffi) makes a lot of sense.

Two requests:
1. There seem to be too many unrelated commits, including a revert commit. Could you prune those?
2. Could yo also rewrite the `execute_process` line using `pkg_get_variable`, so we don't call `pkg-config` every time cmake runs?

dimpase:
1) I can squash, or whoever does the PR merging can use GitHub's option to squash PR into one commit.

2) no, unfortunately pkg-config deals with versions without variables, so this pkg_ command won't work here.

dimpase:
> 2. no, unfortunately pkg-config deals with versions without variables, so this pkg_ command won't work here.

OTOH it's possible to condition calling pkg-config on the value of a CACHE variable, so that
it's only called once.

dimpase:
OK, so this appears to work.
After the 1st run of cmake I see
```
$ grep LIBFFI CMakeCache.txt 
LIBFFI_VERSION:STRING=3.4.4
```
And if I manually edit `CMakeCache.txt` to change it so that, say,
``` 
LIBFFI_VERSION:STRING=3.4.42
```
then this value does not get overwritten on the subsequent run of cmake.

M2/cmake/check-libraries.cmake
```diff
@@ -82,8 +82,11 @@ endif()
 if(WITH_FFI)
   pkg_get_variable(FFI_INCLUDE_DIR libffi includedir)
   find_package(FFI REQUIRED QUIET)
-  execute_process(COMMAND pkg-config --modversion libffi
-    OUTPUT_VARIABLE LIBFFI_VERSION OUTPUT_STRIP_TRAILING_WHITESPACE)
+  if(NOT DEFINED LIBFFI_VERSION)
+    execute_process(COMMAND pkg-config --modversion libffi
+      OUTPUT_VARIABLE LIBFFI_VERSION_ OUTPUT_STRIP_TRAILING_WHITESPACE)
+    set(LIBFFI_VERSION ${LIBFFI_VERSION_} CACHE STRING "FFI version")
+  endif()
```
mahrud, line 89: I think it would be much cleaner if this code block was inside `FindFFI.cmake`. FindModules are usually supposed to set the version variable anyway, so this would be an improvement to the current one.

mahrud:
Give #2974 a try. If it solves your problem, I think this can be closed.

mahrud:
#2974 passes all tests now.

PR #2971, opened 2023-10-21
Add one-argument hilbertFunction method that returns a function

d-torrance:
We add `hilbertFunction(Module)` (and `hilbertFunction(Ideal)`/`hilbertFunction(Ring)`) that returns a ~~memoized~~ function that takes an integer (or list) as an argument.

For example:

```m2
i1 : R = QQ[x,y,z,w];

i2 : I = monomialCurveIdeal(R, {1,2,3})

             2                    2
o2 = ideal (z  - y*w, y*z - x*w, y  - x*z)

o2 : Ideal of R

i3 : h = hilbertFunction I

o3 = h

o3 : FunctionClosure

i4 : apply(10, h)

o4 = {1, 4, 7, 10, 13, 16, 19, 22, 25, 28}

o4 : List
```

### DIFF
--- a/M2/Macaulay2/m2/hilbert.m2
+++ b/M2/Macaulay2/m2/hilbert.m2
@@ -381,3 +381,7 @@ addHook((hilbertFunction, List, Module), Strategy => Default, (L, M) -> (
     f := hilbertSeries(M, Order => 1 + sum(h, L, times));
     U := monoid ring f;
     coefficient(U_L, f)))
+
+hilbertFunction Ring   :=
+hilbertFunction Ideal  :=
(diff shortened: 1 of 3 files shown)


mahrud:
Please don't use `memoize` in Core, it breaks garbage collection (c.f. the discussion in https://github.com/Macaulay2/M2/pull/2955).

d-torrance:
> Please don't use `memoize` in Core, it breaks garbage collection (c.f. the discussion in #2955). You can just return the function and have the user cache it if they'd like, not to mention that caching Hilbert function only saves very small time since `hilbertSeries` is cached!

Done!

PR #2978, opened 2023-11-05
Don't export "programPaths" with "exportMutable"

d-torrance:
It's okay for a user to modify [`programPaths`](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_program__Paths.html), but they shouldn't actually reassign it to something else.

Thanks @mahrud for pointing this out!


DanGrayson:
Won't this prevent the user from modifying the variable `programPaths`?

d-torrance:
> Won't this prevent the user from modifying the variable `programPaths`?

Yes, that's the idea.  It *doesn't* prevent them from modifying the mutable hash table that it is bound to.

Here's the current behavior, which isn't great:

```m2
i1 : programPaths#"normaliz" = "/my/unusual/normaliz/installation/dir"

o1 = /my/unusual/normaliz/installation/dir

i2 : programPaths = 5

o2 = 5

i3 : findProgram "normaliz"
stdio:3:1:(3): error: expected argument 1 to be null, a sequence, list, string, net, hashtable, database, or dictionary
```

And the proposed behavior:

```m2
i1 : programPaths#"normaliz" = "/my/unusual/normaliz/installation/dir"

o1 = /my/unusual/normaliz/installation/dir

i2 : programPaths = 5
stdio:2:14:(3): error: assignment to protected global variable 'programPaths', originally defined at M2/Macaulay2/m2/programs.m2:5:1:(1):

i3 : findProgram "normaliz"

o3 = normaliz

o3 : Program
```

d-torrance:
Also, note that there's a test verifying that `programPaths` works as intended, and it passed:

https://github.com/Macaulay2/M2/blob/86d6c70f8746e6b13d95c53ad20d880e0d60a8dc/M2/Macaulay2/tests/normal/programs.m2#L1-L5

PR #2979, opened 2023-11-05
Update Python package to support Python 3.12

d-torrance:
From [gmpy2](https://github.com/aleaxit/gmpy) 2.2.0a1

---

This should fix the compilation errors we're getting with Python 3.12 ([like this one](https://github.com/Macaulay2/M2/actions/runs/6753249299/job/18359521379)).  I've been getting strange segfaults with Python 3.12 in Debian unstable, but I'm curious what happens with the GitHub builds...


d-torrance:
Converting this to a draft until I figure out the segfault issue.

d-torrance:
I think what was happening is that with Python 3.12, some various thread-related things that Macaulay2 was doing was monkeying around with Python's thread-related things that were set up with `Py_Initialize`.  So we move this from M2 startup to after the `Python` package is loaded, and now no more segfaults!  This makes sense anyway -- what's the point of initializing Python unless you're actually going to use it?

PR #2986, opened 2023-11-07
Use Debian codename for ISSUE_RELEASE if no version number

d-torrance:
This is an issue in Debian testing and unstable, where `lsb_release -s --release` returns the uninformative "n/a".  This causes an extra problem because we end up with a little "a" subdirectory under `usr-dist` during build.

So for example, now we get "Debian-trixie" as the issue.

### DIFF
--- a/M2/configure.ac
+++ b/M2/configure.ac
@@ -96,6 +96,9 @@ then if test -f /usr/bin/sw_vers
      elif test -f /usr/bin/lsb_release
      then ISSUE_FLAVOR=`lsb_release -s --id`
           ISSUE_RELEASE=`lsb_release -s --release`
+	  if test $ISSUE_RELEASE = n/a
+	  then ISSUE_RELEASE=`lsb_release -s --codename`
+	  fi
      elif test -f /etc/os-release
      then ISSUE_FLAVOR=`. /etc/os-release ; echo $ID`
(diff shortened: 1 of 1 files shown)


mahrud:
By the way, this would be the corresponding fix in cmake:
```patch
diff --git a/M2/cmake/flavor.cmake b/M2/cmake/flavor.cmake
index 28eab2894e..8201fab4d8 100644
--- a/M2/cmake/flavor.cmake
+++ b/M2/cmake/flavor.cmake
@@ -26,6 +26,11 @@ if(NOT DEFINED ISSUE)
     execute_process(COMMAND ${LSB_RELEASE} -s --release
       OUTPUT_VARIABLE ISSUE_RELEASE
       ERROR_QUIET OUTPUT_STRIP_TRAILING_WHITESPACE)
+    if(ISSUE_RELEASE STREQUAL "n/a")
+      execute_process(COMMAND ${LSB_RELEASE} -s --codename
+       OUTPUT_VARIABLE ISSUE_RELEASE
+       ERROR_QUIET OUTPUT_STRIP_TRAILING_WHITESPACE)
+    endif()
   elseif(OS_RELEASE)
     file(READ ${OS_RELEASE} _os_release)
     string(REGEX MATCH "ID=([A-Za-z ]*)\n"      _ "${_os_release}")
```

PR #2989, opened 2023-11-09
Implement a workaround for incompatability between libc++ and mpreal

### LINKED ISSUE #2860: eigen compilation fails with latest llvm
If I use latest clang via llvm in homebrew, the compilation of eigen.cpp fails.  This has been noted before, I'd like to fix this.  It is possible that this issue already exists, but I didn't see it explicitly.

The problem appears to be possibly a mismatch of mpreal, eigen, std::complex on this latest version of clang++.  However, it works on apple clang.

Maybe we could use something other than eigen? Or maybe there is a more up to date mpreal? (The one we use is simply checked in to the e directory) @antonleykin @jkyang92 @d-torrance @mahrud Have any of you looked at this problem?

jkyang92:
This implements versions of various functions for `complex<Real>` to work around incompatibility between libc++ and mpreal. This should fix #2860. I've tested that it compiles on linux with clang and libc++, but I'd appreciate someone on an affected Mac testing it as well.

I still need to make sure that all of these handle `NaN` and `inf` correctly, and I suspect they currently don't.

Also, these are the naive algorithms for these operations, so I might have to go and figure out what libc++ does and see if it's worth copying.

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -45,9 +45,9 @@ jobs:
         include:
           - build-system: cmake
             os: ubuntu-22.04
-            compiler: clang-14
-            cxx: clang++-14
-            cc: clang-14
+            compiler: clang-15
+            cxx: clang++-15
+            cc: clang-15
           # This build tests Clang rather than AppleClang (keep)
           - build-system: cmake
             os: macos-12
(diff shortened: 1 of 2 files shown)


mahrud:
I think there's a `;` missing, but this would switch to compiling with clang-15:
[switch-to-clang-15.patch.txt](https://github.com/Macaulay2/M2/files/13303568/switch-to-clang-15.patch.txt)

mahrud:
This seems to work for at least two people using clang-15 on macOS 14 using brew.

mikestillman:
This works for me. @DanGrayson @jkyang92, I'm going to merge it.

mikestillman:
@jkyang92 Thanks for making these changes, it was frustrating not being able to use the latest clang!

Issue #2951, opened 2023-10-04
missing prompt after read()


pzinn:
another minor issue (which I intend to fix): `read()` incorrectly suppresses the input prompt:
```
i1 : read()
     BLAH BLAH

o1 = BLAH BLAH

"missing i2"

o2 = missing i2
```
the issue does not arise with say `read "please type something"`

PR #2991, opened 2023-11-10
fix #2950 + minor texMath

### LINKED ISSUE #2950: texMath MultigradedBettiTally
a minor issue, but the TeX formatting of `MultigradedBettiTally` is incorrect

pzinn:
I fixed #2950, though I'm still not sure whether `compactMatrixForm` is used appropriately, as mentioned in the comments.
I also threw in a bunch of small fixes, including an expression bug I introduced when removing `NonAssociativeProduct` (try `expression 0 * x`) and a double superscript bug.

### DIFF
--- a/M2/Macaulay2/d/actors4.d
+++ b/M2/Macaulay2/d/actors4.d
@@ -854,22 +854,24 @@ readE(f:file):Expr := (
      is e:errmsg do buildErrorPacket(e.message)
      is s:stringCell do Expr(s));
 
+readIO(msg:string):Expr := (
+     readprompt = msg;
+     oldprompt := stdIO.prompt;
+     stdIO.prompt = readpromptfun;
+     r := getLine(stdIO);
+     stdIO.prompt = oldprompt;
+     when r
+     is e:errmsg do buildErrorPacket(e.message)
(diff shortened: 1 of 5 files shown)


pzinn:
oh yeah, I pushed a fix of #2951 too. the olf, buggy, behaviour can still be reproduced with `read stdio`.

pzinn:
don't merge yet, additional problem detected with `multigraded`, cf https://github.com/Macaulay2/M2/issues/2950. needs fixing.

DanGrayson:
It's up to the ultimate user to set `compactMatrixForm`, which controls how matrices are printed.  The default value is `true`.  Are you setting it somewhere?

pzinn:
this PR has nothing to do with `compactMatrixForm` whatsoever. I was pointing out in comments of https://github.com/Macaulay2/M2/issues/2950 that I suspected that the output of `MultigradedBettiTally` with/without `compactMatrixForm` was switched. Ultimately, it's a matter of taste, and I don't particularly care, just thought whoever wrote this code might want to check.

mahrud:
It wasn't switched.

PR #3003, opened 2023-11-19
Dmodules update, new WeylAlgebras package

### LINKED ISSUE #2861: Dmodules package update announcement
We need to better document the split of Dmodules into three packages (Dmodules, HolonomicSystems, BernsteinSato). 

Perhaps also: make Dmodules into BasicDmodules and then have Dmodules load all three packages that came out of the split. 

Finally: maybe a better name for BernsteinSato? (Mike's suggestion)

### LINKED ISSUE #2901: Breaking changes in D-modules, e.g. RatAnn is missing or renamed?
@mahrud @christineBerkesch The breaking changes in the D-modules package make it very hard for people using the previous D-modules package to use 1.22.  Is there a way we can fix this?

The first problem is that it is hard to determine which packages now one needs (BernsteinSato, HolonomicSystems).  This makes it hard for people who have been using the D-modules package in the past, in particular.  The documentation should mention this, at the least.  I know we discussed that in Minneapolis, but that was after we froze the release (I think).

A second issue is that some functions (or at least one) appear to be gone, or perhaps renamed.  I was just asked about one, `RatAnn`, this person was wanting an extension to this function (they are using probably 1.21), but I could not find the function. The changes file doesn't mention that it was removed or renamed.  I know it was a poorly named function, but I'm not sure where it is now, or what it is named.  Perhaps other functions got moved or renamed too?

mahrud:
- renamed Dmodules => WeylAlgebras (Dmodules is now an umbrella package)
- added note to change log about the Dmodules package group

Closes #2861, closes #2901.


DanGrayson:
@mahrud --- can you fix the failing error?

mahrud:
It's not related to this PR.

Issue #3000, opened 2023-11-16
latest llvm compile issues


mikestillman:
@jkyang92 I might have accepted that pull request too soon (#2989).  The part below is failing under llvm 17.0.5.  I *thought* it worked under 17.0.4...  But now I'm not sure about that.

Under cmake, I get the following error, when doing `ninja` in the build tree in a cmake build. (M2 builds fine now, which is good).  

@mahrud Why are mathicgb unit tests being compiled when I do `ninja` (with no target name)? 

```
[27/548] Building CXX object Macaulay2/e/mathicgb/CMakeFiles/mathicgb-unit-tests.dir/test/Range.cpp.o
FAILED: Macaulay2/e/mathicgb/CMakeFiles/mathicgb-unit-tests.dir/test/Range.cpp.o 
[...]

PR #3005, opened 2023-11-20
Update mathicgb to fix compilation issues with clang and libc++

jkyang92:
See https://github.com/Macaulay2/mathicgb/pull/43, Should address #3000.

--
Jay Yang


DanGrayson:
@jkyang92 -- can you fix the error?

jkyang92:
@DanGrayson That error seems to have nothing to do with the patch, I have no idea why it's having a hard time with brew and libffi

DanGrayson:
Okay, I'll rerun the failed job.

PR #3008, opened 2023-11-21
Degree and Monoid changes in engine

mahrud:
- simplified RingElement::multi_degree
- improved IM2_FreeModule_get_degrees
- improved IM2_RingElement_multidegree, RingElement::multi_degree, added RingElement::degree
- changed degree weights to be std::vector (gc_vector?)
- renamed IM2_Monoid_make to rawMonoid
- documented Monoid.hpp, changed order and type of rawMonoid inputs
- moved monoid and monomial orderings out of interface.dd
- improved interface functions in monoid.hpp, added to_degree_vector
- inlined Ring::degree(const ring_elem), removed the rest
- removed unused MonomialIdeal::degree_monoid() method
- reduced calls to get_ring() and degree_monoid()


mahrud:
@mikestillman it's possible that weight vectors should be `gc_vector`s instead. I don't recall if we discussed this or not.

mikestillman:
@mahrud  I'll take a look.

M2/Macaulay2/e/NCAlgebras/FreeAlgebraQuotient.hpp
```diff
@@ -78,7 +78,13 @@ class FreeAlgebraQuotient : public our_new_delete
                      bool p_parens) const;
 
   bool is_homogeneous(const Poly& f) const;
-  void degree(const Poly& f, monomial d) const;
+  // FIXME: copied from ring.hpp because this doesn't inherit from Ring
```
mahrud, line 81: @mikestillman @moorewf Here (and in `FreeAlgebra.hpp`) I had to duplicate an inline function from `ring.hpp` because `FreeAlgebra` and `FreeAlgebraQuotient` don't inherit from each other or from `Ring`. Is there a reason for this?

M2/Macaulay2/e/monideal.hpp
```diff
@@ -124,11 +124,7 @@ class MonomialIdeal : public EngineObject
   void text_out(buffer &o) const;
 
   const PolynomialRing *get_ring() const { return R; }
-  const Monoid *degree_monoid() const
-  {
-    const Ring *S = R;
-    return S->degree_monoid();
-  }
+  const Monoid *degree_monoid() const { return R->degree_monoid(); }
```
mahrud, line 127: Looking at this again, I'm not sure why `MonomialIdeal` needs a `degree_monoid()` method. If it's not used, can I remove it?

mahrud:
I also would like to replace `VECTOR(T)` with `gc_vector<T>` everywhere, but I held off on it to avoid causing unnecessary conflicts in other branches.
https://github.com/Macaulay2/M2/blob/a19008a9446bbd36bb45869507f3a36a79c5579f/M2/Macaulay2/e/newdelete.hpp#L34-L35

mikestillman (changes requested):
@mahrud Thanks for cleaning up this code!  (We still need to add in the newer MonomialOrder code, I'll work on that).

There are a few comments I have, some are just comments, some are possible changes.

M2/Macaulay2/e/interface/monoid.h
```diff
@@ -20,30 +22,37 @@ typedef struct MonomialOrdering MonomialOrdering;
...
+const Monoid * /* or Null */ rawMonoid(const MonomialOrdering *mo,
+                                       const Ring *deg_ring,
                                        M2_ArrayString names,
-                                       const Ring *DegreeRing,
-                                       M2_arrayint degs,
-                                       M2_arrayint hefts);
-/* drg: connected rawMonoid*/
+                                       M2_arrayint degrees,
+                                       M2_arrayint heftvec);
+// TODO: this is wrong and needs to be updated
```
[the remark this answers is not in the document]
mahrud: I think the description is wrong, but I don't remember exactly what was wrong.

mikestillman:
@mahrud I was thinking about the std::vector, gc_vector classes.  Unfortunately, using std::vector<int> (and similar) as fields in the Monoid class lead to memory leaks: the backing memory for the std::vector is allocated using new, and unless we finalize all Monoid objects, the garbage collector will not be able to free that backing memory when it decides to collect the memory later.

There are two solutions that I see: 

1. Make sure all Monoid, Ring objects are finalized (everything that is garbage collected and contains such objects).
2. Use gc_vector for all elements inside the Monoid (and other garbage collected objects, such as Ring, PolynomialRing, etc) classes.  

Unfortunately, std::vector and gc_vector are different classes, so cannot be used interchangeably currently (although we could get that to work with a bit of effort).

mikestillman:
@mahrud Do you understand why the tests failed here?  It isn't related to this pull request at all, I think.

mahrud:
How do we do this exactly?
> * Make sure all Monoid, Ring objects are finalized (everything that is garbage collected and contains such objects).

mikestillman:
@mahrud I have a fix for the finalization of monoids and rings.  Actually, monoids are already being finalized.  I would like to push a quick fix to polynomial rings being finalized.  Can I just push to your branch?  Or do I need to so something first, due to rebasing?

Here is the diff:  Really, it is just commenting out a function body, and another function that is no longer needed.
```
diff --git a/M2/Macaulay2/e/finalize.cpp b/M2/Macaulay2/e/finalize.cpp
index 661a0f422..5d394298d 100644
--- a/M2/Macaulay2/e/finalize.cpp
+++ b/M2/Macaulay2/e/finalize.cpp
@@ -72,30 +72,33 @@ void intern_monideal(MonomialIdeal *G)
             (void *)G);
 }
 //////////////////////////////////////////////////////
-extern "C" void remove_polyring(void *p, void *cd)
-{
-#ifdef MEMDEBUG
-  p = M2_debug_to_inner(p);
-#endif
-  PolynomialRing *G = static_cast<PolynomialRing *>(p);
-  AO_t nremoved = AO_fetch_and_add1(&polyrings_nremoved);
-  if (M2_gbTrace >= 3)
-    fprintf(stderr, "\n -- removing polynomial ring %zd at %p\n", nremoved, G);
-  G->clear();
-}
+// extern "C" void remove_polyring(void *p, void *cd)
+// {
+// #ifdef MEMDEBUG
+//   p = M2_debug_to_inner(p);
+// #endif
+//   PolynomialRing *G = static_cast<PolynomialRing *>(p);
+//   AO_t nremoved = AO_fetch_and_add1(&polyrings_nremoved);
+//   if (M2_gbTrace >= 3)
+//     fprintf(stderr, "\n -- removing polynomial ring %zd at %p\n", nremoved, G);
+//   G->clear();
+// }
 void intern_polyring(const PolynomialRing *G)
 {
-#ifdef MEMDEBUG
-  GC_REGISTER_FINALIZER_IGNORE_SELF(M2_debug_to_outer(const_cast<PolynomialRing *>(G)), remove_polyring, 0, 0, 0);
-#else
-  GC_REGISTER_FINALIZER_IGNORE_SELF(                  const_cast<PolynomialRing *>(G) , remove_polyring, 0, 0, 0);
-#endif
-  AO_t nfinalized = AO_fetch_and_add1(&polyrings_nfinalized);
-  if (M2_gbTrace >= 3)
-    fprintf(stderr,
-            "\n   -- registering polynomial ring %zd at %p\n",
-            nfinalized,
-            (const void *)G);
+  // We are already setting a finalizer for rings in newdelete.hpp,
+  // I believe this one is just unsetting that one.
+  return;
+// #ifdef MEMDEBUG
+//   GC_REGISTER_FINALIZER_IGNORE_SELF(M2_debug_to_outer(const_cast<PolynomialRing *>(G)), remove_polyring, 0, 0, 0);
+// #else
+//   GC_REGISTER_FINALIZER_IGNORE_SELF(                  const_cast<PolynomialRing *>(G) , remove_polyring, 0, 0, 0);
+// #endif
+//   AO_t nfinalized = AO_fetch_and_add1(&polyrings_nfinalized);
+//   if (M2_gbTrace >= 3)
+//     fprintf(stderr,
+//             "\n   -- registering polynomial ring %zd at %p\n",
+//             nfinalized,
+//             (const void *)G);
 }
 //////////////////////////////////////////////////////
 extern "C" void remove_gb(void *p, void *cd)
```

mikestillman:
@mahrud Let's hope the tests run to completion this time!

mikestillman:
All checks passed! I am clicking on "rebase and merge"

Issue #3012, opened 2023-11-26
Addition of matrices over modules with torsion


d-torrance:
Suppose we want to represent the Klein 4-group as a `ZZ`-module.  Adding an element (represented as a matrix) to itself should yield 0, but it doesn't currently.  Multiplication by 2 works as expected, though.

```m2
i1 : K = comodule ideal 2 ++ comodule ideal 2

o1 = cokernel | 2 0 |
              | 0 2 |

                              2
o1 : ZZ-module, quotient of ZZ

i2 : A = map(K, ZZ^1, matrix {{1}, {0}})

o2 = | 1 |
     | 0 |

o2 : Matrix

i3 : A + A

o3 = | 2 |
     | 0 |

o3 : Matrix

i4 : 2*A

o4 = 0

o4 : Matrix
```

Issue #2898, opened 2023-06-29
Slowdown in map(Module,Module,Matrix) on v1.22


mahrud:
Comparing `map(Module,Module,RawMatrix)`:
https://github.com/Macaulay2/M2/blob/5572c8a245d187f6c22f8eaa0b03cac0b8d86e02/M2/Macaulay2/m2/matrix.m2#L19-L31
and `map(Module,Module,Matrix)`:
https://github.com/Macaulay2/M2/blob/5572c8a245d187f6c22f8eaa0b03cac0b8d86e02/M2/Macaulay2/m2/matrix1.m2#L76-L87

Note that the first one only calls `rawMatrixRemake2(...)` but the second calls `reduce(M, rawMatrixRemake2(...))`.

I ran into an example of a 96x1 matrix `B` where `B_{0}` (yes, the whole matrix) was instantaneous in v1.21 but taking a full minute in v1.22 because of this extra `reduce`.

My question is:
1. Why are the two methods above different? Can we not just use this?
```m2
map(Module,Module,Matrix) := Matrix => o -> (M,N,f) -> map(M, N, raw f, o)
```
2. If they are actually different, can we safely remove the `reduce`?

DanGrayson:
The reduce is present so that equality comparisons of matrices will work.  We have to add it to the first method, also.

mahrud:
Perhaps == should call reduce. Why should changing the target of a module induce a groebner basis computation?

PR #3013, opened 2023-11-27
Use "reduce" when adding/subtracting/negating matrices

d-torrance:
This gives us the correct answer when doing arithmetic with torsion elements of modules.

In Zulip, @mahrud suggested that there might be a better solution than using `reduce`, but using it is consistent with some other matrix operations (like `RingElement * Matrix` and `Matrix * Matrix`).

### Before
```m2
i1 : M = comodule ideal 2

o1 = cokernel | 2 |

                              1
o1 : ZZ-module, quotient of ZZ

i2 : A = map(M, ZZ^1, 1)

o2 = | 1 |

o2 : Matrix

i3 : A + A

o3 = | 2 |

o3 : Matrix

i4 : -A

o4 = | -1 |

o4 : Matrix
```

### After
```m2
i1 : M = comodule ideal 2

o1 = cokernel | 2 |

                              1
o1 : ZZ-module, quotient of ZZ

i2 : A = map(M, ZZ^1, 1)

o2 = | 1 |

o2 : Matrix

i3 : A + A

o3 = 0

o3 : Matrix

i4 : -A

o4 = | 1 |

o4 : Matrix
```

Closes: #3012

### DIFF
--- a/M2/Macaulay2/m2/matrix.m2
+++ b/M2/Macaulay2/m2/matrix.m2
@@ -119,7 +119,7 @@ Matrix == RingElement := (m,f) -> m - f == 0		    -- slow!
 Matrix == ZZ := (m,i) -> if i === 0 then rawIsZero m.RawMatrix else m - i == 0
 
 Matrix + Matrix := Matrix => (
-     (f,g) -> map(target f, source f, f.RawMatrix + g.RawMatrix)
+     (f,g) -> map(target f, source f, reduce(target f, raw f + raw g))
      ) @@ toSameRing
 Matrix + RingElement := (f,r) -> if r == 0 then f else f + r*id_(target f)
 RingElement + Matrix := (r,f) -> if r == 0 then f else r*id_(target f) + f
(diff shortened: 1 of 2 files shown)


mahrud:
Both places you changed call `map(Module, Module, RawMatrix)`, so I was thinking that calling `reduce` only once there is more straight forward, as Dan suggested in #2898.

d-torrance:
Just to make sure I understand -- the suggestion is to move `reduce` to `map(Module, Module, RawMatrix)` and then remove it from the various matrix operation methods that currently call it?

Are there any situations where we wouldn't want `reduce`?  For example, it looks like `transpose(Matrix)` calls `map(Module, Module, RawMatrix)` without calling `reduce`.  Should it?

mahrud:
I can see arguments for adding reduce everywhere or removing it everywhere:
* on one hand, there's the problem that you pointed out with addition in the Klein group looking silly (although not technically incorrect);
* on the other hand, `reduce` calls `gb presentation tar`, which is less than ideal if `tar` is huge and you just want to take `tar_0 + tar_0` for instance.

I'm not worried about `transpose` because it only works on free modules, in which case `gb presentation tar` is trivial, but if we want to remove calls to `reduce` in some places I think we should remove it (almost) everywhere and just be very clear that in M2 vectors and matrices in non-free modules are _not necessarily_ displayed in their normal form, or perhaps only in normal form if a gb is already known. We should make sure that things like `==` still work though.

@mikestillman thoughts? This could make M2 much slower in some cases.

PR #3014, opened 2023-11-28
Simplify construction of vectors

d-torrance:
Previously, if we wanted a `Vector` object from some non-free module, we'd need to use `new` and build a matrix:

```m2
i1 : R = QQ[x,y,z,w];

i2 : M = (comodule monomialCurveIdeal(R, {1,2,3}))^3;

i3 : new M from map(M, R^1, {{z^2}, {y*z}, {y^2}})

o3 = | yw |
     | xw |
     | xz |

o3 : cokernel | z2-yw yz-xw y2-xz 0     0     0     0     0     0     |
              | 0     0     0     z2-yw yz-xw y2-xz 0     0     0     |
              | 0     0     0     0     0     0     z2-yw yz-xw y2-xz |
```

This PR gives us an easier way to do this, similar to how `vector(List)` works for free modules:

```m2
i1 : R = QQ[x,y,z,w];

i2 : M = (comodule monomialCurveIdeal(R, {1,2,3}))^3;

i3 : M {z^2, y*z, y^2}

o3 = | yw |
     | xw |
     | xz |

o3 : cokernel | z2-yw yz-xw y2-xz 0     0     0     0     0     0     |
              | 0     0     0     z2-yw yz-xw y2-xz 0     0     0     |
              | 0     0     0     0     0     0     z2-yw yz-xw y2-xz |

```

We also add a few other vector operations (like right scalar multiplication and unary `+`).  Keeping this as a draft for now until I can add unit tests and docs.

### DIFF
--- a/M2/Macaulay2/m2/modules.m2
+++ b/M2/Macaulay2/m2/modules.m2
@@ -117,6 +117,7 @@ lift(Vector,Number) := Vector => o -> (v,S) -> vector (lift(v#0,S))
 
 - Vector := Vector => v -> new class v from {-v#0}
 Number * Vector := RingElement * Vector := Vector => (r,v) -> vector(r * v#0)
+Vector * Number := Vector * RingElement := Vector => (v,r) -> vector(v#0 * r)
 Vector + Vector := Vector => (v,w) -> vector(v#0+w#0)
 Vector - Vector := Vector => (v,w) -> vector(v#0-w#0)
 Vector ** Vector := Vector => (v,w) -> vector(v#0**w#0)
--- a/M2/Macaulay2/m2/modules.m2
+++ b/M2/Macaulay2/m2/modules.m2
@@ -115,6 +115,7 @@ lift(Vector,InexactNumber') :=
 lift(Vector,RingElement) :=
 lift(Vector,Number) := Vector => o -> (v,S) -> vector (lift(v#0,S))
 
++ Vector := Vector => identity
 - Vector := Vector => v -> new class v from {-v#0}
 Number * Vector := RingElement * Vector := Vector => (r,v) -> vector(r * v#0)
 Vector * Number := Vector * RingElement := Vector => (v,r) -> vector(v#0 * r)
--- a/M2/Macaulay2/m2/modules.m2
+++ b/M2/Macaulay2/m2/modules.m2
@@ -154,6 +154,10 @@ new Module from Sequence := (Module,x) -> (
      	       symbol numgens => rawRank rM
      	       })) x
 
+vector(Module, Matrix) := (M, f) -> vector map(M,,entries f)
+vector(Module, List)   := (M, v) -> vector map(M,,apply(splice v, i -> {i}))
+vector(Module, RingElement) := vector(Module, Number) := (M, x) -> vector(M, {x})
+
 -- TODO: deprecate these
 degreesMonoid Module := GeneralOrderedMonoid => M -> degreesMonoid ring M
 degreesRing Module := PolynomialRing => M -> degreesRing ring M
--- a/M2/Macaulay2/m2/modules.m2
+++ b/M2/Macaulay2/m2/modules.m2
@@ -79,6 +79,8 @@ entries Vector := v -> entries ambient v#0 / first
 norm Vector := v -> norm v#0
 expression Vector := v -> VectorExpression apply(flatten entries super v#0,expression)
 net Vector := v -> net expression v
+describe Vector := v -> Describe expression FunctionApplication(
+    vector, (describe module v, describe \ flatten entries matrix v))
 toExternalString Vector :=
 toString Vector := v -> toString expression v
 texMath Vector := v -> texMath expression v
--- a/M2/Macaulay2/m2/modules.m2
+++ b/M2/Macaulay2/m2/modules.m2
@@ -68,6 +68,7 @@ vector Matrix := f -> (
      new target f from {f}
     )
 vector List := v -> vector matrix apply(splice v, i -> {i});
+vector RingElement := vector Number := x -> vector {x}
 
 -----------------------------------------------------------------------------
 
--- a/M2/Macaulay2/m2/modules.m2
+++ b/M2/Macaulay2/m2/modules.m2
@@ -81,7 +81,7 @@ expression Vector := v -> VectorExpression apply(flatten entries super v#0,expre
 net Vector := v -> net expression v
 describe Vector := v -> Describe expression FunctionApplication(
     vector, (describe module v, describe \ flatten entries matrix v))
-toExternalString Vector :=
+toExternalString Vector := toString @@ describe
 toString Vector := v -> toString expression v
 texMath Vector := v -> texMath expression v
 --html Vector := v -> html expression v
--- a/M2/Macaulay2/packages/Macaulay2Doc/doc.m2
+++ b/M2/Macaulay2/packages/Macaulay2Doc/doc.m2
@@ -45,7 +45,6 @@ undocumented {
     (symbol*,  ZZ, ChainComplexMap),
     (symbol*,  Expression, OneExpression),
     (symbol*,  OneExpression, Expression),
-    (symbol*,  Number, Vector)
     }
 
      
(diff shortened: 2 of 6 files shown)


mahrud:
These also, work, no?
```m2
m = map(M, , {{z^2}, {y*z}, {y^2}})
v1 = vector m
v2 = m_0
```

pzinn:
you can just use the usual `M_0` type construction

d-torrance:
Hmm, yeah, I kind of like `List_Module`.

My thinking was that `Module List` would look very similar to `vector(List)`.

Maybe both?

mahrud:
> My thinking was that `Module List` would look very similar to `vector(List)`.

Maybe what you are looking for it `vector(Module, List)` to mimic `matrix(Ring, List)` and `map(Module, Nothing, List)`.

d-torrance:
I like it!  Here's the example now:

```m2
i1 : R = QQ[x,y,z,w];

i2 : M = (comodule monomialCurveIdeal(R, {1,2,3}))^3;

i3 : vector(M, {z^2, y*z, y^2})

o3 = | yw |
     | xw |
     | xz |

o3 : cokernel | z2-yw yz-xw y2-xz 0     0     0     0     0     0     |
              | 0     0     0     z2-yw yz-xw y2-xz 0     0     0     |
              | 0     0     0     0     0     0     z2-yw yz-xw y2-xz |

```

M2/Macaulay2/m2/modules.m2
```diff
@@ -68,6 +68,8 @@ vector Matrix := f -> (
      new target f from {f}
     )
 vector List := v -> vector matrix apply(splice v, i -> {i});
+vector RingElement :=
+vector Number      := x -> vector {x}
```
mahrud, line 72: Methods that apply to `RingElement` and `Number` kinda look better on a single line.

M2/Macaulay2/m2/modules.m2
```diff
@@ -151,6 +155,10 @@ new Module from Sequence := (Module,x) -> (
      	       symbol numgens => rawRank rM
      	       })) x
 
+vector(Module, List) := (M, v) -> vector map(M,,apply(splice v, i -> {i}))
+vector(Module, RingElement) :=
+vector(Module, Number)      := (M, x) -> vector(M, {x})
```
mahrud, line 160: It might also be worth adding `vector(Module, Matrix)` and changing how vectors are printed with `peek` to use that instead of this:
```m2
i36 : v

o36 = | 1 |
      | 1 |

o36 : cokernel | 2 0 |
               | 0 2 |

i37 : peek v

o37 = cokernel | 2 0 |{| 1 |}
               | 0 2 | | 1 |
```

M2/Macaulay2/m2/modules.m2
d-torrance: Yeah I still have docs and tests to add, so I'll fix this when I push those.

M2/Macaulay2/m2/modules.m2
d-torrance: So `vector(Module,  Matrix)` would be something like `(M, f) -> vector map(M, ,f)`?

What are you suggesting for `peek`?  `peek` (which wraps around `peek'`) is pretty low-level and is currently interpreting each `Vector` object as its parent `BasicList`.  There don't appear to be any other mathematical objects with their own `peek'` methods.

mahrud: Ah, true about `peek`. Then let's fix `toString` and friends at least.

> So `vector(Module, Matrix)` would be something like `(M, f) -> vector map(M, ,f)`?

Yes, and perhaps `vector(Module, List)` can call this for simplicity.

d-torrance: `toString(Vector)` is updated now:

```m2
i1 : toString vector {1, 2, 3}

o1 = vector(ZZ^3, {1, 2, 3})
```

I've also added `vector(Module, Matrix)`.  I kept `vector(Module, List)` alone since `map(Module,Nothing,List)` does a bunch of the heavy lifting for us.

M2/Macaulay2/m2/ringmap.m2
```diff
@@ -193,9 +193,7 @@ RingMap MutableMatrix := MutableMatrix => (p,m) -> (
      then error "expected source of ring map to be the same as ring of matrix";
      map(S,rawRingMapEval(raw p, raw m)))
 
-RingMap Vector := Vector => (p,m) -> (
-     f := p new Matrix from m;
-     new target f from f)
+RingMap Vector := Vector => (p,m) -> vector p new Matrix from m
```
mahrud, line 196: Very minor, but I think just `vector p m#0` works here.

Kinda strange to have `new Matrix from Vector` when `matrix Vector` does the same (though perhaps `matrix(Vector, Degree => d)` should use the option).

d-torrance: Yeah, I was just blinding replacing `new target f from` with `vector` without looking at the stuff afterwards.  This definitely could be improved.

M2/Macaulay2/m2/modules.m2
mahrud: Hmm, the difference between `map(Module, Nothing, List)` and `map(Module, Nothing, Matrix)` is kind of mysterious to me. Does this also fix https://github.com/Macaulay2/M2/issues/3016#issuecomment-1830305241?

d-torrance: Yeah, I haven't looked at the code very much.  I just noticed that one did the thing I wanted and the other didn't:

```m2
i1 : R = QQ[x];

i2 : vector map(R^1,, matrix {{5}})
stdio:2:8:(3): error: expected module and matrix for the same ring

i3 : vector map(R^1,, {{5}})

o3 = | 5 |

      1
o3 : R
```

pzinn:
I'm not sure I understand or agree with the changes to `VectorExpression`.
Right now `VectorExpression` follows the same principle as `MatrixExpression`, that is it just records the entries. I'm not sure why one should keep track of the module itself, since in most circumstances it wouldn't be displayed.

d-torrance:
> I'm not sure I understand or agree with the changes to `VectorExpression`. Right now `VectorExpression` follows the same principle as `MatrixExpression`, that is it just records the entries. I'm not sure why one should keep track of the module itself, since in most circumstances it wouldn't be displayed.

This was my attempt at responding to @mahrud's comment about fixing `toString(Vector)`.  I think it's useful to not forget the module, especially when using `value`.  For example:

### Before
```m2
i1 : R = QQ[x];

i2 : v = vector map(R^3,, {{1}, {2}, {3}})

o2 = | 1 |
     | 2 |
     | 3 |

      3
o2 : R

i3 : toString v

o3 = vector {1, 2, 3}

i4 : value oo

o4 = | 1 |
     | 2 |
     | 3 |

       3
o4 : ZZ
```

### After
```m2
i1 : R = QQ[x];

i2 : v = vector map(R^3,, {{1}, {2}, {3}})

o2 = | 1 |
     | 2 |
     | 3 |

      3
o2 : R

i3 : toString v

o3 = vector(R^3, {1, 2, 3})

i4 : value oo

o4 = | 1 |
     | 2 |
     | 3 |

      3
o4 : R
```

This also works using `expression` instead of `toString`.

mahrud:
Although, I think `toExternalString` should give something like `vector(R^3, {1, 2, 3})`. Usually `toString` should just convert the output of `expression` to a string. Here's a rule of thumb:
```m2
net              X :=      net @@ expression
toString         X := toString @@ expression
toExternalString X := toString @@ describe
```

pzinn:
I notice your use of  `vector` everywhere now. In the lot of this code, I had intentionally avoided `vector` because it has a useless test (e.g., when multiplying a matrix with a vector, we know the source is still a free module of rank 1). It's debatable what is better here.

d-torrance:
> I notice your use of `vector` everywhere now. In the lot of this code, I had intentionally avoided `vector` because it has a useless test (e.g., when multiplying a matrix with a vector, we know the source is still a free module of rank 1). It's debatable what is better here.

That makes sense!  I originally had added that commit because the original draft of this PR had a `new Vector from List` method, which broke all of these calls.  But that's not an issue any more, so I've removed that commit.

PR #3015, opened 2023-11-28
Include modules assigned to global variables in Matrix#AfterPrint

d-torrance:
This way, we might be able to see some more information about maps between non-free modules.

### Before
```m2
i1 : R = QQ[x,y,z];

i2 : M = comodule ideal x

o2 = cokernel | x |

                            1
o2 : R-module, quotient of R

i3 : map(M, R^3, {{x, y, z}})

o3 = | 0 y z |

o3 : Matrix
```

What are the source and target?

### After
```m2
i1 : R = QQ[x,y,z];

i2 : M = comodule ideal x

o2 = cokernel | x |

                            1
o2 : R-module, quotient of R

i3 : map(M, R^3, {{x, y, z}})

o3 = | 0 y z |

                    3
o3 : Matrix M <--- R
```

### DIFF
--- a/M2/Macaulay2/m2/matrix1.m2
+++ b/M2/Macaulay2/m2/matrix1.m2
@@ -397,10 +397,12 @@ Number ** RingElement :=
 RingElement ** Number := 
 RingElement ** RingElement := (r,s) -> matrix {{r}} ** matrix {{s}}
 
--- Matrix#AfterPrint = Matrix#AfterNoPrint = f -> (class f, " ", new MapExpression from {target f,source f})
 Matrix#AfterPrint = Matrix#AfterNoPrint = f -> (
     class f,
-    if isFreeModule target f and isFreeModule source f then  (" ", new MapExpression from {target f,source f})
+    (
+	(tar, src) := apply((target f, source f), M -> moduleAbbrv(M, null));
+	if tar =!= null and src =!= null
+	then (" ", new MapExpression from expression \ {tar, src}))
     )
 
 -- precedence Matrix := x -> precedence symbol x
--- a/M2/Macaulay2/m2/modules.m2
+++ b/M2/Macaulay2/m2/modules.m2
@@ -82,6 +82,14 @@ toExternalString Vector :=
 toString Vector := v -> toString expression v
 texMath Vector := v -> texMath expression v
 --html Vector := v -> html expression v
+
+-- helper for Matrix#AfterPrint
+moduleAbbrv = (M, abbrv) -> (
+    if isFreeModule M then M
+    else if hasAttribute(M, ReverseDictionary)
+    then getAttribute(M, ReverseDictionary)
+    else abbrv)
+
 ring Vector := v -> ring class v
 module Vector := v -> target v#0
 leadTerm Vector := v -> new class v from leadTerm v#0
--- a/M2/Macaulay2/m2/modules.m2
+++ b/M2/Macaulay2/m2/modules.m2
@@ -83,12 +83,14 @@ toString Vector := v -> toString expression v
 texMath Vector := v -> texMath expression v
 --html Vector := v -> html expression v
 
--- helper for Matrix#AfterPrint
+-- helper for Matrix#AfterPrint and Vector#AfterPrint
 moduleAbbrv = (M, abbrv) -> (
     if isFreeModule M then M
     else if hasAttribute(M, ReverseDictionary)
     then getAttribute(M, ReverseDictionary)
     else abbrv)
+
+Vector#AfterPrint = v -> moduleAbbrv(module v, Vector)
 
 ring Vector := v -> ring class v
 module Vector := v -> target v#0
--- a/M2/Macaulay2/m2/chaincomplexes.m2
+++ b/M2/Macaulay2/m2/chaincomplexes.m2
@@ -124,7 +124,8 @@ net ChainComplex := C -> (
      else (
 	  a := s#0;
 	  b := s#-1;
-	  horizontalJoin between(" <-- ", apply(a .. b,i -> stack (net C_i," ",net i)))))
+	  horizontalJoin between(" <-- ", apply(a .. b,i ->
+		  stack (net moduleAbbrv(C_i, C_i)," ",net i)))))
 
 texMath ChainComplex := C -> (
      complete C;
--- a/M2/Macaulay2/m2/chaincomplexes.m2
+++ b/M2/Macaulay2/m2/chaincomplexes.m2
@@ -133,7 +133,7 @@ texMath ChainComplex := C -> (
      if # s === 0 then "0" else
      concatenate apply(s,i->(
 	     if i>s#0 then "\\,\\xleftarrow{" | texMath short C.dd_i | "}\\,",
-	     texUnder(texMath C_i,i)
+	     texUnder(texMath moduleAbbrv(C_i, C_i),i)
 	     ))
       )
 
--- a/M2/Macaulay2/m2/expressions.m2
+++ b/M2/Macaulay2/m2/expressions.m2
@@ -1282,7 +1282,7 @@ toString'(Function, MapExpression) := (fmt,x) -> toString'(fmt,new FunctionAppli
 lineOnTop := (s) -> concatenate(width s : "-") || s
 net MapExpression := x-> if #x>2 then horizontalJoin(net x#0, " <--",
 		    lineOnTop net x#2,
-		    "-- ", net x#1) else net x#0 | " <--- " | net x#1
+		    "-- ", net x#1) else net x#0 | " <-- " | net x#1
 texMath MapExpression := x -> texMath x#0 | "\\," | (if #x>2 then "\\xleftarrow{" | texMath x#2 | "}" else "\\longleftarrow ") | "\\," | texMath x#1
 expressionValue MapExpression := x -> map toSequence apply(x,expressionValue)
 
--- a/M2/Macaulay2/packages/Complexes/ChainComplex.m2
+++ b/M2/Macaulay2/packages/Complexes/ChainComplex.m2
@@ -409,6 +409,7 @@ Complex#id = (C) -> (
     result
     )
 
+importFrom(Core, {"moduleAbbrv"})
 net Complex := C -> (
      (lo,hi) := C.concentration;
      if lo > hi then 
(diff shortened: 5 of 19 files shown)


pzinn:
note that this was discussed in https://github.com/Macaulay2/M2/pull/2729
I had proposed the very simple
```
Matrix#AfterPrint = Matrix#AfterNoPrint = f -> (class f, " ", new MapExpression from {target f,source f})
```
but ended up reverting it after @mahrud opposed it

d-torrance:
Yeah, I noticed that.  The proposed change won't print huge modules -- only ones which have been assigned to global variables.

mahrud:
I like Doug's version.

It would be great if you could also change this:
```m2
i64 : M = image [x,y,z]

o64 = image | x y z |

                              1
o64 : R-module, submodule of R

i65 : M_0

o65 = | x |

o65 : image | x y z |
```
to simply print:
```m2
i65 : M_0

o65 = | x |

o65 : M
```

pzinn:
there's a more general problem which is the inconsistency in the way modules are displayed. Despite having ReverseDictionary, it's never used: compare
```
i1 : R=QQ[x]

o1 = R

o1 : PolynomialRing

i2 : M=R^1

      1
o2 = R

o2 : R-module, free
```

d-torrance:
Yeah, an earlier draft of this branch had changes to `net(Module)` that would have addressed those points, but for whatever reason I decided to remove them before pushing.  You've both convinced me to bring them back.  :)

d-torrance:
Here's the new proposed behavior:

```m2
i1 : R = QQ[x];

i2 : M = R^1

o2 = M

o2 : R-module, free

i3 : M_0

o3 = | 1 |

o3 : M

i4 : map(M, M, 5)

o4 = | 5 |

o4 : Matrix M <--- M
```

mahrud:
This is not a good idea for free modules:
```m2
i2 : M = R^1

o2 = M

o2 : R-module, free
```
It leads to nonsense like this:
```m2
i8 : res coker vars R

o8 = M <-- P <-- 0
                  
     0     1     2
```

Also, for non-free modules, `describe Module` should print what used to be printed by `net Module`.

d-torrance:
I've updated things so free modules retain their current `net` behavior:


```m2
i1 : R = QQ[x];

i2 : M = comodule ideal x

o2 = M

                            1
o2 : R-module, quotient of R

i3 : N = R^3

      3
o3 = R

o3 : R-module, free

i4 : map(M, N, {{1, 2, 3}})

o4 = | 1 2 3 |

                    3
o4 : Matrix M <--- R
```

> also note that a lot of documentation and examples will need rewriting

Certainly the `ComputationsBook` tests would need to be updated, but would anything else?

M2/Macaulay2/m2/modules.m2
pzinn, line 1: normally the ReverseDictionary is part of `expression`, not `net`, otherwise the output will depend on the mode. check out `expression PolynomialRing`, for example

d-torrance: Thanks -- fixed!

mahrud:
Could you fix `describe Module`?
```m2
i2 : M = image vars R

o2 = M

                             1
o2 : R-module, submodule of R

i3 : describe M

                 1   {{-1}, {-1}}
o3 = image(map (R , R            , {{x, y}}))
```
Ideally it should print:
```m2
i2 : describe M

o2 = image | x y |

                             1
o2 : R-module, submodule of R
```

d-torrance:
> Could you fix `describe Module`?

Done!

New behavior:

```m2
i1 : R = QQ[x]

o1 = R

o1 : PolynomialRing

i2 : describe image matrix {{x}}

o2 = image | x |

i3 : describe coker matrix {{x}}

o3 = cokernel | x |

i4 : describe subquotient(matrix {{x}}, matrix {{x^2}})

o4 = subquotient (| x |, | x2 |)
```

d-torrance:
Hmm, documentation-wise, I am wondering whether it's a good idea to use global variable names for `net(Module)` without going through and updating a bunch of examples to, say,  use `describe`.  For example, the [subquotient modules](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_subquotient_spmodules.html) node kind of relies on us noticing that `N1` has four more generators than `N`.

M2/Macaulay2/m2/matrix1.m2
```diff
@@ -400,7 +400,12 @@ RingElement ** RingElement := (r,s) -> matrix {{r}} ** matrix {{s}}
 -- Matrix#AfterPrint = Matrix#AfterNoPrint = f -> (class f, " ", new MapExpression from {target f,source f})
 Matrix#AfterPrint = Matrix#AfterNoPrint = f -> (
     class f,
-    if isFreeModule target f and isFreeModule source f then  (" ", new MapExpression from {target f,source f})
+    (
+	-- print source/target if both are free modules and/or have
+	-- been assigned to global variables
+	modulenames := select({target f, source f}, M -> (
+		isFreeModule M or hasAttribute(M, ReverseDictionary)));
+	if #modulenames == 2 then (" ", new MapExpression from modulenames))
```
mahrud, line 408: In response to your worry about documentation, how about revert `expression(Module)` to what it was, but keep printing the global name for matrix afterprints:
```suggestion
        (tar, src) := apply((target f, source f),
            M -> if isFreeModule M then M
            else if hasAttribute(M, ReverseDictionary)
            then getAttribute(M, ReverseDictionary));
        if tar =!= null and src =!= null
        then (" ", new MapExpression from expression \ {tar, src}))
```

mahrud: Actually, it would be nice to also use the global symbol in resolutions. Maybe in addition to the above, `net(Module)` should call `describe` instead, and we make sure that in printing of resolutions we call `expression M` not just `net M`. Would that solve the documentation issue?

mahrud: It's not just about the documentation. I tried using M2 with these changes for a couple of days because I really liked having non-free modules in resolutions being displayed simply by their symbol, but I actually **really** dislike having to type `describe M` every time I want to see the structure of a module outside of a resolution. This is not really the case for rings because typically users don't have a ton of rings around anyway.

Also, adding useful afterprints for matrices and vectors is a small addition, but changing a major behavior like printing of modules probably requires a larger conversation.

d-torrance:
Ok, this PR now just affects `Matrix#AfterPrint` and `Vector#AfterPrint`.  So for example:

```m2
i1 : R = QQ[x,y,z];

i2 : M = image vars R

o2 = image | x y z |

                             1
o2 : R-module, submodule of R

i3 : map(M,, {{1}, {2}, {3}})

o3 = {1} | 1 |
     {1} | 2 |
     {1} | 3 |

                    1
o3 : Matrix M <--- R

i4 : vector oo

o4 = | x+2y+3z |

o4 : M

i5 : (coker vars R)_0

o5 = | 1 |

o5 : Vector
```

M2/Macaulay2/m2/matrix1.m2
d-torrance: > Actually, it would be nice to also use the global symbol in resolutions.

Do we ever have resolutions w/ non-free modules?

mahrud: I do:
```m2
i17 : R = kk[x,y];

i18 : chainComplex apply(1..2, i -> truncate_3 koszul_i vars R)

o18 = image | y3 xy2 x2y x3 | <-- image {1} | y2 xy x2 0  0  0  | <-- image {2} | y x |
                                        {1} | 0  0  0  y2 xy x2 |      
      0                                                               2
                                  1

o18 : ChainComplex
```

mahrud: But happy to add this as a local change if you want to limit the scope of this PR.

d-torrance:
It works for resolutions now too:

```m2
i1 : needsPackage "Truncations";
 -- warning: symbol "Base" in Schubert2.Dictionary is shadowed by a symbol in Complexes.Dictionary
 --   use the synonym Schubert2$Base

i2 : R = QQ[x,y];

i3 : C = chainComplex apply(1..2, i -> truncate_3 koszul_i vars R)

o3 = image | y3 xy2 x2y x3 | <-- image {1} | y2 xy x2 0  0  0  | <-- image {2} | y x |
                                       {1} | 0  0  0  y2 xy x2 |      
     0                                                               2
                                 1

o3 : ChainComplex

i4 : (M, N, P) = apply(0..2, i -> C_i);

i5 : C

o5 = M <-- N <-- P
                  
     0     1     2

o5 : ChainComplex
```

mahrud:
Thank you!

Super minor note: `<--` in complexes vs `<---` in the matrix afterprint.
```m2
i6 : C

o6 = M <-- N <-- P
                  
     0     1     2

o6 : ChainComplex

i7 : C.dd_1

o7 = {3} | 0 0 0 1 0 0 |
     {3} | 1 0 0 0 1 0 |
     {3} | 0 1 0 0 0 1 |
     {3} | 0 0 1 0 0 0 |

o7 : Matrix M <--- N
```
(This may actually be my fault, I don't remember)

mahrud:
I think `<--` is better. There's a note by @pzinn saying:
> note that one can't have a symbol <---

pzinn:
sadly, `ChainComplex` is not expressionified (yet?), so the same change will have to be done for other output types (`texMath`).

d-torrance:
> sadly, `ChainComplex` is not expressionified (yet?), so the same change will have to be done for other output types (`texMath`).

Ah ok.  Maybe I'll go ahead and expressionify it...

pzinn:
> > sadly, `ChainComplex` is not expressionified (yet?), so the same change will have to be done for other output types (`texMath`).
> 
> Ah ok. Maybe I'll go ahead and expressionify it...

might be tricky

pzinn:
I don't see the changes to `describe`. am I missing something?

d-torrance:
> I don't see the changes to `describe`. am I missing something?

They were removed.

d-torrance:
> might be tricky

You're right -- it is!  `net(ChainComplex)` and `texMath(ChainComplex)` behave totally differently (the latter displays the matrices), so I think this is a task for another time.  I went ahead and just updated the `texMath` methods to use the global variable names, too.

d-torrance:
I'm not sure about having `MapExpression` having any control over how the source and target are printed.  It seems like it should just blindly put an arrow between the two things, and the caller should decide how those two things are printed.

Issue #3017, opened 2023-11-28
smithNormalForm bug


pzinn:
found by @mahrud in the discussion of #1273
```
i1 : debugLevel=101; R=QQ[x]

o2 = R

o2 : PolynomialRing

i3 : smithNormalForm matrix{{x^3+1},{x^2+1}}
-- count = 0
-- lead terms in columns (gb    ) : | 0  |
                                    | x2 |
-- lead terms in rows    (not gb) : | x3 |
                                    | x2 |
-- count = 1
-- lead terms in rows    (gb    ) : {0, 3} | 2 |
-- lead terms in columns (not gb) : {0, 3} | 2 |
-- count = 2
-- lead terms in columns (gb    ) : {0, 3} | 2 |
-- lead terms in rows    (not gb) : {0, 3} | 2 |

o3 = (| 2 |, | x+1       -x2-x+1    |, {3} | 1 |)
      | 0 |  | x2+1      -x3-1      |
      | 0 |  | x3+x2+x+1 -x4-x3-x-1 |

o3 : Sequence
```
the problem seems to be at the level of the GB computation:
```
i4 : syz gb (transpose matrix{{x^3+1},{x^2+1}},Syzygies=>true)

o4 = | x2+1  x3+x2+x+1  |
     | -x3-1 -x4-x3-x-1 |
```
which finds 2 syzygies (?)  which confuses the Smith normal form algorithm.

pzinn:
might as well mention the other issue that I found while investigating this: the first matrix of the output of `smithNormalForm` depends on the value of `ChangeMatrix` (even though it probably shouldn't). The reason is pretty simple: when `ChangeMatrix` is turned on (either for rows or columns), it will add extra zero rows/columns, but not if turned off. The obvious fix is to always add those extra zero rows/columns (which means always calling `gb` with `Syzygies=>true`).

pzinn:
note that after merging #3710:
```
i1 : R=QQ[x]

o1 = R

o1 : PolynomialRing

i2 : syz gb (transpose matrix{{x^3+1},{x^2+1}},Syzygies=>true)

o2 = | x2+1  x3+x2+x+1  |
     | -x3-1 -x4-x3-x-1 |

             2      2
o2 : Matrix R  <-- R
```
so the underlying bug is not fixed.

aalmousa:
> note that after merging [#3710](https://github.com/Macaulay2/M2/pull/3710):
> 
> ```
> i1 : R=QQ[x]
[...]

Issue #1273, opened 2020-06-14
(trim, Ideal) should work correctly in single variable polynomial rings


kschwede:
When working over a PID, trim should work correctly, even if the ideal is not homogeneous.  

```
i1 : R = ZZ/7[t];

i2 : I = ideal(t^3*(t+1), t*(t-1))

             4    3   2
o2 = ideal (t  + t , t  - t)

o2 : Ideal of R

i3 : trim I

                2
o3 = ideal (t, t  - t)

o3 : Ideal of R
```
Obviousy the gcd command on the generators works ok, but presumably this should work via trim as well.

This probably isn't a priority for version #1.16, but it should be easy to implement.

pzinn:
I personally use this on my inhomogeneous ideals. it will in particular work for a single variable.
```
-- simplifying inhomogeneous ideals
simplify = x -> (
    R:=ring x;
    local u;
    flag:=true;
    y:=new MutableList from first entries gens x;
    while flag do (
	flag=false; 
	scan(#y, i-> (
		u=y#i;
		y#i=0_R;
		y#i=u%(ideal toList y);
		if y#i =!= u then flag=true;
		));
	);
    trim ideal toList y -- just to remove zeroes
    );
```

kschwede:
That's very nice and I am definitely taking it :-).  Although maybe then there should be a simplify function in Macaulay2 if we don't want this behavior implemented into trim (I imagine for Euclidean domains though, the gcd function might be faster if it's implemented at a lower level, athough I didn't look at the code).

pzinn:
I realize this still hasn't been fixed. it's a bit ridiculous that M2 still can't simplify this:
```
i1 : R=QQ[x]

o1 = R

o1 : PolynomialRing

i2 : ideal(3*x^4-x^3+x^2-x-2,3*x^4+2*x^3+6*x+4)

              4    3    2            4     3
o2 = ideal (3x  - x  + x  - x - 2, 3x  + 2x  + 6x + 4)

o2 : Ideal of R

i3 : trim oo

              3    2             4    3    2
o3 = ideal (3x  - x  + 7x + 6, 3x  - x  + x  - x - 2)

o3 : Ideal of R

```

mikestillman:
I agree! 

In general, how should we trim ideals which are not homogeneous?
- For PID's, it makes sense, but is obviously not being done still (or rather, any more)
- For ideals in local rings, such as exterior algebra, even when not homogeneous, this still makes sense.
[...]

PR #3018, opened 2023-11-29
trim PID

pzinn:
- correct trimming of modules / ideals in PIDs. resolves (at last!) #1273:
```
i1 : R = ZZ/7[t];

i2 : I = ideal(t^3*(t+1), t*(t-1))

             4    3   2
o2 = ideal (t  + t , t  - t)

o2 : Ideal of R

i3 : trim I

o3 = ideal t

o3 : Ideal of R

i4 : R=QQ[x]

o4 = R

o4 : PolynomialRing

i5 : ideal(3*x^4-x^3+x^2-x-2,3*x^4+2*x^3+6*x+4)

              4    3    2            4     3
o5 = ideal (3x  - x  + x  - x - 2, 3x  + 2x  + 6x + 4)

o5 : Ideal of R

i6 : trim oo

               2
o6 = ideal(x + -)
               3

o6 : Ideal of R
```
- several bugs were found in the process. #3017 is *not* fixed in this PR, somebody else needs to take a look at it since a lot is problematic with `smithNormalForm` at the moment.
- instead, a temporary workaround in the `prune` PID strategy was implemented, and similarly for the new `trim` PID strategy.
- tests were added for both `trim` and `prune`.

### DIFF
--- a/M2/Macaulay2/m2/matrix2.m2
+++ b/M2/Macaulay2/m2/matrix2.m2
@@ -247,6 +247,26 @@ trimHelper = ((opts, M) -> (
 addHook((trim, Module), Strategy => Inhomogeneous, (opts, M) -> trimHelper(opts ++ {Strategy => Inhomogeneous}, M))
 addHook((trim, Module), Strategy => Complement,    (opts, M) -> trimHelper(opts ++ {Strategy => Complement},    M))
 
+trimPID := M -> if M.?relations then (if M.?generators then trimPID image generators M else ambient M) / trimPID image relations M else if not M.?generators then M else (
+    f := presentation M;
+    (g,ch) := smithNormalForm(f, ChangeMatrix => {true, false});
+    isunit := r -> r != 0 and degree r === {0};
+    rows := select(min(rank source f,rank target f),i->isunit g_(i,i));
+    rows = rows | toList(rank target f..<rank target g); -- temporary fix for #3017
+    ch = submatrix'(ch,rows,);
+    p:=id_(target ch)//ch;
+    q:=generators M*p;
+    d:=diagonalMatrix apply(rank source q,i->lcm apply(entries (q_i),a->if a==0 then 1 else 1/leadCoefficient a));
+    image (q*d)
+    )
+addHook((trim, Module), Strategy => "PID",
+    (opts, M) -> (
+	R := ring M;
+	if instance(R,PolynomialRing) and numgens R === 1 and isField coefficientRing R and not isHomogeneous M then trimPID M
+	)
+    )
+
+
 syz Matrix := Matrix => opts -> (f) -> (
     c := runHooks((syz, Matrix), (opts, f));
     if c =!= null then return c;
--- a/M2/Macaulay2/m2/matrix2.m2
+++ b/M2/Macaulay2/m2/matrix2.m2
@@ -251,7 +251,7 @@ trimPID := M -> if M.?relations then (if M.?generators then trimPID image genera
     f := presentation M;
     (g,ch) := smithNormalForm(f, ChangeMatrix => {true, false});
     isunit := r -> r != 0 and degree r === {0};
-    rows := select(min(rank source f,rank target f),i->isunit g_(i,i));
+    rows := select(min(rank source g,rank target g),i->isunit g_(i,i));
     rows = rows | toList(rank target f..<rank target g); -- temporary fix for #3017
     ch = submatrix'(ch,rows,);
     p:=id_(target ch)//ch;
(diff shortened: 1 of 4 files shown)


M2/Macaulay2/tests/normal/trim.m2
```diff
@@ -25,6 +25,13 @@ assert (trim(M,Strategy=>Complement) === trim(N,Strategy=>Complement))
 assert (trim P === R^1)
 assert (trim(P,Strategy=>Complement) === R^1)
 
+-- trim PID
+R=QQ[x]
+I = ideal(3*x^4-x^3+x^2-x-2,3*x^4+2*x^3+6*x+4)
+J = trim I
+assert(I == J)
+assert(numgens J == 1)
```
mahrud, line 33: Here's an example that currently fails, if you'd like to use different examples:
```m2
i1 : R = QQ[x]

o1 = R

o1 : PolynomialRing

i2 : I = ideal(x^3-1, x^2+1)

             3       2
o2 = ideal (x  - 1, x  + 1)

o2 : Ideal of R

i3 : trim I
stdio:3:1:(3): error: matrix column index 1 out of range 0 .. 0
```

M2/Macaulay2/m2/matrix2.m2
```diff
@@ -247,6 +247,26 @@ trimHelper = ((opts, M) -> (
...
+    p:=id_(target ch)//ch;
+    q:=generators M*p;
+    d:=diagonalMatrix apply(rank source q,i->lcm apply(entries (q_i),a->if a==0 then 1 else 1/leadCoefficient a));
+    image (q*d)
+    )
+addHook((trim, Module), Strategy => "PID",
+    (opts, M) -> (
+	R := ring M;
+	if instance(R,PolynomialRing) and numgens R === 1 and isField coefficientRing R and not isHomogeneous M then trimPID M
+	)
```
mahrud, line 266: Rather than repeating the prune algorithm in trimPID, how about this instead?
```suggestion
	then if isFreeModule M then M else subquotient(
	    if M.?generators   then (prune image generators M).cache.pruningMap,
	    if M.?relations    then (prune image relations  M).cache.pruningMap))
```

pzinn: I don't think what you wrote works. anyway I did have an earlier version going through `prune`, but it was fairly complicated. I think the current code is cleaner.

M2/Macaulay2/tests/normal/trim.m2
pzinn: weird, works just fine here.

M2/Macaulay2/m2/matrix2.m2
mahrud: Sorry, what doesn't work?

pzinn: I tried your code with `coker matrix {{x}}` and it didn't produce the expected result.

mahrud: What's the expected result, not this?!
```m2
i1 : R = QQ[x]; trim coker matrix {{x}}

o2 = cokernel | x |
```

mahrud: You must have used `M.relations` for both inputs of `subquotient`.

pzinn: no, I copy/pasted your code:
```
i1 : test = M -> if isFreeModule M then M else subquotient(
	    if M.?generators   then (prune image generators M).cache.pruningMap,
	    if M.?relations    then (prune image relations  M).cache.pruningMap)
          
o1 = test

o1 : FunctionClosure

i2 : R=QQ[x];

i3 : test cokernel matrix {{x}}

o3 = subquotient (| x |, | x |)

                               1
o3 : R-module, subquotient of R
```

pzinn: also it's about twice as slow due lots of junk in the `prune` code.

mahrud: These are valid improvements to the PID strategy for prune that I think should be implemented, but the main computation at the center of trim and prune for an ideal or module over PID is exactly the same, and duplicating an algorithm in two places is a recipe for bugs and inconsistencies and harder maintenance in the future.

If you want me to suggest a different change that doesn't have the slow down you describe (pull the prune strategy into a helper function and use it both places) or the denominators (divide by the leading coefficients) I can, but this is your PR and since my impression is that you also don't like hard to parse code I wanted to suggest a way to avoid this mess:
```m2
trimPID := M -> if M.?relations then (if M.?generators then trimPID image generators M else ambient M) / trimPID image relations M else if not M.?generators then M else (
```

If you don't care about code duplication and inconsistency between trim and prune then never mind.

pzinn: I'll try to rewrite to avoid redundancies and ifs

PR #3026, opened 2023-12-05
Improve error message for toExternalString(Function)

d-torrance:
When passed an anonymous function with no global name, it prints the location of the function, but enclosed in parentheses.  The Emacs regular expression thinks that the parentheses are part of the filename, so compilation mode fails to jump to the right place and a file dialog opens.

So we call `net(Function)` to enclose it in the usual Emacs-friendly `-*Function[...]*-`.  We also drop the call the `locate`, since `net` will do that for us.

### Before

```m2
i4 : toExternalString oo
stdio:2:1:(3): error: can't convert anonymous function (foo.m2:1:2-1:5) to external string
```

### After
```m2
i4 : toExternalString oo
stdio:2:1:(3): error: can't convert anonymous function -*Function[foo.m2:1:2-1:5]*- to external string
```

(Note that it's still broken for functions defined in `stdio`, but that's an Emacs issue.)

### DIFF
--- a/M2/Macaulay2/m2/nets.m2
+++ b/M2/Macaulay2/m2/nets.m2
@@ -41,10 +41,9 @@ net Command := toString Command := toExternalString Command := f -> (
      )
 
 toExternalString Function := f -> (
-     if hasAttribute(f,ReverseDictionary) then return toString getAttribute(f,ReverseDictionary);
-     t := locate f;
-     if t === null then error "can't convert anonymous function to external string"
-     else error("can't convert anonymous function (",toString t,") to external string")
+     if hasAttribute(f,ReverseDictionary)
+     then toString getAttribute(f,ReverseDictionary)
+     else error("can't convert anonymous function ",net f," to external string")
      )
 
 net Function := toString Function := f -> (


mahrud:
When the address of a function is too long, the address tends to be abbreviated into:
```m2
i4 : toExternalString oo
stdio:2:1:(3): error: can't convert anonymous function -*Function[blahblah. to external string

i5 : 
```
As you can see, `i5 :` is now grayed out because the `-*` block never ended. Is this easy to fix? (also see https://github.com/Macaulay2/M2-emacs/issues/48#issuecomment-1758659220)

d-torrance:
The location shouldn't be cut off in this case.  For example:

```m2
i1 : file = "~/tmp/" | concatenate apply(200, i -> "a") | ".m2"

o1 = ~/tmp/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
     aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.m2

i2 : file << "x -> x" << endl << close

o2 = /home/profzoom/tmp/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.m2

o2 : File

i3 : input file

ii4 : x -> x

oo4 = -*Function[aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.m2:1:2-1:5]*-

oo4 : FunctionClosure

ii5 : 

i6 : toExternalString oo
stdio:4:1:(3): error: can't convert anonymous function -*Function[aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.m2:1:2-1:5]*- to external string
```
And Emacs jumps to the file just fine.

Long locations will get cut off, causing the Emacs highlighting issue, for things that call `silentRobustNet` (like the "no method found" error in the link).  I'm not sure how to fix that yet.

mahrud:
Thanks for the explanation, I thought that was something about `toExternalString` itself.

> Long locations will get cut off, causing the Emacs highlighting issue, for things that call `silentRobustNet` (like the "no method found" error in the link). I'm not sure how to fix that yet.

Maybe just not using `-* ... *-` for functions? I forget what purpose it serves. Alternatively, maybe something like this?
```m2
     horizontalJoin(part1, if match("-*", toString part1) then "..*-", part2));
```
instead of
https://github.com/Macaulay2/M2/blob/master/M2/Macaulay2/m2/robust.m2#L49

Issue #825, opened 2018-06-26
libreadline on the Mac is replaced by libedit


DanGrayson:
See `man editline` for details.  We should modify our configure script to detect its presence and use it, rather than building libreadline ourselves, or depending on brew to install readline for us.  Maybe all we have to do is to call `int el_set(EditLine *e, int op, ...);` to get interrupts not handled, and everything else will work as is, because editline is a good impersonator of readline (except for not having rl_catch_signals`.)

Under Ubuntu it's provided in the package `libedit-dev`.

mahrud:
I'm rather confused about where is the main source for libedit. Is it https://thrysoee.dk/editline/?

Looking online a bit, people also recommend:
- https://github.com/antirez/linenoise
- https://github.com/AmokHuginnsson/replxx

mahrud:
Also, we should get package names to be included in autocomplete (emacs includes them, but readline doesn't)

mahrud:
Patch for libeditline to support colors: https://github.com/mejedi/libeditline-color

d-torrance:
> **Written by Claude** (Claude Opus 5, via Claude Code), not by @d-torrance, whose account posted it -- please weigh it accordingly.

Closing this as won't-fix, on the packaging side of the argument rather than the technical one.

The proposal here is to use libedit *"rather than building libreadline ourselves, or depending on
brew to install readline for us."* Homebrew readline is not in fact a problem in practice — it
installs, it is current, and it is what the macOS builds use — so the motivation for carrying a
second line-editing backend is thin, and a backend that has to be special-cased is not free to
keep working.

Recording where it would start, for anyone who disagrees later: `M2/configure.ac` has no
`libedit` or `editline` path at all, and the readline detection cascade ends at line 1250 with

```
AC_CHECK_DECL(rl_catch_signals,,BUILD_readline=yes, ...)
```

`rl_catch_signals` is precisely the symbol named in the report as the one libedit lacks. So
libedit is not merely unsupported — if it is what provides `readline/readline.h`, that check
fails and M2 builds its own readline anyway. Reviving this means changing that check first.

### The autocomplete remark has a better home

@mahrud's *"we should get package names to be included in autocomplete (emacs includes them, but
readline doesn't)"* is not being closed with this. It is tracked at
**[Macaulay2/M2-emacs#44](https://github.com/Macaulay2/M2-emacs/issues/44)**, which has the
actual design discussion — @DanGrayson proposing that all package symbols simply be added,
[...]

PR #3028, opened 2023-12-05
Add support for "code" inside stdio 

d-torrance:
Currently `code` won't work for functions defined on the standard input:

```m2
i1 : f = x -> x^2

o1 = f

o1 : FunctionClosure

i2 : code f

o2 = 
```

This proposal uses readline's history to get it to work:

```m2
i1 : f = x -> x^2

o1 = f

o1 : FunctionClosure

i2 : code f

o2 = stdio:1:6-1:11: --source code:
     f = x -> x^2
```

However, there's been some discussion (#825) of removing readline in favor of editline, so I'm not sure if we want to do this or not.  Also, the default in Emacs is to call Macaulay2 with `--no-readline`, and this doesn't work then.   So I'm opening this is a draft for now for discussion.

### DIFF
--- a/M2/Macaulay2/d/M2.d
+++ b/M2/Macaulay2/d/M2.d
@@ -43,22 +43,17 @@ export makearrayint(n:int):arrayint := Ccode(returns, "
   GC_CHECK_CLOBBER(z);
   return z; /* Note that getmem_atomic returns zeroed memory */
 ");
-export tocharstar(s:string):charstar := Ccode(returns, "
-  char *p = getmem_atomic(s->len + 1);
-  memcpy(p,s->array,s->len);
-  p[s->len] = 0;
+export tocharstarn(s:string, n:int):charstar := Ccode(returns, "
+  char *p = getmem_atomic(", n + 1, ");
+  memcpy(p,s->array,", n, ");
+  p[", n, "] = 0;
   GC_CHECK_CLOBBER(p);
   return p;
 ");
-export tocharstarOrNull(s:string):charstarOrNull := Ccode(returns, "
-  char *p;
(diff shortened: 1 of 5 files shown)


pzinn:
interesting idea. for the record, the WebApp mode also has no readline.

d-torrance:
I updated this so we still add the contents of the standard input to the history, even when we're not using readline for anything else, so it works even with `--no-readline`:

```m2
i1 : last commandLine

o1 = --no-readline

i2 : f = x -> x^2

o2 = f

o2 : FunctionClosure

i3 : code f

o3 = stdio:2:6-2:11: --source code:
     f = x -> x^2
```

pzinn:
looking good.
does that work internally by having M2 memorize all past input? does that cost a lot of extra memory?

pzinn:
> I suppose it might add up if you have a lot of input. This has already been the behavior when not using the `--no-readline` option, though.

oh I see, interesting.

d-torrance:
Slightly related issue: `code` only kind of works for functions defined in `currentString`:

```m2
i1 : value "f = x -> x^2
     code f"

o1 = currentString:1:6-1:11: --source code:
     f = x -> x^2

i2 : code f
stdio:3:1:(3): error: no method found for applying separate to:
     argument   :  null (of class Nothing)
```

d-torrance:
I've added an improved error message for the `currentString` situation.  I don't think it's worth trying to cache the code anywhere.

```m2
i1 : value "f = x -> x^2"

o1 = f

o1 : FunctionClosure

i2 : code f
stdio:2:1:(3): error: code no longer available
```

PR #3029, opened 2023-12-05
Remove comment delimiters around net(Function) and friends

d-torrance:
As suggested by @mahrud in https://github.com/Macaulay2/M2/pull/3026#issuecomment-1841326040.  This fixes syntax highlighting in Emacs (and here on GitHub, for that matter) when functions without global names are printed in various contexts using `silentRobustNet`.

### Before
```m2
i1 : ring lookup(rank, Matrix)
stdio:1:1:(3): error: no method found for applying ring to:
     argument   :  -*Function[/usr/share/Macaulay2/. (of class FunctionClosure)

i2 :
```

(Everything after this is highlighted as a comment.)

### After
```m2
i1 : ring lookup(rank, Matrix)
stdio:1:1:(3): error: no method found for applying ring to:
     argument   :  Function[../src/macaulay2/M2/M2/. (of class FunctionClosure)

i2 :
```


pzinn:
I thought `-*...*-` for the standard syntax for unnamed functions? if we change it here, we'll have to change it in a lot more places...

d-torrance:
> I thought `-*...*-` was the standard syntax for unnamed functions? if we change it here, we'll have to change it in a lot more places...

It is the standard syntax, but it's been changed a couple times before.  Looking at the git history, it was previously `--Function[...]--` and then `(*Function[...]*)`.  I think the only place it's hardcoded outside the definition of the corresponding `net` methods is the one place in `document.m2` I've already updated.

pzinn:
what about `hypertext Function`? should it be changed too?

d-torrance:
> what about `hypertext Function`? should it be changed too?

Ah, I missed that one -- thanks!!  I doubt it happens very frequently, but it's possible to get the syntax highlighting issue:

```m2
i1 : debug Core; silentRobustNetWithClass(60, 5, 3, hypertext lookup(rank, Matrix))

o2 = -*FunctionClosure[/usr/share/Macaulay2/Core. (of class SAMP)

i3 : 
```

pzinn:
I'm not against this change, but it does create ambiguities:
```
i1 : code(value,Pseudocode)

o1 = -- code for method: value(Pseudocode)
     function Function:  source code not available
```
this is not so good: when I read it, I believe it's talking about the *Type* Function, not about an instance of that type.
usually, say with BasicList, the ambiguity is removed because it's followed by `{...}`. here, if the code location exists, then it's also fine: `Function[xxx.m2:1:2]` is unambiguous. but if the location is unavailable as in the example, it's not so great.
By the way, the web version (based on `hypertext`) gives slightly more information, and maybe the `net` output should too:
Screenshot from 2023-12-06 13-22-56

d-torrance:
Great points!  I've updated it to match the `hypertext` behavior and use the class name, and now proposing an empty `[]` after the class name for functions with no location.  So for example:

```m2
i1 : code(value, Pseudocode)

o1 = -- code for method: value(Pseudocode)
     function CompiledFunction[]: source code not available
```

pzinn:
looks good to me. just need to update `hypertext Function` to do the `[]` too, and I'll be happy :-P

d-torrance:
>  just need to update `hypertext Function` to do the `[]` too, and I'll be happy :-P

Done!  This also fixes a minor display bug when `hypertext(Function)` was run in the terminal:

### Before
```m2
i1 : hypertext lookup(value, Pseudocode)

o1 = -*CompiledFunctionnull*-
```

### After
```m2
i1 : hypertext lookup(value, Pseudocode)

o1 = CompiledFunction[]
```

pzinn:
oops! good catch

PR #3022, opened 2023-11-30
new MatrixSchubert package

aalmousa:
New package for investigating matrix Schubert varieties and alternating sign matrix varieties by Ayah Almousa, Sean Grate, Daoji Huang, Patricia Klein, Adam LaClair, Yuyuan Luo, and Joe McDonough.

PR #3034, opened 2023-12-09
Add "ForeignStructType ForeignObject" method

d-torrance:
When we added `ForeignStructType HashTable`, it overwrote the `ForeignType ForeignObject` method since `ForeignObject` inherits from `HashTable`.  This caused problems when dealing with foreign struct objects (for example, an `RInterface` test was failing).

---

This fixes the build error noted in https://github.com/Macaulay2/M2/pull/3022#issuecomment-1848771918.


d-torrance:
@DanGrayson @mikestillman Would either of you be willing to merge this soon to fix the builds on the other PR's?  Thanks!

Issue #3035, opened 2023-12-10
reduce breaks algorithms


mahrud:
Since #3013 was merged several pieces of code I use have all ground to a halt. Previously, a change I introduced in #2766 caused `submatrix` to call `reduce`, which slowed down submatrix when the target of a matrix needed a gb computation, even if the computation was trivial (see #2898). Here's an example:
```m2
debug needsPackage "FGLM";
I = cyclic(7, MonomialOrder=>Lex)
I = ideal I_*_{0..5} -- keep only the homogeneous part
R = ring I
v = map(module I, R^1, 0)
v_{}
```
This used to be instantaneous in 1.21.

Now, this happens in many more cases, even when there are no relations involved (which was the initial reasoning in #3012). These simple operations should all be O(1), but induce a gb computation:
```m2
v + v
I_{0}
I_{0}+I_{1}
map(image gens I, , {{1}, {0}, {0}, {0}, {0}, {0}, {0}})
```

To see that these are all caused by calls to `reduce`, try running this and repeat the commands above:
```m2
debug Core
addHook(ReduceHooks, Strategy => Trivial, (tar, rawF) -> rawF)
```
I've had to add the lines above to my `init.m2`, otherwise I can't do anything.

---

Tangentially related to this, here's another method that should be trivial, but actually runs an unnecessary Groebner basis on the target:
```m2
debug needsPackage "FGLM";
I = cyclic(7, MonomialOrder=>Lex)
I = ideal I_*_{0..5}
f = inducedMap(module I, , gens I)
```

d-torrance:
Yikes!  I'll revert the changes from #3013.

mahrud:
I don't think we should revert either #2766 or #3013 or remove `reduce` from more places (at least not yet), rather I think we need to discuss a convention on whether matrices should be reduced or not.

I can think of several options:
1. always reduce, but expose more raw functions for specialized algorithms
2. reduce based on a global flag
3. reduce based on the coefficient ring (e.g. over ZZ or fields)
4. reduce based on existence of relations in the target module
5. never reduce, except before checking equality, and perhaps expose `reduce` as a function that one can use to get a normal form for matrices.

jkyang92:
Personally, I'd say reduce only before checking equality and expose a `reduce` function. Obviously in that circumstance the reduction should be cached. 

[...]

PR #3036, opened 2023-12-10
Revert "reduce" changes in #3013

d-torrance:
This restores the matrix addition/subtraction/negation behavior so we don't get the huge slowdowns reported in #3035.


d-torrance:
Closing pending discussions in #3035.

PR #3037, opened 2023-12-12
Improve net(Iterator) output for return values of iterator(RObject) 

d-torrance:
`net(Iterator)` prints whatever appears first inside `frames`, so we use the `RObject` itself rather than the integer code returned by `TYPEOF`.

### Before
```m2
i2 : iterator RObject {1, 2, 3}

o2 = iterator 13

o2 : Iterator
```

### After
```m2
i2 : iterator RObject {1, 2, 3}

o2 = iterator([1] 1 2 3)

o2 : Iterator
```

### DIFF
--- a/M2/Macaulay2/packages/RInterface.m2
+++ b/M2/Macaulay2/packages/RInterface.m2
@@ -1,7 +1,7 @@
 newPackage("RInterface",
     Headline => "interface to R for statistical computing",
     Version => "0.1",
-    Date => "November 18, 2023",
+    Date => "December 14, 2023",
     Authors => {{
 	    Name => "Doug Torrance",
 	    Email => "dtorrance@piedmont.edu",
--- a/M2/Macaulay2/packages/RInterface.m2
+++ b/M2/Macaulay2/packages/RInterface.m2
@@ -44,7 +44,7 @@ export {
 --------------------
 -- initialization --
 --------------------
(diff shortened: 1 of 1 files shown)


d-torrance:
I also pushed a fix for a bug I found.  If the `R_HOME` environment variable is already set, then a warning message was getting included in the path to the R shared library and we got an error:

### Before
```m2
i1 : needsPackage "ForeignFunctions"

o1 = ForeignFunctions

o1 : Package

i2 : (foreignFunction("putenv", int, charstar))("R_HOME=foo")

o2 = 0

o2 : ForeignObject of type int32

i3 : getenv "R_HOME"

o3 = foo

i4 : needsPackage "RInterface"
/usr/share/Macaulay2/RInterface.m2:48:6:(2):[11]: error: WARNING: ignoring environment value of R_HOME/usr/lib/R/lib/libR.so: cannot open shared object file: No such file or directory
```

### After
```m2
i1 : needsPackage "ForeignFunctions"

o1 = ForeignFunctions

o1 : Package

i2 : (foreignFunction("putenv", int, charstar))("R_HOME=foo")

o2 = 0

o2 : ForeignObject of type int32

i3 : getenv "R_HOME"

o3 = foo

i4 : needsPackage "RInterface"

o4 = RInterface

o4 : Package
```

PR #2977, opened 2023-11-02
Varieties: new package and documentation

mahrud:
This is on top of #2962.

- git mv m2/varieties.m2 packages/Varieties.m2 (and some docs)
- added new package Varieties
- added Varieties to relevant PackageImports lists
- moved all documentation of varieties and sheaves

PR #3042, opened 2023-12-19
Add degree limit option to Hom

### LINKED ISSUE #3039: Hom trims unnecessarily
The key line in `Hom(M, N)` runs:
```m2
     H := trim kernel (transpose presentation M ** N);
```
I suspect `trim` is there to remove generators that give zero homomorphisms, for example. However, this trim can be very slow without even changing the result. Here is an example:
```m2
R = quotient Grassmannian(1,3, CoefficientRing => ZZ/3);
X = Proj R;
-- blackbox what N is, but it has 97 generators and 32 relations
N = module frobeniusPushforward(1, OO_X); -- rank 81
M = R^{0,-1, 43:-1, 52:-2}; -- rank 97
```
Now compare:
```m2
i13 : elapsedTime H = kernel (transpose presentation M ** N);
 -- 0.0469308 seconds elapsed

i14 : elapsedTime H' = trim H;
 -- 167.708 seconds elapsed

i15 : betti H', betti H

                 0    1            0    1
o15 = (total: 9409 3104, total: 9409 3104)
          -2:   52    .     -2:   52    .
          -1: 2332    .     -1: 2332    .
           0: 4641 1664      0: 4641 1664
           1: 2332 1408      1: 2332 1408
           2:   52   32      2:   52   32

```
Which means that there was nothing to trim, but nearly 3 minutes (almost 100% of the computation) was a time waste.

I can think of two solutions:
1. add an optional argument `MinimalGenerators` for `Hom`, perhaps set to false by default, which determines whether to trim the result;
2. never trim the result and leave it to the user to do that.

---

As an aside, we should do this anyway:
- fix `prune` and `trim` to keep cached `homomorphism`, `formation`, `components`, etc. if present (what other cached results are important to keep?)

mahrud:
On top of #2977:
- applied a patch from Complexes to Hom(Module,Module)
- added options for Hom and End
- fixed a broken test in VersalDeformations
- adjusted ComputationsBook/geometry to trim Hom's
- added options for adjoint and homomorphism', since they call Hom
- adjusted an affected test in CompleteIntersectionResolutions
- added MinimalGenerators as option for dual of matrices and complexes
- added new algorithm for Hom(Module,Module)
- changed documentation of optional argument symbols
- moved Hom functor and friends to Hom.m2
- added MinimalGenerators as option for compose
- reworded an error message
- updated documentation of Hom and friends
- fixed documentation of options

Also closes #3039.

cc: @devlinmallory @eisenbud


DanGrayson:
Could you please resolve the merge conflicts, Mahrud?

mahrud:
One important change to keep in mind: with this PR Hom is not trimmed by default, but the option `MinimalGenerators => true` will trim it.

PR #3051, opened 2023-12-28
Use "standalone" class instead of "article" in showTex

d-torrance:
This way, the size of the document will match the size of the object we're viewing.

For example, consider the following code:

```m2
i1 : R = QQ[x,y,z,w]

o1 = R

o1 : PolynomialRing

i2 : showTex res monomialCurveIdeal(R, {1, 2, 3})
+ cd /tmp/M2-720210-0/0/
+ pdflatex -interaction=batchmode show
This is pdfTeX, Version 3.141592653-2.6-1.40.25 (TeX Live 2023/Debian) (preloaded format=pdflatex)
 restricted \write18 enabled.
entering extended mode
```

### Before
image

### After
image

### DIFF
--- a/M2/Macaulay2/m2/latex.m2
+++ b/M2/Macaulay2/m2/latex.m2
@@ -237,7 +237,7 @@ tex MENU := x -> tex drop(redoMENU x, 1)
 -----------------------------------------------------------------------------
 
 -- TODO: incorporate this with packages/Style/M2book.tex.in
-TeXclass := "\\documentclass{article}"
+TeXclass := "\\documentclass{standalone}"
 TeXpackages := {"amsmath", "amssymb"}
 TeXtemplate := src -> concatenate( TeXclass,                newline,
     apply(TeXpackages, pkg -> "\\usepackage{" | pkg | "}"), newline,


DanGrayson:
What's the advantage?  Centering the display seems harmless.

d-torrance:
With the current behavior, we get a large sheet-of-paper-sized pdf file, complete with a page number at the bottom.  That seems like a lot of wasted space for viewing a single Macaulay2 object, especially something small like a polynomial.

I'm thinking of the case where the user might want to convert the pdf into an image for embedding into some other document.   With the current behavior. they'd need to go through an extra step of cropping it.

Another advantage is we're not limited by the width of the paper.  Consider the following example:

```m2
i1 : R = QQ[x_1..x_5]

o1 = R

o1 : PolynomialRing

i2 : showTex koszul vars R
+ cd /tmp/M2-952702-0/0/
+ pdflatex -interaction=batchmode show
This is pdfTeX, Version 3.141592653-2.6-1.40.25 (TeX Live 2023/Debian) (preloaded format=pdflatex)
 restricted \write18 enabled.
entering extended mode
```
### Before
image

### After
image

PR #3053, opened 2023-12-29
Don't build autotools during GitHub workflow

d-torrance:
autoconf, automake, and libtool are already available on the system, so there's no need to build them from scratch.

The one wrinkle is that on macOS, `/usr/bin/libtool` isn't GNU libtool, but if we add the correct directory to the `PATH` environment variable as outlined on the [brew libtool page](https://formulae.brew.sh/formula/libtool), then we're all set.

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -110,6 +110,7 @@ jobs:
                echo `brew --prefix make  `/libexec/gnubin >> $GITHUB_PATH
                echo `brew --prefix llvm`/bin              >> $GITHUB_PATH
                PATH=`brew --prefix llvm`/bin:$PATH
+               echo `brew --prefix libtool`/libexec/gnubin >> $GITHUB_PATH
           fi
           # Necessary for clang to find the right libomp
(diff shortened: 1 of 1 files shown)


mahrud:
Here's a question: why even have the scripts for building these three libraries? (And the same goes for CMake, why even maintain all the code for building external libraries?)

DanGrayson:
> Here's a question: why even have the scripts for building these three libraries? (And the same goes for CMake, why even maintain all the code for building external libraries?)

I must have run into a Linux system where they weren't available.

DanGrayson:
@d-torrance -- if you remove line 163, then the build will fail on systems where those utilities are not available.

mahrud:
> I must have run into a Linux system where they weren't available.

Every software has requirements, and its the responsibility of users to install those requirements. If autotools were some esoteric software it could make sense, but why doesn't it suffice to simply link to download autotools in the INSTALL file?

It makes even less sense to test that autotools compiles fine on Github's virtual machines every night.

d-torrance:
> if you remove line 163, then the build will fail on systems where those utilities are not available.

This PR only deals with the GitHub builds, where autotools is available.

Issue #2956, opened 2023-10-08
Some Github builds hitting 6-hour time limit


d-torrance:
There is a 6-hour time limit on Github actions (see https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits).  I'm not sure what changed, but we've hit this a few times recently (https://github.com/Macaulay2/M2/actions/runs/6446051912, https://github.com/Macaulay2/M2/actions/runs/6415396018, and https://github.com/Macaulay2/M2/actions/runs/6389495572, for example).  It used to usually run in about 4.5 hours I think.

One possible fix might be to break the builds up into multiple "jobs" which hopefully would each run for far less than six hours, and then assign a dependency relationship between them.

mahrud:
Seems like autotools takes about twice as long as cmake. Why not stop running the tests on the autotools build and just run them with cmake?

mahrud:
Seems to have fixed itself. Maybe the Github cache had expired.

mahrud:
Here are timings from a recent build:
- Without tests:
  - [cmake-macos-12-clang-15](https://github.com/Macaulay2/M2/actions/runs/7437548367/job/20235325223#logs), 1h 32m 20s
  - [autotools-macos-12-default](https://github.com/Macaulay2/M2/actions/runs/7437548367/job/20235325100#logs), in 4h 23m 13s
- With tests:
  - [cmake-ubuntu-22.04-clang-15](https://github.com/Macaulay2/M2/actions/runs/7437548367/job/20235325165#logs), in 2h 12m 42s
  - [autotools-ubuntu-22.04-default](https://github.com/Macaulay2/M2/actions/runs/7437548367/job/20235325044#logs), in 3h 34m 29s

Is the cmake build not running some test, or is autotools just significantly slower?

Also, the main branch has been stuck on the same commit [86d6c70](https://github.com/Macaulay2/M2/commit/86d6c70f8746e6b13d95c53ad20d880e0d60a8dc) since **Oct 26** and running the same actions over and over again, and failing it for at least a couple of months. @DanGrayson do you want to remove the nightly build if it's being ignored anyway?

mikestillman:
@DanGrayson Can you replace the main/master branch with development?  Or should I do that?

mikestillman:
@d-torrance @DanGrayson Do you see if there are any obvious speedups for the autotools build?

d-torrance:
[...]

PR #3060, opened 2024-01-07
Attempt at speeding up GitHub autotools builds

d-torrance:
Trying out my ideas from https://github.com/Macaulay2/M2/issues/2956#issuecomment-1880196649.

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -169,7 +169,7 @@ jobs:
       - name: Build Macaulay2 using Make
         if: matrix.build-system == 'autotools'
         run: |
-          make
+          make -j$(nproc 2>/dev/null || sysctl -n hw.logicalcpu)
 
 # ----------------------
 #   Run tests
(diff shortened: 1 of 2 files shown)


d-torrance:
There's some marginal speedup between https://github.com/Macaulay2/M2/actions/runs/7429147236 (the builds from #3058) and the builds for this PR (https://github.com/Macaulay2/M2/actions/runs/7441156789).  4h 12 m -> 3h 58 m on macOS and 3 h 30 m -> 3 h 26 m on Ubuntu.  I think this is probably just from the verbose logging.  Looking at the logs, it doesn't look like anything is being build in parallel.  It might just be the the Makefiles are set up to build the package documentation sequentially, so specifying `-j` might not do anything.

d-torrance:
Ah!  Most of the Makefiles have the special `.NOTPARALLEL` target, which is why `make -j` isn't doing anything.

d-torrance:
Well that definitely did the trick!  The macOS build is down to 1h 45m and the Ubuntu build down to 2h 20m, comparable to the corresponding cmake builds.

PR #3071, opened 2024-01-15
Adding new SheafMap type in Varieties

mahrud:
This is a draft PR to test adding the SheafMap type in the Varieties package, which has been developed with @kellerlv @ritvikrr @ggsmith @Devlin-Mallory and @craicu at the AIM workshop in September 2023.

Also included in this PR:
- improved `BinaryPowerMethod` to also support symmetric powers
- direct sum is now cached in the youngest member (patch from Complexes)
- pullback and pushout for morphisms of modules and sheaves
- a bug in truncation of matrices is fixed

Other side effects:
- Truncations is now preloaded (dependency of Varieties)
- Somehow the order of loading of Elimination has changed, which caused issues in some packages, and the solution was to have Elimination export `resultant` and `discriminant` and `sylvesterMatrix` for other packages to use


ggsmith:
Can I have a couple days to review the details?

mahrud:
No rush at all! I just wanted to make sure nothing unexpected fails, since I wasn't sure how much time I'd have the next few weeks.

mahrud:
I closed this PR because of Greg's objection to Complexes being loaded before it is ready. However, I found that the only reason Complexes was preloaded was because in 77aaa94b17749aab88569b0e9dc8ba865673f718 Mike moved `truncate(List, Complex)` to Truncations, which is a dependency of Varieties, so a simpler solution was to revert his commit.

I updated the PR message to reflect this change.

At the moment, the only way to test Varieties with SheafMaps is to build M2 from this branch.

mahrud:
@mikestillman I heard you may be interested in the pullback and pushout method added to Core in this PR:
https://github.com/Macaulay2/M2/blob/e90c9119dc7b7b191cdd56ed2740c29bdd3646e2/M2/Macaulay2/m2/modules2.m2#L355-L376

M2/Macaulay2/packages/Varieties.m2
```diff
@@ -155,7 +209,7 @@ variety Ring        := S -> if S.?variety then S.variety else error "no variety
 variety Ideal       := I -> Proj(ring I/I) -- TODO: should this be saturated?
 variety RingElement := f -> variety ring f -- TODO: should this be V(f) instead?
```
mahrud, line 210: Should this be deleted?

mahrud: @ggsmith I deleted `variety RingElement`, as you suggested.

M2/Macaulay2/packages/Varieties.m2
```diff
@@ -59,6 +94,9 @@ runLengthEncoding := x -> if #x === 0 then x else (
      p := join({0}, select(1 .. #x - 1, i -> x#i =!= x#(i-1)), {#x});
      apply(#p-1, i -> (p#(i+1)-p#i, x#(p#i))))
 
+-- prints the message only if bool is false and debugLevel > 0
+condition = (bool, msg) -> bool or ( if debugLevel > 0 then printerr msg; false )
+
```
ggsmith, line 99: I would prefer to see a more uniform naming convention used for the various unexported error checking subroutines.  For example, a consistent prefix like ```check``` might work well.  In this particular, case it could be ```checkCondition```.

M2/Macaulay2/packages/Varieties.m2
```diff
@@ -151,11 +205,10 @@ isProjective ProjectiveVariety := X -> true
 -- TODO: instead of an error, return Proj R when there is no variety,
 -- then replace Proj ring M in code for sheaf with variety ring M
 variety = method(TypicalValue => Variety)
-variety Ring        := S -> if S.?variety then S.variety else error "no variety associated with ring"
-variety Ideal       := I -> Proj(ring I/I) -- TODO: should this be saturated?
-variety RingElement := f -> variety ring f -- TODO: should this be V(f) instead?
+variety Ring  := S -> if S.?variety then S.variety else error "no variety associated with ring"
+variety Ideal := I -> Proj quotient I -- TODO: should this be saturated?
```
ggsmith, line 209: I think that common interpretation for the variety of an ideal depends significantly on whether it represents something in projective space or affine space.  Just looking at the ideal seems to be insufficient information.  Specifically, I would be tempted to define ```subvariety(Ideal, Variety)``` rather than simply ```variety Ideal```.

M2/Macaulay2/packages/Varieties.m2
```diff
@@ -151,11 +205,10 @@ isProjective ProjectiveVariety := X -> true
...
 
-sameVariety := Fs -> if not same apply(Fs, variety) then error "expected coherent sheaves on the same variety"
+sameVariety = Fs -> if not same apply(Fs, variety) then error "expected coherent sheaves on the same variety"
```
ggsmith, line 211: How about ```checkSameVariety```?

M2/Macaulay2/packages/Varieties.m2
```diff
@@ -175,6 +228,14 @@ describe ProjectiveVariety := X -> Describe (expression Proj) (expression X.ring
 -- Divisors
 -----------------------------------------------------------------------------
 
+-- overriding the dummy methods defined in Truncations package
+nefCone       Ring := R -> if R.?variety then nefCone       R.variety
+nefGenerators Ring := R -> if R.?variety then nefGenerators R.variety
+
```
ggsmith, line 234: Since a ring doesn't generally have a well-defined notion of a nef cone, this kludge is unfortunate.  Can we not think of a better way to handle this issue?

M2/Macaulay2/packages/Varieties.m2
```diff
@@ -223,6 +287,25 @@ OO = new ScriptedFunctor from {
 OO.texMath = ///{\mathcal O}///
```
ggsmith, line 287: Why does ```sheaf Ring``` assume that one wants to work with a projective variety?  Will caching solve this issue (couldn't one want to work with both the affine variety and the projective variety associated to one ring)?

M2/Macaulay2/packages/Varieties.m2
```diff
@@ -223,6 +287,25 @@ OO = new ScriptedFunctor from {
...
+    and condition(
+	instance(F.variety, Variety) and
+	instance(F.module, Module)   and
+	instance(F.cache, CacheTable),
+	"the hash table does not have the expected values")
+    -- mathematical checks
+    and condition(ring M === ring X,
+	"underlying module and variety do not have the same ring")
+    and condition(not isProjective X or isHomogeneous M,
+	"underlying module of coherent sheaf on a projective variety should be homogeneous")
```
ggsmith, line 306: Is the ```isHomogeneous M``` sufficient?  What sort of gradings do/should we allow the ```Proj``` method to handle?

M2/Macaulay2/packages/Varieties.m2
```diff
@@ -382,40 +482,63 @@ degreeList := M -> (
...
+    -- breaks the test added in 975d780470.
+    -- However, we need to keep the information
+    -- cached in M, for instance if M is a Hom module.
+    M := module F;
     if degreeLength A =!= 1 then error "expected degree length 1";
-    -- quotient by H_m^0(M)
-    M := killH0 cokernel presentation module F;
+    -- quotient by HH^0_m(M) to kill the torsion
+    -- TODO: pass the appropriate irrelevant ideal
+    N := killH0 M;
```
ggsmith, line 504: Why is the second ```killH0``` required?

M2/Macaulay2/packages/Varieties.m2
```diff
@@ -464,11 +587,17 @@ cohomology(ZZ, ProjectiveVariety, CoherentSheaf) := Module => opts -> (p, X, F)
...
     cacheHooks(symbol minimalPresentation, F, (minimalPresentation, CoherentSheaf), (opts, F),
-	(opts, F) -> sheaf(F.variety, HH^0 F(>=0)))) -- this is the default algorithm
+	(opts, F) -> (
+	    -- this is the default algorithm
+	    G := sheaf(F.variety, HH^0 F(>=0));
+	    G.cache.pruningMap = sheafMap F.cache.SaturationMap;
+	    G)))
 
 -----------------------------------------------------------------------------
 -- cotangentSheaf, tangentSheaf, and canonicalBundle
```
ggsmith, line 603: Generalizing to weighted projective spaces introduces several complications.  For example, does one have to deal with the difference between the class group and the Picard group?  Should one work with the stack or the scheme?

M2/Macaulay2/packages/Varieties.m2
```diff
@@ -647,6 +750,12 @@ euler ProjectiveVariety := X -> (
      d := dim X;
```
ggsmith, line 750: The comment with my obsolete email address should be removed.

M2/Macaulay2/packages/Varieties/SheafMaps.m2
```diff
@@ -0,0 +1,890 @@
...
+	instance(f.map, Matrix) and
+	instance(f.degree, ZZ),
+	"the hash table does not have the expected values")
+    -- mathematical checks
+    and condition(ring f === ring X,
+	"underlying matrix and variety do not have the same ring")
+    and condition(same {X, variety F, variety G},
+	"underlying variety does not match that of the source and target")
+    and condition(not isProjective X or isHomogeneous matrix f,
+	"underlying matrix of a map of coherent sheaf on a projective variety should be homogeneous")
```
ggsmith, line 64: Again, is checking ```isHomogeneous``` sufficient.  Which gradings are allowed for projective varieties?

M2/Macaulay2/packages/Varieties/SheafMaps.m2
```diff
@@ -0,0 +1,890 @@
...
+	"expected the matrix to induce a map between a truncation of the underlying modules")
+    )
+
+-- basic methods
+source  SheafMap := CoherentSheaf => f -> f.source
+target  SheafMap := CoherentSheaf => f -> f.target
+variety SheafMap := Variety       => f -> f.source.variety
+ring    SheafMap := Ring          => f -> f.map.ring
+matrix  SheafMap := Matrix => opts -> f -> f.map
+degree  SheafMap := ZZ => f -> f.degree
```
ggsmith, line 81: Calling this numerical invariant of a ```SheafMap``` seems problematic.  As I understand it, it is NOT the degree of the map rather the degree at which the source module was truncated to defined the morphism.

M2/Macaulay2/packages/Varieties/SheafMaps.m2
```diff
@@ -0,0 +1,890 @@
...
+-- isLiftable, lift
+-----------------------------------------------------------------------------
+--general method: checks whether a map phi is in the image of
+--the map Hom(eta,target phi)
+isLiftable = method()
+isLiftable(Matrix, Matrix) := (phi, eta) -> (
+    newPhi := homomorphism' phi;
+    deta := Hom(eta, target phi);
+    (image newPhi / coker deta) == 0)
+
```
ggsmith, line 127: The name ```deta``` seems strange.  What does it mean?

M2/Macaulay2/packages/Varieties.m2
mahrud: I agree that a uniform name is better. In this particular case I'd rather the name of this function be short and hopefully soon it should be moved to Core for use by other packages, so I'll change it to `assert'` (the point being that `assert` throws an error but `assert'` prints a warning if `debugLevel` is high)

M2/Macaulay2/packages/Varieties.m2
mahrud: I'll add a comment to this effect, but I won't change this now because this was simply ported over from old `varieties.m2` and I think we should discuss it more before introducing a potentially breaking change.

M2/Macaulay2/packages/Varieties.m2
mahrud: I agree, and in fact have added a comment about this above that method. I think we should discuss this, and what should `variety Ring` do, at the next meeting.

M2/Macaulay2/packages/Varieties.m2
mahrud: I believe so? Do you have an example in mind?
It's certainly possible that we need more assertions for well-defined-ness of coherent sheaves though.

M2/Macaulay2/packages/Varieties.m2
mahrud: There is just one. The red one was removed and the green one was added.

M2/Macaulay2/packages/Varieties/SheafMaps.m2
mahrud: That's a fair point. I'll change all uses of `degree psi` to `psi.degree` and comment out this method for now, but perhaps we should discuss this at the next meeting.

mahrud: Ah, actually, this seems to be used quite a while and might cause git conflicts. I agree that we should change this, but let's wait until we meet.

M2/Macaulay2/packages/Varieties/SheafMaps.m2
mahrud: @kellerlv do you remember?

mahrud:
I pushed the changes that were easy to do quickly. As discussed, I'll mark this PR as ready for review, but let me know if you see anything else that needs to be fixed before release.

M2/Macaulay2/packages/Varieties/SheafMaps.m2
```diff
@@ -0,0 +1,890 @@
...
+-- inducedMap(CoherentSheaf, CoherentSheaf, SheafMap)
+
+-----------------------------------------------------------------------------
+-- Direct sums and components
+-----------------------------------------------------------------------------
+--WARNING: the current direct sum of sheaves does not cache components
+SheafMap.directSum = args -> (
+    assert(#args>0);
+    X := variety args#0;
+    if not same apply(args, variety) then error "expected maps of sheaves over the same variety";
```
ggsmith, line 184: Shouldn't ```sameVariety``` be used here?

M2/Macaulay2/packages/Varieties/SheafMaps.m2
```diff
@@ -0,0 +1,890 @@
...
+
+-----------------------------------------------------------------------------
+-- Tensors
+-----------------------------------------------------------------------------
+-- TODO: take care of the case when the rings are different
+tensor(SheafMap, SheafMap) := SheafMap => (phi, psi) -> (
+    map(target phi ** target psi,
+	source phi ** source psi,
+	matrix phi ** matrix psi))
+
```
ggsmith, line 205: Does one really not need to consider the truncation degrees when forming tensor products of ```SheafMaps```?

M2/Macaulay2/packages/Varieties/SheafMaps.m2
```diff
@@ -0,0 +1,890 @@
...
+-----------------------------------------------------------------------------
+-- Prune
+-----------------------------------------------------------------------------
+
+-- TODO: is there a better way to do this?
+moveToField = f -> (
+    kk := coefficientRing ring f;
+    map(kk^(numrows f), kk^(numcols f), sub(cover f, kk)))
+
```
ggsmith, line 358: The unexported function should be changed to allow for other coefficient rings.

M2/Macaulay2/packages/Varieties/SheafMaps.m2
```diff
@@ -0,0 +1,890 @@
...
+cohomology(ZZ,                    SheafMap) := Matrix => opts -> (p,    f) -> cohomology(p, variety f, f, opts)
+cohomology(ZZ, ProjectiveVariety, SheafMap) := Matrix => opts -> (p, X, f) -> (
+    -- TODO: need to base change to the base field
+    if p == 0 then moveToField basis(0, matrix prune f) else (
```
ggsmith, line 362: Why do we need the underlying ring to be a field?

M2/Macaulay2/packages/Varieties/SheafMaps.m2
```diff
@@ -0,0 +1,890 @@
...
+    G := directSum apply(toList v, j -> F.cache.components#j);
+    map(F, G, (cover module F)_v))
+
+CoherentSheaf ^ Array := SheafMap => (F, v) -> (
+    v = trans(F,v);
+    G := directSum apply(toList v, j -> F.cache.components#j);
+    map(G, F, (cover module F)^v))
+
+- SheafMap := phi -> map(target phi, source phi, -matrix phi)
+
```
ggsmith, line 540: Should we allow multiplication by any element of the underlying ring?  What about addition when the source and targets agree?

PR #3074, opened 2024-01-16
Update M2-emacs commit

d-torrance:
Several changes were recently proposed and merged in the M2-emacs repository:

* Macaulay2/M2-emacs#50
* Macaulay2/M2-emacs#51
* Macaulay2/M2-emacs#52
* Macaulay2/M2-emacs#53
* Macaulay2/M2-emacs#54

The major changes are:

* There's no more special M2-specific function for <kbd>Return</kbd> in `M2-mode`, which means that we play more nicely with things like[ `electric-indent-mode`](https://www.gnu.org/software/emacs/manual/html_node/emacs/Indent-Convenience.html).
* <kbd>Tab</kbd> behaves more nicely in regards to whitespace in`M2-mode`.  For example, previously pressing <kbd>Tab</kbd> a bunch of times would give you something like `    \t    \t    \t` (i.e., alternating between spaces and tabs), but now it will be more like `\t\t\t\t    ` (i.e., all tabs and then maybe some spaces), which is consistent with most other major modes.
* <kbd>Return</kbd> and <kbd>F11</kbd> are now bound to the same function in `M2-comint-mode`.

I've been running Macaulay2 with these changes for a month or so and haven't noticed anything crazy, but I'd definitely appreciate any feedback before the 1.23 release!


d-torrance:
> Do you want to wait until the highlighting word list has been updated with the new packages and all before updating the submodule hash?

I was thinking that if we update it now, then there's a higher probability of someone catching some bug or unexpected behavior that I introduced before the release.  We can update the symbols in a later PR right before the release.

PR #3044, opened 2023-12-19
Experiment: add AtomicInteger type to top-level M2

d-torrance:
I played around with wrapping C's `atomic_int` type in Macaulay2 to possibly help write thread-safe top-level code, and figured I'd share it for comments.

Here's an example:

```m2
i1 : allowableThreads = maxAllowableThreads

o1 = 12

i2 : x = 0

o2 = 0

i3 : scan(1000, i -> schedule(() -> x = x + 1))

i4 : x

o4 = 895

i5 : y = atomicInteger 0

o5 = 0

o5 : AtomicInteger

i6 : scan(1000, i -> schedule(() -> atomicFetchAdd(y, 1)))

i7 : y

o7 = 1000

o7 : AtomicInteger
```

Issue #1596, opened 2020-11-16
new assignment operator


DanGrayson:
In discussions with Mike, it has occurred to us that it might be nice to introduce the syntax
```m2
X # key ?= foo
```
to denote an expression whose value is `X#key` if the mutable hash table (or mutable list) X has a value stored under that key, and otherwise evaluates the code `foo`, stores the resulting value under that key, and returns that value.

The current alternative coding is `if X#?key then X#key else X#key = foo`, which is more verbose.

@mikestillman

mahrud:
This would be really handy for caching things if `foo` in the example above was lazily computed.

DanGrayson:
Right -- that's what "and otherwise evaluates the code foo" means.

mahrud:
Here's an alternative syntax suggestion. We currently have global assignment with `=` and local declaration with `:=` for variables, but for hash tables only `=` is defined:
```m2
i1 : X = new MutableHashTable from {};

i2 : X.cache := 12
stdio:3:9:(3): error: can't assign a method for this binary operator

i3 : X#"key" := 12
stdio:4:9:(3): error: can't assign a method for this binary operator
```
Why not use `X#key := foo` and `X.cache := foo` for this new operator instead?

d-torrance:
I like this idea a lot, especially for the lazy evaluation!

I think I prefer `?=` over `:=` since it suggests the call to `?#` or `?.`, and also avoids overloading `:=` even more.  (I remember getting confused back in grad school by `:=` meaning both local assignment and method installation.)

Related question: I wonder if adding an `?_` operator would be helpful for user-defined classes?  Then potentially this proposed operator could work there as well.

d-torrance:
Related: https://en.wikipedia.org/wiki/Null_coalescing_operator

It seems some other languages use `??=` for this.  It may also be useful to add `??`, i.e., `x ?? y` returns `x` unless it's null or raises and error, which case we return `y`.

d-torrance:
Adding a null coalescing operator is relatively simple:

```diff
--- a/M2/Macaulay2/d/actors.d
+++ b/M2/Macaulay2/d/actors.d
@@ -1164,6 +1164,14 @@ subvalueQ(lhs:Code,rhs:Code):Expr := (
          else subvalueQ(left,right)));
 setup(SharpQuestionS,subvalueQ);
 
+nullCoalescion(lhs:Code,rhs:Code):Expr := (
+     left := eval(lhs);
+     when left
[...]

PR #3079, opened 2024-01-20
Add augmented assignment operators to Macaulay2 language

d-torrance:
For most flexible binary operators in Macaulay2, we add a corresponding [augmented assignment](https://en.wikipedia.org/wiki/Augmented_assignment) operator.  (This is something that would be very useful if/when we add atomic integers at top-level -- see #3044.)

Here's the proposed documentation for more information:

```m2
i1 : help "augmented assignment"

o1 = augmented assignment
     ********************

     Most binary "operators" have an associated augmented assignment (see
     https://en.wikipedia.org/wiki/augmented_assignment ) operator that modifies
     the given object using the corresponding binary operator.  Essentially, x
     OP= y is equivalent to x = x OP y.

     +---------------+
     |  i1 : x = 2   |
     |               |
     |  o1 = 2       |
     +---------------+
     |  i2 : x += 3  |
     |               |
     |  o2 = 5       |
     +---------------+
     |  i3 : x -= 4  |
     |               |
     |  o3 = 1       |
     +---------------+
     |  i4 : x *= 5  |
     |               |
     |  o4 = 5       |
     +---------------+
     |  i5 : x /= 6  |
     |               |
     |       5       |
     |  o5 = -       |
     |       6       |
     |               |
     |  o5 : QQ      |
     +---------------+

     The following augmented assignment operators are supported.

     +---------------------------------------------------------------------------------+
     |  i6 : importFrom(Core, "augmentedAssignmentOperators");                         |
     +---------------------------------------------------------------------------------+
     |  i7 : augmentedAssignmentOperators                                              |
     |                                                                                 |
     |  o7 = (<<=, >>=, ||=, <==>=, |-=, ==>=, ===>=, |=, ^^=, &=, ..=, ..<=, -=, +=,  |
     |       ------------------------------------------------------------------------  |
     |       ++=, **=, *=, \\=, /=, \=, %=, //=, @=, @@=, ^=, _=, ^**=)                |
     |                                                                                 |
     |  o7 : Sequence                                                                  |
     +---------------------------------------------------------------------------------+

     See also
     ========

       * "installing augmented assignment methods"

i2 :        * "installing augmented assignment methods"

o2 = installing augmented assignment methods
     ***************************************

     In most cases, the default behavior of "augmented assignment" gives the
     desired result.  But in some situations, it may be useful to override this
     behavior and install a custom method for a given type.

     Consider the following example.

     +-------------------------------------------------------+
     |  i1 : Foo = new SelfInitializingType of MutableList;  |
     +-------------------------------------------------------+
     |  i2 : net Foo := x -> net x#0;                        |
     +-------------------------------------------------------+
     |  i3 : Foo + Foo := (x, y) -> Foo {x#0 + y#0};         |
     +-------------------------------------------------------+
     |  i4 : x = Foo {1}                                     |
     |                                                       |
     |  o4 = 1                                               |
     |                                                       |
     |  o4 : Foo                                             |
     +-------------------------------------------------------+
     |  i5 : y = Foo {2}                                     |
     |                                                       |
     |  o5 = 2                                               |
     |                                                       |
     |  o5 : Foo                                             |
     +-------------------------------------------------------+
     |  i6 : x += y                                          |
     |                                                       |
     |  o6 = 3                                               |
     |                                                       |
     |  o6 : Foo                                             |
     +-------------------------------------------------------+

     Note that an intermediate Foo object was created and then assigned to x.
     Instead, it would be more efficient if x was modified directly.

     The first two lines below do exactly the same thing; the second line is
     syntactic sugar for the first.

     +------------------------------------------------------------------+
     |  i7 : installMethod(symbol +=, Foo, (x, y) -> (x#0 += y#0; x));  |
     +------------------------------------------------------------------+
     |  i8 : Foo += (x, y) -> (x#0 += y#0; x);                          |
     +------------------------------------------------------------------+
     |  i9 : x += y                                                     |
     |                                                                  |
     |  o9 = 5                                                          |
     |                                                                  |
     |  o9 : Foo                                                        |
     +------------------------------------------------------------------+

     In some cases, it may be useful to fall back on the default behavior of the
     given operator.  When this is desired, the installed method should return
     the "Default" symbol.

     +-------------------------------------------------------+
     |  i10 : Bar = new SelfInitializingType of List;        |
     +-------------------------------------------------------+
     |  i11 : net Bar := x -> net x#0#0;                     |
     +-------------------------------------------------------+
     |  i12 : Bar * Bar := (x, y) -> Bar {{x#0#0 * y#0#0}};  |
     +-------------------------------------------------------+
     |  i13 : Bar *= (x, y) -> if isMutable x#0 then (       |
     |            print "using custom method";               |
     |            x#0#0 *= y#0#0; x) else Default;           |
     +-------------------------------------------------------+
     |  i14 : x = Bar {new MutableList from {3}}             |
     |                                                       |
     |  o14 = 3                                              |
     |                                                       |
     |  o14 : Bar                                            |
     +-------------------------------------------------------+
     |  i15 : y = Bar {{4}}                                  |
     |                                                       |
     |  o15 = 4                                              |
     |                                                       |
     |  o15 : Bar                                            |
     +-------------------------------------------------------+
     |  i16 : x *= y                                         |
     |  using custom method                                  |
     |                                                       |
     |  o16 = 12                                             |
     |                                                       |
     |  o16 : Bar                                            |
     +-------------------------------------------------------+
     |  i17 : y *= x                                         |
     |                                                       |
     |  o17 = 48                                             |
     |                                                       |
     |  o17 : Bar                                            |
     +-------------------------------------------------------+

     See also
     ========

       * "augmented assignment"
       * "installing methods"
```
Note that not every flexible binary operator is supported, e.g., it wouldn't make sense to have an augmented assignment operator for `==` since `===` is already in use.

As proof of concept, we update a few places in Core with `i = i + 1` to use `i += 1` and also add augmented operator support to the `Python` package.


mahrud:
By the way, have you seen #1596?

d-torrance:
> By the way, have you seen #1596?

Not until you pointed it out, but I like the idea a lot!  The implementation would be pretty similar to this PR, I think.

mahrud:
Yeah, I figured it might be easy to do if you're already familiar with that part of the code.

PR #3080, opened 2024-01-20
Fix SubalgebraBases documentation typos

d-torrance:
When #3073 was opened, it was originally targeting the `master` branch, and so the `codespell` checks were never run since they're only run on PR's to the `development` branch:

https://github.com/Macaulay2/M2/blob/776fd94fa0060f047234ac837464962d3581609f/.github/workflows/lint.yml#L4

But now that it's been merged, other PR's will fail (like #3079) due to some typos.  This PR fixes those typos.

(Why is `master` the default branch when we want everyone to submit to `development`?)

### DIFF
--- a/M2/Macaulay2/packages/SubalgebraBases/Old-SubalgebraBases/Archive/oldDocumentation.m2
+++ b/M2/Macaulay2/packages/SubalgebraBases/Old-SubalgebraBases/Archive/oldDocumentation.m2
@@ -131,7 +131,7 @@ doc ///
 	  maps p_9 to t_1^2*t_2
         Text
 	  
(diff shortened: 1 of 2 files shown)


d-torrance:
> (Why is `master` the default branch when we want everyone to submit to `development`?)

After thinking and googling about this for a bit, I think I agree that we should have `master` as the default branch since there will be more folks cloning than submitting PR's, and we want them to get the more stable branch.  It doesn't appear that GitHub allows the ability to set a distinct "default for PR's" branch, so I guess we're kind of stuck with the current situation of the occasional PR submitted to `master`.

We should probably run `codespell` on those, too -- I'll add another commit to this PR doing that.

Issue #1113, opened 2020-04-30
Removing libatomic_ops


mahrud:
libatomic_ops is a library atomic operations  (i.e. those which cannot be interrupted by another thread) before C11 and C++14 standards. Currently the first line on its [README.md](https://github.com/ivmai/libatomic_ops/blob/master/README.md) says:
> IN NEW CODE, PLEASE USE C11 OR C++14 STANDARD ATOMICS INSTEAD OF THIS PACKAGE.

This issue is for using either the [C++14 atomic](https://en.cppreference.com/w/cpp/atomic) or the [C11 atomic](https://en.cppreference.com/w/c/atomic) operations. These operations are part of the language, so don't require an extra library to be installed. They work much better with the C++ [thread](https://en.cppreference.com/w/cpp/thread) library and are also much better documented than [libatomic_ops](https://www.hpl.hp.com/research/linux/atomic_ops/README.txt).

libatomic_ops mostly used in atomic.d and system, but the uses propagate from there. I tried refactoring atomic.d to use C11 atomic, but ran into issues because atomic.d is later used in C++, where `_Atomic` type is not defined (and `extern "C"` didn't help). On the other hand, C++14 atomic is not usable in C.

Not sure how to proceed, but moving to C11 & C++14 would definitely make multithreading easier.

mahrud:
On a second look, maybe this isn't impossible and the C and C++ atomic libraries are compatible. Here is a test: 
- c11.c:
```
#include <pthread.h>
#include <stdatomic.h>
#include <stdio.h>

atomic_int acnt;
int cnt;

void adding()
{
  for(int i=0; i<10000; i++)
    {
      acnt++;
      cnt++;
    }
  pthread_exit(NULL);
}
```
- cxx14.cpp:
```
#include <vector>
#include <atomic>
#include <thread>

using namespace std;

extern "C" void adding();
extern "C" atomic_int acnt;
extern "C" int cnt;

int main()
{
  vector<thread> v;

  for(size_t i {0}; i < 10; ++i) {
    v.emplace_back(adding);
  }

  for(auto &t : v) { t.join(); }

  printf("the value of acnt is %d\n", acnt.load());
  printf("the value of cnt is %d\n", cnt);

  return 0;
}
```
We should be able to compile separately and link together like so:
```
  gcc -c c11.c -o c11.o --std=c11
  g++ -c cxx14.cpp -o cxx14.o --std=c++14
  g++ cxx14.o c11.o -pthread
[...]

PR #3082, opened 2024-01-22
Replace atomic_ops with std::atomic

d-torrance:
atomic_ops has been deprecated since C11/C++14.  See the message at the top of https://github.com/ivmai/libatomic_ops/blob/master/README.md:

> IN NEW CODE, PLEASE USE C11 OR C++14 STANDARD ATOMICS INSTEAD OF THE CORE LIBRARY IN THIS PACKAGE.

Closes: #1113


d-torrance:
Clang is much pickier about casting `atomic_int`'s to `int`'s than GCC is, but we have finally have all the builds working!  I'm not sure how well things that use atomics are covered in the tests, so hopefully nothing is broken...

mahrud:
Are atomic ints in top-level next?

d-torrance:
> Are atomic ints in top-level next?

Yup!

PR #3085, opened 2024-01-24
Lazily evaluate the right-hand side in augmented assignment

d-torrance:
This is a quick followup to #3079.  We hold off on evaluating `y` in something like `x += y` until we absolutely know we need it.  With this change, `??=` would work out of the box if we were to add `??` (or whatever) as a [null coalescing operator](https://en.wikipedia.org/wiki/Null_coalescing_operator) -- see #1596.


d-torrance:
Wait, what I said above isn't quite correct -- `??=` wouldn't work out of the box for `x ??= y` if evaluating `x` raises an error (e.g., a nonexistent key in a hash table)., but it *would* work out of the box if `x` is null.  It should be an easy change to make it work for errors, though.

PR #3090, opened 2024-01-26
Show unary methods installed using symbols using "methods"

d-torrance:
I noticed the following unusual behavior:

```m2
i1 : methods NewMethod

o1 = {0 => (NewMethod, ChainComplex)}

o1 : NumberedVerticalList
```

It seems like we should be able to go the other direction and find `NewMethod` in the list of methods for `ChainComplex`:

```m2
i2 : methods ChainComplex

o2 = {0 => ((_, =), ChainComplex, ZZ)                                      }
     {1 => (**, ChainComplex, ChainComplex)                                }
     {2 => (**, ChainComplex, ChainComplexMap)                             }
     .
     .
     .
     {37 => (minimalPresentation, ChainComplex)                            }
     {38 => (net, ChainComplex)                                            }
     {39 => (NewFromMethod, ChainComplex, Resolution)                      }
     {40 => (poincare, ChainComplex)                                       }
     {41 => (poincareN, ChainComplex)                                      }
     .
     .
     .
```

It's not there!  It turns out that unary methods that are stored in types using keys that are non-keyword symbols were being skipped.  Now there are lots of symbols that we do want to skip (like `synonym` and `texMath`), but there a few that probably should be listed.

We whitelist a specific list (`AfterEval`, `AfterNoPrint`, `AfterPrint`, `Format`, `InverseMethod`, `NewMethod`, `Wrap`).  Did I miss any?

### After

```m2
i1 : methods ChainComplex

o1 = {0 => ((_, =), ChainComplex, ZZ)                                      }
     {1 => (**, ChainComplex, ChainComplex)                                }
     {2 => (**, ChainComplex, ChainComplexMap)                             }
     .
     .
     .
     {37 => (minimalPresentation, ChainComplex)                            }
     {38 => (net, ChainComplex)                                            }
     {39 => (NewFromMethod, ChainComplex, Resolution)                      }
     {40 => (NewMethod, ChainComplex)                                      }
     {41 => (poincare, ChainComplex)                                       }
     {42 => (poincareN, ChainComplex)                                      }
     .
     .
     .
```

### DIFF
--- a/M2/Macaulay2/m2/code.m2
+++ b/M2/Macaulay2/m2/code.m2
@@ -152,12 +152,23 @@ thingMethods := (T, F) -> nonnull apply(pairs T, (key, func) -> if instance(func
     if isUnaryAssignmentOperator key and isMember(F,        key) then (key, T) else -- unary assignment method, e.g symbol=
     if instance(key, Sequence)       and isMember(F, splice key) then  key)
 
+-- unary methods that are installed using a symbol as a key
+symbolMethods = set {
+    AfterEval,
+    AfterNoPrint,
+    AfterPrint,
+    Format,
+    InverseMethod,
+    NewMethod,
+    Wrap}
+
 sequenceMethods := (T, F, tallyF) -> nonnull apply(pairs T, (key, func) -> if instance(func, Function) then
     if isBinaryAssignmentOperator key and tallyF <= tally splice  key     then  key     else -- e.g T#((symbol SPACE, symbol=), T, T)
     if  isUnaryAssignmentOperator key and tallyF <= tally splice (key, T) then (key, T) else -- e.g T#(symbol+, symbol=)
     if instance(key, Keyword)         and tallyF <= tally splice (key, T) then (key, T) else -- e.g T#(symbol #)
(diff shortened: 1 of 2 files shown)


d-torrance:
This is a pretty hacky solution. I think I might have a better idea.  Closing.

mahrud:
Looking at this PR again, I think I much prefer this solution.

pzinn:
actually, it would make sense if `NewMethod`, etc, were methods (possibly un-capitalising new). I can't see any situation when one would prefer to call them directly rather than use the normal syntax `new ... from`, but it's probably harmless to make them methods. wouldn't that solve the problem?

d-torrance:
That would certainly work, but I decided to go with simple one-line solution for the time being: https://github.com/Macaulay2/M2/pull/3093/commits/eb17ca16497a9e54998c1043e5ef9f05fa613e7d

mahrud:
You raise a good point in the commit message that if `NewMethod` was a keyword this would have worked already. Since `new` is a keyword, we could also get rid of `NewMethod` and friends and store the methods under `(symbol new, Module)` and similar sequences, then the output of `methods Module` would include `(new, Module)` which is much closer to what users need.

PR #3091, opened 2024-01-26
Add new "MethodSymbol" class for symbols which are used for methods

d-torrance:
This is (I think) an improved version of #3090.  Rather than have some whitelist of symbols that we allow to appear in the output of `methods`, we add a new subclass of `Symbol`, `MethodSymbol`, and just check the class.  `NewMethod` and friends are all elements of this proposed new class, and package authors can add their own.

Here's a simple example:

```m2
i1 : X = new Type of HashTable;

i2 : new MethodSymbol from "foo";

i3 : X.foo = x -> 1;

i4 : X.bar = x -> 2;

i5 : methods X

o5 = {0 => (foo, X)}

o5 : NumberedVerticalList
```

Here is a list of the symbols that I'm proposing be turned into method symbols:

```m2
    Fixed objects of class MethodSymbol :
     =====================================

       * "AfterEval" -- top level method applied after evaluation
       * "AfterNoPrint" -- top level method applied after not printing
       * "AfterPrint" -- top level method applied after printing
       * "BeforePrint" -- top level method applied before printing results
       * "GlobalAssignHook" -- hook for assignment to global variables
       * "GlobalReleaseHook"
       * "InverseMethod" -- compute reciprocals
       * "Jupyter" -- top level printing method used in the Jupyter kernel
       * "NewFromMethod"
       * "NewMethod"
       * "NewOfFromMethod"
       * "NewOfMethod"
       * "NoPrint" -- top level method for non-printing results
       * "Print" -- top level method for printing results
       * "Standard" -- the standard top level printing method
       * "TeXmacs" -- the TeXmacs top level printing method
       * "WebApp" -- the web app top level printing method
       * "Wrap" -- a key for methods for wrapping printed output
```
This is a draft for now until I write some documentation.

I'm not sure whether the `topLevelMode` symbols like `Standard` and `WebApp` belong on this list or not, since they're always paired up inside a list with something else.  But they are used for defining methods, so I suppose it makes sense even if `methods`  can't currently find them.

Comments welcome!

### DIFF
--- a/M2/Macaulay2/d/actors5.d
+++ b/M2/Macaulay2/d/actors5.d
@@ -1873,6 +1873,20 @@ foreach s in syms do storeInHashTable(
 storeE = nullE;
 syms = SymbolSequence();
 
+makeMethodSymbol(e:Expr):Expr := (
+    when e
+    is a:Sequence do (
+	if length(a) == 2 then (
+	    when a.0
+	    is HashTable do (
+		when a.1
+		is s:stringCell do Expr(makeMethodSymbol(s.v))
+		else WrongArgString(2))
+	    else WrongArg(1, "a hash table"))
+	else WrongNumArgs(2))
+    else WrongNumArgs(2));
+installMethod(NewFromS, methodSymbolClass, stringClass, makeMethodSymbol);
+
 export fileDictionaries := newHashTable(mutableHashTableClass,nothingClass);
 setupconst("fileDictionaries",Expr(fileDictionaries));
 
--- a/M2/Macaulay2/d/actors5.d
+++ b/M2/Macaulay2/d/actors5.d
@@ -1688,7 +1688,7 @@ readonlyfilesS := dummySymbol;
 setupconst("minExponent",toExpr(minExponent));
 setupconst("maxExponent",toExpr(maxExponent));
 
-StandardS := makeProtectedSymbolClosure("Standard");
+StandardS := makeMethodSymbol("Standard");
 export StandardE := Expr(StandardS);
 export topLevelMode := Expr(StandardS);
 topLevelModeS := dummySymbol;
--- a/M2/Macaulay2/d/binding.d
+++ b/M2/Macaulay2/d/binding.d
@@ -133,6 +133,7 @@ makeKeyword(w:Word):SymbolClosure := (
      -- keywords differ from symbols in that their initial value is null
      entry := makeEntry(w,dummyPosition,globalDictionary);
      entry.Protected = true;
+     entry.Class = keywordClass;
      sc := SymbolClosure(globalFrame,entry);
      globalFrame.values.(entry.frameindex) = Expr(sc);
      sc);
--- a/M2/Macaulay2/d/binding.d
+++ b/M2/Macaulay2/d/binding.d
@@ -63,7 +63,7 @@ export insert(table:SymbolHashTable, newname:Word, entry:Symbol):Symbol := ( -- 
 threadLocal HadError := false;
 
 
-export makeEntry(word:Word,position:Position,dictionary:Dictionary,thread:bool,locallyCreated:bool):Symbol := (
+export makeEntry(word:Word,position:Position,dictionary:Dictionary,thread:bool):Symbol := (
      while dictionary.Protected do (
 	  if dictionary == dictionary.outerDictionary then (
 	       -- shouldn't occur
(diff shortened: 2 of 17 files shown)


mahrud:
Perhaps an easier fix would be this:
https://github.com/Macaulay2/M2/blob/776fd94fa0060f047234ac837464962d3581609f/M2/Macaulay2/m2/code.m2#L149

d-torrance:
> Perhaps an easier fix would be this:
> 
> https://github.com/Macaulay2/M2/blob/776fd94fa0060f047234ac837464962d3581609f/M2/Macaulay2/m2/code.m2#L149

I think that's probably a great idea.    I actually didn't realize until now that unary methods *weren't* stored this way -- I figured they were just based on the output of `methods` and the syntax used when documenting them.

But isn't this comment referring to traditional methods installed using `installMethod` or `f T := g`, where in particular `f` is a `MethodFunction`?  Since the user doesn't handle manually inserting any key/value pairs into the type in that case, we could probably make that change without breaking any code.

My concern for this PR is "methods" that aren't associated with a `MethodFunction` object, but a `Symbol`.  And in that case, users do manually insert the key/value pairs (except for the `New*Method` family of symbols), so I'd be concerned about changing any behavior with these.  For example, we'd have to change all the cases of `T.AfterPrint = g` to `T#(AfterPrint, T) = g`.

Implementing this comment would probably be a good fix for what I really care about -- `NewFrom` not showing up in `methods`, which means it's not linked in a type's documentation page.  I'll take a lot at it!

Meanwhile, somehow this change caused the `MapleInterface` examples to run.  Very strange!

d-torrance:
I think this is ready now.  The previous builds were failing because I didn't read a comment in the code about where to define new classes in the `d` directory and put the new `MethodSymbol` class before `List`.  This changed the hash of the `List` class and consequently every `List` object, including the hashes of the cached examples.  So it thought it needed to regenerate the `MapleInterface` examples without Maple!

I played around a bit with the `T#(f, T)` idea.  It's easy to change `installMethod` to store unary methods this way, but I got lost trying to figure out how to update the code to find them after that!

pzinn:
I haven't read in detail yet, but: does it really make sense to do this at the "d" level. I feel like a class like this `MethodSymbol` should be defined at top level.

d-torrance:
> I haven't read in detail yet, but: does it really make sense to do this at the "d" level. I feel like a class like this `MethodSymbol` should be defined at top level.

Many of the symbols I'm proposing be switched to the new class (like `NewMethod` and friends) are defined at the d level.

mahrud:
Sorry, I think I'm confused.
1. I guess I don't think of the functions that returned when you call something like `lookup(AfterPrint, Matrix)` as "method"s, rather as just a function stored in a hash table.
2. What's the point of introducing a new type of symbol with only a dozen instances if the main issue can be solved another way? (e.g. simply change `AfterPrint` and other similar symbols to actual methods? A capitalized method isn't the end of the world ...)

d-torrance:
I think the functions that are stored under symbols like `AfterPrint` and `NewFromMethod` are "methods" in a similar sense as functions that are stored under keywords like `+` and `-`.  Even the names of a few of them (`NewFromMethod`, `InverseMethod`, etc.) suggest they're methods.  :)

These symbols are very similar to keywords in that they're used to store some important functions in types, often at the d level before we have such a thing as a `MethodFunction`.  But unlike keywords, they're not explicitly called by the user. 

Even though the `methods` output may look strange to a user since they'll never explicitly call things like `NewFromMethod`, `methods` is used to generate the list of methods for a given type on its documentation page.  In this list, each of these "method symbols" is translated into something that a user could explicitly call.  `(NewMethod, X)` becomes `new X` and `(AfterPrint, X)` becomes `X.AfterPrint`.  This is what I'm most interested in.  Right now, `new X` doesn't appear in this list, and I think it wouldn't be a bad idea if the others did, too, for discoverability.

I agree that it's a pretty substantial change for a small upshot, but I think it might be worth it to improve the documentation.  If we support inheritance, we might as well use it!

mahrud:
I agree that it would be good if `new X` appeared in the documentation, though note that `(NewMethod, ChainComplex)` will be deleted eventually when the `Complexes` package is ready.

However, would you say `{Jupyter, InputPrompt}` is a method because `ZZ#{Jupyter, InputPrompt}` is a function, or is `liftDegree` a method because `(QQ[]).liftDegree` is a function? There are probably worse instances.

Similarly, I don't think `(AfterPrint, Module)` needs to appear in the output of `methods Module`, and I think it shouldn't appear in the documentation under "Methods that use a module" (what would it's usage even say, `Module.AfterPrint`?). I don't think that would be an improvement to the documentation (similarly for `InverseMethod`, `GlobalReleaseHook`, `Print`, `Wrap`, `Jupyter`, etc.), some of which are explicitly defined this way so they aren't listed in the documentation :)

As for discoverability, I think how to use `AfterPrint` should be documented under `AfterPrint` and linked by `net` or the node that teaches how to make a new type.

To be clear, my main objection is not any of the above (I'm fine with fixing `methods` to also list these functions if you want), but rather the added complexity of a new type of `Symbol`, which is already poorly documented and confusing in many ways (e.g. package ownership and symbol bodies).

Moreover, even with your new type, I don't think this PR will properly fix the problem of discoverability because some users will just use normal symbols or use `T.blah = x -> ...` and things will be even more wonky and half documented ...

d-torrance:
@mahrud, as always, I can count on you for very insightful comments!  You bring up some excellent points.

In the end, all I really care about is `NewMethod`, specifically for documenting `new AtomicInt` in #3093.  It probably does make sense to keep the other proposed "method symbols" hidden, so I think I'll add a slimmed-down version of #3090 to that PR instead.

Closing.

PR #3092, opened 2024-01-26
replaced null pointers with nullptr

mahrud:
This replaces #2886 and goes with https://github.com/Macaulay2/mathicgb/pull/44.

### DIFF
--- a/M2/Macaulay2/e/Eschreyer.cpp
+++ b/M2/Macaulay2/e/Eschreyer.cpp
@@ -104,7 +104,7 @@ GBMatrix *GBKernelComputation::get_syzygies()
   for (int i = 0; i < syzygies.size(); i++)
     {
       result->append(syzygies[i]);
-      syzygies[i] = 0;
+      syzygies[i] = nullptr;
     }
   return result;
 }
@@ -132,20 +132,20 @@ void GBKernelComputation::strip_gb(const GBMatrix *m)
   int i;
   int *components = newarray_atomic_clear(int, F->rank());
   for (i = 0; i < g.size(); i++)
-    if (g[i] != 0) components[g[i]->comp - 1]++;
+    if (g[i] != nullptr) components[g[i]->comp - 1]++;
 
   for (i = 0; i < g.size(); i++)
     {
       gbvector head;
       gbvector *last = &head;
-      for (gbvector *v = g[i]; v != 0; v = v->next)
+      for (gbvector *v = g[i]; v != nullptr; v = v->next)
         if (components[v->comp - 1] > 0)
           {
             gbvector *t = GR->gbvector_copy_term(v);
             last->next = t;
             last = t;
           }
-      last->next = 0;
+      last->next = nullptr;
       gb.push_back(head.next);
     }
   for (i = 0; i < F->rank(); i++) mi.push_back(new MonomialIdeal(R));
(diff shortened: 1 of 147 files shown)


mahrud:
You didn't merge the accompanying PR in mathicgb.

d-torrance:
Speaking of mathicgb, @DaveBarton pointed out that the submodule is currently pointing at https://github.com/Macaulay2/mathicgb/pull/43/commits/71d4ae1afc41be615b6dacaea818db0684882826 (from @jkyang92 's original PR) instead of https://github.com/Macaulay2/mathicgb/commit/09ea46af614ff0827dd777747755a2edd8147243 (@mikestillman's rebased commit that was created when the PR was accepted).  Dave reported that this causes some issues running mathicgb's tests.

I was going to open a PR updating the mathicgb commit, but if https://github.com/Macaulay2/mathicgb/pull/44 is merged soon, then that updating it to that new commit would do the trick too.

mahrud:
Do you know if there's a way to keep things in sync automatically? e.g. perhaps a GitHub action, either here or in mathicgb's repository that automatically fast-forwards to the latest submodule hash in development perhaps?

DaveBarton:
You guys understand git far better than me, but some notes on the web suggest that you can have the mathicgb submodule of M2 track a branch in the mathicgb repo, instead of a specific commit given by a hash. In the M2 dir M2/submodules/mathicgb, you can do "git checkout -b <new-branch> <mathicgb-repo/branch>" apparently. I wonder if you can avoid the new-branch if you just track the master branch in <mathicgb-repo>. But anyway I think the point is that for a foreign submodule like bdwgc, you want to track a specific commit or tag or something, so we only update what M2 points to when we specifically tell it to. But for mathicgb (and mathic and memtailor) maybe we automatically want to track the head of some target branch. See the second answer in https://stackoverflow.com/questions/10914022/how-do-i-check-out-a-specific-version-of-a-submodule-using-git-submodule for instance. (I hope I've got this right - I am new to this.)

mikestillman:
@jkyang92 @mahrud What all needs to be done here?  Can I merge this pull request?  (with a rebase and merge?) (I am having problems myself with differing versions of mathicgb...). After this is done, then what needs to be done next?  

It would be great to keep these in sync automatically.  I must say, I find the git submodule interface to be quite annoying, spending lots of time fixing problems caused by their use!  This is probably because I somehow am missing how they should really be used.

mikestillman:
@mahrud @jkyang92 It looks like you are saying that if I merge this pull request, it will fix the problem caused by rebasing (not sure how that problem arose)?

mikestillman:
Sorry, I'm talking about merging the mathicgb pull request 
https://github.com/Macaulay2/mathicgb/pull/44

d-torrance:
Yes, if you merge the mathicgb PR and we update the mathicgb commit here in the M2 repository, then it should be good.

I like @mahrud's idea of automatically updating the submodules with a GitHub action.  I'm sure it's possible.  https://github.com/marketplace/actions/github-action-submodule-updates looks promising.

mahrud:
This PR is already merged. I don't think rebasing has caused any issues. You should just merge the accompanying PR in mathicgb without creating a merge commit (i.e. using the rebase and merge option from the drop down menu next to the merge button).

PR #3094, opened 2024-01-29
Add ?? as null coalescing operator

d-torrance:
We add a new keyword `??` for a [null coalescing operator](https://en.wikipedia.org/wiki/Null_coalescing_operator).  Basically, `x ?? y` returns `x` unless `x` is "null", in which case we return `y`, which is lazily evaluated.

There are three cases in which `x` is considered "null":

* It's actually `null`
* It raises a non-interrupting error (i.e., the same kind of error that `try` suppresses)
* `(class x)#(symbol ??) x` returns `null` (so users can define how `??` works with their own types)

Using augmented assignment, `??=` is essentially the operator proposed in #1596:

```m2
i1 : x = new MutableList

o1 = MutableList{}

o1 : MutableList

i2 : x#0 ??= 2

o2 = 2

i3 : peek x

o3 = MutableList{2}

i4 : x#0 ??= 3

o4 = 2

i5 : peek x

o5 = MutableList{2}
```

As a proof of concept, I've used it in a few of my packages.


mahrud:
> * `(class x)#(symbol ??) x` returns `null` (so users can define how `??` works with their own types)

I'm confused about this one. Does `?? x` return a boolean?

> * It raises a non-interrupting error (i.e., the same kind of error that `try` suppresses)

So is it equivalent to `try x else y`?

mikestillman:
Your code suggestion is, e.g.:
```
H#foo ??= 3
```
I suppose this returns H#foo or 3 (in case H#foo is null)? 

 It would be nice to be able to set H#foo only if it is not present too, as in 
```
if H#?foo then H#foo else H#foo = 3
```
Can your code handle this too?

d-torrance:
> > * `(class x)#(symbol ??) x` returns `null` (so users can define how `??` works with their own types)
> 
> I'm confused about this one. Does `?? x` return a boolean?

No, there's no associated prefix operator, since I didn't really expect anyone to call the method directly -- it's just called under the hood by the binary operator. 

It wouldn't really make sense to define a binary method for `x ?? y` since we won't know the class of `y` until we've decided that we're not using `x` since it's lazily evaluated.  So there's just this unary method for the type of the left-hand side.  It should either return `x` or `null`.

For example, the Python and R packages have methods installed that return `null` when a `PythonObject` is `None` and when an `RObject` is `NULL`, respectively.

> > * It raises a non-interrupting error (i.e., the same kind of error that `try` suppresses)
> 
> So, assuming `x` is not null, is it equivalent to `try x else y`?

Yes, exactly.  In fact, I pulled out some of the code that's used for evaluating `try` statements into a helper function that's also used by `??`.

> Your code suggestion is, e.g.:
> 
> ```
> H#foo ??= 3
> ```
> 
> I suppose this returns H#foo or 3 (in case H#foo is null)?
> 
> It would be nice to be able to set H#foo only if it is not present too, as in
> 
> ```
> if H#?foo then H#foo else H#foo = 3
> ```
> 
> Can your code handle this too?

Yes!  Take a look at the example in the top post -- that's exactly what's going on there.

mahrud:
Mike, if I understand correctly, there are two different operators:
- `H#foo ?? 3` returns `H#foo` if it is not null, otherwise `3` (no changes to `H`).
- `H#foo ??= 3` returns `H#foo` if it is not null, otherwise sets `H#foo = 3`.

And I think Doug is suggesting a third form:
- `?? X` returns true or false, as determined by a custom method on `class X`.

mahrud:
> No, there's no associated prefix operator, since I didn't really expect anyone to call the method directly -- it's just called under the hood by the binary operator.

I can't think of a specific need for this, but it may be nice to have a specific method or unary operator for the sake of making it clear how to install a `isDefined` method that is used by `X ?? Y` for custom types (e.g. for one type it could be based on `isNonEmpty` and for another `isWellDefined`).

This could be `X ??` or `?? XX` (not to be confused with `? X`) or just `isDefined` if overloading `??` is a bad idea.

d-torrance:
Adding `??` as a prefix operator should be easy, and then we'd be able to use `?? X := x -> f` as syntactic sugar for `installMethod(symbol ??, X, f)`.  And there might be uses for something like `?? x` -- you could turn an an error into `null` for example.  Sounds good to me!

Sticking with `??` would make it compatible out of the box with the code that's already written.  But `?? Thing` probably should be left undefined to avoid an unnecessary function call.

mahrud:
Can the interpreter differentiate `?? X` from `null ?? X` or `X ??` from `X ?? null`?

d-torrance:
Just pushed a few commits adding `??` as a prefix operator.  For example:

```m2
i1 : ?? 2

o1 = 2

i2 : ?? null

i3 : ??(1/0)

i4 : 
```

d-torrance:
> Can the interpreter differentiate `?? X` from `null ?? X` or `X ??` from `X ?? null`?

It's no different than any other keyword that acts as both a prefix and binary operator like `+` and `-`.

mahrud:
Is this expected behavior?
```m2
i1 : a ??= 12

o1 = a

o1 : Symbol

i2 : a

o2 = a

o2 : Symbol

i3 : a = null

i4 : a ??= 12

o4 = 12

i5 : a

o5 = 12
```
I would have expected the first use of `??=` to also set `a = 12`.

Also, something weird is happening here:
```m2
i1 : clearAll ()
stdio:1:1:(3): error: assignment to protected variable 'buckets'

i2 : a ??= 12

o2 = a

o2 : Symbol

i3 : a ??= 12

o3 = a

o3 : Symbol

i4 : clearAll ()
stdio:4:1:(3): error: expected argument 1 to be null, a sequence, list, string, net, hashtable, database, or dictionary

i5 : a ??= 12

o5 = a

o5 : stdio:5:3:(3): error: after print: expected argument 1 to be null, a sequence, list, string, net, hashtable, database, or dictionary
```

d-torrance:
> Is this expected behavior?

It is, but it doesn't have to be.  :)  In an earlier draft, I actually had it consider unprotected symbols as "null" (in fact, I never removed that blurb from the commit message...) and it would have set `a` to be 12 in your example.

I'm wondering if the current behavior assigns the symbol the value of a symbol in some low-level way that's different than just calling `a = a`, unleashing all sorts of weird shenanigans.  I'll play around with it.

Thanks for testing this!

d-torrance:
I'm not getting the strange `clearAll` errors.  Is your build just based on the `development` branch?

mahrud:
AH, it is unrelated .. I'll make a new issue.

d-torrance:
I think I broke `try` on some machines with this PR.  Some of the PPA builds were failing, so I built locally in a 32-bit chroot:

```m2
i1 : try 5
-- SIGSEGV
-* stack trace, pid: 589289
 0# stack_trace(std::ostream&, bool) at ../../../../Macaulay2/d/main.cpp:119
 1# segv_handler at ../../../../Macaulay2/d/main.cpp:235
 2# 0xF7ECE590 in linux-gate.so.1
 3# hashtables_lookup at ../../../../Macaulay2/d/hashtables.dd:341
 4# hashtables_lookup_2 at ../../../../Macaulay2/d/hashtables.dd:353
 5# readeval3(parse_TokenFile_struct*, char, parse_DictionaryClosure_struct*, char, char, char) at ../../../../Macaulay2/d/interp.dd:273
 6# loadprint(M2_string_struct*, parse_DictionaryClosure_struct*, char) at ../../../../Macaulay2/d/interp.dd:346
 7# commandInterpreter_2(tagged_union*) at ../../../../Macaulay2/d/interp.dd:468
 8# evaluate_evalraw at ../../../../Macaulay2/d/evaluate.d:1443
 9# evaluate_evalraw at ../../../../Macaulay2/d/evaluate.d:1399
10# evaluate_evalraw at ../../../../Macaulay2/d/evaluate.d:1399

```

PR #3105, opened 2024-02-04
Document changes for version 1.23

d-torrance:
I went back through the list of merged pull requests since 1.22 was released and tried to document most of the changes to `Core` or to the build process that I had a hand in and users might be interested in.


mahrud:
Not a comment about this PR in particular, but I find personally find the changelog style very hard to use. For instance, each version has its own page, so often I'd have to open half a dozen different pages to look for something, at which point I usually give up and look at the `changes.m2` source instead! Also, it's hard to track changes to a particular package or feature.

PR #3106, opened 2024-02-05
added PlaneCurveLinearSeries.m2 and updated the list of packages

eisenbud:
A new package for dealing with linear series on a smooth curve from information about a (usually singular) plane model. This will be referenced in the forthcoming book, "The Practice of Algebraic Curves" by David Eisenbud and Joe Harris, to be published by the American Mathematical society. The script implements and gives examples of various aspects of
Chapters 4,5 and 14.

### DIFF
--- a/M2/Macaulay2/packages/PlaneCurveLinearSeries.m2
+++ b/M2/Macaulay2/packages/PlaneCurveLinearSeries.m2
@@ -1,0 +1,1189 @@
+newPackage(
+          "PlaneCurveLinearSeries",
+          Version => "1.0",
+          Date => "February 4, 2024",
+          Headline => "Linear series on the normalization of a plane curve",
+          Authors => {{ Name => "David Eisenbud", 
+		  Email => "de@berkeley.edu", 
+		  HomePage => "eisenbud.io.github.com"}},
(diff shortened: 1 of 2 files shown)


M2/Macaulay2/packages/PlaneCurveLinearSeries.m2
```diff
@@ -0,0 +1,1189 @@
...
+   setRandomSeed 0
+   I = ideal random(3, intersect(pS,oS))
+   E = S/I   
+   geometricGenus E
+   q = o
+   netList ({o} | apply(5, i-> q = addition(o,p,q,E)))
+  Text
+   A cubic with a node or cusp is also arithmetic genus 1; in the case
+   of a node, the smooth points are in correspondence with P^1 minus {0, infinity}
+   and the Jacobian is the multiplicative troup of the field.
```
mahrud, line 407: @DanGrayson I don't understand why the spellchecker didn't run on this PR, but the misspelling of "group" is causing tests to fail on other PRs.

d-torrance: The spellchecker didn't run because this PR originally targeted `master` and currently those tests are only run against PR's to `development`.  This shouldn't be a problem once #3080 is merged into `master`.

PR #3111, opened 2024-02-06
Complexes for 1 23

mikestillman:
The latest version of Complexes.  This includes freeResolution documentation, and implementation which is independent of ChainComplex'es.


mikestillman:
@DanGrayson @mahrud This pull request says that it cannot be rebased due to conflicts.  But I see no conflicts?  Perhaps I should have merged development in one more time before doing the pull request.  Maybe i will do that right now anyway...

mikestillman:
@DanGrayson I screwed up: I think the mathicgb in here is for an older commit.  I was trying to include that one...

mikestillman:
@DanGrayson Actually, scratch that, it seems to have the correct commit for mathicgb.

mikestillman:
@DanGrayson All the tests ran to completion, except the lint test.  But the mispellings it shows are for another package that just got accepter earlier today (or yesterday?)

For some reason, it is now rerunning the tests.  I say we can accept this pull request now (actually, it is mine...).  What do you think?

DanGrayson:
Pulled.

PR #3113, opened 2024-02-06
Update M2-emacs submodule

d-torrance:
Fixes bug where indentation was reset when there were blank lines (see https://github.com/Macaulay2/M2-emacs/pull/55)


d-torrance:
Another typo snuck in in a PR that was originally targetted at `master`:

> M2/Macaulay2/packages/PlaneCurveLinearSeries.m2:407: troup ==> troupe

Here it is in context:

>  A cubic with a node or cusp is also arithmetic genus 1; in the case
   of a node, the smooth points are in correspondence with P^1 minus {0, infinity}
   and the Jacobian is the multiplicative troup of the field.
   To allow the program to consider this as a curve of arithmetic genus 1, 
   use the optional argument Conductor.

Should that be multiplicative *group*?

mahrud:
I don't know why spellchecker didn't run on that PR.

PR #3124, opened 2024-02-09
Add MPFR support to ForeignFunctions package

d-torrance:
For example:

```m2
i1 : needsPackage "ForeignFunctions";

i2 : mpfrAdd = foreignFunction("mpfr_add", void, {mpfrT, mpfrT, mpfrT, int})

o2 = mpfr_add

o2 : ForeignFunction

i3 : x = mpfrT 0p100

o3 = 0

o3 : ForeignObject of type mpfr_t

i4 : mpfrAdd(x, numeric(100, pi), exp 1p100, 0)

i5 : x

o5 = 5.85987448204883847382293085463

o5 : ForeignObject of type mpfr_t

i6 : value oo

o6 = 5.85987448204883847382293085463

o6 : RR (of precision 100)

### DIFF
--- a/M2/Macaulay2/d/ffi.d
+++ b/M2/Macaulay2/d/ffi.d
@@ -357,6 +357,10 @@ ffiRealType(e:Expr):Expr := (
     else WrongArgZZ());
 setupfun("ffiRealType", ffiRealType);
 
+-- returns pointer to real number object with given value
+-- inputs: (x:RR, y:ZZ)
+-- x = value
+-- y = type (0 = mpfr_t, 32 = float, 64 = double)
 ffiRealAddress(e:Expr):Expr := (
     when e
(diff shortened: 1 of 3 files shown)


mahrud:
I'm a bit confused. Ideally one shouldn't need any changes to the interpreter/engine to connect to a new library using foreign functions, right? Is this just an optimization then?

d-torrance:
> I'm a bit confused. Ideally one shouldn't need any changes to the interpreter/engine to connect to a new library using foreign functions, right? Is this just an optimization then?

Correct.  This is so we can call functions from other libraries that might use MPFR.  The specific case I have in mind is the ball arithmetic stuff in FLINT.  It would be possible to call everything just using `double`'s, but it would be nice to allow for arbitrary precision.

PR #3127, opened 2024-02-13
Add parallelism for minimalBetti and non-minimal resolutions

mikestillman:
This pull request (done with @moorewf) adds parallelism in the engine for minimalBetti, and non minimal resolutions (along with refactoring the code to handle more finite fields).  Some undocumented features of non-minimal resolutions have been removed, including some hacks for working with essentially products of rings.  This PR also contains functionality to read directly (i.e. without going through the front-end parser) matrices and msolve output.  Also reinstated and refactored is the Algorithm => LinearAlgebra option to `gb`, but this code is not yet parallelized.

@DanGrayson Please don't accept this until we ok it (which we will do within the day).  I first want to see if pass all the tests.

I would like (if possible) to have this in the upcoming release.


mikestillman:
@DanGrayson @mahrud  Any reason it says I cannot do "rebase and merge" due to conflicts?  I just merged development into my branch a few minutes ago.

Also, the builds are failing.  It appears due to ffpack/openmp things?  I haven't changed anything with that, at least I thought I had not...

mikestillman:
@d-torrance The failures seem also to have something to do with python include files?  This compiled fine on macos on my machine...?

d-torrance:
I don't see any Python-related errors.  The builds all seem to be complaining about `std::memcmp`, e.g.,:

```
In file included from ../../../../Macaulay2/e/BasicPoly.cpp:2:
../../../../Macaulay2/e/gb-f4/MonomialView.hpp: In member function ‘bool newf4::MonomialView::operator==(const newf4::MonomialView&) const’:
../../../../Macaulay2/e/gb-f4/MonomialView.hpp:44:37: error: ‘memcmp’ is not a member of ‘std’; did you mean ‘wmemcmp’?
   44 |     return ((size() != 0) && ((std::memcmp(mData + 1,
      |                                     ^~~~~~
      |                                     wmemcmp
```

mikestillman:
@d-torrance Thanks, I did not look closely enough. Fixed (I hope).

mahrud:
> Any reason it says I cannot do "rebase and merge" due to conflicts? I just merged development into my branch a few minutes ago.

I think it's exactly because you merged with the development. If you plan on using "rebase and merge" on GitHub, avoid using merges with development locally and only rebase on top of development whenever there are new commits. Currently there are 10 commits that just say "Merge remote-tracking branch 'origin/development' into refactor-f4" and GitHub doesn't seem to know what to do with these.

mahrud:
If you want to use a rebase (which I still think would be nice), you would need to rebase on top of development locally and then force push to this branch.

Alternatively, I already tried rebasing to see how bad were the conflicts (very mild, only a few deletions that conflicted with my nullptr PR) and the result is in [this branch](https://github.com/mahrud/M2/tree/refactor-f4) which you can use for a PR instead.

mikestillman:
> If you want to use a rebase (which I still think would be nice in order to avoid 10 merge commits), you would need to rebase on top of development locally and then force push to this branch.
> 
> Alternatively, I already tried rebasing to see how bad were the conflicts (very mild, only a few deletions that conflicted with my nullptr PR) and the result is in [this branch](https://github.com/mahrud/M2/tree/refactor-f4) which you can use for a PR instead.

Hi @mahrud , the problem is that Frank Moore and I have been working on this together, and rebasing with others has in my experience caused lots of issues.  Merging in development was necessary, to get fixes needed, and also because this was a long running branch.  In fact, I was hoping that the "merge and rebase" would just clean all those merges up on its own.  Usually this is not a problem (well, I have seen problems before), I think if I merge the development branch in, and there are no conflicts that my workflow seemed ok.  However, every now and then there are conflicts, such as with your nullptr commit, and those seem to confuse it a lot.  

I just pushed one more change due to a failing test.  If this works, maybe we can switch over to your suggested pull request (after placing this last commit in there).

PR #3129, opened 2024-02-13
VirtualResolutions minor update

mahrud:
Quick update that fixes a few issues and adds some speedups.

- raised the limit of FourTiTwo for printing debug info
- added quick test for non-quasilinearity in LinearTruncations
- improved multigradedRegularity
- added new test for multigradedRegularity
- quickfix in VirtualResolutions
- fixed internal function plotRegion

One important change in Core:
- improved RingMap ** Module for modules without relations (e.g. an image module remain an image module)

### DIFF
--- a/M2/Macaulay2/m2/matrix.m2
+++ b/M2/Macaulay2/m2/matrix.m2
@@ -634,12 +634,13 @@ inducedMap(Module,Module,Matrix) := Matrix => opts -> (N',M',f) -> (
      N := target f;
      M := source f;
      if ring N' =!= ring M' or ring N' =!= ring f then error "inducedMap: expected modules and map over the same ring";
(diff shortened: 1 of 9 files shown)


mahrud:
@mikestillman @DanGrayson currently `RingMap Module` only works for free modules and image modules while `RingMap ** Module` will convert any module to a cokernel module since it computes `cokernel f presentation M`.

Is there any reason why both `RingMap Module` and `RingMap ** Module` should not work for all modules by just applying the ring map to the generators and relations and creating an equivalent subquotient module?

This PR is changing `RingMap ** Module` to do as above, and I expected some test to fail if this was on purpose, but it doesn't seem to be the case:
https://github.com/Macaulay2/M2/blob/060e16fa0a4dc132ebe857f21e582dee92d129be/M2/Macaulay2/m2/ringmap.m2#L200-L219

mahrud:
@mikestillman I added a fix for the bug from David's email here.

PR #3131, opened 2024-02-14
comment out use of <concepts> header file

mikestillman:
Forgot to comment out this header file for use with <= c++17...

### DIFF
--- a/M2/Macaulay2/e/PolynomialStream.hpp
+++ b/M2/Macaulay2/e/PolynomialStream.hpp
@@ -1,9 +1,5 @@
 #pragma once
 
-#if 0
-#include <concepts>
-#endif
-
 namespace newf4
 {
 
--- a/M2/Macaulay2/e/PolynomialStream.hpp
+++ b/M2/Macaulay2/e/PolynomialStream.hpp
@@ -1,6 +1,8 @@
 #pragma once
 
+#if 0
 #include <concepts>
+#endif
 
 namespace newf4
 {


mahrud:
Could you use this instead?
```cpp
#if __cplusplus >= 202002L
    // C++20 (and later) code
#endif
```

mikestillman:
@mahrud @d-torrance We decided to not try to maintain two different c++ versions for the moment.  We are already maintaining TBB and no TBB.  Actually, now that TBB has settled down, perhaps we can insist on TBB support.   Thoughts?

mahrud:
I'm not asking for supporting two different versions. I just don't like the use of `#if 0` because it's unclear what it means. Is this code deprecated and should be deleted? Is it experimental and not ready yet? Is it disabled because it's wrong and should be fixed?

d-torrance:
>We are already maintaining TBB and no TBB. Actually, now that TBB has settled down, perhaps we can insist on TBB support. Thoughts?

That's fine with me -- the Debian/Ubuntu packages build with TBB support in all cases.

mahrud:
> > We are already maintaining TBB and no TBB. Actually, now that TBB has settled down, perhaps we can insist on TBB support. Thoughts?
> 
> That's fine with me -- the Debian/Ubuntu packages build with TBB support in all cases.

Is it tbb 2020 in all cases though?

d-torrance:
> Is it tbb 2020 in all cases though?

Yeah.  I had to backport TBB 2020 for a couple of the older releases.

Issue #3009, opened 2023-11-22
Flint 3.0.0 build issue


jmcdonough98:
Attempting to build the latest release version on Arch fails at NCAlgebras:
```
CXX NCAlgebras/NCF4.cpp
In file included from ./aring-translate.hpp:16,
                 from ./aring-glue.hpp:7,
                 from ./VectorArithmetic.hpp:9,
                 from ./NCAlgebras/NCF4.hpp:14,
                 from NCAlgebras/NCF4.cpp:1:
./aring-zz-flint.hpp: In member function ‘unsigned int M2::ARingZZ::computeHashValue(const ElementType&) const’:
./aring-zz-flint.hpp:46:38: error: ‘fmpz_get_ui’ was not declared in this scope; did you mean ‘mpz_get_ui’?
   46 |     return static_cast<unsigned int>(fmpz_get_ui(&a));
      |                                      ^~~~~~~~~~~
      |                                      mpz_get_ui
```
This is the same issue as #2972. I was able to successfully build the development branch, but would it be possible to push this fix to master?

feynhat:
What's the fix? (Including "fmpz.h" doesn't work.)

d-torrance:
Building the `development` branch should work.  (`git checkout development` after cloning)

feynhat:
Yeah thanks. I was building 1.22. 
Dev branch is building fine (so far).

ghbrown:
I ran into this problem (and #2849) when building from source via `make` on Arch Linux according to the instructions in `INSTALL`, which recommend changing to the latest release branch (`release-1.22-branch` for me).
Is there something that can be done to ensure the instructions in `INSTALL` result in a successful build? The fact that the "official" instructions result in a failed build seems problematic. Perhaps a new release, a general warning to try `development` in INSTALL, etc.?

d-torrance:
Yeah, I'd be very hesitant to require older versions of dependencies. Usually, newer versions are backwards compatible.  FLINT 3 was a relatively rare exception.

mahrud:
M2 doesn't compile with most recent version of flint:
```cpp
/tmp/macaulay2-20240304-13320-pqsrxl/M2-release-1.23-rc1/M2/Macaulay2/e/mat-linalg.hpp:867:15: fatal error: no matching function for call to 'fq_nmod_mat_rref'
  long rank = fq_nmod_mat_rref(A1.fq_nmod_mat(), A.ring().flintContext());
[...]

PR #3141, opened 2024-02-22
A few updates to branch/tag information in INSTALL

d-torrance:
In response to https://github.com/Macaulay2/M2/issues/3009#issuecomment-1958230775, we add some clarification regarding building the `master` and `development` branches v. the most recent `release-*` tag.

(Since this PR only touches `INSTALL` and doesn't affect the build, the CI tests are skipped.)

### DIFF
--- a/M2/INSTALL
+++ b/M2/INSTALL
@@ -61,6 +61,11 @@ source code:
     git submodule sync
     git submodule update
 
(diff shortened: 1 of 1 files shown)


ghbrown:
Thanks for these edits, I think this will help avoid users needing to go scouring GitHub issues to find fixes to some build issues caused by dependency version mismatches.

PR #3143, opened 2024-02-22
fixed is_equal to ignore Schreyer order on free modules

### LINKED ISSUE #3135: Equality of modules with different monomial orders
Can we change `Module === Module` in the interpreter to ignore Schreyer orders? This is causing too many problems, almost exclusively for free modules:
```m2
  R = QQ[x,y,z]/(x^2,y^2,z^2)
  C = res coker vars R
  M = ambient coker C.dd_3
  F = R^{6:-2}
  -- both free modules:
  M ==  F -- true
  M === F -- false
  intersect(M, F) -- error
  remainder(M, F) -- error
  inducedMap(M, F, id_(2 * F)) -- error
```
And here is the problem: (scroll to the right)
```m2
i49 : peek M

o49 = Module of Vector{cache => CacheTable{...4...}                                                                 }
                       numgens => 6
                       RawFreeModule => free(rank 6 degrees = {T2, T2, T2, T2, T2, T2}z2.0 yz.1 xz.2 y2.3 xy.4 x2.5)
                       ring => R

i50 : peek F

o50 = Module of Vector{cache => CacheTable{...5...}                                    }
                       numgens => 6
                       RawFreeModule => free(rank 6 degrees = {T2, T2, T2, T2, T2, T2})
                       ring => R
```
All of these can be solved with an appropriate change of `===` to `==` in Core, such as in 1338599bee952ce060d207b4fe5e10468f6b1b74 and 7f1c585cfb58574afb1dd8f5a493452ebaf7f1ed, but this will have a negative effect on speed when the modules are not free and there needs to be a gb computation instead.

mahrud:
Also partially reverts changes from 1338599 and 7f1c585.

Closes #3135.

### DIFF
--- a/M2/Macaulay2/e/freemod.cpp
+++ b/M2/Macaulay2/e/freemod.cpp
@@ -163,8 +163,10 @@ bool FreeModule::is_equal(const FreeModule *F) const
     for (i = 0; i < rank(); i++)
       if (0 != D->compare(degree(i), F->degree(i))) return false;
 
+  /* free modules with same ranks and degrees should be equal
   if (schreyer != nullptr) return schreyer->is_equal(F->schreyer);
   if (F->schreyer != nullptr) return false;
+  */
 
   return true;
 }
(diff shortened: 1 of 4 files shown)


mikestillman:
@DanGrayson I'm not sure it is possible, but if so, this would be a good (and simple) fix to get in this release (assuming it passes the tests)

mahrud:
The server running the autotools macos test seems to have crashed. I'll restart that test once everything finishes.

PR #3146, opened 2024-02-27
Update prism.js (and vim) symbols for 1.23

d-torrance:
This updates the file `prism.js` in the `Style` package (used for syntax highlighting the html documentation) with the new symbols added in 1.23.

The vim symbols were also updated in the process, so I committed those changes for good measure.

This is a draft for now, as I intend to also update the M2-emacs submodule commit if https://github.com/Macaulay2/M2-emacs/pull/61 is merged.


d-torrance:
The M2-emacs submodule commit has been updated, so I've marked this as ready for review.

PR #3147, opened 2024-02-27
Fix typo in compareExchange docs (z is stored in x, not y)

d-torrance:
Thanks to @pzinn for catching this!

@DanGrayson -- would it be possible to include this small fix in the release?


DanGrayson:
> @DanGrayson -- would it be possible to include this small fix in the release?

Yes.

Issue #603, opened 2017-05-10
Bug in quotient


moorewf:
Each of the first two sequences of commands work fine:

M = matrix {{1},{0}}
N = matrix {{0},{1}}
M // N

M = sub(matrix {{1},{0}}, QQ)
N = sub(matrix {{0},{1}}, QQ)
M // N

However, when I try to factor a matrix through another over a finite field:

M = sub(matrix {{1},{0}}, ZZ/32003)
N = sub(matrix {{0},{1}}, ZZ/32003)
M // N

I get the error:
stdio:9:3:(3): error: no method found for applying matrix to:
     argument   :  null (of class Nothing)

DanGrayson:
This is a top level problem, thanks!

moorewf:
I know this is an old issue, but I would be happy to try to fix it but I don't completely understand how the hooks system works (i.e. how different strategies are assigned/selected)  Any suggestions as to where to look?

mahrud:
> I know this is an old issue, but I would be happy to try to fix it but I don't completely understand how the hooks system works (i.e. how different strategies are assigned/selected) Any suggestions as to where to look?

You can always run hooked functions with `debugLevel=1` to see which strategies are attempted:
```m2
i6 : debugLevel=1

o6 = 1

i7 : quotient(M, N)
 -- (quotient,Matrix,Matrix) with Strategy => Default from -*Function[../linuxbrew/.linuxbrew/share/Macaulay2/Core/matrix2.m2:328:70-353:39]*-
stdio:7:1:(3): error: quotient: no method implemented for this type of input
```
You can also see all available strategies (in the order they are attempted) using `hooks` and read the code using `code`:
```m2
i4 : hooks(quotient, Matrix, Matrix)

o4 = {0 => (quotient, Matrix, Matrix, Strategy => Default)}

o4 : NumberedVerticalList

i5 : code 0

o5 = -- code for method: quotient(Matrix,Matrix)
     ../linuxbrew/.linuxbrew/share/Macaulay2/Core/matrix2.m2:328:70-353:39:
     --source code:
...
```
In particular, here there is only one strategy available (the `LocalRings` package adds another).

In this case, it seems like the code attempts to run and return `solve(M, N)`, which returns `null`, but when working with hooks `null` means the strategy "doesn't work", so the error is returned that no strategy is applicable.

[...]

PR #3149, opened 2024-03-03
Attempted fix for bug in quotient for matrices over finite field.

moorewf:
This is an attempted fix for #603.

Feel free to suggest alternative solutions, but at this point if solve return non-null, then that value is returned, otherwise the hook for quotient continues as usual.  This results in "throwing away" the work from solve if a solution is not found(which arises from a call to rawLinAlgSolve of a mutable dense matrix), which may be undesirable.


moorewf:
Ok, I can add some checks.  I poked around and it wasn't obvious where they should go.  Any suggestions?

mahrud:
Maybe one of the files with quotient in the name in `Macaulay2/tests/normal`?

PR #3153, opened 2024-03-07
Further corrections/improvements to RealRoots:-realRootIsolation

cel34-bath:
Correction to some issues that arose (if polynomial was of the form Ax^2-Bx, Lagrange bound would miss the root |x|=|B/A|; the case where the polynomial is one term (i.e Ax^n) was improved by simply setting the bound to 0 to avoid any further calls to bisection algorithm). Actual comparisons between Cauchy and Lagrange bound revealed if the Lagrange bound is ever better, is it better by at most one, so Lagrange bound has been removed (this also avoids roots being missed in the above case).

Knuth bound was also introduced, which is much tighter in cases of coefficient blowup. This bound was relaxed slightly to keep the bounds/intervals in QQ.


cel34-bath:
I've had a look at the build check failure and there are a large number of errors in cf_factor.cc where fmpz_poly_factor and similar identifiers are undeclared, is any of this related to changes I've made or does something else need correcting? 

If it has been caused by the changes I've made, I would assume it's because I've lifted some elements to QQ (so maybe the case for when elements are in ZZ is never called, so these are never initialised), but I don't really know what I'm talking about!

Please let me know if there is anything I need to correct, otherwise please rerun the checks.

d-torrance:
The build errors are unrelated to your pull request.  I'm pretty sure it's because we're building an older version of factory using the latest version FLINT, which introduced some breaking changes.

DanGrayson:
Could you fix the build error? Perhaps merge with the development branch and then push it up stream.

PR #3154, opened 2024-03-14
VNumber - a package for computing the v-number and v-function

EmanueleSgroi:
We provide a package to calculate the v-number of homogeneous ideals and the v-function of monomial ideals. The concept of v-number was introduced by Cooper et al. in

"S.M. Cooper, A. Seceleanu, S.O. Tohaneanu, M. Vaz Pinto, R.H. Villarreal. Generalized minimum distance functions and algebraic invariants of geramita ideals. Advances in Applied Mathematics, 112:101940, 2020".

We call v(I^k) the v-function of I, where I is a graded ideal of a graded polynomial ring S.

The package consists of several functions; the main functions are vNumberP, vNumber, vFunctionP and vFunction.


d-torrance (commented):
Thanks!

d-torrance:
> Hello @d-torrance, I noticed that the check related to autotools-macos-12-default was not successful, but I'm unsure what might have caused it. Could it be the VNumber package or something else? Thanks in advance!

It's totally unrelated -- don't worry about it!

EmanueleSgroi:
> > Hello @d-torrance, I noticed that the check related to autotools-macos-12-default was not successful, but I'm unsure what might have caused it. Could it be the VNumber package or something else? Thanks in advance!
> 
> It's totally unrelated -- don't worry about it!

Great, thank you!

DanGrayson:
Could you fix the build error?  Perhaps merge with the development branch and then push it up stream.

PR #3159, opened 2024-03-21
Update factory to 4.3.0 in autotools build

d-torrance:
We update the autotools build so that it used (a patched version of) factory 4.3.0 with FLINT 3 support, just like the cmake build.

I think this should fix the macOS autotools builds that are failing.  What's currently happening is that the homebrew factory isn't linked properly (it still depends on flint 3.0.1 even though it was built against flint 3.1.0 or something like that).  So even though we try to install factory from homewbrew in the GitHub builds, `pkg-config` isn't able to find it during `configure` and we try to build it from source.  But we were building an older version without support for FLINT 3 and the build was failing.

So until the homebrew package starts working, this should at least fix it so that we can build factory.


mahrud:
Isn't factory at 4.3.2 now?

d-torrance:
> Isn't factory at 4.3.2 now?

It is, but 4.3.0 is the latest version with a tarball hosted at https://www.singular.uni-kl.de/ftp/pub/Math/Factory.

> Also, the patch name was `4.3.0...dcca183` because it's the git diff between tag `Release-4.3.0` and commit `dcca183`. I think that's more informative. Right now brew adds the patch for `Release-4-3-0..151235c5`, which fixed flint compatibility issues.

I just kept the naming convention of the other patches in the `libraries` directory.  Another approach to documenting things like the git diff might be to add some metadata to the headers of patches.  I really like Debian's [patch tagging guidelines](https://dep-team.pages.debian.net/deps/dep3/), for example.

PR #3160, opened 2024-03-22
Fix link to JSAG copy of BettiCharacters package

d-torrance:
When I was preparing the 1.23 release, I noticed that the link to the `BettiCharacters` source code on the JSAG webpage was broken.  I was able to find it by playing around with the URL, and I used the link that worked in the `Certification` section.

I reported the broken link to JSAG.  They fixed it, but now the link I used in the package is broken!


mahrud:
Since the change is minor I cancelled the builds. Would be good if there were different levels of checks:
1. no check at all
2. only lint / spellcheck
3. also run package examples / tests using latest nightly build
4. also compile first

PR #3183, opened 2024-04-16
New operators

mahrud:
- **added postfix ops `^! _! ^~ _~`**
- **added binary ops `^> ^>= ^< ^<= _> _>= _< _<=`**

I also wanted to add postfix op `^#` (for instance used in math for the ring map corresponding to a map of ringed spaces) and `_#` but currently this would conflict with things like `2^#L`. Perhaps they could be changed to `2 ^ #L`, but this could be an annoying breaking change.


mahrud:
@d-torrance if you'd like to switch `~` to a prefix op in this PR, feel free to push commits to do that here.

d-torrance:
I think if we do switch `~` to a prefix operator, it should wait for at least another release to give folks time to transition from `M~` to `M^~` for sheafification.

For the failed builds, I think either the other four operators should be added to the augmented assignment blacklist in `binding.d` or their augmented versions should be added to `exports.m2`.

DanGrayson:
An operator that begins with _ or ^ should be unary and postfix.

mahrud:
Why? I'd like to use the binary operators for truncations of modules, complexes, and filtrations in general.

d-torrance:
We already have several binary operators that begin with `_` or `^`:  `_` and `^` themselves, as well as `^**`, `^^`, and all the corresponding augmented assignment operators.

PR #3186, opened 2024-04-22
Parallelization updates

d-torrance:
This branch includes a number of improvements to Macaulay2's code related to parallelization, primarily by @DaveBarton.  The work began at the AIM workshop ["Macaulay2: expanded functionality and improved efficiency"](https://aimath.org/pastworkshops/macaulay2efie.html) last fall.

Improvements include:

* Changed initial value of `currentAllowedThreads` from 2 to 4 to 5, added 1 to `maxAllowableThreads`
* Use default `GC_FREE_SPACE_DIVISOR` of 3 not 12, initial heap size 70MB -> 150MB + `maxAllowableThreads` * 8MB
* `elapsedTime`: also show process, thread, gc time.  Also change `cpuTime()` to return process, not thread, time.
* Have `taskResult()` wait for the task to finish
* Added `parallelApply` - make it take `allowableThreads` as an optional parameter!?
* Comment out some `GC_FREE()`s
* Add patch for GC bug fix
* Add `getThreadIOMode` function and improve the existing `setIO*` functions

### DIFF
--- a/M2/Macaulay2/c/README
+++ b/M2/Macaulay2/c/README
@@ -289,7 +289,7 @@ Expressions:
 
 	export a := b				export a variable or function (by declaring it in *.sig)
 	import a				import a variable (used mainly in *.sig files)
-	thread a := b				declare a thread local variable
+	threadLocal a := b			declare a thread local variable
 
 	f(x,y)					function application
 	when e 					examine e
--- a/M2/Macaulay2/c/README
+++ b/M2/Macaulay2/c/README
@@ -266,7 +266,7 @@ Expressions:
 	if a then b
 	until a do b
 	while a do b			evaluates b as long as a is true
-	while a list b			produces a list of the values of b, as long as a is true
+	while a list b			produces a list of the values of b, as long as a is true (not implemented!?)
 	foreach a in b do c		
 	foreach a at i in b do c
 	foreach a at i in b by s do c
@@ -330,7 +330,7 @@ Useful functions
 	+	-	*	/	%
 	^^					exclusive or
 	^					power (for doubles)
-	::					nothing
+	::					nothing (not implemented!?)
 
 Operators
 
--- a/M2/Macaulay2/c/README
+++ b/M2/Macaulay2/c/README
@@ -210,7 +210,7 @@ Types:
 
 	Pointer "bar"			a new pointer type, declared as "bar" in the C or C++ code produced
 	atomicPointer "bar"		a new atomic pointer type, declared as "bar" in the C or C++ code produced
-	malloc(T)			returns a pointer of type T pointing to new memory on the heap
+	GCmalloc(T)			returns a pointer of type T pointing to new memory on the heap
 					the number of bytes allocated is sizeof(*((T)0))
 					if T is a pointer type, the memory is cleared by GC_MALLOC
 					if T is an atomic pointer type, the memory is not cleared by GC_MALLOC_ATOMIC
--- a/M2/Macaulay2/d/M2mem.c
+++ b/M2/Macaulay2/d/M2mem.c
@@ -69,7 +69,7 @@ void freememlen(void *s, size_t old) {
 #ifdef MEMDEBUG
      M2_debug_free(s);
 #else     
-     GC_FREE(s);
+     /* GC_FREE(s); */
 #endif
 }
 
@@ -80,7 +80,7 @@ void freemem(void *s) {
 #ifdef MEMDEBUG
      M2_debug_free(s);
 #else     
-     GC_FREE(s);
+     /* GC_FREE(s); */
 #endif
 }
 
@@ -192,7 +192,7 @@ char *getmoremem_atomic (char *s, size_t old, size_t new) {
      size_t min = old<new ? old : new;
      if (p == NULL) outofmem2(new);
      memcpy(p, s, min);
-     GC_FREE(s);
+     /* GC_FREE(s); */
 #    ifndef NDEBUG
      {
        int excess = new - min;
--- a/M2/Macaulay2/d/pthread.d
+++ b/M2/Macaulay2/d/pthread.d
@@ -217,6 +217,7 @@ setupfun("taskResult",taskResult);
 setupfun("setIOSynchronized",setIOSynchronized);
 setupfun("setIOUnSynchronized",setIOUnSynchronized);
 setupfun("setIOExclusive",setIOExclusive);
+setupfun("getIOThreadMode", getIOThreadMode);
 
 -- Local Variables:
 -- compile-command: "echo \"make: Entering directory \\`$M2BUILDDIR/Macaulay2/d'\" && make -C $M2BUILDDIR/Macaulay2/d pthread.o "
--- a/M2/Macaulay2/d/pthread.d
+++ b/M2/Macaulay2/d/pthread.d
@@ -203,7 +203,9 @@ taskResult(e:Expr):Expr := (
      when e is c:TaskCell do
      if c.body.resultRetrieved then buildErrorPacket("task result already retrieved")
      else if !taskKeepRunning(c.body.task) then buildErrorPacket("task canceled")
-     else if !taskDone(c.body.task) then buildErrorPacket("task not done yet")
+     else if !taskDone(c.body.task) then (
+	  Ccode(voidPointer, "waitOnTask(",c.body.task,")");
+	  taskResult(e))
      else (
 	  r := c.body.returnValue;
 	  c.body.returnValue = nullE;
--- a/M2/Macaulay2/d/pthread.d
+++ b/M2/Macaulay2/d/pthread.d
@@ -217,6 +217,15 @@ setupfun("taskResult",taskResult);
 setupfun("setIOSynchronized",setIOSynchronized);
 setupfun("setIOUnSynchronized",setIOUnSynchronized);
 setupfun("setIOExclusive",setIOExclusive);
+
+export getIOThreadMode(e:Expr):Expr := (
+    when e
+    is a:Sequence do (
+	if length(a) == 0
+	then toExpr(getFileThreadMode(stdIO))
+	else WrongNumArgs(0, 1))
+    is f:file do toExpr(getFileThreadMode(f))
+    else WrongArg("a file or ()"));
 setupfun("getIOThreadMode", getIOThreadMode);
 
 -- Local Variables:
--- a/M2/Macaulay2/bin/main.cpp
+++ b/M2/Macaulay2/bin/main.cpp
@@ -68,8 +68,14 @@ int main(/* const */ int argc, /* const */ char *argv[], /* const */ char *env[]
   while (env[++envc] != NULL) { /* iterate over environ until you hit NULL */ }
 
   GC_INIT();
+  size_t heap_size = GC_get_heap_size(),
+	 min_heap_size = 150'000'000 + getMaxAllowableThreads() * 8'000'000;
+		// each thread is currently started with 8 MB of stack space
+  if (heap_size < min_heap_size)
+    GC_expand_hp(min_heap_size - heap_size);
   GC_call_with_alloc_lock(GC_start_performance_measurement_0, NULL);
     // record total time of (full) gcs for GCstats()
+
   IM2_initialize();
 
 #ifndef NDEBUG
--- a/M2/Macaulay2/bin/main.cpp
+++ b/M2/Macaulay2/bin/main.cpp
@@ -56,6 +56,11 @@ void* interpFunc(ArgCell* vargs);
 void* profFunc(ArgCell* p);
 void* testFunc(ArgCell* p);
 
+static void * GC_start_performance_measurement_0(void *) {
+  GC_start_performance_measurement();
+  return NULL;
+}
+
 int main(/* const */ int argc, /* const */ char *argv[], /* const */ char *env[])
 {
   /* find the number of environment variables defined */
@@ -63,6 +68,8 @@ int main(/* const */ int argc, /* const */ char *argv[], /* const */ char *env[]
   while (env[++envc] != NULL) { /* iterate over environ until you hit NULL */ }
 
   GC_INIT();
+  GC_call_with_alloc_lock(GC_start_performance_measurement_0, NULL);
+    // record total time of (full) gcs for GCstats()
   IM2_initialize();
 
 #ifndef NDEBUG
--- a/M2/Macaulay2/c/dictionary.c
+++ b/M2/Macaulay2/c/dictionary.c
@@ -220,7 +220,7 @@ static void psymbol(node s){
      if (s->body.symbol.flags & defined_F) put(" initialized");
      if (s->body.symbol.flags & export_F) put(" export");
      if (s->body.symbol.flags & import_F) put(" import");
-     if (s->body.symbol.flags & threadLocal_F) put(" thread");
+     if (s->body.symbol.flags & threadLocal_F) put(" threadLocal");
      if (s->body.symbol.flags & const_F) put(" const");
      if (s->body.symbol.flags & global_F) put(" global");
      if (s->body.symbol.flags & literal_F) put(" literal");
--- a/M2/Macaulay2/d/M2lib.c
+++ b/M2/Macaulay2/d/M2lib.c
@@ -27,16 +27,23 @@ void system_handleInterruptsSetup(M2_bool handleInterrupts) {
   oursignal(SIGALRM,handleInterrupts ? alarm_handler : SIG_DFL);
 }
 
-static __thread double startTime;
+static double startTime;
 double system_cpuTime(void) {
   struct timespec t;
-  int err = clock_gettime(CLOCK_THREAD_CPUTIME_ID, &t);
+  int err = clock_gettime(CLOCK_PROCESS_CPUTIME_ID, &t);
   if (err) return 0; /* silent about error */
   double u = t.tv_sec + t.tv_nsec * 1e-9;
   return u - startTime;
 }
 void system_cpuTime_init(void) {
   startTime = system_cpuTime();
+}
+
+double system_threadTime(void) {
+  struct timespec t;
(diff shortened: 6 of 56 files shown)


DaveBarton:
Doug, how is this branch different from your parallelization branch? It seems like you left out a lot of the commits that accomplished the points you mentioned. Also the most important point is that in the parallelization branch, parallelization actually works, i.e. it's thread-safe, doesn't crash, and runs faster. See https://github.com/d-torrance/M2/compare/new-parallelization...d-torrance:M2:parallelization for the differences. Why not just add your last 2 commits to the parallelization branch?

The 8 points above were based on notes I sent you, which were supposed to just be notes/issues that you should be aware of and ideally discuss with someone, if/when you do merge. I guess some version of some of them should also go in Macaulay2's release notes. I would add here:

1. Changed GCstats()

2. TODO: comment out all GC_FREE()s!?

3. Need to update bdwgc submodule & include my bug fix

4. setIOUnSyncronized (a shared buffer without mutexes) and setIOExclusive (separate buffers that lose data) are totally broken, leaving only setIOSynchronized. These need to be eliminated or totally redesigned and reimplemented.

Good luck! Hope this helps, Dave

d-torrance:
It's just the `parallelization` branch rebased on top of `development`.  I don't think I missed any commits, but I'll double check!

d-torrance:
It looks like we have two failing tests:

```m2
i1 : check(80, "Complexes", Verbose => true)
 -- running   check(80, "Complexes")             
 ulimit -c unlimited; ulimit -t 700; ulimit -m 850000; ulimit -s 8192; ulimit -n 512;  cd /tmp/M2-2402714-0/1-rundir/; GC_MAXIMUM_HEAP_SIZE=400M "/home/profzoom/src/macaulay2/M2/M2/BUILD/doug/usr-dist/x86_64-Linux-Ubuntu-23.10/bin/M2-binary" --no-randomize --no-readline --silent --stop --print-width 77 -e 'needsPackage("Complexes",Reload=>true,FileName=>"/home/profzoom/src/macaulay2/M2/M2/Macaulay2/packages/Complexes.m2")' <"/tmp/M2-2402714-0/0.m2" >>"/tmp/M2-2402714-0/0.tmp" 2>&1
/tmp/M2-2402714-0/0.tmp:0:1:(3):[12]: (output file) error: Macaulay2 exited with status code 1
/tmp/M2-2402714-0/0.m2:0:1: (input file)
M2: *** Error 1
 -- 2.46926 seconds elapsed; 0.00194321 process (cpu time); 0.00167225 thread; 0. gc (process)
../../macaulay2/M2/M2/Macaulay2/packages/Complexes/FreeResolutionTests.m2:710:1-738:1 error:
 -- 
 --           1      4      10      11      5      1
 -- o9 = (0, R  <-- R  <-- R   <-- R   <-- R  <-- R )
 --                                                
 --          0      1      2       3       4      5
 -- 
 -- o9 : Sequence
 -- 
 -- i10 :   assert BinaryOperation(symbol >, usedtime3, 5*usedtime2)
 -- stdio:20:3:(3): error: assertion failed:
 -- 0>0 is false
 -- 
stdio:1:1:(3): error: test(s) #80 of package Complexes failed.
```

This is probably due to the timing changes.

Also, we end up hanging at:

```m2
i1 : check(6, "NumericalImplicitization")
 -- capturing check(6, "NumericalImplicitization")
```

mahrud:
Couple of feedback:
1. can we keep `elapsedTime` as is and add the new functionality to `time` (and `timing`) instead?
2. I sometimes use `cpuTime` for benchmarking; is there a way to get the thread cpu time now?
3. how does `taskResult` wait for the task to finish? assuming it is not running a loop that repeatedly checks if the task is finished, can we add a synonym `await` and also add `async = f -> (x -> schedule(f, x))`? This is the [Async/Await](https://en.wikipedia.org/wiki/Async/await) pattern. It would be great if these were keywords instead
4. instead of `parallelApply` can we make a keyword which can turn a normal apply to run in parallel? (similar to OpenMP patterns)
```m2
parallel apply(...)
```

mahrud:
> * `time` and `timing` involve a data structure, which would be a breaking change. `elapsedTime` just prints stuff. At least that was my thinking.

I don't think this is right. `time` and `elapsedTime` simply print a message, `timing` and `elapsedTiming` return a data structure with the timing info. My primary ask was to keep `elapsedTime` intact and change the output of `time` instead (both without any data structures involved`. I think this is more appropriate because `elapsedTime` is the "wall clock" so it shouldn't depend on threads at all, whereas `time` measures cpu time so it should be thread specific.

The secondary ask was to also return the gc and thread time in the data structure returned by `timing`, but that doesn't feel very urgent and as you said would be a slight breaking change.

> * The new `elapsedTime` prints the time used by the current thread.

I think this is confusing and potentially misleading.

DaveBarton:
You're right, `timing` and `elapsedTiming` return a data structure, I misspoke.

"cpu time", by standard usage, counts the total cpu time used by all cores in a calculation. If you have a parallel algorithm that hands off the real work to other threads, the current thread's cpu time may be very small. If you don't use multiple cores, the total cpu time should be approximately equal to the current thread's cpu time. Thus it seems like `cpuTime()` should return the total for all cores, and this is indeed what you should use for benchmarking. (If a lot of time is being wasted by the task system on threads that don't compute anything useful, that should be fixed, but not by just returning one thread's total time.) Also BTW, the gc time can be gotten programmatically now by the new `GCstats()`.

I see your point though, which function should show the extra info?

Here's the doc for the new `elapsedTime`:
```
     Key => "elapsedTime",
     Headline => "time a computation including time elapsed",
     Usage => "elapsedTime e",
     TT "elapsedTime e", " evaluates ", TT "e", ", prints the amount of time
     elapsed, and returns the value of ", TT "e", ". The cpu time used by the Macaulay2
     process, the current thread, and garbage collection during the evaluation of ", TT "e",
     " is also shown.",
     EXAMPLE "elapsedTime sleep 1",
```

Here's Doug's output above from a failing test:
```
 -- 2.46926 seconds elapsed; 0.00194321 process (cpu time); 0.00167225 thread; 0. gc (process)
```

It does take a bit of getting used to, for instance the gc time is total process (cpu) time, which is divided pretty evenly between the threads. One upshot is that I think you'll all get a better feel for how much time gc is taking for things, and how it's being spread out. But maybe this could all be presented or documented better.

My thinking about `time` vs. `elapsedTime` though was that if you want to see just one number, it should be total cpu time (all tasks, including gc). That's the basic `time` you should be benchmarking most often. If you want details, you want all 4 numbers. But just seeing a wall clock time (having `elapsedTime` show just that) is often misleading: it depends on how loaded your machine is, and how many threads you're using for this computation. So to my mind, it does depend on the number of threads. I guess you could have `time` and `elapsedTime` both show all 4 numbers, but that seems overkill and redundant to me. But anyway it's up to you, do whatever you guys thinks is best.

Thanks for the feedback!

mahrud:
> Here's Doug's output above from a failing test:

I definitely appreciate the additional information (which I had asked for in the past) but this change has messed up the formatting of the output of `installPackage` and `check`:
image
image

> My thinking about `time` vs. `elapsedTime` though was that if you want to see just one number, it should be total cpu time (all tasks, including gc)

I think most M2 users don't care about benchmarking, but they do care about wall time (e.g. if I'm running an algorithm that took an hour last time, I should come back to check on it in an hour).

Consider this example:
```m2
i3 : time capture examples minimalBetti;
     -- used 12.4161 seconds

i4 : elapsedTime capture examples minimalBetti;
 -- 10.085 seconds elapsed; 12.2704 process (cpu time); 10.0572 thread; 0. gc (process)
```
In reality, this tool 10 seconds to run on my machine. If I run this again, it won't be on a different cpu or different setup, so I would expect it to take about 10 seconds. Also note that the cpu time that is now added is simply the usual output of `time`, which is redundant, so I think the output should instead be:
```m2
i3 : time capture examples minimalBetti;
 -- used 12.4161 seconds (cpu time); 10.0572 thread; 0. gc (process)

i4 : elapsedTime capture examples minimalBetti;
 -- 10.085 seconds elapsed
```

Also, I think having `cpuTime()` return the total cpu time is fine, but for my purposes I would still like a `threatTime()`, or perhaps we could have `cpuTime` take an optional integer to return the thread-specific cpu time. This would be useful along with an integer parameter like `currentThread` so that `cputTime(currentThread)` is always the current thread's cpu time.

d-torrance:
Locally, the two failing tests are working when I don't use `capture`.  So I've added `no-capture-flag` to them both.  Let's see if the GitHub builds work now...

DaveBarton:
I'm getting your point, thanks. There are over 1000 calls in M2 to each of `time` and `elapsedTime`. Maybe we'll mess up formatting or shock folks if we grow the output of either one, especially since almost no one will be using parallelism at first (except for gc)? We could have an `allTimes` or `showTimes` or `timeInfo` (too bad just `times` is already used for something) that prints the 4 times. Maybe the best order would be: cpu time, gc time (before elapsed time since it's a total cpu time across tasks), elapsed time, time on this thread. I did want people to learn about gc time though, sigh.

Could the formatting changes be messing up Doug's tests? (Sorry - not really - for the newbie questions.)

P.S. The counter-argument to adding a new function is I guess if you think enlarging the output of `time` wouldn't mess up any formatting, and if you thought everyone (or most folks) that call `time` really would like to see the extra output, e.g. gc time and elapsed time.

d-torrance:
Avoiding `capture` fixed the `NumericalImplicitization` test, but not the `Complexes` one.  (Plus it was already not using `capture` because it uses `gbTrace`, so adding `no-capture-flag` was redundant.)

mahrud:
At least in the Core, there are only two calls to `elapsedTime` (one for examples and one for tests) and zero calls to `time`. Packages use both of course, but I can't think of any other cases where formatting would break because of either.

I don't think there's a need to add a new keyword function, especially since changing `time` would also teach people who use it that the number displayed is cpu time.

d-torrance:
The builds are all working!  I'll convert this to a draft until the other comments are addressed.

d-torrance:
I've moved the thread and gc time over to `time` as discussed:

```m2
i1 : time capture examples minimalBetti;
 -- used 5.18674 seconds (cpu time); 4.60075 (thread); 0. gc (process)

i2 : elapsedTime capture examples minimalBetti;
 -- 5.18838 seconds elapsed
```

It also now prints to `stderr` instead of `stdout` for consistency with `elapsedTime` and most other logs and warning messages.

d-torrance:
The `time` changes were causing the `ComputationsBook` tests to fail, so I've pushed a commit fixing that.  I also updated the output slightly:

```m2
i1 : time capture examples minimalBetti;
 -- used 5.11699 s (cpu); 4.43841 s (thread); 0. s (gc)
```
This way, there are fewer total characters, but we also have units on all three times.

mahrud:
Is there a space before the "s"?

d-torrance:
> Is there a space before the "s"?

There is -- I think that's the SI convention -- but I suppose we could save three more characters by removing them.

DaveBarton:
Doug, thanks for all the test fixes and coding the time vs. elapsedTime fix!

I didn't know about the "raw" naming rule. Feel free to choose a better name.

mahrud:
> There is -- I think that's the SI convention -- but I suppose we could save three more characters by removing them.

Oh interesting. I don't care about the savings though: I would pick `0.0s` over the `0. s`, which looks a bit weird. Unix `time` command outputs like
```
real	0m0.007s
user	0m0.001s
sys	0m0.006s
```
So I feel like the space is unnecessary, unless there's any specific benefit to it.

d-torrance:
Dropping the space sounds good to me.  I agree that `0.` looks strange.

d-torrance:
I've removed the space and also simplified the printing of doubles.  Now we just use `snprintf` from the C standard library instead of rolling our own function (`tostring5`, which according to the git history, used to be used by `format`, but more recently has *only* been used for printing doubles).

So now we have the following:
```m2
i1 : time capture examples minimalBetti;
 -- used 5.96428s (cpu); 4.31929s (thread); 0s (gc)
```

DaveBarton:
Maybe just 3 or 4 digits of precision (%.3g or %.4g I think) instead of 6? For both `time` & `elapsedTime`?

d-torrance:
I'm not sure about changing the precision.  Using `%g` seems to keep things pretty consistent with the current behavior.  Also, this routine is for turning *any* double into a string, not necessarily just timing results.  (Although I'm not sure if it's even called outside timing results or not -- the vast majority of the floating point numbers in Macaulay2 are `mpfr_t` objects, not doubles.)

M2/Macaulay2/c/README
```diff
@@ -266,7 +266,7 @@ Expressions:
 	if a then b
 	until a do b
 	while a do b			evaluates b as long as a is true
-	while a list b			produces a list of the values of b, as long as a is true
+	while a list b			produces a list of the values of b, as long as a is true (not implemented!?)
```
mahrud, line 269: I think this was meant to be `while a do provide b`.

M2/Macaulay2/c/README
```diff
@@ -289,7 +289,7 @@ Expressions:
 
 	export a := b				export a variable or function (by declaring it in *.sig)
 	import a				import a variable (used mainly in *.sig files)
-	thread a := b				declare a thread local variable
+	threadLocal a := b			declare a thread local variable
```
mahrud, line 292: Can we also change the top level name to `threadLocal` instead of `threadVariable`? It'll be a breaking change, but lots of other things are changing with parallelism anyway. (I should say, I would prefer making them both just be `thread` for brevity, but I particularly dislike `threadVariable`)

M2/Macaulay2/c/README
DaveBarton: There's also [`while p list x do z`](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_while.html) in top-level m2 [which you probably already know], which does create a list of `x` values.

M2/Macaulay2/d/M2mem.c
```diff
@@ -80,7 +80,7 @@ void freemem(void *s) {
 #ifdef MEMDEBUG
      M2_debug_free(s);
 #else     
-     GC_FREE(s);
+     /* GC_FREE(s); */
```
mahrud, line 83: I'm a little confused. Doesn't this make `freemem` and `freememlen` entirely useless? Also, keeping `trapchk` and `M2_debug_free` there would mean when compiling in memory debug mode you get in traps in places that normally wouldn't call `GC_FREE` anymore, so shouldn't those be commented too?

M2/Macaulay2/d/pthread.d
```diff
@@ -218,6 +220,16 @@ setupfun("setIOSynchronized",setIOSynchronized);
 setupfun("setIOUnSynchronized",setIOUnSynchronized);
 setupfun("setIOExclusive",setIOExclusive);
 
+export getIOThreadMode(e:Expr):Expr := (
```
mahrud, line 223: How can I use this info at the top-level?

d-torrance: This returns 0, 1, or 2, depending on whether it's unsynchronized, synchronized, or exclusive, respectively.  Previously, 0 was the default, but in this PR, the proposed default is 1.

Here's an example:

```m2
i1 : getIOThreadMode() -- the new default, try to synchronize how threads print

o1 = 1

i2 : scan(20, i -> schedule(() -> print "Hello, world!"))
Hello, world!
Hello, world!
Hello, world!
Hello, world!
Hello, world!
Hello, world!

Hello, world!
i3 : Hello, world!
Hello, world!
Hello, world!Hello, world!
Hello, world!
Hello, world!
Hello, world!
Hello, world!

Hello, world!Hello, world!

Hello, world!Hello, world!

Hello, world!
setIOExclusive() -- only the main thread can print

i4 : getIOThreadMode()

o4 = 2

i5 : scan(20, i -> schedule(() -> print "Hello, world!")) -- notice nothing is printing (yet)

i6 : setIOUnSynchronized() -- go crazy!  the current default, which leads to fun things like this:

i7 : Hello, world!
Hello, world!Hello, world!
Hello, world!Hello, world!
Hello, world!Hello, world!
Hello, world!Hello, world!
Hello, world!Hello, world!
Hello, world!Hello, world!
Hello, world!Hello, world!
Hello, world!Hello, world!
Hello, world!Hello, world!
ello, world!

internal error: beginning of line marker not within buffer
internal error: beginning of line marker not within buffer
internal error: beginning of line marker not within buffer
internal error: beginning of line marker not within buffer
Aborted (core dumped)

Process M2 exited abnormally with code 134
```
As Dave mentioned above, this is all a mess and should be scrapped eventually.  But I figured it would be nice to at least have a function that lets us find out the current thread mode for the time being.

mahrud: Thanks! I see, I didn't know you could change the thread mode this at the top level. I agree that the whole printing system is not really suited for threads.

M2/Macaulay2/d/M2mem.c
DaveBarton: Right, this makes `freemem()`/etc. into noops. But Mike has talked about removing gc from the engine, so you might change your mind later, and want the `freemem()` calls to still exist. Also, I don't really know how Macaulay2 debugging works or is used. I tried to note in a TODO in this github discussion that someone should ideally look at all the calls to `GC_FREE()` and decide what to do. But you certainly want to comment out the major calls to `GC_FREE()` at least, and it helps (a lot with parallelism) not hurts, so I did that.

d-torrance:
Just pushed a few more commits un-exporting `parallelApplyRaw` and renaming `threadVariable` to `threadLocal`.

DaveBarton:
Maybe rename `parallelApplyRaw` instead of un-exporting it?

d-torrance:
I added a `Strategy => "raw"` option to `parallelApply` that calls `parallelApplyRaw`directly without needing to export it.

At this point, I think we've addressed most of the comments, so removing the draft status.

PR #3189, opened 2024-04-25
Remove unncessary hash counter increments

pzinn:
Right now every call to `newHashTable` will increase the hash counter, even if it is not needed -- say if the hash table is unmutable or a cache. This causes massive inflation of the hash counter, leading to the dreaded `hash code serial number counter overflow`. The simple fix in this PR empirically divides by 4 the hash counter increase in examples.

### DIFF
--- a/M2/Macaulay2/d/expr.d
+++ b/M2/Macaulay2/d/expr.d
@@ -164,11 +164,15 @@ export newHashTable(Class:HashTable,parent:HashTable):HashTable := (
 	  -- we start with four empty buckets.  It is important for the 
 	  -- enlarge/shrink code in hashtable.dd that the number of buckets
 	  -- (here four) is a power of two
-	  Class,parent,0,nextHash(),
+	  Class,parent,0,0,
 	  true,				  -- mutable by default; careful: other routines depend on this
 	  false,
 	  uninitializedSpinLock
 	  ); init(ht.mutex); ht);
+export newHashTableWithHash(Class:HashTable,parent:HashTable):HashTable := (
+       ht:=newHashTable(Class,parent);
+       ht.hash=nextHash();
+       ht);
 
 export newCompiledFunction(fn:fun):CompiledFunction := (
     cf := CompiledFunction(fn, 0);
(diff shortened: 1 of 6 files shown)


mahrud:
But that's how `youngest` decides which hash table is younger, isn't that going to break?

pzinn:
This PR doesn't actually change the ordering of hash numbers; it simply removes unnecessary hash increments. i.e., unmutable hash tables and cache tables are unaffected by it; the *order* of the hashes of mutable hash tables is also preserved, but they're more densely packed.

mahrud:
Ah right, I forgot we only care about the age of mutable hash tables. This is great. I had noticed some strange things about how they increment. For instance, notice how it goes up by two:
```m2
i1 : print hash new MutableHashTable from {};
1132881

i2 : print hash new MutableHashTable from {};
1132883
```
Why the extra bump? But wait it get's worse, if I remove the semicolon it jumps by three!
```m2
i3 : hash new MutableHashTable from {}

o3 = 1132885

i4 : hash new MutableHashTable from {}

o4 = 1132888
```
Now if I do something to bump the line number by one, it still bumps by three:
```m2
i5 : null

i6 : hash new MutableHashTable from {}

o6 = 1132891
```
But if I even enter "0" it bumps by four!!
```m2
i7 : 0

o7 = 0

i8 : hash new MutableHashTable from {}

o8 = 1132895
```
Why does this happen?

Does your PR help with any of these?

pzinn:
it does:
```
i1 : print hash new MutableHashTable from {};
1130098

i2 : print hash new MutableHashTable from {};
1130099

i3 : hash new MutableHashTable from {}

o3 = 1130100

i4 : hash new MutableHashTable from {}

o4 = 1130102

i5 : null

i6 : hash new MutableHashTable from {}

o6 = 1130104

i7 : 0

o7 = 0

i8 : hash new MutableHashTable from {}

o8 = 1130107
```
the reason there are still some gaps is that the past output gets its symbols say `o3`.

M2/Macaulay2/d/expr.d
```diff
@@ -164,11 +164,15 @@ export newHashTable(Class:HashTable,parent:HashTable):HashTable := (
 	  -- we start with four empty buckets.  It is important for the 
 	  -- enlarge/shrink code in hashtable.dd that the number of buckets
 	  -- (here four) is a power of two
-	  Class,parent,0,nextHash(),
+	  Class,parent,0,0,
```
d-torrance, line 167: Are there any hash tables that might sneak by with a hash of 0 after this change?  Or is it always set to something else elsewhere?

pzinn: this needs to be checked carefully, but I believe the answer is yes. `newHashTable(...)` is called by `makenew` which sets the hash, and by a bunch of other functions in say `actors4.d` and `evaluate.d` which call `sethash` afterwards.

PR #3193, opened 2024-05-02
New operator for restriction

mahrud:
- **added binary op |_**
- **documented the new operators**


M2/Macaulay2/d/binding.d
```diff
@@ -536,12 +538,13 @@ offset := 0;
...
     -- update the subtracted number here to match number of operators below
-    len length(opsWithBinaryMethod) - 17 at i
+    len length(opsWithBinaryMethod) - 18 at i
     do (
 	-- to avoid ambiguity and syntax errors, we don't create augmented
 	-- assignment operators for ==, <==, <===, :, SPACE, or, and, xor, ?
-	-- also _< _<= _> _>= ^< ^<= ^> ^>=
+	-- also |_ _< _<= _> _>= ^< ^<= ^> ^>=
 	while (
+	    opsWithBinaryMethod.(i + offset) === BarUnderscoreS           ||
```
d-torrance, line 547: Why exclude this operator from having an augmented version?

mahrud: I feel like it's the kind of mathematical notation that doesn't really lend itself to augmentation. `F|_=Z` seems strange and potentially misleading, no?

d-torrance: My thinking when adding the augmented assignment operators was that every flexible binary operator should get one unless it was going to create some ambiguity, e.g, we already had `<===` so we shouldn't add one for `<==`.  Anyone can assign methods for any flexible binary operator in their code, and perhaps augmented assignment might make sense for their method, even if it doesn't make sense for any methods installed for that operator in `Core`.

For this particular operator, am I correct that the intended behavior is that if `F` is a matrix, then `F|_Z` would return a matrix with source `Z` (assuming that `Z` is a submodule of `source F`)?  If so, then perhaps maybe someone doesn't really care about the original `F` and only plans on using the restriction going forward.  Then `F|_=Z` might be useful.

mahrud: The application I have in mind is for restriction of sheaves: if F is a sheaf on X and Y is a closed subvariety (e.g. a curve embedded in a projective space) then `F|_Y` will be a sheaf on `Y` which is the restriction.

PR #3197, opened 2024-05-03
Certify packages published in JSAG volume 14

d-torrance:
We certify the following packages that were published in JSAG volume 14 ahead of the next release:

* `CotangentSchubert`
* `GeometricDecomposability`
* `InvariantRing` (this is a re-certification -- an earlier version was previously published In JSAG)
* `MultiplicitySequence`
* `Probability`
* `TropicalToric` *(new package)*

I had to make one small change from the published version to get `TropicalToric` to load without errors, likely due to the recent `Varieties` package changes.

The JSAG website says that volume 14 is "not yet complete".  I'll keep my eyes peeled and add any new ones that pop up before the release.

### DIFF
--- a/M2/Macaulay2/packages/TropicalToric.m2
+++ b/M2/Macaulay2/packages/TropicalToric.m2
@@ -23,7 +23,20 @@ newPackage(
 	AuxiliaryFiles => true,
 	CacheExampleOutput => true,
   OptionalComponentsPresent => true,
-  UseCachedExampleOutput => true
+  UseCachedExampleOutput => true,
+  Certification => {
+      "journal name" => "Journal of Software for Algebra and Geometry",
+      "journal URI" => "https://msp.org/jsag/",
+      "article title" => "Tropical computations for toric intersection theory in Macaulay2",
+      "acceptance date" => "2023-09-14",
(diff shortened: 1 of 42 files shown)


M2/Macaulay2/packages/TropicalToric.m2
```diff
@@ -0,0 +1,98 @@
...
+    "Package$gfanInterface"
+  },
+  PackageImports => {
+    "FourierMotzkin"
+  },
+	DebuggingMode => true,
+	AuxiliaryFiles => true,
+	CacheExampleOutput => true,
+  OptionalComponentsPresent => true,
+  UseCachedExampleOutput => true,
```
mahrud, line 26: Why is this using cached examples?

d-torrance: Good question.  I just copied this verbatim from JSAG.  I don't think it needs it.

mahrud: I realized it technically needs Tropical, which needs Polymake. `OptionalComponentsPresent` should probably be transitive, no?

d-torrance: Ah, ok.  Yes, that would definitely make sense.

Issue #3198, opened 2024-05-04
Precedence of subscripts and superscripts


mahrud:
There's a comment in `HyperplaneArrangements` which I think we should fix by changing the precedence of _ and ^:
```m2
Flat ^ Flat := Flat => meet  -- ooh, cool.  But note L_1^L_2 isn't L_1^(L_2) !
```

--
Initially this issue was the following, but I switched it because I think the above is much more important.
> `join` for posets is called `posetJoin`, but would be more intuitive if `join` was available as  a method. Similarly, flats in `HyperplaneArrangements` have `meet` (or `^`) and `vee` (or `|`) rather than `join`.

mahrud:
Similarly, flats in `HyperplaneArrangements` have `meet` (or `^`) and `vee` (or `|`) rather than `join`.

Also, there's a comment there which I think we could fix by changing the precedence, if it doesn't break anything else:
```m2
Flat ^ Flat := Flat => meet  -- ooh, cool.  But note L_1^L_2 isn't L_1^(L_2) !
```
cc: @d-torrance do you think this makes sense?

d-torrance:
Yes, I think it would make sense if `_` and friends had higher precedence than  `^` and friends.  Subscripting having higher precedence than exponentiation seems to be common in other languages -- I checked R, Python, and Javascript.

d-torrance:
> **Written by Claude** (Claude Opus 5, via Claude Code), not by @d-torrance, whose account posted it -- please weigh it accordingly.

Closing as `wontfix`, on a ground the thread never got to: the change would break `dd^C_i`.

That idiom — the `i`-th differential of a complex — works *because* `^` binds tighter than `_`, so
`dd^C_1` parses as `(dd^C)_1`. Under the proposed precedence it would parse as `dd^(C_1)`. It
appears **323 times across 23 files** in the distributed packages and Core, with another 95
occurrences in the `Complexes` documentation, and it turns up in essentially every piece of
workshop and textbook material that touches complexes.

@d-torrance's point above is sound in isolation — R, Python and JavaScript do all subscript before
exponentiating — but M2 has built a very widely used notation on the current order, and changing it
would silently alter the meaning of existing code rather than erroring on it.

Worth recording that the current behaviour does *not* silently compute the wrong thing. On 1.26.06:

```m2
i9 : y_1^y_2

stdio:8:3:(3):[1]: error: no method for binary operator ^ applied to objects:
            y  (of class S)
             1
[...]

PR #3199, opened 2024-05-04
Give subscript operators higher precedence than superscript ones

d-torrance:
This is a response to https://github.com/Macaulay2/M2/issues/3198#issuecomment-2094071558.

In many languages (or at least the three I checked -- Python, R, and Javascript), subscripting/indexing operators have higher precedence than exponentiation.  But to this point in Macaulay2, they have the same precedence.

We re-order the precedence so that all of the subscript operators (`.`, `#`, `_`, etc.) have higher precedence than  the superscript ones (`^`, etc.)

### Before

```m2
i1 : L = {1,2,3,4}

o1 = {1, 2, 3, 4}

o1 : List

i2 : L#1^L#2
stdio:2:4:(3): error: no method for binary operator ^ applied to objects:
--            2 (of class ZZ)
--      ^     {1, 2, 3, 4} (of class List)
```

### After
```m2
i2 : L#1^L#2

o2 = 8
```


d-torrance:
After playing around with this for a bit, I think this is too much of a breaking change.  Things like `dd^C_i` are everywhere in the `Complexes` package, and this should be parsed like `(dd^C)_i`, not `dd^(C_i)`.

mahrud:
Was Complexes the main issue? I think you've convinced me that this is something that should definitely be fixed in M2, so I'm wondering if there's a workaround for `dd^C_i`.

mahrud:
cc: @mikestillman @ggsmith

d-torrance:
Yeah, `Complexes` seemed to be the main issue that I ran into.  There were a few other instances `A^i_j` that needed changing to `(A^i)_j` (mostly columns of powers of matrices and generators of Cartesian powers of modules), and this seems reasonable since the notation is otherwise ambiguous.  But `dd^C_i` is pretty ubiquitous in `Complexes`.

mahrud:
Fortunately `dd^C_i` is a new notation, so it's not used very widely. I'm a bit confused reading the [documentation](https://macaulay2.com/doc/Macaulay2-1.23/share/doc/Macaulay2/Complexes/html/_differential_spof_spa_spchain_spcomplex.html), since `dd_C` is also listed, but doesn't work.

Issue #3200, opened 2024-05-04
"published article DOI" unused for certified packages


d-torrance:
In https://github.com/Macaulay2/M2/commit/337ddb0d1ee181d7ef93b7857ab196f3299f9e13, "published article DOI" was added to the template for certified packages.  However, this field currently isn't  used anywhere when generating the documentation.  Was there a plan to add support at some point?

Also, packages that do use it are inconsistent -- some include "https://dx.doi.org/"  and some don't.

d-torrance:
Fixed in #3204

PR #3204, opened 2024-05-05
Add link to DOI for certified packages in documentation

d-torrance:
We add a link to the DOI for each certified package to the documentation.  We also drop the "https://doi.org" where present in some packages for consistency and add DOI's for packages that didn't have them.

Closes: #3200


DanGrayson:
The test builds are failing...

d-torrance:
Oops, I was working from an older copy of M2 without the new `TropicalToric` package, so its DOI was missing.  I'll fix it here in a moment.

PR #3209, opened 2024-05-08
added union, intersect, intersection for sets

mahrud:
Also removed (unique, Sequence), which returned a list!


DanGrayson:
Test builds are failing.

mahrud:
Also, I think this is rather confusing:
```m2
i12 : set {"a", "A", "B", "b"}

o12 = set {A, a, B, b}

o12 : Set

i13 : keys oo

o13 = {a, A, b, B}

o13 : List
```

PR #3212, opened 2024-05-08
Build Normaliz 3.10.2

d-torrance:
Normaliz 3.10.2 was released on March 1 -- see https://github.com/Normaliz/Normaliz/releases/tag/v3.10.2.  It appears to be compatible with Macaulay2.  I uploaded it to Debian shortly after it was released and nothing has broken yet.  :)

This PR updates both the autotools and cmake builds to build this latest version when Normaliz isn't already present on the user's system.

Note that the GitHub builds don't actually test this code, as they install Normaliz via apt or homebrew, but I've tested both builds locally and they worked.

Cc: @ggsmith


mahrud:
Ah I was gonna say nauty also needs to be updated.

PR #3213, opened 2024-05-09
New MultigradedBGG Package

sashahbc:
This package implements the multigraded BGG functors, including an implementation of differential modules, their minimal free resolutions, and a method for computing strongly linear strands of multigraded free resolutions.

PR #3223, opened 2024-05-10
Add MultigradedBGG package

d-torrance:
This is just #3213 again.  I was trying to rebase it on top of `development` after #3219 was merged to fix the failing tests, but somehow I ended up closing the PR, and then it wouldn't let me push to the same branch again.

Cc: @sashahbc


sashahbc:
Thanks for the help Doug!

PR #3224, opened 2024-05-11
Fix a bug in submatrix when source is a subquotient

mahrud:
- **added and documented all(BasicList)**
- **fixed all(List,List,Function) to avoid creating new list**
- **fixed a bug in submatrix when source is a subquotient**

See the added test for the case that was failing before.


mahrud:
I skipped the github tests on the last commit because it's only rewriting a test so nothing would change.

PR #3233, opened 2024-05-15
Deprecate support for MPIR

mahrud:
- **deprecated support for MPIR**


d-torrance:
Why FLINT 2.8.5?  Ubuntu 22.04 still has FLINT 2.8.4, and Macaulay2 still builds fine using FLINT 2.6.3 (which I backported for the  Ubuntu 18.04 and 20.04 PPA's).

DanGrayson:
3  test builds are failing

DanGrayson:
Could you fix the failing test builds?

d-torrance:
Looks good to me, but I suppose we should probably do the same thing for the autotools build.  There's still a bunch of mpir cruft in configure.ac, for example.  I can work on it and push to your branch, if you'd like.

mahrud:
Sounds good!

d-torrance (approved):
Autotools build should be MPIR-free now too.

M2/LICENSING-NOTES
```diff
@@ -49,4 +48,4 @@ modifications by Paul Zinn-Justin			public domain
 
 for M2 binaries:
 
-We are prevented from using GPL 2 by mpir and mpfr
+We are prevented from using GPL 2 by mpfr
```
mahrud, line 51: Huh, [mpfr.org](https://www.mpfr.org/) says:
> MPFR is free. It is distributed under the [GNU Lesser General Public License](https://www.gnu.org/licenses/lgpl-3.0.html) (GNU Lesser GPL), version 3 or later (2.1 or later for MPFR versions until 2.4.x).

I didn't know this file existed.

d-torrance: Apparently GPL-2 and LGPL-3 aren't compatible: https://www.gnu.org/licenses/gpl-faq.html#AllCompatibility

mahrud: Ah, I see.

PR #3241, opened 2024-05-16
Add check to see if we need to link against -latomic

d-torrance:
This is necessary to handle 64-bit atomics (which are used by TBB) on certain 32-bit architectures (e.g., powerpc and m68k) due to a [GCC bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81358).

Otherwise, we end up with the following linking error ([full log](https://buildd.debian.org/status/fetch.php?pkg=macaulay2&arch=powerpc&ver=1.24.05%2Bds-1&stamp=1715757140&raw=0)):

```
/usr/bin/ld: ../e/NCAlgebras/NCF4.o: undefined reference to symbol '__atomic_fetch_add_8@@LIBATOMIC_1.0'
/usr/bin/ld: /lib/powerpc-linux-gnu/libatomic.so.1: error adding symbols: DSO missing from command line
```

@mahrud: Do you think it's worth adding an equivalent check to the cmake build?  This is only an issue on certain old-school 32-bit architectures.

### DIFF
--- a/M2/configure.ac
+++ b/M2/configure.ac
@@ -1863,6 +1863,21 @@ AC_LINK_IFELSE([AC_LANG_SOURCE([int main() {}])],
 rm -f conftest.mapfile
 LDFLAGS=$SAVE
 
+# https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81358
+AC_MSG_CHECKING([whether we need to link with -latomic])
+AC_LANG([C++])
+AC_LINK_IFELSE([AC_LANG_PROGRAM([
+	#include <atomic>
+	], [std::atomic<int64_t> x; x++])],
+	[AC_MSG_RESULT([no])],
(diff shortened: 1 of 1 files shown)


mahrud:
Unless you want to try using the CMake build on those architectures, I doubt it. At some point I was under the impression that the plan was to eventually fully transition to the CMake build, but that doesn't seem to be the case. It seems silly to have separate build systems.

d-torrance:
Yeah, it definitely is a lot of extra work trying to make things work for both build systems!

I use the autotools build for the Debian packages because that's what I'm familiar with.  I took a stab at getting it to work with the cmake build a couple years ago but never finished.  Maybe I'll give it another go this summer.

mahrud:
Happy to take a look if you run into any unusual issues. Both the brew formula and the actions workflow should have all the steps.

Issue #3254, opened 2024-05-22
Depth off-by-one when ring includes "a" as variable


ab2422:
When using the `depth` function from the Depth package, the depth of a ring seems to be one larger than it is supposed to be whenever the ring is a quotient ring of a polynomial ring that includes "a" as a variable. In particular, it is sometimes giving that depth is larger than dimension! Some small examples:
```
i1 : needsPackage("Depth");
i2 : depth(GF(2)[x]/ideal(x^2))
o2 = 0
i3 : depth(GF(2)[a]/ideal(a^2))
o3 = 1
i4 : depth(GF(2)[b]/ideal(b^2))
o4 = 0
```

A larger example:
```
i5 : depth(GF(5)[x,y,z]/ideal(x^3+y^3+z^3))
o5 = 2
i6 : depth(GF(5)[a,b,c]/ideal(a^3+b^3+c^3))
o6 = 3
```

But notice a polynomial ring itself using "a" works fine:
```
i7 : depth(GF(5)[a,b,c])
o7 = 3
```
This issue of course carries down to other functions that use `depth`, such as `isCM`.

d-torrance:
I think the problem isn't with the `Depth` package, but with Galois fields.  It works just fine with other coordinate fields:

```m2
i2 : depth(QQ[a]/ideal(a^2))

o2 = 0

i3 : depth(ZZ/2[a]/ideal(a^2))

o3 = 0
```
The  following code is likely to blame:
https://github.com/Macaulay2/M2/blob/ec9e9ac60ed4a8e791448942202f077a88a87e15/M2/Macaulay2/m2/galois.m2#L99

d-torrance:
After playing around with this, I think this is actually the intended behavior, but the documentation for `GF` should probably be improved to explain what's going on.

If `GF` is called without the `Variable` option, then `a` is set globally as a generator of the field.  So we shouldn't be using it as a variable for a polynomial ring over the field.

```m2
i1 : GF 2

o1 = GF 2

o1 : GaloisField

i2 : a

o2 = 1

o2 : GF 2
```

d-torrance:
@ab2422 - Do you think the fix in #3259 (adding documentation as to why the variable `a` is special when using Galois fields) is sufficient to close this issue?

d-torrance:
Related: #897

ab2422:
@d-torrance Sure, the documentation fix is helpful & seems good enough to close. Thanks for looking at it!

PR #3255, opened 2024-05-22
Use local variable for Galois field instead of getSymbol

d-torrance:
Otherwise, we run into issues when "a" is used elsewhere.  After the fix:

```m2
i2 : depth(GF(2)[a]/ideal(a^2))

o2 = 0

i3 : depth(GF(5)[a,b,c]/ideal(a^3+b^3+c^3))

o3 = 2
```

Closes: https://github.com/Macaulay2/M2/issues/3254


d-torrance:
It turns out that this is kind of a breaking change because a couple packages assumed that `a` is always available as a generator of the field after calling `GF`.  But this was undocumented and led to bugs.

d-torrance:
I'm closing this -- this is too much of a breaking change.  I think maybe `GF` just needs better documentation.

PR #3257, opened 2024-05-22
add CoefficientRing to smoothFanoVariety(d,0)

Devlin-Mallory:
In NormalToricVarieties, smoothFanoToricVariety(d,i) took a CoefficientRing option, which was previously erroneously ignored when i = 0 (i.e., for P^d). This just adds CoefficientRing => opts.CoefficientRing in that case.


mahrud:
Why not also pass the variable option? You can just pass `opts` which includes both.

Devlin-Mallory:
> Why not also pass the variable option? You can just pass `opts` which includes both.

Ah, good call - added that now.

PR #3259, opened 2024-05-23
Document use of global variable "a" as generator of Galois field

d-torrance:
We update the `GF` and `finite fields` documentation nodes to mention the fact that if `GF` is called without the `Variable` option, then the variable `a` is used as the generator of the field.  My earlier attempt at changing the behavior led to lots of failing tests, so this behavior is widely used -- see #3255.

This was previously undocumented, leading to confusion trying to construct polynomial rings over Galois fields with `a` as a variable as in #3254.

We also link these two documentation nodes to one another.

### DIFF
--- a/M2/Macaulay2/packages/Macaulay2Doc/functions/GF-doc.m2
+++ b/M2/Macaulay2/packages/Macaulay2Doc/functions/GF-doc.m2
@@ -10,7 +10,9 @@ document {
 	  Usage => "GF(p,n)\nGF(q)",
 	  Inputs => {
 	       "p" => "a prime number", "n",
-	       Variable => Symbol => "the name to use for the generator of the field",
+	       Variable => Symbol => {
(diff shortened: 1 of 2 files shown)


mahrud:
I think this PR broke this example in a funny way: https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/SwitchingFields/html/_field__Extension.html

Do you want to fix the docs, or maybe even simplify the function using `map`?

d-torrance:
I don't think this PR did anything -- it just updated the documentation a bit.

It's weird that the html docs for `fieldExtension` don't display the output of the `try ... then ... else` clause.  Is that the "funny" thing you' mentioned?

When I run the code manually, `map(L2, K2)` works without an error.  I suppose that when the package was written, it raised an error?  Has `map` been updated in the meantime?

mahrud:
Oh you're right, I thought you actually changed the behavior. Who fixed this then?!

Issue #3239, opened 2024-05-16
"error: degree list should be of length 34" in check(6, "NumericalImplicitization")


d-torrance:
I've seen this a couple times in recent PPA builds:

```m2
NumericalImplicitization/tests.m2:125:1-139:1 error:
 -- 
 -- i10 : elapsedTime PW = pseudoWitnessSet(F,I,p, Repeats=>2, Threshold=>3, MaxThreads=>allowableThreads) -- ~ 40 seconds
 -- --error message bumped: degree list should be of length 34
 -- Tracking monodromy loops ...
 -- Points found: 2
 -- Points found: 4
 -- Points found: 8
 -- Points found: 14
 -- Finished tracking 14 paths in parallel, in 2.2066 seconds
 -- /<<PKGBUILDDIR>>/M2/Macaulay2/m2/monoids.m2:595:9:(1):[11]: error: degree list should be of length 34/<<PKGBUILDDIR>>/M2/Macaulay2/m2/monoids.m2:595:9:(1):[11]: error: degree list should be of length 34
 -- 
 -- /<<PKGBUILDDIR>>/M2/Macaulay2/m2/methods.m2:130:40:(1):[10]: --back trace--
 -- /<<PKGBUILDDIR>>/M2/Macaulay2/m2/methods.m2:130:40:(1):[10]: --back trace--
 -- /<<PKGBUILDDIR>>/M2/Macaulay2/m2/methods.m2:146:9:(1):[9]: --back trace--
 -- /<<PKGBUILDDIR>>/M2/Macaulay2/m2/methods.m2:146:9:(1):[9]: --back trace--
 -- /<<PKGBUILDDIR>>/M2/Macaulay2/m2/option.m2:17:8:(1):[7]: --back trace--
 -- /<<PKGBUILDDIR>>/M2/Macaulay2/m2/option.m2:17:8:(1):[7]: --back trace--
 -- /<<PKGBUILDDIR>>/M2/Macaulay2/m2/methods.m2:154:80:(1):[6]: --back trace--
 -- /<<PKGBUILDDIR>>/M2/Macaulay2/m2/methods.m2:154:80:(1):[6]: --back trace--
 -- /<<PKGBUILDDIR>>/M2/Macaulay2/m2/option.m2:17:8:(1):[5]: --back trace--
 -- /<<PKGBUILDDIR>>/M2/Macaulay2/m2/option.m2:17:8:(1):[5]: --back trace--
[...]

PR #3260, opened 2024-05-24
Various output routine updates

pzinn:
- moved some TeX-related code from `expressions.m2` to `latex.m2`
- fixed an annoying double superscript bug for things like `x'_1^2`
- harmonised tex output of zero vectors
- added `ReverseDictionary` to iterators
- added `texMath RRi`


M2/Macaulay2/m2/latex.m2
```diff
@@ -72,20 +79,84 @@ sectionType = sectionNumber -> (
     "\\subsubsubsubsubparagraph");
 
 -----------------------------------------------------------------------------
+keywordTexMath = applyKeys(hashTable { -- both unary and binary keywords
+    symbol |- => "\\vdash ",
```
mahrud, line 83: Two comments:
1. The indenting is incorrect with the `applyKeys(..., symbolBody)` addition. Could you do that on the next line perhaps? (e.g. rewrite it with `keywordTexMath = applyKeys(keywordTexMath, symbolBody)`)
2. On that note, could you also adjust the entries so the `=>` are all aligned?

M2/Macaulay2/m2/latex.m2
```diff
@@ -72,20 +79,84 @@ sectionType = sectionNumber -> (
...
+},last)
```
mahrud, line 140: Could you break these into multiple lines? (Also I think the indentation is off)

M2/Macaulay2/m2/latex.m2
pzinn: what do you mean, the indenting is incorrect? do we have rules for indentation laid out somewhere?

pzinn:
those inter-dependencies of various files in the loadsequence are a pain

pzinn:
so with this PR: `iterators.m2` needs `expressions.m2` which needs `max.m2` which needs `fold.m2` which needs `iterators.m2`...
how are we supposed to deal with those circular dependencies?

d-torrance:
I suppose the `accumulate` and `fold` methods involving `Thing` (i.e., the ones that use iterators) could get moved from `fold.m2` to `iterators.m2` to remove that dependency.

M2/Macaulay2/m2/latex.m2
mahrud: [Yes](https://github.com/Macaulay2/M2-emacs/blob/90ce16a15ea1f4fad08e8ba0b74b5045f4a72e8f/M2.el#L39-L42), four spaces for each unmatched parenthesis, etc. (and ideally tabs for each eight spaces, but this is more aesthetic).

DanGrayson:
Could you fix the broken test-builds, please?

pzinn:
a `NumericalImplicitization` test runs oom, don't think it's related to this PR?

d-torrance:
> a `NumericalImplicitization` test runs oom, don't think it's related to this PR?

Probably #3239

mahrud:
I reran that test and it passed; LGTM!

PR #3262, opened 2024-05-24
De-duplicate three or more duplicate variables

d-torrance:
We refactor the `dedupSymbols` helper function, which previously only de-duplicated up to two duplicate variables.

### Before
```m2
i1 : QQ[x,x]

o1 = QQ[x ..x ]
         0   1

o1 : PolynomialRing

i2 : QQ[x,x,x]
stdio:2:1:(3): error: baseName: no base name available
```

### After
```m2
i1 : QQ[x,x,x]

o1 = QQ[x ..x ]
         0   2

o1 : PolynomialRing
```

Closes: #3261

Issue #2895, opened 2023-06-29
Complexity of list operations


mahrud:
Core methods involving lists could use some auditing, because some are not optimal at all. Here is an example:
```m2
Macaulay2/Core/lists.m2:103:45-103:89: --source code:
all(BasicList,BasicList,Function) := (x,y,p) -> not any(apply(x,y,identity), ij -> not p ij)
```
Here the use of `apply` means we're iterating over all entries in x and y even if every single pair will return false!!
```m2
elapsedTime all(10^7:false, 10^7:false, (i,j) -> i and j) -- 1.71222 seconds elapsed
```

Related: #2561

mahrud:
These methods involving lists are very simple but having them in the top-level makes them very slow, especially considering that `isTable` is called by `matrix(List)`, which significantly slows down creating a very large list.

https://github.com/Macaulay2/M2/blob/5db7db6cecad7cff7a559ed322ae344171563322/M2/Macaulay2/m2/lists.m2#L300-L308

PR #3265, opened 2024-05-25
Added simple fix for deduplication of multiple variables

### LINKED ISSUE #3261: Inconsistent behavior when defining polynomial rings with duplicate variables
The following behavior seems reasonable, I suppose:

```m2
i1 : QQ[x,x]

o1 = QQ[x ..x ]
         0   1

o1 : PolynomialRing
```

But as soon as we try including three or more variables, we get the following:

```m2
i2 : QQ[x,x,x]
stdio:2:1:(3): error: baseName: no base name available
```

Shouldn't this give us `QQ[x_0,x_1,x_2]` for consistency?

mahrud:
Reverts #3262, closes #3261.

### DIFF
--- a/M2/Macaulay2/m2/monoids.m2
+++ b/M2/Macaulay2/m2/monoids.m2
@@ -478,7 +478,7 @@ dedupSymbols = varlist -> if 0 == repeats varlist then varlist else while 0 < re
     mapping := hashTable toList pairs varlist;
     -- TODO: applyPairs with collision handler, and accepting MutableHashTable
     counter := new MutableHashTable from applyKeys(mapping,
-	i -> varlist#i, dups -> makeVars(#dups, first dups));
+	i -> varlist#i, dups -> makeVars(#flatten dups, last dups));
     varlist = values applyValues(mapping, var -> if instance(counter#(name := var), List)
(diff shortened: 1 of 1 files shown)


d-torrance:
One of the reasons I chose to use a different data structure in #3262 was to avoid creating new lists with `drop`.  Instead, each value in the `data` hash table is a mutable list containing an index and the list of newly indexed variables for each variable, and we use the index to figure out where we are in the list when constructing the new deduplicated list.

The performance increase is noticeable when there are a lot of duplicated variables.  `dedupSymbols1` is the code from #3262 and `dedupSymbols2` is the code from this PR.

```m2
i43 : elapsedTime dedupSymbols1 apply(100, i -> x);
 -- 0.00101386 seconds elapsed

i44 : elapsedTime dedupSymbols2 apply(100, i -> x);
 -- 0.0151957 seconds elapsed
```

mahrud:
I'm skeptical that this difference is noticeable in real life, but point taken: using tally is definitely the right way to go.

By the way, `drop` being not efficient is an issue in itself, see #2895. (As well as it being named differently from `remove` and not returning the dropped value)

d-torrance (approved):
Looks good to me!

PR #3266, opened 2024-05-27
Update "remove" to return the removed value

d-torrance:
This change was suggested by @mahrud in a comment in the code added in #3265.

### Before
```m2
i1 : x = new MutableList from {2, 4, 6, 8, 10}

o1 = MutableList{...5...}

o1 : MutableList

i2 : remove(x, -1) -- no return value

i3 :
```

### After
```m2
i2 : remove(x, -1)

o2 = 10

i3 : 
```

### DIFF
--- a/M2/Macaulay2/d/actors5.d
+++ b/M2/Macaulay2/d/actors5.d
@@ -603,9 +603,10 @@ remove(x:List,i:int):Expr:= (
      else if i >= n || i < -n then ArrayIndexOutOfBounds(i, n - 1)
      else (
 	  if i < 0 then i = n + i;
+	  ret := x.v.i;
 	  for j from i to n - 2 do x.v.j = x.v.(j + 1);
 	  Ccode(void, x.v, "->len = ", n - 1);
-	  nullE));
+	  ret));
 
 removefun(e:Expr):Expr := (
      when e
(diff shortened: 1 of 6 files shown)


mahrud:
Thank you, this is great!

Other features that I think would be nice to add at some point, though not necessarily now if it's too much work:
- make this consistent with `take`, accepting a list of integers, removing those and returning a list of removed values 
- make this consistent with `delete`
- make this also accept any VisibleList or HashTable as the first input.


(ps: while you're editing monoids.m2, I realized dedupSymbols can be further simplified by using a simple `while true` and removing the whole `mapping` line and also `else varlist` at the end. I was really distracted while making that PR!)

d-torrance:
`take` and `delete` already have return values (new lists with the corresponding entries removed), so I'm not sure if changing these would be a good idea.

How would something like `remove(x, i)` work if `x` is immutable?  We could return the `i`th element of `x`, I suppose, but we shouldn't actually remove it from `x` since it's immutable.

mahrud:
I should have elaborated:
1. `take`:
    - `remove(L, {i,j,..})` should change `L` by removing the i-th entry, etc. and return a _list_ `{L#i,L#j,...}` corresponding to the removed values.
    - (less optimistic about this, but) `take(L, {1,3})` should just give you those two items, but `take(L, (1,3))`, `take(L, 1..3)`, and `take(L, {1..3})` should give you the range of three objects.
    - both `take(L, {i..j})` and `remove(L, {i..j})` should do what you expect
    - the second parameter should allow repetitions (similar to `List_List`), in particular `take(L, {5:2})` and `remove(L, {5:2})` should do what you expect
    - if the first input is a mutable hash table, `take(H, L)` should return a list with the values corresponding to the items in `L`, and `remove(H, L)` should do the same but also change `H` by removing those key-value pairs.
2. `delete`:
    - this function is confusing because the list or hash table is the second input
    - `delete(L, a)` should return a **new** list with `a` removed, similarly
    - `delete(L, {a,b,...})` should do what you expect,
    - `delete(H, {key1,key2,...})` again what you expect but with `H` a hash table
    - this should also work for a mutable hash table, but making a copy
3. Like you said, for immutable lists or hash tables, `remove` should basically be a synonym for `take` (which, as mentioned above, also should accept hash tables)
4. Eventually, something like `append(L, remove(L, i))` should remove an entry and add it to the end without a copy, similarly `join(L, remove(L, {i,j,...}))`, `insert(L, j, remove(L, {i,k,...}))`, etc.

I'm open to different interpretations for most of these, but I think we agree that consistency is good, and ability to pipe different functions without unnecessary third variables or copying in an intuitive way is good.

PR #3269, opened 2024-05-30
Export changeDirectory

d-torrance:
I've often wanted the option to change the current working directory inside M2 (e.g, if without realizing I run `M-x M2` inside Emacs when the active buffer is visiting some file in some strange directory).

It turns out that a `changeDirectory` function already exists, but it just wasn't exported!

We update it slightly (allowing the expansion of `~` to the user's home directory and changing to the home directory when there's no input), document it, and add unit tests.

```m2
i1 : currentDirectory()

o1 = /home/profzoom/tmp/

i2 : makeDirectory "foo"

i3 : changeDirectory "foo"

i4 : currentDirectory()

o4 = /home/profzoom/tmp/foo/
```

### DIFF
--- a/M2/Macaulay2/d/actors5.d
+++ b/M2/Macaulay2/d/actors5.d
@@ -1316,11 +1316,6 @@ readlinkfun(e:Expr):Expr := (
      else WrongArgString());
 setupfun("readlink",readlinkfun);
 
-changeDirectory(e:Expr):Expr := (
-     when e is filename:stringCell do if chdir(filename.v) == -1 then buildErrorPacket(syscallErrorMessage("changing directory")) else nullE
-     else WrongArgString());
-setupfun("changeDirectory",changeDirectory);
-
 realpathfun(e:Expr):Expr := (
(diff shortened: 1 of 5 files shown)


M2/Macaulay2/d/actors5.d
```diff
@@ -1316,9 +1316,19 @@ readlinkfun(e:Expr):Expr := (
      else WrongArgString());
 setupfun("readlink",readlinkfun);
 
+changeDirectory(dir:string):Expr := (
+    if chdir(expandFileName(dir)) == -1
+    then buildErrorPacket(syscallErrorMessage("changing directory"))
+    else nullE);
```
mahrud, line 1322: You might want to have it return whatever `currentDirectory()` returns. This is useful to know the exact new address.

On that note, it would be good if `makeDirectory()` also returned the address so you could do `changeDirectory makeDirectory "~/foo"`.

d-torrance: Great ideas!

PR #3270, opened 2024-05-30
Bump line number after syntax errors

d-torrance:
This adds consistency with other errors.  For example, this is what happens when we have higher-level errors:

```m2
i1 : 1/0
stdio:1:2:(3): error: division by zero

i2 : 1/0
stdio:2:2:(3): error: division by zero

i3 : 1/0
stdio:3:2:(3): error: division by zero

i4 : currentPosition()

o4 = stdio:5:0

o4 : FilePosition
```

The current line number increases by 1 each time, as does the file position.

However, if it's a syntax error, this doesn't happen:

```m2
i1 : )
stdio:1:1:(3): error: syntax error at ')'

i1 : )
stdio:1:2:(3): error: syntax error at ')'

i1 : )
stdio:1:3:(3): error: syntax error at ')'

i1 : currentPosition()

o1 = stdio:2:0

o1 : FilePosition
```

### After

```m2
i1 : )
stdio:1:1:(3): error: syntax error at ')'

i2 : )
stdio:2:1:(3): error: syntax error at ')'

i3 : )
stdio:3:1:(3): error: syntax error at ')'

i4 : currentPosition()

o4 = stdio:5:0

o4 : FilePosition
```

### DIFF
--- a/M2/Macaulay2/d/interp.dd
+++ b/M2/Macaulay2/d/interp.dd
@@ -142,6 +142,8 @@ readeval4(file:TokenFile,printout:bool,dictionary:Dictionary,returnLastvalue:boo
 		    	      if debugLevel == 123 then stderr <<  "-- error during parsing" << endl;
 			      if fileError(file) then return buildErrorPacket(fileErrorMessage(file));
 			      if stopIfError || returnIfError then return buildErrorPacket("--backtrace: parse error--");
+			      file.posFile.line = file.posFile.line + 1;
+			      file.posFile.column = ushort(0);
 			      );
 			 )
 		    else (
--- a/M2/Macaulay2/d/interp.dd
+++ b/M2/Macaulay2/d/interp.dd
@@ -108,6 +108,8 @@ readeval4(file:TokenFile,printout:bool,dictionary:Dictionary,returnLastvalue:boo
 	       	    if fileError(file) then return buildErrorPacket(fileErrorMessage(file));
 	       	    if stopIfError || returnIfError then return buildErrorPacket("--backtrace: token read error--");
 		    if isatty(file) || file.posFile.file.fulllines then flushToken(file);
+		    file.posFile.line = file.posFile.line + 1;
+		    file.posFile.column = ushort(0);
 		    )
 	       )
 	  else (


pzinn:
I would be very careful changing this behaviour. I spent a lot of time fixing line numbers in error messages and I think I got it under control now (still haven't made a PR sadly) . Please give me some time to take a look at this. (I'm on the plane for the next 24h).

pzinn:
Looks good to me. I'm wondering if you also want to do something about this:
```
i1 : `
stdio:1:1:(3): error: invalid character

i1 : 
```

d-torrance:
> Looks good to me. I'm wondering if you also want to do something about this:
> 
> ```
> i1 : `
> stdio:1:1:(3): error: invalid character
> 
> i1 : 
> ```

Yes, definitely -- thanks for pointing this out!

d-torrance:
The "invalid character" error is one of a larger class of "token errors".  I noticed the code earlier but wasn't sure how to reproduce one.  Another example is trying to parse a real number that has `e` but no exponent.

### Before
```m2
i1 : `
stdio:1:1:(3): error: invalid character

i1 : 1e
stdio:1:4:(3): error: exponent missing in floating point constant

i1 : 
```

### After
```m2
i1 : `
stdio:1:1:(3): error: invalid character

i2 : 1e
stdio:2:3:(3): error: exponent missing in floating point constant

i3 : 
```

mahrud:
Are you sure the behavior you're changing wasn't intentional? The `stdio` line number is changing to signal where in the source an error occurred in order to fix it, but it's not necessarily supposed to match the `i` number of the input. Here's an example:
```m2
i1 : 1; 2; 3; 4; 5; 6; 7; 8; 9; )
stdio:1:28:(3): error: syntax error at ')'

i10 : 
```

d-torrance:
I agree that the `stdio` line number shouldn't necessarily match the `i` number.  And it won't necessarily agree after this PR either:

```m2
i1 : 1; 2; 3; 4; 5; 6; 7; 8; 9; )
stdio:1:28:(3): error: syntax error at ')'

i11 : )
stdio:2:1:(3): error: syntax error at ')'
```

Note that the `i` number increased by 10 (the 9 evaluated expressions and the 1 syntax error) and the `stdio` line number increased by 1.  Currently, the `i` number only increases 9 times for the evaluated expressions and the `stdio` line number doesn't change.

I noticed this issue after `code` didn't work properly for functions defined on the standard input after some errors.  This is because we were adding a new entry to readline's history but not increasing the `stdio` line number, so the numbers got out of sync.

Currently:

```m2
i1 : )
stdio:1:1:(3): error: syntax error at ')'

i1 : f = x -> x^2

o1 = f

o1 : FunctionClosure

i2 : code f

o2 = stdio:1:7-1:12: --source code:
     )
```

But with this PR:

```m2
i1 : )
stdio:1:1:(3): error: syntax error at ')'

i2 : f = x -> x^2

o2 = f

o2 : FunctionClosure

i3 : code f

o3 = stdio:2:6-2:11: --source code:
     f = x -> x^2
```

d-torrance:
Ok, that makes sense.  I'll revert the `i` line number stuff.

PR #3271, opened 2024-05-31
Add MultigradedImplicitization package

bkholler:
This package implements the main algorithm from our paper "Computing Implicitizations of Multi-Graded Polynomial Maps"
(https://arxiv.org/abs/2311.07678) which was recently accepted for presentation at MEGA. 

This is essentially an algorithm to compute a (partial) generating set for the kernel of a polynomial map by turning it into many small linear algebra problem, with one for each homogeneous component of the kernel. This strategy is partriculalry useful for implicitization problems in algebraic statistics.

### DIFF
--- a/M2/Macaulay2/packages/=distributed-packages
+++ b/M2/Macaulay2/packages/=distributed-packages
@@ -271,3 +271,4 @@ SchurVeronese
 VNumber
 TropicalToric
 MultigradedBGG
+MultigradedImplicitization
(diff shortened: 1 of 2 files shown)


d-torrance:
The call to `newPackage` should have a `Keywords` option, e.g.:

```m2
newPackage("MultigradedImplicitation",
   ...
   Keywords => {"Foo"},
   ...
```
For a list of keywords currently in use, see the headings at https://www.macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_packages_spprovided_spwith_sp__Macaulay2.html.

d-torrance:
The checks are still failing, again because of an option to `newPackage`.  The `DebuggingMode` option should be set to `false` (or just removed, since that's the default value).

bkholler:
Sorry about that! I should've read the checks more carefully. Thanks for the help!

PR #3279, opened 2024-06-04
Add PR template with some instructions

d-torrance:
When new PR's are opened, this will give some instructions inside the "Write" field (target the `development` branch, don't forget `=distributed-packages`, `Keywords`, and `DebuggingMode`).  They're inside a comment, so they won't actually appear on the PR page after it's been created.

See [Creating a pull request template for your repository](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository).

Skipping the builds and targeting `master` since this needs to be on the default branch to work.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,5 @@
+<!--
+Thank you for contributing to Macaulay2!
+
+Please read https://github.com/Macaulay2/M2/wiki/Pull-requests for instructions.
+-->


mahrud:
I like this, but instead of adding the instructions directly, it might be better to have a link to a page in the wiki with instructions, which would be easier to update as we go.

d-torrance:
Sounds good!  The template now links to the [Pull requests](https://github.com/Macaulay2/M2/wiki/Pull-requests) wiki page.

d-torrance:
This was intended to be merged into `master` so that the template would actually work. It needs to be on the default branch.

mahrud:
@d-torrance do you want to cherry-pick this to the main branch?

d-torrance:
Good idea!  Done.

Issue #1292, opened 2020-06-17
switch autotool build from libraries/flint to submodules/flint2


DanGrayson:
... to get the latest version.

And then delete libraries/flint from the source tree.

d-torrance:
Closed by #3281

PR #3281, opened 2024-06-05
Use submodule for flint build (autotools)

d-torrance:
We update the autotools build to use the submodule, matching the cmake behavior.

I tried this a few months ago without any luck because the current practice in the autotools build for submodules is to do out-of-tree builds, which flint doesn't support.  So instead, we use the existing code in the `libraries` directory, but instead of downloading and extracting a tarball, we update the submodule and copy its contents to the build directory.

This should also make moving any other libraries from tarballs to submodule much easier.  Adding `SUBMODULE = true` to the appropriate Makefile might be sufficient.  (I'm also tempted to move some of the existing submodules to this interface -- I think it's a lot cleaner than the current hodgepodge of targets in GNUMakefile.)

Closes: #1292

### DIFF
--- a/M2/configure.ac
+++ b/M2/configure.ac
@@ -996,6 +996,7 @@ then AC_LANG(C)
 fi
 if test $BUILD_flint = yes
 then BUILTLIBS="-lflint $BUILTLIBS -lm"
+     AC_DEFINE([HAVE_FLINT_NMOD_H])
 fi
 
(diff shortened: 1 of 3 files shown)


mahrud:
I don't really like using submodules, to be honest, and would prefer to see them removed, especially now that flint has more regular releases.

PR #3282, opened 2024-06-05
Remove libatomic_ops submodule; no longer necessary

d-torrance:
I neglected to do this in #3082.

Skipping the test builds.


mahrud (approved):

DanGrayson:
Why are you skipping the test builds???

d-torrance:
Because it doesn't affect them at all. We don't link against this library any more, and even when we did, we used system copies instead of building the submodule.

PR #3289, opened 2024-06-06
Submodule updates for autotools build

d-torrance:
Rather than building tarballs and submodules completely differently using the autotools build, we do everything in the `libraries` directory, as was done for flint in #3281.  This simplifies maintenance, as all we need to do to switch back and forth between tarball and submodule builds is set the `SUBMODULE` variable in the approriate Makefile.  It also allows us to remove a bunch of code from GNUmakefile.

We also update the autotools build to use the same submodules that the cmake build does, adding bdwgc, frobby, and googletest.

### DIFF
--- a/M2/GNUmakefile.in
+++ b/M2/GNUmakefile.in
@@ -83,10 +83,6 @@ ifneq (@host_alias@,)
 # it was ../configure --build=@host_alias@ (for cross compiling)
 SUBMODULE_CONFIGOPTIONS += --host=@host_alias@
 endif
-copy-license-files-from-memtailor: LICENSE_FILES = README.md license.txt
-
-# the Makefile is the target used to run "configure" in the submodule:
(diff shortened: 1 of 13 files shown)


mahrud:
I'll reiterate that I'd prefer not using submodules, and if I had more time I would move cmake to not use submodules for flint, bdwgc, and googletest. The main reason we had flint as a submodule was that we needed a lot of patches to get it to work with mpir, which is out of the picture now.

Also, to be clear, cmake doesn't actually build mathic, etc. anymore. They get built as a part of the engine.

d-torrance:
I have no strong opinion either way -- I always just use packaged libraries and don't worry about building anything other than M2 itself.  With the commits switching bdwgc, frobby, and googletest, I was mostly just trying to make the autotools build more consistent with the cmake one.

If we switch back to tarballs for any of these things with the cmake build, with this PR it will be really easy to do the same for the autotools build -- just remove the `SUBMODULE = true` line from the corresponding Makefile.

PR #3294, opened 2024-06-08
Add an mpfi patch

d-torrance:
I just built a Fedora 40 package for 1.24.05 and posted it to the website.   I took some notes on the process and updated INSTALL accordingly.  One of the few dependencies without an existing Fedora package is mpfi, and it needed a patch ([cherry-picked from upstream's git repo](https://gitlab.inria.fr/mpfi/mpfi/-/issues/21774)) to compile with the latest gcc.

I also built a Rocky Linux 9.4 (compatible with RHEL 9.4) package, but didn't think to take notes about what I did until afterwards...  There was lots of fiddling with FFLAS-FFPACK flags!

### DIFF
--- a/M2/libraries/mpfi/Makefile.in
+++ b/M2/libraries/mpfi/Makefile.in
@@ -3,6 +3,7 @@
 VERSION = 1.5.4
 LICENSEFILES = README COPYING.LESSER
 TARFILE = $(LIBNAME)-$(VERSION).tar.gz
+PATCHFILE = @abs_srcdir@/patch-$(VERSION)
 
 #############################################################################
 CONFIGOPTIONS += --disable-thread-safe
--- a/M2/libraries/mpfi/patch-1.5.4
+++ b/M2/libraries/mpfi/patch-1.5.4
@@ -1,0 +1,60 @@
+From a02e3f9cc10767cc4284a2ef6554f6df85e41982 Mon Sep 17 00:00:00 2001
+From: REVOL Nathalie <nathalie.revol@inria.fr>
+Date: Sat, 19 Mar 2022 18:09:45 +0100
+Subject: [PATCH] incorret types: corrected
+
+---
+ mpfi/src/div_ext.c | 16 ++++++++--------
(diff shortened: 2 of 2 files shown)


d-torrance:
I canceled the builds (meant to skip them but forgot) since this PR doesn't affect them.  We're just editing a text file and patching a library that GitHub doesn't build.

mahrud:
Would it be better if the building instructions were on the wiki and we just had links to those pages in a single INSTALL.md file in git? (Thinking of also moving the CMake instructions to wiki)

d-torrance:
Sounds good to me!

d-torrance:
I went ahead and added some links on the wiki page for installing M2 in various ways.  I started the Linux instructions and plan on working on the autotools ones soon.

d-torrance:
I removed the INSTALL changes from this PR -- it's now on the wiki (which still needs some formatting, but it's in progress.)  It's now just the mpir patch.

Still skipping the GitHub builds since they don't build mpir.

jamesjer:
I don't understand.  Fedora has had an [mpfi package](https://src.fedoraproject.org/rpms/mpfi) since 2008.  I maintain it.  The linked patch has been applied to it, which makes sense since I am both the author of the patch and the maintainer of the Fedora mpfi package. :-)

d-torrance:
Weird -- maybe I fat-fingered something when I tried to install it.  I'll try playing around with it again.

Regardless, we should include this patch for anyone building Macaulay2 on a system that doesn't have an mpfi package.

d-torrance:
Found the `mpfi-devel` package!  Not sure how I missed it...

d-torrance:
@DanGrayson -- is this okay to merge?  As I mentioned above, I skipped the builds because the GitHub builds use packaged versions of mpfi and this patch wouldn't be tested.

DanGrayson:
I don't see any test runs.

d-torrance:
Because I skipped them. As I explained above, this patch wouldn't be tested since we don't build mpfi as part of the github builds.

PR #3297, opened 2024-06-10
Remove nullary changeDirectory example

d-torrance:
It changes the directory to the home directory, but this directory might not actually exist on the build system.

In #3287, I did the same thing for the `changeDirectory` tests, but I forgot about the docs!


mikestillman:
@d-torrance I am going to accept your series of simple pull requests.  Is using "rebase and merge" ok under this scenario?

d-torrance:
Yes, that should be ok!  From what I understand, GitHub's "rebase and merge" isn't exactly the same thing as running `git rebase`, but it does avoid lots of extra merge commits in the git history.

PR #3298, opened 2024-06-11
Add Dependabot support and update KaTeX

d-torrance:
We add support for Dependabot to keep our GitHub actions and the npm modules used for the two Javascript syntax highlighting lexers up-to-date.

Once the file `.github/dependabot.yml` has been merged into the `master` branch, we will get automated PR's on occasion when any of these modules are updated.

For example, here are the PR's that I got on my branch when I tested it out:

* https://github.com/d-torrance/M2/pull/3
* https://github.com/d-torrance/M2/pull/4
* https://github.com/d-torrance/M2/pull/5
* https://github.com/d-torrance/M2/pull/6
* https://github.com/d-torrance/M2/pull/7

I went ahead merged all these and included the commits in this PR.


mikestillman:
@mahrud Could you review this PR?

mahrud:
I'm not very familiar with dependabot, but few suggestions that come to mind:
- Would changing the schedule to every month or even 6 months be better?
- Could we combine all the PRs into one? Seems like it's [possible](https://github.com/dependabot/dependabot-core/issues/2265#issuecomment-1742364365).
- There are a few other javascript packages embedded in the packages, namely prism and katex in Style and a whole bunch in Visualize. Is it possible to keep some of those updated as well?

d-torrance:
Yeah, less frequent updates is probably better. The examples all use weekly, so I just ran with it.

I think Visualize and Style support would be good, too. I had draft branches for both.

d-torrance:
I've changed the interval to monthly and also set the PR's to be grouped.

d-torrance:
I went ahead and added the KaTeX update to this PR.  Visualize is going to need quite a bit of work to update.

.github/dependabot.yml
```diff
@@ -0,0 +1,37 @@
...
+    groups:
+      highlightjs-dependencies:
+        patterns:
+          - "*"
+
+  - package-ecosystem: "npm"
+    directory: "/M2/Macaulay2/packages/Style"
+    schedule:
+      interval: "monthly"
+    groups:
```
mahrud, line 34: Will this group dependency updates for prism and highlightjs into one PR? That's what I had in mind, based on the first example [here](https://github.blog/2023-08-24-a-faster-way-to-manage-version-updates-with-dependabot/#example-1-group-by-dependency-type), but I'm not sure.

d-torrance: Currently, prism and highlightjs get separate PR's.  I'll play around with it some more and see if we can combine them.

d-torrance: It looks like it's one group per element of the "updates" list in yaml file (so prism and highlightjs are separate).  I tried moving the "groups" key up a level and I got an error: "The property '#/' contains additional properties ["groups"] outside of the schema when none are allowed"

mahrud (approved):
Let's try this for a while, worse case we can disable it if there are too many PRs.

d-torrance:
I realized that we don't need all the dependencies for the highlight.js any more -- that was just cruft left over from before we switched to prism for syntax highlighting the documentation.  I went ahead and removed all that.

M2/Macaulay2/editors/highlightjs/README.md
```diff
@@ -2,114 +2,10 @@ Macaulay2 in highlight.js
...
-```
-
-Using the CMake build, run the following:
-
-```
-ninja M2-highlightjs
-```
-
-Both build systems will save the file in `Macaulay2/editors/highlightjs`.
+See https://github.com/d-torrance/highlightjs-macaulay2 for more information.
```
mahrud, line 11: If it's just a README file and no actual generation code then maybe just delete this directory and put a link to your repo on wiki.

d-torrance: We still need macaulay2.js.in to re-generate the source with new symbols every release.

.github/dependabot.yml
```diff
@@ -0,0 +1,28 @@
...
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    groups:
+      github-action-dependencies:
+        patterns:
+          - "*"
+
+  - package-ecosystem: "npm"
+    directory: "/M2/Macaulay2/editors/prism"
```
mahrud, line 13: Would it make more sense to also move every file other than `macaulay2.js.in` and perhaps `README.md` to Style? Then prism and katex can update together and the build target in this directory will generate the grammar file and copy it into the Style package.

d-torrance: Yeah, I though about that too.  I'll play around with it.

M2/Macaulay2/editors/highlightjs/README.md
mahrud: Since we don't actually use highlight.js anymore, does that need to be in this repository? We could even move the `generateGrammar` function from `make-M2-symbols.m2` to Style so that any M2 binary can generate macaulay2.js, without needing the whole repository, and it doesn't have to be at release, just whenever someone wants it.

mahrud:
Here's a suggestion:
- publish `make-M2-symbols.m2` in Style (which currently doesn't have any M2 code, but this makes sense as an addition).
- have `makePackageIndex` also regenerate `prism.js` based on preloaded packages along with `.Macaulay2/index.html`.
- could even regenerate the emacs wordlist, either in-place or have an add-on list in `.Macaulay2`.

Issue #2398, opened 2022-02-27
Broken html-check-links


d-torrance:
There are currently two issues with the `html-check-links` utility:

* Currently, `make -C Macaulay2/html-check-links` only checks xhtml files, of which there are none.  This was because of a mistake I made in [#1875](https://github.com/Macaulay2/M2/pull/1875).
* After restoring the original behavior of the check, we get the following on the autotools GitHub builds, e.g., [this one](https://github.com/Macaulay2/M2/runs/5347144612?check_suite_focus=true).  I can't reproduce this locally, and I don't understand why it thinks that `YYSTYPE` is `int`.  That's what it's initially defined as in `grammar-tab.h` by bison, but it's supposed to be redefined as `char *` in `grammar.h`.  Why isn't it?
```
gcc -std=gnu11 -g3 -O2 -Wall -g -I/home/runner/work/M2/M2/M2/BUILD/build/../../include -I/home/runner/work/M2/M2/M2/BUILD/build/include -I/home/runner/work/M2/M2/M2/BUILD/build/usr-host/include -isystem /usr/include/libxml2     -I/usr/src/gtest  -I/home/linuxbrew/.linuxbrew/include -I/home/linuxbrew/.linuxbrew/opt/tbb@2020/include -DNDEBUG  -I/usr/include/cdd -I/usr/include/eigen3 -I/usr/include/python3.8 -I/usr/include -DBOOST_STACKTRACE_LINK -I../../../../Macaulay2/html-check-links -I.  -c grammar.fixed.c -o grammar.fixed.o
      |               ^
../../../../Macaulay2/html-check-links/lex.l:38:15: warning: assignment to ‘YYSTYPE’ {aka ‘int’} from ‘char *’ makes integer from pointer without a cast [-Wint-conversion]
   38 | <tag>src   COUNT; yylval = strperm(yytext); return SRC;
      |               ^
../../../../Macaulay2/html-check-links/lex.l:39:16: warning: assignment to ‘YYSTYPE’ {aka ‘int’} from ‘char *’ makes integer from pointer without a cast [-Wint-conversion]
   39 | <tag>{STRING}   COUNTN; yylval = strperm(yytext); return WORD;
      |                ^
../../../../Macaulay2/html-check-links/lex.l:40:16: warning: assignment to ‘YYSTYPE’ {aka ‘int’} from ‘char *’ makes integer from pointer without a cast [-Wint-conversion]
[...]

PR #3303, opened 2024-06-12
Use xargs (with -r option) again when running html-check-links

d-torrance:
Back in #1875, I broke `make -C Macaulay2/html-check-links`, which finds all the html files we just built and checks them for broken links.  It previously used `xargs` to call  `ls` and sort them from oldest to newest.

This was a problem when we didn't build any html documentation -- we called `ls` on the empty list and ended up checking all the non-html files for broken links.  So I switched away from `xargs`, inadvertently completely breaking everything.  We haven't actually checked for any broken links for over three years!

I tried fixing it in #2392, but there were some strange bison-related errors I couldn't figure out.  They appear not to be an issue anymore, so we're trying again!

Basically, we return to using `xargs`, but now with the `-r` (short for `--no-run-if-empty`) option so we don't run `ls` on the empty list again.  We also stop searching for xhtml files since there won't be any.

For now, the autotools builds should fail because there are lots and lots of broken links thanks to a mistake I made in #3277.  But that's a good thing because it will mean that the check is working again!  I'll rebase this on top of #3299 once it's merged.

Closes: #2398


mahrud:
(I think you already removed this change from this PR, but for the record, I hate how the filenames in the documentation are nearly garbled ... I'd love a different system that doesn't have a _dozen __under____scores_in_it____!____!)

Issue #579, opened 2017-02-17
select on hashtables is confusing


DanGrayson:
See the discussion at #578 . 

Decree that `select` no longer works for hash tables and introduce new functions `selectKeys` and `selectValues`.   Maybe also `selectPairs`.

d-torrance:
Fixed in #3305

Issue #578, opened 2017-02-17
select(Set, Function) isn't what you might expect


denham0:
Check this out:

i837 : select(set{0,1,2},odd)
o837 = set {0, 1, 2}
o837 : Set

i840 : select(set{0,1,2},even)
o840 = set {}
o840 : Set

Now that's not a bug, because a Set is a HashTable, so select applies the function to the values in the hash table.  Which, for a set, are all 1.  But I think most people would expect select(Set, Function) to select a subset (i.e., apply the function to the keys of the hash table, rather than the values.)  It's not a big deal, but I think it would be better if the user didn't have to know how Set was implemented in order to avoid this little trap.

DanGrayson:
Good point!  I think I was thinking of a list as being like a hash table whose keys are 0,1,...,n, as in this example:

```
i1 : f = method();

i3 : f List := v -> new HashTable from apply ( #v, i -> i => v#i );

i5 : f {4,6,11,13}

o5 = HashTable{0 => 4 }
               1 => 6
               2 => 11
               3 => 13

o5 : HashTable

i6 : select (  {4,6,11,13}, even )

o6 = {4, 6}

o6 : List

i7 : select ( f {4,6,11,13}, even )

o7 = HashTable{0 => 4}
               1 => 6

o7 : HashTable
```
We could decree that `select` no longer works for hash tables and introduce new functions `selectKeys` and `selectValues`.  We already do the analogous thing for `apply` and `scan`.

denham0:
That sounds like a good solution.  Then one could use `selectKeys` in order to select from sets as one would from lists.  But `select(Set,Function)` would fail, which should help keep the user out of trouble.  

And it would be nice to select from hash tables both by keys or by values.

d-torrance:
Fixed in #3305

PR #3305, opened 2024-06-13
Add select functions for hash tables

d-torrance:
We add three new methods (`selectKeys`, `selectValues`, and `selectPairs`) for hash tables, partially addressing #579.  That issue also suggested dropping hash table support from `select`.  But that would be a breaking change (there's some code in `Macaulay2Doc` that uses it, for example), so I'm proposing that we instead keep it around as an alias for `selectValues`.

We also add `select` support for sets, using keys instead of values, addressing #578.

For example:

```m2
i1 : x = hashTable apply(5, i -> (i, i + 1))

o1 = HashTable{0 => 1}
               1 => 2
               2 => 3
               3 => 4
               4 => 5

o1 : HashTable

i2 : selectKeys(x, odd)

o2 = HashTable{1 => 2}
               3 => 4

o2 : HashTable

i3 : selectValues(x, odd)

o3 = HashTable{0 => 1}
               2 => 3
               4 => 5

o3 : HashTable

i4 : selectPairs(x, (k,v) -> odd(k + 2*v))

o4 = HashTable{1 => 2}
               3 => 4

o4 : HashTable

i5 : select(set(1..10), odd)

o5 = set {9, 1, 3, 5, 7}

o5 : Set
```


mahrud (approved):
I like the special routine for `Set`!

Issue #1644, opened 2020-12-02
Selecting preloaded packages in init.m2


mahrud:
> A general comment, I think it would be good if the user could control what packages were loaded by default.

_Originally posted by @kschwede in https://github.com/Macaulay2/M2/issues/641#issuecomment-309957744_

This would likely involve some changes here:

https://github.com/Macaulay2/M2/blob/1e47260755100936bafd60d64013672dedd3941f/M2/Macaulay2/m2/last.m2#L30-L54

mahrud:
@d-torrance can you think of a nice way to do this? It would be nice to customize the list. For instance, I don't know why `ConwayPolynomials, InverseSystems, LLLBases, OnlineLookup, ReesAlgebra, TangentCone` are loaded by default.

mahrud:
That seems worth a try!

d-torrance:
Fixed in #3312

PR #3312, opened 2024-06-17
Allow modification of preloaded packages in init.m2

d-torrance:
We run the start functions (which, among other things, loads the `User` package that loads all of the preloaded packages) *after* loading `init.m2` rather than before it.  This allows us to modify the list of preloaded packages.

For example, if we put the following in `init.m2`:

```m2
Core#"pre-installed packages" = {"FirstPackage"}
```

Then we get:

```m2
Macaulay2, version 1.24.05-160-gc1436f9cbf-dirty (pre-installed-packages)
with packages: FirstPackage

i1 : 
```

Closes: #1644


mahrud:
Isn't "pre-installed packages" a misnomer? They are all pre-installed. Maybe we should also rename this option to "preloaded packages"?

d-torrance:
Yeah, that was bugging me too.  Renamed!

mahrud:
Thanks! Maybe along with that, we should rename the "m2/installedpackages" file, or better yet define the list in Core.m2?

mahrud (approved):

mahrud:
I think we broke the ability to do `importFrom_Core "blah"` in `init.m2`!

d-torrance:
Oops, we did!  I think the problem is that now we load `init.m2` before we load the `User` package, and so `currentPackage` is null.  And in particular, there's no dictionary to add the imported symbols to.

mahrud:
Yes, I figured this much, but wasn't sure what's the best solution. Thoughts on setting currentPackage to be Core before init.m2 is loaded? Then importFrom essentially acts like export!

mahrud:
No that doesn't work ... surprisingly using `export` in `init.m2` also fails ...

PR #3313, opened 2024-06-18
Added Normaliz as a linked library

mahrud:
- **added support for torsion class group for truncate**
- **renamed truncate option Nef to Cone**
- **fixed caching in truncationMonomials**
- **simplified hilbertBasis Cone in Polyhedra**
- **modified isVeryAmple ToricDivisor to use hilbertBasis Cone**
- **modified latticePoints ToricDivisor to use latticePoints Cone**
- **added Nauty as a library in CMake**
- **added Normaliz as a library in CMake**
- **updated installation documents and docker scripts**
- **added rawHilbertBasis using libnormaliz**
- **modified truncate, hilbertBasis Cone to use rawHilbertBasis**
- **moved hilbertBasis to Polyhedra**
- **fixed broken tests due to new order of hilbertBasis**
- **fixed a dependency in ReactionNetworks**
- **fixed a dependency in a GroebnerStrata example**
- **added rawFourierMotzkin, calling libnormaliz**
- **changed findHeft to use rawFourierMotzkin**
- **brushed up Normaliz, in preparation for overhaul**
- **moved isSmooth to Core**
- **moved isVeryAmple to Core**
- **fixed a conflict over isReflexive**
- **removed pesky printout**
- **added libnauty2-dev and libnormaliz-dev to github builds**
- **fixed broken examples in FourierMotzkin**
- **fixed symbol conflicts in Normaliz examples**
- **fixed bug in heft of degree zero rings**
- **fixed symbol conflicts in MonodromySolver examples**
- **wip: fix configure.ac for libnormaliz**

@d-torrance the last commit needs work for getting autotools to find and link with Normaliz. Could you help with this?

### DIFF
--- a/M2/Macaulay2/e/interface/cone.cpp
+++ b/M2/Macaulay2/e/interface/cone.cpp
@@ -24,6 +24,7 @@ const Matrix /* or null */ *rawFourierMotzkin(const Matrix *C)
 {
   try
     {
+      // TODO: generalize the input type, in particular to allow lineality space
       const Ring *R = C->get_ring();
       const size_t c = C->n_cols();  // rank of ambient lattice
       const size_t r = C->n_rows();  // number of cone inequalities
--- a/M2/Macaulay2/e/interface/cone.cpp
+++ b/M2/Macaulay2/e/interface/cone.cpp
@@ -1,0 +1,62 @@
+// Written in 2021 by Mahrud Sayrafi
+
+#include "interface/cone.h"
+
+#include <M2/math-include.h>
+
+#include "debug.hpp"
+#include "interface/gmp-util.h"
+#include "interface/matrix.h"
+#include "matrix-con.hpp"
+#include "matrix.hpp"
+#include "relem.hpp"
+
+#include <libnormaliz/cone.h>
+#include <vector>
+
+typedef mpz_class Integer;
+
+/**
+ * \ingroup cones
+ */
+
+const Matrix /* or null */ *rawHilbertBasis(const Matrix *C)
+{
+  try
+    {
+      // TODO: Check that C is over ZZ
+      // TODO: for cones over QQ, lift to ZZ first
+      // TODO: Normaliz also supports algebraic cones defined over
+      // algebraic number fields embedded in RR
+      const Ring *R = C->get_ring();
+      const size_t c = C->n_cols();  // rank of ambient lattice
+      const size_t r = C->n_rows();  // number of cone rays
+
+      auto rays = libnormaliz::Matrix<Integer>(r, c);
+      for (size_t i = 0; i < r; i++)
+        for (size_t j = 0; j < c; j++)
+          rays[i][j] = static_cast<Integer>(C->elem(i, j).get_mpz());
+
+      auto cone = libnormaliz::Cone<Integer>(libnormaliz::Type::cone, rays);
+      // cone.compute(libnormaliz::ConeProperty::HilbertBasis,
+      //              libnormaliz::ConeProperty::DefaultMode);
+      auto HB = cone.getHilbertBasis();
+      size_t n = HB.size();  // number of basis elements
+
+      MatrixConstructor mat(R->make_FreeModule(n), c);
+      for (size_t i = 0; i < n; i++)
+        for (size_t j = 0; j < c; j++)
+          {
+            mpz_ptr z = newitem(__mpz_struct);
+            mpz_init_set(z, HB[i][j].get_mpz_t());
+            mpz_reallocate_limbs(z);
+            mat.set_entry(i, j, ring_elem(z));
+          }
+
+      return mat.to_matrix();
+  } catch (const exc::engine_error &e)
+    {
+      ERROR(e.what());
+      return nullptr;
+  }
+}
--- a/M2/Macaulay2/e/interface/cone.cpp
+++ b/M2/Macaulay2/e/interface/cone.cpp
@@ -20,6 +20,42 @@ typedef mpz_class Integer;
  * \ingroup cones
  */
 
+const Matrix /* or null */ *rawFourierMotzkin(const Matrix *C)
+{
+  try
+    {
+      const Ring *R = C->get_ring();
+      const size_t c = C->n_cols();  // rank of ambient lattice
+      const size_t r = C->n_rows();  // number of cone inequalities
+
+      auto ineqs = libnormaliz::Matrix<Integer>(r, c);
+      for (size_t i = 0; i < r; i++)
+        for (size_t j = 0; j < c; j++)
+	  // libnormaliz uses A*x >= 0, Macaulay2 uses A*x <= 0
+          ineqs[i][j] = (-1) * static_cast<Integer>(C->elem(i, j).get_mpz());
+
+      auto cone = libnormaliz::Cone<Integer>(libnormaliz::Type::inequalities, ineqs);
+      auto rays = cone.getExtremeRays();
+      size_t n = rays.size();  // number of extremal rays
+
+      MatrixConstructor mat(R->make_FreeModule(n), c);
+      for (size_t i = 0; i < n; i++)
+        for (size_t j = 0; j < c; j++)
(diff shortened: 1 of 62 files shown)


d-torrance:
Will do!

d-torrance:
Just pushed a commit updating the autotools build.  It's not quite working yet when we're building normaliz, but it should work when it already exists on the system.

mahrud:
Interesting. We don't need antic for now, but seems like Nauty is needed also:
```
Undefined symbols for architecture x86_64:
  "_densenauty", referenced from:
      libnormaliz::nauty_result<long> libnormaliz::compute_automs_by_nauty_Gens_LF<long>(libnormaliz::Matrix<long> const&, ...
```
When building it ourselves we certainly don't need either.

mahrud:
That seems to have done the trick. At some point making sure that libnormaliz builds successfully would be good, but I think this is good for now.

M2/Macaulay2/packages/NormalToricVarieties/Divisors.m2
```diff
@@ -531,10 +528,7 @@ vertices ToricDivisor := Matrix => D -> (
 latticePoints ToricDivisor := Matrix => D -> (
     V := vertices D;
     if V === null then return null;
-    d := numRows V;
-    V = transpose (normaliz (transpose V,"polytope"))#"gen";
-    s := select (numColumns V, i -> V_(d,i) === 1);
-    c := (V_s)^{0..d-1};
+    c := matrix(vector \ latticePoints convexHull V);
```
mahrud, line 531: @ggsmith with this PR, `isVeryAmple` and `latticePoints` for toric divisors are instead using `rawHilbertBasis` (called by Polyhedra after some hoops). Other changes in `NormalToricVarieties` are superficial (e.g. the symbol `isSmooth` is moving to Core, resulting in a bunch of documentation changes).

ggsmith:
I have a few questions:

- In ```FourierMotzkin```: I don't understand some of the changes the documentation node "Finding the facets of the cyclic polytope".  Why compute the vertices of the cyclic polytope, rather than just constructing them explicitly?  The original code seems more accessible to me.  Perhaps the best option is to include both by verifying that the more elementary construction coincides with the one obtain from the ```cyclicPolytope``` construction.
- In ```NormalToricVarieties```: as I understand it, the methods ```isVeryAmple``` and ```latticePoints``` now both depend on the ```Polyhedra``` package.  The earlier code was explicitly designed to avoid this.  Can this code not be altered to call ```rawHilbertBasis``` directly?  If not, then I would much prefer to wait to make this change until all the Normalize routines are available.
- In ```NormalToricVarieties```: in the documentation nodes where you removed links to the ```(isSmooth, NormalToricVariety)``` page, did you check that this link already appears in the ```SeeAlso``` section?
- In ```NormalToricVarieties/Tests.m2```: Why did you change the ```sort``` to a ```set```?

mahrud:
> * In `FourierMotzkin`: I don't understand some of the changes the documentation node "Finding the facets of the cyclic polytope".  Why compute the vertices of the cyclic polytope, rather than just constructing them explicitly?  The original code seems more accessible to me.  Perhaps the best option is to include both by verifying that the more elementary construction coincides with the one obtain from the `cyclicPolytope` construction.

I changed this because of the conflict in redefining `cyclicPolytope` from `Polyhedra`, but I'm happy to expand the documentation to show the explicit construction.

> * In `NormalToricVarieties`: as I understand it, the methods `isVeryAmple` and `latticePoints` now both depend on the `Polyhedra` package.  The earlier code was explicitly designed to avoid this.  Can this code not be altered to call `rawHilbertBasis` directly?  If not, then I would much prefer to wait to make this change until all the Normalize routines are available.

The Polyhedra methods `latticePoints` and `hilbertBasis` are a little obfuscated but are not doing anything fancy, only post-processing the output of `rawHilbertBasis`. Yes, the toric variety methods `isVeryAmple` and `latticePoints` can be altered to call `rawHilbertBasis` directly, but I'm against duplicating the raw post-processing code in Polyhedra. Specifically, computing lattice points of a polytope using a Hilbert basis computation is a piece of code that I think should be defined in one efficient routine rather than several places.

I should add that with this PR, NormalToricVarieties doesn't import the package Normaliz anymore, so I don't understand what would change when all Normaliz routines are available. Do you have another suggestion?

> * In `NormalToricVarieties`: in the documentation nodes where you removed links to the `(isSmooth, NormalToricVariety)` page, did you check that this link already appears in the `SeeAlso` section?

I did not, but instead I changed them all to link to the correct node.

> * In `NormalToricVarieties/Tests.m2`: Why did you change the `sort` to a `set`?

I'm actually not sure why I changed it so drastically, I'll change it back, but importantly `rawHilbertBasis` from Normaliz and 4ti2 have different output orders, which actually results in a desingularization with different order of rays in this example, so in the maximal cones 3 and 5 will be swapped.

mahrud:
@jkyang92 could you review the parts of this PR in the engine? I'm not confident if I dealt with memory matters safely.

@mikestillman could you enable requesting reviews from anyone? Currently I can't ask Jay to leave a review because it's only limited to [this team](https://github.com/orgs/Macaulay2/teams/macaulay2-contributors).

mikestillman:
@mahrud Do you know how I enable requesting reviews from anyone?  I don't immediately see where to do so...

mahrud:
I think `Settings > Moderation Options`.

mikestillman:
@mahrud, on github repo settings for M2, I look at 
```
Restrict users who are permitted to approve or request changes on pull requests in this repository.

Limit to users explicitly granted read or higher access
[]When enabled, only users explicitly granted access to this repository will be able to submit pull request reviews that "approve" or "request changes". All users able to submit comment pull request reviews will continue to be able to do so.
```
and there is a checkmark box, which is unchecked, for this option.  This seems to be the correct setting, but perhaps it should be checked?

M2/Macaulay2/e/interface/cone.cpp
```diff
@@ -0,0 +1,99 @@
...
+const Matrix /* or null */ *rawFourierMotzkin(const Matrix *C)
+{
+  try
+    {
+      // TODO: generalize the input type, in particular to allow lineality space
+      const Ring *R = C->get_ring();
+      const size_t c = C->n_cols();  // rank of ambient lattice
+      const size_t r = C->n_rows();  // number of cone inequalities
+
+      auto ineqs = libnormaliz::Matrix<Integer>(r, c);
```
jkyang92, line 32: It's slightly better to write ```libnormalize::Matrix<Integer> ineqs(r,c);```. Similarly later for ```libnormalize::Cone<Integer>```. This in theory avoids a copy/move, although here it's almost certainly too cheap to matter.

jkyang92:
@mahrud The memory management is fine. I added a comment on a minor unrelated thing I noticed.

M2/Macaulay2/e/interface/cone.cpp
mahrud: Hmm I think I understand what you mean, but I feel like this is more readable than `ineqs(r, c)` (it almost looks like `ineqs` is a function with two inputs) or `cone(libnormaliz::Type::inequalities, ineqs)`. Do you think the compiler would take care of avoiding this copy/move for us?

jkyang92: It's pretty standard c++ notation although for c++11 some suggest to always use ```libnormalize::Matrix<Integer> ineqs{r,c};``` note the braces. This avoids the following pitfall: https://en.wikipedia.org/wiki/Most_vexing_parse.

As for avoiding the copy, usually. In c++17, the compiler will always get rid of the copy (the copy doesn't exist in c++17 because of how the standard implements initialization), but prior to that, it's allowed to and usually it does, but it's not technically required.

DanGrayson:
Could you resolve the merge conflicts?

d-torrance:
@mahrud: Should any changes from this PR be mentioned in the 1.24.11 changelog?

PR #3317, opened 2024-06-19
Fixes vector overrun when defining the trivial monoid

### LINKED ISSUE #3316: Vector overrun in monoid.cpp
The Fedora project builds packages with `-D_GLIBCXX_ASSERTIONS`, which enables various assertions in gcc's C++ headers.  I just did a test build of version 1.24.05 and one such assertion failed:
```
/usr/include/c++/14/bits/stl_vector.h:1225: std::vector<_Tp, _Alloc>::const_reference std::vector<_Tp, _Alloc>::front() const [with _Tp = int; _Alloc = std::allocator<int>; const_reference = const int&]: Assertion '!this->empty()' failed.
```
The assertion is triggered on line 191 of `monoid.cpp`, on this line in `Monoid::set_degrees()`:
```
  auto *iter = &mDegrees.front();
```
In this case, `mDegrees` is empty, so the call to the `front` method is illegal.  This can be worked around as follows:
```
  auto *iter = mDegrees.data();
```
However, that just silences the assertion.  It doesn't solve the underlying issue, which is that in this case, `iter` is a pointer that cannot be dereferenced safely.  That is the answer to the question on line 212:
```
      // TODO: what is iter in this case?!
```
Perhaps iter should be set to `nullptr` in this case, so that attempts to dereference it will trigger a segfault.  What will happen on line 214 when `mDegreeMonoid->from_expvector(iter, m)` is called, and iter is a pointer that cannot be dereferenced?

mahrud:
- **added _GLIBCXX_ASSERTIONS to cmake debug builds**
- **fixed vector overrun when defining the trivial monoid**

Closes #3316. cc: @jamesjer.

### DIFF
--- a/M2/Macaulay2/e/BasicPolyListParser.cpp
+++ b/M2/Macaulay2/e/BasicPolyListParser.cpp
@@ -1,7 +1,6 @@
 #include "BasicPolyListParser.hpp"
 
 #include <iostream>
-#include <sstream>
 #include <fstream>
 
 std::string readEntireFile(const std::string &fileName)
@@ -10,7 +9,7 @@ std::string readEntireFile(const std::string &fileName)
 
   std::ifstream::pos_type fileSize = ifs.tellg();
   ifs.seekg(0, std::ios::beg);
-  
+
   std::vector<char> bytes(fileSize);
   ifs.read(bytes.data(), fileSize);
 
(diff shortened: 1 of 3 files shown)


mahrud:
@d-torrance it's not an issue with this PR, but I tried installing Macaulay2Doc with this debug flag and it seems like the example for "parallelism in engine computations" hangs. Could you take a look?

d-torrance:
Will do!

mahrud:
Actually it ended, but took quite a while, so maybe it's just because of a lot of glibc assertions:
```m2
 -- capturing example results for "parallelism in engine computations"       -- 965.351 seconds elapsed
```

PR #3319, opened 2024-06-20
Stop using PPA for Ubuntu GitHub builds

d-torrance:
This was only necessary when we were using Ubuntu 20.04 and not all of the dependencies were in the official archives.  This is no longer an issue in 22.04 and beyond.

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -80,7 +80,6 @@ jobs:
       - name: Install requirements for Linux
         if: runner.os == 'Linux'
         run: |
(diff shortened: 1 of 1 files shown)


d-torrance:
That PPA flint package was flawed because we didn't also have updated packages for factory, etc. 

Here's a list of the packages that are hosted in the PPA:  https://launchpad.net/~macaulay2/+archive/ubuntu/macaulay2

Outside of Macaulay2 itself, the only packages that are hosted are dependencies that aren't available in Ubuntu 18.04 or 20.04.  So we definitely don't need it for 24.04 either.

Issue #3251, opened 2024-05-21
error "incorrect Hilbert function given" in PrimaryDecomposition


paolo-lella:
While running the code
```macaulay2
R=QQ[x_1,x_2,x_3,y_1,y_2,z_1,z_2,w_1,w_2,Degrees=>{1..3,3:1..2}];
[...]

Issue #3176, opened 2024-04-08
primaryDecomposition issue with nonstandard grading


eloisagrifo:
input:

```
R = QQ[x,y,z, Degrees => {25,72,29}];
P = ker map(QQ[t],R,{t^(25),t^(72),t^(29)});
primaryDecomposition(P^3)
```

output:

`[stdio:3:20:(3)](https://www.unimelb-macaulay2.cloud.edu.au/#editor:stdio:3:20:(3)): error: array index 0 out of bounds 0 .. -1
`

This used to work as recently as version 1.16, but it's been broken for the last few versions of M2, and it is still broken in version 1.23.0.1. There was a discussion about this in the google groups in March 2022 and the conclusion seemed to be that the issue is with the Birational strategy in minimal primes.

d-torrance:
Reproducing this with `errorDepth` set to 2, we get:

```m2
i4 : primaryDecomposition (P^3)
/usr/share/Macaulay2/MinimalPrimes/AnnotatedIdeal.m2:202:34:(2):[56]: error: array index 0 out of bounds 0 .. -1
/usr/share/Macaulay2/MinimalPrimes/AnnotatedIdeal.m2:202:34:(2):[56]: --entering debugger (type help to see debugger commands)
/usr/share/Macaulay2/MinimalPrimes/AnnotatedIdeal.m2:202:18-202:34: --source code:
        S := ring I.LexGBOverBase#0; -- should be of the form kk(indepvars)[fibervars]

ii5 : peek I

oo5 = AnnotatedIdeal{Birational => true                                         }
                     DecomposeMonomials => true
                     Factorization => true
                     IndependentSet => ({z}, QQ[y, x, z], frac(QQ[z])[y, x])
                     isPrime => YES
                     LexGBOverBase => {}
                     LexGBSplit => true
                     Linear => true
                     Linears => {}
                     NonzeroDivisors => {}
                                      3    4 4   29     18   7     7 2 14    25
                     Ideal => ideal (y  - x z , x   - 3x  y*z  + 3x y z   - z  )
                                   11      7
                     Inverted => {x   - y*z , z}
```
So the bug seems to be that `I.LexGBOverbase` is an empty list, but we're trying to access its 0th element.

PR #3320, opened 2024-06-20
This fixes github issues #3251, #3176. 

mikestillman:
The problem: nonstandard gradings , when using the Independent sets code in M2 for computing minimal primes, was using a presumed Hilbert function that was not correct as the ideal was not even homogeneous.

This fix improves the check for homogeneity in splitIdeals#IndependentSets, and also in the function makeFiberRings in MinimalPrimes.m2, it now uses the original grading when it permutes the variables, keeping the new ideal homogeneous.

This problem has been found by others as well, including Frank Schreyer.

A couple of tests have been added.


d-torrance:
@mikestillman - Should we mention the changes from this PR in the 1.24.11 changelog?

PR #3334, opened 2024-06-25
Moved installation notes to the wiki

mahrud:
As discussed in #3294:

- deleted INSTALL.mac in favor of [Building-M2-from-source-on-macOS](https://github.com/Macaulay2/M2/wiki/Building-M2-from-source-on-macOS)
- deleted INSTALL in favor of [Building-M2-from-source-using-Autotools](https://github.com/Macaulay2/M2/wiki/Building-M2-from-source-using-Autotools)
- deleted INSTALL-CMake.md in favor of [Building-M2-from-source-using-CMake](https://github.com/Macaulay2/M2/wiki/Building-M2-from-source-using-CMake)
- renamed README -> README.md
- deleted obsolete gawk script

The autotools install page on wiki still needs some work, but at least it contains all the information previously in `INSTALL`.

Skipping CI builds because only touching text files and an obsolete gawk file not referenced from anywhere (used to be used by GNUmakefile.in but it has long been removed).


M2/LICENSING-NOTES
```diff
@@ -1,5 +1,24 @@
...
+                                   Macaulay2
+                   by Daniel R. Grayson <dan@math.uiuc.edu>
+                and Michael E. Stillman <mike@math.cornell.edu>
+
+                     available from http://Macaulay2.com/
+
+We thank the United States National Science Foundation for generous funding
+to develop Macaulay2.
+
+Macaulay2 is copyright 1993-2016 by Daniel R. Grayson and Michael E. Stillman.
```
d-torrance, line 10: I suppose this should be 1993-2024?  The date is also in `M2/Macaulay2/README.in`, so I'm not sure how many times we want to duplicate it.

mahrud: Should I just delete it then?

d-torrance: I think `LICENSING-NOTES` is maybe a more natural place to put the date, so maybe leave it here to keep updated and remove it from the README?

DanGrayson:
Could you fix the merge conflicts, please?

PR #3335, opened 2024-06-26
Use unsigned 64-bit integers for hash codes

d-torrance:
This should decrease the probability of running out significantly.  Currently, we can only create 2147483647 mutable hash tables before we get a "hash code serial number counter overflow".  This bumps that up to 18446744073709551615.

We use a new integer type, `hash_t` (a `#define` of `uint64_t`), to simplify things if we ever decide to change types again.

This is something we've discussed a few times at M2internals.  It was a big game of whack-a-mole to find all the places we had hard-coded the type of hash codes, but I finally got around to doing it!


d-torrance:
TODO:

- [ ] Figure out the weird "'RTLD_DEFAULT' undeclared" bug on the Ubuntu builds
- [ ] Update the hash codes of the cached examples so we don't try to regenerate them

M2/Macaulay2/d/expr.d
```diff
@@ -35,10 +35,12 @@ export nextHash():hash_t := (
      HashCounter);
 
 -- Knuth, Art of Computer Programming, Section 6.4
+-- constant is approx. 2^64 / phi
 export fibonacciHash(k:hash_t,p:int):hash_t := (
-    Ccode(hash_t, "(2654435769 * ",k,") >> (32 - ",p,")"));
+    Ccode(hash_t, "(11400714819323198485ull * ",k,") >> (64 - ",p,")"));
 
 -- hash codes for mutable objects that don't use nextHash
+-- we use 9 since #buckets ZZ = 2^9
```
mahrud, line 43: Why is `#buckets ZZ` relevant here? `#buckets Matrix = 2^10`.

d-torrance: It was just an arbitrary decently large hash table.  The algorithm says that if $2^n$ is the number of bits in our hash code, multiply $2^n/\varphi$ by the input, and then shift that by $n - m$, where $2^m$ is the size of the hash table.  But our hash tables vary in size and we won't know which hash table we might be stuffing this thing into, so I just picked `ZZ`.

mahrud:
> * Something's goofy about hashes of mutable hash tables on 32-bit architectures (e.g., `hash ZZ` gives 1053710097531506 instead of the expected 1000050).  Does this have something to do with how thread local variables work?

Was this resolved?

d-torrance:
Not yet

Issue #3207, opened 2024-05-07
Building with Nix - various problems with the autoconf build process


pdg137:
Hello,

As a personal project, I've been trying to build Macaulay2 with the [Nix build system](https://nixos.org/). The advantage of a Nix build is that it's available on many platforms - for example it would automatically support almost every Linux distribution - and gives a well-defined, predictable result, independent of the state of the OS it's built on. It's also a kind of extremely dependable documentation for *how* to build a package, replacing the vague or incomplete instructions in INSTALL-type documents.

Look at [the math packages](https://github.com/NixOS/nixpkgs/tree/nixos-23.11/pkgs/applications/science/math) already available in Nix to see how much they are missing M2!

I am not a Nix expert, don't know much about Macaulay2 either, and have not managed to get it all the way through the build process, but I think I've already uncovered some difficulties that might be worth addressing. Also, in case there are any NixOS users in the M2 community, maybe someone will be interested in using this as a starting point:

https://github.com/pdg137/macaulay2-build

Note that I am using the autoconf build process. Here's what I have run into:

* Several dependencies are not checked by autoconf and will cause build failures much later than necessary: bison, perl, texinfo, time (the time command isn't even necessary; I just removed it from the build)
* There are several places where the build process tries to run git commands, not recognizing that I have already checked out all the source code. This fails since a) I don't have network access at build time and b) I am not providing git as a build input.
* One spot where a shell script `autogen.sh` is assumed to be executable; it's more reliable to run it with `sh`.
* Failure to generate some version string when double quotes are returned by `lsb_release`.
* Some mysterious problem running `dlopen "libmpfr.so"` (this is what I'm working on now - it fails to generate the documentation for openSharedLibrary). See [my issue about it](https://github.com/pdg137/macaulay2-build/issues/1).

Most of these are addressed in a blunt way with my [patch file](https://github.com/pdg137/macaulay2-build/blob/master/macaulay2.patch). For example, I just removed any git stuff that was getting in the way of the build. But it would be great to see them fixed more cleanly here in the main repository.

An additional problem is this:

* The documentation takes a really long time to build, maybe more than half of the build time.

Is there a way to build and install the M2 binary without the docs? Breaking it up this way would allow me to say I've successfully built an mostly-working M2 that might already be useful to others.

Related:

*  https://github.com/NixOS/nixpkgs/pull/251814
* https://github.com/NixOS/nixpkgs/issues/186823

d-torrance:
Very cool!

[...]

PR #3338, opened 2024-06-28
Check for a few more programs in autotools builds

d-torrance:
There are a number of programs (`time`, `patch`, `bison`, and `install-info`) that are required to build Macaulay2 using the autotools build, but we don't tell the user this when running `configure`.  This leads to surprises later on.  A few of these were mentioned in #3207.

We add checks for each of these.


d-torrance:
I keep finding little things to add to this PR while building Macaulay2 on various Docker images while I work on the autotools wiki page.

PR #3339, opened 2024-06-30
changed caching in minimalBetti

mahrud:
- **cached the ChainComplex, instead of Resolution, in minimalBetti**
- **simplified caching for res and minimalBetti of ideals**

### DIFF
--- a/M2/Macaulay2/m2/betti.m2
+++ b/M2/Macaulay2/m2/betti.m2
@@ -212,13 +212,14 @@ minimalBetti = method(
 	})
 minimalBetti Module := BettiTally => opts -> M -> (
     R := ring M;
+    weights := heftvec(opts.Weights, heft R);
     degreelimit := resolutionDegreeLimit(R, opts.DegreeLimit);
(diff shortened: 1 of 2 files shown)


mahrud:
Hmm this is unfortunate, because the changes from Varieties package during the workshop also involve a lot of changes to Complexes. I was hoping there would be some communication when this was going to happen.

mikestillman:
When I say: "as we speak", that means we are working on it.  Currently, on a branch, we can preload Complexes automatically, but it will still take some effort to tease out the nontrivial use of ChainComplex.  That said, what other changes do you envision for Complexes? (other than this and PR #3342 ?

mahrud:
Those are the only ones where the work is already done, but I don't mind repeating the changes in this PR later. Only #3342 could be a lot more work if there are conflicts.

Other things involve future changes:
- retire `PruneComplex` and merge it in `Complexes` because `minimize` is currently too slow.
- we need `(resolution, Complex)` to be functorial, and I want to talk to you and Greg about how to do this.

mahrud:
Ah, I don't think Mike wanted this to be merged ...

PR #3342, opened 2024-07-02
Changes to Complexes from Utah workshop

mahrud:
These are changes to the Complexes (and Truncations) package that I extracted from the Varieties branch from the recent workshop. Since Mike mentioned changes to the Complexes are happening now, I extracted these commits now, hoping to prevent a big merge conflict in the future.

- **added formation for Ext modules in Complexes**
- **added length limits for connectingExtMap**
- **added method for truncating source of map more than target**
- **added inducedTruncationMap for truncating complexes**
- **added better algorithm for truncate(ZZ, Complex)**
- **added auxiliary methods for subtruncate**
- **fixed load paths in Complexes**
- **added basis for complexes**
- **removed strange definition of cohomology of complex of modules**
- **removed calls to length Complex in resolutionMap**
- **fixed basis and truncate for complexes**
- **fixed map(Complex, Complex, ZZ) to allow multiplication maps**

cc: @mikestillman @ggsmith


M2/Macaulay2/packages/Complexes/ChainComplexDoc.m2
```diff
@@ -2017,25 +2017,24 @@ doc ///
...
 
 
 
 doc ///
    Key
      (homology,ZZ,Complex)
-     (cohomology,ZZ,Complex)
    Headline
-     homology or cohomology module of a complex
+     homology of a complex
```
mahrud, line 2029: `HH^i C` was removed because we want to use this for complexes of sheaves, and it wasn't doing anything deep here, just `HH_-i C` which the user can do as well.

mahrud:
@mikestillman I'm gonna close this and make it directly to your branch instead.

PR #3343, opened 2024-07-03
Use VERSION instead of INSTALL for AC_CONFIG_SRCDIR

d-torrance:
INSTALL was removed in #3334, but the builds were skipped so we didn't notice this was an issue.


d-torrance:
@DanGrayson - I'm going to go ahead and merge this so the autotools builds in other PR's aren't broken.

mahrud:
Ah, my bad!

PR #3347, opened 2024-07-04
A couple autotools libnormaliz detection updates

d-torrance:
This is a quick followup to #3313 to improve the detection of libnormaliz in the autotools build.

Two changes:

* Instead of including `libnormaliz.h`, we include `cone.h`, since the former doesn't include the latter on older systems (e.g., Ubuntu 18.04) and we get a compile error trying to construct a `Cone` object.
* Check whether libnormaliz was built with e-antic (by checking if `ENFNORMALIZ` was defined in `nmz_config.h`), and if so, add the appropriate linker flags.

### DIFF
--- a/M2/configure.ac
+++ b/M2/configure.ac
@@ -1018,7 +1018,7 @@ AS_IF([test $BUILD_normaliz = no],
 	       LIBS="$LIBS -lnormaliz -lnauty -lgmp"
 	       AC_LINK_IFELSE(
 		   [AC_LANG_PROGRAM([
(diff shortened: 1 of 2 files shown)


d-torrance:
I added one more commit for when we're *building* libnormaliz.  We add a bunch of `--without-*` options to `configure`, as otherwise it will check to see if these libraries are present on the system, and if they are, it will use them.  But we wouldn't know which flags to add when linking Macaulay2.  The cmake build already does this.

PR #3349, opened 2024-07-05
Moved containment and changed equality of modules

### LINKED ISSUE #3346: bug in equality of complexes (really equality of zero modules)
Here's an unfortunate situation that happened today:
```m2
restart
needsPackage "Complexes"

S = kk[a..d]
psi = map(kk, S)
I = ideal"ab, ad, bc, c3"
F = freeResolution comodule I

C = psi complex applyValues(F.dd.map, basis_0)
D = part(0, F)
prune C == D -- true
C == D       -- false!!!!!!
```
Looking at the output, you might guess that it's because concentrations aren't equal:
```m2
i6 : C = psi complex applyValues(F.dd.map, basis_0)

o6 = image | 1 | <-- image 0 <-- image 0 <-- image 0
                                              
     0               1           2           3

o6 : Complex

i7 : D = part(0, F)

       1
o7 = kk
      
     0

o7 : Complex
```
But in reality it's because M2 treats the zero submodule in different ambient modules to be different:
```m2
M = source C.dd_1
N = source D.dd_1
M == 0 -- true
N == 0 -- true
M == N -- false
```
And this is why the modules are considered not the same:
```m2
i12 : M = source C.dd_1

o12 = image 0

                                4
o12 : kk-module, submodule of kk

i13 : N = source D.dd_1

o13 = 0

o13 : kk-module
```
I can imagine the argument that these two modules should not be equal, but I think the equality of the complexes above failing is obviously a bug. So I would like to know the arguments for having these modules be unequal. Does it break something? Should we just change them to be equal?

I personally think they should be equal, keeping in mind that they are not === equal.

cc: @mikestillman

mahrud:
I've been meaning to review this part of the modules code and and add hooks for it, and issue #3346 provided an opportunity to do this.

- simplified issub using tryHooks
- rewrote and hookified equality of modules
- added tests for equality of modules
- moved and cached Module == ZZ
- moved and simplified isSubset
- fixed equality of zero modules
  - fixes #3346 
- fixed incorrect ambient module in saturation
- added GroebnerBasis == GroebnerBasis
- combined generic documentation of equality

The fix for equality of zero modules is based on the following factors:
1. we don't want to compute a Groebner basis of two huge, non-zero module if for degree reasons they can't be the same;
2. we don't want to say two zero submodules of (possibly different) free modules are non-equal because the degrees of ambient modules don't match.

So I implemented a compromise, which is that two zero modules are equal if they are equal as cosets, i.e. `gens M | relations M == 0 and gens N | relations N == 0`. This means:
```m2
0 * S^1 == 0 * S^{2}
0 * S^1 == 0 * S^{1,1}
0 * S^1 != subquotient(matrix"x", matrix"x")
0 * S^1 != subquotient(matrix{{0_S}}, matrix"x")
```
The only implication is that zero submodules of any free module are now the same, but zero submodules of arbitrary modules are only the same if they have the same ambient module.


mahrud:
@mikestillman this was merged, but I think you should still review the change to equality of modules.

PR #3353, opened 2024-07-09
MatrixSchubert - Fixing isPatternAvoiding

### LINKED ISSUE #3291: isPatternAvoiding sometimes incorrect
The function `isPatternAvoiding` (and the related `avoidsAllPatterns`) from the `MatrixSchubert` package sometimes gives the incorrect output. For example:

```
i1 : needsPackage("MatrixSchubert");
i2 : isPatternAvoiding({3,1,2,4},{3,1,2})
o2 = true
i3 : isPatternAvoiding({3,1,2,4},{3,1,2,4})
o3 = true
```
Of course, sometimes it *is* also correct!
```
i4 : isPatternAvoiding({3,1,2,4},{2,1})
o4 = false
i5 : isPatternAvoiding({1,2,3,4}, {1,2,3,4})
o5 = false
```

jmcdonough98:
isPatternAvoiding was checking if the permutation avoided the inverse of the given pattern instead of the actual pattern (all of the tests patterns were involutions so this wasn't spotted). It now properly checks if a permutation avoids the right pattern.
This fixes #3291

### DIFF
--- a/M2/Macaulay2/packages/MatrixSchubert/permutationMethods.m2
+++ b/M2/Macaulay2/packages/MatrixSchubert/permutationMethods.m2
@@ -224,18 +224,14 @@ isPatternAvoiding (List,List) := Boolean => (perm, pattern) -> (
     --input validation
     if not (isPerm perm) then error(toString perm | " is not a permutation.");
     --assume permutation is pattern-avoiding, break if not true
-    isAvoiding := true;
     for idx in subsets(0..#perm-1, #pattern) do {
-        vals := perm_(sort idx);
+        vals := perm_(idx);
         sortedVals := sort(vals);
         relPositions := hashTable toList apply(0..#vals-1, i -> {sortedVals#i, i});
         p := toList apply(vals, i -> (relPositions#i) + 1); 
-        if p == pattern then {
-            isAvoiding = false;
-            break;
-        };
+        if p == pattern then return false;
     };
-    isAvoiding
+    true
 )
 
 --------------------------------
--- a/M2/Macaulay2/packages/MatrixSchubert/permutationMethods.m2
+++ b/M2/Macaulay2/packages/MatrixSchubert/permutationMethods.m2
@@ -226,10 +226,10 @@ isPatternAvoiding (List,List) := Boolean => (perm, pattern) -> (
     --assume permutation is pattern-avoiding, break if not true
     isAvoiding := true;
     for idx in subsets(0..#perm-1, #pattern) do {
-        vals = perm_(sort idx);
-        sortedVals = sort(vals);
-        relPositions = hashTable toList apply(0..#vals-1, i -> {sortedVals#i, i});
-        p = toList apply(vals, i -> (relPositions#i) + 1); 
+        vals := perm_(sort idx);
+        sortedVals := sort(vals);
+        relPositions := hashTable toList apply(0..#vals-1, i -> {sortedVals#i, i});
+        p := toList apply(vals, i -> (relPositions#i) + 1); 
         if p == pattern then {
             isAvoiding = false;
             break;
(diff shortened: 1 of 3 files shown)


M2/Macaulay2/packages/MatrixSchubert/permutationMethods.m2
```diff
@@ -226,10 +226,14 @@ isPatternAvoiding (List,List) := Boolean => (perm, pattern) -> (
     --assume permutation is pattern-avoiding, break if not true
     isAvoiding := true;
     for idx in subsets(0..#perm-1, #pattern) do {
```
d-torrance, line 228: This isn't really a comment for this PR specifically, but just a comment in general regarding `subsets`.  It would be great if it returned at iterator (maybe as an option) so we wouldn't need to pre-compute *all* $\binom{n}{k}$ subsets at the beginning.  If we find out that it's not pattern avoiding early in the process, then we would have computed all the other subsets for no reason.

M2/Macaulay2/packages/MatrixSchubert/permutationMethods.m2
```diff
@@ -226,10 +226,14 @@ isPatternAvoiding (List,List) := Boolean => (perm, pattern) -> (
...
-        sortedIdx := sort(idx);
-        pairwiseComparison := apply(pattern_{0..#pattern-2}, pattern_{1..#pattern-1}, (i,j) -> perm#(sortedIdx#(i-1)) < perm#(sortedIdx#(j-1))); -- pairwise comparison of permutation according to pattern
-        isAvoiding = not all(pairwiseComparison, i -> i == true); -- true if there was one inequality that failed, else all inequalities are true and so not pattern-avoiding
-        if not isAvoiding then break;
+        vals := perm_(sort idx);
```
d-torrance, line 229: I'm almost certain that `idx` will already be sorted, so this call to `sort` is likely redundant.

M2/Macaulay2/packages/MatrixSchubert/permutationMethods.m2
```diff
@@ -226,10 +226,14 @@ isPatternAvoiding (List,List) := Boolean => (perm, pattern) -> (
...
+        sortedVals := sort(vals);
+        relPositions := hashTable toList apply(0..#vals-1, i -> {sortedVals#i, i});
+        p := toList apply(vals, i -> (relPositions#i) + 1); 
+        if p == pattern then {
+            isAvoiding = false;
+            break;
```
d-torrance, line 235: I don't think the `isAvoiding` variable is necessary.  We could have:

```m2
  if p == pattern then return false
  )
true
)

**Comment by jmcdonough98** [comment]
Thanks for the suggestions! I implemented them but now it looks like an unrelated package is running out of memory during its tests on one of the builds. 
I poked around a bit and this seems to be related to #3239

**Comment by mahrud** [comment]
Joe if you get a chance, do you mind also fixing these warnings?
```
[691/692] Installing package MatrixSchubert
 -- warning: newPackage: use the Contributors or Acknowledgement keywords to acknowledge contributors in the package documentation
 -- warning: document: documentation headlines must be less than 100 characters long:
 --   "compute a table of rank conditions that determines a Schubert determinantal ideal or, more generally, an alternating sign matrix ideal."
 -- warning: document: documentation headlines must be less than 100 characters long:
 --   "find the Rothe diagram of a partial alternating sign matrix together with the rank table determining the alternating sign matrix variety"
 -- warning: document: documentation headlines must be less than 100 characters long:
 --   "compute the essential set in the Rothe Diagram for a partial alternating sign matrix or a permutation."
 -- warning: document: documentation headlines must be less than 100 characters long:
 --   "find the essential set of a partial alternating sign matrix or a permutation together with the rank conditions determining the alternating sign matrix variety"
 -- warning: document: documentation headlines must be less than 100 characters long:
 --   "compute the Castelnuovo-Mumford regularity of the quotient by a Schubert determinantal ideal or ASM ideal"
```

Issue #3351, opened 2024-07-07
bug in ??= and mutable hash tables


mahrud:
Why is this happening?
```m2
i1 : X = Proj QQ[x];

i2 : X.sheaf ??= ring X

o2 = QQ[x]

o2 : PolynomialRing

i3 : X.sheaf ??= ring X
stdio:3:9:(3): error: expected method to be a function or TYPE => function'
```
cc: @d-torrance

d-torrance:
It's because `QQ[x]` is a `PolynomialRing`, which inherits from `Type`, and `Type ??=` is syntactic sugar for calling `installMethod` to install a `??=` method for a given type.  So it's yelling at us that `ring X` isn't a function.

Your use case definitely seems more important than having that particular syntactic sugar.  I'll open a PR soonish removing it (probably in the next few days).

mahrud:
Hmm shouldn't the syntax for installing `??=` methods have a key like `((??=, =), Type, Type)`? What is an example of using this syntactic sugar?

d-torrance:
Right now they key is just `??=`.  So `installMethod(symbol ??=, T, f)` will set it so that `x ??= y` (when `x` is an instance of `T`) will call `f(x, y)`.

I don't think the syntactic sugar is currently used anywhere outside of the documentation (see [installing augmented assignment methods](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_installing_spaugmented_spassignment_spmethods.html)

mahrud:
I'm not sure if I understand. In the bug above, is `Type ??= Type` triggering a syntactic sugar for installing a method on `(symbol ??=, Type)`? I can't figure out how to use this syntactic sugar without `installMethod`.

d-torrance:
It's installing a method on `(symbol ??=, QQ[x])`

PR #3355, opened 2024-07-10
Remove syntactic sugar for installing augmented assignment methods

d-torrance:
We may want to actually use an augmented assignment operator to augmentedly assign something to a type, and the current syntactic sugar of `T += f` as a shortcut for `installMethod(symbol +=, T, f)` may cause unexpected behavior.

Closes: #3351

### DIFF
--- a/M2/Macaulay2/tests/normal/augmented-assignment.m2
+++ b/M2/Macaulay2/tests/normal/augmented-assignment.m2
@@ -126,7 +126,7 @@ assert Equation(x#0, 5)
 
 -- install custom method
 Bar - Bar := (x, y) -> Bar(x#0 - y#0)
-Bar -= (x, y) -> if even y#0 then x#0 = 0 else Default
+installMethod(symbol -=, Bar, (x, y) -> if even y#0 then x#0 = 0 else Default)
 x = Bar 5
 x -= Bar 2
 assert Equation(x#0, 0)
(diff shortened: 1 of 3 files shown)


M2/Macaulay2/tests/normal/augmented-assignment.m2
```diff
@@ -126,7 +126,7 @@ assert Equation(x#0, 5)
 
 -- install custom method
 Bar - Bar := (x, y) -> Bar(x#0 - y#0)
-Bar -= (x, y) -> if even y#0 then x#0 = 0 else Default
```
mahrud, line 129: Ohhh wait I see why I've been confused about this. I agree that this looks strange and should be removed, even if it wasn't for the problem I ran into. I think method keys for installing a method involving operators should be consistent, and currently they're all `((1OP, =), X)` or `((2OP, =), X, Y)`, not just `(OP, X)` or `(OP, X, Y)`. I actually would prefer changing the `=` to `:=` to make it absolutely clear that this is a method installation.

Perhaps an alternative syntactic sugar here could be:
```m2
Bar -= Bar := (x,y) -> ...
```

d-torrance: All the `((1OP, =), X)` and `((2OP, =), X, Y)` methods actually involve the `=` operator, though, e.g., `((+, =), X, Y)` tells you what to do when there's something like `x + y = z`, where `x` and `y` are instances of `X` and `Y`, respectively.    Note also that we don't worry about the type of `z` when installing the method.

PR #3357, opened 2024-07-12
Update PositivityToricBundles to v1.8

a-hochenegger:
Fixed a bug in an internal method used for `compatibleBases` and `toricChernCharacter`:
the method made an assumption on the form how the filtration steps are ordered in the bundles generated by the package `ToricVectorBundles`. Usually true, this assumption does not apply, if the bundle arises by using the method `dual` of `ToricVectorBundles` (e.g. cotangent bundles).


mahrud:
Could you add an example where the previous version fails as an additional test?

a-hochenegger:
There is no such a test (that would have failed in the previous versions).

d-torrance:
I think the macOS builds were failing because this PR originally targeted `master` and so an older version of factory w/o FLINT 3 support was being built.  I just rebased it on top of `development`, which should fix this.

a-hochenegger:
@d-torrance thank you very much! I was already wondering about this failing builds (but I had no idea how my changes could have caused them). In order to avoid this errors in the future: should I always choose `development` as the target of my pull requests?

d-torrance:
Yes, always target `development` for new pull requests.  (If you forget, then one of us will likely change it for you so it gets merged into the correct branch.  But that doesn't help tests, unfortunately.)

Issue #3358, opened 2024-07-14
Hash counter should be atomic, not thread local


d-torrance:
Consider the following example:

```m2
i1 : Foo = taskResult schedule(() -> new Type of HashTable)

o1 = Foo

o1 : Type

i2 : youngest(Foo, Matrix)

o2 = Matrix

o2 : Type
```

That doesn't seem right!  Any user-defined classes should definitely be younger than ones defined in `Core` so that garbage collection works as expected.

But indeed, `Foo` has a much smaller hash code:

```m2
i3 : hash Foo

o3 = 1050004

i4 : hash Matrix

o4 = 1150524
```

This is because the hash counter is thread local, and we only have a gap of 10,000 between the starting point for each thread:

https://github.com/Macaulay2/M2/blob/ec9e9ac60ed4a8e791448942202f077a88a87e15/M2/Macaulay2/d/expr.d#L27

Instead, I think we might consider using an atomic hash counter.

d-torrance:
Fixed in #4048

PR #3359, opened 2024-07-14
 Add additional space to store 64-bit hash codes on 32-bit machines

d-torrance:
This is a followup to #3335, checking off the final checkbox from the TODO list in https://github.com/Macaulay2/M2/pull/3335#issuecomment-2191677295.

On 32-bit machines, thread local variables are stored in arrays of 32-bit chunks of memory.  But now that we use 64-bit hash codes, we were getting the hash counter by concatenating two adjacent elements of the array, giving us unexpected results.

We add a new thread local variable to store these extra 32 bits, fixing this issue.  This will be unnecessary if we switch to an atomic hash counter (see #3358).

Now that we can actually see what the expected result should be, we also update the hash code test for `RR` objects for 32-bit machines.

### DIFF
--- a/M2/Macaulay2/d/expr.d
+++ b/M2/Macaulay2/d/expr.d
@@ -27,6 +27,9 @@ threadCounter := 0;
 threadLocal HashCounter := (
     threadCounter = threadCounter + 1;
     hash_t(1000000 + 3 + (threadCounter-1) * 10000 ));
+-- give 32-bit machines enough space to store a 64-bit hash code
+-- TODO: instead, allow 64-bit entries in the array of thread local variables
+threadLocal HashCounterExtraBits := 0;
 
 export nextHash():hash_t := (
      if HashCounter == Ccode(hash_t, "18446744073709551615ull") -- check for integer overflow
--- a/M2/Macaulay2/m2/basictests/hashcodes.m2
+++ b/M2/Macaulay2/m2/basictests/hashcodes.m2
@@ -22,7 +22,7 @@ assert( (hash true) === 444777 )
 assert( (hash false) === 777333 )
 
 -- these might change if our floating point implementation changes, but let's check anyway:
-assert( hash 1.23p200 === 18446744072207201388 -* big endian *- or hash 1.23p200 == -640232547 -* little endian *- )
+assert( hash 1.23p200 === 18446744072207201388 -* 64-bit *- or hash 1.23p200 == 18446744073069319069 -* 32-bit *- )
 
 -- Local Variables:
 -- compile-command: "make -C $M2BUILDDIR/Macaulay2/packages/Macaulay2Doc/basictests hashcodes.okay"


mahrud:
Wait so you're saying we earmark 32-bits extra on the stack so that when we read 64 bits from the hash counter we don't get junk? I'm concerned that we're reading a 64-bit integer on 32-bit systems! Why not make threadlocal variables in interpreter all 64 bits?

Also, is the problem with using `size_t` that hashes will be different on 32-bit vs 64-bit systems? Maybe we should make a `hash_t` typedef so this is clearer?

Tangentially, are there users out there with 32-bit systems who use the most recent M2 versions? (i.e. not users still running M2 v1.9 or something)

d-torrance:
> Wait so you're saying we earmark 32-bits extra on the stack so that when we read 64 bits from the hash counter we don't get junk?

Yes, exactly.

> I'm concerned that we're reading a 64-bit integer on 32-bit systems! Why not make threadlocal variables in interpreter all 64 bits?

That might be a better solution, but I think all of the other thread local variables are 32-bit ints.

> Also, is the problem with using `size_t` that hashes will be different on 32-bit vs 64-bit systems? Maybe we should make a `hash_t` typedef so this is clearer?

hash_t is already a typedef for uint64_t.

> Tangentially, are there users out there with 32-bit systems who use the most recent M2 versions? (i.e. not users still running M2 v1.9 or something)

Probably not. I occasionally run it on my 32-bit raspberry pi for giggles. I need it to build on several 32-bit architectures for Debian, though.

mahrud:
> That might be a better solution, but I think all of the other thread local variables are 32-bit ints.

I could be wrong, but I think they're just `int`s, so which means somewhere down the line we might run into a situation involving thread local variables where the result (or at least a hash) is different between a 32-bit and 64-bit machine.

> I need it to build on several 32-bit architectures for Debian, though.

Does Debian require all packages to build on 32-bit archs? I've always been a bit concerned about tests or examples that were removed because they failed on rare architectures.

d-torrance:
> I could be wrong, but I think they're just `int`s, so which means somewhere down the line we might run into a situation involving thread local variables where the result (or at least a hash) is different between a 32-bit and 64-bit machine.

Theoretically, an `int` could be as little as 16 bits, but in practice, it's 32 bits on all architectures that anyone actually uses.  (It's `long` that's the annoying one that varies.)

> Does Debian require all packages to build on 32-bit archs? I've always been a bit concerned about tests or examples that were removed because they failed on rare architectures.

Kind of.  There are three officially supported 32-bit architectures (i386, armhf, and armel), plus a bunch of non-official ones.  If Macaulay2 stops building on any of the three official ones, then it won't migrate from unstable to testing on *any* of the architectures, 32-bit or 64-bit.  There are ways around this if we ever decide to completely remove 32-bit support one day, though.

---

Converting this to a draft.  I like the idea of making the array of thread-local variables hold 64 bits per slot instead of 32.

d-torrance:
I've looked at this a little more, and I'm not sure how to change the number of bits per element of the array without some more serious study of how pthreads work.  For now, I'm proposing that we keep this hacky solution so that hashes work properly on 32-bit systems.  I'll add a TODO comment.

mahrud (approved):
Sounds good! I'll merge once tests past.

Issue #1026, opened 2019-10-09
value and toExternalString... deprecated?


antonleykin:
Is `toExternalString` deprecated? 
`Serialization` package seems to be a newer solution to "pickling"... but it is not mentioned by `help toExternalString`.

It seems that to be able to pickle and unpickle a matrix (with `toExternalString`) one needs to have the ring and its generators as global variables, otherwise...
 
```
i1 : QQ[x]

o1 = QQ[x]

o1 : PolynomialRing

i2 : toExternalString matrix{{x}}

o2 = map((QQ[x])^1,(QQ[x])^{{-1}},{{x}})

i3 : value oo
currentString:1:1:(3):[1]: error: modules over incompatible rings
currentString:1:1:(3):[1]: --entering debugger (type help to see debugger commands)
currentString:1:1-1:32: --source code:
map((QQ[x])^1,(QQ[x])^{{-1}},{{x}})
```

DanGrayson:
It's not deprecated, but it does have its shortcomings, as you can see.  At least it covers lots of cases.  It might be worth putting a cross reference to Serialization, as you note, which works well, at least on this example:
```

i5 : QQ[x]

o5 = QQ[x]

o5 : PolynomialRing

i6 : f = matrix {{x}}

o6 = | x |

                   1             1
o6 : Matrix (QQ[x])  <--- (QQ[x])

i7 : serialize f

o7 = -- -*- mode: M2; coding: utf-8 -*-
     s0:=QQ -- Ring : QQ
     s1:=global QQ -- Symbol : QQ
     s2:=monoid[x, Degrees => {1}, Heft => {1}, MonomialOrder => VerticalList{MonomialSize => 32, GRevLex => {1}, Position => Up}, DegreeRank => 1] -- GeneralOrderedMonoid : monoid[x, Degrees => {1}, Heft => {1}, MonomialOrder => VerticalList{MonomialSize => 32, GRevLex => {1}, Position => Up}, DegreeRank => 1]
     s3:=s0 s2 -- PolynomialRing : QQ[x]
     s4:=s3^1 -- Module : (QQ[x])^1
     s5:=s3^{{-1}} -- Module : (QQ[x])^1
     s6:=s3_{1} -- QQ[x] : x
     s7:=map(s4,s5,{{s6}}) -- Matrix : matrix {{x}}
     s7

i8 : "/tmp/foo" << o7 << close

o8 = /tmp/foo

o8 : File

i9 : 

Process M2 finished

[...]

PR #3362, opened 2024-07-15
Add toExternalString(MutableMatrix)

d-torrance:
### Before
```m2
i1 : toExternalString mutableMatrix {{1, 2}, {3, 4}}
stdio:1:1:(3): error: can't convert local symbol or invisible global symbol 'RawMutableMatrix' to external string
null: here is the first use of 'RawMutableMatrix'
```

### After
```m2
i1 : toExternalString mutableMatrix {{1, 2}, {3, 4}}

o1 = mutableMatrix(map(ZZ^2,ZZ^2,{{1, 2}, {3, 4}}))
```

### DIFF
--- a/M2/Macaulay2/m2/mutablemat.m2
+++ b/M2/Macaulay2/m2/mutablemat.m2
@@ -16,6 +16,7 @@ precision MutableMatrix := precision @@ ring
 expression MutableMatrix := m -> MatrixExpression append(applyTable(entries m, expression), symbol MutableMatrix => true)
 texMath MutableMatrix := m -> texMath expression m
 net MutableMatrix := m -> net expression m
+toExternalString MutableMatrix := lookup(toExternalString, MutableHashTable)
 
 map(Ring,RawMutableMatrix) := opts -> (R,m) -> (
      new MutableMatrix from {
--- a/M2/Macaulay2/m2/enginering.m2
+++ b/M2/Macaulay2/m2/enginering.m2
@@ -237,10 +237,12 @@ reduce := (r,s) -> (
 	  );
      (a,b))
 
+-- printing
 expression EngineRing := R -> if hasAttribute(R,ReverseDictionary) then expression getAttribute(R,ReverseDictionary) else expression toString R.RawRing -- should never be used
-texMath EngineRing := R -> texMath expression R
+
 toString EngineRing := toString @@ expression
-net EngineRing := net @@ expression
+net      EngineRing :=      net @@ expression
+texMath  EngineRing :=  texMath @@ expression
 
 ZZ _ EngineRing := 
 RR _ EngineRing :=
@@ -279,7 +281,6 @@ coefficientRing FractionField := F -> coefficientRing last F.baseRings
 	    dim FractionField := F -> 0
      expression FractionField := F -> if hasAttribute(F,ReverseDictionary) then expression getAttribute(F,ReverseDictionary) else (expression frac) (expression last F.baseRings)
      describe FractionField := F -> Describe (expression frac) (describe last F.baseRings)
-     toExternalString FractionField := F -> toString describe F
 
 -- freduce := (f) -> (numerator f)/(denominator f)
 
--- a/M2/Macaulay2/m2/galois.m2
+++ b/M2/Macaulay2/m2/galois.m2
@@ -7,9 +7,6 @@ needs "polyrings.m2"
 GaloisField = new Type of EngineRing
 GaloisField.synonym = "Galois field"
 
-toExternalString GaloisField := k -> toString describe k
-toString GaloisField := toString @@ expression
-net GaloisField := net @@ expression
 expression GaloisField := F -> if hasAttribute(F,ReverseDictionary) then expression getAttribute(F,ReverseDictionary) else (expression GF) (expression F.order)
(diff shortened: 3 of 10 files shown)


M2/Macaulay2/m2/mutablemat.m2
```diff
@@ -16,6 +16,9 @@ precision MutableMatrix := precision @@ ring
 expression MutableMatrix := m -> MatrixExpression append(applyTable(entries m, expression), symbol MutableMatrix => true)
 texMath MutableMatrix := m -> texMath expression m
 net MutableMatrix := m -> net expression m
+describe MutableMatrix := m -> Describe FunctionApplication(
+    mutableMatrix, describe matrix m)
+toExternalString MutableMatrix := toString @@ describe
```
mahrud, line 21: I think I'd rather have a method that checks whether a matrix is dense or sparse in memory, and if it's sparse describe and toExternalMatrix should do something different. Do you need this change now, or are you okay with keeping this open until we can expand it?

d-torrance: I'm in no rush -- I just noticed that the current behavior raised an error!

mahrud: Well, now that you point it out, maybe it _should_ raise an error, just a different error:
```m2
i1 : toExternalString new MutableList from {}
stdio:1:1:(3): error: anonymous mutable list cannot be converted to external string

i2 : toExternalString new MutableHashTable from {}
stdio:2:1:(3): error: anonymous mutable hash table cannot be converted to external string

i3 : toExternalString new MutableMatrix from id_(ZZ^2)
stdio:3:1:(3): error: can't convert local symbol or invisible global symbol 'RawMutableMatrix' to external string
null: here is the first use of 'RawMutableMatrix'

```

d-torrance: What's the rationale for the mutable list/hash table errors?

mahrud: I think the point of `toExternalString` is that it should satisfy `X === value toExternalString X`, which can never be the case for mutable objects.

d-torrance: Ok, that makes sense

d-torrance:
I've re-written this so that `toExternalString` for mutable hash tables, lists, and matrices all call the same function that uses `synonym` in the error message.

### Before
```m2
i1 : toExternalString new SelfInitializingType of HashTable
stdio:1:1:(3): error: anonymous type cannot be converted to external string

i2 : toExternalString mutableMatrix {{0}}
stdio:2:1:(3): error: can't convert local symbol or invisible global symbol 'RawMutableMatrix' to external string
null: here is the first use of 'RawMutableMatrix'
```

### After

```m2
i1 : toExternalString new SelfInitializingType of HashTable
stdio:1:1:(3): error: anonymous self initializing type cannot be converted to external string

i2 : toExternalString mutableMatrix {{0}}
stdio:2:1:(3): error: anonymous mutable matrix cannot be converted to external string
```

M2/Macaulay2/m2/mutablemat.m2
mahrud: Actually what I said can't be true because of these:
```m2
i16 : select(methods toExternalString, m -> ancestor(MutableHashTable, m#1))

o16 = {0 => (toExternalString, FractionField)   }
      {1 => (toExternalString, GaloisField)     }
      {2 => (toExternalString, Monoid)          }
      {4 => (toExternalString, PolynomialRing)  }
      {5 => (toExternalString, QuotientRing)    }
      {7 => (toExternalString, Variety)         }
```
But I think these are sort of the exceptions?

mahrud:
Looks good. Mind if I send in a couple of commits, clearing up unnecessary toExternalString methods for different types of rings?

d-torrance:
Go for it!

mahrud:
Now we have:
```m2
i1 : select(methods toExternalString, m -> ancestor(MutableHashTable, m#1))

o1 = {0 => (toExternalString, Monoid)          }
     {1 => (toExternalString, MutableHashTable)}
     {2 => (toExternalString, Ring)            }
     {3 => (toExternalString, Variety)         }
```
But among these, the mutable hash table one is just an error:
```m2
i3 : code o1#1

o3 = -- code for method: toExternalString(MutableHashTable)
     ../../Macaulay2/m2/nets.m2:73:39-76:43: --source code:
     toExternalString MutableList      := s -> (
         if hasAttribute(s,ReverseDictionary)
         then toString getAttribute(s,ReverseDictionary)
         else error("anonymous ", synonym class s,
```
And the other three are convenient exceptions because you can make a matrix, list, or hash table immutable, but you can't turn a monoid, ring, or variety immutable.

Tangentially:
1. typing `code 1` above gives the code for `(toExternalString, BasicList)`, because `select` didn't update the cached methods list. Maybe we should have the after print of `NumberedVerticalList` automatically update a unified variable "lastlist", which can be used by all `help`, `code`, etc?
2. the last line of `(toExternalString, MutableList)` is dropped for some reason.

mahrud (approved):
I'm happy with this. If the changes I added are okay, feel free to merge!

mahrud:
Related: have you seen #1026?

Issue #3354, opened 2024-07-10
Catching uncaught type errors


mahrud:
In the documentation node [Tutorial: Divisors](https://www.macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/___Tutorial_co_sp__Divisors.html), there's this code:
```m2
i54 : effective = (D) -> (
          LB := globalSections D;
          L := LB#0;  -- the matrix of numerators
          if numgens source L == 0 
          then error(toString D + " is not effective")
          else divisor sectionIdeal(L_(0,0), LB#1, D));
i55 : effective(2 R - P)

o55 = Divisor{ideal (z, x), ideal 1}

o55 : Divisor
```
which seems fine on the surface, until you realize that `toString D + " is not effective"` is an error, since `String + String` is not defined! Perhaps this used to be defined, but was removed at some point and because that branch in the code was never tested in the documentation, nobody realized that this is an error now.

I'm curious whether other languages have mechanisms to detect type errors like this. @DanGrayson was there ever a plan for the interpreter to catch errors like this using the `typicalValues` hash table or `pseudocode` or `disassemble`?

I understand that a user may define a function like this and define the missing methods _later_, but is there a way to check if every method call in a function is well-defined at a given point in time? e.g. `isWellDefined effective`?

For instance, `disassemble effective` yields a lisp like string and I think this section contains the error:
```
(adjacent (global-fetch 410) (2-OP + (adjacent (global-fetch 1067) (fetch 0 0)) " is not effective"))
```
I can guess what `adjacent` and `2-OP` mean and I presume `global-fetch 410` is `error` and `global-fetch 1067` is `toString`. Is there anyway to get the function corresponding to these numbers in the top level? If so, then from `typicalValues#toString` it would be apparent that the binary operator `+` is not defined for two strings.

This would be a very useful check for packages, and to make sure changes in the engine don't break untested code.

mahrud:
@pzinn since you've been looking at this code, do you have any insight on this?

mahrud:
But the type of the local variables should be determinable (at least in some cases, and maybe eventually in most/all cases), no?

pzinn:
how? M2 is not a typed language... except for arguments of methods I suppose?

mahrud:
Like this:
```m2
i2 : typicalValues#toString

o2 = String

o2 : Type
```

pzinn:
[...]

PR #3363, opened 2024-07-17
Position/locate/pseudocode revamp

pzinn:
- The positioning of code has been greatly improved, allowing for correct location/quoting code/error. compare say
```
x -> "blah
blah"
```
before and after. Various hacks that are no longer necessary have been correspondingly removed.
- The `Pseudocode` type has been revamped. It has been split into two (`Pseudocode` and `PseudocodeClosure`), has been renamed in d for consistency and also to avoid confusion (the old name `CodeClosure` didn't make it clear that the type only existed for debugging purposes, not for execution). `disassemble` has been retired: all pseudocode is automatically disassembled and displayed in a more human-friendly way.
- `tempPosition` has been introduced when exporting symbols in a package, so their position is only determined when the symbol is actually used.
- some code cleanup

This PR is the first part of a series involving debugging and error messages. Hopefully it is self-contained.


pzinn:
I hate when that happens.

pzinn:
weird, builds fine on my comp.

M2/Macaulay2/d/expr.d
```diff
@@ -278,12 +275,13 @@ export compiledFunctionClass := newtypeof(functionClass);
 export compiledFunctionClosureClass := newtypeof(functionClass);
 export symbolClass := newbasictype();
 export keywordClass := newtypeof(symbolClass);
-export codeClass := newbasictype();
+export pseudocodeClass := newbasictype();
+export pseudocodeClosureClass := newtypeof(pseudocodeClass);
```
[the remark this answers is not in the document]
pzinn: I see -- so I should leave `pseudocodeClass` where it is, but move `pseudocodeClassClosure` down? It's a bit unfortunate, but I suppose it doesn't really matter...

pzinn:
another issue, inevitably I'll have to fix the output of `ComputationsBook` tests to reflect the improved code location. I know I've asked before (@mahrud), but how do I populate the `ComputationsBook` tests when using cmake?

pzinn:
> another issue, inevitably I'll have to fix the output of `ComputationsBook` tests to reflect the improved code location. I know I've asked before (@mahrud), but how do I populate the `ComputationsBook` tests when using cmake?

aha, I think I found an old thread:

> not directly related to this PR: [@mahrud](https://github.com/mahrud) I'm struggling to fix a test, but I'm running blind because locally I can't get the ComputationsBook tests to run; when I do ctest, I get error messages about the .trim files not existing. I vaguely remember having had this problem, but don't remember how to fix it.
> 
> You have to run ninja M2-tests or something like that first. I'm not proud of it but couldn't figure out how to make it work with ctest alone ...
maybe this should be in the wiki

mahrud:
It's two steps:
```m2
ninja M2-tests-ComputationsBook -- generate output files
ctest -R  ComputationsBook -- compare against expected output files

File: M2/Macaulay2/d/actors.d [hunk 2]
```diff
@@ -1010,6 +1014,7 @@ installMethodFun2(arg1:Expr,args:CodeSequence):Expr := (
      else buildErrorPacket("expected left hand parameter to be a function, type, or a hash table"));
 installMethodFun(args:CodeSequence):Expr := installMethodFun2(eval(args.1),args);
 InstallMethodFun = installMethodFun;
+setup(ColonEqualS,InstallMethodFun);
```
mahrud, line 1017: Was this really not a symbol before and is being added now?!

M2/Macaulay2/d/actors2.dd
```diff
@@ -660,21 +660,162 @@ unSingleton(e:Expr):Expr := (
      else e);
 setupfun("unsequence",unSingleton);
 
+disassembleCode(c:Code):Expr := ( -- add locations?
```
mahrud, line 663: Since you're editing almost all of the code for `disassemble` and `pseudocode`, I would be very happy if you moved it out of `actors2.dd`, perhaps `debugging.dd` or something like that?

M2/Macaulay2/d/actors4.d
```diff
@@ -1586,123 +1587,30 @@ precision(e:Expr):Expr := (
 setupfun("precision0",precision);
 
 -- locate:
+locate(p:Position):Expr := (
```
mahrud, line 1590: Same here, could you move this to `debugging.dd` (or maybe `code.dd`)? Aligning the names with the filename in Core would be ideal, but at the very least not in a jungle of actors is great.

M2/Macaulay2/d/basic.d
```diff
@@ -18,7 +18,8 @@ export hash(e:Expr):hash_t := (
      is MysqlConnectionWrapper do hash_t(237489) -- improve this later!
      is MysqlFieldWrapper do hash_t(23748) -- improve this later!
      is MysqlResultWrapper do hash_t(2374) -- improve this later!
-     is CodeClosure do hash_t(73889)       -- improve this later!
+     is PseudocodeClosure do hash_t(73889)       -- improve this later!
```
mahrud, line 21: Maybe align the "-- improve this later!" comments? (unless you want to just remove them)

M2/Macaulay2/packages/Macaulay2Doc/debugging.m2
```diff
@@ -50,7 +50,7 @@ document {
      code f
      f 0
      return
-     disassemble current
+     current
```
mahrud, line 53: Did you mean `pseudocode current` here?

M2/Macaulay2/packages/Macaulay2Doc/repl.m2
```diff
@@ -969,57 +969,51 @@ document { Key => symbol OutputDictionary,
...
-	  "It may be useful to disassemble code during debugging, as in the following demonstration."
+	  "It may be useful to look at code during debugging, as in the following demonstration."
 	  },
      EXAMPLE lines ///
      load "Macaulay2Doc/demo1.m2"
      code g
      g 2
      code current
-     disassemble current
+     current
```
mahrud, line 1004: Same here! (Is this a duplicate?)

mahrud:
@d-torrance and I have talked a bit about commit message and granularity standards for Macaulay2. I think we both prefer more granular commits (the first commit could have been several bite sized ones), for instance because it's easier to figure out what change broke something using `git bisect`.

I don't necessarily want you to go back and split the commit, but on the other hand I'd like to push for better commit message standards, especially outside of personal packages:
``` Could you squash all your commits into one?

**Review by mahrud — CHANGES_REQUESTED** [review]
@d-torrance and I have talked a bit about commit message and granularity standards for Macaulay2. I think we both prefer more granular commits (the first commit could have been several bite sized ones), for instance because it's easier to figure out what change broke something using `git bisect`.

I don't necessarily want you to go back and split the commit, but on the other hand I'd like to push for better commit message standards, especially outside of personal packages:
```
Position/locate/pseudocode revamp
change compile order
and again
moved pseudocodeClosureClass def down, fixed test
fix test
```
Could you squash all your commits into one? I think that would be the easiest thing to do for now.

File: M2/Macaulay2/packages/Macaulay2Doc/debugging.m2 (hunk 6 above)
Comment (pzinn) [reply M2/Macaulay2/packages/Macaulay2Doc/debugging.m2:53]: no I think it's correct as is. current is already pseudocode, it's just that now there's no need to `disassemble` it.

File: M2/Macaulay2/packages/Macaulay2Doc/repl.m2 (hunk 7 above)
Comment (pzinn) [reply M2/Macaulay2/packages/Macaulay2Doc/repl.m2:1004]: yeah the code is a duplicate, but that was already the case before.

**Comment by pzinn** [comment]
> It's two steps:
> 
> ```m2
> ninja M2-tests-ComputationsBook -- generate output files
> ctest -R  ComputationsBook -- compare against expected output files
> ```
maybe include this in https://github.com/Macaulay2/M2/wiki/Building-M2-from-source-using-CMake

File: M2/Macaulay2/d/actors.d (hunk 2 above)
Comment (pzinn) [reply M2/Macaulay2/d/actors.d:1017]: try `pseudocode` on something like
```
() -> (
toList List := identity
)
```
(+disassembling before this PR).
Currently it doesn't have a symbol for `InstallMethodFun` so will display nothing. after this PR it will show the symbol `:=".

File: M2/Macaulay2/d/basic.d (hunk 5 above)
Comment (pzinn) [reply M2/Macaulay2/d/basic.d:21]: yeah I'm not sure what these comments are meant for

**Comment by pzinn** [comment]
OK I'll move things around and then force push into a single commit

File: M2/Macaulay2/d/actors2.dd (hunk 3 above)
Comment (pzinn) [reply M2/Macaulay2/d/actors2.dd:663]: what's the `.dd` stand for? (as opposed to `.d`)

Comment (mahrud) [reply M2/Macaulay2/d/actors2.dd]: `.d` is translated to `.c` programs, `.dd` to `.cc` (cpp).

File: M2/Macaulay2/c/error.c [hunk 8]
```diff
@@ -117,11 +117,11 @@ void fail(char *filename, int lineno) {
 void downpos(node n){
      struct POS *p = pos(n);
      if (p != NULL && p->filename != NULL) {
-	  fprintf(stderr,errfmt,p->filename,p->lineno,p->column+1,"");
+	  fprintf(stderr,errfmt,p->filename,p->lineno,p->column,"");
```
mahrud, line 120: What does this do again?

M2/Macaulay2/d/binding.d
```diff
@@ -766,13 +767,13 @@ export bind(e:ParseTree,dictionary:Dictionary):void := (
      when e
      is i:IfThen do (
 	  bind(i.predicate,dictionary);
-	  -- i.thenclause = bindnewdictionary(i.thenclause,dictionary);
-	  bind(i.thenclause,dictionary);
+	  -- i.thenClause = bindnewdictionary(i.thenClause,dictionary);
+	  bind(i.thenClause,dictionary);
```
mahrud, line 771: This had deeply annoyed me in the past ... thanks for fixing it!

M2/Macaulay2/c/error.c
pzinn: you mean the +1? it was part of the numerous hacks to circumvent the inaccuracy of code location. no longer needed.

mahrud (approved):

mahrud:
@d-torrance I already reviewed the changes here, but it involves the interpreter so probably two reviews would be more appropriate.

d-torrance:
Yes, I definitely plan on reviewing this (and #3382) soon.

d-torrance:
I'm hesitant about the changes to how pseudocode is displayed.  While the current lisp-like display can be hard to read, the parentheses are great for delineating each `Code` element.

For example, consider the following current behavior:
```m2
i1 : disassemble pseudocode(() -> if true then for x in {} do while true do try null)

o1 = (function restargs: false numparms: 0 framesize: 0 frameID: 8611 (if (global-fetch 218) then: (for in: (list ) from: (null) to: (null) when: (null) list:
     (null) do: (while (global-fetch 218) do: (try (global-fetch 220) (null) (null)))) else: (null)))
```

Versus the proposed behavior:

```m2
i1 : pseudocode(() -> if true then for x in {} do while true do try null)

o1 = function {restargs false                                           }
              {numparms 0                                               }
              {framesize 0                                              }
              {frameID 8894                                             }
              {body if {global-fetch 219                               }}
              {        {then for {framesize 1                         }}}
              {        {         {frameID 8895                        }}}
              {        {         {in list {}                          }}}
              {        {         {from                                }}}
              {        {         {to                                  }}}
              {        {         {when                                }}}
              {        {         {list                                }}}
              {        {         {do while {global-fetch 219         }}}}
              {        {         {         {do try {global-fetch 221}}}}}
              {        {         {         {       {then            }}}}}
              {        {         {         {       {else            }}}}}
              {        {else                                           }}

o1 : PseudocodeClosure
```
With the proposed behavior, it's not clear that `if`, `for`, `while`, and `try` are actually the starts of new `Code` objects, but there are opening parentheses in the current behavior that tell us that.

Also, in an editor like Emacs, we can move our cursor to a closing parenthesis and it will tell us where the corresponding opening parenthesis is with the current behavior.  But in the proposed behavior, there are a bunch of curly braces everywhere.

Maybe a compromise would be to keep the basic lisp style, but with proper indentation?  Maybe something like the following, which is just the current behavior as indented by Emacs in elisp mode:

```m2
(function restargs: false
	  numparms: 0
	  framesize: 0
	  frameID: 8611
	  (if (global-fetch 218)
	      then:
	    (for in: (list )
		 from: (null)
		 to: (null)
		 when: (null)
		 list: (null)
		 do: (while (global-fetch 218)
		       do: (try (global-fetch 220)
				(null)
				(null))))
	    else: (null)))
```



Also, maybe we could define `disassemble Pseudocode := net` or similar for backward compatibility?

mahrud:
Oh I didn't realize this is the display in Emacs. The html display is very different: https://github.com/Macaulay2/M2/issues/3354#issuecomment-2236329588

For some reason I thought the new output was of type Descent, or a similar traversable data structure, but if it's not then I agree with Doug that all the braces are counterproductive.

I'm wondering maybe the following would be the most useful:
- `pseudocode` should return a `PseudocodeClosure` object which can be turned back into a `FunctionClosure` using `value`
- `disassemble` should return a nested lisp-style sequence or even something like json.

The second item can happen later, but it's very useful for modern editors.

pzinn:
Yeah the display uses `VerticalList` which is ugly in standard mode. (the reason it looks very different -- and nicer -- in webapp mode is a separate story -- I did try to imitate the standard mode but failed).
I can easily make it look like you want @d-torrance but do you really want all the parentheses?? and turn pseudocode into lisp...
as to all your suggestions to reinstate `disassemble`, that kinda goes again the philosophy of this PR. Also I doubt we need backward compatibility, are there users who ever used this stuff?

pzinn:
also do you want the "null"s displayed? this is something that M2 is not very consistent about across situations and modes...

pzinn:
also, I object to all the unncessary colons.

pzinn:
we could use `toString` to produce the old, ugly, one-line with lots of parentheses, output...

mahrud:
> also, I object to all the unncessary colons.

I don't know about Doug but I don't care about the colons.

> and turn pseudocode into lisp...

I think this is the most important part, because lisp is basically an abstract syntax tree, so _some_ way of traversing the tree (either by following parenthesis in a string like the old output, or following a nested sequence, or a Descent-like structure) is necessary.

> Yeah the display uses `VerticalList` which is ugly in standard mode.

Can you have `disassemble` just return the list? Keywords like `if, else, ...` can be either strings or symbols, like this:
```m2
("try", ("global-fetch", 220), null, null)
```

pzinn:
yeah `disassemble` just rerturns a nested list, that's why it's easy to tweak the output at the M2 level any way one likes.

mahrud:
Didn't you say it was retired? I'm confused.

pzinn:
> Ooh, I like the idea of returning an object like a Descent that we could programatically traverse! That's a lot better than a > lisp-like string, I think. This could be the new behavior of disassemble, I suppose?
I'm confused, this is already the new behavior, cf the 2nd example on https://www.unimelb-macaulay2.cloud.edu.au/usr/share/doc/Macaulay2/Macaulay2Doc/html/_pseudocode.html?user=221974

pzinn:
all I can say is the obvious which is that `Pseudocode` is *not* a list -- it's ... pseudo-code.
so things like `toSequence` or `_` won't work out of the box, you need to program them.  I used `_` instead of `#` because `#` can't be programmed at the M2 level.

pzinn:
attempted compromise:
```
i1 : pseudocode(() -> if true then for x in {} do while true do try null)

o1 = function restargs false
              numparms 0
              framesize 0
              frameID 8944
              body if global-fetch 219
                      then for framesize 1
                               frameID 8945
                               in list 
                               from null
                               to null
                               when null
                               list null
                               do while global-fetch 219
                                        do try global-fetch 221
                                               then null
                                               else null
                      else null

o1 : PseudocodeClosure

i2 : toString oo

o2 = function (restargs false numparms 0 framesize 0 frameID 8944 body if (global-fetch 219 then for (framesize 1 frameID 8945 in list () from null to null when null
     list null do while (global-fetch 219 do try (global-fetch 221 then null else null))) else null))
```

d-torrance:
Ah, cool -- I didn't notice the `_` behavior earlier.  With the `net` changes, it looks good to me now.  :)

d-torrance (approved):

mahrud:
Did you want a parenthesis after global-fetch? This doesn't look right:
```
try (global-fetch 221 then null else null)
```

M2/Macaulay2/m2/code.m2
```diff
@@ -271,6 +262,33 @@ debuggerHook = entering -> (
 	  )
      )
 
+-- disassemble
+CodeList := new Type of List
```
mahrud, line 266: Oooh you've opened a can of worms (to be dealt with in future, not this PR) .. now I want `code` to return a `CodeList` with numbered corresponding to the line of code!

M2/Macaulay2/m2/exports.m2
```diff
@@ -630,7 +631,6 @@ export {
 	"difference",
 	"dim",
 	"directSum",
-	"disassemble",
```
mahrud, line 633: I think you should keep this exported, especially since it's an easier way to traverse Pseudocode.

Also, is there a reason why `disassemble` returns mostly sequences, but one list near the beginning, and doesn't disassemble all the way? See:
```m2

i2 : X = disassemble pseudocode(() -> try true);

i3 : class \ X

o3 = (String, List)

o3 : Sequence

i4 : class \ X#1#4

o4 = (String, Pseudocode)

o4 : Sequence

i5 : disassemble X#1#4
stdio:5:11:(3): error: expected pseudocode

i6 : disassemble X#1#4#1

o6 = (try, {global-fetch 219, (then, null), (else, null)})

o6 : Sequence
```

mahrud (dismissed):

M2/Macaulay2/m2/exports.m2
pzinn: it doesn't disassemble all the way because that's how it works. `Pseudocode` is not a list, so if you want to see one of its parts, you have to disassemble only at one level.
I don't think `disassemble` should be exported, it's only meant to be used internally.
`disassemble` returns a sequence which is simply a list of objects to be printed next to each other.

pzinn:
> Did you want a parenthesis before try and one after global-fetch? This doesn't look right:
> 
> ```
> try (global-fetch 221 then null else null)
> ```
> 
> compare with:
> 
> ```
> (try (global-fetch 221) then null else null)
> ```

I disagree. the parentheses follow the internal structure. `tryCode` contains a bunch of arguments which are listed in `(...)`.

pzinn:
the last commit introduces `Pseudocode_*`, which I think is what @mahrud wants by exporting `disassemble` (which is not the way to go).

M2/Macaulay2/m2/exports.m2
mahrud: What is the rationale behind these decisions? To me they seem arbitrary. `disassemble` used to print the S-expression (aka "lisp version") of the code for the user, and your proposed `Pseudocode` type is pretty but doesn't replace that.

mahrud:
> I disagree. the parentheses follow the internal structure. `tryCode` contains a bunch of arguments which are listed in `(...)`.

If you're suggesting that the output should be converted from [S-expressions](https://en.wikipedia.org/wiki/S-expression) to [M-expressions](https://en.wikipedia.org/wiki/M-expression), then maybe you should just use `code` not `pseudocode`. This string display is not nearly as useful for debugging:
```m2
ii6 : current

oo6 = 2-OP / 1
             fetch 0 0

oo6 : PseudocodeClosure

ii7 : toString current

oo7 = 2-OP / (1 fetch 0 0)
```
This used to be `(2-OP / 1 (fetch 0 0))`, which is an S-expression.

M2/Macaulay2/m2/exports.m2
pzinn: `toString` will produce the same lisp version as what `disassemble` used to do, with perhaps a few useless parentheses removed. The rest will be a string just like `disassemble` used to produce, which means fairly useless per se (why should one have to parse a string?). I believe the current format is much more structured. I'm sure one can it be improved, but the principle seems sound to me.

pzinn:
ugh... this is ugly as hell but sure one can use S-expressions...

pzinn:
actually, correction, there does seem to be something wrong with the parenthesising. investigating.

mahrud:
I mean, this is how almost any interpreter works! That's the whole point, and in some languages you can explicitly view this:
```m2
julia> Meta.show_sexpr(Meta.parse("4 + 4 / 2"))
(:call, :+, 4, (:call, :/, 4, 2))
```

M2/Macaulay2/m2/exports.m2
mahrud: This is _not lisp_:
```m2
i3 : toString pseudocode(() -> (4 + 4) / 2)

o3 = function (restargs false numparms 0 framesize 0 frameID 11325 body 2-OP / (2-OP + (4 4) 2))
```
This is:
```m2
i1 : disassemble pseudocode(() -> (4 + 4) / 2)

o1 = (function restargs: false numparms: 0 framesize: 0 frameID: 8748 (2-OP / (2-OP + 4 4) 2))
```
One pair of parentheses are replaced by "body", but you've also moved every other parenthesis. You seem to prefer the `Pseudocode` display, so why do you insist on changing the lisp format?

I agreed with you about removing the colons after "then" and "else" but you also removed the colons after "restargs", etc. which definitely make this output look _less_ structured.

pzinn:
None of these changes have anything to do with how the interpreter works. In any case, I've changed the parentheses which should match more closely the required form.

mahrud:
I think the parentheses/braces between toString and disassemble are off still. Can you explain why you prefer disassemble to produce a sequence possibly containing lists which possibly contain another `Pseudocode`? What's the difference between `CodeList` and `CodeSequence`?
```m2
i1 : debug Core; load "Macaulay2Doc/demo1.m2"; g 2
../../Macaulay2/packages/Macaulay2Doc/demo1.m2:8:11:(3):[2]: error: division by zero
../../Macaulay2/packages/Macaulay2Doc/demo1.m2:8:11:(3):[2]: --entering debugger (type help to see debugger commands)
../../Macaulay2/packages/Macaulay2Doc/demo1.m2:8:10-8:13: --source code:
     b := 1/x;

ii4 : current

oo4 = 2-OP / 1
             fetch 0 0

oo4 : PseudocodeClosure

ii5 : toString current

oo5 = (2-OP / 1 (fetch 0 0))

ii6 : disassemble current

oo6 = (2-OP, /, {1, fetch 0 0})

oo6 : Sequence

ii11 : class \ flatten oo6

oo11 = (String, SymbolBody, Pseudocode, Pseudocode)

oo11 : Sequence
```
You said it's for internal use, but I don't understand what changes if everything was nested sequences without any `Pseudocode`. Colloquially if someone asked me for to disassemble a compiled binary code I wouldn't show them a file with some C written in there, it would be all assembly.

> None of these changes have anything to do with how the interpreter works.

I know you're not changing how the interpreter works. The point of `disassemble` was to translate a piece of code to the lowest level that the interpreter understands, which for M2 is lisp-like. It's not meant to be particularly human-readable or efficient, but unambiguous and simple to parse. Introducing new human-readable format for `Pseudocode` is useful and appreciated, but it's not the same thing.

For that reason, I think `disassemble` should be exported and we should still show its output in the documentation in addition to the `pseudocode` output, because different people (e.g. you and I) prefer different outputs and there's no reason to hide one in Core. If that's too much work, I can also export and document it myself after this is merged.

pzinn:
`CodeSequence` and `CodeList` are just for output purposes, they have no particular meaning. Clearly you have a different idea than I what the "purpose" of `disassemble` is. If so, perhaps you should write your own disassemble routine.

mahrud:
Yes, that's true, but the idea and purpose I have in mind for `disassemble` is the existing one, as documented [here](https://macaulay2.com/doc/Macaulay2-1.24.05/share/doc/Macaulay2/Macaulay2Doc/html/_disassemble.html), which this PR takes away. Maybe the internal-only tool you call `disassemble` should have a different name and you should define `disassemble Pseudocode := toString`, and leave its documentation intact? Why do you say this is "not the way to go"?

pzinn:
Fine, I can rename `disassemble` something else and have `disassemble` point to `toString`. For the rest, if you have a precise example where you think the parenthesising can be improved, let me know.

M2/Macaulay2/m2/exports.m2
pzinn: the problem with colons is, one can put them everywhere or remove them everywhere but anything in between is fairly arbitrary.

pzinn:
instead of `disassemble`, `toList` is used to turn `Pseudocode` into a list that's amenable to printing / using `_`. `disassemble` is exported and acts as `toString` on `Pseudocode`.

M2/Macaulay2/d/actors2.dd
```diff
@@ -660,24 +645,6 @@ unSingleton(e:Expr):Expr := (
...
 
-disassemble(e:Expr):Expr := (
-     when e
-     is f:FunctionClosure do toExpr(tostring(Code(f.model)))
-     is f:functionCode do toExpr(tostring(Code(f)))
-     is c:CodeClosure do toExpr(tostring(c.code))
-     is s:SpecialExpr do disassemble(s.e)
-     else WrongArg("pseudocode or a function closure derived from Macaulay 2 code")
-     );
-setupfun("disassemble", disassemble);
```
d-torrance, line 671: With the new `disassemble` behavior, we've lost some of the previous functionality, like accepting `Function` and `FunctionBody` objects as input.

d-torrance:
I think I actually kind of miss the colons...

Previously, open parentheses indicated the start of a new pseudocode object:

```m2
i1 : disassemble(() -> if true then 5)

o1 = (function restargs: false numparms: 0 framesize: 0 frameID: 8610 (if
     (global-fetch 218) then: 5 else: (null)))
```

But now, there are lots more opening parentheses:
```m2
i1 : disassemble pseudocode(() -> if true then 5)

o1 = (function (restargs false) (numparms 0) (framesize 0) (frameID 10541) (body
     (if (global-fetch 219) (then 5) (else null))))
```

It looks like maybe `restargs`, `numparams`, etc, might indicating new pseudocode objects, but they're really just members of the function struct.

Here are a couple possible lisp-y alternatives:

* Dotted pair notation:
  ```lisp
  (function (restargs . false) (numparms . 0) (framesize . 0) (frameID . 10541) (body .
  (if (global-fetch 219) (then . 5) (else . null))))
  ```
* Leading colons, like lisp keyword symbols:
  ```lisp
  (function :restargs false :numparms 0 :framesize 0 :frameID 10541 :body
  (if (global-fetch 219) :then 5 :else null)))
  ```

d-torrance:
Also, what's the difference between `Pseudocode` and `PseudocodeClosure`?

M2/Macaulay2/d/actors2.dd
pzinn: put back those methods

pzinn:
You guys make up your mind which notation you want for `toString`, I don't particularly care. For the record the documentation says nothing about which particular form of parenthesisation `disassemble` should produce.

pzinn:
it used to be that `Pseudocode` was the m2 equivalent of the d class `CodeClosure`. That was not very consistent.
Now `CodeClosure` -> `PseudocodeClosure` and `Code` -> `Pseudocode`.

mahrud:
Paul, I'm mostly happy with this now, but the PR is making a lot of changes in d (in particular many changes in the first commit), and I'm nervous about finding a bug in the future and the difficulty of figuring out which changes caused the bug. Do you mind if I make a new PR where I take exactly your changes, but break them up into smaller commits? (All commits still keeping you as the author.) This would make bisecting easier. I can also fix the merge conflicts.

Doug, I hear your point, but I think we can come back and fix the string printing later. Does that sound okay?

mahrud:
One thing i hadn't thought of before: moving pseudocode and disassemble out to code.m2 means that if there was a problem earlier, say in startup.m2, or in any of the files that need to be loaded before code.m2, we can't use pseudocode/disassemble to debug it. I think at least one of them needs to be in the interpreter.

d-torrance:
> Doug, I hear your point, but I think we can come back and fix the string printing later. Does that sound okay?

Oh totally -- I'm just splitting hairs.

mahrud:
> it used to be that `Pseudocode` was the m2 equivalent of the d class `CodeClosure`. That was not very consistent. Now `CodeClosure` -> `PseudocodeClosure` and `Code` -> `Pseudocode`.

Should `Pseudocode` really be `PseudocodeBody`?
```m2
i1 : pseudocode(() -> 1/0)

o1 = function restargs false
              numparms 0
              framesize 0
              frameID 11336
              body 2-OP / 1
                          0

o1 : PseudocodeClosure

i2 : pseudocode functionBody(() -> 1/0)

o2 = 2-OP / 1
            0

o2 : Pseudocode
```

pzinn:
the correspondence between d classes and m2 classes is not particularly consistent. `FunctionBody` in m2 is just `functionCode` in d (also, annoying lack of  capitalisation).  so, yes, one could have `PseudocodeBody`, and then both it and `PseudocodeClosure` could descent from a `Pseudcode` which would be the pseudocode analogue of `Function`. I decided to keep things simpler but it's debatable what is better.

DanGrayson:
Could you resolve the conflicts?

pzinn:
I think @mahrud rewrote this PR -- I will check out the "split" version soon. closing.

PR #3375, opened 2024-07-22
Added try .. then .. clause

### LINKED ISSUE #2572: try..then without else clause
Currently this doesn't work:
```m2
i1 : try true then 2
stdio:1:1:(3): error: syntax error : expected 'else' to match this 'try'
```
I would have expected it to be treated like this:
```m2
i2 : try true then 2 else null

o2 = 2

i3 : try error() then 2 else null

i4 : 
```
Similar to how `if..then..else` works:
```m2
i1 : if true then 2

o1 = 2

i2 : if true then 2 else null

o2 = 2

i3 : if false then 2 else null

i4 : 
```

mahrud:
Closes #2572.

### DIFF
--- a/M2/Macaulay2/d/binding.d
+++ b/M2/Macaulay2/d/binding.d
@@ -889,6 +889,10 @@ export bind(e:ParseTree,dictionary:Dictionary):void := (
 	  -- i.alternate = bindnewdictionary(i.alternate,dictionary);
 	  bind(i.alternate,dictionary);
 	  )
+     is i:TryThen do (
+	  bind(i.primary,dictionary);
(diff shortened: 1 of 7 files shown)


pzinn:
note that there will probably be a merge conflict with #3363 due to changes to `convertr.d`. though it should be easy to fix.

mahrud:
Good point, I'll draft it until yours is merged.

d-torrance (approved):

mahrud:
I think the other PR might take some time, so undrafting this. Resolving the conflicts this way shouldn't be too hard either.

PR #3370, opened 2024-07-21
Set warnings to be always printed

mahrud:
Before:
```m2
$ ninja install-Msolve
[1/1] Installing package Msolve
 -- making example results for "msolveEliminate"                             -- 1.79479 seconds elapsed
 -- making example results for "msolveSaturate"                              -- 1.6628 seconds elapsed
 -- making example results for "msolveRealSolutions"                         -- 2.01426 seconds elapsed
 -- making example results for "msolveLeadMonomials"                         -- 1.76609 seconds elapsed
 -- making example results for "msolveRUR"                                   -- 1.79487 seconds elapsed
 -- making example results for "msolveGB"                                    -- 1.7215 seconds elapsed
 -- warning: found 6 documentation node(s) not listed as a subnode
```
After:
```m2
$ ninja install-Msolve
[1/1] Installing package Msolve
 -- warning: document: documentation headlines must be less than 100 characters long:
 --   "Macaulay2 interface for msolve; computes real solutions, Groebner basis (in GRevLex), saturations, and elimination ideals."
 -- warning: document: documentation headlines must be less than 100 characters long:
 --   "Uses symbolic methods to compute a certified solution interval for each real solution to a zero dimensional polynomial system."
 -- warning: document: documentation headlines must be less than 100 characters long:
 --   "Uses symbolic methods to compute the rational univariate representation of a zero dimensional polynomial system."
 -- warning: document: documentation headlines must be less than 100 characters long:
 --   "Computes a Groebner basis for the saturation of an ideal by a single polynomial in GrevLex order; only works over a finite field."
 -- warning: found 6 documentation node(s) not listed as a subnode
 -- warning: ("MaybeQuotedTT{msolveEliminate(Ideal,List)}", is an instance of an intermediate markup type ,"MaybeQuotedTT", with no qname, appearing in hypertext during validation)
 -- warning: ("MaybeQuotedTT{msolveEliminate(Ideal,List)}", is an instance of an intermediate markup type ,"MaybeQuotedTT", with no qname, appearing in hypertext during validation)
 -- warning: ("MaybeQuotedTT{msolveEliminate(Ideal,RingElement)}", is an instance of an intermediate markup type ,"MaybeQuotedTT", with no qname, appearing in hypertext during validation)
 -- warning: ("MaybeQuotedTT{msolveEliminate(Ideal,RingElement)}", is an instance of an intermediate markup type ,"MaybeQuotedTT", with no qname, appearing in hypertext during validation)
[...]

PR #3376, opened 2024-07-22
Use lookup in validate for getting qname

d-torrance:
This allows us to use inheritance for getting the qname of hypertext types (e.g, `MaybeQuotedTT` is a child of `TT` and should inherit its qname).

This fixes all the `MaybeQuotedTT` warnings that are cropping up in #3370.  (@mahrud - would you rather I push this to that branch?)

We also fix a "no method" error that was occurring when we encountered this "no qname" warning with a positive `debugLevel`.

### DIFF
--- a/M2/Macaulay2/m2/format.m2
+++ b/M2/Macaulay2/m2/format.m2
@@ -179,11 +179,12 @@ net  PRE  :=
 net   TT  :=
 net CODE  :=
 net SAMP  :=
-net  KBD  :=
 info TT   :=
 info SAMP :=
-info  KBD :=
 info CODE :=  x -> horizontalJoin apply(noopts x,net)
+
+net  KBD :=
+info KBD := x -> formatNoEscaping toString horizontalJoin apply(noopts x,net)
 
 info PRE  := x ->
     wrap(printWidth, "-", concatenate apply(noopts x,toString @@ info))
--- a/M2/Macaulay2/m2/help.m2
+++ b/M2/Macaulay2/m2/help.m2
@@ -64,8 +64,8 @@ optTO := key -> (
 	-- TODO: figure out how to align the lists using padding
 	-- ref = pad(ref, printWidth // 4);
 	(format ptag, fkey, next(), fixup (
-		if currentHelpTag === ptag then MaybeQuotedTT fkey
-		else SPAN {MaybeQuotedTT fkey, " -- see ", TOH{ptag}})))
+		if currentHelpTag === ptag then KBD fkey
+		else SPAN {KBD fkey, " -- see ", TOH{ptag}})))
     -- need an alternative here for secondary tags such as (export,Symbol)
     else (fkey, fkey, next(), TOH{tag}))
 -- this isn't different yet, work on it!
--- a/M2/Macaulay2/m2/help.m2
+++ b/M2/Macaulay2/m2/help.m2
@@ -46,12 +46,6 @@ seeAbout := (f, i) -> (
 --   the last member is the corresponding hypertext entry in the UL list
 -----------------------------------------------------------------------------
 
--- we want quotes around elements of the "Ways to use" list so that
--- "* String" works when running "help" inside Macaulay2, but we don't
--- need the quotes otherwise
-MaybeQuotedTT = new IntermediateMarkUpType of TT
-net MaybeQuotedTT := x -> formatNoEscaping x#0
-
 counter := 0
 next := () -> counter = counter + 1
 optTO := key -> (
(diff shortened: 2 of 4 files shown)


mahrud:
Now that I look at it I have more problems with `MaybeQuotedTT`:
1. hypertext types shouldn't be defined outside hypertext.m2
2. `<tt>` is already deprecated in html5, so adding a subtype seems strange

I think we should be slowly removing TT and using it's replacements instead:
> Consider the [`<kbd>`](https://www.w3schools.com/tags/tag_kbd.asp) element (for keyboard input), the [`<var>`](https://www.w3schools.com/tags/tag_var.asp) element (for variables), the [`<code>`](https://www.w3schools.com/tags/tag_code.asp) element (for computer code), the [`<samp>`](https://www.w3schools.com/tags/tag_samp.asp) element (for computer output), or use CSS instead.

d-torrance:
Good point regarding `<tt>`.  I switched `MaybeQuotedTT` to `MaybeQuotedM2CODE`, which will also give us syntax highlighting in the HTML "ways to use" list.  I also switched `M2CODE` from a method to a subclass of `CODE`.

mahrud:
I think my preference is to keep the hypertext types as close to html elements as possible. So let's go back to `M2CODE` being a method. Why not have KBD (for `<kbd>`) be always syntax highlighted and use that instead? Afterall, if someone types on a keyboard in emacs it'll be syntax highlighted.

d-torrance:
Currently, the css for KBD elements gives us something that looks like a literal keyboard key, much like GitHub's css, e.g., <kbd>Esc</kbd>.

mahrud:
That sounds great, doesn't it? It would look like this:

* <kbd>new Type of HashTable</kbd>
* <kbd>new Type of BasicList</kbd>

@pzinn could even turn them clickable.

d-torrance:
Right, it already looks like that! https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Text/html/___K__B__D.html

pzinn:
currently, on Macaulay2Web, what makes something clickable is that it's code (preferably with language set to be "Macaulay2"). so `KBD` is not clickable:
https://www.unimelb-macaulay2.cloud.edu.au/usr/share/doc/Macaulay2/Text/html/___K__B__D.html

mahrud:
Right now the css needs a little adjusting to get them not to intersect:
image
But I would be in favor of this:
image

d-torrance:
I've removed the `M2CODE` stuff, switched to `KBD` for the "ways to use" list, and updated the `KBD` css.  I'll let someone more artistic fiddle with it if they'd like, but adjacent `KBD` elements in a list now no longer collide:

image

The `qname` change now doesn't really matter since `MaybeQuotedTT` is gone, but I think it's still a good change if we ever add any other children of existing `Hypertext` classes.

pzinn:
What's the point of this change? As mentioned above, it used to be that only <code> was considered code, and in particular clickable in the web interface. Is <kbd> supposed to replace it somehow?

mahrud:
There are two separate things going on:
1. `<tt>` is deprecated from html5, replaced by code, kbd, and var
2. we currently have a non-standard `MaybeQuotedTT` type which turns into `"blah"` in terminal but just `blah` on the web:
```m2
i16 : net MaybeQuotedTT "blah"

o16 = "blah"

i17 : html MaybeQuotedTT "blah"

o17 = <span class="tt">blah</span>
```
`<kbd>` is replacing `MaybeQuotedTT`, not `<code>`. This solves both problems in one in a sensible, non-hacky way, and it actually looks good.

pzinn:
oh okay I misunderstood. I thought `M2CODE` was completely removed.

mahrud:
No, at some point `M2CODE` was turned from a method into a subclass of `CODE`, but no longer. There are a few other "fake" hypertext types that I wouldn't mind getting rid of, but no need to worry about that here.

mahrud:
I changed the css a little, here's what it looks like now.
image
Probably all should be `<kbd>` types. Here's the net output:
```
       * "quotient(Ideal,Ideal)"
       * "quotient(Ideal,RingElement)"
       * "quotient(Module,Ideal)"
       * "quotient(Module,Module)" -- ideal or submodule quotient
       * "quotient(Module,RingElement)"
       * "quotient(MonomialIdeal,RingElement)"
```

pzinn:
to clarify: should `kbd` content be syntax-hilighted? for example, on Macaulay2Web, at the moment, only `code.language-macaulay2` is.

mahrud:
I don't have a strong preference either way. I would, however, prefer everything under "Exports" to be `kbd` not just the secondary tags. The simple way to do this is to have `TO` always use `kbd` but I don't know how to prevent the nodes with string keys (e.g. "heft vectors") from being turned to `kbd`.

pzinn:
I agree that for consistency, it would be nice to have all exports in the same style, say using `kbd`.

PR #3379, opened 2024-07-25
Undocument a bunch of expression methods, cf #1668

pzinn:
hopefully this reduces the number of undocumented methods, cf #1668

### DIFF
--- a/M2/Macaulay2/m2/expressions.m2
+++ b/M2/Macaulay2/m2/expressions.m2
@@ -14,10 +14,12 @@ lprec = prec = x -> (getParsing x)#0
 rprec = strength2 = x -> (getParsing x)#1
 uprec = strength1 = x -> (getParsing x)#2
 
--- local variables
-EmptyName := symbol EmptyName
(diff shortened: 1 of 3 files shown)


mahrud:
Ah, sorry I should have been clear that I'm already working on this on a branch, I just wanted to know if you have a particular preference about how to do it.

For instance, do you specifically want things like `Sum + Sum` to appear in the documentation (perhaps along with the documentation of +), or is it okay if we undocument all default operations on Expressions other than some exceptions? (e.g. `denominator Divide`, etc.

pzinn:
well, you can see my proposed solution, that should answer your question

mahrud:
It doesn't really answer this question:

> For instance, do you specifically want things like `Sum + Sum` to appear in the documentation (perhaps along with the documentation of +), or is it okay if we undocument all default operations on Expressions other than some exceptions? (e.g. `denominator Divide`, etc.)

Which nodes involving expressions do you think should be listed in the documentation?

PR #3381, opened 2024-07-27
Removed dummy definitions from typicalvalues.m2

mahrud:
- **merged tvalues-make.m2 into typicalvalues.m2**
- **simplified generateTypicalValues**
- **set M2 to autogenerate tvalues.m2**
- **removed --no-tvalues optional argument**
- **added debug info to tvalues.m2**
- **removed a bunch of dummy definitions in typicalvalues.m2**
- **removed a bunch more dummy definitions**
- **removed 'any' from typicalvalues.m2**
- **moved a few compiled functions to classes.m2**
- **removed some dummy Net and Set methods**
- **wip: removed more dummy methods from typicalvalues.m2**
- **wip: separate inputs and outputs in tvals**
- **tabulated the tvalues.m2 file**
- **switched to autogenerating tvalues.m2**


mahrud:
How on earth did this happen?!
```m2
 -- i16 : assert ( hash res =!= hash syz )
 -- stdio:25:1:(3): error: assertion failed
```

M2/Macaulay2/m2/typicalvalues.m2
```diff
@@ -199,6 +201,8 @@ generateTypicalValues = (srcdir) -> (
 
 if isMember("--no-tvalues", commandLine) then end
 
+if not fileExists typicalValuesSource then generateTypicalValues(currentFileDirectory | "../d/")
```
d-torrance, line 128: I don't know how likely this situation would be, but suppose that `tvalues.m2` doesn't exist and we're reading this file from someplace other than the source repository so that `../d` doesn't exist.  Should the user get a more informative error message that the one they'll get from `readDirectory`?

M2/Macaulay2/m2/typicalvalues.m2
```diff
@@ -199,8 +200,6 @@ generateTypicalValues = (srcdir) -> (
 	if 0 < #extracted then outfile << comment << endl << stack extracted << endl);
     close outfile)
 
-if isMember("--no-tvalues", commandLine) then end
```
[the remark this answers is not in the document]
mahrud: I _think_ this wouldn't happen anymore, but more generally this option is _never_ useful for non-developers, and for us, we can just comment out the `load typicalValuesSource` line to debug. Does this seem reasonable? Alternatively, we could use an environment flag, so we don't have to keep the option in `M2 --help`.

M2/Macaulay2/m2/typicalvalues.m2
mahrud: I think this is the error you'd get:
```m2
../../Macaulay2/m2/debugging.m2:18:6:(1):[12]: error: unable to find the source code for the interpreter
```
(not sure why it says it's coming from `debugging.m2` ... I find this so annoying.)

M2/Macaulay2/m2/typicalvalues.m2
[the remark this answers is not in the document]
d-torrance: Good point -- that works

M2/Macaulay2/m2/typicalvalues.m2
d-torrance: Oh, I missed that check! It was added it a commit I hadn't looked at yet

mahrud: No worries. Also, if you or @pzinn have any clues about why the error points to debugging.m2 rather than the real source, I'd be very curious.

d-torrance: No clue, but this has always bugged me too. Sometimes I've found that if run the code that raised the error a second or third time, I'll eventually get a useful backtrace with the actual source of the error. It's very weird!

mahrud: Maybe once this PR is merged I'll make an issue with this isolated example which seems to be pretty reproducible.

M2/Macaulay2/m2/set.m2
```diff
@@ -81,9 +81,16 @@ Set.synonym = "set"
...
 -- constructor
 new Set from List := Set => (X,x) -> set x -- compiled function
-elements Set := List => x -> keys x
+
+-- compiled functions which will be documented
+-- TODO: why doesn't this work?
+-- # Set        := Boolean => lookup(symbol #,  HashTable)
+-- Set #? Thing := Boolean => lookup(symbol #?, HashTable, Thing)
+installMethod(symbol #,  Set,        Boolean => x -> x)
+installMethod(symbol #?, Set, Thing, Boolean => x -> x)
```
d-torrance, line 95: Should these installed methods raise an error like the `dummy` ones in `typicalvalues.m2`?

M2/Macaulay2/m2/typicalvalues.m2
```diff
@@ -15,21 +15,10 @@ typicalValues#(symbol symbol) = Symbol
 typicalValues#(symbol ?) = Boolean
 
 dummy := x -> error("dummy method function called")
-installMethod(symbol #?, HashTable, Thing, Boolean => x -> (dummy x;))
-installMethod(symbol #?, Set, Thing, Boolean => x -> (dummy x;))
-installMethod(symbol #?, Database, String, Boolean => x -> (dummy x;))
+-- TODO: where are these defined in the interpreter?
```
d-torrance, line 16: I think these `BasicList` methods are the same as the corresponding `List` ones:

* `BasicList #? ZZ`: `subvalueQ` in hashtables.dd
* `BasicList # ZZ`: `subvalue` in hashtables.dd
* `# BasicList`: `lengthFun` in actors.d

d-torrance:
General question: what determines whether a "method" that's really a compiled function should be declared using a comment in the d directory v. in typicalvalues.m2?

mahrud:
> General question: what determines whether a "method" that's really a compiled function should be declared using a comment in the d directory v. in typicalvalues.m2?

Whether I've got around to adding it in the d directory or not ;)
I don't like files that contain just piles of unrelated things, so I'd like to remove all of them from typicalvalues.m2.

M2/Macaulay2/m2/set.m2
mahrud: I don't follow. Raise an error for what?

M2/Macaulay2/m2/typicalvalues.m2
mahrud: So is `List` in the interpreter the same as `BasicList` in the top-level?

d-torrance: Kind of but not exactly, since `Sequence` also inherits from `BasicList` at top-level, but `List` and `Sequence` are distinct in the interpreter.  (A `Sequence` is just a sequence of `Expr` objects, and a `List` contains a `Sequence` and also keeps track of its class, its hash code, and whether it's mutable.)

So if a compiled function gets an instance of some subclass of `BasicList` as input, there are two possible branches it might end up in inside the `when` clause -- `Sequence` if it's a sequence and `List` for everything else.

M2/Macaulay2/m2/set.m2
d-torrance: I was just suggesting that maybe for consistency, the installed method should be `x -> (dummy x;)`, where `dummy` is defined as `x -> error("dummy method function called")` like the other ones in typicalvalues.m2.

Why add these here rather than a comment in the d directory?  And do we even need to bother since `Set` inherits from `HashTable`, and it's already registered using a comment in actors.d?

Also, wouldn't the typical value for `# Set` be `ZZ` instead of `Boolean`?

M2/Macaulay2/m2/undocs.m2
```diff
@@ -0,0 +1,327 @@
+restart
```
d-torrance, line 1: Was this file intended to be commited to git?

mahrud: Certainly not haha

PR #3382, opened 2024-07-27
Various updates to methods / installPackage / package / etc.

mahrud:
- **set warnings to be always printed**
- **added mutable symbols to items in package docs**
- **added briefDocumentation for method keys**
- **added routine that checks everything is documented**
- **overhauled warning and errors in installPackage**
- **fixed an issue in (package, Symbol), (package, Sequence)**
- **changed default errorDepth in CMake targets**
- **fixed methods to not show unexported methods**
- **simplified importFrom and added warning**
- **deleted unused key in Package**
- **added missing documentation for MonomialIdeal:RingElement**
- **moved and fixed (sort, List, Function)**
- **moved (jacobian, RingElement) to Core**
- **added helpers subclasses, descendants, allobjects**
- **downgraded a fatal error to an error signal**
- **undocumented all methods of Expressions**
- **added error when an unexported type is documented**
- **exported missing Bertini symbol**
- **exported missing VectorGraphics symbol**


mahrud:
@pzinn are the changes related to documentation of expressions here acceptable to you?

pzinn:
As far as I understand you've only "undocumented" things that were not documented anyway. so I can't imagine anyone would object to that.

mahrud:
> As far as I understand you've only "undocumented" things that were not documented anyway. so I can't imagine why anyone would object to that.

Not quite, there are two differences:
1. There are <160 nodes like `- Expression` which are currently just pointing to the documentation for `Expression`, this PR removes those from the index.
2. Even a missing documentation will appear under "Methods that use an expression" on the documentation for `Expression`. Marking them as undocumented removes this.

pzinn:
> 2\. Even a missing documentation will appear under "Methods that use an expression" on the documentation for `Expression`. Marking them as undocumented removes this.

I see. That's a bit radical. Maybe we do want to keep some of these methods listed? is there any way to do that w/o actually documenting them?

mahrud:
Exactly, this is why I've been trying to ask you which things are important and should appear in the documentation! Adding them as secondary keys is the only other way I can think of (e.g. with this pr, `value(Expression)` still appears under `Expression` because it's a secondary of that node while `assert(Expression)` appears under `Expression` and links to `assert` because it's a secondary of `assert`:

image
https://github.com/Macaulay2/M2/blob/18a83cd0d3c9ab85bc088b25357070a5c65a8d32/M2/Macaulay2/packages/Macaulay2Doc/debugging.m2#L315-L318
https://github.com/Macaulay2/M2/blob/18a83cd0d3c9ab85bc088b25357070a5c65a8d32/M2/Macaulay2/packages/Macaulay2Doc/doc12.m2#L285

I think this way undocumenting by default, except for nodes intentionally added as secondaries is probably easier to maintain.

ps: note that all the other methods still appear under `methods Expression`, they're just not listed online.

M2/Macaulay2/m2/help.m2
```diff
@@ -501,33 +501,25 @@ setAttribute(infoHelp#0, ReverseDictionary, symbol infoHelp)
...
-    synopsis,
-    waystouse,
-    technical }
-)
+    try symb := getGlobalSymbol toString key then (
+	waystouse := documentationValue(symb, key);
+	technical := getTechnical(symb, key)) else ();
+    DIV nonnull {
+	PARA {format tag, commentize title},
+	synopsis, waystouse, technical })
```
d-torrance, line 528: What happens here if we end up in the `else` statement above and `waystouse` and `technical` are undefined?

M2/Macaulay2/m2/installPackage.m2
```diff
@@ -592,21 +591,31 @@ isDocumentationComplete = pkg -> (
     -- and is run if CheckDocumentation => true is given.
     -- Takes ~22s for Macaulay2Doc
     resetCounters();
+    print toString pkg;
```
d-torrance, line 594: Should this be here?

mahrud: lol no, thank you

M2/Macaulay2/m2/help.m2
mahrud: `nonnull` gets rid of them.

d-torrance: Okay, I think maybe I don't completely understand how local variables work.  I would have expected that `waystouse` and `technical` would be symbols if the call to `getGlobalSymbol` raised an error.  But they're `null` instead?

The following seems to indicate that that's the case:

```m2
i1 : f = () -> if false then x := 2 else x

o1 = f

o1 : FunctionClosure

i2 : f() === null

o2 = true
```

mahrud: I think when the function is parsed, every instance of `:=` causes a new local variable with initial value null to be created in the function closure. This is a part of the interpreter that I don't fully understand because it's not documented and the source is very readable either. For instance, this is wild to me:
```m2
i1 : f = x -> y -> (print x; x = y);

i2 : g = f(2);

i3 : g(3)
2

o3 = 3

i4 : g(4)
3 -- it's changed!!!

o4 = 4
```

d-torrance: Ah, ok -- interesting

d-torrance (approved):

mahrud:
Can this be merged? #3388 is on top of this.

M2/Macaulay2/tests/normal/methods.m2
```diff
@@ -120,3 +120,10 @@ assert ( chainComplex (new X, FOO => 123 ) === (new OptionTable from {FOO => 123
 assert ( chainComplex (new X) === (new OptionTable from {FOO => BAR},new X from {}) )
 chainComplex X := identity
 assert( chainComplex (new X) === new X )
+
+-- this should list (net/info, HypertextContainer),
+-- even though HypertextContainer is not exported.
+assert(2 == length methods parent class help())
```
d-torrance, line 126: Was there a reason to call `help()` here and not just use `parent DIV`?  Or if we're trying to avoid importing `DIV`, `code` would work since it also returns a `DIV` object.

Calling `help()` can take a fair amount of memory (I seem to remember an issue about this but I can't find it...) and [one of the PPA builds](https://launchpadlibrarian.net/743391375/buildlog_ubuntu-bionic-i386.macaulay2_1.24.05+git202408112135-0ppa202408100259~ubuntu18.04.1_BUILDING.txt.gz) just crashed here.

mahrud: Oh sure, we can replace this with `code` of something.

M2/Macaulay2/m2/installPackage.m2
```diff
@@ -634,8 +666,10 @@ installPackage Package := opts -> pkg -> (
     verboseLog := if opts.Verbose or debugLevel > 0 then printerr else identity;
 
     use pkg;
-    -- TODO: make this more functional
-    chkdoc = opts.CheckDocumentation;			    -- oops, this will have a lingering effect...
+    if chkdoc then checkForErrors pkg;
```
d-torrance, line 669: Why was this call to `checkForErrors` added at the beginning of `installPackage`?  We haven't reset the counters yet, so it might complain about an old, out-of-date error.

Quick example:  Suppose we have the following silly package:

```m2
newPackage("Foo", Headline => "foo",)
beginDocumentation()

doc ///
  Key
    Foo
  Description
    Example
      1/0
///
```
Of course, the first call to `installPackage` fails.  So we go back and change that `1/0` to, say, `1/2`, save the file, and try again.  But it still thinks there's an error:

```m2
i2 : installPackage("Foo", FileName => "Foo.m2")
stdio:2:14:(3): error: installPackage: 1 error(s) occurred in running examples for package Foo
```
If we comment out the call to `checkForErrors` here, then the second call to `installPackage` works as expected -- it tries running the example again, and this time it's successful.

mahrud: I think the idea was to signal silent documentation errors from `loadPackage`.

> So we go back and change that 1/0 to, say, 1/2, save the file, and try again.

I don't think there's any guarantee that this would work. e.g. if the package called `needs` inside, the second time they might not be loaded at all. It would be good to get it to work, but there are numerous points of failure.

PR #3341, opened 2024-06-30
Format better "about" output

pzinn:
This is an attempt to improve the output of `about`.
before:
```
i1 : about resolution

o1 = {0 => ChainComplexExtras::resolution(ChainComplex)             }
     {1 => Complexes::resolution(Complex)                           }
     {2 => EagonResolution::resolution(EagonData)                   }
     {3 => Macaulay2Doc::resolution                                 }
     {4 => Macaulay2Doc::resolution(...,DegreeLimit=>...)           }
     {5 => Macaulay2Doc::resolution(...,FastNonminimal=>...)        }
     {6 => Macaulay2Doc::resolution(...,HardDegreeLimit=>...)       }
     {7 => Macaulay2Doc::resolution(...,LengthLimit=>...)           }
     {8 => Macaulay2Doc::resolution(...,PairLimit=>...)             }
     {9 => Macaulay2Doc::resolution(...,SortStrategy=>...)          }
     {10 => Macaulay2Doc::resolution(...,StopBeforeComputation=>...)}
     {11 => Macaulay2Doc::resolution(...,Strategy=>...)             }
     {12 => Macaulay2Doc::resolution(...,SyzygyLimit=>...)          }
     {13 => Macaulay2Doc::resolution(Ideal)                         }
     {14 => Macaulay2Doc::resolution(Matrix)                        }
     {15 => Macaulay2Doc::resolution(Module)                        }
     {16 => NCAlgebra::resolution(NCMatrix)                         }

o1 : NumberedVerticalList
```
after:
```
i1 : about resolution
 -- warning: symbol "minimize" in Core.Dictionary is shadowed by a symbol in ChainComplexExtras.Dictionary
 --   use the synonym Core$minimize

o1 = {0 => "resolution" -- projective resolution                                                                                                     }
     {1 => "resolution(...,DegreeLimit=>...)" -- compute only up to this degree                                                                      }
     {2 => "resolution(...,FastNonminimal=>...)" -- compute a non-minimal graded free resolution                                                     }
     {3 => "resolution(...,HardDegreeLimit=>...)"                                                                                                    }
     {4 => "resolution(...,LengthLimit=>...)" -- stop when the resolution reaches this length                                                        }
     {5 => "resolution(...,PairLimit=>...)" -- stop when this number of pairs has been handled                                                       }
     {6 => "resolution(...,SortStrategy=>...)"                                                                                                       }
     {7 => "resolution(...,StopBeforeComputation=>...)" -- whether to stop the computation immediately                                               }
[...]

PR #3387, opened 2024-08-02
Reverted `about`, added `headlines`

mahrud:
- **changed makeDocumentTag to strip -- comments**
- **reverted previous behavior of about**
- **added headlines as a help function**

Here's the upshot: you can get links to documentation along with the headlines:
```m2
i1 : headlines about firstFunction

o1 = 0 "firstFunction"      -- a silly first function
     1 "firstFunction(ZZ)"  -- a silly first function

i2 : headlines apropos "hilbert"

o2 = 0 "hilbertBasis"       -- computes the Hilbert basis of a Cone
     1 "hilbertFunction"    -- the Hilbert function                
     2 "hilbertPolynomial"  -- compute the Hilbert polynomial      
     3 "hilbertSeries"      -- compute the Hilbert series          

i3 : headlines methods syz

o3 = 0 "syz(GroebnerBasis)"  -- retrieve the syzygy matrix
     1 "syz(Matrix)"         -- compute the syzygy matrix
```
The output is a Table, but `help ZZ` and `code ZZ` still work.

On the other hand, `about` has been reverted to it's previous format so the entries are syntax highlighted:
```m2
i4 : about heft

o4 = {0 => Macaulay2Doc :: heft                }
     {1 => Macaulay2Doc :: heft vectors        }
     {2 => Macaulay2Doc :: heft(Monoid)        }
     {3 => Macaulay2Doc :: heft(PolynomialRing)}
     {4 => Macaulay2Doc :: heft(QuotientRing)  }
     {5 => Macaulay2Doc :: heft(Ring)          }
```

Hopefully this is a good compromise.

### DIFF
--- a/M2/Macaulay2/m2/document.m2
+++ b/M2/Macaulay2/m2/document.m2
@@ -137,8 +137,9 @@ DocumentTag ? DocumentTag := (x, y) -> x.Format ? y.Format
 DocumentTag ? String      := (x, y) -> x.Format ? y
 String      ? DocumentTag := (x, y) -> x        ? y.Format
 
--- helper for parsing " pkg :: key " to ("pkg", "key")
+-- helper for parsing " pkg :: key -- headline" to ("pkg", "key")
 parseDocumentTag := key -> (
+    key = first separate(" -- ", key); -- the spaces are there so "--" can be a valid key
     segments := separate("^[[:space:]]*|[[:space:]]*::[[:space:]]*|[[:space:]]*$", key);
     segments  = select(segments, segment -> segment =!= "");
     -- this is important, because the names of info nodes get extracted from text where
--- a/M2/Macaulay2/m2/document.m2
+++ b/M2/Macaulay2/m2/document.m2
@@ -198,7 +198,6 @@ makeDocumentTag String      := opts -> key -> (
     then error ("mismatching packages ", pkg, " and ", toString opts#Package, " specified for key ", key);
     if pkg === null then pkg = opts#Package;
     (makeDocumentTag' new OptionTable from {Package => pkg}) key)
-makeDocumentTag TOH := opts -> first
 
 -- before creating links, we recreate the document tag as a hack to
 -- correct its package, if it is incorrect (e.g. truncate, quotient)
--- a/M2/Macaulay2/m2/exports.m2
+++ b/M2/Macaulay2/m2/exports.m2
@@ -746,6 +746,7 @@ export {
 	"groupID",
 	"hash",
 	"hashTable",
+	"headlines",
 	"heft",
 	"height",
 	"help",
--- a/M2/Macaulay2/m2/help.m2
+++ b/M2/Macaulay2/m2/help.m2
@@ -347,6 +347,10 @@ headline = method(Dispatch => Thing)
 headline Thing := key -> getOption(fetchRawDocumentationNoLoad makeDocumentTag key, Headline)
 headline DocumentTag := tag -> getOption(fetchRawDocumentation getPrimaryTag tag, Headline)
 
+headlines = method()
+headlines List := L -> netList(Boxes => false, HorizontalSpace => 1,
+    apply(#L, i -> {i, TO2(tag := makeDocumentTag L#i, format tag), commentize headline tag}))
+
 -- Compare with SYNOPSIS in document.m2
 getSynopsis := (key, tag, rawdoc) -> (
     if rawdoc === null then return null;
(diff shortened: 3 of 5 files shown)


mahrud:
Actually, I'm just using `netList` to tabulate the output, so it's possible that the links are lost in the web interface ... @pzinn is `netList` outputting an html table on the web interface?

pzinn:
`netList` only produces a `Net`, so yeah, this is definitely a regression on the HTML side.
also, why `NumberedVerticalList.AfterPrint = null`? this makes it different from any other list, e.g., its ancestor `VerticalList`.

pzinn:
For reference: (links are in orange)
- before #3341:
Screenshot from 2024-08-06 16-28-48
- after #3341:
Screenshot from 2024-08-06 16-29-21
- after this PR:
Untitled

mahrud:
> `netList` only produces a `Net`, so yeah, this is definitely a regression on the HTML side.

I guess I wouldn't be opposed to changing `netList` to return a `TABLE` on the web, but it's okay I'll think of an alternative.

> also, why `NumberedVerticalList.AfterPrint = null`? this makes it different from any other list, e.g., its ancestor `VerticalList`.

We have plenty of types whose afterprint is silenced, usually because either it's obvious (e.g. integers) or annoying when it shows up (e.g. we don't want a random `DIV` appear after calling `help`!). I think the latter logic applies here: it's kind of annoying when a random `NumberedVerticalList` appears after calling `methods`. I think `NumberedVerticalList` was created specifically for `methods` anyway, and now is also used by `hooks` and `oeis`, but literally no other mention of it anywhere. Does this make sense?

pzinn:
Before this PR, the only difference between `NumberedVerticalList` and `OL` was that the former had an afterprint and the latter not (following the general principle that `List` types have afterprint but `Hypertext` types don't). After the PR, there won't be any difference, so might as well retire `NumberedVerticalList`...

mahrud:
The fix was trivial actually. Now it should look like this:
image

> following the general principle that `List` types have afterprint but `Hypertext` types don't

I don't know what you mean by general principle. I think this was an arbitrary choice, probably mainly because the Hypertext types aren't exported by default. Net and Describe also don't have an afterprint. I think this is just convenience, but if you feel strongly about this I'm happy to revert that commit.

> After the PR, there won't be any difference, so might as well retire `NumberedVerticalList`...

There is a huge difference! The whole point is that debugging tools that accept lists can accept the output of `methods`:
```m2
i1 : locate methods syz

o1 = {0 => (gb.m2:203:47-203:91)    }
     {1 => (matrix2.m2:260:29-263:4)}

i2 : makeDocumentTag methods syz

o2 = {0 => (Macaulay2Doc :: syz(GroebnerBasis))}
     {1 => (Macaulay2Doc :: syz(Matrix))       }

i3 : help methods syz

o3 = syz(GroebnerBasis) -- retrieve the syzygy matrix
     ************************************************
...
     ------------------------------------------------------------------------------------------------------------------------

     syz(Matrix) -- compute the syzygy matrix
     ****************************************
...
```
Currently implementing these is as simple as:
```m2
locate List        := List     => x -> apply(x, locate)
```
But if `methods` returned an `OL` then you'd have to go through and look in each `LI` item only to realize that sequences have been turned into lists:
```m2
i27 : OL methods syz

o27 =   0. syzGroebnerBasis
        1. syzMatrix

i28 : peek OL methods syz

o28 = OL{LI{syz, GroebnerBasis}, LI{syz, Matrix}}
```

mahrud:
Absolutely not. `methods` is fundamentally separate from documentation, and is used very frequently in Core. Many, many combinations (like `code methods`, `locate methods`, `locate makeDocumentTag methods`, `help methods`) are out there and they only work because of the simplicity of `methods`.

pzinn:
correction, this was about `about`, so my point was, who needs `about firstFunction` when you can have
`headlines about firstFunction`.

mahrud:
I thought about that, but I actually really like how `headlines about blah` or `headlines apropos "foo"` sound like plain English. In particular, a feature I really like in certain languages is when you can just learn the basics (e.g. `methods`, `apropos`, `about`, `headlines`, `makeDocumentTag`, `locate`, etc.) then combine them like legos, e.g:
```m2
headlines about "hilbert"
locate makeDocumentTag about "hilbert"
```
and you can switch out `about` with `methods` or `apropos` and you get other useful results. This interoperability makes the language easy to learn.

More practically, `headlines` is a really simple routine, whereas `about` is pretty complicated and has an option and changes `lastabout` ... I prefer to keep them separate.

mahrud:
@pzinn Does this have your approval now? If so, could you submit a review? (Files Changes Tab > Green Review Changes Button)

pzinn (approved):

PR #3390, opened 2024-08-05
Move 3 newPackage options to beginDocumentation

### LINKED ISSUE #3369: Move OptionalComponentsPresent from newPackage to beginDocumentation
I generally don't appreciate packages that run an external program upon being loaded (which also ends up happening when `Macaulay2Doc` is being installed, c.f. #3292).

Perhaps `OptionalComponentsPresent` should be set _after_ `newPackage` in the call to `beginDocumentation`, so that loading could stop there if the optional component is not present.

(Originally from https://github.com/Macaulay2/M2/pull/3366#discussion_r1685557557)

mahrud:
This is a draft for now to see what breaks. Comments welcome.

- **fixed newPackage to permit terminal null entries**
- **moved 3 newPackage options to beginDocumentation**
- **adjusted existing packages with optional components**
- **corrected package options in TropicalTorics**
- **removed FirstPackage stuff from Polymake.m2**
- **unexported two undefined gfanInterface symbols**

Closes #3369.

### DIFF
--- a/M2/Macaulay2/m2/packages.m2
+++ b/M2/Macaulay2/m2/packages.m2
@@ -209,7 +209,6 @@ newPackage = method(
     Options => {
 	Authors                   => {},
 	AuxiliaryFiles            => false,
-	CacheExampleOutput        => null,
 	Certification             => null,
 	Configuration             => {},
 	Date                      => null,
@@ -218,11 +217,9 @@ newPackage = method(
 	HomePage                  => null,
 	InfoDirSection            => "Macaulay2 and its packages",
 	Keywords                  => {"Uncategorized"},
-	OptionalComponentsPresent => null,
 	PackageExports            => {},
 	PackageImports            => {},
 	Reload                    => false,
-	UseCachedExampleOutput    => null,
 	Version                   => "0.0"
 	})
 newPackage Sequence := opts -> x -> newPackage(x#0, opts) -- to allow null entries
@@ -316,8 +313,6 @@ newPackage String := opts -> pkgname -> (
 		    DebuggingMode => loadOptions.DebuggingMode}, last)));
     --
     if opts.DebuggingMode and not debuggingMode then opts = opts ++ {DebuggingMode => false};
-    if opts.OptionalComponentsPresent === null  then opts = opts ++ {OptionalComponentsPresent => opts.CacheExampleOutput =!= true};
-    if opts.UseCachedExampleOutput === null     then opts = opts ++ {UseCachedExampleOutput => not opts.OptionalComponentsPresent};
     --
     packagePrefix := (
 	-- Try to detect whether we are loading the package from an installed version.
@@ -521,7 +516,15 @@ endPackage String := title -> (
      pkg.PackageIsLoaded = true;
      pkg)
 
-beginDocumentation = () -> (
+beginDocumentation = {
+    CacheExampleOutput        => null,
+    OptionalComponentsPresent => null,
+    UseCachedExampleOutput    => null, -- TODO: not used by any package, retire this!
+    } >> opts -> () -> (
+    -- the options are used in installPackage and check
+    if opts.OptionalComponentsPresent === null  then opts = opts ++ {OptionalComponentsPresent => opts.CacheExampleOutput =!= true};
+    if opts.UseCachedExampleOutput === null     then opts = opts ++ {UseCachedExampleOutput => not opts.OptionalComponentsPresent};
+    currentPackage.Options ++= opts;
     pkgname := currentPackage#"pkgname";
     if loadPackageOptions#?pkgname and not loadPackageOptions#pkgname.LoadDocumentation
     and currentPackage#?rawKeyDB and isOpen currentPackage#rawKeyDB then (
--- a/M2/Macaulay2/m2/packages.m2
+++ b/M2/Macaulay2/m2/packages.m2
@@ -205,6 +205,7 @@ getpkgNoLoad = pkgname -> if isPackageLoaded pkgname then value PackageDictionar
 -----------------------------------------------------------------------------
 
 newPackage = method(
+    Dispatch => Thing,
     Options => {
 	Authors                   => {},
 	AuxiliaryFiles            => false,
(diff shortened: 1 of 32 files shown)


M2/Macaulay2/m2/packages.m2
```diff
@@ -224,6 +225,7 @@ newPackage = method(
 	UseCachedExampleOutput    => null,
 	Version                   => "0.0"
 	})
+newPackage Sequence := opts -> x -> newPackage(x#0, opts) -- to allow null entries
```
d-torrance, line 225: Would `newPackage(String, Nothing)` make sense instead to solve the specific issue of trailing commas?

With the new behavior, I'd be a little concerned that something like `newPackage("Foo", 1, 2, 3, 4, 5)` would be allowed.

M2/Macaulay2/m2/packages.m2
```diff
@@ -521,7 +516,15 @@ endPackage String := title -> (
      pkg.PackageIsLoaded = true;
      pkg)
 
-beginDocumentation = () -> (
+beginDocumentation = {
+    CacheExampleOutput        => null,
+    OptionalComponentsPresent => null,
+    UseCachedExampleOutput    => null, -- TODO: not used by any package, retire this!
```
d-torrance, line 522: I use this option in the Debian package to fix some reproducibility issues and to cache some examples that occasionally crash, so I'd prefer to keep it around.

M2/Macaulay2/packages/RInterface/doc.m2
```diff
@@ -2,7 +2,7 @@
 -- RInterface documentation --
 ------------------------------
 
-beginDocumentation()
+beginDocumentation(OptionalComponentsPresent => RPresent)
```
d-torrance, line 5: The builds are failing right now because the `RInterface` tests are loaded before this call to `beginDocumentation()` and so the `OptionalComponentsPresent` package option doesn't exist yet, which causes a problem when calling `check`.

A band-aid fix for this specific package would be to switch the order of the two `load` lines at the bottom of `RInterface.m2`, but I wonder if `TEST` should raise an error if the option doesn't exist yet, so we know that we're calling it after `beginDocumentation()`.

d-torrance: Actually, that's not quite right.  We're calling `end` since R isn't found before `beginDocumentation()`.  This package is unusual in that regard.

But regardless, there should probably be some kind of check (either in `TEST` or `check`) to make sure that the `OptionalComponentsPresent` option exists to avoid strange "key not found" error messages.

d-torrance: I just pushed a fix for the RInterface-specific issue.

M2/Macaulay2/m2/packages.m2
mahrud: I feel like `methods newPackage` showing `(newPackage, Sequence)` is a lot less surprising than `(newPackage, String, Nothing)`. I can assert that the other entries are null if you want?

M2/Macaulay2/m2/packages.m2
mahrud: `UseCachedExampleOutput`? Can you say more about how you use it? Maybe I don't understand it well. If cached examples are present, wouldn't M2 automatically use them? (I think it does, e.g. see the single `realrank` example that I added)

M2/Macaulay2/packages/RInterface/doc.m2
mahrud: Hmm, I think it shouldn't matter where `beginDocumentation()` is called. Actually, I tend to prefer defining tests _before_ calling `beginDocumentation()`, because I've ran into weird issues where the tests weren't loaded since M2 stopped at `beginDocumentation()`. I'm not sure what specific situation caused this though.

For this issue, I think I might prefer treating the absence of `OptionalComponentsPresent` as it being true. I think this would solve the failure, no?

d-torrance: Yeah, you're right -- it doesn't matter since we won't be calling `check` until after the package is loaded.

Yes, your idea is much better than manually setting it for a specific package like I was doing!

mahrud:
For now, I just wanted to make sure all tests passed, but I would like to simplify this further before moving it out of draft mode. Here are a few goals, but feel free to add more:
- [ ] keep `beginDocumentation()` simple, ideally one option only.
- [ ] maybe allow use of cached examples opportunistically, or even provide a function for generating a single cached example, so we can cache only the necessary examples (e.g. `realrank` is the only one from `CoincidenceRootLoci` that needs to be cached)
- [ ] simplify the logic in the background, maybe debug mystery issue I described above with tests disappearing after `beginDocumentation()`
- [ ] think about whether `CacheExampleOutputs` should be transitive.
- [ ] think about whether `readPackage` should know whether optional dependencies are present, because with this PR it will not
- [ ] maybe have one distinguished package for each optional dependency so we don't check for existence of the same software in multiple places
- [ ] along the way, stop using `type` or `command` and use `findProgram`
- [ ] bring back the `newPackage` options for backwards compatibility, but have them give a deprecation warning.

M2/Macaulay2/m2/packages.m2
d-torrance: I think it needs to be true in order to copy any cached examples over:

https://github.com/Macaulay2/M2/blob/ec9e9ac60ed4a8e791448942202f077a88a87e15/M2/Macaulay2/m2/installPackage.m2#L556-L562

I use it in the Debian package by [patching several packages](https://salsa.debian.org/math-team/macaulay2/-/blob/debian/latest/debian/patches/use-cached-examples.patch) so that `OptionalComponentsPresent` is true, and then copying the cached examples from [this directory](https://salsa.debian.org/math-team/macaulay2/-/tree/debian/latest/debian/examples) over to the corresponding package directories before building the docs.

M2/Macaulay2/packages/RInterface/doc.m2
d-torrance: I just pushed an alternate fix based on your suggestion.  One of the builds was failing even after the RInterface-specific fix, so it must not be the only package where this was an issue.

d-torrance:
> or even provide a function for generating a single cached example, so we can cache only the necessary examples

This would be awesome.  I have a [pretty hacky function](https://salsa.debian.org/math-team/macaulay2/-/blob/debian/latest/debian/scripts/Debian.m2#L105) that does this for the Debian package, but a cleaner solution would be great.

M2/Macaulay2/m2/packages.m2
d-torrance: Sure, that works

PR #3392, opened 2024-08-07
msolveRealSolutions

antonleykin:
I've edited this method options and the docs... as [PR #3366](https://github.com/Macaulay2/M2/pull/3366) got merged.

### DIFF
--- a/M2/Macaulay2/packages/Msolve.m2
+++ b/M2/Macaulay2/packages/Msolve.m2
@@ -169,7 +169,7 @@ addHook((saturate, Ideal, RingElement), Strategy => Msolve,
     (opts, I, f) -> try ideal msolveSaturate(I, f))
 
 msolveRealSolutions = method(TypicalValue => List,
-    Options => msolveDefaultOptions ++ { "output type" => "rationalInterval" })
+    Options => msolveDefaultOptions ++ { "output type" => "rational interval" })
 msolveRealSolutions Ideal := opt -> I0 -> (
     (S, K, I) := toMsolveRing I0;
     mOut := msolve(S, K, I_*, "", opt); -- optional: -p precision
@@ -178,9 +178,9 @@ msolveRealSolutions Ideal := opt -> I0 -> (
     if solsp_0>0 then (error "Input ideal not zero dimensional, no solutions found.";);
     if (solsp_1)_0>1 then (print "unexpected msolve output, returning full output"; return solsp;);
     sols:=(solsp_1)_1;
-    if opt#"output type"=="rationalInterval" then return sols;
-    if opt#"output type"=="floatInterval" then return (1.0*sols);
-    if opt#"output type"=="float" then return (for s in sols list(for s1 in s list sum(s1)/2.0));
+    if opt#"output type"=="rational interval" then return sols;
+    if opt#"output type"=="float interval" then return (1.0*sols);
+    if opt#"output type"=="float middle point" then return (for s in sols list(for s1 in s list sum(s1)/2.0));
     );
 
 msolveRUR = method(TypicalValue => List, Options => msolveDefaultOptions)
@@ -389,33 +389,26 @@ Node
 	    value is also represented by a two element list of rational numbers, {a, b},
 	    this means that that coordinate of the solution has a value greater than or
 	    equal to a and less than or equal to b. This interval is computed symbolically
-	    and its correctness is guaranteed by exact methods. Note that using the option
-	    Output one may specify the output in terms of either a float interval with
-	    "floatInterval" or an average of the interval endpoints as a single float with "float".
-
+	    and its correctness is guaranteed by exact methods.
+	Text
+	    Option "output type" (default = "rational interval") gives alternatives ways to provide output 
+	    eitherusing @TO RRi@ ("float interval")
+	    or by taking a floting point approximation of the average of the inteval endpoints ("float middle point").
     	Text
 	    First an example over a finite field
 	Example
-	    R = QQ[x_1..x_3]
-	    f = (x_1-1)*x_1
-	    g = (x_2-5/2)*(x_2-1/2)
-	    h = (x_3-2)*(x_3-1)
-	    I = ideal (f,g,h)
-	    sols=msolveRealSolutions(I)
-	    floatSolsInterval=msolveRealSolutions(I,"output type"=>"floatInterval")
-	    floatAproxSols=msolveRealSolutions(I,"output type"=>"float")
-	    	    
+	    R = QQ[x,y]
+	    I = ideal {(x-1)*x, y^2-5}
+	    sols = msolveRealSolutions I
+	    floatSolsInterval = msolveRealSolutions(I,"output type"=>"float interval")
+	    floatAproxSols = msolveRealSolutions(I,"output type"=>"float middle point")	    	    
 	Text
-	    Note in cases where solutions have multiplicity this is not returned, but the presence
-	    of multiplicity also does not reduce accuracy or reliability in any way.
+	    Note in cases where solutions have multiplicity this is not reflected in the output.
+	    While the solver does not return multiplicities,
+	    it reliably outputs the verified isolating intervals for multiple solutions.  
 	Example 
-	    f = (x_1-1)*x_1^3
-	    g = (x_2-5/2)*(x_2-1/2)
-	    h = (x_3-2)*(x_3-1)
-	    I = ideal (f,g,h)
-	    sols=msolveRealSolutions(I)
-	    floatSolsInterval=msolveRealSolutions(I,"output type"=>"floatInterval")
-	    floatAproxSols=msolveRealSolutions(I,"output type"=>"float")
+	    I = ideal {(x-1)*x^3, y^4-25}
+	    floatAproxSols=msolveRealSolutions(I,"output type"=>"float interval")
 Node 
     Key
     	msolveRUR
--- a/M2/Macaulay2/packages/Msolve.m2
+++ b/M2/Macaulay2/packages/Msolve.m2
@@ -392,7 +392,7 @@ Node
 	    and its correctness is guaranteed by exact methods.
 	Text
 	    Option "output type" (default = "rational interval") gives alternatives ways to provide output 
-	    eitherusing @TO RRi@ ("float interval")
+	    either using @TO RRi@ ("float interval")
 	    or by taking a floating point approximation of the average of the interval endpoints ("float middle point").
     	Text
 	    First an example over a finite field
--- a/M2/Macaulay2/packages/Msolve.m2
+++ b/M2/Macaulay2/packages/Msolve.m2
@@ -393,7 +393,7 @@ Node
 	Text
 	    Option "output type" (default = "rational interval") gives alternatives ways to provide output 
 	    eitherusing @TO RRi@ ("float interval")
-	    or by taking a floting point approximation of the average of the inteval endpoints ("float middle point").
+	    or by taking a floating point approximation of the average of the interval endpoints ("float middle point").
     	Text
 	    First an example over a finite field
 	Example
--- a/M2/Macaulay2/packages/Msolve.m2
+++ b/M2/Macaulay2/packages/Msolve.m2
@@ -43,11 +43,9 @@ msolveProgram = findProgram("msolve", "msolve --help",
     Verbose => debugLevel > 0)
 
 if msolveProgram === null then (
-    if not (options currentPackage).OptionalComponentsPresent
-    then ( printerr "warning: msolve cannot be found; ending"; end )
-    else ( printerr("warning: msolve found but its version is older than v" | msolveMinimumVersion);
-	-- note: msolve -h returns status code 1 :/
-	msolveProgram = findProgram("msolve", "true", Verbose => debugLevel > 0)))
+    printerr("warning: could not find msolve with version at least v" | msolveMinimumVersion);
+    -- note: msolve -h returns status code 1 :/
+    msolveProgram = findProgram("msolve", "true", Verbose => debugLevel > 0))
 
 msolveDefaultOptions = new OptionTable from {
     Threads => allowableThreads,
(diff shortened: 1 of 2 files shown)


mahrud:
Ah sorry didn't realize you had more changes coming in.

M2/Macaulay2/packages/Msolve.m2
```diff
@@ -389,33 +389,26 @@ Node
...
 	    this means that that coordinate of the solution has a value greater than or
 	    equal to a and less than or equal to b. This interval is computed symbolically
-	    and its correctness is guaranteed by exact methods. Note that using the option
-	    Output one may specify the output in terms of either a float interval with
-	    "floatInterval" or an average of the interval endpoints as a single float with "float".
-
+	    and its correctness is guaranteed by exact methods.
+	Text
+	    Option "output type" (default = "rational interval") gives alternatives ways to provide output 
+	    eitherusing @TO RRi@ ("float interval")
```
mahrud, line 395:
```suggestion
	    either using @TO RRi@ ("float interval")
```

mahrud: @fchapoton I'm surprised codespell didn't catch this. Do you know what happened?

antonleykin:
I've noticed `codespell` catching a proper subset of typos before. In fact, we observed some dismal ~50% "success" rate with @Martin-Helmer when the docs had a lot of typos.

mahrud:
Anton, how open are you to changing the options and outputs of `msolveRealSolutions` to be an `RRi` interval type? If you want I can take a stab at it and open a PR on your fork, but if you prefer it this way because this is more useful for your workflow then I don't need to touch it. I just think it's strange that the documentation explicitly mentions the `RRi` type, but the output is _not_ and `RRi` object.

Martin-Helmer:
I currently have code I am working on that uses the "output type"=>"float" option. And I think that the rational interval, which is the raw output from msolve, should be preserved.  

So for me the two output types I really want at the the approximate roots as floating point numbers and the rational intervals. 

As far as I know the msolve documentation doesn't mention the RRi type at all..? I didn't write anything about RRi anyway.

So I have no objection to adding some option that allows for an RRi type to be returned, but I can't see myself ever using it. As I said, I do have a package I am working on for some real algebraic geometry things that will use the msolveRealSolutions a lot, but only ever the "output type"=>"float" option.

antonleykin:
> Anton, how open are you to changing the options and outputs of `msolveRealSolutions` to be an `RRi` interval type? If you want I can take a stab at it and open a PR on your fork, but if you prefer it this way because this is more useful for your workflow then I don't need to touch it. I just think it's strange that the documentation explicitly mentions the `RRi` type, but the output is _not_ and `RRi` object.

Yep, I think, actually it would be fine if the "float interval" to get`RRi`. I've typed that into the documentation, but... I guess we should change the code 😄 
unless @Martin-Helmer has objections.
Martin, I also renamed option values into more straightforward strings "rational interval", "float interval", "float middle point". I hope no objections here.

antonleykin:
> I currently have code I am working on that uses the "output type"=>"float" option. And I think that the rational interval, which is the raw output from msolve, should be preserved.
> 
> So for me the two output types I really want at the the approximate roots as floating point numbers and the rational intervals.
> 
> As far as I know the msolve documentation doesn't mention the RRi type at all..? I didn't write anything about RRi anyway.
> 
> So I have no objection to adding some option that allows for an RRi type to be returned, but I can't see myself ever using it. As I said, I do have a package I am working on for some real algebraic geometry things that will use the msolveRealSolutions a lot, but only ever the "output type"=>"float" option.

`RRi` is something from M2, not msolve. I think [interval(QQ,QQ)](https://www.macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_interval.html) would be an appropriate constructor to use --- just give it two rational endpoints.

Martin-Helmer:
changing "float interval" (or some other equivalently named option if preferred) to return RRi sounds good to me! 

Calling the midpoint option  "float middle point" is also good for me.

mahrud:
> So for me the two output types I really want at the the approximate roots as floating point numbers and the rational intervals.

Here's a mockup of what I have in mind:
```m2
i1 : needsPackage "Msolve";

i2 : R = QQ[x,y];

i3 : I = ideal "x2-2,y3-3";

o3 : Ideal of R

i4 : sols = msolveRealSolutions(I, RRi_53) -- could also just be RRi

o4 = {{[-1.41421,-1.41421], [1.44225,1.44225]}, {[1.41421,1.41421], [1.44225,1.44225]}}

o4 : List

i5 : class \ flatten oo

o5 = {RRi, RRi, RRi, RRi}

o5 : List

i6 : sols#0#0

o6 = [1.41421356237309,1.4142135623731]

o6 : RRi (of precision 53)

i7 : midpoint oo

o7 = -1.41421356237309

o7 : RR (of precision 53)

i8 : floatAproxSols = msolveRealSolutions(I, RR) -- or RR_53 or whatever

o8 = {{-1.41421, 1.44225}, {1.41421, 1.44225}}

o8 : List

i9 : class \ flatten oo

o9 = {RR, RR, RR, RR}

o9 : List

i10 : msolveRealSolutions(I, QQi)

         481231938336009023090067544955250113853  7519249036500140985782305389925783029    490772097474911374874256598976936699839  245386048737455687437128299488468349921        481231938336009023090067544955250113855    240615969168004511545033772477625056927    7668314023045490232410259359014635935  490772097474911374874256598976936699841
o32 = {{[---------------------------------------, -------------------------------------], [---------------------------------------, ---------------------------------------]}, {[- ---------------------------------------, - ---------------------------------------], [-------------------------------------, ---------------------------------------]}}
         340282366920938463463374607431768211456  5316911983139663491615228241121378304    340282366920938463463374607431768211456  170141183460469231731687303715884105728        340282366920938463463374607431768211456    170141183460469231731687303715884105728    5316911983139663491615228241121378304  340282366920938463463374607431768211456

o32 : List

i33 : class \ flatten oo

o33 = {QQi, QQi, QQi, QQi}

o33 : List
```

Though `QQi` above is not an existing type, and I just realized that the `mpfi` library that we use only allows floating point intervals, not rational intervals, but we can make it a type in the package (i.e. without the ability to define matrices, ideals, etc. over them like `RRi`).

If we accept `RR_53` or `RRi_53` or whatever as input, we can read the precision of the field and pass it to `msolve` via `-p precision` (default 32) also, rather than have yet another option.

@MichaelABurr do you think adding `QQi` via [MPRIA](https://www.gnu.org/software/mpria/) would be useful, or would it just bloat the dependencies and we should instead just use a pair of rationals or define a rational interval type in the top level?

mahrud:
> I think [interval(QQ,QQ)](https://www.macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_interval.html) would be an appropriate constructor to use --- just give it two rational endpoints.

Is it okay that M2 converts the rational numbers to 53bit floats here? This is not going to be as precise as a true `QQi` type would be.

mahrud:
Actually, returning each solution as a `Point` object from `NAGtypes` would also be nice, but again, I realize that I'm not the primary audience of this method.

Martin-Helmer:
Using RR to get the midpoint float is good, and etc. i.e. your mock up looks good to me. 

For adding a QQi structure, this is fine, I don't mind either way. I am also not sure if it adds anything? 

For setting the msolve precision parameter, I think as long as the RR option gives numbers in either RR or RR_53 but uses the msolve default precision of 32, i.e. doesn't force a higher than default precision without the user explicitly requesting that. 

For my use I want floats in RR or RR_53 returned, but don't want to up the precision on msolve in most cases (but would still prefer to have the outputs in at least RR_53 so that subsequent calculations don't suffer from round off issues). 

It would also be okay if a NAGType point was returned, but I am also not sure if I see any advantage to that. Since there is no additional info beyond the coordinates to store, and if people want a point type they can create one using the list.

Martin-Helmer:
@mahrud @antonleykin : A more significant issue. It seems like we have introduced a bug in msolveEliminate, pretty much any non-trival example seems not to be working correctly.  There seems to be two things

1. Something with the new switching rings stuff that Mahrud added is scrambling variables (i.e. the output polynomials are the correct ones if one permutes certain variables). This I don't immediately know how to fix since I am not quite sure how this is meant to work and would need to stare at it awhile. 
2. It seems like over Q msolve is now returning all generators of the GB in the elimination order, not just the generators of the elimination ideal in the smaller ring. Or... it may be returning all, but the coefficients are only correct for the ones in the smaller ring.... I am not yet sure. This one I know how to fix once I figure out which of these two is happening. 

For 2. this must have changed between 0.6.4 and 0.7.1 (the two versions I have for testing). While this is better, we then need to rework the output. 

Here is an example which produces bad behavior:

```
needsPackage "Msolve";
 R=QQ[x..z,t]
K=ideal(x^6+y^6+x^4*z*t+z^3,36*x^5+60*y^5+24*x^3*z*t,-84*x^5+10*x^4*t-56*x^3*z*t+30*z^2,-84*y^5-6*x^4*t-18*z^2,48*x^5+10*x^4*z+32*x^3*z*t,48*y^5-6*x^4*z,14*x^4*z+8*x^4*t+24*z^2)
W1=msolveEliminate(R_0,K)
W2=eliminate(R_0,K)
sub(W1,R)==W2 
```

Note that, e.g., the first polynomial of W1 and of W2 is the same if you swap x->y, y->z, z->x etc.

Martin-Helmer:
Currently I am not even 100% sure if this is an us bug or an msolve bug, but I think it is an us bug. I tried switching the eliminate command to always do the finite feild case, i.e. to assume the whole GB of the ideal was being returned, but that made new and weird problems.

mahrud:
I created https://github.com/antonleykin/M2/pull/31.

> For adding a QQi structure, this is fine, I don't mind either way. I am also not sure if it adds anything?

Probably not, but it's a thin layer over `List` type, so it wouldn't get in they way at all, but maybe look a little better.

> Can others confirm?

Yes, and I think it's a bug in the engine code that parses msolve's output. The output ideal has 11 generators, the first 6 are processed fine, and this is the seventh in msolve's output file:
```
142809155371008000*p_2^8-816052316405760000*p_2^7*p_3-10669301142493593600*p_2^6*p_3^2-329892659597401915392*p_2^5*p_3^3+18635283636736000*p_1*p_2^3*p_3^4-185242799667232333824*p_2^4*p_3^4+468447388748134531992*p_1*p_2^4-1981763745539690457421*p_2^5+125788164547968000*p_1*p_2^3*p_3-1250301544861771053312*p_2^4*p_3
```
But this is what M2 reads:
```m2
-1869611008z8-360710144z7t-1476395008z6t2+1073741824z5t3+359976960yz3t4+1002741760z4t4+870368152yz4+388110003z5+1356102656yz3t+1745012480z4t 802816z5t4+458752z4t5+13125z6+31850496y5t-806400z5t+6193152z4t2+11943936z3
```
I think the problem is that the digits are too large?

mahrud:
Ah, yes, observe:
```math
(142809155371008000 - (-1869611008)) \mod 2^{32} = 0
```

Martin-Helmer:
Thanks Mahrud! Yes this definitely seems like the issue. 

@mikestillman : I guess the read msolve core function is limited to using 32-bit fixed size integers somewhere? I am not sure if just using larger integers will fix it? Or if that will cause any other issues?

mahrud:
It's just not done yet:
>  //  it should be able to read arbitrary precision ints too... 

https://github.com/Macaulay2/M2/blob/ec9e9ac60ed4a8e791448942202f077a88a87e15/M2/Macaulay2/e/BasicPolyListParser.cpp#L41-L61

But I'm still confused ... why are there 7 more generators than M2's `eliminate`? msolve doesn't seem to be returning the extra generators in other, simpler examples over QQ, only ZZ/p, except for this example. Maybe it's also an msolve bug? Or an eliminate bug?

Martin-Helmer:
I think the only bug is the coefficient overflow. If you do a GB in the ring  R=QQ[x..z,t, MonomialOrder=>{1,3}], and then select the first 11 generators you get the same ideal as that returned by eliminate (but with a different generating set, since eliminate uses a different monomial order than msolveEliminate). 

So I think we just need to fix the coefficient overflow bug.

Martin-Helmer:
So in particular, as expected over Q, msolve is only returning the generators in the smaller ring... but our reading is going wrong.

mikestillman:
Ah, I'll look at the reading matrix code, it should definitely handle arbitrary size integer input

mahrud:
In the [PR](https://github.com/antonleykin/M2/pull/31) on Anton's fork, I also fixed up the top-level msolve parsing code, so now if you set `use'readMsolveOutputFile := false` you'll get the right coefficients for this example, but this routine is kinda hacky so some other tests fail.

mahrud:
The msolveRealSolutions example just failed on another build:
```m2
i7 : floatSolsInterval=msolveRealSolutions(I,"output type"=>"floatInterval")
sh: line 1: 83025 Cputime limit exceeded: 24  /usr/local/bin/msolve -t 5 -v 0 -f /var/folders/sy/86bqx33x44j7mfsyzj1snqmh0000gn/T/M2-82879-0/2-in.ms -o /var/folders/sy/86bqx33x44j7mfsyzj1snqmh0000gn/T/M2-82879-0/2-out.ms > /var/folders/sy/86bqx33x44j7mfsyzj1snqmh0000gn/T/M2-82879-0/3.out 2> /var/folders/sy/86bqx33x44j7mfsyzj1snqmh0000gn/T/M2-82879-0/3.err
stdio:7:37:(3): error: msolve returned an error (24):
```
This is a tiny example, has this ever happened before?

mahrud:
Doug pointed out something that didn't make sense in https://github.com/Macaulay2/M2/pull/3397#pullrequestreview-2229900862 so I pushed a fix.

Martin-Helmer:
@mahrud: No the msolve real solutions hasn't failed like that before... I am also confused... as you say it is a very simple example.

d-torrance (approved):

Issue #2581, opened 2022-08-24
List ^ ZZ for permutations


mahrud:
Here's something that might be nice for combinatorialists working with permutations:
```m2
i1 : w = {2,0,1}

o1 = {2, 0, 1}

o1 : List

i2 : w^2 -- essentially w_w

o2 = {1, 2, 0}

o2 : List

i3 : w^3 -- essentially (w_w)_w

o3 = {0, 1, 2}

o3 : List

i4 : w^-1 -- essentially inversePermutation w

o4 = {1, 2, 0}

o4 : List

i5 : w^-2 -- essentially (inversePermutation w)_(inversePermutation w)

o5 = {2, 0, 1}

o5 : List
```
Here's a simple implementation:
```m2
List ^ ZZ := List => (w, n) -> fold(
    if n < 0 then (-n):inversePermutation w else n:w,
    toList(0 ..< #w),
    (w, v) -> v_w)
```
What's nice is that if the entries are not integers `List _ List` handles the errors correctly.

DanGrayson:
Good idea, but introduce a new type of list called "Permutation" to implement these things for.

mahrud:
> Good idea, but introduce a new type of list called "Permutation" to implement these things for.

Part of the goal is to appeal to people closer to combinatorics by keeping the syntax as close to the standard syntax as possible (in the same way that, for instance, `QQ[x,y,z]` appeals to others).

TheGrateSalmon:
I have written some code to resolve this and add some other permutation methods (e.g., inversions), so there should be a pull request in the next few weeks (hopefully). 

However, I think there is a convention choice that has to be made. Namely, in any math that I have seen, permutations are 1-indexed, but functionally in code, permutations are 0-indexed; i.e., should permutations be permutations on 0..n-1 or 1..n? While helper functions to go between the two conventions is possible so that the user has choice in which convention they use, this seems very unwieldy. Are there any thoughts on this?

mahrud:
I think we should go with the 0-indexed convention, because there's actually a very easy way to convert the 1-indexed notation into 0-based: simply prepend a 0 and in the end drop the 0!

Here's what I mean: let's say you want to check that (3,1,2)^2 is (2,3,1) in the 1-based notation, you can do it this way:
```m2
i3 : w = {0,3,1,2}; w_w

o4 = {0, 2, 3, 1}
```

PR #3393, opened 2024-08-07
New Permutations package

TheGrateSalmon:
New package implementing a `Permutation` type by Sean Grate. Its intent is to have a standard and consistent type for permutations as more combinatorics-focused packages get added to Macaulay2. Much of the inspiration for the implemented methods came from the [Sage implementation](https://doc.sagemath.org/html/en/reference/combinat/sage/combinat/permutation.html). This also resolves [Issue #2581](https://github.com/Macaulay2/M2/issues/2581).


mahrud:
1. Add your package to the `=distributed-packages` file, otherwise it's basically ignored.
2. I think having both `Permutations.m2` and `Permutations/Permutations.m2` could lead to hard to decipher errors (e.g. try `needsPackage "Permutations"` inside the `Permutations` directory). My suggestion would be to put most of the content of  `Permutations/Permutations.m2` in `Permutations.m2`, and name the other two files `tests.m2` and `docs.m2`.
3. Typos:
```
M2/Macaulay2/packages/Permutations/PermutationsDOC.m2:241: postive ==> positive
M2/Macaulay2/packages/Permutations/PermutationsDOC.m2:622: simulataneously ==> simultaneously
```

M2/Macaulay2/packages/Permutations.m2
```diff
@@ -0,0 +1,98 @@
...
+        {Name => "Sean Grate", 
+         Email => "sean.grate@auburn.edu", 
+         HomePage => "https://seangrate.com/"}
+  },
+    Headline => "functions for working with permutations",
+    PackageExports => {
+        -- "SimplicialComplexes",
+        -- "SimplicialDecomposability",
+        -- "Posets",
+        -- "MinimalPrimes"
```
d-torrance, line 17: Is there a reason all these packages in `PackageExports` are commented out?

M2/Macaulay2/packages/Permutations.m2
```diff
@@ -0,0 +1,98 @@
...
+    },
+    DebuggingMode => true
```
d-torrance, line 19: `DebuggingMode` should either be set to `false`, or this option removed completely (since `false` is the default).

M2/Macaulay2/tests/normal/permutations.m2
```diff
@@ -1,5 +0,0 @@
-f = id_(ZZ^5)
-v = {1,2,4,3,0}
-w = {3,1,2,4,0}
-assert(f_(v_w)==(f_v)_w)
-assert(f_(v_w)==(f_v)*(f_w))
```
d-torrance, line 5: Why is this test file being removed?

mahrud: It's moved to the package tests.

TheGrateSalmon: Originally, I had tried to implement all of this directly into M2, i.e., not as a package, so I had written all of its tests here. Upon recommendation, I then tried to refactor all of permutations code I had written into a package and I simply forgot that M2 actually had this file to begin with, so it got deleted during the refactor.

d-torrance: Oh, ok -- I see it in the package tests now.  But it doesn't use anything from the package, so should it stay where it was?

M2/Macaulay2/packages/Permutations/Permutations.m2
```diff
@@ -0,0 +1,311 @@
...
+Permutation = new Type of BasicList
+Permutation.synonym = "permutation"
+
+isValidPermutation = method(TypicalValue => Boolean)
+isValidPermutation List := Boolean => w -> (((sort w) == toList(1..#w)) and not (#w == 0))
+
+new Permutation from BasicList := (typeofPermutation,w) -> (
+    if not isValidPermutation(w) then error(toString w | " is not a valid permutation in one-line notation.");
+    w
+)
```
mahrud, line 23: A better approach is this: don't check for correctness in the constructor, because it will slow down people's code (e.g. if they're making lots of very large permutations, and they know their permutations are definitely valid, then there's no need to check); instead change your verification method to `isWellDefined Permutation`, so people can verify that the permutation they constructed is correct.

Also, I think you might want `VisibleList` instead of `BasicList` in both places?

M2/Macaulay2/packages/Permutations/Permutations.m2
```diff
@@ -0,0 +1,311 @@
...
+
+permutation = method(TypicalValue => Permutation)
+permutation List := Permutation => w -> new Permutation from w
```
mahrud, line 26: Once you declared `TypicalValue => Permutation`, you don't need to list `Permutation => ` in front of every method.

(On the other hand, you should think about whether anyone else might want to use the method name `permutation` for a different purpose and have it return a different object, in which case fixing `TypicalValue => Permutation` would make their life harder)

M2/Macaulay2/packages/Permutations.m2
TheGrateSalmon: No, a byproduct of copying from MatrixSchubert I believe. Should be fixed in some new commits.

M2/Macaulay2/packages/Permutations.m2
TheGrateSalmon: This should be corrected in some new commits.

M2/Macaulay2/tests/normal/permutations.m2
TheGrateSalmon: I think it staying where it was makes sense as it uses M2's base functionality.

M2/Macaulay2/packages/Permutations/Permutations.m2
TheGrateSalmon: That makes sense and should be corrected in some new commits. Related, but can I also use `VisibleList` as a type anywhere I would expect a `List` and `Sequence` to behave identically? I would hope that this would mean some code could be condensed and reduce the total number of overloads for a method.

M2/Macaulay2/packages/Permutations/Permutations.m2
TheGrateSalmon: I didn't think about that side effect of `TypicalValue`, so it has been more carefully thought out and corrected in some new commits.

TheGrateSalmon:
I should also mention that this package includes code for pattern avoidance and is taken directly from the `MatrixSchubert` package. The idea is that since pattern avoidance is inherent to permutations, it'll go into `Permutations`, and following the adoption of this package, the `MatrixSchubert` package might be overhauled to use the `Permutation` type from this package for consistency.

M2/Macaulay2/packages/Permutations/Permutations.m2
mahrud: Yes, List, Sequence, Array, and AngleBarList inherit from VisibleList. On the other hands, loads of things inherit from BasicList (think Option, Command, URL, etc.)

M2/Macaulay2/packages/Permutations.m2
```diff
@@ -0,0 +1,383 @@
...
+records = method()
+records Permutation := List => w -> (
+    to1Index positions(1 .. #w, i -> all(1 ..< i, j -> w_j < w_i))
+)
+
+------------------------------------
+-- Pattern avoidance
+------------------------------------
+avoidsPattern = method(TypicalValue => Boolean)
+avoidsPattern (Permutation,List) := (w, pattern) -> (
```
mahrud, line 252: You should mention here that this code is borrowed from MatrixSchubert's `isPatternAvoiding` (and any others that may be in similar situation), and also get in touch with their authors to see how the two packages should interoperate (though that doesn't have to happen in this PR).

M2/Macaulay2/packages/Permutations/docs.m2
```diff
@@ -0,0 +1,1296 @@
...
+    avoidsPatterns
+    isCartwrightSturmfels
+    isCDG
+    isVexillary
+///
+
+-- avoidsPatterns
+doc ///
+  Key
+    avoidsPatterns
```
mahrud, line 619: Under `SeeAlso`, you can link to `"MatrixSchubert::avoidsAllPatterns"` and `"MatrixSchubert::isPatternAvoiding"`.

M2/Macaulay2/packages/Permutations.m2
```diff
@@ -0,0 +1,383 @@
...
+------------------------------------
+-- Permutation type declarations and basic constructors
+------------------------------------
+Permutation = new Type of VisibleList
+Permutation.synonym = "permutation"
+
+new Permutation from VisibleList := (typeofPermutation,w) -> w
+
+permutation = method()
+permutation List := Permutation => w -> new Permutation from w
```
mahrud, line 65: If you want, I think it would be reasonable to move `permutations`, `inversePermutation`, and (possibly?) `uniquePermutations` into this package (along with their tests and documentation), and comment that they were moved from Core. This may break some tests in other packages, in which case we will add this package as their dependency.

M2/Macaulay2/packages/Permutations.m2
```diff
@@ -0,0 +1,383 @@
...
+}
+
+-----------------------------------------------------------------------------
+-- **CODE** --
+-----------------------------------------------------------------------------
+------------------------------------
+-- Local utilities
+------------------------------------
+to1Index := w -> (w / (i -> i+1))
+to0Index := w -> (w / (i -> i-1))
```
mahrud, line 54: This is fine, but I'm curious if there was a problem with my suggestion about prepending 0 to switch to 0-index: https://github.com/Macaulay2/M2/issues/2581#issuecomment-2148157938

pzinn:
One incompatibility with the built-in permutations is that they start at 0, not 1.

mahrud:
> One incompatibility with the built-in permutations is that they start at 0, not 1.

Yes, although MatrixSchubert is also 1-indexed. I realize I'm not the main audience of this, so I don't have a strong opinion. Would it be much harder to check whether a permutation is 0 or 1 indexed in every algorithm?

M2/Macaulay2/packages/Permutations.m2
TheGrateSalmon: Ah, I didn't think to mention it in the code itself because of my comment above.

> I should also mention that this package includes code for pattern avoidance and is taken directly from the `MatrixSchubert` package. The idea is that since pattern avoidance is inherent to permutations, it'll go into `Permutations`, and following the adoption of this package, the `MatrixSchubert` package might be overhauled to use the `Permutation` type from this package for consistency.

As one of the authors myself, I will be coordinating with Ayah to manage the overhaul for `MatrixSchubert`.

M2/Macaulay2/packages/Permutations.m2
TheGrateSalmon: I think this make sense and I'll look into it.

M2/Macaulay2/packages/Permutations.m2
mahrud: Oh I didn't realize you are one of the authors. Never mind then haha.

M2/Macaulay2/packages/Permutations.m2
TheGrateSalmon: No, those helper functions seemed like the most obvious thing to do at the time I had written the bulk of this code (December 2023).

TheGrateSalmon:
> > One incompatibility with the built-in permutations is that they start at 0, not 1.
> 
> Yes, although MatrixSchubert is also 1-indexed. I realize I'm not the main audience of this, so I don't have a strong opinion. Would it be much harder to check whether a permutation is 0 or 1 indexed in every algorithm? Or maybe have easy to use functions for going forth and back?

I think a good solution would be some sort of configuration option when loading the package. So, it defaults to $1$-indexing, but the user can switch to $0$-indexing if they prefer. Something like `needsPackage("Permutations", ZeroIndex=>true)`.

d-torrance:
@TheGrateSalmon, @mahrud: What's the status of this PR?  Should we include this package in the release as it is, or is still waiting from some changes?

TheGrateSalmon:
> 1. `VisibleList_Permutation` (and `MutableList`), `Matrix_Permutation` and `Matrix^Permutation` (and `MutableMatrix`) for permuting elements of lists and columns or rows of matrices.

This is somewhat implemented in the package, though with a different syntax. I think your suggestion makes a bit more sense. 

> 2. A comparison operator `Permutation ? Permutation` based on the Bruhat order (is this available elsewhere in M2?)
> 
> 3. A method to get the poset of Bruhat order and the Hasse diagram of S_n (including maybe a way to get the subposet below or above a given permutation)

These  items would be wonderful to have. I'm not sure what the best way to display the Hasse diagram would be, however. I don't think I will get around to implementing them in the current release though.

> 4. `matrix Permutation` to get the permutation matrix (I know `permToMatrix` in `MatrixSchubert` does this, but given a `Permutation` type you could just define `matrix` on it)

This also exists as `toMatrix`, but this syntax makes a bit more sense as well, so it should be an easy change.



> @TheGrateSalmon, @mahrud: What's the status of this PR? Should we include this package in the release as it is, or is still waiting for some changes?

I should be able to incorporate some of @mahrud's suggestions this weekend. After that, from my understanding, it sounds like this package will be ready for its initial release.

d-torrance:
> These items would be wonderful to have. I'm not sure what the best way to display the Hasse diagram would be, however. I don't think I will get around to implementing them in the current release though.

There's a `hasseDiagram` method already in the [`Posets`](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Posets/html/index.html) package that you could use.

mahrud:
> There's a `hasseDiagram` method already in the [`Posets`](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Posets/html/index.html) package that you could use.

How do I get the poset of $$S_n$$ to begin with?

d-torrance:
> How do I get the poset of S n to begin with?

How's the partial order defined?  If it's easy to either list all the pairs $(\sigma,\tau)$ with $\sigma\leq\tau$ or write a function that returns true when $\sigma\leq\tau$, then the [`poset`](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Posets/html/_poset.html) method should work.

mahrud:
$S_n$ has $n!$ elements so listing all ordered pairs _works_ but definitely not the optimal solution. The Hasse diagram has far fewer edges but I don't know how many exactly.

TheGrateSalmon:
I've added @mahrud's request for `VisibleList`, though I'm not sure how to do the same for `MutableList` (mainly because I'm not sure what the expected behavior with something like `Bag` should be). I've also changed `toMatrix` to `matrix`.

> S n has n ! elements so listing all ordered pairs _works_ but definitely not the optimal solution. The Hasse diagram has far fewer edges but I don't know how many exactly.

If I am remembering correctly, the Bruhat order gives a _graded_ Hasse diagram with a unique maximal element, the longest word permutation $(n,n-1,\dots,1)$. So, I would think a breadth-first search from that element should be good enough, though this is without thinking about the problem too carefully.

mahrud:
This looks good! Could you resolve the conflict so the tests can run?

PR #3288, opened 2024-06-06
Added arm64 actions build

mahrud:

PR #3396, opened 2024-08-09
arm64 testing on macos-14

mahrud:
This reverts commit 8bd3db78d4afe5999002a835f791745773537418 from #3288.

The issue was than an example involving `MonomialSize => 16` was failing, but we couldn't reproduce it.


mahrud:
@mikestillman could you take a look at why the `Lex` example is failing?
```m2
     R = QQ[a..d, MonomialOrder => Lex];
     a^3 + a^2*b^2 + b*c
     B = QQ[a..d,MonomialOrder=>Lex,MonomialSize=>16];
     a^(2^15-1)
     C = QQ[a..d,MonomialOrder=>Lex,MonomialSize=>8];
     try a^(2^15-1) else "failed"
     a^(2^7-1)
     B = QQ[a..d,MonomialSize=>16,MonomialOrder=>{Weights => {1,2,3,4}, Lex}];
     a^2 + b+ c + b*d
```

I suspect you might need to run it in a debugger. Probably there's a very subtle bug.

mikestillman:
I hit this bug today, but I forget which compiler (might have been llvm clang).  I’ll take a look.  

I want to set up a project for testing M2 builds on Sonoma (i.e. I want different check boxes for different tries, where I can put notes in, and include thiings that don’t work.  I just tried creating a project (in macaulay2 org), but it isn’t doing what I want (I want checkmarks and ability to put in problems that arise, it seems very much not able to do this).  What do you suggest I use?  Should I just make an issue and use that?  A pull request, or draft PR might give me what I want visually, but it isn’t really a PR….

mahrud:
You can make a checkbox in markdown like this:
```
- [x] blah
- [ ] foo
- [ ] bar
```
Which gives:
- [x] blah
- [ ] foo
- [ ] bar

Once you submit the comment, you can still check/uncheck the boxes and github will save the changes.

Is this what you want?

mahrud:
There is a new failure now, this time from RunExternalM2. @d-torrance any ideas?

d-torrance:
No clue. Is there a way to get the external M2 that's failing to print its output to the build log or save it in the workflow artifacts?

d-torrance:
I've been playing around with this a bit.  Here's the code that `RunExternalM2` is running:

```m2
spin = (x,t) -> (
  startTime:=cpuTime();
	while(cpuTime()-startTime<t) do ();
	return(x);
);
```

`spin(5, 6)` is called (so it should spin for 6s of cpu time), but we call M2 with `ulimit -t 2` (so we should kill it 2s in, before it finishes spinning).  The test is failing because we're actually successfully finishing the 6s spin and M2 is exiting with a return value of 0:

```m2
-- running   check(4, "RunExternalM2")       -- running failed
 -- error log:  /var/folders/y6/nj790rtn62lfktb1sh__79hc0000gn/T/M2-72296-0/8.tmp:0:1:(3):[14]: 
 -- input code: /Users/runner/work/M2/M2/M2/Macaulay2/packages/RunExternalM2.m2:1051:5-1070:3
*** error: M2 exited with status code 1       -- 6.71609s elapsed
=============================================================================
/Users/runner/work/M2/M2/M2/Macaulay2/packages/RunExternalM2.m2:1051:5-1070:3 error:
 --      );
 --      ///<<endl<<close;
 -- 
 -- i3 : 
 --      r=runExternalM2(fn,"spin",(5,6),PreRunScript=>"ulimit -t 2");
 -- Running (ulimit -t 2 && (/Users/runner/work/M2/M2/M2/BUILD/build/usr-dist/arm64-Darwin-macOS-14.7.6/bin/M2-binary  --stop --no-debug --silent  -q  <"/var/folders/y6/nj790rtn62lfktb1sh__79hc0000gn/T/M2-72790-0/1.m2" >"/var/folders/y6/nj790rtn62lfktb1sh__79hc0000gn/T/M2-72790-0/1.out" 2>&1 ))
 -- Finished running.
 -- 
 -- i4 : assert(not(r#"return code"===0));
 -- stdio:13:6:(3): error: assertion failed
```

So maybe something about `ulimit -t` and `cpuTime` (which calls `clock_gettime(CLOCK_PROCESS_CPUTIME_ID, &t);` under the hood) is different on ARM Macs, I guess?

PR #3397, opened 2024-08-09
Check for msolve version in autotools build

d-torrance:
Since the msolve package requires version 0.7.0, we check the msolve version during `configure` in the autotools build.  If msolve exists on the system but is too old, then we build the latest msolve.

~~I also fixed a bug in the warning message that occurs in the package when msolve is too old -- we were trying to concatenate the output of `printerr` (which is null) with a string.~~ *Edit:* This bug is fixed in https://github.com/antonleykin/M2/pull/31.

### DIFF
--- a/M2/configure.ac
+++ b/M2/configure.ac
@@ -1112,10 +1112,21 @@ then AC_MSG_RESULT([no, will build])
 fi
 
 AS_IF([test $BUILD_msolve = no],
-    [AC_MSG_CHECKING(whether msolve is installed)
+    [AC_MSG_CHECKING([whether msolve is installed])
      AS_IF([command -v msolve > /dev/null],
-         [AC_MSG_RESULT([yes])
-          FILE_PREREQS="$FILE_PREREQS `command -v msolve`"],
+         [AS_IF([msolve -V > /dev/null 2>&1],
+             [msolve_actual_version=$(msolve -V)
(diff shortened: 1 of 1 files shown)


M2/Macaulay2/packages/Msolve.m2
```diff
@@ -44,7 +44,7 @@ msolveProgram = findProgram("msolve", "msolve --help",
 if msolveProgram === null then (
     if not (options currentPackage).OptionalComponentsPresent
```
d-torrance, line 45: I don't think this will ever be true now that `OptionalComponentsPresent` stuff was removed.

mahrud:
Did you see the other msolve PR (and my PR on Anton's branch)?

d-torrance:
I did see #3392 but hadn't noticed https://github.com/antonleykin/M2/pull/31.  It has an equivalent fix for the `printerr` issue, so I'll remove that commit from this PR.

M2/Macaulay2/packages/Msolve.m2
mahrud: Wait, we haven't merged that PR yet!

mahrud: Oh I see it was removed from this package's headers. Yeah you're right

mahrud (approved):

PR #3400, opened 2024-08-10
Fixed quotient and modulus for non-integer domains

### LINKED ISSUE #3340: quotient and modulus for non-integers are wrong
Unless I'm way off about what we want `//` and `%` to mean, I think these are all wrong:
https://github.com/Macaulay2/M2/blob/ec9e9ac60ed4a8e791448942202f077a88a87e15/M2/Macaulay2/m2/reals.m2#L213-L226

Some examples: (afterprints are silenced)
```m2
i1 : 13.3 // 4.0

o1 = 3.325

i2 : 13.3 % 4.0

o2 = 0

i3 : 13.3 // 4

o3 = 3.325

i4 : 13.3 % 4

o4 = 0

i5 : 13.3 // (4/1)

o5 = 3.325

i6 : 13.3 % (4/1)

o6 = 0

i7 : (pi*ii) // 3.0

o7 = 1.0471975511966*ii

i8 : (pi*ii) % 3.0

o8 = 0
```
What's the point of having `//` be a synonym for `/`, and why even define `%` for non-integers if the answer is always zero anyway?!

mahrud:
Closes #3340.

### DIFF
--- a/M2/Macaulay2/packages/WeylGroups.m2
+++ b/M2/Macaulay2/packages/WeylGroups.m2
@@ -539,19 +539,19 @@ scalarProduct = method()
 --the scalar product of the i-th fundamental weight with the j-th
 scalarProduct(RootSystem,ZZ,ZZ) := (R,i,j) -> 
 	(
-	((R.CartanMatrixTrInv)_(j-1,i-1))*(R.RootNorms#(j-1))//2
+	((R.CartanMatrixTrInv)_(j-1,i-1))*(R.RootNorms#(j-1)) / 2
 	)
 
 --the scalar product of two general weights
 scalarProduct(RootSystem,Weight,Weight) := (R,u,v) -> 
 	(
-	sum apply(apply(entries((R.CartanMatrixTrInv)*u),entries(v),(x,y)->x*y),R.RootNorms,(x,y)->x*y)//2
+	sum apply(apply(entries((R.CartanMatrixTrInv)*u),entries(v),(x,y)->x*y),R.RootNorms,(x,y)->x*y) / 2
 	)
 
 --the scalar product of the i-th fundamental weight with a general weight
 scalarProduct(RootSystem,ZZ,Weight) := (R,i,v) -> 
 	(
-	(((R.CartanMatrixTrInv)*v)_(i-1))*(R.RootNorms#(i-1))//2
+	(((R.CartanMatrixTrInv)*v)_(i-1))*(R.RootNorms#(i-1)) / 2
 	)
 
 --the dual of a root evaluated at something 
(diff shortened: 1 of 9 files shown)


M2/Macaulay2/packages/WeylGroups.m2
```diff
@@ -545,7 +545,7 @@ scalarProduct(RootSystem,ZZ,ZZ) := (R,i,j) ->
 --the scalar product of two general weights
 scalarProduct(RootSystem,Weight,Weight) := (R,u,v) -> 
 	(
-	sum apply(apply(entries((R.CartanMatrixTrInv)*u),entries(v),(x,y)->x*y),R.RootNorms,(x,y)->x*y)//2
+	sum apply(apply(entries((R.CartanMatrixTrInv)*u),entries(v),(x,y)->x*y),R.RootNorms,(x,y)->x*y) / 2
```
mahrud, line 548: @baptistecalmes could you take a look at this change, please?

I'm trying to clarify the difference between `QQ // ZZ` and `QQ / ZZ` in Macaulay2.
For instance, until now `(35/2) // 2` gave `35/4`, but to make `//` behave uniformly, I'd like to change it to `(35/2) // 2 = floor(32/4) = 8`. This changed caused an example in the package `HighestWeights` to fail, and I tracked it down to this division here.

Should this division result in a rational or integer number? Was there a reason `//` was used rather than `/`? Also, there are a couple of more uses of `// 2` a few lines above and below here. Should those also be changed to `/ 2`?

mahrud: Actually, looking at the following tests, I think the rational number result was intended in all three cases, so I'll go ahead and change those as well.
```m2
assert(scalarProduct(rootSystemA(2),1,2)==1/6)

R=rootSystemA(2);
v=weight(R,{1,2});
assert(scalarProduct(R,1,v)==2/3)

R=rootSystemA(2)
u=weight(R,{1,2})
v=weight(R,{0,3})
assert(scalarProduct(R,u,v)==5/2)
```
I appreciate your thorough test coverage!

baptistecalmes: @mahrud I'm sure the coherent behavior of // that you are trying to reach has been discussed among core developers of M2, but I personally do not quite get what it's supposed to be outside of Euclidean rings or maybe slight generalization of them. In particular, it's unclear to me that QQ // QQ should return anything at all, and if it does, I assume it should be an integer. Then, I do not see how general this behavior should be. What should be the output ring when // is defined on some ring?

Relatedly, concerning your changes to WeylGroups, could you please add test cases for various definitions of eval ensuring that eval always returns an integer, just in case // is modified further in the smoothing process, since you've removed the floor function in the code that used to ensure that.

mahrud: I initiated the discussion about this in #3340.

I will add more tests for `eval`!

mahrud: Actually, it might be simpler to leave floor there as is.

PR #3402, opened 2024-08-10
Update PositivityToricBundles to v1.9

a-hochenegger:
Added the method `wellformedBundleFiltrations`, updated documentation accordingly and added a corresponding test. 
Several methods make the implicit assumption that the filtration steps are stored in an ascending manner. This is not true for bundles which are created by the methods `dual` or `tensor` (maybe others) from the package `ToricVectorBundles`. The method `wellformedBundleFiltrations` ensures that the filtration steps are stored in a well-formed manner, suitable for the methods of this package.


M2/Macaulay2/packages/PositivityToricBundles.m2
```diff
@@ -480,7 +486,8 @@ if separatesJets(tvb, Verbosity=>opts#Verbosity) >= 0 then true else false;
 --           check if the vector bundle is very ample, using [RJS, Cor. 6.7]
 --   INPUT : 'tvb', toric vector bundle
 --  OUTPUT :  'true' if very ample, otherwise 'false'
-isVeryAmple ToricVectorBundleKlyachko := true >> opts -> tvb -> if separatesJets(tvb, opts) >= 1 then true else false
+--isVeryAmple = method( Options => true ) -- conflicts with Polyhedra
+isVeryAmple (ToricVectorBundleKlyachko) := (tvb) -> if separatesJets(tvb) >= 1 then true else false;
```
d-torrance, line 490: I think the failing builds are due to this `isVeryAmple` change.  It's a `MethodFunctionWithOptions`, so it should accept options.

a-hochenegger: Thanks for the hint, I hope it's fixed now.

M2/Macaulay2/packages/PositivityToricBundles.m2
```diff
@@ -396,6 +397,10 @@ graphToricChernCharacter (ToricVectorBundleKlyachko) := {Verbosity => 0} >> opts
 -- METHOD: separateJets
 ------------------------------------------------------------------------------
 
+-- hT .. hashTable
+-- fun .. function that takes key and value as input and returns true or false
+--selectPairs = (hT, fun) -> hashTable select(pairs hT, (k,v) -> fun(k,v) )
```
d-torrance, line 402: Is there a reason to re-introduce this `selectPairs` comment?  It was removed when `selectPairs` was added to Core in #3305.

M2/Macaulay2/packages/PositivityToricBundles.m2
```diff
@@ -486,8 +482,7 @@ if separatesJets(tvb, Verbosity=>opts#Verbosity) >= 0 then true else false;
 --           check if the vector bundle is very ample, using [RJS, Cor. 6.7]
 --   INPUT : 'tvb', toric vector bundle
 --  OUTPUT :  'true' if very ample, otherwise 'false'
---isVeryAmple = method( Options => true ) -- conflicts with Polyhedra
---isVeryAmple (ToricVectorBundleKlyachko) := (tvb) -> if separatesJets(tvb) >= 1 then true else false;
+--isVeryAmple = method( Options => true ) -- already defined in Polyhedra
 isVeryAmple (ToricVectorBundleKlyachko) := {Verbosity => 0} >> opts -> tvb -> if separatesJets(tvb, Verbosity=>opts#Verbosity) >= 1 then true else false
```
mahrud, line 486: This is fine, but just so you know, this also works:
```suggestion
isVeryAmple (ToricVectorBundleKlyachko) := {Verbosity => 0} >> opts -> tvb -> separatesJets(tvb, opts) >= 1
```

Also, it might be better to check the global variable `debugLevel` for printing extra information rather than adding a `Verbosity` option.

a-hochenegger: Thanks for the hint! I will think about using `debugLevel` and probably implement it (hopefully soon).

a-hochenegger: May I ask a question about using `debugLevel`? 
One reason for using the `Verbosity` option is that many methods (more complicated than the `isVeryAmple` above) pass `Verbosity-1` to invoked other methods. So the value of `Verbosity` decides how far you want to follow into the rabbit hole of nested methods. 
It seems to me that this behaviour is not possible with `debugLevel` alone, or is there some counter of the level of depth? It seems to me that `interpreterDepth` or `recursionDepth()` count something else...

mahrud: Well, in a high level method you can check `if debugLevel > 0`, on nested methods you can check `if debugLevel > 1`, and on really deep ones you could even do something like `if debugLevel == 12` or something like that.

The benefit is that suppose someone wrote a method using `isVeryAmple` but didn't use the `Verbosity` option. Then there would be no way to get debugging info from it.

a-hochenegger: Thanks for the answer! I guess, I will change at some point to use `debugLevel` (but not for this pull request).

PR #3403, opened 2024-08-11
Added persistent history using readline

### LINKED ISSUE #2819: Persistent history and shortcuts using libreadline/libedit
We should enable persistent history across M2 instances run from a terminal. This is pretty common in repl shells and makes many lives easier.

Related: https://github.com/Macaulay2/M2/issues/825

---

Related, but separate enhancement:

When using a terminal, or M2 inside a terminal, you can go back in command history with up/down arrow keys, but also Ctrl + P and Ctrl + N (and Ctrl + R for reverse search).

On Emacs, only Ctrl + arrow keys work, which in particular conflicts with a macOS shortcut Ctrl + Up (Ctrl + R works for different reasons).

Meta + P/N keys also work in Emacs, but they don't work in the terminal. The discrepancy makes it hard to tell users to just remember one shortcut.

We could easily enable Ctrl + P/N in Emacs to instead go through history, but as Doug points out below, that's not a great idea. Instead, we should enable scrolling through history with Meta + P/N when M2 is being used within terminal.

mahrud:
This adds persistent history for M2 running in terminal.

- **moved historyGet to getHistory in interp.dd**
- **moved rl_catch_signals to initReadlineVariables**
- **added persistent history using libreadline**
- **simplified getting libreadline's version**

There are a few kinks to be resolved, but I think it's fine for now. Closes #2819.

### DIFF
--- a/M2/Macaulay2/d/interp.dd
+++ b/M2/Macaulay2/d/interp.dd
@@ -669,6 +669,22 @@ getHistory(e:Expr):Expr := (
     else WrongArgZZ());
 setupfun("getHistory", getHistory);
 
+readHistory(e:Expr):Expr := (
+    when e is filename:stringCell
+    do toExpr(readHistory(tocharstar(expandFileName(filename.v))))
+    else WrongArgString(1));
+setupfun("readHistory", readHistory);
+
+appendHistory(e:Expr):Expr := (
+    when e is s:Sequence do if length(s) != 2 then WrongNumArgs(2) else
+    when s.0 is n:ZZcell do if !isInt(n) then WrongArgSmallInteger(1) else
+    when s.1 is filename:stringCell
+    do toExpr(appendHistory(toInt(n), tocharstar(expandFileName(filename.v))))
+    else WrongArgString(2)
+    else WrongArgZZ(1)
+    else WrongNumArgs(2));
+setupfun("appendHistory", appendHistory);
+
 -- Local Variables:
 -- compile-command: "echo \"make: Entering directory \\`$M2BUILDDIR/Macaulay2/d'\" && make -C $M2BUILDDIR/Macaulay2/d interp.o "
 -- End:
--- a/M2/Macaulay2/d/interp.dd
+++ b/M2/Macaulay2/d/interp.dd
@@ -662,6 +662,13 @@ runBasicTests(e:Expr):Expr := (
      else WrongNumArgs(0));
 setupfun("runBasicTests",runBasicTests);
 
+getHistory(e:Expr):Expr := (
+    when e is n:ZZcell do if isInt(n)
+    then toExpr(tostring(getHistory(toInt(n))))
+    else WrongArgSmallInteger()
+    else WrongArgZZ());
+setupfun("getHistory", getHistory);
+
 -- Local Variables:
 -- compile-command: "echo \"make: Entering directory \\`$M2BUILDDIR/Macaulay2/d'\" && make -C $M2BUILDDIR/Macaulay2/d interp.o "
 -- End:
--- a/M2/Macaulay2/m2/Core.m2
+++ b/M2/Macaulay2/m2/Core.m2
@@ -119,16 +119,16 @@ needs = filename -> if not filesLoaded#?filename then load filename else (
 -----------------------------------------------------------------------------
 -- Setup persistent history
 -----------------------------------------------------------------------------
-historyFilename = "~/.Macaulay2-history"
+historyFilename = "history.m2"
 
-if not gotarg "--no-readline" then (
-    addStartFunction(() -> readHistory historyFilename);
+if not noinitfile and not gotarg "--no-readline" then (
+    addStartFunction(() -> readHistory(applicationDirectory() | historyFilename));
     -- TODO: find a better alternative to addEndFunction, because
     -- exiting with Ctrl+D duplicates the last line of history file,
     -- but if we use lineNumber-1, then exit and restart miss the first
     -- entry from the current session. Moreover, maybe we shouldn't save
     -- exit and restart in the history anyway.
-    addEndFunction(() -> appendHistory(lineNumber, historyFilename));
+    addEndFunction(() -> appendHistory(lineNumber, applicationDirectory() | historyFilename));
     )
 
 -----------------------------------------------------------------------------
--- a/M2/Macaulay2/m2/Core.m2
+++ b/M2/Macaulay2/m2/Core.m2
@@ -117,6 +117,21 @@ needs = filename -> if not filesLoaded#?filename then load filename else (
      if filetime < fileTime filepath then load filepath)
 
 -----------------------------------------------------------------------------
+-- Setup persistent history
+-----------------------------------------------------------------------------
+historyFilename = "~/.Macaulay2-history"
+
+if not gotarg "--no-readline" then (
+    addStartFunction(() -> readHistory historyFilename);
+    -- TODO: find a better alternative to addEndFunction, because
+    -- exiting with Ctrl+D duplicates the last line of history file,
+    -- but if we use lineNumber-1, then exit and restart miss the first
+    -- entry from the current session. Moreover, maybe we shouldn't save
+    -- exit and restart in the history anyway.
+    addEndFunction(() -> appendHistory(lineNumber, historyFilename));
+    )
+
+-----------------------------------------------------------------------------
 -- Load the rest of Core
 -----------------------------------------------------------------------------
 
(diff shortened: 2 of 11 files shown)


M2/Macaulay2/d/interp.dd
```diff
@@ -662,6 +662,13 @@ runBasicTests(e:Expr):Expr := (
      else WrongNumArgs(0));
 setupfun("runBasicTests",runBasicTests);
 
+getHistory(e:Expr):Expr := (
+    when e is n:ZZcell do if isInt(n)
+    then toExpr(tostring(getHistory(toInt(n))))
+    else WrongArgSmallInteger()
+    else WrongArgZZ());
+setupfun("getHistory", getHistory);
```
d-torrance, line 670: Thank you for renaming this!  I've regretted my decision to call it `historyGet` ever since I added it.  I think I was just trying to mimic `history_get`, but `getHistory` makes way more sense.

mahrud: readline developers had no sense of style ...

M2/Macaulay2/d/interp.dd
```diff
@@ -669,6 +669,22 @@ getHistory(e:Expr):Expr := (
...
+    when e is filename:stringCell
+    do toExpr(readHistory(tocharstar(expandFileName(filename.v))))
+    else WrongArgString(1));
+setupfun("readHistory", readHistory);
+
+appendHistory(e:Expr):Expr := (
+    when e is s:Sequence do if length(s) != 2 then WrongNumArgs(2) else
+    when s.0 is n:ZZcell do if !isInt(n) then WrongArgSmallInteger(1) else
+    when s.1 is filename:stringCell
+    do toExpr(appendHistory(toInt(n), tocharstar(expandFileName(filename.v))))
```
d-torrance, line 682: Should we raise an error if `appendHistory` has a nonzero return value?

M2/Macaulay2/m2/Core.m2
```diff
@@ -116,6 +116,21 @@ needs = filename -> if not filesLoaded#?filename then load filename else (
...
+-----------------------------------------------------------------------------
+historyFilename = "~/.Macaulay2-history"
+
+if not gotarg "--no-readline" then (
+    addStartFunction(() -> readHistory historyFilename);
+    -- TODO: find a better alternative to addEndFunction, because
+    -- exiting with Ctrl+D duplicates the last line of history file,
+    -- but if we use lineNumber-1, then exit and restart miss the first
+    -- entry from the current session. Moreover, maybe we shouldn't save
+    -- exit and restart in the history anyway.
```
d-torrance, line 130: For consistency with Emacs, I think it would make sense to save `exit` and `restart`.

M2/Macaulay2/d/interp.dd
mahrud: Maybe, but not here. This function is returning the return code intact. We _could_ check it in `Core.m2` and signal a warning, but this wasn't a priority for me.

M2/Macaulay2/m2/Core.m2
mahrud: Sure, that's fine, but I'd like to figure out a way to detect when we exit by Ctrl-D. Python, Julia, etc. handle this fine (and save "exit"), so I'm sure there's a way. Maybe `lineNumber` is not the right counter to use, or maybe it should be incremented by `exit` and `restart` and `end` before they evaluate?

M2/Macaulay2/d/interp.dd
d-torrance: One advantage of checking it here is we could call `strerror` to get an more informative error message (since it returns `errno` when there's an error.)

mahrud: Does `libhistory` somehow store the meaning of the error somewhere for `strerror` to access? I didn't see anything about this in the documentation!

d-torrance: Looking at the [source](https://sources.debian.org/src/readline/8.2-4/histfile.c/), I think it just passes along the value of `errno` after it's been set by one the standard library functions it calls.  So it will just be one of the various standard library error messages.

mahrud: Not sure where exactly to look in there. I'm not very familiar, but if you think it would be helpful and not depend on the system (or different versions of readline on mac, etc) feel free to make changes.

M2/Macaulay2/m2/Core.m2
```diff
@@ -116,6 +116,21 @@ needs = filename -> if not filesLoaded#?filename then load filename else (
      (filepath, filetime) := filesLoaded#filename;
      if filetime < fileTime filepath then load filepath)
 
+-----------------------------------------------------------------------------
+-- Setup persistent history
+-----------------------------------------------------------------------------
+historyFilename = "~/.Macaulay2-history"
```
d-torrance, line 122: Perhaps this could be configurable in some way?  For example, Python 3.13 adds support for an environment variable (`PYTHON_HISTORY`) that users can use to set the path of Python's equivalent history file.

M2/Macaulay2/d/interp.dd
d-torrance: I just pushed few commits dealing with this.  The first raises an error if `appendHistory` (or `readHistory`) fails.  That was in turn raising errors if we tried reading or appending to a nonexistent history file, so I added a couple more commits dealing with those cases.

d-torrance: So for example:

```m2
i2 : readHistory "foo"
stdio:2:1:(3): error: readHistory: No such file or directory

i3 : appendHistory(0, "foo")
stdio:3:1:(3): error: appendHistory: No such file or directory
```

M2/Macaulay2/m2/Core.m2
mahrud: Ehh, we have a lot of other things that aren't configurable, I don't see the point of starting with this one. But maybe `.Macaulay2/history.m2` would make more sense.

M2/Macaulay2/m2/Core.m2
```diff
@@ -116,6 +116,28 @@ needs = filename -> if not filesLoaded#?filename then load filename else (
...
+    -- exiting with Ctrl+D duplicates the last line of history file,
+    -- but if we use lineNumber-1, then exit and restart miss the first
+    -- entry from the current session. Moreover, maybe we shouldn't save
+    -- exit and restart in the history anyway.
+    addEndFunction(() -> (
+	    if not fileExists historyFilename
+	    then (
+		dir := directoryName historyFilename;
+		if not fileExists dir then makeDirectory dir;
+		historyFilename << close);
```
mahrud, line 137: I don't like the added complexity here. These falls were silently failing before anyway, and if a someone used one of these (who?!) and actually cared about error checking they could have checked the integer that they return for the error. I would rather revert.

M2/Macaulay2/m2/Core.m2
d-torrance: That's true.  I do like the idea of putting it in `applicationDirectory()`, since that is configurable by changing `applicationDirectorySuffix`.

M2/Macaulay2/m2/Core.m2
d-torrance: Fair enough.  I added a lot of this with my environment variable idea in mind, but we wouldn't need to check if the directory exists if it's in either `$HOME` or `applicationDirectory()` since we already know they exist.

d-torrance: I've reverted the three commits I added, but there should probably still be something that creates `historyFilename` if it doesn't exist before calling `appendHistory`.  Otherwise the history isn't saved.

mahrud: Hmm, I thought this was working when I tested it.
I just added a commit that creates the file and also adds an initial like which users might run into and learn about where the file is stored:
```m2
-- This is the beginning of your Macaulay2 log stored at /home/mahrud/.Macaulay2/history.m2
```

M2/Macaulay2/m2/system.m2
```diff
@@ -112,6 +112,10 @@ setUpApplicationDirectory = () -> (
      f := (n,c) -> (n = dir|n; if not fileExists n then n << c << close);
      f("init.m2", sampleInitFile);
      f("README", readmeFile);
+     f(historyFilename,
+	 concatenate(
+	     "-- This is the beginning of your Macaulay2 log stored at ",
+	     historyFile, dir, historyFilename, newline));
```
d-torrance, line 118: What is `historyFile`?  Should this line just be `dir, historyFilename, newline));`?

mahrud: Opps typo.

d-torrance (approved):

d-torrance:
Is this ready to merge, or would you like to figure out the Ctrl-D stuff first?

mahrud:
I'm happy to merge this now and figure that out later.

PR #3404, opened 2024-08-12
Moved git state from version.dd to version.m2

### LINKED ISSUE #3250: version-tmp.cc recompiles after every commit
I realized a major drawback of including anything more than just the version in the binary is that any change in git will cause M2-binary to recompile. It's not urgent to change this, but I think a possible solution is to move the most frequently changed numbers into a non-compiled file (e.g. a variable set in an m2 file).

mahrud:
Finally closes #3250, which has been _such_ a pain ...

The main difficulty was rearranging startup.m2 to:
1. set prefixDirectory
2. set currentLayout
3. and based on those find the Core directory where version.m2 lives

all before the version is printed.

If `version.m2` is not found, for whatever reason, we get:
```m2
Macaulay2, version 1.24.05-bin (version.m2 not found)
```
The "-bin" indicates that this is the version stored in the binary only.

### DIFF
--- a/M2/Macaulay2/d/CMakeLists.txt
+++ b/M2/Macaulay2/d/CMakeLists.txt
@@ -134,11 +134,6 @@ foreach(_d_source IN LISTS DLIST)
   list(APPEND CLIST ${_c_source})
 endforeach()
 
-# To avoid unnecessary recompiling, only version.dd is given GIT_DESCRIPTION
-set_property(SOURCE version-tmp.cc APPEND PROPERTY COMPILE_DEFINITIONS
-  GIT_DESCRIPTION="${GIT_DESCRIPTION}"
-  GIT_BRANCH="${GIT_BRANCH}")
-
 ###############################################################################
 ## Compile the interpreter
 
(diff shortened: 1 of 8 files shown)


mahrud:
I hope the last commit fixed the autotools issue, but if it didn't work, @d-torrance do you mind looking into what's missing?

d-torrance:
Figured it out!  We need to use `AC_SUBST` instead of `AC_DEFINE` now.

mahrud:
Apparently neither cmake nor autotools have been detecting the git description and branch correctly on development when building on GitHub. I don't know where this started, but on autotools we've just been getting:
```m2
configure: configuring Macaulay2 version 1.24.05
...
configure: with  GIT_DESCRIPTION   = release-1.24.05
configure: with  GIT_BRANCH        = 
```
And on cmake:
```m2
     M2 version        = 1.24.05
     Git description   = 
     Git branch        = 
```

mahrud:
It seems like it's because github uses a shallow checkout, which doesn't pull tags. This makes sense and I don't necessarily think we should change it (though for future reference, here is how one would change it: https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches)

d-torrance (approved):

PR #3405, opened 2024-08-12
Use code() instead of help() for "methods" test

d-torrance:
We just want something that returns a `DIV`, and calling `help()` during a test can run out of memory on some systems.

See https://github.com/Macaulay2/M2/pull/3382#discussion_r1713098508


mahrud (approved):

mahrud:
Honestly, no need to run the whole build over this (famous last words...)

PR #3414, opened 2024-08-18
Remove unused semicolonfun (; is not a binary operator)

d-torrance:
This would only be used if `x ; y` was converted into a `binaryCode` object, but it's converted into a `semiCode` object instead.


mahrud (approved):
Almost want to say we should keep this as an archaeological clue to Macaulay2's pre-(git)-history!

Issue #3415, opened 2024-08-18
Incorrect code location for functions defined in standard input


d-torrance:
Since #3403:

```m2
i1 : f = x -> x^2

o1 = f

o1 : FunctionClosure

i2 : code f

o2 = stdio:1:6-1:11: --source code:
     -- This is the beginning of your Macaulay2 log stored at /home/profzoom/.Macaulay2/history.m2
```

I suppose the solution would be to keep track of how long the history is when we start M2 and add that to the line number when finding the code.

mahrud:
Ah yes I should have noticed this from the getHistory lines in code.m2.

mahrud:
This is actually more complicated than I expected.
I added a call to `where_history` in this [commit](https://github.com/mahrud/M2/commit/9eed58c5d9c5ad317c672f6223cba8243fd1a42e), but:
1. it seems to return zero if it's called before the first prompt, even if it's after `readHistory` apparently!
2. it seems to be off by an inconsistent number (up to 4 even) from the right number to get `code` working in your example.

I'm not sure if I have time to debug this. We can disable history if this is too complicated to resolve.

d-torrance:
I think I figured it out -- see #3416.

d-torrance:
Closed by #3416

PR #3416, opened 2024-08-18
Fix code location when reading history from a file

d-torrance:
At startup, we use a new compiled function (`historyLength()`, which returns the value of readline's `history_length` variable) to determine how many lines of history we've saved from previous sessions.  We add this value to the line numbers of any code from `stdio` when calling `code` so we get the correct lines back when calling `getHistory`.

### Before
```m2
i1 : f = x -> x^2;

i2 : code f

o2 = stdio:1:6-1:11: --source code:
     -- This is the beginning of your Macaulay2 log stored at /home/profzoom/.Macaulay2/history.m2
```

### After
```m2
i1 : f = x -> x^2;

i2 : code f

o2 = stdio:1:4-1:12: --source code:
     f = x -> x^2;
```

Closes: #3415

### DIFF
--- a/M2/Macaulay2/d/M2lib.c
+++ b/M2/Macaulay2/d/M2lib.c
@@ -143,6 +143,8 @@ char *system_getHistory(const int n)
   return NULL;
 }
 
+int system_historyLength() { return history_length; }
+
 void system_initReadlineVariables(void) {
   static char readline_name[] = "M2";
   static char basic_word_break_characters[] = "!\"#$%&'()*+,-./:;<=>?@[\\]^_`{|}~ \t\n\r";
--- a/M2/Macaulay2/m2/Core.m2
+++ b/M2/Macaulay2/m2/Core.m2
@@ -123,9 +123,8 @@ historyFilename = "history.m2"
 historyOffset = 0;
 
 if not noinitfile and not gotarg "--no-readline" then (
-    addStartFunction(() -> (
-	    readHistory(applicationDirectory() | historyFilename);
-	    historyOffset = historyLength()));
+    readHistory(applicationDirectory() | historyFilename);
+    historyOffset = historyLength();
     -- TODO: find a better alternative to addEndFunction, because
     -- exiting with Ctrl+D duplicates the last line of history file,
     -- but if we use lineNumber-1, then exit and restart miss the first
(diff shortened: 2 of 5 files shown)


M2/Macaulay2/d/M2lib.c
```diff
@@ -143,6 +143,8 @@ char *system_getHistory(const int n)
   return NULL;
 }
 
+int system_historyLength() { return history_length; }
```
mahrud, line 146: Is it possible to directly export this variable to the top-level?

mahrud:
Do we have a mechanism to test this? Not sure how exactly to test readline ...

mahrud (approved):
Two possible improvements:
- export history length directly, not sure how easy this is
- add readline tests to make sure this doesn't break again

M2/Macaulay2/d/M2lib.c
d-torrance: That was my first thought, too.  But I couldn't figure it out -- if we do something like `setupvar("historyLength", toExpr(Ccode(int, "history_length")))`, then it will just be 0 since it gets assigned at startup and never updates.

mahrud: Maybe it needs to be a pointer?

d-torrance: But we'd still need to dereference it when we need its value, so we'd have to call some kind of function.  I suppose we could add some variation of `setupvar` that does this, but I'm not sure if it's worth it.

M2/Macaulay2/m2/Core.m2
```diff
@@ -120,9 +120,12 @@ needs = filename -> if not filesLoaded#?filename then load filename else (
 -- Setup persistent history
 -----------------------------------------------------------------------------
 historyFilename = "history.m2"
+historyOffset = 0;
 
 if not noinitfile and not gotarg "--no-readline" then (
-    addStartFunction(() -> readHistory(applicationDirectory() | historyFilename));
+    addStartFunction(() -> (
+	    readHistory(applicationDirectory() | historyFilename);
+	    historyOffset = historyLength()));
```
mahrud, line 128: I'm actually blanking on why `addStartFunction` was needed here. Why not just do this when `Core` is loading?

d-torrance: Yeah, it works just fine without `addStartFunction`.  I just pushed a new commit removing it.

mahrud (approved):
Feel free to merge once tests pass.

PR #3419, opened 2024-08-20
Use a list instead of a hash table for storing package tests

d-torrance:
This way, `tests` will work like other similar functions like `methods` and `about`.

As suggested by @mahrud in https://github.com/Macaulay2/M2/pull/3418#issuecomment-2298982452.

### Before

```m2
i1 : tests "FirstPackage"

o1 = HashTable{0 => TestInput[/usr/share/Macaulay2/FirstPackage.m2:54:1-57:1]}

o1 : HashTable
```

### After

```m2
i1 : tests "FirstPackage"

o1 = {0 => (TestInput[/home/dtorrance/src/macaulay2/M2/M2/Macaulay2/packages/FirstPackage.m2:54:1-57:1])}

o1 : NumberedVerticalList
```

Another neat consequence that we get for free is negative indices work:

```m2
i2 : tests(-1, "FirstPackage")

o2 = TestInput[/home/dtorrance/src/macaulay2/M2/M2/Macaulay2/packages/FirstPackage.m2:54:1-57:1]

o2 : TestInput
```


mahrud (changes requested):
Why doesn't `code tests Saturation` work?

The only other requested change is the one in the documentation, but here are several other desiderata:
1. the printing of `TestInput` could use simplifying:
   - Why is it in parenthesis inside `NumberedVerticalList`?
   - The full path to the test seems unnecessary, can we cap it to path from `packages` directory? e.g. `TestInput[FirstPackage.m2:54:1-57:1]`. Full location can be queried with `locate N` or `locate tests Saturation`.
2. I'd expect `capture tests Saturation` to work.
3. Why are `code first tests Saturation` and `code locate first tests Saturation` are so different? I understand that the former is the string that we will eventually run, but even the location line is different, and the extra lines before and after look weird. Maybe `toString TestInput` should give the actual string so `code` can just show the exact lines in the file?
```m2
i5 : code locate first tests Saturation

o5 = ../../Macaulay2/packages/Saturation/quotient-test.m2:1:1-11:1: --source code:
     TEST ///
       R  = ZZ[x, y];
       I  = ideal(6*x^3, 9*x*y, 8*y^2);
       J1 = ideal(-3, x^2);
       J2 = ideal(4*y);
       assert(intersect(I:J1, I:J2) == ideal(8*y^2, 3*x*y, x*y^2, 6*x^3))
       assert(I : (J1+J2) == ideal(8*y^2, 3*x*y, 6*x^3))
       for strategy in {Quotient, Iterate} do
       assert(quotient(I, J1, Strategy => strategy) == ideal(8*y^2, 3*x*y, x*y^2, 6*x^3))
     ///


i6 : code first tests Saturation

o6 =  -- /home/mahrud/Projects/M2/quickfix/M2/Macaulay2/packages/Saturation/quotient-test.m2:11:1: location of test code

       R  = ZZ[x, y];
       I  = ideal(6*x^3, 9*x*y, 8*y^2);
       J1 = ideal(-3, x^2);
       J2 = ideal(4*y);
       assert(intersect(I:J1, I:J2) == ideal(8*y^2, 3*x*y, x*y^2, 6*x^3))
       assert(I : (J1+J2) == ideal(8*y^2, 3*x*y, 6*x^3))
       for strategy in {Quotient, Iterate} do
       assert(quotient(I, J1, Strategy => strategy) == ideal(8*y^2, 3*x*y, x*y^2, 6*x^3))


i7 :
```

M2/Macaulay2/packages/Macaulay2Doc/functions/tests-doc.m2
```diff
@@ -17,11 +17,11 @@ doc ///
     i:ZZ
     pkg:{Package, String}
   Outputs
-    :{HashTable, TestInput}
+    :{NumberedVerticalList, TestInput}
   Description
     Text
-      Returns @ofClass HashTable@ containing the tests for the given
-      package.  Each key of this hash table is an integer, which would
+      Returns @ofClass NumberedVerticalList@ containing the tests for the given
```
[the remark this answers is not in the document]
d-torrance: I've updated the docs accordingly.

d-torrance:
> Why doesn't `code tests Saturation` work?

It works now!

> The only other requested change is the one in the documentation, but here are several other desiderata:
> 
> 1. The printing of `TestInput` could use simplifying:
>    
>    * Why is it in parenthesis inside `NumberedVerticalList`?

That was because `precedence(TestInput)` wasn't defined.  Adding that removed the parentheses!

>    * The full path to the test seems unnecessary, can we cap it to path from `packages` directory? e.g. `TestInput[FirstPackage.m2:54:1-57:1]`. Full location can be queried with `locate N` or `locate tests Saturation`.

I'd rather stick with the current behavior for Emacs clickability.

> 2. I'd expect `capture tests Saturation` to work.

I've added `capture(TestInput)`, so one test can now be captured at a time.  `capture(List)` expects a list of strings, so it would need to be rewritten to allow `TestInput` objects.

> 3. Why are `code first tests Saturation` and `code locate first tests Saturation` are so different? I understand that the former is the string that we will eventually run, but even the location line is different, and the extra lines before and after look weird. Maybe `toString TestInput` should give the actual string so `code` can just show the exact lines in the file?

`toString(TestInput)` now works as you suggested, and `code(TestInput)` now returns a `DIV` like all the other `code` methods (it's just `code @@ locate` now).

M2/Macaulay2/m2/testing.m2
```diff
@@ -35,13 +23,21 @@ editMethod TestInput := editMethod @@ locate
...
 TEST String := opts -> teststring -> (
-    n := currentPackage#"test number";
-    currentPackage#"test inputs"#n = TestInput if opts.FileName then (
-        testCode := get teststring;
-        (minimizeFilename teststring, depth net testCode + 1, testCode)
-        ) else
-        (minimizeFilename currentFileName, currentRowNumber(), teststring);
-    currentPackage#"test number" = n + 1;)
+    n := #currentPackage#"test inputs";
+    if opts.FileName then (
```
mahrud, line 27: Is this option ever used? I have seen `TEST get "test.m2"` but never this option. Wouldn't be opposed to deprecating it for simplicity.

mahrud: Oooh actually this seems like a bug:
```m2
i1 : TEST get "../../Macaulay2/packages/Macaulay2Doc/demo1.m2"

i2 : tests User

o2 = {0 => TestInput[stdio:-13:1-1:80]}

i3 : locate oo

o3 = {0 => (stdio:-13:1-1:80)}

i4 : code oo

o4 = stdio:-13:1-1:80: --source code:














     -- This is the beginning of your Macaulay2 log stored at /home/mahrud/.Macaulay2/history.m2
```
Sorry to be making this PR more complicated for you!

d-torrance: The `FileName` option is used for all the `Core` tests.  `JSON` too.  I'm not sure if there are any others -- it's relatively new.

d-torrance: Hmm, yeah that's a tricky bug.  I think it's probably always existed for `locate`, since it assumes we're calling `TEST` on a literal string and not the output of some function.  It's just now that `code` is calling `locate`, it's showing up there, too.

I guess the documentation for `TEST` could discourage the use of `get` for this reason, and we could modify the handful of tests that currently use it.  (I'm only counting 13.)

mahrud: Let's first try to find a way to fix the bug because I'd rather keep allowing `TEST get ...`, it's much simpler than using an option.

mahrud:
> I'd rather stick with the current behavior for Emacs clickability.

Can you say more about how you use this? e.g. I would imagine `locate tests Saturation` would still produce a clickable list. I think eventually it would be good if the printing of a `TestInput` looked something closer to a document tag, telling you that this is a test for such and such method.

d-torrance:
My initial rationale for introducing `tests` was for debugging failing builds.  We'd have some log that said such and such test failed, but it was non-trivial to figure out what exactly that test was and where it was located in the source.  I figured using something similar to `net(Function)` would be useful.  A test isn't a function, but it's kind of similar in that they both consist of some hunk of code.

So now, if we see on some log that `check(37, "Foo")` failed, we can just run `tests "Foo"` and click on the `37 => TestInput[...]` link that shows up to start debugging.

d-torrance:
I've fixed the test location bug, which required turning `TEST` into a keyword so that we could figure out the location of its argument while it's still an unevaluated `Code` object.

I also added the ability for `TEST` to accept a `File` object.  This way, the location is actually the source of the test rather than the line that calls `get`.  For example:

```m2
i1 : TEST get "/usr/share/Macaulay2/Macaulay2Doc/demo1.m2"

i2 : TEST openIn "/usr/share/Macaulay2/Macaulay2Doc/demo1.m2"

i3 : tests User

o3 = {0 => TestInput[stdio:1:5-1:53]                                     }
     {1 => TestInput[/usr/share/Macaulay2/Macaulay2Doc/demo1.m2:1:1-14:9]}

o3 : NumberedVerticalList

i4 : code oo

o4 = stdio:1:5-1:53: --source code:
     TEST get "/usr/share/Macaulay2/Macaulay2Doc/demo1.m2"
     /usr/share/Macaulay2/Macaulay2Doc/demo1.m2:1:1-14:9: --source code:
     -- when running examples, we usually specify the --stop option
     -- here we reverse that so we can demonstrate the debugger in an example
     stopIfError = false
     debuggingMode = true

     f := x -> (
          a := "hi there";
          b := 1/x;
          b+1)

     g = y -> (
          c := f(y-1);
          d := f(y-2);
          c+d)
```

d-torrance:
> In principle this sounds useful, but I can totally imagine it becoming a pain to figure out where a test file was loaded from!

Great point!   I'll remove the `File` method.

mahrud:
By the way, one benefit of turning TEST into a keyword is that the content in front of it doesn't have to be a string anymore. It could literally be `TEST ( ... )` and the stuff in front could be kept for future evaluation. 

Not necessarily suggesting doing this now, but worth considering.

M2/Macaulay2/packages/LLLBases.m2
```diff
@@ -1248,6 +1248,7 @@ TEST ///
 
 TEST 
 /// -- DON'T TEST YET??
+    -- no-check-flag (encountered an unknown key or option: Engine)
```
d-torrance, line 1251: @mikestillman - This test appears to never have been run since `TEST` and the test string were on separate lines.  After changing `TEST` to a keyword, the test does get run, but it fails because `LLL` doesn't have an `Engine` option.

I'm proposing that we continue skipping it for now (which `no-check-flag` takes care of), but I figured I'd tag you in case you'd like to take a look.

M2/Macaulay2/d/binding.d
```diff
@@ -311,6 +311,7 @@ bumpPrecedence();
      export elapsedTimeS := special("elapsedTime",unaryop,precSpace,wide);
      export elapsedTimingS := special("elapsedTiming",unaryop,precSpace,wide);
      export shieldS := special("shield",unaryop,precSpace,wide);
+     export testS := special("TEST",unaryop,precSpace,wide);
```
mahrud, line 314: I'm a little unsure if `testS` is the right term here, especially if we wanted an actual `test` keyword in the future. `TESTS` also seems wrong, but maybe `TestS`?

M2/Macaulay2/tests/normal/names.m2
```diff
@@ -360,7 +360,7 @@ assert( (toString SyzygyRows) === "SyzygyRows" )
 assert( (toString T) === "T" )
 assert( (toString TABLE) === "TABLE" )
 assert( (toString TD) === "TD" )
-assert( (toString TEST) === "TEST" )
+assert( (toString symbol TEST) === "TEST" )
```
mahrud, line 363: I think you should just remove this line.

M2/Macaulay2/packages/LLLBases.m2
mahrud: Was there an engine implementation that got removed or never got added? Seems strange.

M2/Macaulay2/d/common.d
```diff
@@ -110,14 +110,6 @@ export setup(e:SymbolClosure,fn:ternop):void := (
 export setup(e:SymbolClosure,fn:multop):void := (
      multopNameList = multopNameListCell(fn,e.symbol,multopNameList);
      );
-export setupfun(name:string,fun:unop):Symbol := (
```
mahrud, line 113: What's the rationale for removing this?

M2/Macaulay2/m2/testing.m2
```diff
@@ -73,9 +63,14 @@ loadTestDir := pkg -> (
...
-            match("\\.m2$", file)), test -> testDir | test),
-            FileName => true);
+	scan(sort select(readDirectory testDir, file -> match("\\.m2$", file)),
+	    test -> (
+		testpath := testDir | test;
+		teststr := get testpath;
+		testlines := lines teststr;
+		addTest(teststr,
+		    new FilePosition from(testpath, 1, 1,
+			#testlines, #testlines#-1))));
```
mahrud, line 73: This seems range here. Maybe add another method to `addTest` that takes a filename and call it here? And in that method, I don't think you need to store an actual `FilePosition`, just the string filename (possibly with ":0:0" appended) should suffice. This is actually a good reason why having `FilePosition` with only a start point is useful.

M2/Macaulay2/packages/JSON.m2
```diff
@@ -358,8 +358,5 @@ for tst in sort select(tsts, f -> match("^y_", f)) do (
 close outfile
 ///
 
-TEST(currentPackage#"source directory" | "JSON/test-parse.m2",
-    FileName => true)
-
-TEST(currentPackage#"source directory" | "JSON/test-encode.m2",
-    FileName => true)
+TEST get(currentPackage#"source directory" | "JSON/test-parse.m2")
```
mahrud, line 361: I _think_ just `TEST "./JSON/..."` should always work, no?

M2/Macaulay2/packages/Macaulay2Doc/functions/check-doc.m2
```diff
@@ -5,21 +5,13 @@
 doc ///
 Node
   Key
-     TEST
-    (TEST, String)
-    (TEST, List)
-    [TEST, FileName]
+     symbol TEST
```
mahrud, line 8: I haven't tested this, do both `help symbol TEST` and `help "TEST"` work this way? I think other keywords are usually documented as strings.

M2/Macaulay2/d/binding.d
d-torrance: Sounds good -- I'll switch it to `TestS`

M2/Macaulay2/d/common.d
d-torrance: It's not called anywhere, and I'm not sure it even works.  While I was playing around with these changes, I unsuccessfully tried using it to define `TEST` as a keyword.

M2/Macaulay2/packages/JSON.m2
d-torrance: It doesn't :(  I get the following:

```m2
i1 : check "JSON"
../src/macaulay2/M2/M2/Macaulay2/packages/JSON.m2:361:0:(2):[18]: error: opening input file "./JSON/test-parse.m2" failed: No such file or directory
```

I think it depends on the value of `currentDirectory()`.  If I `changeDirectory` into the packages directory, then it works just fine.

M2/Macaulay2/packages/Macaulay2Doc/functions/check-doc.m2
d-torrance: Yes, they both work

M2/Macaulay2/d/common.d
mahrud: Could this be used for something like `not`?

M2/Macaulay2/packages/JSON.m2
mahrud: Hmm but `load "./..." works correctly. What's the difference Does `TEST(currentFileDirectory | ...)` work?

d-torrance: Yes, it works with `currentFileDirectory`

M2/Macaulay2/d/common.d
d-torrance: I played around with this a bit more, and keywords created with this function end up being left associative instead of right associative like we'd probably expect.  So if we were to use it to define `not`, something like `not f x` would be parsed as `(not f) x` instead of `not (f x)`.

M2/Macaulay2/d/common.d
```diff
@@ -100,24 +100,12 @@ export setup(e:SymbolClosure,fun1:unop,fun2:unop):void := (
      e.symbol.unary = fun1;
      e.symbol.postfix = fun2;
      );
-export setupop(s:SymbolClosure,fun:unop):void := (
```
mahrud, line 103: A very good reason to keep `setupop` is for ease of searching through the codebase for places where operators are declared.

mahrud:
I'm hesitant about the broader interpreter changes in this PR. Could you scale back this PR to just the new `TEST` keyword?

M2/Macaulay2/packages/JSON.m2
```diff
@@ -358,8 +358,5 @@ for tst in sort select(tsts, f -> match("^y_", f)) do (
 close outfile
 ///
 
-TEST(currentPackage#"source directory" | "JSON/test-parse.m2",
-    FileName => true)
-
-TEST(currentPackage#"source directory" | "JSON/test-encode.m2",
-    FileName => true)
+TEST get(currentPackage#"auxiliary files" | "test-parse.m2")
```
mahrud, line 361: Reading this line, I would be more confused about where to look for that file than before. You can of course go with whatever solution you prefer in your package.

mahrud (approved):

M2/Macaulay2/packages/JSON.m2
mahrud: Ah, the difference is that `load` looks everywhere under `path` but `get` doesn't. A little strange to be honest.

mahrud:
Hmm one last thing that comes to mind: is there an easy way to make sure none of the undistributed-packages stop working because of this change to `TEST`? I doubt it's the case, but it's possible. Maybe we should also run the tests / examples of those packages on GitHub.

d-torrance:
I checked the undistributed packages locally on this branch.  A good chunk of them don't even load -- perhaps why they're undistributed -- but the tests pass for the rest of them:

```m2
 -- warning: CacheFileName has no tests
 -- warning: CodepthThree did not load
 -- warning: CustomEngineTests did not load
 -- capturing check(0, "ExampleFreeResolutions") -- .102555s elapsed
 -- capturing check(1, "ExampleFreeResolutions") -- .0819185s elapsed
 -- capturing check(2, "ExampleFreeResolutions") -- .105076s elapsed
 -- running   check(3, "ExampleFreeResolutions") -- 1.68029s elapsed
 -- warning: ExampleIdeals has no tests
 -- warning: FastLinearAlgebra did not load
 -- warning: FreeResolutions has no tests
 -- warning: FrobeniusMultiplicities did not load
 -- warning: Functoriality did not load
 -- capturing check(0, "GenerateD")              -- .0782202s elapsed
 -- capturing check(1, "GenerateD")              -- .078785s elapsed
 -- capturing check(2, "GenerateD")              -- .082572s elapsed
 -- capturing check(3, "GenerateD")              -- .0757686s elapsed
 -- capturing check(4, "GenerateD")              -- .0796765s elapsed
 -- running   check(5, "GenerateD")              -- 3.36057s elapsed
 -- capturing check(6, "GenerateD")              -- .0784762s elapsed
 -- warning: MemoryLeaks did not load
 -- capturing check(0, "MGBInterface")           -- .114109s elapsed
 -- capturing check(1, "MGBInterface")           -- .091242s elapsed
 -- running   check(2, "MGBInterface")           -- 3.44805s elapsed
 -- running   check(3, "MGBInterface")           -- 1.77514s elapsed
 -- running   check(4, "MGBInterface")           -- 1.74921s elapsed
 -- running   check(5, "MGBInterface")           -- 1.79347s elapsed
 -- running   check(6, "MGBInterface")           -- 4.34543s elapsed
 -- warning: Mysql did not load
 -- capturing check(0, "NAGtools")               -- 1.14616s elapsed
 -- warning: ParameterSchemes did not load
 -- warning: Polyhedra2 did not load
 -- warning: PolyhedralObjects did not load
 -- warning: PolymakeInterface did not load
 -- warning: RandomSearch has no tests
 -- warning: RisaAsir did not load
 -- warning: Sage did not load
 -- warning: SchurRingsOld has no tests
 -- warning: SecondPackage did not load
 -- warning: SuffixTrees has no tests
 -- warning: ToricCohomology did not load
```

mahrud:
> A good chunk of them don't even load -- perhaps why they're undistributed

This is a matter for another time, but this isn't great. If the undistributed packages don't load, what's the point of keeping them?

M2/Macaulay2/tests/normal/testing.m2
```diff
@@ -7,14 +7,17 @@ loadPackage("TestPackage", FileName => testpkg)
...
 assert instance(pkgtest, TestInput)
-assert Equation(toSequence locate pkgtest, (testpkg, 3, 1, 4, 1,,))
-assert Equation(toString pkgtest, testpkg | ":3:1-4:1")
-assert Equation(net pkgtest, "TestInput[" | testpkg | ":3:1-4:1]")
-expectedCode =  " -- " | toAbsolutePath testpkg |
-	":4:1: location of test code" | newline | "assert Equation(1 + 1, 2)"
-assert Equation(code pkgtest, expectedCode)
-assert Equation(code 0, expectedCode)
-
+assert Equation(toSequence locate pkgtest, (testpkg, 3, 5, 3, 32, 3, 5))
```
mahrud, line 10: This happened in a brew build:
```m2
 -- running   check(357, "Core")              -- .380111s elapsed
/opt/homebrew/Cellar/macaulay2/1.24.11-rc1/share/doc/Macaulay2/Core/tests/testing.m2:1:1 error:
 -- i5 : pkgtest = tests(0, "TestPackage")
 -- 
 -- o5 = TestInput[../../M2-93313-0/0.m2:3:5-3:32]
 -- 
 -- o5 : TestInput
 -- 
 -- i6 : assert instance(pkgtest, TestInput)
 -- 
 -- i7 : assert Equation(toSequence locate pkgtest, (testpkg, 3, 5, 3, 32, 3, 5))
 -- stdio:10:6:(3): error: assertion failed:
 -- (../../M2-93313-0/0.m2, 3, 5, 3, 32, 3, 5) == (/private/tmp/M2-93313-0/0.m2, 3, 5, 3, 32, 3, 5) is false
 -- 
stdio:1:5:(3): error: test(s) #333 of package Core failed.
```

d-torrance: Oh, that makes sense.  Calling `minimizeFilename` should fix it.  I'll open a PR.

PR #3423, opened 2024-08-22
Few typos packages

fchapoton:
just a few typos fixed in packages


M2/Macaulay2/editors/emacs
mahrud, line 1: Your PR shouldn't change the submodule. If you'd like to fix typos in the submodule the PR should appear in [M2-emacs](https://github.com/Macaulay2/M2-emacs) first.

mahrud (changes requested):

mahrud (approved):

mahrud:
Also, do you exactly which typos are caught by codespell and which aren't? e.g. see https://github.com/Macaulay2/M2/pull/3392#discussion_r1707261601. Are documentation typos not caught?

fchapoton:
As far as I know, `codespell` is using a long list of "frequent typos" with their correct replacement. This list is constantly evolving, but of course does not catch everything. This is not sophisticated enough to find concatenated words. I used myself here a manual process starting with `egrep` command to extract, for example, all words ending with `tion` and then looked at this list to find bad guys.

Issue #3424, opened 2024-08-22
Msolve issues


mahrud:
- [x] As mentioned in https://github.com/Macaulay2/M2/pull/3392#issuecomment-2274679356, the functioni `readInteger` in [BasicPolyListParser.cpp](https://github.com/Macaulay2/M2/blob/ec9e9ac60ed4a8e791448942202f077a88a87e15/M2/Macaulay2/e/BasicPolyListParser.cpp#L41-L61) should read in arbitrary precision integers. (cc: @mikestillman)
- [x] This error should be improved:
```m2
i1 : needsPackage "Msolve"

o1 = Msolve

o1 : Package

i2 : S = QQ[x,y,h]

o2 = S

o2 : PolynomialRing

i3 : I = ideal"3x2h-x3-y3,3y2h-2x3-2y3"

               3    3     2       3     3     2
o3 = ideal (- x  - y  + 3x h, - 2x  - 2y  + 3y h)

o3 : Ideal of S

i4 : msolveRealSolutions I -- FIXME
stdio:4:19:(3): error: parallel assignment: expected a sequence of 2 values

i5 : I = ideal"3x2-x3-y3,3y2-2x3-2y3"

               3    3     2      3     3     2
o5 = ideal (- x  - y  + 3x , - 2x  - 2y  + 3y )

o5 : Ideal of S

i6 : msolveRealSolutions I -- FIXME
stdio:6:19:(3): error: parallel assignment: expected a sequence of 2 values
```
This is probably introduced by me, but in fairness I think the documentation of Msolve is unclear or has a mistake about the output in this case.

cc: @Martin-Helmer @antonleykin

Martin-Helmer:
This error use to be caught, for example when I run both these examples on my slightly older version of the Msolve package:

```
msolveRealSolutions I
stdio:5:1:(3): error: Input ideal not zero dimensional, no solutions found.
```

The lines that caught it were:
```
solsp := readMsolveList get mOut;
if solsp_0>0 then (error "Input ideal not zero dimensional, no solutions found.";);
if (solsp_1)_0>1 then (print "unexpected msolve output, returning full output"; return solsp;);
```

mahrud:
Yes, I realized that was on purpose and fixed this in #3425. The first issue I think Mike probably intends to fix. so maybe I'll assign this issue to him.

mahrud:
Fixed in #3763.

PR #3425, opened 2024-08-22
Fixed incorrect error in msolveRealSolutions

mahrud:
Fixes part 2 of #3424.

### DIFF
--- a/M2/Macaulay2/packages/Msolve.m2
+++ b/M2/Macaulay2/packages/Msolve.m2
@@ -220,9 +220,11 @@ msolveRealSolutions(Ideal, Ring)       := opt -> (I0, F) -> (
     -- if precision is not specified, we want to use msolve's default precision
     prec := if precision F === infinity then msolveDefaultPrecision else precision F;
     mOut := msolve(S, K, I_*, "-p " | prec, opt);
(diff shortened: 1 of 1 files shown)


M2/Macaulay2/packages/Msolve.m2
```diff
@@ -220,9 +220,11 @@ msolveRealSolutions(Ideal, Ring)       := opt -> (I0, F) -> (
     -- if precision is not specified, we want to use msolve's default precision
     prec := if precision F === infinity then msolveDefaultPrecision else precision F;
     mOut := msolve(S, K, I_*, "-p " | prec, opt);
-    -- format: [dim, [numlists, [ solution boxes ]]]
-    (d, solsp) := toSequence value readMsolveList get mOut;
+    -- format: [dim, [numlists, [ solution boxes ]]] when zero-dimensional, otherwise [1?, numgens, -1, []]
+    mSeq := toSequence value readMsolveList get mOut;
```
d-torrance, line 224: So was the previous parallel assignment error occurring because `mSeq` might not have two elements?

mahrud: Yes, if the input is not zero dimensional the output has weirdly a different format entirely.

PR #3433, opened 2024-08-25
Update several cached examples

d-torrance:
This updates the version numbers of PHCpack and Python that appear in some cached examples, and also removes some verbose polymake output (which I don't think has appeared since way back in #1373).


mahrud (approved):
I couldn't think of anyone better in this particular case, but it would be good to slowly diversify the reviewers.

PR #3435, opened 2024-08-26
Fixed order of FilePosition components

mahrud:
- **moved a dummy declaration to expr.d**
- **fixed the order of FilePosition components**

c.f. https://github.com/Macaulay2/M2/pull/3431#issuecomment-2309138711

### DIFF
--- a/M2/Macaulay2/d/evaluate.d
+++ b/M2/Macaulay2/d/evaluate.d
@@ -1305,11 +1305,11 @@ steppingFurther(c:Code):bool := steppingFlag && (
      if p == dummyPosition || p.loadDepth < errorDepth then return true;
      if stepCount >= 0 then (
 	  if lastCodePosition.filename != p.filename
-	  || lastCodePosition.line != p.line
+	  || lastCodePosition.lineF != p.lineF
 	  then (
      	       stepCount = stepCount - 1;
      	       lastCodePosition.filename = p.filename;
-	       lastCodePosition.line = p.line;
+	       lastCodePosition.lineF = p.lineF;
 	       if debugLevel == 1001 && stepCount >= 0 then printErrorMessage(p,"--evaluating: "+present(tostring(c)));
 	       );
 	  stepCount >= 0)
--- a/M2/Macaulay2/d/stdiop.d
+++ b/M2/Macaulay2/d/stdiop.d
@@ -118,7 +118,7 @@ export tostring(w:Position) : string := (
 	     if c == ' ' then (
 		 filename = "\"" + filename + "\"";
 		 break));
-	 errfmt(filename ,int(w.line),int(w.column),int(w.loadDepth))));
+	 errfmt(filename, int(w.lineF), int(w.columnF), int(w.loadDepth))));
 export (o:file) << (w:Position) : file := o << tostring(w);
 export (o:BasicFile) << (w:Position) : BasicFile := o << tostring(w);
 threadLocal export SuppressErrors := false;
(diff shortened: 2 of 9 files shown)


M2/Macaulay2/d/stdiop.d
```diff
@@ -118,7 +118,7 @@ export tostring(w:Position) : string := (
 	     if c == ' ' then (
 		 filename = "\"" + filename + "\"";
 		 break));
-	 errfmt(filename ,int(w.line),int(w.column),int(w.loadDepth))));
+	 errfmt(filename, int(w.lineL), int(w.columnL), int(w.loadDepth))));
```
mahrud, line 121: Should this point to `lineF` and `columnF` (the focused location) instead, or just the beginning?

pzinn: I'm confused. the current `line`, `column`, *are* the focus position.

mahrud: Sure, that was your choice, but it's misleading given the top-level format for the _same object_ and the existing documentation, so we move the focus location to the end instead, and make the corresponding changes everywhere else a position is constructed. The names `line1`, `line2`, etc. shouldn't have been used in the first place, especially given that there were no commentary provided and the documentation of the object online pointed to the last pair as the focus position.

mahrud:
To be clear, this PR doesn't change any top-level behavior whatsoever. You can verify with the tests I added, and if you find anything that is changed we should add them as tests to make sure nothing changes by accident.

M2/Macaulay2/d/stdiop.d
pzinn: if you change the ordering in the definition of `export Position`, don't you have to change every line of code that defines a new Position to take into account that new ordering? I don't see those changes.

pzinn:
yes I will provide more tests, I have a list of them, I just forgot to include them in my PR

M2/Macaulay2/d/stdiop.d
mahrud: In all the other ones all three locations are the same. I changed all the ones where they're different. Have I missed any?

pzinn: OK I missed the unexpanded changes in `convertr.d`. looks good.

pzinn: anyway back to the start of this thread, I do believe it should be `lineF` and `columnF`, yes.

M2/Macaulay2/tests/normal/locate.m2
pzinn, line 1: thinking about tests -- by defining both `f` and `F`, are you suggesting they should have different locations? I think I vaguely considered that at some point (maybe `F` should only have the right of `->`?) but rejected it, don't remember why though.

pzinn: the simplest one that has all locations distinct is probably
```
() -> 1 + 2
```

pzinn: most of my "tests" actually involve errors -- not sure how to turn them into `assert` type tests.

mahrud: > are you suggesting they should have different locations

I can see it both ways, but I'm just adding the test lines so if they change unintentionally we find out.

M2/Macaulay2/d/evaluate.d
```diff
@@ -1305,11 +1305,11 @@ steppingFurther(c:Code):bool := steppingFlag && (
      if p == dummyPosition || p.loadDepth < errorDepth then return true;
      if stepCount >= 0 then (
 	  if lastCodePosition.filename != p.filename
-	  || lastCodePosition.line != p.line
+	  || lastCodePosition.lineF != p.lineF
 	  then (
      	       stepCount = stepCount - 1;
      	       lastCodePosition.filename = p.filename;
-	       lastCodePosition.line = p.line;
+	       lastCodePosition.lineF = p.lineF;
```
mahrud, line 1312: This is another place where I'm not sure if the correct location should be left-point or focused-point. Technically all three were the same before, so which one is important here?

M2/Macaulay2/tests/normal/locate.m2
mahrud: > most of my "tests" actually involve errors -- not sure how to turn them into `assert` type tests.

You can use `capture`. There are several examples of this already.

M2/Macaulay2/d/evaluate.d
pzinn: I'd stick to focus for now. one may want to rewrite some of this to take advantage of the 3 positions, but I don't think there's any rush to do so.

M2/Macaulay2/tests/normal/locate.m2
pzinn: OK. I think this can wait till my next PR on this kind of stuff.

pzinn, line 1: am I misreading, or are lines 1 and 23 identical?

mahrud: Do you know how `generateAssertions` works? The test is only lines 1 through 13, then `end` is called. The bottom portion is for autogenerating the top portion (i.e. I didn't manually fill in the numbers ...)

pzinn: I see.

mahrud: If you already have tests that you know what to expect and are correct now, just paste them here or send them to me and I can put them in.

pzinn (approved):

M2/Macaulay2/tests/normal/locate.m2
mahrud: I was not a fan of it at first, but then I figured out how to make it very smooth to update the test (see the commented out block in the middle) and now I'm a fan.

Issue #3428, opened 2024-08-23
"make check" fails for msolve (autotools)


d-torrance:
When building msolve using autotools and running `make check`:

```
$ make -C libraries/msolve check
.
.
.
make[4]: Entering directory '/home/profzoom/src/macaulay2/M2/M2/BUILD/doug/libraries/msolve/build/msolve-0.7.1'
PASS: neogb_io
./test-driver: line 112: 2744164 Segmentation fault      (core dumped) "$@" >> "$log_file" 2>&1
FAIL: fglm_build_matrixn_radical_shape-31
./test-driver: line 112: 2744194 Segmentation fault      (core dumped) "$@" >> "$log_file" 2>&1
FAIL: fglm_build_matrixn_nonradical_shape-31
./test-driver: line 112: 2744223 Segmentation fault      (core dumped) "$@" >> "$log_file" 2>&1
FAIL: fglm_build_matrixn_nonradical_radicalshape-31
FAIL: test/diff/diff_cp_d_3_n_4_p_2.sh
FAIL: test/diff/diff_cyclic5-16.sh
FAIL: test/diff/diff_cyclic5-31.sh
FAIL: test/diff/diff_cyclic5-qq.sh
FAIL: test/diff/diff_eco6-16.sh
FAIL: test/diff/diff_eco6-31.sh
FAIL: test/diff/diff_eco6-qq.sh
FAIL: test/diff/diff_henrion5-qq.sh
FAIL: test/diff/diff_eco10-31.sh
FAIL: test/diff/diff_truncate_gb.sh
FAIL: test/diff/diff_elim-31.sh
FAIL: test/diff/diff_elim-qq.sh
FAIL: test/diff/diff_F4SAT-31.sh
FAIL: test/diff/diff_F4SAT-byone-31.sh
FAIL: test/diff/diff_kat6-31.sh
FAIL: test/diff/diff_kat7-qq.sh
FAIL: test/diff/diff_multy-16.sh
FAIL: test/diff/diff_multy-31.sh
FAIL: test/diff/diff_multy-qq.sh
FAIL: test/diff/diff_nonradical_shape-qq.sh
FAIL: test/diff/diff_nonradical_radicalshape-qq.sh
FAIL: test/diff/diff_nonradical_radicalshape-no-square-31.sh
FAIL: test/diff/diff_nonradical_radicalshape-no-square-qq.sh
FAIL: test/diff/diff_one-16.sh
FAIL: test/diff/diff_one-31.sh
[...]

PR #3436, opened 2024-08-26
Use unlimited virtual memory when running msolve tests (autotools)

d-torrance:
Most of the tests fail using the default 400000 kB.

Closes: #3428

Here's a successful run of the tests after this change: https://github.com/d-torrance/M2/actions/runs/10552691192/job/29231880174

### DIFF
--- a/M2/libraries/msolve/Makefile.in
+++ b/M2/libraries/msolve/Makefile.in
@@ -6,5 +6,7 @@ LICENSEFILES = README.md COPYING
 
 PRECONFIGURE = ./autogen.sh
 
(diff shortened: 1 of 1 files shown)


mahrud:
Perhaps you could request one of the authors to review?

mikestillman:
We have the limit in for a reason, so that tests (mostly) don't need to be run on large memory machines.  How much memory is the msolve executable taking?

d-torrance:
Ok, that makes sense.  I'll do some experiments and find a finite upper bound.

PR #3437, opened 2024-08-26
Only run lint workflow on pull requests (or workflow dispatch)

d-torrance:
This way, it won't be run on all pushes of arbitrary branches to developers' forks unless explicitly requested using a workflow dispatch.

Minus the weekly run, the proposed settings are the same as the "Build and Test" workflow.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,5 +1,10 @@
 name: Lint
-on: [ push, pull_request ]
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - master
+      - development
 
 jobs:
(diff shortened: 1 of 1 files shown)


mahrud:
This is a pretty harmless action though, and it's helpful for people to see the typos before they make a PR, no?

d-torrance:
I was mostly thinking of consistency between the two workflows.  Every time I push to my fork, the lint workflow runs, but the build/test one doesn't.  But of course, this one is much, much smaller, and as Mahrud said, it's helpful to catch these things before making a PR.

I don't feel very strongly about this -- I'll go ahead and close it.

PR #3438, opened 2024-08-26
Add finite limit to virtual memory used when checking msolve

d-torrance:
Rather than allowing unlimited virtual memory, we increase the upper bound to 1.5 GB, which should still allow all the tests to pass.

Here's an example of the tests passing after this commit: https://github.com/d-torrance/M2/actions/runs/10561928026/job/29258736811

I ran the tests a bunch of times locally, incrementing the upper bound by 100,000 until they all passed.  By 1.3 GB, they all passed but one (test/diff/diff_nonradical_shape-qq.sh), and then they all passed at 1.4 GB.  I figured I'd round up a bit more to 1.5 GB just in case.

### DIFF
--- a/M2/libraries/msolve/Makefile.in
+++ b/M2/libraries/msolve/Makefile.in
@@ -6,7 +6,7 @@ LICENSEFILES = README.md COPYING
 
 PRECONFIGURE = ./autogen.sh
 
-VLIMIT = unlimited
+VLIMIT = 1500000
 
 include ../Makefile.library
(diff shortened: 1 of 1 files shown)


mahrud:
(I just cancelled the builds because they weren't going to test this change anyway.)

d-torrance:
I'm going to go ahead and merge this and a couple other small autotools build-related PR's that have been open for a while (#3464 and  and #3466) since there have been no objections.  Feel free to revert!

PR #3440, opened 2024-08-26
Complexes for version 1-24-11, part 1.

mikestillman:
This is the first PR for converting the ChainComplex and ChainComplexMap types from the Core to the types Complex, ComplexMap in Complexes package.  Here are the steps we envision to do the switch:

- [ ] (This pull result) Make changes to Complexes and minor changes elsewhere so that it will be easy to set Complexes to be a preloaded package. Also includes some bug fixes and minor improvements.
- [ ] A series of PR's that change each package to run with Complexes, rather than ChainComplexes.
- [ ] Finally, we remove ChainComplex and ChainComplexMap from the Core, including `m2/res.m2`, `m2/chaincomplexes.m2`, and move/remove documentation and tests from the Core.
- [ ] Make Complexes a preloaded package.
- [ ] Probably (needs discussion), change the names Complex, ComplexMap in Complexes, back to ChainComplex, ChainComplexMap.  Rename the Complex package to ChainComplexes.


M2/Macaulay2/m2/shared.m2
```diff
@@ -24,6 +24,10 @@ isEmpty = method(TypicalValue => Boolean)
 isSmooth = method(TypicalValue => Boolean, Options => true)
 isVeryAmple = method(TypicalValue => Boolean, Options => true)
 
+isNormal = method()
+
+normalCone = method(Options => true)
```
mahrud, line 29: These two should be added in `packages/Macaulay2Doc/shared.m2`, ideally with references to `IntegralClosure::isNormal` and other packages using these symbols.

M2/Macaulay2/packages/ReesAlgebra.m2
```diff
@@ -1256,10 +1278,8 @@ doc ///
 
 doc ///
   Key
-    normalCone
-    (normalCone, Ideal)
     (normalCone, Ideal, RingElement)
-    
+    (normalCone, Ideal)
```
mahrud, line 1282: You can document `associatedGradedRing` and its methods here, since it is now missing documentation.

M2/Macaulay2/packages/ReesAlgebra.m2
```diff
@@ -209,26 +207,50 @@ isLinearType(Module, RingElement):= o-> (N,a)->(
...
+-- )
+-- normalCone(Ideal) := o -> I -> (
+--      RI := reesAlgebra(I,o);
+--      RI/promote(I,RI)
+--      )
+
+-- normalCone(Ideal, RingElement) := o -> (I,a) -> (
+--      RI := reesAlgebra(I,a,o);
+--      RI/promote(I,RI)     
+--      )
```
mahrud, line 228: Is there a reason this section is duplicated here? It's all just below, or am I missing something?

M2/Macaulay2/packages/ReesAlgebra.m2
```diff
@@ -1980,7 +2000,6 @@ doc ///
     [reesAlgebra,Strategy]
     [isLinearType,Strategy]
     [isReduction, Strategy]    	  
-    [normalCone, Strategy]
```
mahrud, line 1983: You can document the strategy for `associatedGradedRing` here instead, and the for the other options,

M2/Macaulay2/packages/ReesAlgebra.m2
```diff
@@ -209,26 +207,50 @@ isLinearType(Module, RingElement):= o-> (N,a)->(
...
      RI/promote(I,RI)
      )
 
-normalCone(Ideal, RingElement) := o -> (I,a) -> (
+normalCone(Ideal, RingElement) := Ring => normalConeOptions >> o  -> (I,a) -> (
      RI := reesAlgebra(I,a,o);
      RI/promote(I,RI)     
      )
 
+associatedGradedRing = method(Options => normalConeOptions)
```
mahrud, line 250:
```suggestion
associatedGradedRing = method(TypicalValue => Ring, Options => normalConeOptions)
```

mahrud (changes requested):
I only reviewed the non-Complexes parts of this PR.

Devlin-Mallory:
I noticed what I think is a small bug - I was going to make a separate pull request but maybe it makes more sense to change it as part of this PR:

In the "extend" method in ChainComplexMap.m2, lines 795 -- 799 are intended to check if the output of the preceding lines is really a map of chain complexes. However, since line 795 is `if false and opts.Verify`, the user has no way to actually trigger this, no matter what the value of Verify is. I would suggest simply removing "false and", but is a comment saying "TODO: the following line: "false and" should be removed when we switch Verify to have default value false", so perhaps I am missing some context.

M2/Macaulay2/m2/exports.m2
```diff
@@ -954,6 +955,7 @@ export {
 	"nextPrime",
 	"nextkey",
 	"norm",
+        "normalCone",
```
antonleykin, line 958:
```suggestion
        "NormalCone",
```

M2/Macaulay2/m2/shared.m2
```diff
@@ -24,6 +24,10 @@ isEmpty = method(TypicalValue => Boolean)
 isSmooth = method(TypicalValue => Boolean, Options => true)
 isVeryAmple = method(TypicalValue => Boolean, Options => true)
 
+isNormal = method()
```
antonleykin, line 27: isNormal is weird

antonleykin (commented):
This is a DEMO!

M2/Macaulay2/packages/ReesAlgebra.m2
mikestillman: We placed the return type into each usage, which we feel is a better place to document the return type

M2/Macaulay2/packages/ReesAlgebra.m2
mikestillman: Done

M2/Macaulay2/packages/ReesAlgebra.m2
mikestillman: commented out code removed, thanks for noticing!

M2/Macaulay2/m2/shared.m2
mikestillman: doc added

M2/Macaulay2/packages/ReesAlgebra.m2
mahrud: Is this a general preference against using `TypicalValue` option of `method`? Can I ask why?

M2/Macaulay2/packages/Complexes/ChainComplexDoc.m2
```diff
@@ -4363,6 +4372,66 @@ doc ///
...
+        (Ext,Module,Ring)
+    Headline
+        total Ext module
+    Usage
+        Ext(M, N)
+    Inputs
+        M:Module
+            or ofClass{Ideal,Ring}, that is homogeneous
+        N:Module
+            or ofClass{Ideal,Ring} over the same ring as $M$, that is also homogeneous
```
mahrud, line 4391: I don't think `ofClass{Ideal,Ring}` works like this ...
```suggestion
        M:{Ring,Ideal,Module}
            that is homogeneous
        N:{Ring,Ideal,Module}
            over the same ring as $M$, that is also homogeneous
```

M2/Macaulay2/packages/Complexes/ChainComplexDoc.m2
```diff
@@ -4363,6 +4372,66 @@ doc ///
...
+    Outputs
+        :Module
+            the $\operatorname{Ext}$ module of $M$ and $N$, as a
```
mahrud, line 4394: In principle, I'm in favor of using latex in the documentation, but this is also displayed in emacs/terminal and it doesn't display well there. We could define macros for `\Ext` (and `\Hom`, `\Tor`, etc.) in the header in `m2/html.m2` though.

We can also actively decide that we don't care about how the documentation looks in emacs/terminal, but I don't think that's the consensus, right?

M2/Macaulay2/packages/Complexes/ChainComplexDoc.m2
```diff
@@ -4363,6 +4372,66 @@ doc ///
...
+            S = ring H;
+            H
+            isHomogeneous H
+            rank source basis( { -2,-3 }, H)
+            rank source basis( { -3 }, Ext^2(N,N) )
+            rank source basis( { -4,-5 }, H)
+            rank source basis( { -5 }, Ext^4(N,N) )
+            hilbertSeries H
+            hilbertSeries(H,Order=>11)
+        Text
```
mahrud, line 4425: You can also use the `References` keyword instead of `Text` (and undo indentation by one level).

M2/Macaulay2/packages/Complexes/ChainComplex.m2
```diff
@@ -592,6 +622,11 @@ poincareN Complex := C -> (
     f
     )
 
+rank Complex := ZZ => C -> (
+    (lo, hi) := concentration C;
+    sum for i from lo to hi list (-1)^i * rank C_i
```
mahrud, line 627: Out of curiosity, is this different than the Euler characteristic of the complex?

mikestillman:
@d-torrance Any reason why this one Core test is failing?  I don't see any artifacts with the failing tests.  It runs fine on my mac too...  It doesn't seem to me like it uses too much memory either?

d-torrance:
This is the failing test:

```m2
i1 : tests(73, "Core")

o1 = TestInput[/usr/share/doc/Macaulay2/Core/tests/ext-total.m2:1:1]
```

I see that this PR moves some of the Ext code to the `Complexes` package.  Could that be affecting the computation somehow?  It also runs just fine on my system.

M2/Macaulay2/packages/Complexes/ChainComplexDoc.m2
mikestillman: Thanks for reminding/teaching us about this syntax!

M2/Macaulay2/packages/Complexes/ChainComplexDoc.m2
mikestillman: For the moment, we will leave this in.  We might want a more systematic addition of macros to katex, or we might want to have some of the tex functions become no-ops in terminal/emacs output.  I'm not sure which is best.

M2/Macaulay2/packages/Complexes/ChainComplexDoc.m2
mikestillman: In this particular case, I think we like having the reference in place where it is,

M2/Macaulay2/packages/Complexes/ChainComplexDoc.m2
```diff
@@ -2341,10 +2351,8 @@ doc ///
    Usage
      D = C1 ** C2
    Inputs
-     C1:Complex
-       or @ofClass Module@
-     C2:Complex
-       or @ofClass Module@
+     C1:{Module}
+     C2:{Module}
```
mahrud, line 2355: I think you meant to also add `Complex` to these lists, no? Otherwise you could just write `C1:Module`. Currently the documentation would not mention that the inputs can be complexes (which I would have thought would signal an error)
```suggestion
     C1:{Module,Complex}
     C2:{Module,Complex}
```

M2/Macaulay2/packages/Complexes/ChainComplex.m2
mikestillman: That might be true for vector spaces, but in general it is similar.  We only added this for backward compatibility, as there was a function in Core doing this, and code using that.

M2/Macaulay2/packages/Complexes/ChainComplexDoc.m2
mikestillman: Since the first key is e.g. (Hom, Complex, Complex), writing C1:Module gives an error, and it turns out that using {Complex,Module} as suggested, adds in an extra Complex in the doc line.  I'm not sure if that was the planned behavior (not sure when this got added, actually).  We wanted the doc to add extra types on top of Complex, and this seemed to be the way that worked. Perhaps that should be revisited, but probably not for this PR.

mahrud: Wait, sorry, it seems like with what you have the documentation _does_ mention module and complex, but this is a bug that I will fix before the next release! You should still definitely list all possible input types here, so for instance `C1:{Module,Complex}` (even though right now it will probably say something silly like "a complex, a module, or a complex" or something like that).

mikestillman:
> This is the failing test:
> 
> ```m2
> i1 : tests(73, "Core")
> 
> o1 = TestInput[/usr/share/doc/Macaulay2/Core/tests/ext-total.m2:1:1]
> ```
> 
> I see that this PR moves some of the Ext code to the `Complexes` package. Could that be affecting the computation somehow? It also runs just fine on my system.

The problem seemed to be the changes we made in nullHomotopy.  The general case was slower (and probably took significantly more memory), so we added in the keyword FreeToExact in two calls to nullHomotopy in Ext(Module,Module).  However, with that change, the code would not run on the version before Complexes is loaded.  So we moved the test to where it should be anyway (and occurs after Complexes is loaded).

M2/Macaulay2/packages/Complexes/ChainComplexDoc.m2
d-torrance: It's possible to render TeX in Emacs (see https://github.com/Macaulay2/M2/wiki/Previewing-TeX-in-Macaulay2-Emacs-buffers), so I'd rather not make them no-ops.

mikestillman:
After this is merged, we will make a second PR that fixes about 8-10 of the packages, the ones that require mostly trivial changes.  After that, another group...  continuing as in the first comment of this PR.

M2/Macaulay2/packages/Complexes/ChainComplexDoc.m2
mahrud: Can the output of `help` be rendered as TeX automatically?

M2/Macaulay2/packages/Complexes/ChainComplexDoc.m2
mikestillman: Yes, that's exactly what it says! (Up to order).  How will you fix this bug?  Currently, it does some checking of types (like C:Module would fail in the above case).  How are you planning on modifying these checks?  (one way: make sure the types in the first key actually appear in the list, but allow others?)

mahrud: I opened an issue here: https://github.com/Macaulay2/M2/issues/3465.

For now, please add `Complex`, otherwise when `SimpleDoc` is fixed the documentation will not mention complexes anymore.

mahrud: I need to look at the code to figure out what's the best fix, but the current behavior is definitely wrong and will signal an error.

mikestillman: I'll make the changes.

mikestillman:
I made the small doc changes.  The github tests are rerunning.

mahrud (approved):

M2/Macaulay2/packages/ReesAlgebra.m2
mahrud: By the way, did you intentionally flip the order of `(normalCone, Ideal)` and `(normalCone, Ideal, RingeElement)` methods in the list? This causes the headline to say:
> normalCone(Ideal,RingElement) -- The normal cone of a subscheme

which is unusual because the second parameter is not mentioned in the headline and is optional.

mahrud:
I think this is ready to merge. Should I do it or do you want to?

mikestillman:
@d-torrance What else should we do before this PR is merged?

d-torrance:
Looks good to me!

mikestillman:
Should I just do the merge?  Should I do "rebase and merge"?  Will that screw up future PR's based on the same branch?

mikestillman:
Oops, all merged!  Thanks!

d-torrance:
No problem! Using the same branch for future PR's should be fine.

d-torrance:
@mikestillman, @ggsmith: Do you have any suggestions for how the complexes changes should be described in the changelog for the new release?

PR #3442, opened 2024-08-27
small NormalToricVarieties fix

Devlin-Mallory:
Several methods in NormalToricVarieties (notably toricBlowup and thus makeSmooth and makeSimplicial) were discarding the CoefficientRing and Variable of the input toric variety. This fixes that by retaining these options from the input toric variety.

### DIFF
--- a/M2/Macaulay2/packages/NormalToricVarieties/ToricVarieties.m2
+++ b/M2/Macaulay2/packages/NormalToricVarieties/ToricVarieties.m2
@@ -699,7 +699,7 @@ toricBlowup (List, NormalToricVariety, List) := NormalToricVariety => (s, X, v) 
 	if all (s, i -> member (i,t)) then continue
 	else t | {n}
 	);
-    Z := normalToricVariety (rays X | {v}, coneList | coneList');
+    Z := normalToricVariety (rays X | {v}, coneList | coneList', CoefficientRing => X.cache.CoefficientRing);
     Z.cache.toricBlowup = X;
     Z
     );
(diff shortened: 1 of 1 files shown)


M2/Macaulay2/packages/NormalToricVarieties/ToricVarieties.m2
```diff
@@ -693,13 +693,13 @@ toricBlowup (List, NormalToricVariety, List) := NormalToricVariety => (s, X, v)
...
-    	return normalToricVariety (rays X, coneList | coneList') 
+	return normalToricVariety (rays X, coneList | coneList', CoefficientRing => X.cache.CoefficientRing, Variable => X.cache.Variable)
 	);
     coneList' = for t in clStar list (
 	if all (s, i -> member (i,t)) then continue
 	else t | {n}
 	);
-    Z := normalToricVariety (rays X | {v}, coneList | coneList');
+    Z := normalToricVariety (rays X | {v}, coneList | coneList', CoefficientRing => X.cache.CoefficientRing, Variable => X.cache.Variable);
     Z.cache.toricBlowup = X;
```
mahrud, line 703: Do you know why the base `X` is cached here but not above?

Devlin-Mallory: Not sure I understand the question - where above would/should X be cached?

Devlin-Mallory: Oh, I see now - I'm not sure why that differs. I could add the base into the cache, unless @ggsmith or anyone else has objections?

ggsmith: For consistency, I think that the base toric variety should also be cached in the ```#s === 1``` case.  @Devlin-Mallory can you also make this change?

Devlin-Mallory: Done!

ggsmith (approved):

mahrud:
Thanks for the fixes!

Issue #1579, opened 2020-11-04
check_11 "SumsOfSquares" fails on i386


d-torrance:
On a build of the `master` branch on a Debian unstable i386 chroot:

```m2
i2 : check_11 "SumsOfSquares"
-* running test 11 of package SumsOfSquares in file:
   ../../Macaulay2/packages/SumsOfSquares.m2:1376:1:
   rerun with: check_11 "SumsOfSquares" *-
 -- making test results
 ulimit -c unlimited; ulimit -t 700; ulimit -m 850000; ulimit -s 8192; ulimit -n 512;  cd /tmp/M2-2519019-0/31-rundir/; GC_MAXIMUM_HEAP_SIZE=400M "/root/M2/M2/BUILD/doug/usr-dist/i686-Linux-Debian-unknown/bin/M2-binary" --int --no-randomize --no-readline --silent --stop --print-width 77 -e 'needsPackage("SumsOfSquares", Reload => true, FileName => "/root/M2/M2/Macaulay2/packages/SumsOfSquares.m2")' <"/tmp/M2-2519019-0/30.m2" >>"/tmp/M2-2519019-0/30.tmp" 2>&1
/tmp/M2-2519019-0/30.tmp:0:1: (output file) error: Macaulay2 exited with status code 1
/tmp/M2-2519019-0/30.m2:0:1: (input file)
M2: *** Error 1
stdio:2:6:(3): error: test #11 of package SumsOfSquares failed

i3 : get "/tmp/M2-2519019-0/30.tmp"

o3 = -- -*- M2-comint -*- hash: 245165013
     --warning: file doc.css not installed yet in package Style
     --warning: file katex not installed yet in package Style

     i1 : --/root/M2/M2/Macaulay2/packages/SumsOfSquares.m2:1376: location of test code
           --lowerBound
              debug needsPackage "SumsOfSquares"

     i2 :     results := checkLowerBound("CSDP")
     -- warning: experimental computation over inexact field begun
     --          results not reliable (one warning given per session)

     o2 = {true, false, true, true, true, true, true}

     o2 : List

     i3 :     assert all(results,t->t=!=false);
     stdio:5:5:(3): error: assertion failed
```
Note that we're using the Debian `coinor-csdp` package.

d-torrance:
The same test is failing on Fedora rawhide:

```m2
i1 : debug needsPackage "SumsOfSquares";
[...]

PR #3444, opened 2024-08-30
Fix check(11, "SumsOfSquares") on Fedora rawhide

d-torrance:
As reported in https://github.com/Macaulay2/M2/issues/1579#issuecomment-2282749223, `check(11, "SumsOfSquares")` is failing on Fedora rawhide due an update in the csdp package.  (For an example of a failing build, see https://github.com/d-torrance/M2-workflows/actions/runs/10609828175/job/29406195683.)

The problem was that it gave up trying to round a real matrix to a rational matrix after 32 attempts.  If we bump this to 33, then it passes.  (For a successful run on Fedora rawhide, see https://github.com/d-torrance/M2/actions/runs/10632956367/job/29476969319.)

Also, a slightly related change: we stop exporting the `Verbosity` symbol in the SemidefiniteProgramming package (which is exported by SumsOfSquares and handles the csdp interface) since it's now exported by Core.

Cc: @diegcif, @tom111

### DIFF
--- a/M2/Macaulay2/packages/SemidefiniteProgramming.m2
+++ b/M2/Macaulay2/packages/SemidefiniteProgramming.m2
@@ -37,7 +37,6 @@ export {
     "changeSolver",
     "criticalIdeal",
 --Method options
-    "Verbosity",
     "Solver",
     "Scaling"
 }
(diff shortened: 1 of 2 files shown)


tom111:
Seems fine, although I don't exactly remember the details of this `MaxRoundTol` variable. Thanks.

Issue #3446, opened 2024-09-01
Undistributed packages that don't load


d-torrance:
_Originally posted by @mahrud in https://github.com/Macaulay2/M2/issues/3419#issuecomment-2323351361_:

> If the undistributed packages don't load, what's the point of keeping them?

I investigated each of the undistributed packages that fail to load so we'd have a starting point for discussion about what to do with them:

- [ ] `CodepthThree`:  Raises an error saying that it has been renamed to `TorAlgebra`.
- [ ] `CustomEngineTests`: Raises an error saying that it has been deprecated.
- [ ] `FastLinearAlgebra`: Raises an error saying that it has been deprecated.
- [ ] `FrobeniusMultiplicities`: Mutable unexported symbol (`k` on line 179), and `tensorF` is documented twice.
- [ ] `MemoryLeaks`: No call to `newPackage`
- [x] `Mysql`: Doesn't export `MysqlResult` (or `MysqlField`) from `Core`, so we get an error trying to assign a synonym on line 18.
- [x] `ParameterSchemes`: Lots of mutable unexported symbols.
- [ ] `Polyhedra2`: Loads `PolyhedralObjects` (see below).
- [x] `PolyhedralObjects`: Empty `Example`entries in documentation.
- [ ] `RisaAsir`: Empty `Example` entry in documentation.
- [ ] `Sage`: See [#1022](https://github.com/Macaulay2/M2/issues/1022).
- [x] `SecondPackage`: `Authors` option is malformed.
- [ ] `ToricCohomology`: Lots of mutable unexported symbols.

PR #3447, opened 2024-09-01
Fix a few undistributed packages

d-torrance:
This fixes some of the low-hanging fruit from #3446 to get some of the undistributed packages to load.


d-torrance:
I added a couple more related commits:

* Fix the html validation bug mentioned in [#3448 (comment)](https://github.com/Macaulay2/M2/pull/3448#issuecomment-2323556698).
* Remove the `supplanted-packages` directory.  It just contained old versions of packages that still exist in the main packages directory.   That's not necessary using git -- we can just check out an old commit if we want to look at an older version of a package.

mikestillman:
@d-torrance OK, I'm finally going to make an attack on PR's and issues.  But I am having trouble with this one, for some reason.  Besides having deleted files (which are likely fine), I am having trouble with libomp not being found.  That works for me now in development, but not this branch (which I though I had merged into development...). Not sure yet what the problem is...

d-torrance:
@mikestillman - I've just rebased it onto `development`, so maybe that will help.  Thanks for taking a look!

.github/workflows/test_build.yml
```diff
@@ -209,7 +209,7 @@ jobs:
               make validate-html
             else
             # otherwise, just validate HTML docs for packages that have been updated
-              PACKAGES=$(git diff --stat origin/development HEAD -- ../../Macaulay2/packages/ | grep -Po "(?<=Macaulay2/packages/)[^/\.]*(?=\.m2|/)" | uniq | xargs)
```
mikestillman, line 212: What is this doing?  Is it now going to try to install these packages?

M2/Macaulay2/packages/undistributed-packages/RisaAsir.m2
```diff
@@ -67,7 +67,6 @@ Headline
   an interface to the Risa/Asir website and system
```
mikestillman, line 67: This package I would like to flesh out, as Risa/Asir has a few very good algorithms

M2/Macaulay2/packages/undistributed-packages/ParameterSchemes.m2
```diff
@@ -243,11 +243,11 @@ parameterIdeal(Ideal,Ideal) := (M,family) -> (
      -- M is a monomial ideal in a polynomial ring
      -- family is the result of a call to 'parameterFamily'
      R := ring M;
-     time G = forceGB gens family;
-     time syzM = substitute(syz gens M, ring family);
-     time eq = compress((gens family * syzM) % G);
+     time G := forceGB gens family;
+     time syzM := substitute(syz gens M, ring family);
```
mikestillman, line 247: This package is perhaps superseded by GroebnerStrata.m2, but in fact there might be some things here that I want to include there.  I had forgotten that this package was here.

mikestillman (approved):
Thanks Doug for doing this!  I think this is ready to go, but I'd like to see what you say about a question I had in the review.  Once you answer those (in any way, really!), lets merge this PR.

.github/workflows/test_build.yml
d-torrance: No -- for every pull request, we do a `git diff` to see if any packages were updated, and if so, validate the html of these packages' docs.  In particular, we look for any changes to an .m2 file of `M2/Macaulay2/packages` or something in a subdirectory of `M2/Macaulay2/packages` to generate a list of packages to check.

However, `undistributed-packages` is a subdirectory of `M2/Macaulay2/packages` but it isn't a package!  So we were trying to validate the html docs of a nonexistent package, leading to occasional failing GitHub builds.  So now we filter this case out.

mikestillman: Ah, thanks, that sounds good!

mikestillman:
Feel free to merge it.  Do you tend to do merge PR's or rebase PR's?

d-torrance:
I usually rebase since that gives us a cleaner looking git history without a bunch of merge commits.

I'm assuming all the tests will pass, but I'll hold off on merging until they do just in case.

PR #3448, opened 2024-09-01
fixing some typos in packages

fchapoton:
just fixing a few typos, found using something like

`egrep -Rw --no-filename -o '[a-z]+ules?' packages/ | sort | uniq -c | sort -rn`


mahrud (approved):

d-torrance:
This PR uncovered a bug in the html validation part of the workflow -- it thinks that `undistributed-packages` and `supplanted-packages` are packages whose documentation needs to be validated...

Issue #3456, opened 2024-09-05
Trouble defining a complex matrix


pzinn:
I thought we'd resolved all these issues involving `Constant` vs `Number`:
```
i1 : matrix {{ ii }}
stdio:1:6:(3): error: common ring not found
```

Issue #3184, opened 2024-04-17
Promotion issues


pzinn:
#3177 let M2 create fraction fields of iterated rings. Even though this works now, there is an issue which is not easily fixed: promotion. For example:
```
i1 : R=QQ[x]

o1 = R

o1 : PolynomialRing

i2 : S=R[y]

o2 = S

o2 : PolynomialRing

i3 : 1/x+1/y
stdio:3:3:(3): error: expected pair to have a method for '+'

i4 : 1/x

     1
o4 = -
     x

o4 : frac R

i5 : 1/y

     1
o5 = -
     y

o5 : frac S
```
Easy, add promotion from `frac R` to `frac S`? but that's not currently possible because one can only promote from one ring to another if the former is in the base rings of the latter, and it's not the case here:
```
i7 : (frac S).baseRings

o7 = {ZZ, QQ, R, S}
```
Basically, the current system is based on a "linear chain" of promotion whereas in the situation above we have a graph of the form
![Screenshot from 2024-04-16 22-18-11](https://github.com/Macaulay2/M2/assets/4632711/af0aad6a-a0f6-417e-a623-c3b84e11b66e)
One possibility would be to replace the current test of "promotability" which is pretty much what I wrote above with a more flexible
```
isPromotable = (R,S) -> S#?(promote,R,S);
```
This would have other advantages, like define "natural" maps between rings say A and B which should be applied automatically when multiplying an element of A and an element of B. (typically situation is when A is a subring of B, since M2 doesn't have a concept of subring).
It wouldn't solve every problem (what if one tries to add an element of `S` and an element of `frac R` in the situation above?) but it would be a start.

pzinn:
I'm working on this. Here's another example that fails and should work:
```
i1 : R=QQ[x,y]

o1 = R

o1 : PolynomialRing

i2 : S=QQ[z]

o2 = S

o2 : PolynomialRing

i3 : T=R**S

o3 = T

o3 : PolynomialRing

i4 : R_0*T_2
stdio:4:3:(3): error: expected pair to have a method for '*'
```

pzinn:
also, some of these remarks also apply to `lift`.
incidentally, shouldn't `liftable` be renamed `isLiftable`?

mahrud:
> incidentally, shouldn't `liftable` be renamed `isLiftable`?

This was brought up before too, I think. I think it would be good to deprecate `liftable` in favor of `isLiftable`.

d-torrance:
[...]

PR #3457, opened 2024-09-05
Add ring and promote methods for Constant objects

d-torrance:
Add support for creating matrices with Constant entries.  Also add a unit test.

### Before

```m2
i1 : matrix {{ii}}
stdio:1:6:(3): error: common ring not found
```

### After
```m2
i1 : matrix {{ii}}

o1 = | ii |

                1         1
o1 : Matrix CC    <-- CC
              53        53
```

Closes: #3456


pzinn:
looking good. it still misses some cases, though maybe we're reaching the limit of the `promote` system, cf #3184:
```
i1 : f=map(CC[x],CC[y,z],{x,ii})
stdio:1:5:(3): error: no method found for applying promote to:
     argument 1 :  ii (of class Constant)
     argument 2 :  CC  [x]
                     53
```

d-torrance:
That case was straightforward enough.  I added `promote(Constant, RingElement)`, so now your example works:

```m2
i1 : map(CC[x],CC[y,z],{x,ii})

o1 = map (CC  [x], CC  [y..z], {x, ii})
            53       53

o1 : RingMap CC  [x] <-- CC  [y..z]
               53          53
```

pzinn (approved):

PR #3462, opened 2024-09-07
Remove unused "setup" functions in interpreter

d-torrance:
This is based on a few commits that I'd originally proposed in #3419.

We remove some of the `setup` family of functions in `common.d` (which are used for associating a symbol with its function) that weren't being used anywhere:

* All of them that accept a `Word` argument.
* `setup(SymbolClosure, unop, unop)` (for unary operators that are both prefix and postfix, of which we have none)
* `setupfun(string, unop)`

We also remove `setup(SymbolClosure, unop)` since it was identical to `setupop`.

### DIFF
--- a/M2/Macaulay2/d/actors.d
+++ b/M2/Macaulay2/d/actors.d
@@ -810,7 +810,7 @@ export notFun(rhs:Code):Expr := (
      else if a == True then False
      else if a == False then True
      else unarymethod(a,notS));
-setup(notS,notFun);
+setupop(notS,notFun);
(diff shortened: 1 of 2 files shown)


mahrud:
I probably won't have enough time to do this review for a few weeks, but then I'd like to take a better look at how and where setup-like functions are used, mainly because before removing them I want to make sure we can't make better simplifications elsewhere using them.

mahrud:
I'd like to make further simplifications in this area, but probably best not to keep this PR until that happens.

mahrud (approved):

Issue #3468, opened 2024-09-11
Order in keys and values for identical hash tables


mahrud:
Shouldn't the order of keys and values of a hash table depend only on the hash? Why can I get two identical hash tables with different keys and values?
```m2
i1 : partition(even, {1,2,3,4,5}, {true, false})

o1 = HashTable{false => {1, 3, 5}}
               true => {2, 4}

o1 : HashTable

i2 : (keys oo, values oo)

o2 = ({true, false}, {{2, 4}, {1, 3, 5}})

o2 : Sequence

i3 : partition(even, {1,2,3,4,5}, {false, true})

o3 = HashTable{false => {1, 3, 5}}
               true => {2, 4}

o3 : HashTable

i4 : (keys oo, values oo)

o4 = ({false, true}, {{1, 3, 5}, {2, 4}})

o4 : Sequence

i5 : o1 === o3

o5 = true

i6 : keys o1 === keys o3

o6 = false

i7 : values o1 === values o3

o7 = false
```

Of course, I see that the buckets are different:
```m2
i13 : buckets o1

o13 = {{}, {(true, {2, 4}), (false, {1, 3, 5})}, {}, {}}

o13 : List

i14 : buckets o3

o14 = {{}, {(false, {1, 3, 5}), (true, {2, 4})}, {}, {}}

o14 : List
```
So is this just because the first element, `1`, in the call to `partition` was odd? I don't think the output hash table itself should be affected by this.

mahrud:
By the way, there's some code depending on this behavior, so we should be careful about changing it:
https://github.com/Macaulay2/M2/blob/1fd238210e9a37ddaec3b700ccd22f06ff5f28f5/M2/Macaulay2/m2/document.m2#L521

pzinn:
simpler example:
```
i1 : h1=hashTable{false=>1,true=>2}

o1 = HashTable{false => 1}
               true => 2

o1 : HashTable

i2 : h2=hashTable reverse{false=>1,true=>2}

o2 = HashTable{false => 1}
               true => 2

o2 : HashTable

i3 : keys h1

o3 = {true, false}

o3 : List

i4 : keys h2

o4 = {false, true}

o4 : List

i6 : h1===h2

o6 = true
```

d-torrance:
For this particular case, it's because the hash table is small (the minimum size of 4 buckets), and so `true` and `false` get stuffed into the same bucket:

```m2
i1 : hash true % 4

o1 = 1

[...]

PR #3471, opened 2024-09-11
Ensure that "true" and "false" won't go in same bucket in hash table

d-torrance:
This doesn't completely address #3468, but it at least does when the keys of the hash table are `true` and `false`.

### Before:

```m2
i2 : buckets tally {true, false}

o2 = {{}, {(false, 1), (true, 1)}, {}, {}}

o2 : List
```

### After

```m2
i2 : buckets tally {true, false}

o2 = {{(false, 1)}, {(true, 1)}, {}, {}}

o2 : List
```

~~We also refactor the code mentioned in https://github.com/Macaulay2/M2/issues/3468#issuecomment-2342524704 so it doesn't rely on the order of the elements returned by `values`.~~ *Edit:* I removed this commit because it's already dealt with in #3469.

*Edit no. 2:*  I adjusted the hash of `false` to be congruent to 2 (mod 4) instead of 0 (mod 4) so that the `true` bucket would go before the `false` bucket and the builds wouldn't fail w/o #3469.


mahrud:
What if we treated true and false as 0/1 like in C?

mahrud:
I think just wait until the other one is merged first.

mahrud (approved):

PR #3472, opened 2024-09-11
Fix error location for functions

### LINKED ISSUE #3470: Incorrect error location
This error location is not helpful, nor correct, I believe.
```m2
f = (a,b) -> (
    --
    a -- <<< line 3
    --
    )

g = x -> (
    --
    f x
    --
    )

end
restart
needs "test.m2"

g 1 -- test.m2:3:4:(3): error: expected 2 arguments but got 1
```
cc: @pzinn

mahrud:
- **fixes #3470**
- **aligned code for codePosition**

TODO: needs a test.


pzinn (approved):

mahrud:
I fixed the changed test. By the way, I think `locate FunctionBody` is wrong, too. Isn't it?

pzinn:
I think we discussed this already. Yes, one could change `locate FunctionBody` to be literally the body of the function (i.e., no arguments). But that would possibly break the test you've just edited, I believe.

mahrud:
> But that would possibly break the test you've just edited, I believe.

I think of the test as a way to find out when something changes unintentionally, not to avoid changing it at all.

mahrud:
I think there's a semantic issue that should be fixed: if `functionCode` in the interpreter is truly the equivalent of `FunctionBody` in the top-level, then it should _not_ know about the information in `functionDescription`.

pzinn:
My point was, if you change the `position` of `functionCode`, then it will undo the fix to the issue this PR is for. But you did something in your last commit: you changed the `locate` of `FunctionBody`. The problem with this is that it breaks the correspondence of d/m2 types: now the location of a `FunctionBody` is no longer that of the underlying `functionCode`, but rather that of the top of the tree structure of its code. I'm not sure I agree with that.

mahrud:
> you changed the `locate` of `FunctionBody`.

That's exactly what I said was wrong above, so I fixed it. Now these are equivalent:
```m2
M2                     <-> d
locate FunctionClosure <-> locate FunctionClosure
locate FunctionBody    <-> locate functionCode
```
Does this break something?

I think eventually it would be good to actually change the value of `position` sitting inside `functionCode` and remove the extra data, which should be stored in `FunctionClosure`. (That is, assuming we want `functionCode` to be `FunctionBody` in the top level. We could also forget about this correspondence, but I don't think that's what you're advocating)

pzinn:
the point is, the line in `debugging.dd` that used to say "the location of a `functionCode` is its `position`", as should be, now says "the location of a `functionCode` is *not* its position but rather the position of the first code you meet inside it", which is not very logical.

mahrud:
> the first code you meet inside it

What do you mean by "first" code? It's the location of the body of the `functionCode`, which is what I would expect when I type `locate functionBody f`. Why is this not logical?

mahrud:
I would in fact argue that the previous behavior, which returned the same position was not logical, given the clear semantic difference between a function closure and a function body.
```m2
    is f:FunctionClosure do locate(f.model.position)
    is f:functionCode    do locate(f.position)
```
The correct solution would be to have the position of a function closure in `f` itself and the position of a function body should be the position of `f.body.`

pzinn:
once again, I'm not disagreeing with having a different position for `functionCode` and `FunctionClosure`. But that should be done by changing the `position` field of `functionCode`, not by hacking the locate function. I'm willing to make that change myself if you give me a few.

mahrud:
Okay, I think we're in agreement about the permanent solution then. I'm happy to wait for your PR, or we can merge this PR a temporary solution, then change the position in functionCode in your PR. Either way is fine with me.

pzinn:
Actually I had a quick look and my suggested solution is not going to be easy.
The problem is, the `FunctionClosure` (i.e., the code with the frame) is only created when the function code is *evaluated*, not when it's parsed (as should be). So at that stage the only position available is that of the `functionCode`. So making `FunctionClosure` and `functionCode` have different positions, as natural as that may be, would require some weird stuff.
I'll think about it a little more.

pzinn:
Incidentally, anyone know the difference between `wrongModel1` and `WrongNumArgs` in `evaluate.d`? looks to me like they're doing the same thing...

mahrud:
Yeah, I also don't think it would be easy, and at least in the short term with this PR the behavior is correct (even if which position is stored where is not ideal).

pzinn:
My only suggested change is this: 38fbf77919f1e2818a152f6d26cf69b17c137a37
before:
```
i1 : f= x -> (
     x+1
     )

o1 = f

o1 : FunctionClosure

i2 : locate f

o2 = stdio:1:3-3:1

o2 : FilePosition

i3 : locate functionBody f

o3 = stdio:2:0-2:3

o3 : FilePosition

i4 : f=(a,b)->1

o4 = f

o4 : FunctionClosure

i5 : f 2
stdio:6:7:(3): error: expected 2 arguments but got 1
```
after:
```
i1 : f= x -> (
     x+1
     )

o1 = f

o1 : FunctionClosure

i2 : locate f

o2 = stdio:1:3-3:1

o2 : FilePosition

i3 : locate functionBody f

o3 = stdio:1:8-3:1

o3 : FilePosition

i4 : f=(a,b)->1

o4 = f

o4 : FunctionClosure

i5 : f 2
stdio:6:2:(3): error: expected 2 arguments but got 1
```
note how the location of the body is more accurate. and also the last error is positioned more correctly.

mahrud:
I agree with the second change, but not the first one. The function body is just a piece of code, which in this case is `x+1`. The surrounding parentheses aren't relevant. This is also true everywhere else in your code:
```m2
i6 : P = pseudocode(() -> if (    true    ) then (     1     ) else (     0     ))

o6 = function restargs false
              numparms 0
              framesize 0
              frameID 8896
              body if global-fetch 221
                      then 1
                      else 0

o6 : PseudocodeClosure

i7 : locate P_4_0, locate P_4_1, locate P_4_2

o7 = (stdio:6:29-6:33, stdio:6:50-6:51, stdio:6:69-6:70)
```

mahrud:
I found a different way to determine the position for different types of code, by centralizing it in `treePosition`. I think it's simpler and more readable.

Could you test if this changes any error results in your tests?

pzinn:
I don't remember why I didn't do the positioning at the time of parsing as you suggest. It does seem to work equally well.
Let me test this some more.

pzinn:
I think the reason I didn't do it this way is, and correct if I'm wrong, this requires repeatedly going through the tree for each piece of the code; so if you have nested things like (a+(a+... (n times) a+()))) the number of operations will grow like n^2?

mahrud:
I think you're right that it's less efficient (not sure exactly the O-complexity), but M2's nesting limit is ~300 so it's a minuscule difference, which I haven't noticed when running this locally (note that this may slow down _parsing_ by a small amount, but _execution_ time is unaffected). I think the readability of code is worth the less than optimal algorithm in this case.

mahrud:
I think I should not add any more commits to this PR, but I have further changes, so could you review this again?

pzinn (approved):

PR #3473, opened 2024-09-12
Update VNumber.m2

EmanueleSgroi:
We made a minor change to the function "soc"

### DIFF
--- a/M2/Macaulay2/packages/VNumber.m2
+++ b/M2/Macaulay2/packages/VNumber.m2
@@ -3,7 +3,7 @@
 newPackage(
     "VNumber",
     Version => "1.0",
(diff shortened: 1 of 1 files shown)


M2/Macaulay2/packages/VNumber.m2
```diff
@@ -3,7 +3,7 @@
 newPackage(
     "VNumber",
     Version => "1.0",
```
d-torrance, line 5: Did you want to increase the version number as well?

EmanueleSgroi: If it is not a problem, we can leave the current version. It was just a formal fix.

Issue #1229, opened 2020-06-04
add the possibilty to give method, eg. random further options. In case of random I would like to specify my own favorate distribution on e.g. QQ


foschreyer:

DanGrayson:
If we change
```
random = method(Options => {
	  MaximalRank => false,
	  Density => 1.,
	  UpperTriangular => false,
	  Height => 10
	  })
```
to
```
random = method(Options => true)
```
then each method function, when installed, can offer its own list of
optional arguments.

DanGrayson:
Frank, how would you like to specify your favorite distribution?  By writing entirely new code for `random QQ`, I assume.  That doesn't sound like an option to existing code.

foschreyer:
Dan, my first choice would be instead of rationals to use integers in bounded range  equidistributed (with zero median). At least for application in projective geometry I do not see any reason to use denominators. I can image that draw integers basically with a Gaussian distribution could also be interesting.

Of course you are right, that is more than option but would require a new instance
random(QQ,Distribution):= ...

DanGrayson:
Well, one thing we could do trivially is to fix this so no denominators appear:
```
i8 : random (5,QQ[x,y])

     2 5   5 4      3 2    2 3   4   4   3 5
o8 = -x  + -x y + 4x y  + x y  + -x*y  + -y
     7     7                     7       4

o8 : QQ[x, y]
```
It seems to me that this, however, should still return denominators:
```
i9 : for i to 10 list random QQ

         2      9  1  7        10   7
o9 = {1, -, 2, --, -, -, 1, 1, --, --, 1}
         5     10  9  4         7  10

o9 : List
```

As for `random(QQ,Distribution):= ...`, it's hard to imagine the code in `...` doing anything nontrivial, so there doesn't seem to be any reason to have it.

d-torrance:
> **Written by Claude** (Claude Opus 5, via Claude Code), not by @d-torrance, whose account posted it -- please weigh it accordingly.

@foschreyer — this is available now, through the
[`Probability`](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Probability/html/) package,
which ships with Macaulay2. It lets you define whatever distribution you want and draw from it.

Your stated first choice in the thread above was *"integers in bounded range equidistributed (with
zero median)"*, and *"I do not see any reason to use denominators"*. That is one line:

```m2
i1 : needsPackage "Probability";

[...]

PR #3478, opened 2024-09-14
Probability package updates

### LINKED ISSUE #999: random QQ
Frank Schreyer would like "random QQ" to yield negative numbers about half the time (currently, none are provided) and that the various numbers occur with equal probability (currently, fractions of small height occur more often).

Moreover, when working with polynomial rings over the rationals, it seems senseless to provide any denominators:
```

i21 : R=QQ[x,y]

o21 = R

o21 : PolynomialRing

i22 : random (10, R)

      3 10   5 9    1 8 2   5 7 3   1 6 4   5 5 5   9 4 6   1 3 7    2 8      9   10 10
o22 = -x   + -x y + -x y  + -x y  + -x y  + -x y  + -x y  + -x y  + x y  + x*y  + --y
      5      7      8       3       7       2       4       7                      9

o22 : R
```

d-torrance:
Currently, the probability distribution of `random(QQ)` is kind of unusual.  If _X_ and _Y_ both have the discrete uniform distribution on {1,...,_h_} for some positive integer _h_, then we get a sample from the distribution of _X_/_Y_.   This maybe isn't great.  For example, _P_(_X_/_Y_ = 1) = 1/_h_.  Indeed:

```m2
i1 : tally apply(100, i -> random(QQ, Height => 3))

o1 = Tally{1 => 29}
           2 => 12
           3 => 11
           1
           - => 9
           3
           1
           - => 12
           2
           2
           - => 14
           3
           3
           - => 13
           2
```

Instead, we propose sampling from the standard normal distribution and rounding the results to the nearest rational number with denominator bounded by _h_ (using the [Farey sequence](https://en.wikipedia.org/wiki/Farey_sequence)).  I think I mentioned this idea recently in some discussion, but I can't find it.  After the change:

```m2
i1 : tally apply(100, i -> random(QQ, Height => 3))

o1 = Tally{-1 => 10 }
           -2 => 1
           0 => 12
           1 => 6
           2 => 2
             1
           - - => 5
             3
             1
           - - => 8
             2
             2
           - - => 12
             3
             3
           - - => 3
             2
             4
           - - => 6
             3
             5
           - - => 3
             3
             7
           - - => 1
             3
           1
           - => 6
           2
           1
           - => 15
           3
           2
           - => 3
           3
           3
           - => 3
           2
           4
           - => 2
           3
           7
           - => 1
           2
           7
           - => 1
           3
```

This PR also has a few other related changes:

* Make the Farey sequence approximation routine available at top-level (but unexported).
* Add a function to the engine that returns a random `RR` from the standard normal distribution and use it in the `Probability` package.
* Improve the `Probability` package documentation.

This possibly closes #999, although that suggests using a uniform distribution.

### DIFF
--- a/M2/Macaulay2/d/interface.dd
+++ b/M2/Macaulay2/d/interface.dd
@@ -63,6 +63,12 @@ export rawRandomRR(e:Expr):Expr := (
      else toExpr(Ccode(RR, "rawRandomRR(", toULong(prec.v), ")"))
      else WrongArgZZ());
 setupfun("rawRandomRR",rawRandomRR);
+export rawRandomRRNormal(e:Expr):Expr := (
+     when e
+     is prec:ZZcell do if !isULong(prec.v) then WrongArgSmallInteger()
+     else toExpr(Ccode(RR, "rawRandomRRNormal(", toULong(prec.v), ")"))
+     else WrongArgZZ());
+setupfun("rawRandomRRNormal",rawRandomRRNormal);
 export rawRandomCC(e:Expr):Expr := (
      when e
      is prec:ZZcell do if !isULong(prec.v) then WrongArgSmallInteger() 
--- a/M2/Macaulay2/e/interface/random.cpp
+++ b/M2/Macaulay2/e/interface/random.cpp
@@ -104,6 +104,15 @@ gmp_RR rawRandomRR(unsigned long precision)
   return moveTo_gmpRR(result);
 }
 
+gmp_RR rawRandomRRNormal(unsigned long precision)
+/* returns a normally distributed random real with the given precision */
+{
+  mpfr_ptr result = getmemstructtype(mpfr_ptr);
+  mpfr_init2(result, precision);
+  mpfr_nrandom(result, state, MPFR_RNDN);
+  return moveTo_gmpRR(result);
+}
+
 gmp_CC rawRandomCC(unsigned long precision)
 /* returns a uniformly distributed random complex in the box [0.0,0.0],
  * [1.0,1.0] */
--- a/M2/Macaulay2/e/interface/random.h
+++ b/M2/Macaulay2/e/interface/random.h
@@ -40,6 +40,9 @@ gmp_RR rawRandomRR(unsigned long prec);
 /* returns a uniformly distributed random real with the given precision, in
  * range [0.0,1.0] */
 
+gmp_RR rawRandomRRNormal(unsigned long prec);
+/* returns a normally distributed random real with the given precision */
+
 gmp_CC rawRandomCC(unsigned long prec);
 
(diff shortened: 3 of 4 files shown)


mahrud:
Also see https://github.com/Macaulay2/M2/issues/1229. Is there a standard data type for distributions?

d-torrance:
> Is there a standard data type for distributions?

There is in the Probability package -- see https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Probability/html/___Probability__Distribution.html

mahrud:
I know it'll be a lot of work, but I think it would be useful down the road if I could pass a distribution as an option to random.

d-torrance:
> I know it'll be a lot of work, but I think it would be useful down the road if I could pass a distribution as an option to random.

That's already possible in the `Probability` package -- see [random(ProbabilityDistribution)](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Probability/html/_random_lp__Probability__Distribution_rp.html).  

I suppose you're suggesting things like random ring elements and random matrices?  That would be pretty easy to implement in the package, I think, by adding more methods like `random(Module, Module, ProbabilityDistribution)`.

mahrud:
I guess I want everything to be compatible, so random(QQ, ...) can be changed to random(QQ^n, ...) or random(QQ^m, QQ^n, ...), and further random(S = QQ[x], d, ...) or random(S^n, S^{...}, ...) would also pull coefficients from the same distributions. I wouldn't be opposed to preloading the Probability package if necessary.

More specifically about this PR, why did you choose normal distribution over uniform distribution?

d-torrance:
Good ideas!  I'm going to remove most of the engine stuff from this PR (keeping the `Probability` package updates) and mull it over a bit for a future PR.

I went with normal because of the central limit theorem -- it shows up so frequently in applications!  And also we don't have to worry about the support -- what interval would we use for a uniform distribution?  But if we were to change the default for `random QQ`, then we should probably also change it for `RR` and `CC` as well.

d-torrance:
I've simplified this to now just add `rawRandomRRNormal` to the engine, use it in the Probability package, and improve some documentation.

mahrud:
> what interval would we use for a uniform distribution?

I think since `random ZZ` uses a uniform distribution in [0,10), a uniform distribution by default for QQ/RR/CC makes the most sense, and I would imagine the support should be fixed to be the interval [0,1), allowing users to scale or translate it as they like. Also, I liked using the Farey sequence here!

d-torrance:
That makes sense!  I'll work on a revised `random(QQ)` in another PR.

mahrud (approved):
Not necessarily in this PR, but it would be good to rename rawRandomRR to rawRandomRRuniform for clarity.

Issue #2089, opened 2021-05-20
Crash on random with Height => 0


mahrud:
```m2
Macaulay2, version 1.17.2
with packages: ConwayPolynomials, Elimination, IntegralClosure, InverseSystems,
               LLLBases, MinimalPrimes, PrimaryDecomposition, ReesAlgebra,
               Saturation, TangentCone

i1 : random(ZZ^2, ZZ^2, Height => 0)
/home/linuxbrew/.linuxbrew/bin/M2: line 2: 1349503 Floating point exception(core dumped) LD_LIBRARY_PATH=`dirname "$0"`/../lib/Macaulay2/lib:$LD_LIBRARY_PATH `dirname "$0"`/M2-binary "$@"
```

DanGrayson:
```
(gdb) where
#0  raise (sig=<optimized out>) at ../sysdeps/unix/sysv/linux/raise.c:49
#1  0x00007ffff6e08a4d in __gmp_exception () from /lib/x86_64-linux-gnu/libgmp.so.10
#2  0x00007ffff6e08a94 in __gmp_divide_by_zero () from /lib/x86_64-linux-gnu/libgmp.so.10
#3  0x00007ffff6e1c69c in __gmpz_urandomm () from /lib/x86_64-linux-gnu/libgmp.so.10
#4  0x0000555555a6ea7d in rawRandomInteger (maxN=0x0) at /home/dan/src/M2/M2.git/M2/Macaulay2/e/interface/random.cpp:59
#5  0x00005555557b9e32 in RingZZ::random (this=0x7ffff3659c80) at /home/dan/src/M2/M2.git/M2/Macaulay2/e/ZZ.cpp:69
#6  0x00005555559287a0 in Matrix::random (R=0x7ffff3659c80, r=2, c=2, density=1, special_type=0) at /home/dan/src/M2/M2.git/M2/Macaulay2/e/matrix.cpp:718
#7  0x0000555555a6112d in IM2_Matrix_random (R=0x7ffff3659c80, r=2, c=2, fraction_non_zero=1, special_type=0, preference=0)
    at /home/dan/src/M2/M2.git/M2/Macaulay2/e/interface/matrix.cpp:207
#8  0x0000555555c2ed0c in interface_rawRandomConstantMatrix (e=0x7ffff2cafd40) at /home/dan/src/M2/M2.git/M2/Macaulay2/d/interface.dd:2209
[...]

Issue #1993, opened 2021-03-15
memoize GF


jieaosong:
Hi! I saw this in the [documentation](https://faculty.math.illinois.edu/Macaulay2/doc/Macaulay2-1.17/share/doc/Macaulay2/Macaulay2Doc/html/_random_lp__Type_rp.html) which I guess is not the desired behavior.
```macaulay2
i3 : tally for i to 100 list random GF 11

o3 = Tally{-1 => 1}
           -1 => 1
           -1 => 1
           -1 => 1
           -1 => 1
           -1 => 1
           -1 => 1
           -1 => 1
           -1 => 1
           -2 => 1
           -2 => 1
           -2 => 1
           -2 => 1
           -2 => 1
           -2 => 1
           -2 => 1
           -2 => 1
           -2 => 1
           -2 => 1
           -2 => 1
           -2 => 1
           -2 => 1
           -2 => 1
           -2 => 1
           -3 => 1
...
```
This could be fixed by making `GF` memoized. I saw this trick in the package "Cyclotomic", where the same number would always give the exact same cyclotomic field.

DanGrayson:
Probably the reason we didn't memoize it originally was to allow for garbage collection of unused Galois fields.  One can imagine experiments that make a lot of them.  If that's not a concern, memoization might be good now.  And it might be good to make `GF 11` and `ZZ/11` give the same answer.

8d1h:
That would indeed be nice. Also I think currently the arithmetic over `ZZ/11` is faster than `GF 11`.
```macaulay2
i23 : ZZ/11[x]; time for i to 10000 do factor x^2-1;
     -- used 0.451269 seconds  
                                                   
i25 : GF 11[x]; time for i to 10000 do factor x^2-1;
     -- used 2.07047 seconds
```
And `isFinitePrimeField GF 11` returns `false` since it checks with `isQuotientRing`.

mahrud:
A possible solution is https://github.com/Macaulay2/M2/commit/0471672b94c3d96be80d04de33c4c4814637e269.

PR #3481, opened 2024-09-17
Update random(QQ) to use standard uniform distribution

d-torrance:
This is split off from #3478.  I think it makes sense to keep them separate since that deals with the Probability package and this deals with how rationals are randomly generated in the engine.  The big change to the commits that were removed from that PR is the proposed probability distribution (no longer normal).

Rather than using the quotient of two discrete random variables with the uniform distribution on {1, ..., h}, we use the continuous uniform distribution on [0, 1], but rounded to the nearest rational number with denominator bounded by the number given by the `Height` option (default 10).  For example:

```m2
i1 : random(QQ^3, QQ^3, Height => 3)

o1 = | 2/3 2/3 1/3 |
     | 1   1   2/3 |
     | 1   2/3 1/2 |

              3       3
o1 : Matrix QQ  <-- QQ

i2 : random(QQ^3, QQ^3)

o2 = | 1/8 1   7/8  |
     | 1/3 5/6 9/10 |
     | 1   3/4 1/7  |

              3       3
o2 : Matrix QQ  <-- QQ

i3 : random(QQ^3, QQ^3, Height => 100)

o3 = | 61/65 33/59 3/16  |
     | 15/26 7/26  31/39 |
     | 3/26  7/48  62/95 |

              3       3
o3 : Matrix QQ  <-- QQ
```


M2/Macaulay2/packages/Macaulay2Doc/functions/random-doc.m2
```diff
@@ -77,12 +77,14 @@ Node
...
-      is @TO "ZZ"@ then the integers returned are in the range @TT "[0, h)"@;
-      for @TO "QQ"@ the numerator and denominator are in the range @TT "[1, h]"@.
+      is @TO "ZZ"@ then the integers returned are in the range @TT "[0, h)"@.
+      If @TT "T"@ is @TO QQ@, then the results are drawn from the standard
+      uniform distribution and rounded to the nearest rational number with
+      denominator bounded by @TT "h"@.
     Example
       random RR
       random CC_100
-      tally for i to 100 list random GF 11
```
mahrud, line 85: I hope GF is memoized, no?

mahrud (approved):

M2/Macaulay2/packages/Macaulay2Doc/functions/random-doc.m2
d-torrance: No, it's not

d-torrance:
I think the build failures are coming from examples that use `random` in some way and that have some nonzero probability of failing.  This PR is changing things so that we're only calling the pseudorandom number generator once instead of twice for each call to `rawRandomQQ`, we're getting different random numbers and getting some unexpected results.

In particular, `CoincidentRootLoci::realrank` sometimes calls QEPCAD, which isn't available in the macOS builds, and `GraphicslModelsMLE::solverMLE` sometimes results in division by zero.

I was able to reproduce both of these pretty quickly using Macaulay2 1.24.05 on Debian unstable, i.e., *before* the proposed changes:

```m2
i1 : needsPackage "CoincidentRootLoci";

i2 : realrank randomBinaryForm(6, 4, 4)

o2 = 4

i3 : realrank randomBinaryForm(6, 4, 4)

o3 = 4

i4 : realrank randomBinaryForm(6, 4, 4)
stdio:4:10:(3):[1]: error: the ideal is not apolar

i5 : realrank randomBinaryForm(6, 4, 4)

o5 = 4

i6 : realrank randomBinaryForm(6, 4, 4)

o6 = 4

i7 : realrank randomBinaryForm(6, 4, 4)

o7 = 4

i8 : realrank randomBinaryForm(6, 4, 4)
/bin/sh: 1: qepcad: not found
stdio:8:1:(3): error: error occurred while executing QEPCAD. Please make sure that it is installed and configured correctly.
```

```m2
i1 : needsPackage "GraphicalModelsMLE";
-- storing configuration for package Graphs in /root/.Macaulay2/init-Graphs.m2
-- storing configuration for package NumericalAlgebraicGeometry in /root/.Macaulay2/init-NumericalAlgebraicGeometry.m2
-- storing configuration for package Bertini in /root/.Macaulay2/init-Bertini.m2

i2 : G = mixedGraph(digraph {{1,2},{1,3},{2,3},{3,4}},bigraph {{3,4}});

i3 : U = matrix{{6.2849049, 10.292875, 1.038475, 1.1845757}, {3.1938475, 3.2573, 1.13847, 1}, {4/5, 3/2, 9/8, 3/10}, {10/7, 2/3,1, 8/3}};

                4         4
o3 : Matrix RR    <-- RR
              53        53

i4 : solverMLE(G,U,RealPrecision=>10)

o4 = (1.83311, | 4.52904   7.81123   -.0250185 .309951   |, 2)
               | 7.81123   14.3732   -.0382332 .473665   |
               | -.0250185 -.0382332 .00337164 -.0418254 |
               | .309951   .473665   -.0418254 .742627   |

o4 : Sequence

i5 : solverMLE(G,U,RealPrecision=>10)
stdio:5:1:(3): error: atttempt to divide by zero
```
I'm thinking of replacing the failing examples with canned ones for the time being and also opening issues for both of these examples.  Does that sound reasonable?

mahrud:
Instead of canning the examples, try setting the random seed to something that prevents failure.

M2/Macaulay2/packages/Macaulay2Doc/functions/random-doc.m2
mahrud: I would prefer leaving this line unchanged and, separately, memoize GF.

d-torrance:
For now, I've just added a `setRandomSeed` line to the offending examples, but I wonder if that might be confusing to a user reading the documentation?

Maybe at some point we should add some way of setting the random seed for a particular example (or test)?

M2/Macaulay2/m2/galois.m2
```diff
@@ -117,7 +117,9 @@ findGalois(ZZ,ZZ) := RingElement => opts -> (p,n) -> (
      else error (opts.Strategy | " is not a valid argument for' Strategy' option")
      )
 
-GF(ZZ,ZZ) := GaloisField => opts -> (p,n) -> (
+GF(ZZ,ZZ) := GaloisField => (
+    memo := new CacheTable;
+    opts -> (p,n) -> memo#(p, n, opts) ??= (
```
mahrud, line 122: Does `memoize` ignore options?

mahrud (approved):

M2/Macaulay2/m2/galois.m2
d-torrance: Yeah, it doesn't work well here.  If I use `memoize(opts -> (p, n) -> ...`, then it uses the function `(p.n) -> ...` as the keys for the memo instead of the actual input.

```m2
i1 : scan({2, 3, 5, 7, 11}, GF)

i2 : peek memoizeValues lookup(GF, ZZ, ZZ)

o2 = MutableHashTable{OptionTable{PrimitiveElement => FindOne} => FunctionClosure[../src/macaulay2/M2/M2/Macaulay2/m2/galois.m2:120:44-129:6]}
                                  SizeLimit => 10000
                                  Strategy => null
                                  Variable => null

```

mahrud:
> For now, I've just added a `setRandomSeed` line to the offending examples, but I wonder if that might be confusing to a user reading the documentation?
> 
> Maybe at some point we should add some way of setting the random seed for a particular example (or test)?

This kind of situation is fairly rare, so I think it's fine. Also, I think it's helpful for users to see that they can produce the results identical to the documentation.

mahrud:
I'm happy with this PR, but given that it's changing the behavior of `random` rather significantly, I think it would be good to have a second opinion from @mikestillman or @antonleykin.

d-torrance:
> I'm happy with this PR, but given that it's changing the behavior of `random` rather significantly, I think it would be good to have a second opinion from @mikestillman or @antonleykin.

Yes, I agree!

mahrud:
While we're here, could you also add something like `random RR` for `QQ`? e.g.
```m2
random QQ := QQ => opts -> x -> x * random(QQ, opts)
```
A problem with this method is that the height is then skewed (e.g. `6 * random(QQ, Height => 3)` always gives integers), so maybe we want `random x + random(QQ, opts)` instead? I'm not sure what's the best approach.

d-torrance:
I pushed a commit adding `random(QQ)` by first calling `random(RR)` and then rounding it.  That seems to work pretty well:

```m2
i1 : apply(20, i -> random(10/1))

      19  11  59  25  7  5  17  89  4  8  33  31  31  11  53     17  64  31  53
o1 = {--, --, --, --, -, -, --, --, -, -, --, --, --, --, --, 4, --, --, --, --}
      10   5   6   8  6  2   3  10  7  5   4  10   4   3   9      6   7   6   6

o1 : List
```

d-torrance:
I went ahead and added more `random` methods, so now pretty much any `Number` object or pair of `Number` objects should so something sensible.

For the complex numbers, if we give it $a + bi$, it will give us something in the rectangle $[0, a]\times [0, b]$ on the complex plane, and if we give it $a + bi$ and $c + di$, it will give us something in $[a,c]\times[b, d]$.  (The order on the endpoints may be reversed if necessary.)

```m2
i1 : random(1, pi)

o1 = 1.47764022754653

o1 : RR (of precision 53)

i2 : random(1/2, ii)

o2 = .398515040066406+.230213567526911*ii

o2 : CC (of precision 53)
```

mahrud:
> I went ahead and added more `random` methods, so now pretty much any `Number` object or pair of `Number` objects should do something sensible.

I don't like the "pair of numbers" methods and would even advocate for deprecating `random(ZZ, ZZ)`. They overcrowd the methods on `random` for very little gain, since they are simply for non-integers `random(min, max) = min + max * random class max` or `min + random max` for integers. The reason I wanted `random QQ` is that otherwise getting the height correct is nontrivial.

d-torrance:
Fair enough -- I'll remove them. It was redundant with `random uniformDistribution(a, b)` from the Probability package anyway

d-torrance:
I'm wondering if maybe the support of `random(QQ)` should be [0, opts.Height] instead of [0, 1]? That would keep it consistent with the current behavior, and might cut down on the random seed failures since I think now there's a higher probability of singular random rational matrices, etc, with the smaller support

mahrud:
Yeah, maybe. Height of a rational number a/b is usually defined as max(a,b) so you're right that that would make more sense. I guess I'm just annoyed that Height has different effects for ZZ and QQ and seemingly no effect for RR or CC. Can we make these consistent somehow and document them all in the same page?

Also, should `random CC` give a random complex number on the unit circle instead of a random complex number in the unit square? Not sure which is preferable.

d-torrance:
I've updated the support so that `random QQ` returns a random rational number in [0, opts.Height].  At least locally, this fixes it so that only one of the tests/examples that previously needed a different random seed still needs one.

So for `RR` and `CC`, are you suggesting that we adjust it so that the support is [0, opts.Height] and [0, opts.Height] x [0, opts.Height], respectively (or [0, opts.Height] x [0, 2$`\pi`$] in polar coords if we switch to a circle in the complex plane)?.

d-torrance:
Just pushed another commit fixing a bug I found working on this:

### Before
```m2
i1 : random(QQ, Height => 0)
Floating point exception (core dumped)
```

### After
```m2
i1 : random(QQ, Height => 0)
stdio:1:6:(3): error: expected a positive height
```

M2/Macaulay2/d/interface.dd
```diff
@@ -66,8 +66,7 @@ export rawFareyApproximation(e:Expr):Expr := (
 setupfun("rawFareyApproximation", rawFareyApproximation);
 export rawRandomQQ(e:Expr):Expr := (
      when e
-     is Nothing do toExpr(Ccode(QQ, "rawRandomQQ(", "0)"))
-     is ht:ZZcell do toExpr(Ccode(QQ, "rawRandomQQ(", ht.v, ")"))
+     is ht:ZZcell do toExpr(Ccode(QQorNull, "rawRandomQQ(", ht.v, ")"))
```
mahrud, line 69: I think it might be simpler to just assert that Height is nonzero in the top-level, no?

d-torrance: My concern was that we also want to catch this when making random matrices, which is a different call to the engine that isn't specific to the coefficient ring.  Is height always positive for all the various meanings of height we might want for `random`?  Or could it be zero/negative in other contexts?

d-torrance:
I keep re-thinking what the support of `random QQ` should be lol...

Considering that the height of a rational number is the max of the absolute values of its numerator and denominator, and the proposed behavior would allow larger numerators, then maybe the current behavior is basically correct, or at least has the correct support.  But instead maybe we should have a uniform distribution on that support, rather than the current distribution where 1 is more likely than anything else.  So in other words, instead of the distribution being $\frac{X}{Y}$ where $`X,Y\in\{1,\ldots,h\}`$, we have $\frac{X}{Y}$ where $`X,Y\in\{1,\ldots,h\}`$ and $\gcd(X,Y)=1$.

So `rawRandomQQ` could be something like (pseudocode):

```
while (
    x = random_int(1, h)
    y = random_int(1, h)
    gcd(x, y) != 1) do nothing
return x/y
```

d-torrance:
So where did we land on the support?  [0, 1] or [0, height]?

mahrud:
I think new option `Support` with default value `[-1,1]`?

mahrud:
If it makes sense to you, I think we should merge this and work on the optional arguments in follow ups.

d-torrance:
I'm not sure if the current proposed behavior is what we want, though, since the numerator can exceed the height.

mahrud:
Related: #999 and #2089.

mikestillman:
@d-torrance Thanks very much for fixing the previous behavior of random QQ!

The height of a rational number is defined as the max of the absolute values of the numerator and denominator, I believe.  The behavior now is *much* better than it was, but is there any way we could have it return something within the bound of the given height?  If not, I am fine with this behavior.

Is it possible to allow negative numbers too? (we don't need full ranges, but allowing negatives would be useful.  Several people have had to write random functions to work around this behavior (well, at least one person did!))

Also, the doc node for (random, QQ) is not a subnode, as this existed before that requirement :)

I want to look closer at the specific code changed in the engine, but from top level (I'm glad it works in random matrices too, including in polynomial rings over QQ!) it looks very good.

M2/Macaulay2/m2/galois.m2
```diff
@@ -117,7 +117,9 @@ findGalois(ZZ,ZZ) := RingElement => opts -> (p,n) -> (
      else error (opts.Strategy | " is not a valid argument for' Strategy' option")
      )
 
-GF(ZZ,ZZ) := GaloisField => opts -> (p,n) -> (
+GF(ZZ,ZZ) := GaloisField => (
+    memo := new CacheTable;
```
mikestillman, line 121: Why has this been memo-ized?

M2/Macaulay2/e/interface/random.cpp
```diff
@@ -67,34 +70,119 @@ gmp_ZZ rawRandomInteger(gmp_ZZ maxN)
...
 }
 
-gmp_QQ rawRandomQQ(gmp_ZZ height)
-/* returns random a/b, where 1 <= b <= height, 1 <= a <= height */
-/* if height is the null pointer, use the default height */
+void rawSetFareyApproximation(mpq_ptr result, gmp_RR x, gmp_ZZ height)
+/* sets result = the nearest rational to x w/ denominator <= height */
+{
+  int sgn;
+  mpfr_t fracpart, tmp1, tmp2;
```
mikestillman, line 77: I *think* there is no memory leak here, but a bit hard to tell...

M2/Macaulay2/d/interface.dd
mikestillman: Would this be too much of a hack?  Negative height means height is the absolute value, but includes a random sign?

mikestillman (changes requested):
I like the changed behavior.  I had a few comments and requests in the review.  Once we have discussed those, I am likely good with this PR.

Once again, thanks!

d-torrance:
Here are several possible ways I can think of that this could be rewritten to ensure that "height" actually makes sense.

1.  Sample two random integers $X$ and $Y$ from the uniform distribution on $`\{1,...,h\}`$ (so what we're currently doing), but then check if they're relatively prime.  If so, return $X/Y$.  Otherwise, re-roll.
2.  Sample a random real number from the uniform distribution on $[0, h]$ and then find the Farey approximation (i.e., the current proposal).  If the numerator is greater than $h$, re-roll.
3. Compute the number $n$ of rational numbers with a numerator and denominator bounded by $h$, sample an integer $X$ from the uniform distribution on $`\{1,\ldots,n\}`$, and then pick the $X$th element from our list of rationals somehow.

M2/Macaulay2/m2/galois.m2
d-torrance: I forget why this ended up in this particular PR , but GF currently creates a new object every time:

```m2
i1 : tally apply(20, i -> GF 2)

o1 = Tally{GF 2 => 1}
           GF 2 => 1
           GF 2 => 1
           GF 2 => 1
           GF 2 => 1
           GF 2 => 1
           GF 2 => 1
           GF 2 => 1
           GF 2 => 1
           GF 2 => 1
           GF 2 => 1
           GF 2 => 1
           GF 2 => 1
           GF 2 => 1
           GF 2 => 1
           GF 2 => 1
           GF 2 => 1
           GF 2 => 1
           GF 2 => 1
           GF 2 => 1

o1 : Tally
```

Compare to:

```m2
i1 : tally apply(20, i -> ZZ/2)

           ZZ
o1 = Tally{-- => 20}
            2

o1 : Tally
```

mikestillman: Yes, for rings, that is the expected behavior.  Otherwise, we have to keep pointers to all of these fields, and therefore they will never be garbage collected.  I guess currently, we are doing something similar for ZZ/p though, so those are likely not being garbage collected either.  In reality, we should have GF(p) be identical to ZZ/p, but it currently isn't...

As these elements are in different rings, that behavior is expected.  Here is what is the intended way to use rings (create it once):
```m2
kk = GF 8
tally apply(20, i -> random kk)
```

d-torrance: Ok -- I'll remove that commit.

According to the commit message, this is why I made the change:  https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_random_lp__Type_rp.html

I guess the `tally for i to 100 list random GF 11` should be changed to something like `kk = GF 11; tally for i too 00 list random kk`?

mikestillman: Yes, that would be a good change!

d-torrance:
I've just pushed the latest changes.  In particular:

* Rebased onto `development`
* The `random(QQ)` doc is now a subnode of `random`
* Added comments to the Farey approximation code
* Removed the `GF` memoization, but fixed the `random(Type)` doc so it uses only a single `GF` object

mikestillman:
All looks good!  @d-torrance, please merge this.  I'm still unclear which kind of merge commit is best...! (i.e. cleanest is probably rebase and merge, or squash and merge, but does that sometimes cause issues...?)

M2/Macaulay2/m2/galois.m2
mahrud: This was requested before in #1993.

PR #3489, opened 2024-09-20
Bump msolve to v0.7.2

d-torrance:
A new msolve version was released a couple days ago.  We update both the autotools and cmake builds to download and build it if needed.


mahrud:
I don't think we build msolve on GitHub, so this could skip the builds.

PR #3494, opened 2024-09-23
Promote

pzinn:
This PR addresses some of the issues raised in #3184.
before:
```
i1 : R=QQ[x]

o1 = R

o1 : PolynomialRing

i2 : S=R[y]

o2 = S

o2 : PolynomialRing

i3 : 1/x+1/y
stdio:3:3:(3): error: expected pair to have a method for '+'

i4 : matrix{{1/x}}*matrix{{1/y}}
stdio:4:13:(3): error: maps over incompatible rings

i5 : promote(1/x_R,frac S)
stdio:5:7:(3): error: no method found for applying promote to:
                   1
     argument 1 :  - (of class frac R)
                   x
     argument 2 :  frac S

i6 : lift(1/x_S,frac R)
stdio:6:4:(3): error: no method found for applying lift to:
                   1
     argument 1 :  - (of class frac S)
                   x
     argument 2 :  frac R
```
and
```
i1 : R=QQ[x]

o1 = R

o1 : PolynomialRing

i2 : T=R**QQ[z]

o2 = T

o2 : PolynomialRing

i3 : x_R*z
stdio:3:3:(3): error: expected pair to have a method for '*'
```
after:
```
i1 : R=QQ[x]

o1 = R

o1 : PolynomialRing

i2 : S=R[y]

o2 = S

o2 : PolynomialRing

i3 : 1/x+1/y

     y + x
o3 = -----
      x*y

o3 : frac S

i4 : matrix{{1/x}}*matrix{{1/y}}

o4 = | 1/xy |

                    1             1
o4 : Matrix (frac S)  <-- (frac S)

i5 : promote(1/x_R,frac S)

     1
o5 = -
     x

o5 : frac S

i6 : lift(1/x_S,frac R)

     1
o6 = -
     x

o6 : frac R
```
and
```
i1 : R=QQ[x]

o1 = R

o1 : PolynomialRing

i2 : T=R**QQ[z]

o2 = T

o2 : PolynomialRing

i3 : x_R*z

o3 = x*z

o3 : T
```

In particular it introduces a new method `setupPromote` which makes it easy to (automatically) promote from A -> B given a map f: A -> B.


M2/Macaulay2/m2/exports.m2
```diff
@@ -871,7 +871,7 @@ export {
 	"left",
 	"length",
 	"lift",
-	"liftable",
+	"isLiftable",
```
mahrud, line 874: This list is sorted, so this line should be next to other "is" symbols.

M2/Macaulay2/m2/newring.m2
```diff
@@ -11,6 +11,21 @@ needs "matrix2.m2" -- for lift
 -- new polynomial ring or quotient ring from old --
 ----------------------------------
 
+-- automate promotion
+promoteFromMap = method()
```
mahrud, line 15: Maybe call this `setupPromotionMethods` or something like that?

M2/Macaulay2/m2/quotring.m2
```diff
@@ -281,7 +281,7 @@ char QuotientRing := (stashValue symbol char) ((S) -> (
      g := generators gb relns;
      if g == 0 then return char ring g;
      m := g_(0,0);
-     if liftable(m,ZZ) then lift(m,ZZ) else 0))
+     lift(m,ZZ,Verify=>false) ?? 0))
```
mahrud, line 284: Not sure if I understand the motivation in this change. I like `??` and `??=` in certain situations (e.g. caching) but I'm concerned that in places like this it might make debugging more difficult. For instance, I imagine we _would_ like to see an error if `isLiftable` returned true but `lift` failed here, because that would be a bug and an error would help us locate it, but `??` hides the error here and silently returns zero.

M2/Macaulay2/m2/rationals.m2
```diff
@@ -48,9 +48,9 @@ QQ.Engine = true
 assert (hash ZZ < hash QQ)
 
 lift(QQ,ZZ) := opts -> (r,o) -> if denominator r === 1 then numerator r else if opts.Verify then error "rational number is not an integer"
-liftable(QQ,ZZ) := (r,o) -> denominator r === 1
+isLiftable(QQ,ZZ) := (r,o) -> denominator r === 1
 lift(QQ,QQ) := opts -> (r,QQ) -> r
-liftable(QQ,QQ) := (QQ,QQ) -> true
+isLiftable(QQ,QQ) := (QQ,QQ) -> true
```
mahrud, line 53: Why does this line even exist? We don't have `isLiftable(ZZ,ZZ)`, etc.. Maybe there should be `isLiftable(Ring, Ring) := (R, S) -> R === S` that is used as a last resort if R and S don't have an `isLiftable` method installed?

M2/Macaulay2/m2/quotring.m2
pzinn: in general I would agree with you but here's a pretty clear-cut case where we're basically calling `lift` twice for no reason.

pzinn:
there is `isLiftable` in `Varieties.m2`. should it be merged with the core `isLiftable`?

M2/Macaulay2/m2/rationals.m2
pzinn: you're right, this line is useless I believe.
right now for most (all?) rings we have `lift(R,R):=o->(x,R)->x` so `isLiftable(R,R)` will give `true` without any need to override the default behaviour.

M2/Macaulay2/m2/newring.m2
```diff
@@ -77,7 +92,11 @@ tensor(QuotientRing,   QuotientRing) := monoidTensorDefaults >> optns -> (R, S)
      fg := substitute(f,(vars AB)_{0 .. m-1}) | substitute(g,(vars AB)_{m .. m+n-1});
      -- forceGB fg;  -- if the monomial order chosen doesn't restrict, then this
                      -- is an error!! MES
-     AB/image fg)
+     RS := AB/image fg;
+     promoteFromMap map(RS,R,(vars AB)_{0 .. m-1});
+     if S =!= R then promoteFromMap map(RS,S,(vars AB)_{m .. m+n-1});
```
mahrud, line 97: Is it natural that when `R == S` we favor the first ring map over the second?

M2/Macaulay2/m2/newring.m2
mahrud: And perhaps `commonEngineRingInitializations` should call it, similar to the `lift` methods?

M2/Macaulay2/m2/newring.m2
pzinn: yeah this is annoying. no there isn't. in fact there are 3 choices here:
- favour the first
- favour the second
- decide that it's ambiguous and not call `promoteFromMap` at all.

I'm not sure what's best.

M2/Macaulay2/m2/ringmap.m2
```diff
@@ -85,7 +83,7 @@ map(Ring, Ring, Matrix)  := RingMap => opts -> (R, S, m) -> (
 		    " into a degree of length ", toString degreeLength R);
 	       opts.DegreeMap
 	       )
-	  else if workable (() -> promote({},S,R)) then (d -> first promote({d},S,R))
+	  else if (pr:=lookup(promote,List,S,R)) =!= null then (d -> first pr({d},S,R))
```
mahrud, line 86: Huh, I didn't know `promote(List,S,R)` is for promoting the degree till now ...

mahrud:
> there is `isLiftable` in `Varieties.m2`. should it be merged with the core `isLiftable`?

Sure, you just need to remove `isLiftable = method()` right? That should be fine.

M2/Macaulay2/m2/ringmap.m2
pzinn: For the record I fixed this line because it could create infinite loops. It pretty much never occurred before because there were few `promote`, but now with the added flexibility of `promoteFromMap` (or whatever we're going to end up calling), it can easily happen.

M2/Macaulay2/m2/quotring.m2
mahrud: I'm actually a little confused about what's happening in this code, but sure.

M2/Macaulay2/m2/newring.m2
mahrud: I wonder if there's a general lesson to be learned from the way NormalToricVarieties handles natural maps:
```m2
i1 : needsPackage "NormalToricVarieties";

i2 : X = toricProjectiveSpace 1;

i3 : Y = toricProjectiveSpace 2;

i4 : Z = X ** Y;

i5 : Z_[0]

o5 = | 1 |
     | 0 |
     | 0 |

o5 : ToricMap Z <--- X

i6 : Z_[1]

o6 = | 0 0 |
     | 1 0 |
     | 0 1 |

o6 : ToricMap Z <--- Y

i7 : Z^[1]

o7 = | 0 1 0 |
     | 0 0 1 |

o7 : ToricMap Y <--- Z

i8 : Z^[0]

o8 = | 1 0 0 |

o8 : ToricMap X <--- Z
```
I wonder if we could have a notation for specific ring maps.

mahrud:
A similar approach to your `promoteFromMap` that I thought about a while ago but never tried implementing is based on allowing users to install a _natural ring map_ between any two given rings and having internal methods simply call `map(R, S)` to access it. e.g. I could define two completely different rings where I have a mathematically natural map between them, but defining a ring map and passing it to every function is really cumbersome, but if I could literally type:
```m2
map(R, S) := map(R, S, { ... })
```
and have it cache that ring map (there is a syntax for allowing this), then `promote` and `lift` could also use it. Or perhaps we could allow something like:
```m2
lift(R, S) := ...
promote(S, R) := ...
```
to define these methods in one place and have it be used throughout.

The main motivation for this was having a better was to cache a ring map that specifically has a _degree map_ or _degree lift map_ that is then used here:
https://github.com/Macaulay2/M2/blob/1fd238210e9a37ddaec3b700ccd22f06ff5f28f5/M2/Macaulay2/m2/basis.m2#L200-L203
In many situations `map(R, S)` is defined, but the degree map that M2 uses by default is incorrect, which causes all sorts of errors here.

pzinn:
I'm not sure if this is what you want but `map(R,S)` will work after calling `promoteFromMap`:
```
i1 : R=QQ[x]

o1 = R

o1 : PolynomialRing

i2 : S=QQ[y]

o2 = S

o2 : PolynomialRing

i3 : promoteFromMap map(R,S,{x^2})

i4 : map(R,S)

                  2
o4 = map (R, S, {x })

o4 : RingMap R <-- S
```

M2/Macaulay2/m2/ringmap.m2
mahrud: I think I would be in favor of deprecating `promote(List, Ring, Ring)` and instead given a ring map `f` make use of `f.degreeMap` and `f.degreeLift`.

mahrud: Or maybe just `degreeGroup f` defined as `degreeGroup RingMap := f -> f.degreeMap`, which is similar to how NormalToricVarieties supports `picardGroup ToricMap` to give the map of Picard groups functorially.

pzinn:
> > there is `isLiftable` in `Varieties.m2`. should it be merged with the core `isLiftable`?
> 
> Sure, you just need to remove `isLiftable = method()` right? That should be fine.

That seems to cause problems. I haven't looked in detail but the Core `isLiftable` is a `method(TypicalValue => Boolean, Dispatch => {Thing, Type, Type})` whereas the one in Varieties isn't, so the `isLiftable(Matrix, Matrix)` there probably doesn't work.

mahrud:
Okay, then either change the method type or don't rename it?

pzinn:
not sure what you mean by either option. Do you want to change `isLiftable(Matrix, Matrix)` to take different arguments?

mahrud:
No, change the isLiftable in Core. For instance you could use hooks to check `lift(r, R) =!= null` by default:
```m2
isLiftable(Number,      Ring) :=
isLiftable(RingElement, Ring) := (r, R) -> ring r === R or tryHooks(
    LiftabilityHooks, (r, R), (r, R) -> lift(r, R) =!= null))
```
Then you only need to install the special cases where there's legitimately a way to check liftability without actually lifting like this:
```m2
addHook(LiftabilityHooks, (r, R) -> if ring r === QQ and R === ZZ then denominator r === 1)
```
Actually, I would love to use hooks like this for `lift` and `promote` and stop this mess:
```m2
i49 : #methods lift

o49 = 103

i50 : #methods promote

o50 = 124
```

pzinn:
I have to disagree with you here. I don't think hooks should be used as a replacement for the traditional method paradigm based on inheritance. And changing the definition of the Core `isLiftable` to not be `Dispatch => {Thing, Type, Type}` would be a major change which would need to be discussed with all involved developers. In any case, definitely outside the scope of this PR.

mahrud:
Okay, then don't rename isLiftable.

pzinn:
OK I can revert that part of the PR. Though at some point we'll want to do the rename, and then the problem will arise again.

mahrud:
Well, I provided a way for that to happen, but you disagreed, and I don't see another option.

> I don't think hooks should be used as a replacement for the traditional method paradigm based on inheritance.

The problem with `lift`, `promote` and `liftable` is specifically that they do _not_ use inheritance because e.g. `QQ[x]` does _not_ a subtype of `QQ`, so why are they implemented as methods in the first place? The whole `Dispatch` system seems to be kind of a hack, given that rings and modules don't have inheritance whatsoever.

Besides, we use hooks in places where a function depends on the specifics of the ring of its inputs several places; look up `ReduceHooks` and `ContainmentHooks` for instance. This seems like the natural application to me.

pzinn:
I don't disagree that `Dispatch` is a bit of a hack. If M2internals weren't at 3am for me, I'd be happy to suggest a general discussion about it at some point.

pzinn:
reverting `liftable` -> `isLiftable` 7689295449e5036a534d0726da5408f378bb862c e9e77da2693fedcadde05cccfcc957e0f444e745  a7e9349fe533bc49bda54387554786a5730c5e70

M2/Macaulay2/m2/newring.m2
pzinn: how about just `setupPromote`? so it's not too long

mahrud: Up to you. My plan is to eventually allow the notation `map(R, S) = f`.

mahrud:
Are you aiming for this PR to make it to the upcoming release, or the next one?

pzinn:
I think it should be ready for this release.

pzinn:
seems I'm getting both #3239 and #1579 ?

M2/Macaulay2/packages/Macaulay2Doc/functions/setupLift-doc.m2
```diff
@@ -0,0 +1,43 @@
+--- status: DRAFT
+--- author(s): PZJ
+--- notes: 
+
+undocumented methods setupLift
```
mahrud, line 5: Is there a reason you undocument these rather than just list the method keys as secondary keys?

M2/Macaulay2/packages/Macaulay2Doc/functions/setupLift-doc.m2
```diff
@@ -0,0 +1,43 @@
...
+
+doc ///
+ Key
+  setupLift
+ Headline
+  automatic lift from one ring to another
```
mahrud, line 11: In what sense is this "automatic"? Seems like the purpose of this function is to _manually_ setup lifting method.

M2/Macaulay2/packages/Macaulay2Doc/functions/setupLift-doc.m2
mahrud: If it's because you get weird documentation errors, look at the `Synopsis` keyword used in the documentation of `preimage`, for instance.

M2/Macaulay2/packages/Macaulay2Doc/functions/setupLift-doc.m2
pzinn: oops. copy-pasted too quickly the `setupPromote` doc.

M2/Macaulay2/m2/enginering.m2
```diff
@@ -225,7 +225,40 @@ commonEngineRingInitializations = (F) -> (
...
+    else d -> d|toList(n:0)
+    )
+
+-- automate promote
+setupPromote = method()
+setupPromote (Function,Ring,Ring,Function) := (f,R,S,degmap) -> (
+    promote(R,S) := (a,S) -> f a;
+    promote(List,R,S) := (m,R,S) -> apply(m,degmap);
+    -- promote(Matrix,R,S) :=
+    -- promote(MutableMatrix,R,S) := -- doesn't work, cf https://github.com/Macaulay2/M2/issues/2192
```
mahrud, line 243: I'm confused about these commented lines. Don't you have these exact methods implemented two lines below?

pzinn: yes, but not as `S ** M`

M2/Macaulay2/packages/Macaulay2Doc/doc5.m2
```diff
@@ -360,7 +360,7 @@ document { Key => {(eagonNorthcott,Matrix),eagonNorthcott},
 
 document { Key => {(selectVariables,List,PolynomialRing),selectVariables},
      Headline => "make a subring of a polynomial ring generated by selected variables",
-     Usage => "(S,f) = selectVariables(v,R)",
+     Usage => "(S,F) = selectVariables(v,R)",
```
mahrud, line 363: Was this change included by accident? `f` makes a lot more sense for a ring map than `F`.

pzinn: one of the variables is called `f`

M2/Macaulay2/packages/Macaulay2Doc/functions/setupLift-doc.m2
pzinn: mostly out of laziness. Also, some of the methods are really meant for internal use only, so I need to decide whether to document those or not.

M2/Macaulay2/packages/Macaulay2Doc/doc5.m2
mahrud: Ah .. still, maybe changing ring variables to x_1..x_8 would be better, but it's fine.

M2/Macaulay2/packages/Macaulay2Doc/functions/setupLift-doc.m2
mahrud: My complaint applies there, too. How is that automatic?

M2/Macaulay2/packages/Macaulay2Doc/functions/setupLift-doc.m2
mahrud: Yeah, I would vote to keep both of these methods unexported for now (of course you can do `importFrom_Core "setupPromote"` in your own package to use it).

M2/Macaulay2/packages/Macaulay2Doc/functions/setupLift-doc.m2
pzinn: it is automatic in the sense that you don't need to invoke `promote` explicitly, see the examples provided for `setupPromote`

M2/Macaulay2/packages/Macaulay2Doc/functions/setupLift-doc.m2
pzinn: When I said "some of the methods are really meant for internal use only", I did not mean that `setupPromote` and `setupLift` were for internal use only. I meant some of their methods (as listed by say `methods setupPromote`) are.

M2/Macaulay2/m2/enginering.m2
mahrud: Why would it be? That wouldn't even work because `Ring ** Matrix` calls `Ring ** Module` which then calls `promote(Matrix, Ring)`, so it would be a loop.

M2/Macaulay2/editors/emacs
mahrud, line 1: This hash changed again.

M2/Macaulay2/packages/Macaulay2Doc/functions/setupLift-doc.m2
mahrud: Well you've listed two things under usage, but so they should be listed as keys.

M2/Macaulay2/packages/Macaulay2Doc/functions/setupPromote-doc.m2
```diff
@@ -0,0 +1,29 @@
...
+--- author(s): PZJ
+--- notes: 
+
+undocumented methods setupPromote
+
+doc ///
+ Key
+  setupPromote
+ Headline
+  automatic promotion from one ring to another
```
mahrud, line 11: Your force push made my comment disappear, but still, I don't think this headline describes what this function does. If you don't want to remove "automatic", at least add "set up" to the beginning? Also:
- list at least one method key
- maybe link to this node from `promote`? (and similarly from `lift` to `setupLift`)

M2/Macaulay2/m2/enginering.m2
```diff
@@ -466,6 +503,8 @@ swap = (x,y) -> (y,x)
 promoterightexact   = swap @@ promoteleftexact   @@ swap
 promoterightinexact = swap @@ promoteleftinexact @@ swap
 
+isPromotable = (R,S) -> lookup(promote,R,S) =!= null
```
mahrud, line 506: Should be defined next to `promote`?

M2/Macaulay2/m2/enginering.m2
```diff
@@ -225,7 +225,40 @@ commonEngineRingInitializations = (F) -> (
...
+setupPromote (Function,Ring,Ring) := (f,R,S) -> setupPromote(f,R,S,defaultDegMap(R,S));
+
+-- automate (to some extent) lift
+setupLift = method()
+setupLift (Function,Ring,Ring,Function) := (f,R,S,degmap) -> (
+    lift(R,S) := opts -> (a,S) -> if opts.Verify then f a else try f a;
+    lift(List,R,S) := opts -> (m,R,S) -> apply(m,degmap);
+    lift(Module,R,S) := opts -> (M,R,S) -> S ** M;
+    lift(Matrix,R,S) := opts -> (m,R,S) -> map(lift(target m,S),lift(source m,S),applyTable(entries m,x->lift(x,S)));
+    lift(MutableMatrix,R,S) := opts -> (m,R,S) -> mutableMatrix applyTable(entries m,x->lift(x,S));
```
mahrud, line 257: Why do these call `lift(x,S)` and not `lift(x, R, S)`?

Also, do we have _any_ `lift` methods of matching e.g. `lift(Module, R, S)` that _isn't_ defined here automatically? Because if not, I think all of these should just be e.g. `lift(Module, Ring, Ring)`, and we can rename `lift(x,S)` to something else that is defined per ring.

M2/Macaulay2/editors/emacs
pzinn: this is a pain, we should set up the emacs stuff differently so it doesn't occur...

M2/Macaulay2/packages/Macaulay2Doc/functions/setupPromote-doc.m2
pzinn: will rewrite doc

M2/Macaulay2/m2/enginering.m2
pzinn: good question. `lift(Module,...)` only works for free modules anyway, it seems the definition I give could work universally. that being said it's not how it's done for the usual internal `lift`, and I don't dare to change too much.

M2/Macaulay2/m2/enginering.m2
pzinn: I think the comment (which I wrote a while ago) only applies to the `MutableMatrix`. I'll just remove these lines.

M2/Macaulay2/m2/enginering.m2
mahrud: Okay. At the next internals meeting I'll propose deprecating lift/promote for everything other than ring elements. The current system is obfuscated for very little benefit.

M2/Macaulay2/packages/Macaulay2Doc/functions/lift-doc.m2
```diff
@@ -122,5 +122,5 @@ document {
      .0001^QQ
      .0001_QQ
      ///,
-     SeeAlso => {baseRings,liftable,promote}
+     SeeAlso => {baseRings,liftable,promote,setupLift}
```
mahrud, line 125: I would even put `setupLift` under `Subnodes` of `lift`, just to give it more structure.

mahrud: Same with `liftable`.

mahrud (approved):
Looks good to me.

Tasks for future:
- rename liftable to isLiftable
- allow the notation `map(R, S) = map(R, S, {x,y,z}, DegreeMap => ..., DegreeLift => ...)`
- perhaps limit ring-dependent lift/promote methods to only ring elements

pzinn:
> Looks good to me.
> 
> Tasks for future:
> 
> * rename liftable to isLiftable
> * allow the notation `map(R, S) = map(R, S, {x,y,z}, DegreeMap => ..., DegreeLift => ...)`
> * perhaps limit ring-dependent lift/promote methods to only ring elements

Yes this PR is definitely the end of the story, and some of the content may even be short-lived (e.g., if you get rid of all `promote` beyond `RingElement` and have this new syntax `map(R,S)=` then `setupPromote` can probably go) but I think it's useful to have this as a first step.

mahrud:
Yes, I think it's great to have this for now.

mahrud:
Could you push https://github.com/Macaulay2/M2/pull/3519/commits/bbc9f099484946762ef4f667a860be286665a891 to your branch? Or I can do it if that's okay.

pzinn:
please do

d-torrance:
@pzinn: Do you have a suggestion for the changelog entry for these promotion updates?

mahrud:
This PR broke this:
```m2
R = CC_53[x,y]
sub(0 * vars R, CC_53)
```

mahrud:
@pzinn @d-torrance could you make sure this is either fixed or this PR is reverted before the release until the issue can be fixed?

d-torrance:
What's the bug?  I'm getting the following on the release branch (which is on top of development):

```m2
i1 : R = CC_53[x,y];

i2 : sub(0 * vars R, CC_53)

o2 = | 0 0 |

                1         2
o2 : Matrix CC    <-- CC
              53        53
```

Isn't that the expected result?

pzinn:
can't reproduce the bug either.
as to the blurb:
- allow promotion and lifting between two arbitrary  rings using `setupPromote` and `setupLift`.

mahrud:
I would maybe prefer not advertising `setupPromote` and `setupLift` yet, since we may switch to different syntax soon?

mahrud:
This is what I'm getting:
```m2
i1 : R = CC_53[x,y]; sub(0 * vars R, CC_53)
stdio:1:19:(3): error: no method found for applying promote to:
     argument 1 :  | x y | (of class Matrix)
     argument 2 :  R
     argument 3 :  CC
                     53
```

mahrud:
Ah never mind, I think it's a bad mixed reaction of two changes.

pzinn:
fine, something more generic then:
relax the restriction on promotion/lifting so one can promote/lift between any two rings, and apply this to fraction fields and tensor products.

PR #3498, opened 2024-09-25
Improve some interpreter-level error messages

d-torrance:
Mostly changing "expected a number" to specifically listing the types of number that are expected.


mahrud (dismissed):
I'm a little worried that eventually (with the addition of complex intervals and real and complex balls) this error is going to get really unwieldy. Also, most of these functions still work for integers and rationals. Maybe we just want to say `expected a real or imaginary number`?

d-torrance:
> Also, most of these functions still work for integers and rationals.

The compiled functions themselves don't.  They only work after we install methods that call `numeric` on the inputs.  This is the current behavior, which I think is confusing:

```m2
$ M2 --no-core
Macaulay2, version 1.24.05-645-g9d1e0c2dbb (development)

i1 : cos 0
stdio:1:4:(3): error: expected a number
```
I agree that adding a `WrongArgRRorRRiorRRballorCCorCCiorCCball` function would be unwieldy if/when we add these extra types, but we haven't yet.  :)

M2/Macaulay2/d/actors3.d
```diff
@@ -730,23 +730,23 @@ sin(e:Expr):Expr := (
      is x:CCcell do toExpr(sin(x.v))				    -- # typical value: sin, CC, CC
      is x:RRcell do toExpr(sin(x.v))				    -- # typical value: sin, RR, RR
      is x:RRicell do toExpr(sin(x.v))				    -- # typical value: sin, RRi, RRi
-     else buildErrorPacket("expected a number")
+     else WrongArgRRorRRiorCC()
```
mahrud, line 733: I'm fine with everything in this PR except the uses of `WrongArgRRorRRiorCC`. I don't think that function should exist. Either the error should be less informative (e.g. just say `number` or `real or complex number` or call `WrongArgRRorCC`) or there should be a more flexible alternative, like `WrongArg(RR,RRi,CC)` which can be easily adjusted.

d-torrance: Fair enough -- `WrongArgRRorCC` works for me!

This is all pretty minor anyways since the vast majority of the time, users will be using the methods that call these functions and never see these error messages anyway.

mahrud: > This is all pretty minor anyways since the vast majority of the time, users will be using the methods that call these functions and never see these error messages anyway.

Exactly, which is why I prefer minimal and maintainable interpreter code over verbose.

M2/Macaulay2/d/actors3.d
```diff
@@ -1062,7 +1062,7 @@ atan(e:Expr):Expr := (
      is x:RRicell do toExpr(atan(x.v))				    -- # typical value: atan, RRi, RRi
      is a:Sequence do if length(a) == 2 then buildErrorPacket("atan(x,y) has been replaced by atan2(y,x)")
```
mahrud, line 1063: Pretty sure we can delete this line by now ... but no need to do it here and run the actions all over again.

M2/Macaulay2/d/actors3.d
```diff
@@ -1213,13 +1213,13 @@ log(e:Expr):Expr := (
                     else
                         buildErrorPacket("Not defined")
 		    )
-	       else WrongArg(1,"a number"))
-	  else WrongArg(2,"a number"))
+	       else WrongArgRRorRRi(2))
+	  else WrongArgRRorRRi(1))
```
mahrud, line 1217: Did the 1 and 2 flip here?

mahrud: Ah it is corrected, nvm.

mahrud (approved):

PR #3501, opened 2024-09-29
Minor change to insert, switch, replace, RingMap acting on lists

pzinn:
This PR fixes two inconsistencies in allowed types:
- one can now write `f \ l` for `f` a `RingMap` and `l` any `VisibleList`, not just `List` (same as `f \ l` when `f` is a `Function`)
- `insert`, `switch` and `replace`, which are all based on `take`, accept `BasicList`, not just `VisibleList`, like `take` itself.


pzinn:
can this be reviewed/merged? I have other PRs coming that depend on this.

M2/Macaulay2/editors/emacs
d-torrance, line 1: Did you mean to update the emacs submodule?

pzinn: argh! no

d-torrance:
I've always interpreted `BasicList` objects as things we don't really think of as lists, but just happen to be implemented as lists. In particular, I'm not sure how many "listy" methods outside of the ones defined in the interpreter we want installed for them.

pzinn:
well, `replace` is pretty useful no matter what type of `BasicList` you're dealing with.

d-torrance:
> well, `replace` is pretty useful no matter what type of `BasicList` you're dealing with.

Sounds good to me!

d-torrance (approved):

Issue #2811, opened 2023-04-14
multiplication became broken in noncommutative rings recently


mikestillman:
I am surprised this became broken (or rather, I'm surprised that there are no tests to make sure it is working correctly).
```
needsPackage "AssociativeAlgebras"
A = QQ <|x,y|>
-- these are fine
x*y
y*x
M = matrix {{x}}
N = matrix {{y}}
-- these are not
y * M
M * y
```

The problem appears to have been added in commit 36fdb90298, and no test revealed that it broke non-commutative multiplication.

I think we need to change the code in `m2/matrix.m2`: 
```
Matrix * RingElement := (m,r) -> r*m
```

back to what it was:

```
Matrix * RingElement := (m,r) -> (
     r = promote(r,ring m);
     map(target m, source m, reduce(target m, raw m * raw r)))
```

pzinn:
oops! my bad

moorewf:
I think the fix is slightly different -- I think one should take the code from the multiplication in the other order (RingElement * Matrix) and reverse the order of multiplication.  this is because the RingElement * Matrix code contains the fix that @pzinn was fixing with 36fdb90.

mahrud:
Probably worth checking everything in `methods symbol*`.

Incidentally, I noticed that `RingElement * MonomialIdeal` is terribly wrong!! Try `(x + y) * monomialIdeal x` and it gives `x^2`!

PR #3503, opened 2024-10-01
Add scalar division methods

d-torrance:
I was using M2 to double-check some computations while grading calculus homework and was surprised to get the following error:

```m2
i1 : vector {1, 2, 3} / 4
stdio:1:17:(3): error: no method for binary operator / applied to objects:
                              3
            | 1 | (of class ZZ )
            | 2 |
            | 3 |
      /     4 (of class ZZ)
```

We add scalar division methods for vectors and matrices by numbers and ring elements.  So afterward:

```m2
i1 : vector {1, 2, 3} / 4

o1 = | 1/4 |
     | 1/2 |
     | 3/4 |

       3
o1 : QQ
```

### DIFF
--- a/M2/Macaulay2/m2/matrix.m2
+++ b/M2/Macaulay2/m2/matrix.m2
@@ -85,6 +85,8 @@ Matrix * Number :=
 Matrix * RingElement := (m,r) -> (
     if ring r =!= ring m then try r = promote(r,ring m) else m = promote(m,ring r);
      map(target m, source m, reduce(target m, raw m * raw r)))
+Matrix / Number      :=
+Matrix / RingElement := (m,r) -> m * (1/r)
 
 toSameRing = (m,n) -> (
      if ring m =!= ring n then (
--- a/M2/Macaulay2/packages/Macaulay2Doc/doc.m2
+++ b/M2/Macaulay2/packages/Macaulay2Doc/doc.m2
@@ -403,6 +403,33 @@ document {
      SeeAlso => { "//"}
      }
 
+doc ///
+  Key
+    (symbol /, Matrix, Number)
+    (symbol /, Matrix, RingElement)
+    (symbol /, Vector, Number)
+    (symbol /, Vector, RingElement)
+  Headline
+    scalar division
+  Usage
+    v / c
+  Inputs
+    v:{Matrix, Vector}
+    c:{Number, RingElement}
+  Outputs
+    :{Matrix, Vector}
+  Description
+    Text
+      This operation is equivalent to right scalar multiplication by the
+      multiplicative inverse of @CODE "c"@, i.e., @CODE "v * (1/c)"@.
+    Example
+      R = QQ[a,b,c,d]
+      matrix {{d, -b}, {-c, a}} / (a * d - b * c)
+  Caveat
+    The base ring of the output will be a field containing the base ring
+    of @CODE "v"@.
+///
+
 document {
      Key => {symbol %,
 	  (symbol %, CC, CC),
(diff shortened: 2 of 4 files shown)


M2/Macaulay2/m2/matrix.m2
```diff
@@ -85,6 +85,8 @@ Matrix * Number :=
 Matrix * RingElement := (m,r) -> (
     if ring r =!= ring m then try r = promote(r,ring m) else m = promote(m,ring r);
      map(target m, source m, reduce(target m, raw m * raw r)))
+Matrix / Number      :=
+Matrix / RingElement := (m,r) -> 1/r * m
```
mahrud, line 89: I think this is intentionally not defined:
```m2
i3 : R = kk[x,y]

o3 = R

o3 : PolynomialRing

i4 : matrix{{x}} / x

o4 = | 1 |

                    1             1
o4 : Matrix (frac R)  <-- (frac R)
```

mahrud: But this is defined:
```m2
i3 : matrix{{x}} // x

o3 = | 1 |

             1      1
o3 : Matrix R  <-- R
```

mahrud: Even over integers:
```m2
i2 : Matrix / Number      := (m,r) -> 1/r * m

o2 = FunctionClosure[stdio:2:30-2:39]

o2 : FunctionClosure

i3 : matrix{{2}} / 2

o3 = | 1 |

              1       1
o3 : Matrix QQ  <-- QQ
```

d-torrance: I'm proposing that we define them.  :)  I think the result being over the fraction field seems like reasonable behavior.  That's how `ZZ / ZZ` and `RingElement / RingElement` work.

mahrud: I guess my point is that using `1/r * m` instead makes it clear that you _should_ expect something over the fraction field, so maybe at least caveat this in the documentation.

pzinn:
I think it was intentionally undefined. In linear algebra, we learn that we can takes "scalar * vector", but not the other way around.

pzinn:
actually, I'm confused, multiplication already works -- `matrix{{3}}*3` or `vector{3}*3`...

pzinn:
never mind my previous comment. then I guess there's no reason not to have division work as well.

mahrud:
Actually, how this behaves in the noncommutative case is a great point. There are two _different_ methods for scalar multiplication on left and right on purpose:
https://github.com/Macaulay2/M2/blob/1fd238210e9a37ddaec3b700ccd22f06ff5f28f5/M2/Macaulay2/d/engine.dd#L177-L188
The `opposite_mult` argument (whether or not it's actually honored in the engine) distinguishes them:
https://github.com/Macaulay2/M2/blob/1fd238210e9a37ddaec3b700ccd22f06ff5f28f5/M2/Macaulay2/e/matrix.hpp#L117
So at the very least division should be defined as `m * (1/r)` not the other way around.

pzinn:
ah, yes, cf #2811

d-torrance:
Makes sense!  I've updated it to use right multiplication.

M2/Macaulay2/packages/Macaulay2Doc/doc.m2
```diff
@@ -403,6 +403,33 @@ document {
...
+  Description
+    Text
+      This operation is equivalent to right scalar multiplication by the
+      multiplicative inverse of @CODE "c"@, i.e., @CODE "v * (1/c)"@.
+    Example
+      R = QQ[a,b,c,d]
+      matrix {{d, -b}, {-c, a}} / (a * d - b * c)
+  Caveat
+    The base ring of the output will be the fraction field of the base ring
+    of @CODE "v"@.
```
pzinn, line 430: I'm not sure I agree with this "Caveat". For example, `matrix{{1}}/2.` will return a matrix in `RR`...

d-torrance: Hmm, good point.  Maybe something more general like "the base ring of the output will be a field containing the base ring of v"?

d-torrance: I went ahead and made this change

d-torrance:
@pzinn , @mahrud: Any additional thoughts on this PR?  Is this okay to merge?

mahrud (approved):

PR #3504, opened 2024-10-02
SCMAlgebras - a package to work with sequentially Cohen-Macaulay modules or ideals

ErnestoLax:
SCMAlgebras is a package to check whether a module or an ideal is sequentially Cohen-Macaulay, by computing the modules of deficiency and the filter ideals.

### DIFF
--- a/M2/Macaulay2/packages/SCMAlgebras.m2
+++ b/M2/Macaulay2/packages/SCMAlgebras.m2
@@ -342,13 +342,13 @@ assert(isCM canonicalModule(M))
 
 
 --========================
--- minDim test
+-- minimumDimension test
 --========================
 TEST ///
 S = QQ[x_1..x_4,y_1..y_4]
 E = {{1, 2}, {1, 3}, {1, 4}, {2, 3}, {3, 4}}
 J = ideal(for e in E list x_(e#0)*y_(e#1)-x_(e#1)*y_(e#0));
-assert(minDim(J)==4)
+assert(minimumDimension(J)==4)
 ///
 
 
--- a/M2/Macaulay2/packages/SCMAlgebras.m2
+++ b/M2/Macaulay2/packages/SCMAlgebras.m2
@@ -3,9 +3,9 @@
 newPackage(
     "SCMAlgebras",
     Version => "1.0",
-    Date => "October 2, 2024",
+    Date => "October 3, 2024",
     Authors => {{Name => "Ernesto Lax", Email => "erlax@unime.it", HomePage => "https://www.researchgate.net/profile/Ernesto-Lax"}},
-    Headline => "a package to work with sequentially Cohen-Macaulay modules or ideals",
+    Headline => "sequentially Cohen-Macaulay modules or ideals",
     Keywords => {"Commutative Algebra"},
     PackageExports => {"Depth","MinimalPrimes"},
     DebuggingMode => false
@@ -18,7 +18,7 @@ export{
     "filterIdeal",
     "isSCM",
     -- Service
-    "minDim"
+    "minimumDimension"
 };
 
 
@@ -28,7 +28,7 @@ Node
   Key
     SCMAlgebras
   Headline
-    a package to work with sequentially Cohen-Macaulay modules or ideals
+    sequentially Cohen-Macaulay modules or ideals
   Description
     Text
       SCMAlgebras is a package to check whether a module or an ideal is sequentially Cohen-Macaulay,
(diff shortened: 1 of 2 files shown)


M2/Macaulay2/packages/SCMAlgebras.m2
```diff
@@ -0,0 +1,394 @@
+-- -*- coding: utf-8 -*-
+
+newPackage(
+    "SCMAlgebras",
+    Version => "1.0",
+    Date => "October 2, 2024",
+    Authors => {{Name => "Ernesto Lax", Email => "erlax@unime.it", HomePage => "https://www.researchgate.net/profile/Ernesto-Lax"}},
+    Headline => "a package to work with sequentially Cohen-Macaulay modules or ideals",
```
d-torrance, line 8: The tests are currently failing because this headline includes the redundant "a package to ...".  From the [Package writing style guide](https://github.com/Macaulay2/M2/wiki/Package-Writing-Style-Guide):

> The Headline option to "newPackage" should be a brief phrase stating the topic of the package. For example, the headline for the package "Tropical" is the phrase "tropical geometry". It would be redundant to say something such as "a package for tropical geometry", "computations in tropical geometry", or "a package for computations in tropical geometry".

M2/Macaulay2/packages/SCMAlgebras.m2
```diff
@@ -0,0 +1,394 @@
...
+-------------------------------------------------------------------------
+deficiencyModule(Module,ZZ) := (M,i) -> (
+    S:=ring M;
+    n:=dim S;
+    t:=depth M;
+    d:=dim M;
+    if i<t or i>d then (
+        Di:=Ext^(n+1)(M,S^{-n})
+    ) else Di=Ext^(n-i)(M,S^{-n});
+    return Di
```
d-torrance, line 89: `return` on the last line of a function is unnecessary.  This line be simplied to just `Di`.

The same is true for many of the other functions in this package.

M2/Macaulay2/packages/SCMAlgebras.m2
```diff
@@ -0,0 +1,394 @@
...
+      S = QQ[x_1..x_10,y_1..y_10];
+      E = {{1,2},{1,3},{1,4},{1,5},{1,6},{1,7},{1,8},{1,9},{1,10},{6,7},{8,9},{8,10},{9,10}};
+      J=ideal(for e in E list x_(e#0)*y_(e#1)-x_(e#1)*y_(e#0));
+      minDim J
+  SeeAlso
+    filterIdeal
+    isSCM
+///);
+-------------------------------------------------------------------------
+minDim = method(TypicalValue=>ZZ);
```
d-torrance, line 182: From the style guide:

> Abbreviations in exported names should be avoided. Correct: "formalDepth". Incorrect: "fD".

This should probably be `minimumDimension`.

M2/Macaulay2/packages/SCMAlgebras.m2
```diff
@@ -0,0 +1,394 @@
...
+    if i>-2 and i<d0 then (
+        Ii:=I;
+    ) else if i>=d0 and i<d then ( 
+        Q:=decompose I;
+        L:={};
+        for k from 0 to #Q-1 do (
+            Qk:=Q#k;
+            Pk:=radical Qk;
+            dk:=dim Pk;
+            L=L|{{dk,Qk}};
```
d-torrance, line 241: From the style guide:

> Do not convert an order n algorithm into an order n^2 algorithm by repeated concatenation of lists or of matrices. Similarly, do not repeatedly append or prepend something to a list, and do not repeatedly remove single items from a list. Hash tables are your friend here. Notice also that the for ... list ... command offers a convenient way to make a list.

I think something like `apply(decompose I, J -> {P := radical J, dim P})` would work here.

The same goes for the construction of `Pi` a few lines below.

M2/Macaulay2/packages/SCMAlgebras.m2
```diff
@@ -0,0 +1,394 @@
...
+///);
+-------------------------------------------------------------------------
+isSCM = method(TypicalValue=>Boolean);
+-------------------------------------------------------------------------
+isSCM(Module) := M -> (
+    d:=dim M;
+    for i from 0 to d-1 do ( 
+        Oi:=deficiencyModule(M,i);
+        if  Oi != 0 then (
+            if (isCM(Oi) == false) or (dim(Oi)!=i) then (
```
d-torrance, line 302: From the style guide:

> Don't test a Boolean value for equality with true or false, because x===true is equivalent to x and x===false is equivalent to not x.

Here, you can use `if not isCM Oi or ...`.

ErnestoLax:
Thanks for your comments! I updated the code according to your suggestions

d-torrance (approved):

PR #3509, opened 2024-10-06
Assorted fixes

mahrud:
- **combined documentation of catch and throw**
- **fixed a typical value typo**
- **deleted duplicate code**
- **changed MonomialIdeal ^ ZZ to use BinaryPowerMethod**
- **removed dead code**
- **added lift and promote for IndexedVariable**

### DIFF
--- a/M2/Macaulay2/packages/Macaulay2Doc/ov_language.m2
+++ b/M2/Macaulay2/packages/Macaulay2Doc/ov_language.m2
@@ -606,18 +606,17 @@ document {
 		},
 	SeeAlso => {"if", "then"}
      }
-document { Key => "catch",
-     Headline => "catch a thrown exception", SeeAlso => {"throw"},
-     Usage => "catch c",
-     Outputs => {{"the value obtained by evaluating the code ", TT "c", ", or, if a ", TO "throw", " was executed during the evaluation of ", TT "c", ",
-	       the argument given to ", TO "throw", "."}},
-     EXAMPLE lines ///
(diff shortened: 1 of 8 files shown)


M2/Macaulay2/packages/Macaulay2Doc/ov_language.m2
```diff
@@ -606,18 +606,17 @@ document {
...
-     EXAMPLE lines ///
-          catch scan(0..10, i -> if i == 5 then throw 18 else print i)
-     ///}
-document { Key => "throw",
-     Headline => "throw an exception", SeeAlso => {"catch"},
-     Usage => "throw x", 
-     Consequences => {{"the flow of control is passed to the surrounding ", TO "catch", ", and ", TT "x", " is returned as its value"}},
+document {
+    Key => {"throw", "catch"},
+    Headline => "throw and catch exceptions",
```
d-torrance, line 611: Is "exception" the right term to use  here?  This makes it seem like these are used for dealing with errors.

mahrud: An exception doesn't have to be an error, you catch exceptions so they don't _become_ errors. Also the previous titles were "throw an exception" and "catch a thrown exception".

d-torrance: Works for me!

d-torrance (approved):

PR #3510, opened 2024-10-07
Bump normaliz to 3.10.4

d-torrance:
[Normaliz 3.10.4](https://github.com/Normaliz/Normaliz/releases/tag/v3.10.4) was released earlier today, so we update the version when building it ourselves.

This is a draft for now -- I plan on removing the second commit (which forces building normaliz on GitHub) if the builds are successful.

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -224,6 +224,7 @@ jobs:
            path: |
              # Autotools
              M2/BUILD/build/config.log
+             M2/BUILD/build/check-configure/tmp/config.log
              M2/BUILD/build/include/*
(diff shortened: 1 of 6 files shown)


d-torrance:
@mahrud: Shouldn't `-DBUILD_LIBRARIES=NORMALIZ` tell the cmake build to build normaliz?  It's not working on the GitHub builds.

mahrud:
It needs to also be in the `LIBRARY_OPTIONS` list in `cmake/check-libraries.cmake`.

mahrud:
Okay I think everything is working now, if you want to rebase without the "REMOVE ME" commit (I cancelled the builds because a couple of them failed due to url timeout.)

PR #3512, opened 2024-10-07
Build with Python support by default

d-torrance:
The Python package is pretty stable, and I think it might be useful to some users for it to be more widely available (there was a recent question about using numpy in Zulip, for example), so I'm proposing that we build with Python support by default.  It would still be possible to turn this off using `--without-python` (autotools) or `-DWITH_PYTHON=OFF` (cmake).


mahrud (approved):

mahrud:
```c++
/usr/local/opt/python@3.12/Frameworks/Python.framework/Versions/3.12/include/python3.12/pyconfig.h:1454:9: warning: 'HAVE_UNISTD_H' macro redefined [-Wmacro-redefined]
 1454 | #define HAVE_UNISTD_H 1
      |         ^
/Users/runner/work/M2/M2/M2/BUILD/build/include/M2/config.h:280:9: note: previous definition is here
  280 | #define HAVE_UNISTD_H
      |         ^
```

Issue #1091, opened 2020-04-11
Delete unused files


mahrud:
This directory contains several simple utilities that either are not used or can be replaced with simpler tools:
- args.c: not used
- echoout.c: used once in Macaulay2/d to reverse a list. Possible alternative:
  - `alias tac='perl -e "print reverse(<>)"'` (in case `tac` doesn't already exist)
  - `(for d in ${MY_LIST}; do echo $$d; done) | tac`
- gc_demo.c: not used, though perhaps useful to keep it somewhere else for pedagogy.
- gc_test.c: used as a test, can be moved to Macaulay2/tests
- linkexec*: seems to be an overly complicated solution to making the programs relocatable
- setup.c: windows legacy
- timestamp.c: compiled, but not used as far as I can tell

mahrud:
Is Macaulay2/d/basictests/testgc.c related to Macaulay2/util/gc_test.c?

DanGrayson:
Yes, they test for the same libgc bug, but in different ways.

mahrud:
Not specifically related to `util`, but on the topic of deleting unused cruft, @d-torrance can we:
1. move `distributions` under `BUILD` (and maybe delete the unused/outdated files)
2. delete `check-configure` and `usr-build` or move contents to `config`?
3. maybe rename `config` to `autotools` to match `cmake` subdirectory and avoid confusion that these are configurations for Macaulay2 itself?
4. delete `files` directory or move contents elsewhere?
5. do we need both `Makefile` and `GNUmakefile.in`?

d-torrance:
> 1. move `distributions` under `BUILD` (and maybe delete the unused/outdated files, subsuming [readme #1211](https://github.com/Macaulay2/M2/issues/1211))
> 2. delete `check-configure` and `usr-build` or move contents to `config`?
> 3. maybe rename `config` to `autotools` to match `cmake` subdirectory and avoid confusion that these are configurations for Macaulay2 itself?
> 4. delete `files` directory or move contents elsewhere?
> 5. do we need both `Makefile` and `GNUmakefile.in`?

These all seem like good ideas.

I think renaming `GNUmakefile.in` to just `Makefile.in` and replacing `Makefile` with a shell script (probably named `autogen.sh` and `boostrap.sh`) that generates the configure script would make a lot of sense and put us in line with most other autotools-based projects.

mahrud:
I like "autogen.sh" better.

mahrud:
[...]

PR #3520, opened 2024-10-15
Various updates to the autotools build

d-torrance:
This addresses some of the issues raised in #1091.  So far, we:

* Replace `Makefile` with a much simpler `autogen.sh` that generates the `configure` script and updates the M2-emacs submodule.  (We drop the various autotools version checks since they'll almost certainly be satisfied.  For example, we were checking for GNU make 3.81, but it was released in 2006!)
* Rename `GNUmakefile` -> `Makefile` for consistency.
* Remove several autotools-generated files from version control.
* Rename `config` -> `m4` to avoid confusion.

I'll probably add a few more things to this PR.


.gitignore
mahrud, line 1: We could also add the submodules here so users don't accidentally add them (we can still update them with `git add -f`), though on the other hand people might not realize that the submodules are out of date ...

M2/autogen.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+echo "-- Generating configure script"
+autoreconf --verbose --force --install
+
+echo "-- Updating M2-emacs submodule"
+git submodule update --init Macaulay2/editors/emacs
```
mahrud, line 7: Why just this one, as opposed to all of them?

d-torrance: We may not need the others.  In the autotools build, the submodules that `configure` tells us we need to build M2 are updated in the corresponding Makefile in `M2/libraries/`.

.gitignore
d-torrance: There's apparently an `ignore` option that we could add to `.gitmodules` that might also help out with this (see https://git-scm.com/docs/gitmodules).  I'll play around with that.

M2/.gitignore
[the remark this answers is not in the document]
d-torrance: That sounds like a good idea.  We currently have 20 different `.gitignore` files!  A bunch of them look redundant.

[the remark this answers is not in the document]
mahrud: Some may be useful to keep separate (e.g. one in the BUILD directory), but yeah 20 seems way too much.

.gitignore
mahrud: I think we want to see in `git status` when it has changed, but not add it with `git commit -a`.

M2/autogen.sh
d-torrance: It would.  But so would all the other submodule stuff!

(In the Debian package, where we do build from a tarfile, this step is skipped entirely and a different Debian-specific program -- `dh_autoreconf` -- is run to generate the configure script.)

mahrud: I think since this script is just two commands, it would be worth checking if `../.git` exists (either as a directory or file in case of a worktree) and if not skipping the second command and giving a warning that some other tarfiles may need to be downloaded.

d-torrance: Thinking about this some more, I don't think a warning is sufficient.  If we're not working in a clone of the git repository, then the build will certainly fail because we rely on git submodules, and we don't have a good way of fetching them without git.  I'll add an error message.

It might be good to distribute a tarball (probably generated using `git archive`) that includes all the submodules to support git-less builds.  The configure script could be pre-generated (which is standard for autotools projects) so that users wouldn't need to run this script.

mahrud: But if someone has done this already, why give an error and fail?

M2/autogen.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+set -e
+
+if ! git rev-parse --is-inside-work-tree > /dev/null 2>&1
```
mahrud, line 5: This is not good enough. e.g. I have track my home directory dot-files with git, so if I extract a tarfile in my home directory, it's still in a git tree. I think you specifically want to check that `../.git` exists as a file or directory.

That said, I still don't think this should straight up fail.

M2/Macaulay2/tests/ComputationsBook/.gitignore
mahrud, line 1: I think leave this here.

M2/Macaulay2/editors/.gitignore
mahrud, line 1: I think leave these here too. They are very specifically for this directory, and if someone made a change they might not notice that they also need to change the root level .gitignore.

.gitignore
```diff
@@ -1,18 +1,145 @@
...
+*.out
+*.trim
+*~
+.#*
+.*.swp
+.DS_Store
+/M2/.completed
+/M2/.idea
+/M2/BUILD/CommonStagingArea
+/M2/BUILD/build/Macaulay2
```
mahrud, line 14: I think I should emphasize that combining literally every .gitignore is a bad idea, for several reasons:
1. if I make a copy of the `BUILD/build` directory, I want the .gitignore to be copied with it and stay local to that directory.
2. certain directory specific things are harder to maintain if they're clumped together in this file.
3. if we rename a directory, e.g. `M2/distributions`, we want the .gitignore to move with it.

Also, I personally think I prefer 20 .gitignores to a giant laundry list like this.

M2/autogen.sh
d-torrance: Ah, ok -- good point re. `git rev-parse --is-inside-work-tree` v. checking for `../.git`.

Let me make sure I'm clear about your original suggestion.  It basically the following?

```
if inside a clone of the M2 git repo:
    update M2-emacs submodule
else:
    print some sort of warning
```
What would the warning entail?  Right now, it's not possible to build M2 using just a tarball unless you make some modifications ahead of time to avoid the use of git (e.g., for the Debian package, the M2-emacs submodule is vendored inside the tarball and we use Debian packages for everything else).

This script (and the Makefile that preceded it) is designed to be used by folks who have cloned the repository and plan on using the autotools to build M2.  So it really only makes sense to be used inside the git repository.

.gitignore
d-torrance: Fair enough.  I'll revert that commit and just go back to the one with the `*~` files.  It was outside of the original scope of this PR anyway.

M2/autogen.sh
mahrud: If I extract a tarball, and also extract tarballs of a handful of submodules, then run this script, why would the build system fail? At least the CMake build doesn't, but maybe there's a step in the autotools build that I'm forgetting.

M2/autogen.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+
+set -e
+
+echo "-- Generating configure script"
+autoreconf --verbose --force --install
+
+if test ! -f Macaulay2/editors/emacs/M2.el
+then
+    if test -d ../.git
```
mahrud, line 10: If you use git's worktree feature, `.git` can be a file with the address of the real `.git` directory in it, hence why I've been saying the check should be directory _or_ file.

M2/autogen.sh
```diff
@@ -0,0 +1,23 @@
...
+    then
+	echo "-- Updating M2-emacs submodule"
+	git submodule update --init Macaulay2/editors/emacs
+    else
+	echo "-- Warning: not in git repository"
+	echo "-- Downloading and extracting M2-emacs tarball"
```
mahrud, line 16: Is this necessary? e.g. what if they didn't want the emacs mode? I think a warning should suffice, no?

mahrud:
I may not be the best person to approve this, given that I don't use the autotools build, but this looks good to me!

Issue #3345, opened 2024-07-04
bug in part(Complex)


mahrud:
The code:
```m2
needsPackage "Complexes"

R = QQ[x,y]
C = complex R
part(0, C)
part(0, C[1] ++ C[-1])
```
Outputs:
```m2
i4 : part(0, C)

       1
o4 = QQ
      
     0

o4 : Complex

i5 : part(0, C[1] ++ C[-1])

             1
o5 = 0 <-- QQ
            
     0     1

o5 : Complex
```

Also, in conversation with @ggsmith today we realized this is very similar to the `basis` command, except it's oddly not defined for graded modules and maps between them, which I think should be defined.

mahrud:
Fixed in #3522.

Issue #3352, opened 2024-07-08
bug in map and image


mahrud:
What is happening here?
```m2
i1 : f = vars(QQ[x,y]);

i2 : map(image f, , f)
stdio:2:1:(3): error: expected matrix and new target module to agree
```
Note: I understand that `inducedMap` works here, but that method can be computationally expensive (e.g. computes Grobner bases in some cases) which seems unnecessary for what the code above should do.

PR #3522, opened 2024-10-16
Add `part` for graded objects and other improvements

mahrud:
Summary of changes:
- **added part for modules, matrices, ideals, and rings**
  - this is the same as `part(ZZ, Complex)`, but for other graded objects
- **added basis(ZZ, Complex)**
- **added basis(ZZ, ComplexMap)**
  - this is a similar to `part(ZZ, Complex)`, but the base ring is the original ring, not its coefficient ring
- **simplified part(ZZ, Complex)**
  - also fixes https://github.com/Macaulay2/M2/issues/3345
- **simplified part(ZZ, ComplexMap)**
- **added better algorithm for basis(ZZ, Complex)**
- **added better algorithm for truncate(ZZ, Complex)**
- **added better algorithm for truncate(ZZ, ComplexMap)**
  - this is way faster for large enough complexes
- **added tests for basis of Complex and ComplexMap**
- **added method for truncating source of map more than target**
  - we use this frequently in Varieties

The rest are optimizations in line with https://github.com/Macaulay2/M2/issues/3352 to eliminate unnecessary gb computations, though there's more work left.
- **cached subquotient gens in M.cache.Monomial**
- **optimized basis(ZZ, Matrix)**
- **used ??= to cache image, coker, coimage**
- **added ways to avoid gb calls in inducedMap**

### DIFF
--- a/M2/Macaulay2/m2/matrix1.m2
+++ b/M2/Macaulay2/m2/matrix1.m2
@@ -311,14 +311,15 @@ subquotient(Nothing,Matrix) := (null,relns) -> (
 	  Mparts = append(Mparts, symbol relations => relns);
 	  );
      new Module of Vector from hashTable Mparts)
-subquotient(Matrix,Nothing) := (subgens,null) -> (
-     R := ring subgens;
-     E := target subgens;
-     rE := E.RawFreeModule;
-     subgens = align matrix subgens;
+subquotient(Matrix, Nothing) := (subgens0, null) -> (
+    R := ring subgens0;
+    E := target subgens0;
+    rE := E.RawFreeModule;
+    subgens := align matrix subgens0;
      if E.?generators then subgens = E.generators * subgens;
      Mparts := {
 	  symbol cache => new CacheTable from { 
+	    symbol Monomials => subgens0,
 	       cache => new MutableHashTable	    -- this hash table is mutable, hence has a hash number that can serve as its age
 	       },
 	  symbol RawFreeModule => rE,
@@ -330,16 +331,16 @@ subquotient(Matrix,Nothing) := (subgens,null) -> (
 	  Mparts = append(Mparts, symbol relations => E.relations);
 	  );
      new Module of Vector from hashTable Mparts)
-subquotient(Matrix,Matrix) := (subgens,relns) -> (
-     R := ring relns;
-     E := target subgens;
+subquotient(Matrix, Matrix) := (subgens0, relns) -> (
+    R := ring relns;
+    E := target subgens0;
      if E =!= target relns then error "expected maps with the same target";
      rE := E.RawFreeModule;
      n := rawRank rE;
      if n == 0 then new Module from (R,rE)
      else (
 	  relns = align matrix relns;
-	  subgens = align matrix subgens;
+	subgens := align matrix subgens0;
 	  if E.?generators then (
 	       relns = E.generators * relns;
 	       subgens = E.generators * subgens;
@@ -347,6 +348,7 @@ subquotient(Matrix,Matrix) := (subgens,relns) -> (
 	  if E.?relations then relns = relns | E.relations;
 	  Mparts := {
 	       symbol cache => new CacheTable from { 
+		symbol Monomials => subgens0,
 		    cache => new MutableHashTable	    -- this hash table is mutable, hence has a hash number that can serve as its age
 		    },
 	       symbol RawFreeModule => rE,
--- a/M2/Macaulay2/packages/Complexes/ChainComplex.m2
+++ b/M2/Macaulay2/packages/Complexes/ChainComplex.m2
@@ -750,16 +750,24 @@ truncate(ZZ, Complex) := Complex => {} >> opts -> (e, C) -> truncate({e}, C)
 --------------------------------------------------------------------
 -- basis -----------------------------------------------------------
 --------------------------------------------------------------------
+importFrom_Core { "inducedBasisMap" }
+
 -- returns the graded component of the complex in the given degree
 -- as a complex over the same ring (as opposed to the coefficient ring)
 -- TODO: also define basis given a degree range and infinite ranges
--- TODO: also define for ComplexMap
 basis(ZZ,   Complex) :=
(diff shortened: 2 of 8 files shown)


M2/Macaulay2/packages/Complexes/ChainComplex.m2
```diff
@@ -472,6 +472,24 @@ texMath Complex := String => C -> (
...
+    if lo == hi
+    then complex(super C_lo, Base => lo)
+    else complex applyValues(C.dd.map, super))
+
+ambient Complex := Complex => C -> (
+    (lo, hi) := concentration C;
+    if lo == hi
+    then complex(ambient C_lo, Base => lo)
+    else complex applyValues(C.dd.map, ambient))
+
```
ggsmith, line 492: It seems to me that ``cover``, ``super``, and ``ambient`` would not typically produce well-defined complexes.  Am I missing something?

M2/Macaulay2/packages/Complexes/ChainComplex.m2
```diff
@@ -717,30 +735,54 @@ canonicalTruncation(Complex,InfiniteNumber,InfiniteNumber) :=
...
-            );
-        if # keys maps === 0 then complex(psi source basis(deg, C_lo), Base => lo)  else complex maps
-        )
-    )
-part(ZZ, Complex) := Complex => (deg, C) -> part({deg}, C)
+	-- this construction requires ~half as many truncations
+	f := truncate(degs, dd^C_lo, opts);
+	complex hashTable for i from lo+1 to hi list i => (
+	    f = inducedTruncationMap(source f, truncate(degs, C_i, opts), dd^C_i))
+    ))
```
ggsmith, line 754: Can we avoid calling ``Truncation`` within ``part``?

M2/Macaulay2/packages/Complexes/ChainComplexMap.m2
```diff
@@ -247,6 +247,13 @@ net ComplexMap := Net => f -> (
 
 texMath ComplexMap := String => f -> texMath expression f
 
+cover ComplexMap := ComplexMap => f -> (
+    map(cover target f, cover source f, i -> cover f_i, Degree => degree f))
+super ComplexMap := ComplexMap => f -> (
+    map(super target f, super source f, i -> super f_i, Degree => degree f))
+ambient ComplexMap := ComplexMap => f -> (
+    map(ambient target f, ambient source f, i -> ambient f_i, Degree => degree f))
+
```
ggsmith, line 256: Again, I suspect that these methods are generally not well-defined.

M2/Macaulay2/packages/Complexes/ChainComplex.m2
```diff
@@ -717,30 +735,54 @@ canonicalTruncation(Complex,InfiniteNumber,InfiniteNumber) :=
...
+    then complex(image basis(deg, C_lo, opts), Base => lo)
+    -- this is the simplest way to take the basis of the whole complex:
+    -- else complex applyValues(C.dd.map, f -> basis(deg, f, opts)))
+    else (
+	-- this construction requires ~half as many basis computations
+	f := basis(deg, dd^C_lo, opts);
+	complex hashTable for i from lo+1 to hi list i => (
+	    f = inducedBasisMap(source f, image basis(deg, C_i, opts), dd^C_i))
+    ))
+
```
ggsmith, line 776: The phrase ``basis`` of complex seems particularly unfortunate.

M2/Macaulay2/packages/Complexes/ChainComplex.m2
mahrud: Do you have an example in mind? I think they should be functorial, but maybe I'm missing something.

M2/Macaulay2/packages/Complexes/ChainComplex.m2
mahrud: We do not call truncate within part. We do however call basis.

mahrud:
> From my perspective, the proposed changes need (at a minimum) documentation and tests before being accepted.

Tests have been included, I'm happy to add some documentation this week if otherwise the changes are acceptable.

> In particular, I suspect that several of the methods do not typically return well-defined complexes or complex maps.

I'm happy to remove `super` and `ambient` for now if you suspect those are not well-defined (though I'm still curious to know why you suspect that or whether there are obvious examples I'm missing), but I use `cover` in the simplified version of `part`. Are there specifically issues for `cover` of a complex?

Alternatively, would it make sense to have an internal method for `cover Complex` that is used in `part`?

mahrud:
I removed cover, super, ambient, and changed part to use an unexported `cover'` method.

mahrud:
@ggsmith if all the changes you requested are addressed, is this okay to be merged?

M2/Macaulay2/m2/matrix1.m2
```diff
@@ -311,14 +311,15 @@ subquotient(Nothing,Matrix) := (null,relns) -> (
...
-     subgens = align matrix subgens;
+subquotient(Matrix, Nothing) := (subgens0, null) -> (
+    R := ring subgens0;
+    E := target subgens0;
+    rE := E.RawFreeModule;
+    subgens := align matrix subgens0;
      if E.?generators then subgens = E.generators * subgens;
      Mparts := {
 	  symbol cache => new CacheTable from { 
+	    symbol Monomials => subgens0,
```
mikestillman, line 322: Why are `subgens0` called Monomials?  subgens0 can be any matrix at all?

mikestillman (approved):
There are a few things I still need to puzzle out, including the exact lines, and what is F.cache.Monomials supposed to represent?

But it looks clean, and tests run, and having the truncate sorted out, along with the induced maps, is good.

I suggested adding a couple of tests, and I asked about the Monomials usage too.  Perhaps a comment about that and how it is used would help the next person who needs to check this code in the future.

I agree with Greg that the name `basis` is unfortunate for Complexes, but this comes back to a name I decided to use in Macaulay, when I was a postdoc...  So maybe soon we can make a synonym for it, but we can think about that later.

Thanks for the contribution!

PR #3528, opened 2024-10-19
Include bison-generated header files first in html-check-links

d-torrance:
Otherwise, we may end up including them from elsewhere (e.g., there's a "grammar.h" distributed with some versions of Python).

Closes: #2398

(Draft for now -- this works locally but I'd also like to check on a RHEL 9 GitHub build.)


d-torrance:
This worked!

https://github.com/d-torrance/M2-workflows/actions/runs/11420709080/job/31776724515

(The build still ended up failing during `make check`, but `html-check-links` worked!)

d-torrance:
I'm going to go ahead and merge this small patch.

PR #3529, opened 2024-10-20
Detect TOPCOM package on Arch Linux

d-torrance:
It prepends "topcom-" to just some of the binaries.  This means we need to check two binaries (one with the prefix and one without) to distinguish the Arch package from the Debian one (which prepends "topcom-" to everything).

From https://gitlab.archlinux.org/archlinux/packaging/packages/topcom/-/blob/main/PKGBUILD?ref_type=heads:

```shell
  for _prog in binomial cross cube cyclic lattice; do
    mv "$pkgdir"/usr/bin/{,topcom-}$_prog
  done
```


mikestillman:
@d-torrance This worked for me, I OK this pull request.

mikestillman (approved):
I approve this pull request

PR #3530, opened 2024-10-20
Minimize filename during TestInput test

d-torrance:
This should fix https://github.com/Macaulay2/M2/pull/3419#discussion_r1807967394

PR #3532, opened 2024-10-21
Minimize filename during TestInput tests (take 2)

d-torrance:
The filename appears in several different checks, so we minimize it at the beginning.

This *should* fix https://github.com/Macaulay2/M2/pull/3530#issuecomment-2427876700.  Previously, if I ran M2 from `/tmp` (so the relative path to the temporary file would be missing the "/tmp"), it would fail, but after this change, it passes.


mahrud (dismissed):

mahrud:
Something is still off. Try downloading this file in `/tmp` and running it there. It keeps failing regardless of how I run it.

mahrud:
Ah never mind, I was using an old M2 to test it.

PR #3385, opened 2024-07-31
Updated release checklist

mahrud:
- **limited package headlines to 100 characters**
  - also fixes the top node check, which wasn't working.
- **removed Reload => true from packages**
- **implemented uniformity in headlines (cf. 7a7e6f96)**
- **removed line breaks from package headlines**

I'll seek approval from authors of affected packages in the comments.

PR #3539, opened 2024-10-25
Update RealRoots package to version 1.0 to match JSAG

d-torrance:
This PR updates the RealRoots package so that it (almost) matches the [published version](https://msp.org/jsag/2024/14-1/jsag-v14-n1-x10-RealRoots.m2) in JSAG.  This is so we can properly add it to the list of "certified" packages for the forthcoming 1.24.11 release.

The one difference between the proposed version of the package and the JSAG-published version is that the `Headline` was shortened in #3385.

One thing I'm not sure about -- @cel34-bath submitted a couple PR's with updates to this package earlier this year (#3100, #3153).  This part of the code (the `realRootsIsolation` method) was also changed in the version published in JSAG.  Does the published version incorporate these changes?  Or should they be restored?

Cc: @JordyLopez27, @kmaluccio, @FrankSottile, @tjyahl, @cel34-bath

### DIFF
--- a/M2/Macaulay2/packages/RealRoots.m2
+++ b/M2/Macaulay2/packages/RealRoots.m2
@@ -1065,7 +1065,7 @@ document {
 		hurwitzMatrix(f,4)
 	        hurwitzMatrix(f,3)	      
 	 	 ///,
-	PARA{"We can also use mutliple variables to represent unknown coefficients. Note that we create another ring ",TT "S"," so 
+	PARA{"We can also use multiple variables to represent unknown coefficients. Note that we create another ring ",TT "S"," so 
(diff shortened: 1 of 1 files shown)


cel34-bath:
> One thing I'm not sure about -- @cel34-bath submitted a couple PR's with updates to this package earlier this year (#3100, #3153). This part of the code (the `realRootsIsolation` method) was also changed in the version published in JSAG. Does the published version incorporate these changes? Or should they be restored?

Hi, sorry for the delay. I have merged these updates together in https://github.com/Macaulay2/M2/pull/3708. Let me know if this is correct and wanted, as a package I am in the process of submitting relies on these updates.

cel34-bath:
Yes, I think this looks good to me! Thank you!

Issue #3417, opened 2024-08-20
debugger fails to find location of calls to "error" in Core


mahrud:
@mikestillman pointed out something which I had also observed and noted on other issues here, which is that the debugger often doesn't end up in the right place in the Core.

Here is how to reproduce: insert `blah = x -> error "debug me"` somewhere in Core, then run it with `errorDepth` one or less. Here is the result for me:
```m2
i1 : debug Core

i2 : code blah

o2 = ../../m2/debugging.m2:6:7-6:28: --source code:
     blah = x -> error "debug me"

i3 : errorDepth=1

o3 = 1

i4 : blah 12
../../m2/debugging.m2:20:13:(1): error: debug me
../../m2/debugging.m2:20:13:(1): --entering debugger (type help to see debugger commands)
../../m2/debugging.m2:20:5-20:30: --source code:
     olderror processArgs args)

ii5 : currentString

ii6 : code current

oo6 = ../../m2/debugging.m2:20:5-20:30: --source code:
           olderror processArgs args)

ii7 : locate current

oo7 = ../../m2/debugging.m2:20:5-20:30

oo7 : FilePosition
```
I don't seem to be able to get the original location in anyway.
```m2
ii8 : listLocalSymbols

oo8 = symbol  class  value  location of symbol
      ------  -----  -----  ------------------

ii9 : listUserSymbols

oo9 = symbol  class  value  location of symbol
      ------  -----  -----  ------------------

ii10 : x

oo10 = x

oo10 : Symbol
```
One interesting factor is that avoiding tail-end optimization of `error` leads to the symbols remaining accessible. If I instead add `blah = x -> (error "debug me";)` then I get:
```m2
i2 : code blah

o2 = ../../m2/debugging.m2:6:7-6:31: --source code:
     blah = x -> (error "debug me";)

i3 : errorDepth=1

o3 = 1

i4 : blah 12
../../m2/debugging.m2:20:13:(1):[1]: error: debug me
../../m2/debugging.m2:20:13:(1):[1]: --entering debugger (type help to see debugger commands)
../../m2/debugging.m2:20:5-20:30: --source code:
     olderror processArgs args)
```
So far identical, but:
```m2
ii8 : listLocalSymbols

oo8 = symbol              class             value                                                 location of symbol
      ------              -----             -----                                                 ------------------                
      x                   ZZ                12                                                    ../../m2/debugging.m2:6:7-6:8     
[...]

PR #3544, opened 2024-10-26
Revert "fix error location in tail call" to fix the debugger

mahrud:
Reverts Macaulay2/M2#3144 in order to fix https://github.com/Macaulay2/M2/issues/3417 (debugger not finding the correct location of errors in Core).

### DIFF
--- a/M2/Macaulay2/d/evaluate.d
+++ b/M2/Macaulay2/d/evaluate.d
@@ -29,7 +29,6 @@ export nextS := setupvar("next", nullE);
 export applyIteratorS := setupvar("applyIterator", nullE);
 export joinIteratorsS := setupvar("joinIterators", nullE);
 
-handleError(c:Code,e:Expr):Expr;
 eval(c:Code):Expr;
 applyEE(f:Expr,e:Expr):Expr;
 export evalAllButTail(c:Code):Code := while true do c = (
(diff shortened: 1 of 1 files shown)


mahrud:
@pzinn I'll try to find an alternative fix for this issue before the next release, but I believe fixing the debugger is more important than the error location. If you can think of another fix let me know.

d-torrance:
This is what I'm getting after building this branch:

```m2
i1 : debug Core; errorDepth = 1; blah 12
stdio:1:32:(3): error: debug me
```

I put `blah` on line 2 of "debugging.m2" just like the example in #3417.   It's not giving us the `olderror processArgs args` output any more, but is `stdio` the expected output?

mahrud:
I noticed this too, but try turning `blah` into a method. It would be good to fix that as well.

d-torrance:
Ah, ok!  As a method, this works great:

```m2
i4 : blah 12
/home/profzoom/src/macaulay2/M2/M2/Macaulay2/m2/debugging.m2:7:21:(1):[1]: error: debug me
/home/profzoom/src/macaulay2/M2/M2/Macaulay2/m2/debugging.m2:7:21:(1):[1]: --entering debugger (type help to see debugger commands)
/home/profzoom/src/macaulay2/M2/M2/Macaulay2/m2/debugging.m2:7:16-7:32: --source
code:
blah ZZ := x -> error "debug me"
```

d-torrance (approved):

pzinn:
OK. I didn't have time to look into #3417 myself.

Issue #3524, opened 2024-10-18
bug in try, alarm, and ??= (augmented null coalescing assignment operator)


mahrud:
Try this, taken from `tests/normal/alarm.m2`:
```m2
R = QQ[vars(0..24)]
f = () -> (alarm 4; try res coker vars R else "ran out of time")
time f()

cokernel Matrix := Module => f -> f.cache.cokernel ??= subquotient(null, f)

R = QQ[vars(0..24)]
f = () -> (alarm 4; try res coker vars R else "ran out of time")
time f()
```
After the first run we get:
```m2
i3 : time f()
 -- used 5.55149s (cpu); 3.82313s (thread); 0s (gc)

o3 = ran out of time
```
But the second run never finishes.
I believe this is what's failing in #3517 and #3522 (removed commits: c07017ec1a3f89b5aa0fba22f93a502ea5a64454 and 9b29e3df4df0eecae5c34e91b1460df934d79eda).

@d-torrance could you take a look?

d-torrance:
I think I've zeroed in on the problem a bit more.  Certain errors break `alarm` using `try` or `??`:

```m2
i1 : alarm 1; try ZZ.foo else sleep 2

o2 = 0

i3 : alarm 1; ZZ.foo ?? sleep 2

o4 = 0
```
But other errors are fine:

```m2
i5 : alarm 1; try 1/0 else sleep 2
stdio:3:9:(3): error: alarm occurred

i7 : alarm 1; 1/0 ?? sleep 2
stdio:4:13:(3): error: alarm occurred
```

mahrud:
That's interesting!
(ps: I know I added this to 1.24.11 milestone, but if you don't have time, it's not an absolute necessity)

d-torrance:
I should have some time this week -- I'd like to get this figured out!

It looks like it's just "key not found" errors.  Also, I checked the 1.24.05 package in Debian unstable and this bug wasn't present.  So this might be related to the changes from #3409...

d-torrance:
I think I figured it out!  When keys aren't found, we call `Thing.RobustPrintMethod`, which itself calls `try`:

```m2
i1 : debug Core

i2 : code Thing.RobustPrintMethod

o2 = /usr/share/Macaulay2/Core/robust.m2:62:26-63:76: --source code:
     Thing.RobustPrintMethod = (msg, obj) -> try toString stack { msg | ":",
[...]

PR #3545, opened 2024-10-26
Don't use robust "key not found" error when errors are suppressed

d-torrance:
The robust error message involves calling `alarm`, which will cancel any alarms we might be trying to catch using `try` or `??`.

Closes: #3524

### DIFF
--- a/M2/Macaulay2/d/hashtables.dd
+++ b/M2/Macaulay2/d/hashtables.dd
@@ -318,10 +318,11 @@ lookup(object:HashTable, key:Expr):Expr; -- forward declaration
 KeyNotFound(object:string, key:Expr):Expr := (
     -- TODO: implement a similar trick to call synonym(object)
     msg := "key not found in " + object;
-    see := lookup(Class(key), RobustPrintE);
-    if see != notfoundE then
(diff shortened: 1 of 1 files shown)


M2/Macaulay2/d/hashtables.dd
```diff
@@ -318,10 +318,11 @@ lookup(object:HashTable, key:Expr):Expr; -- forward declaration
 KeyNotFound(object:string, key:Expr):Expr := (
```
mahrud, line 318: Does this work?
```suggestion
KeyNotFound(object:string, key:Expr):Expr := (
    if SuppressErrors then return nullE;
```
No reason to go through `buildErrorPacket` if we don't have to.

mahrud (approved):

M2/Macaulay2/d/hashtables.dd
d-torrance: That would work fine for `??` but not `try`.  we still want it to raise an error.  For example, with this change:

```m2
i1 : try ZZ.foo then "no error" else "error"

o1 = no error
```

PR #3547, opened 2024-10-27
Added file: AbstractSimplicialComplexes.m2

n-m-g:
This package gives a way to work with simplicial complexes implemented as graded lists.  It complements the existing M2 package SimplicialComplexes.m2.

PR #3548, opened 2024-10-28
Update AbstractSimplicialComplexes.m2 and SpectralSequences.m2

n-m-g:

### DIFF
--- a/M2/Macaulay2/packages/SpectralSequences.m2
+++ b/M2/Macaulay2/packages/SpectralSequences.m2
@@ -1,6 +1,6 @@
 -- -*- coding: utf-8 -*-
 --------------------------------------------------------------------------------
--- Copyright 2012  Gregory G. Smith
+-- Copyright 2024  the joint authors
 --
 -- This program is free software: you can redistribute it and/or modify it under
 -- the terms of the GNU General Public License as published by the Free Software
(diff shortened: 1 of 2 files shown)


d-torrance:
This PR is using the same branch as #3547 -- closing

M2/Macaulay2/packages/SpectralSequences.m2
```diff
@@ -1546,325 +1476,111 @@ doc ///
 
 doc ///
     Key
-      "Identifying anti-podal points of the two sphere"
+      "Spectral sequences and non-Koszul syzygies"
```
mahrud, line 1479: Is there just some reordering of the documentation nodes happening here? The differences looks very large so it's hard to tell.

n-m-g:
No, this is a larger update making the package forward compatible with the Complexes update.  So many parts of the code have changed and become reordered in various parts.

Best,
Nathan

mahrud:
I see. I would suggest splitting the package into several smaller files, similar to `NormalToricVarieties` or `Complexes` or `Varieties`, or even much simpler:
```
SpectralSequences.m2
SpectralSequences/docs.m2
SpectralSequences/tests.m2
```

n-m-g:
Many thanks for these suggestions.  Eventually, we may try to do that with future updates.

Best,
Nathan

PR #3551, opened 2024-10-29
Update SpectralSequences.m2

n-m-g:
Make forward compatible following suggestion of Mike and Greg.

### DIFF
--- a/M2/Macaulay2/packages/SpectralSequences.m2
+++ b/M2/Macaulay2/packages/SpectralSequences.m2
@@ -1,6 +1,6 @@
 -- -*- coding: utf-8 -*-
 --------------------------------------------------------------------------------
--- Copyright 2012  Gregory G. Smith
+-- Copyright 2024  the joint authors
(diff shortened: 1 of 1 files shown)


M2/Macaulay2/packages/SpectralSequences.m2
```diff
@@ -2943,52 +2730,50 @@ doc ///
      	  Text
 	     Produces the filtered complex obtained by successively truncating the complex.
 	  Example 
-	    needsPackage "SpectralSequences"
+	    needsPackage "SpectralSequences2"
```
mahrud, line 2733: Here's an example of "SpectralSequences2" in the package. You could either search & replace all to remove the "2" or, better yet, simply remove the whole line, because examples don't need to reload the package.

n-m-g: Oh.  Yes.  Those need to be removed!

PR #3552, opened 2024-10-29
Update SpectralSequences.m2

n-m-g:

### DIFF
--- a/M2/Macaulay2/packages/SpectralSequences.m2
+++ b/M2/Macaulay2/packages/SpectralSequences.m2
@@ -1,6 +1,6 @@
 -- -*- coding: utf-8 -*-
 --------------------------------------------------------------------------------
--- Copyright 2012  Gregory G. Smith
+-- Copyright 2024  the joint authors
 --
 -- This program is free software: you can redistribute it and/or modify it under
 -- the terms of the GNU General Public License as published by the Free Software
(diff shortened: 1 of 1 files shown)


mahrud:
In general, you don't need to open new pull requests. You can push (or force push) to the same branch to update an existing pull request. I closed the other one for you.

n-m-g:
Thanks.  I’ve updated the AbstractSimplicialComplexes in my branch.  Can you see those changes?  Do I need to issue a pull request for that?

Best,
Nathan

n-m-g:
I've closed this as the SpectralSequences update has been pushed elsewhere and the AbstractSimplicialComplexes pull request is continuing on another pull request.

mahrud:
I don't see another pull request with SpectralSequences currently.
@mikestillman are you going to include this in the complexes PR?

n-m-g:
Hi,

Again, I am happy to issue a new pull request with the updated SpectralSequences package but I believe that the updates we made have (at the request of Mike and Greg) already been pushed into the development branch.  Again, we are limited somewhat with what we can do to make everything entirely forward compatible because we don’t know 100% what the “core static” features of the “Complexes” package actually will be.  The other main issue is the current “Complexes” compatibility with the existing “SimplicialComplexes” package.  The intent is to make additional updates in due time as our time allows but also as things become in a more static form within the M2 system.

Best,
Nathan

Issue #2907, opened 2023-07-21
RingMap from WeylAlgebra to AssociativeAlgebra is broken


mahrud:
Here's a minimal example, note the swapped order `x,dx` and `dy,y`:
```m2
needsPackage "AssociativeAlgebras"

D = ZZ/101[x,dx, WeylAlgebra => {x => dx}]
R = ZZ/101<|dy,y|>/(dy*y - 1 - y*dy)

f = map(R, D, {y,dy})
f(dx*x) -- gives y*dy + 2, should be y*dy+1
f(x*dx) -- gives y*dy + 1, should be y*dy
```

Related: #2828

cc: @christineBerkesch

mdebellevue:
After chatting with @mikestillman about this, I tried changing ringmap.cpp to add an explicit check for the weyl algebra, with an accompanying block that is the same as for other noncommutative rings, but with a multiplication reversed in order.  It didn't seem to fix the problem though.  But I don't trust that I compiled from source correctly/that my new block is actually executing, because a print command that I put in isn't triggering when it should.  [See here for the change I attempted](https://github.com/mdebellevue/M2/commit/9f39d3dd868ee6350b1906633be9685b6dcd75f5).

mahrud:
@mdebellevue do you think you could open a pull request with your changes? That would run the tests and Mike could review it.

mdebellevue:
Sorry for dropping the ball on this for so long, I'm assuming that the pull request you made supersedes mine?

PR #3556, opened 2024-10-30
Fixing ring maps from WeylAlgebra to AssociativeAlgebra

mahrud:
@mikestillman this is the change suggested in https://github.com/Macaulay2/M2/issues/2907#issuecomment-1741490215 by @mdebellevue. Could you have a look at why it's not working?

I emailed you two weeks ago that we need to have #2907 fixed before the upcoming workshop. If that's not possible, we need to know so we can change plans.

### DIFF
--- a/M2/Macaulay2/e/M2FreeAlgebra.hpp
+++ b/M2/Macaulay2/e/M2FreeAlgebra.hpp
@@ -50,6 +50,7 @@ public:
   virtual const M2FreeAlgebraOrQuotient * cast_to_M2FreeAlgebraOrQuotient()  const { return this; }
   virtual       M2FreeAlgebraOrQuotient * cast_to_M2FreeAlgebraOrQuotient()        { return this; }
 
+  bool is_commutative_ring() const { return false; }
 };
 
 class M2FreeAlgebra : public M2FreeAlgebraOrQuotient
--- a/M2/Macaulay2/e/M2FreeAlgebraQuotient.hpp
+++ b/M2/Macaulay2/e/M2FreeAlgebraQuotient.hpp
@@ -45,7 +45,9 @@ public:
   
   int numVars() const { return monoid().numVars(); }
   int n_vars() const { return numVars(); }
-  
+
+  bool is_quotient_ring() const { return true; }
+
   // these are all the functions from Ring that must exist for M2FreeAlgebraQuotient to be instantiated
   virtual int index_of_var(const ring_elem a) const;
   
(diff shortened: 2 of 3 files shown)


mikestillman:
@mahrud Sorry, I'm confused.  This example works on my machine (with your changes).  I thought you needed me to fix a bug?  If not, let's just accept this pull request (I guess we should make sure it passes all tests, they are running now).

mikestillman:
Ah, I see, you were able to fix it after asking me about it.  That's good!

mahrud:
This is failing a couple of assertions in AssociativeAlgebras. One of them is fixable (probably the test was wrong or ambiguous) but this one seems like a problem:
```m2
  needsPackage "AssociativeAlgebras"
  R = QQ<|b,c,d|>
  F2 = map(R,R,{c+b,c,d})
  F2(b*c) -- gives c*(b+c), should be (b+c)*c
  F2(c*b) -- gives (b+c)*c, should be c*(b+c)
```

mdebellevue:
Sorry, this is probably just due to my inexperience with git and/or totally loosing the thread on what's going on with this issue, but just to check that I have the situation straight: the fix I originally wrote added an explicit check for WeylAlgebras to ringmap.cpp, and your fix instead adds an IsCommutative and IsQuotient to the free algebra class.  Looking at this with my eyes now, I'm not sure why either of these changes would have fixed the issue?  Unless I'm missing something. in the original ringmap.cpp is_commutative_ring() should be returning false for both WeylAlgebras and FreeAlgebra (or quotients theoreof), and the call to cast_to_SchurRing() shouldn't be affecting anything.  So in either of our fixes, the code used to evaluate a ring map is still the chunk from 171 of ringmap.cpp:
```
      for (index_varpower i = vp; i.valid(); ++i)
        {
          int v = first_var + i.var();
          int e = i.exponent();
          ring_elem g;
          if (e >= 0)
            g = _elem[v].bigelem;
          else
            g = R->invert(_elem[v].bigelem);
          for (int j = 0; j < e; j++)
            {
              assert(v < nvars);
              ring_elem tmp = R->mult(g, result);
              R->remove(result);
              result = tmp;
            }
        }

moorewf:
I think I have fixed this in a new PR -- the idea for the fix was correct, but the is_commutative_ring() was placed in the wrong spot (it should be in M2FreeAlgebra and M2FreeAlgebraQuotient since those are the front-end interfaces to the FreeAlgebra and FreeAlgebraQuotient back-end classes).

mahrud:
Closing this in favor of yours.

mahrud:
> Unless I'm missing something. in the original ringmap.cpp is_commutative_ring() should be returning false for both WeylAlgebras and FreeAlgebra

You're right about everything except the last part: is_commutative_ring for FreeAlgebra was actually true! Now (either with this PR or with Frank's) it is false, so the same block handles it.

PR #3560, opened 2024-10-31
Updated AbstractSimplicialComplexes 

n-m-g:
Updated AbstractSimplicialComplexes to overload "==" and "dim"

### DIFF
--- a/M2/Macaulay2/packages/AbstractSimplicialComplexes.m2
+++ b/M2/Macaulay2/packages/AbstractSimplicialComplexes.m2
@@ -396,7 +396,7 @@ inducedSimplicialChainComplexMap(AbstractSimplicialComplex,AbstractSimplicialCom
 (
     h := simplicialChainComplex H;
     l := simplicialChainComplex L;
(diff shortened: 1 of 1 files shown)


M2/Macaulay2/packages/AbstractSimplicialComplexes.m2
```diff
@@ -400,7 +396,7 @@ inducedSimplicialChainComplexMap(AbstractSimplicialComplex,AbstractSimplicialCom
 (
     h := simplicialChainComplex H;
     l := simplicialChainComplex L;
-    if areEqual(abstractSimplicialComplex {{}},H)==true then return map(l,h,zero)
+    if ((abstractSimplicialComplex {{}}) == H)==true then return map(l,h,zero)
```
d-torrance, line 399: It's redundant to use `if x == true then ...`.  You can drop the `== true` and just use `if x then ...`.

n-m-g:
Yes.  It’s true!  I can remove that as well.

Best,
Natthan

n-m-g:
This is great.  Do let me know if anything else is needed prior to having this distributed.  Many thanks again for everyone's feedback.

PR #3565, opened 2024-10-31
Preload PackageCitations package

d-torrance:
As discussed in M2internals this afternoon


mahrud:
Should we just rename it to `Citation`?

d-torrance:
> Should we just rename it to `Citation`?

I think the standard convention for packages is plural, so maybe `Citations`?  I'm generally not a huge fan of renaming things, though.  I suppose we could have a stub "PackageCitations" package that just exports "Citations" for backwards compatibility?

PR #3566, opened 2024-11-01
Fix annoying typo in VectorGraphics

pzinn:
fix an unfortunate typo in VectorGraphics which breaks `GraphicsObject ++ List`.


d-torrance:
Just in time!  I'll squeeze this into the release

PR #3573, opened 2024-11-04
Run autogen.sh for re-generating configure script from Makefile

d-torrance:
I've run into this bug a few times in the last day or so.  Whenever make thinks that `configure` needs to be updated, it tries to regenerate it the old-fashioned way instead of running autogen.sh.  I missed this change back in #3520.


mahrud:
I'm can't test this, but it is small enough that if you feel confident about it I'm fine with you merging it.

Issue #2467, opened 2022-04-30
Making sets Sets


jchen419:
Currently, the `set` method is only defined on `VisibleList`s. Trying to turn a `Set` into a `Set` gives an error:
```
i1 : methods set

o1 = {0 => (set, VisibleList)}

o1 : NumberedVerticalList

i2 : set set{1,2}
stdio:2:1:(3): error: expected a visible list
```
It seems sensible for `set` to act as the identity on a `Set` (which is how e.g. `sequence` and `toList` behave). Can this be done at top-level?
```
i3 : set Set := Set => identity

o3 = Set => identity

o3 : Option

i4 : methods set

o4 = {0 => (set, Set)        }
     {1 => (set, VisibleList)}

o4 : NumberedVerticalList

i5 : set set{1,2}
stdio:5:1:(3): error: expected a visible list
```

pzinn:
reminds me of https://github.com/Macaulay2/M2/issues/1499. if we made `ideal` idempotent, we could fix this...

DanGrayson:
This is a good idea.   Other things might come along later that can be converted to sets, too, so it would be good to change it from a compiled function to a method function.  I did that recently for `toList`, this way:
```diff
diff --git a/M2/Macaulay2/d/actors2.dd b/M2/Macaulay2/d/actors2.dd
index 761c7fe98..6215a3a34 100644
--- a/M2/Macaulay2/d/actors2.dd
+++ b/M2/Macaulay2/d/actors2.dd
@@ -64,7 +64,7 @@ toList(e:Expr):Expr := (
 	       	    false)))
      is s:stringCell do list(strtoseq(s))
      else WrongArg("list, sequence, set, or string"));
-setupfun("toList",toList);
+setupfun("toList1",toList);
 values(e:Expr):Expr := (
      when e
      is oc:DictionaryClosure do (
diff --git a/M2/Macaulay2/m2/methods.m2 b/M2/Macaulay2/m2/methods.m2
index d578af659..6902dc846 100644
--- a/M2/Macaulay2/m2/methods.m2
+++ b/M2/Macaulay2/m2/methods.m2
@@ -419,6 +419,12 @@ depth String := s -> 0
 
 -----------------------------------------------------------------------------
 
+toList = method(Dispatch => Thing)
+toList List := List => x -> x
+toList BasicList := toList Set := toList String := List => toList1
+
+-----------------------------------------------------------------------------
+
 oldflatten := flatten
 erase symbol flatten
[...]

Issue #1499, opened 2020-09-14
misleading help leadTerm


pzinn:
`help "leadTerm(Ideal)"` says 
```
Compute a Gröbner basis and return the ideal generated by the lead terms of the Gröbner basis elements.`
```
but it's not true: it returns instead the generators of init(ideal) as a matrix.
I'm assuming it's the documentation that's wrong and the method works as intended? (though it would seem to make more sense to return an `Ideal`)

DanGrayson:
Good catch.  Here's the code:
```m2
o16 = -- code for method: leadTerm(Ideal)
      /Users/dan/src/M2/M2.git/M2/Macaulay2/m2/matrix1.m2:513:33-513:59: --source code:
      leadTerm Ideal := Matrix => (I) -> leadTerm generators gb I
```

@mikestillman -- do you have an opinion?

By the way, if leadTerm works for ideals, it should also work for submodules of free modules.

pzinn:
Certainly returning an Ideal would make a lot more sense, but I'm afraid it might break existing code.
(a hacky way to avoid the latter would be to define `ideal Ideal := identity` but that's not great)

DanGrayson:
Well, let's just fix the documentation.  Upward compatibility is important.

PR #3574, opened 2024-11-09
leadTerm Ideal returning an Ideal

pzinn:
This one has been bothering me for a long time.
`leadTerm RingElement` returns a `RingElement`, `leadTerm Matrix` returns a `Matrix`, `leadTerm Vector` returns a `Vector`, but `leadTerm Ideal` ... returns a `Matrix`. Despite the documentation saying it should return an `Ideal`. This obviously needs fixing, but the problem is, a lot of existing code relies on this current incorrect implementation.

I decided to finally take a look at this and fix it the best I could:
- As suggested in the comments of #1499, I added `ideal Ideal := identity` which should mitigate most issues of upwards compatibility.
- I corrected some of the existing code in packages when this wasn't enough.
Incidentally, packages `ExteriorIdeals` and `TSpreadIdeals` define a method `initialIdeal` which by now is strictly equivalent to `leadTerm`. I believe these methods should be deprecated.
- I took this opportunity to also implement the suggestion of #2467.

### DIFF
--- a/M2/Macaulay2/m2/gb.m2
+++ b/M2/Macaulay2/m2/gb.m2
@@ -202,6 +202,7 @@ syz = method(TypicalValue => Matrix, Options => syzDefaults)
 -- rawGBSyzygies doesn't sort its columns, so we do that here
 syz              GroebnerBasis  := Matrix => o -> G -> map(ring G, rawsort rawGBSyzygies G.RawComputation)
 leadTerm         GroebnerBasis  := Matrix =>      G -> map(ring G, rawGBGetLeadTerms(raw G, -1))
+leadTerm     (ZZ,GroebnerBasis) := Matrix =>   (i,G)-> map(ring G, rawGBGetLeadTerms(raw G, i))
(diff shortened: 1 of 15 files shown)


pzinn:
Actually the documentation states that it should return an `Ideal`, so doesn't need changing.
I haven't marked `initialIdeal` as deprecated, though. That could be added to this PR.

d-torrance:
The new methods like `ideal(Ideal)` and `leadTerm(ZZ, GroebnerBasis)` should probably be documented, even if it's a simple as adding keys to existing nodes.

pzinn:
ah, yes, forgot about that. will do.

d-torrance (approved):

d-torrance:
This looks good to me.  Let's go ahead and give this a few more days to give any additional authors of the affected packages a change to take a look.

lkastner (approved):

lucamata (approved):

midipasq (approved):

midipasq (commented):
Thanks for doing this - I think it makes good sense.

PR #3579, opened 2024-11-12
Use brew's readline in macOS autotools GitHub builds

d-torrance:
It's keg-only to avoid conflicting with libedit provided by the system, so we need to specify the paths in order for it to be detected.  Otherwise, we end up building it every time!

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -157,8 +157,12 @@ jobs:
           ../../autogen.sh
           export PYVERSION=`python3 -c "from sys import version_info; \
             print(f'{version_info.major}.{version_info.minor}')"`
-          export CPPFLAGS="-I`brew --prefix`/include -I`brew --prefix libomp`/include"
-          export  LDFLAGS="-L`brew --prefix`/lib -L`brew --prefix libomp`/lib \
+          export CPPFLAGS="-I`brew --prefix`/include \
+            -I`brew --prefix libomp`/include         \
+            -I`brew --prefix readline`/include"
+          export  LDFLAGS="-L`brew --prefix`/lib \
+            -L`brew --prefix libomp`/lib         \
+            -L`brew --prefix readline`/lib       \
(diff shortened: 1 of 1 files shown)


d-torrance:
### Before

From https://github.com/Macaulay2/M2/actions/runs/11762892834/job/32766260824:

```
checking for readline/readline.h... yes
checking for library containing rl_set_prompt... -lreadline
checking for library containing rl_completion_matches... none required
checking for library containing readline... none required
checking for library containing add_history... none required
checking whether readline library is new enough (version at least 6)... no, will build it
configure: readline library will be compiled
```

### After

From: https://github.com/Macaulay2/M2/actions/runs/11789232513/job/32837650836?pr=3579

```
checking for readline/readline.h... yes
checking for library containing rl_set_prompt... -lreadline
checking for library containing rl_completion_matches... none required
checking for library containing readline... none required
checking for library containing add_history... none required
checking whether readline library is new enough (version at least 6)... yes
```

mahrud:
I don't like a setup in which users "in the know" know which flags they need to provide for the build to succeed and others either get stuck or have to build extra libraries.

I also don't care how autotools works really so I'll abstain from reviewing, but my preferred fix would be something like this:
https://github.com/Macaulay2/M2/blob/6a1ee7c608f4a424a5f362010023d40b3ddaf34e/M2/cmake/FindReadline.cmake#L26

d-torrance:
Ooh, I like how cmake checks the brew directories by default!  This PR came out of my experiments putting together the macOS dependencies portion of the [autotools wiki page](https://github.com/Macaulay2/M2/wiki/Building-M2-from-source-using-Autotools), and poor macOS users have to add a *ton* of flags manually because of all these keg-only brew packages.

I'll go ahead and merge this for now to avoid building readline unnecessarily, but I would like to work toward a better solution soon.

mahrud:
I should say, it only checks the brew directories by default for readline and gdbm. At this point it's probably safe to look there for everything there also, but cmake doesn't do that yet.

Issue #3434, opened 2024-08-25
Feature request: cdot or ⋅ as a binary operator


mahrud:
Currently we have a number of unicode synonyms defined in [exports.m2](https://github.com/Macaulay2/M2/blob/ec9e9ac60ed4a8e791448942202f077a88a87e15/M2/Macaulay2/m2/exports.m2#L1277-L1302). I think it would be useful to introduce the [interpuct](https://en.wikipedia.org/wiki/Interpunct) ($\cdot$) as a binary operator. This could be implemented as a synonym to a binary keyword `cdot` (similar to `and`, `or`, etc. but bound to a top-level method).

However, it would even be more interesting if M2-mode (and other editors) could automatically convert `\cdot` to $\cdot$ similar to, for instance, in [agda-mode](https://agda.readthedocs.io/en/v2.7.0/tools/emacs-mode.html#unicode-input), and we could freely type `C⋅D` (without spaces) and have M2 recognize this.

This would be pretty straightforward, mostly duplicating sections related to `and` in the interpreter, but there are two tricky things:
1. What's the best way to allow installing methods from top-level?
2. How do we parse `⋅` as a keyword operator in the interpreter so that `C⋅D` works?

The main applications for this is intersection product and of course dot product of vectors. If this goes well, I think replacing `@@` with `circ` and $\circ$ would be amazing!

mahrud:
This is of course related to #1069 for working with unicode characters. @pzinn have you tried something like this?

d-torrance:
Parsing without spaces doesn't work:

```m2
i4 : v·w

o4 = v·w

o4 : Symbol
```

But `locate` seems to work:

```m2
i6 : methods symbol ·

o6 = {0 => ((·, =), Type, Type)  }
     {1 => ((·, =), Thing, Thing)}
     {2 => (·, Thing, Thing)     }
     {3 => (·, Vector, Vector)   }

o6 : NumberedVerticalList

i7 : locate 3

o7 = stdio:1:20-1:60

o7 : FilePosition

i8 : code oo

o8 = stdio:1:20-1:60: --source code:
     Vector · Vector := (v, w) -> ((transpose v#0) * w#0)_(0, 0)
```

mahrud:
That's not what I meant, is this output the same if you replace + with cdot?
```m2
i2 : peek locate (pseudocode functionBody(() ->   1     +     1))

o2 = FilePosition{stdio, 2, 45, 2, 58, 2, 51}
```

[...]

PR #3584, opened 2024-11-18
Add interpunct (·) as a keyword for dot product, etc.

d-torrance:
For this to work, we no longer count unicode characters beginning with 194 as alphabetic (just like how we don't count characters beginning with 226 as alphabetic).  These are the "[Latin-1](https://en.wikipedia.org/wiki/Latin-1_Supplement) Punctuation and Symbols". 

So afterwards, we can use this new keyword to define things like dot products without having to worry about surrounding it with spaces:

```m2
i1 : Vector · Vector := (v, w) -> ((transpose v#0) * w#0)_(0, 0);

i2 : v = vector {1, 2, 3}; w = vector {4, 5, 6};

       3
o2 : ZZ

       3
o3 : ZZ

i4 : v·w

o4 = 32
```

Closes: #3434

### DIFF
--- a/M2/Macaulay2/d/ctype.d
+++ b/M2/Macaulay2/d/ctype.d
@@ -25,8 +25,7 @@ foreach c in " \t\r"	                   do setchartype(c,WHITE);
 foreach c in "\n"                          do setchartype(c,NEWLINE);
 foreach c in "$'"                          do setchartype(c,ALNUMEXTRA);
 
-for c from 128 to 225	       	    	   do setchartype(char(c),ALPHA);  -- 226 is unicode math symbols
-for c from 227 to 255	       	    	   do setchartype(char(c),ALPHA);
+for c from 128 to 255                      do setchartype(char(c),ALPHA);
 					      setchartype('\"',QUOTE);
 
 chartype(c:int):int := if (c & ~255) == 0 then int(chartypes.c) else 0;
@@ -52,10 +51,28 @@ export isspace    (c:char):bool := (chartype(c) & SPACE    ) != 0;
 export isnewline  (c:char):bool := (chartype(c) & NEWLINE  ) != 0;
 export isquote    (c:char):bool := (chartype(c) & QUOTE    ) != 0;
 
-export isvalidsymbol  (s:string):bool := (
-     if int(uchar(s.0)) == 226 && length(s) == 3 then return true; -- ugly unicode math symbol hack
+-- c = two bytes concatenated
+export ismathoperator(c:int):bool := (
+    (c & 0xffe0) == 0xc2a0 || -- latin-1 punctuation/symbols
+    c == 0xc397            || -- multiplication sign
+    c == 0xc3b7            || -- division sign
+    (c & 0xfffc) == 0xe288 || -- mathematical operators
+    (c & 0xfffc) == 0xe2a8 || -- supplemental mathematical operators
+    c == 0xe29f            || -- misc. mathematical symbols A
+    (c & 0xfffe) == 0xe2a6 || -- misc. mathematical symbols B
+    (c & 0xfffc) == 0xe28c    -- misc. technical
+    );
+
+ismathoperator(c1:char, c2:char):bool := (
+    ismathoperator((int(uchar(c1)) << 8) | int(uchar(c2))));
+
+export isvalidsymbol(s:string):bool := (
      if !isalpha(s.0) then return false;
-     foreach c in s do if !isalnum(c) then return false;
+     if ismathoperator(s.0, s.1) && length(s) == utf8charlength(s.0)
+     then return true;
+     for i from 0 to length(s) - 1 do (
+	 if !isalnum(s.i) || ismathoperator(s.i, s.(i + 1))
+	 then return false);
      true);
 
 -- Local Variables:
--- a/M2/Macaulay2/m2/latex.m2
+++ b/M2/Macaulay2/m2/latex.m2
@@ -126,7 +126,8 @@ keywordTexMath = applyKeys(hashTable { -- both unary and binary keywords
 	symbol ||  => "\\mid\\mid",
 	symbol ^** => "{}^{\\otimes}", -- temporary solution to KaTeX issue https://github.com/KaTeX/KaTeX/issues/3576
 	symbol _*  => "{}_*", -- temporary solution to KaTeX issue https://github.com/KaTeX/KaTeX/issues/3576
-	symbol ^*  => "{}^*" -- temporary solution to KaTeX issue https://github.com/KaTeX/KaTeX/issues/3576
+	symbol ^*  => "{}^*", -- temporary solution to KaTeX issue https://github.com/KaTeX/KaTeX/issues/3576
+	symbol ·   => "\\cdot",
 	},symbolBody)
 
 bbLetters := set characters "kABCDEFGHIJKLMNOPQRSTUVWXYZ"
@@ -141,7 +142,8 @@ texMathLiteralTable := merge(texLiteralTable,
 	"𝔩" => "\\mathfrak{l}","𝔪" => "\\mathfrak{m}","𝔫" => "\\mathfrak{n}","𝔬" => "\\mathfrak{o}","𝔭" => "\\mathfrak{p}","𝔮" => "\\mathfrak{q}","𝔯" => "\\mathfrak{r}","𝔰" => "\\mathfrak{s}","𝔱" => "\\mathfrak{t}","𝔲" => "\\mathfrak{u}","𝔳" => "\\mathfrak{v}",
 	"𝔴" => "\\mathfrak{w}","𝔵" => "\\mathfrak{x}","𝔶" => "\\mathfrak{y}","𝔷" => "\\mathfrak{z}","𝔄" => "\\mathfrak{A}","𝔅" => "\\mathfrak{B}","𝔆" => "\\mathfrak{C}","𝔇" => "\\mathfrak{D}","𝔈" => "\\mathfrak{E}","𝔉" => "\\mathfrak{F}","𝔊" => "\\mathfrak{G}",
 	"𝔋" => "\\mathfrak{H}","𝔌" => "\\mathfrak{I}","𝔍" => "\\mathfrak{J}","𝔎" => "\\mathfrak{K}","𝔏" => "\\mathfrak{L}","𝔐" => "\\mathfrak{M}","𝔑" => "\\mathfrak{N}","𝔒" => "\\mathfrak{O}","𝔓" => "\\mathfrak{P}","𝔔" => "\\mathfrak{Q}","𝔕" => "\\mathfrak{R}",
-	"𝔖" => "\\mathfrak{S}","𝔗" => "\\mathfrak{T}","𝔘" => "\\mathfrak{U}","𝔙" => "\\mathfrak{V}","𝔚" => "\\mathfrak{W}","𝔛" => "\\mathfrak{X}","𝔜" => "\\mathfrak{Y}","𝔝" => "\\mathfrak{Z}"
+	"𝔖" => "\\mathfrak{S}","𝔗" => "\\mathfrak{T}","𝔘" => "\\mathfrak{U}","𝔙" => "\\mathfrak{V}","𝔚" => "\\mathfrak{W}","𝔛" => "\\mathfrak{X}","𝔜" => "\\mathfrak{Y}","𝔝" => "\\mathfrak{Z}",
+	"×" => "\\times", "÷" => "\\div", "±" => "\\pm",
 	},last)
 texMathLiteral = texLiteral1 texMathLiteralTable
 -- TODO: expand and document this behavior
--- a/M2/Macaulay2/d/actors5.d
+++ b/M2/Macaulay2/d/actors5.d
@@ -1581,23 +1581,24 @@ fillnodes(n:LexNode):void := (
 fillnodes(baseLexNode);
 setupconst("operatorNames",Expr(operatorNames));
 
+createSymbol(w:Word, d:Dictionary, s:string):Expr := (
+    if !isvalidsymbol(s) then buildErrorPacket("invalid symbol")
+    else if d.Protected then (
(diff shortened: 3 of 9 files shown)


mahrud:
A couple of questions:
1. While you're doing this, could you also add exceptions for U+00D7 and U+00F7 for \times and \div?
2. I don't understand where 194 is coming from. Latin-1 punctuation and symbols starts at U+00A0 which is 160, no?

mahrud:
Also, this doesn't need to happen right now, but `texMath symbol ·` should give `\cdot` (and maybe same for a handful others that have tex names).

d-torrance:
> 1. While you're doing this, could you also add exceptions for U+00D7 and U+00F7 for \times and \div?

The problem with those two is that they're in with a bunch of characters that I think are definitely alphabetic:

```m2
i1 : apply(splice(150..160,180..190), x -> ascii {195, x})

o1 = (Ö, ×, Ø, Ù, Ú, Û, Ü, Ý, Þ, ß, à, ô, õ, ö, ÷, ø, ù, ú, û, ü, ý, þ)
```

I'm not sure which is more important -- allowing characters like à in symbol names or × and as ÷ as keywords.

> 2. I don't understand where 194 is coming from. Latin-1 punctuation and symbols starts at U+00A0 which is 160, no?

From what I understand from [Wikipedia](https://en.wikipedia.org/wiki/UTF-8#Description), since A0 is between 80 and 7FF  (and can be represented in 11 bits), it gets encoded in two bytes.  The first byte is 110 followed by the first 5 bits, and the second byte is 10 followed by the last 6 bits.  In this case, A0 -> 00010100000, and so the first byte is 11000010, or 194.

M2/Macaulay2/m2/latex.m2
```diff
@@ -126,7 +126,8 @@ keywordTexMath = applyKeys(hashTable { -- both unary and binary keywords
 	symbol ||  => "\\mid\\mid",
 	symbol ^** => "{}^{\\otimes}", -- temporary solution to KaTeX issue https://github.com/KaTeX/KaTeX/issues/3576
 	symbol _*  => "{}_*", -- temporary solution to KaTeX issue https://github.com/KaTeX/KaTeX/issues/3576
-	symbol ^*  => "{}^*" -- temporary solution to KaTeX issue https://github.com/KaTeX/KaTeX/issues/3576
+	symbol ^*  => "{}^*", -- temporary solution to KaTeX issue https://github.com/KaTeX/KaTeX/issues/3576
+	symbol ·   => "\\cdot"
```
mahrud, line 130: FYI trailing nulls are skipped by `hashTable` so you could put a comma after this.

mahrud:
> The problem with those two is that they're in with a bunch of characters that I think are definitely alphabetic:

Can't we single out those specific characters rather than the whole range?

pzinn:
tangentially related, while I was testing this:
```
i1 : getSymbol "⟎"

o1 = ⟎

o1 : Symbol

i2 : getSymbol "⟎⟎"
stdio:2:9:(3): error: invalid symbol

i3 : getGlobalSymbol "⟎"

o3 = ⟎

o3 : Symbol

i4 : getGlobalSymbol "⟎⟎"
stdio:4:15:(3): error: attempted to create symbol in protected dictionary
```
Clearly the second error message is incorrect. This can be traced back to `actors5.d`
```
getglobalsym(d:Dictionary,s:string):Expr := (
     w := makeUniqueWord(s,parseWORD);
     when lookup(w,d.symboltable) is x:Symbol do Expr(SymbolClosure(globalFrame,x))
     is null do (
          if !isvalidsymbol(s) then return buildErrorPacket("invalid symbol");
	  if d.Protected then return buildErrorPacket("attempted to create symbol in protected dictionary");
	  t := makeSymbol(w,tempPosition,d);
	  globalFrame.values.(t.frameindex)));

getglobalsym(s:string):Expr := (
     w := makeUniqueWord(s,parseWORD);
     when globalLookup(w)
     is x:Symbol do Expr(SymbolClosure(if x.thread then threadFrame else globalFrame,x))
     is null do (
	  if globalDictionary.Protected then return buildErrorPacket("attempted to create symbol in protected dictionary");
	  t := makeSymbol(w,tempPosition,globalDictionary);
	  globalFrame.values.(t.frameindex)));
```
the second def of getglobalsym is missing the line about invalid symbol.
but why can't the second definition just call the first with `globalDictionary`? the only difference is this thread thing which I don't fully understand.

pzinn:
also, I'm confused about `help getGlobalSymbol`:
```
If dict is omitted, then the first symbol found in the dictionaries listed in [dictionaryPath](http://localhost:8002/home/pzinn/M2/M2/BUILD/fedora/usr-dist/common/share/doc/Macaulay2/Macaulay2Doc/html/_dictionary__Path.html) will be returned. If none is found, one will be created in the first dictionary listed in dictionaryPath, unless it is not mutable, in which case an error will be signalled; perhaps that behavior should be changed.
```
first dictionary listed in `dictionaryPath`??? it's `Varieties.Dictionary`...

d-torrance:
> Can't we single out those specific characters rather than the whole range?

We definitely can!  I've pushed an updated version with support for this.

pzinn:
I haven't looked at the code in detail yet, but there are definitely differences in behaviour. For example, in the example above, `getSymbol "⟎⟎"` no longer causes an error message.

pzinn:
something's not right:
before
```
i1 : a⊗b
stdio:1:1:(3): error: no method for binary operator ** applied to objects:
            a (of class Symbol)
     **     b (of class Symbol)
```
after
```
i1 : a⊗b

o1 = a⊗b

o1 : Symbol
```

moorewf:
I have wanted to define shuffle products using (ш), so I am following this (and the previous) thread.

On a related note:  Is it possible to define subscripted operators in Macaulay2?  Something like ⊗_p or ш_p?

d-torrance:
Oops -- I definitely messed up mathematical operator characters inside symbols!  I'll take a look.

M2/Macaulay2/d/ctype.d
```diff
@@ -52,9 +51,24 @@ export isspace    (c:char):bool := (chartype(c) & SPACE    ) != 0;
...
-     if int(uchar(s.0)) == 226 && length(s) == 3 then return true; -- ugly unicode math symbol hack
+export ismathoperator(c1:int, c2:int):bool := (
+    c1 == 194      || -- latin-1 punctuation/symbols
+    c1 == 195 && (
+	c2 == 151  || -- \times
+	c2 == 183) || -- \div
+    c1 == 226);       -- mathematical operators
+
+-- number of bytes used to encode unicode character
+export numbytes(c:int):int := (
```
pzinn, line 62: note that there's already a function `utf8charlength` that does pretty much the same as your `numbytes`.

M2/Macaulay2/d/ctype.d
```diff
@@ -52,9 +51,24 @@ export isspace    (c:char):bool := (chartype(c) & SPACE    ) != 0;
...
+    if      c < 192 then 1
+    else if c < 224 then 2
+    else if c < 240 then 3
+    else                 4);
+
+export isvalidsymbol(s:string):bool := (
      if !isalpha(s.0) then return false;
+     if (ismathoperator(int(s.0), int(s.1)) &&
+	 length(s) == numbytes(int(s.0))) then return true;
```
pzinn, line 71: does this mean say `getSymbol "⟎⟎"` is not a valid symbol? this used to cause an error but no longer now.

mahrud:
> On a related note: Is it possible to define subscripted operators in Macaulay2? Something like ⊗_p or ш_p?

It's technically possible to define keywords `**_` or `⊗_` or `ш_`, but I think they would have to be a new kind of trenary operator with new parsing and evaluation logic to handle `A **_ B C`.

Also I should mention that `M **_R N` is defined as `tensor(M, f, N)` where M and N are modules and f is a ring map from R, but it's essentially evaluating `tensor(M, tensor(f, N))`.

M2/Macaulay2/d/ctype.d
d-torrance: Oh cool -- thanks!

M2/Macaulay2/d/ctype.d
```diff
@@ -52,9 +51,24 @@ export isspace    (c:char):bool := (chartype(c) & SPACE    ) != 0;
 export isnewline  (c:char):bool := (chartype(c) & NEWLINE  ) != 0;
 export isquote    (c:char):bool := (chartype(c) & QUOTE    ) != 0;
 
-export isvalidsymbol  (s:string):bool := (
-     if int(uchar(s.0)) == 226 && length(s) == 3 then return true; -- ugly unicode math symbol hack
+export ismathoperator(c1:int, c2:int):bool := (
```
pzinn, line 54: if you're going to discriminate further, then you should probably also reduce the range of `c1 == 226`, only a subset of which are actual mathematical symbols, cf https://en.wikipedia.org/wiki/Mathematical_operators_and_symbols_in_Unicode

mahrud: It's up to Doug but I would rather this PR be limited and uncomplicated. Your suggestion could happen in a future PR.

pzinn: actually, it's literally a one-line change, in line with what was done for `c1 == 194`

d-torrance:
This is in better shape now.  I've fixed it so that `getSymbol` should once again raise an error if its argument contains a mathematical operator characters as a proper substring.  We also check for some more mathematical operators, including ш.

M2/Macaulay2/d/ctype.d
```diff
@@ -52,10 +51,29 @@ export isspace    (c:char):bool := (chartype(c) & SPACE    ) != 0;
...
 export isquote    (c:char):bool := (chartype(c) & QUOTE    ) != 0;
 
-export isvalidsymbol  (s:string):bool := (
-     if int(uchar(s.0)) == 226 && length(s) == 3 then return true; -- ugly unicode math symbol hack
+-- c = two bytes concatenated
+export ismathoperator(c:int):bool := (
+    (c & 0xffe0) == 0xc2a0 || -- latin-1 punctuation/symbols
+    c == 0xc397            || -- multiplication sign
+    c == 0xc3b7            || -- division sign
+    c == 0xd188            || -- sha (shuffle product)
```
pzinn, line 59: Unfortunately 0xd188 (ш) is a ... Russian letter. you can't declare unilaterally that one letter in some alphabet is a mathematical symbol.
In any case at this stage it wouldn't do the user any good, since they can't define binary operators.
Once they can, they can name their opertaors ш, or привет, or any other sequence of characters they like.

pzinn: actually my comment is not entirely correct. without the line I singled out, they still won't be able to write `1ш2`. but I think that's just the way it is, as I said we can't prevent Russian users from naming their variables `шутник`...

d-torrance: I went ahead and removed ш for now.  We can cross that bridge later if we really want it as a keyword

moorewf:
How about this symbol?

https://decodeunicode.org/U+29E2

That doesn't seem to be the Cyrillic letter and comes from the Miscellaneous Mathematical Symbols block.

pzinn (approved):

d-torrance:
I suppose it's as good a time as any to make ⧢ a keyword.  @moorewf -- what should its precedence be? Maybe the same as `**`?

d-torrance:
> It's technically possible to define keywords `**_` or `⊗_` or `ш_`, but I think they would have to be a new kind of trenary operator with new parsing and evaluation logic to handle `A **_ B C`.

I think we might be able to implement this as a binary operator without making too many changes to the interpreter.  `A **_ B` could return a function that takes `C` as an argument.

mahrud:
Yes, I considered that, but it feels like a hack for the sake of syntactic sugar, and at least in the case of tensor I decided it's not worth it.

mahrud:
Could you also add ⊠ (\[boxtimes](https://en.wiktionary.org/wiki/%E2%8A%A0)) as a binary operator?

mahrud:
It would also be good to (eventually) add equivalent unicodes for all existing symbols, like ≥ as synonym for >=, etc.

pzinn:
already the case I believe:
```
i1 : symbol ≥ 

o1 = >=

o1 : Keyword
```

d-torrance:
I went ahead and added both `⊠` and `⧢` as keywords.  Same precedence as `**` -- does that seem ok?

pzinn:
I think this is getting out of hand. We can't add math symbols one at a time, there are thousands of them. We should let the user define them. (in some other PR, that is -- I understand there are complications with that)

mahrud:
We've only asked for like 5 total symbols in this thread because we have applications for them ...

pzinn:
sure but there are plenty more. For example, it would be nice to have `∈` as a binary-method synonym for `isMember`.

pzinn:
actually that one is fairly universal, so would make sense in this PR.

mahrud:
Hmm actually I think `∈` should be a synonym for `in` so you can write `for x ∈ L do ...`, though maybe we can also add new syntactic sugar `x in L` which calls `isMember`!

d-torrance:
Any further comments?  Is this ready to merge?

mahrud:
I'm happy with it.

pzinn:
a bit late to the party as usual, but there's still something not-quite-right with this PR: it's the changes in `latex.m2`.
why is `symbol ·   => "\\cdot` added to `keywordTexMath` whereas the other new symbols are added to `texMathLiteralTable`? to be fair this part of `latex.m2` is somewhat cryptic (I wrote it recently, so my fault). The immediate consequence is that `·` is not texified. But I do feel like the separation between keywords and symbols is stupid and needs rethinking.

pzinn:
I've just noticed belatedly that this PR breaks this:
```
i1 : 1⇒2
stdio:1:1:(3): warning: character '�' immediately following number
stdio:1:1:(3): error: no method for adjacent objects:
            1 (of class ZZ)
    SPACE   ⇒2 (of class Symbol)
```
vs before
```
i1 : 1⇒2

o1 = 1 => 2

o1 : Option
```
I might try to fix this in #3593

Issue #3570, opened 2024-11-01
Issue error when using multithreading with --no-threads


mahrud:
> Perhaps a clear error should be signaled whenever a thread-related function (e.g. `schedule`) is called but `--no-threads` is given.

_Originally posted by @mahrud in https://github.com/Macaulay2/M2/issues/3458#issuecomment-2332645824_

d-torrance:
Closed by #3587

Issue #1147, opened 2020-05-13
documentation for threads and tasks needs improving


mikestillman:
The documentation for the following should all refer to each other.
* Task
* "parallel programming with threads and tasks"
* schedule
* createTask
* addCancelTask
* addDependencyTask
* addStartTask
* cancelTask
* isCanceled (change name!?)
* taskResult
* isReady
* nanosleep
* allowableThreads
* maxAllowableThreads

It would also be useful to have a few examples of higher level use of these low level facilities.  I believe we used to have something like this, (perhaps even in the Goettingen workshop from a number of years ago?).

This is part of the effort to make task more usable in Macaulay2.

d-torrance:
> **Written by Claude** (Claude Opus 5, via Claude Code), not by @d-torrance, whose account posted it -- please weigh it accordingly.

On "I believe we used to have something like this" — there were examples, and
there is a record of what happened to them.

`bugs/dan/0-task-examples`, from the pre-GitHub `bugs/` tree, is Dan Grayson
removing them, with the diff. The `(addCancelTask,Task,Task)` node lost this:

```m2
allowableThreads = 3
u = schedule ( () -> while true do null )
sleep 1
u
t = createTask ( () -> null )
addCancelTask(t,u)
u
schedule t
while not isReady t do sleep 1
u
taskResult t
```

His note on why:

> I removed failing task examples from the documentation. They failed, for
> example, under ubuntu 12.04. Why?
>
> Part of it is probably the garbling of printed output, since separating
> examples into lines can get confused.
>
> Sometimes tasks run forever, using cpu time. But they're too simple!

The guess looks right. #2021 is the same failure ten years later on an armhf
build, diagnosed there as a timing race in example capture and settled by
lengthening a `sleep`.

The example itself works on 1.26.06-8-g34d5846039 — `u` reaches
`<<task, canceled>>`, `t` becomes ready, `taskResult t` is `null` — but it still
emits an interleaved backtrace from the cancelled thread:

```
error: interrupted
--back trace--
```

which lands in the captured output at a point that depends on scheduling. That
is the garbling, and it is presumably why these have stayed out.

So the node at `M2/Macaulay2/packages/Macaulay2Doc/ov_threads.m2:158` still has
no `Example` block, and restoring one means either suppressing that backtrace or
accepting nondeterministic example output.

[...]

PR #3587, opened 2024-11-21
Raise an error when creating a task but M2 was called w/ --no-threads

d-torrance:
### Before

We could create tasks, but they'd never be run:

```m2
$ M2 --silent --no-threads

i1 : t = schedule(() -> 3);

i2 : sleep 5

o2 = 0

i3 : isReady t

o3 = false
```

### After

```m2
$ M2 --silent --no-threads

i1 : t = schedule(() -> 3);
stdio:1:12:(3): error: thread support has been disabled
```

Closes: #3570 

---
I also noticed another bug while working on this.  Previously, if we call `taskResult` on a `Task` object that was created with `taskResult` but we never called `schedule`, then Macaulay2 would hang:

```m2
i1 : t = createTask(() -> 3);

i2 : taskResult t
^C^C
Exit (y=yes/n=no/a=abort/b=backtrace)? b
-* stack trace, pid: 3305389
 0# stack_trace(std::ostream&, bool) at /usr/src/macaulay2-1.24.05+git202410290107-0ppa202410281907~ubuntu24.04.1/M2/Macaulay2/d/main.cpp:136
 1# interrupt_handler at /usr/src/macaulay2-1.24.05+git202410290107-0ppa202410281907~ubuntu24.04.1/M2/Macaulay2/d/main.cpp:300
 2# 0x00007CA6F4A45320 in /lib/x86_64-linux-gnu/libc.so.6
 3# __GI___futex_abstimed_wait_cancelable64 at ./nptl/futex-internal.c:139
 4# pthread_cond_wait at ./nptl/pthread_cond_wait.c:627
 5# ThreadTask::waitOn() at /usr/src/macaulay2-1.24.05+git202410290107-0ppa202410281907~ubuntu24.04.1/M2/Macaulay2/system/supervisor.cpp:211
 6# taskResult_1 at /usr/src/macaulay2-1.24.05+git202410290107-0ppa202410281907~ubuntu24.04.1/M2/Macaulay2/d/pthread-tmp.c:465
```

Now we raise an error instead:

```m2
i1 : t = createTask(() -> 3);

i2 : taskResult t
stdio:2:10:(3): error: task not scheduled yet
```

### DIFF
--- a/M2/Macaulay2/d/pthread.d
+++ b/M2/Macaulay2/d/pthread.d
@@ -97,8 +97,12 @@ nextTaskSerialNumber():int := (
      taskSerialNumber = taskSerialNumber+1;		    -- race condition here, solve later
      taskSerialNumber);
 
+NoThreadsError() ::= buildErrorPacket("thread support has been disabled");
+
 createTask2(fun:Expr,arg:Expr):Expr :=(
      if !isFunction(fun) then return WrongArg(1,"a function");
+     if Ccode(bool, "!isThreadSupervisorInitialized()")
+     then return NoThreadsError();
      tc := TaskCell(TaskCellBody(nextHash(),nextTaskSerialNumber(),Ccode(taskPointer,"((void *)0)"), false, fun, arg, nullE ));
      blockingSIGINT();
      -- we are careful not to give the new thread the pointer tc, which we finalize:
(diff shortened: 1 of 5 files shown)


M2/Macaulay2/d/pthread.d
```diff
@@ -97,8 +98,12 @@ nextTaskSerialNumber():int := (
      taskSerialNumber = taskSerialNumber+1;		    -- race condition here, solve later
      taskSerialNumber);
 
+NoThreadsError() ::= buildErrorPacket("thread support has been disabled");
+
 createTask2(fun:Expr,arg:Expr):Expr :=(
```
mahrud, line 103: Why is this called `createTask2`?

mahrud:
It might be worth adding a couple of tests.

mahrud (approved):

M2/Macaulay2/d/pthread.d
d-torrance: I think because it takes two arguments?

d-torrance:
> It might be worth adding a couple of tests.

Definitely!  The thread documentation could certainly use some work -- see also #1147.

Issue #3507, opened 2024-10-03
fatal error: no member named 'Low' in 'rint<K>'


d-torrance:
The `cmake-macos-13-clang-15` builds have just started failing with the following compile error:

```c
[216/697] Building CXX object Macaulay2/e/CMakeFiles/M2-engine.dir/engine.cpp.o
FAILED: Macaulay2/e/CMakeFiles/M2-engine.dir/engine.cpp.o 
[...]

PR #3594, opened 2024-11-26
Update givaro submodule w/ fix for GCC 15 bug

d-torrance:
The givaro developers have merged https://github.com/linbox-team/givaro/pull/233, which fixes the "no member named 'Low' in 'rint<K>'" compilation failure we were getting with GCC 15.

Closes: #3507


d-torrance:
The givaro devs just tagged a new release candidate (4.2.1rc0), so I've updated this PR to use that commit instead.

mikestillman (approved):
This one looks ready to go.  @mahrud, do you think so too?

@d-torrance Do we need to change anything else?  Is this working on both cmake and autotools?

d-torrance:
Yes, it should be good.  In particular, the build now works when we build givaro ourselves with GCC 15.

d-torrance:
Givaro just released v4.2.1, so I've updated the submodule commit again to reflect that.

I added a temporary commit that should build givaro (and ffpack) and use it in the GitHub builds to double check that everything's still working.

d-torrance:
The builds all passed (with the exception of the autotools Ubuntu builds, but that's because of an autotools bug I should fix in another PR), so I'm going to remove the last commit.

PR #3595, opened 2024-11-27
Define gcd & lcm of single element

d-torrance:
Previously, `gcd` and `lcm` raised an error if we passed a single element.  This seems a bit odd, especially when we use parentheses, since it works with any other number of arguments:

```m2
i1 : gcd()

o1 = 0

i2 : gcd(3)
stdio:2:3:(3): error: no method found for applying gcd to:
     argument   :  3 (of class ZZ)

i3 : gcd(3, 6)

o3 = 3

i4 : gcd(3, 6, 9)

o4 = 3
```

After defining `gcd(ZZ)`:

```m2
i1 : gcd(3)

o1 = 3
```

### DIFF
--- a/M2/Macaulay2/m2/integers.m2
+++ b/M2/Macaulay2/m2/integers.m2
@@ -51,6 +51,7 @@ gcd(QQ,ZZ) := QQ => (y,x) -> gcd(x * denominator y, numerator y) / denominator y
 gcd(QQ,QQ) := QQ => (x,y) -> (
      d := denominator x * (denominator y // gcd(denominator x, denominator y));
      gcd(numerator (x * d), numerator (y * d)) / d)
+gcd ZZ := gcd QQ := identity
 
 abs = method()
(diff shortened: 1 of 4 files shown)


M2/Macaulay2/m2/integers.m2
```diff
@@ -68,6 +69,7 @@ lcm(QQ,QQ) := (f,g) -> (
     d := gcd(f, g);
     if d == 0 then 0_QQ
     else abs f * (abs g / d))
+lcm ZZ := lcm QQ := identity
```
mahrud, line 72: Number, perhaps?

d-torrance: I thought about that.  I stuck with `ZZ` and `QQ` since those are the ones we have binary methods for.  It seems like it might be a bit strange if, say, `gcd(3.0)` worked but `gcd(3.0, 6.0)` didn't.

pzinn (approved):

PR #3597, opened 2024-11-27
gcdCoefficents improvements

d-torrance:
A few related changes:

* GMP has a built-in extended GCD function, so we use it instead of the current one that's implemented at top-level, improving performance:

   ### Before
   ```m2
   i1 : elapsedTime gcdCoefficients(46, 240)
    -- .000109342s elapsed

   o1 = {2, 47, -9}

   o1 : List
   ```

   ### After
   ```m2
   i1 : elapsedTime gcdCoefficients(46, 240)
    -- .00001076s elapsed

   o1 = {2, 47, -9}

   o1 : List
   ```
   
* `gcdCoefficients(ZZ, RingElement)` and `gcdCoefficients(RingElement, ZZ)` have now been defined, promoting the integer to appropriate ring just like `gcd`.

* The documentation has been updated (converting to SimpleDoc, adding the new methods, and adding some examples).

### DIFF
--- a/M2/Macaulay2/packages/Macaulay2Doc/doc_arithmetic.m2
+++ b/M2/Macaulay2/packages/Macaulay2Doc/doc_arithmetic.m2
@@ -72,15 +72,42 @@ document {
 	  "conjugate 3"
 	  }
      }
-document {
-     Key => {gcdCoefficients,(gcdCoefficients, RingElement, RingElement),(gcdCoefficients, ZZ, ZZ)},
-     Headline => "gcd with coefficients",
-     TT "gcdCoefficients(a,b)", " -- returns ", TT "{d,r,s}", " so that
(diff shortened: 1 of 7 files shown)


M2/Macaulay2/packages/Macaulay2Doc/doc_arithmetic.m2
pzinn, line 1: I think the name of French mathematician Bézout needs to appear in this doc :)

d-torrance: Sounds good!  Updated

d-torrance:
I feel like this method should return a sequence instead of a list so that the caller can use parallel assignment, but I'm hesitant to make a breaking change.

pzinn:
it's only used 3 times in packages, AFAICT, so it's worth a shot.

pzinn (approved):

Issue #3593, opened 2024-11-26
User-created keywords


pzinn:
Following #3584, compare
```
i1 : symbol ×

o1 = ×

o1 : Symbol

i2 : symbol ⧢

o2 = ⧢

o2 : Keyword
```
Why the difference? because an arbitrary list of symbols has been declared as a keyword (something that I'd rather call a binary operator, but sadly the word `Binary` is used for something completely different in M2. I digress.), i.e., something that's used like this:
```
i3 : 1 ⧢ 2
stdio:3:2:(3): error: no method for binary operator ⧢ applied to objects:
            1 (of class ZZ)
      ⧢     2 (of class ZZ)
```
But realistically most users are never going to use `⧢` so why this symbol rather than another?
I think it'd make much more sense for users/packages to be able to define their own Keywords. Then the Core should stick to a minimal set of Keywords.
Incidentally note that `new Keyword from ×` will not produce an error (though currently, it probably should, since the result is not what you would expect).

pzinn:
I intend to rewrite `binding.d` and `actors5.d` which have become a mess, due to the overabundance of keywords/symbols.

PR #3599, opened 2024-11-28
1st attempt: m2 level makeKeyword

pzinn:
This is very much work in progress, just to demonstrate how to have user-defined keywords: (cf #3593)
```
i1 : makeKeyword("×")

o1 = ×

o1 : Keyword

i2 : ZZ × ZZ := times

o2 = times

o2 : CompiledFunction

i3 : 1×2

o3 = 2
```
The most important change is to have a default operation (which applies to 99% of existing keywords, and to the newly created user-defined keywords) which is to look up the corresponding method. so no need to define all these identical functions one-by-one.
Right now only binary operations are properly created, I'm just doing a PR for the tests.
TODO
- simplify drastically actors5.d by removing all unnecessary functions
- transfer the creation of keywords to m2 level (core)
- get rid of `opsWithBinaryMethod` etc
- get rid of `getBinopName` etc and rewrite `pseudocode` accordingly


pzinn:
I might stop here, to keep this PR relatively small and to avoid the more controversial items on my list.

M2/Macaulay2/d/actors5.d
```diff
@@ -13,26 +14,44 @@ getParsing(e:Expr):Expr := (
...
+	   uprec:=nopr;
+	   bprec:=nopr;
+	   if b1.v || b2.v then (
+	   if b1.v then (t=binaryop; bprec=prec);
+	   if b2.v then (u=unaryop; uprec=prec);
+	   ) else t=postfixop;
+	   w:=makeUniqueWord(s.v, parseinfo(prec,bprec,uprec,parsefuns(u,t)));
+	   when globalLookup(w) is x:Symbol do buildErrorPacket("symbol already in use")
+	   else (
+	   install(s.v,w);  -- TODO check whether install is really needed (for mathematical symbols as opposed to words)
```
d-torrance, line 44: Are there any updates on whether we need to call `install` here?

M2/Macaulay2/d/binding.d
```diff
@@ -190,7 +190,7 @@ special(s:string,f:function(Token,TokenFile,int,bool):ParseTree,lprec:int,rprec:
 -- created below are all in the global dictionary.
 
 -- new operators must be:
---   set up as an "actor" with "setup()"
+--   set up as an "actor" with "setup()" -- or not TODO explain
```
d-torrance, line 193: Are you planning on adding this explanation in this PR?

M2/Macaulay2/d/actors5.d
pzinn: yeah this is rather subtle. my understanding is, it's harmless to always have `install` here (except maybe for a tiny slowdown of the parsing?). The point is, this is useless for keywords that are actual words (like `and`) because you need to separate them with spaces anyway (so they behave more similarly to regular symbols). But at the moment I've put limited constraints on allowed keywords (see `isvalidkeyword` -- which may require more restrictions), so people can have keywords like `@*@` for which `install` is definitely necessary.

pzinn: I rephrased slightly the comment, but haven't actually changed anything

pzinn: I need to think about this some more. Relatedly, at this stage nothing would prevent us from removing the section in `lex.d` which you wrote `else if ismathoperator(peek2(file)) then (` etc except maybe unicode synonyms.

pzinn: answering myself: no that doesn't work. math operators now have a slightly ambiguous status of being both "alphanumeric" (i.e., usable in a regular symbol) and have special parsing rules (so they can be used in keywords, e.g., binary operators).

pzinn: it would be so much easier if we could declare math symbols to be non alphanumeric. In fact this was my original idea in my first PR on unicode symbols, but that breaks an existing package which defined the sum and product unicode symbols, so I had to change it. now we're paying the price for this backwards compatibility.

mahrud: What package is this?

d-torrance: I think it's MultiprojectiveVarieties, which defines `∏`, `⋂`, and `⋃` methods.  I think they'd probably still work as unary operators created using `makeKeyword` after this PR, right?

pzinn: well, right now, this PR doesn't change the status of math operators, so everything works fine as written.
but yes, as you pointed out, a nice feature of this PR is one can even define unary (prefix) operators which for all practical purposes work the same as a `MethodFunction`. so there would definitely be one way forward. might want to have a separate PR for this whole mess. In the meantime, I will fix the `install` thing today so we can resolve this particular comment.

M2/Macaulay2/m2/methods.m2
```diff
@@ -732,6 +732,22 @@ registerFinalizer' = registerFinalizer
...
+protect Precedence
+protect Syntax
+makeKeyword'=makeKeyword
+makeKeyword=method(TypicalValue => Keyword, Options => { Precedence => symbol *, Syntax => Binary })
+makeKeyword String := o -> s -> (
+    pr:=if instance(o.Precedence,Symbol) then (getParsing o.Precedence)#0 else o.Precedence;
+    makeKeyword'(s,
+    pr,
+    o.Syntax === Binary or (class o.Syntax === List and isMember(Binary,o.Syntax)),
+    o.Syntax === Prefix or (class o.Syntax === List and isMember(Prefix,o.Syntax))
```
d-torrance, line 748: Should there more checking of valid inputs for the `Syntax` option?   For example, currently we have:

```m2
i1 : makeKeyword("⇗", Syntax => foo)

o1 = ⇗

o1 : Keyword
```

But shouldn't it complain that `foo` is not one of the expected options?

M2/Macaulay2/d/actors5.d
mahrud: > it would be so much easier if we could declare math symbols to be non alphanumeric. In fact this was my original idea in my first PR on unicode symbols, but that breaks an existing package which defined the sum and product unicode symbols, so I had to change it. now we're paying the price for this backwards compatibility.

I don't follow this then. How does declaring math symbols to be non-alphanumeric break backwards compatibility with a package that uses non-alphanumeric method names? Unless you meant declare any non-alphanumeric string to be a math operator automatically?

mahrud: I agree that we should aim for the ability to declare different types of operators, binary, unary left/right, with prescribed precedents.

pzinn: > > it would be so much easier if we could declare math symbols to be non alphanumeric. In fact this was my original idea in my first PR on unicode symbols, but that breaks an existing package which defined the sum and product unicode symbols, so I had to change it. now we're paying the price for this backwards compatibility.
> 
> I don't follow this then. How does declaring math symbols to be non-alphanumeric break backwards compatibility with a package that uses non-alphanumeric method names? 
you mean, with a package that uses math-operator method names?
because methods have ordinary symbols (as opposed to keywords), and these are required to be alphanumeric.

mahrud:
Since lots of things are being refactored here, what do you think about also changing the terminology? My understanding is that the classes `Symbol` and `Keyword` are identical, except that keywords are symbols which are reserved by the language and can't be assigned values, etc. This means things like `and, then`, but also operators like `+`, etc. But with this PR users can define _new_ keywords, which implies they are no longer reserved.

So I think we should create a new class `Operator` for only non-alphanumeric math operators which are distinct from reserved alphanumeric keywords.

pzinn:
sure.

pzinn:
though it would be nice to have a common ancestor, since I don't want to have two different methods `makeKeyword` and `makeOperator` which would do the exact same thing.

mahrud:
I guess what I'm saying is we should _not_ have `makeKeyword` at all, since by definition keywords are the preserved symbols.

M2/Macaulay2/m2/expressions.m2
```diff
@@ -1250,15 +1250,10 @@ net Set := net @@ expression
...
-ldots=new Dots from symbol ldots
-texMath Dots := x -> "\\" | simpleToString x -- note that \vdots has bad spacing in ordinary LaTeX
-toString Dots := x -> "..."
-net Dots := x -> if x === vdots then "."||"."||"." else if x === ddots then ".  "||" . "||"  ." else "..."
-
+⋯:=symbol ⋯
+⋱:=symbol ⋱
+⋮:=symbol ⋮
+…:=symbol …
 -- used e.g. in chaincomplexes.m2
```
mahrud, line 1257: I'm confused about what's happening here, and also don't think this particular syntax is good practice.

M2/Macaulay2/m2/set.m2
```diff
@@ -183,9 +183,6 @@ fixedOperators = unique toList fixedOperators
 allOperators = unique toList allOperators
 
 protect Flexible
-protect Binary
-protect Prefix
-protect Postfix
```
mahrud, line 188: How come Flexible is kept here?

M2/Macaulay2/m2/expressions.m2
mahrud: For instance, maybe `cdots = ⋯ = new Operator from symbol ⋯` is preferable, and later in this file it would be better to use the synonym `cdots` rather than `⋯`.

M2/Macaulay2/m2/set.m2
mahrud: Tangentially, I think a syntax like:
```m2
otimes = ⊗ = new BinaryOperator from symbol ⊗
otimes.precedence = precedence symbol *
```
would be nice, though perhaps for flexible operators it's a bit more tricky.

pzinn:
oh yeah this PR also fixes this silly bug: `getGlobalSymbol ""`

> I guess what I'm saying is we should _not_ have `makeKeyword` at all, since by definition keywords are the preserved symbols.
> 
> edit: to clarify, a `makeKeyword` function to simplify declaring alphanumeric only reserved symbols (e.g. `and, or` etc.) in the interpreter is totally fine, I'm saying there shouldn't be a top level one. On the other hand, a `makeOperator` function to simplify things in the interpreter and also exporting it to allow users to define new operators is fine.

err? that's exactly what this PR is doing... Example
```
makeKeyword "implies"
Boolean implies Boolean := (a,b) -> not a or b
false implies true
```

M2/Macaulay2/m2/set.m2
pzinn: dunno I didn't touch this

M2/Macaulay2/m2/expressions.m2
pzinn: well, what was there before was worse, since it defined a sub-Type of `Symbol`, which is not a goo idea in general. this has the merit that it works.

M2/Macaulay2/m2/methods.m2
pzinn: done

M2/Macaulay2/m2/set.m2
pzinn: the problem with using the syntax `new ... from symbol ...` is that there is no room for options, and I'm not sure how to implement this `.precedence` thing.

pzinn: > How come Flexible is kept here?
dunno I didn't touch this

pzinn:
oh yeah this PR also fixes this silly bug: `getGlobalSymbol ""`

M2/Macaulay2/m2/set.m2
pzinn: should I have moved `Flexible` too? I'm not even clear what it means.

M2/Macaulay2/m2/expressions.m2
pzinn: but OK with synonyms, will do

mahrud:
If a user can define `implies` as in your example, then it is by definition not a _reserved_ symbol, therefore it shouldn't be called a _keyword_. It's just a new type of method attached to a symbol, which is very useful, but is different.

M2/Macaulay2/m2/set.m2
mahrud: > dunno I didn't touch this

You removed `protect Binary` and `Prefix` and `Postfix` but kept `Flexible` here. Why?

pzinn:
sure, I'm happy to name it something else. but the point is, for all practical purposes, in that example, `and` and `implies` behave in the exact same way, and therefore should probably have the same class.

M2/Macaulay2/m2/set.m2
pzinn: well, besides what I wrote above, the practical reason was that I didn't need `Flexible`.

mahrud:
> for all practical purposes, in that example, `and` and `implies` behave in the exact same way, and therefore should probably have the same class.

Then users should not be able to define them in the top level. They should be defined in the interpreter, compiled, and fixed in the language.

pzinn:
> > for all practical purposes, in that example, `and` and `implies` behave in the exact same way, and therefore should probably have the same class.
> 
> Then users should not be able to define them in the top level. They should be defined in the interpreter, compiled, and fixed in the language.

well, I disagree. It certainly doesn't harm to give users that power.

mahrud:
It certainly does harm if it's hastily implemented without much longer discussion and evaluation. What happens if a package makes `foo` into a keyword and another makes it a symbol for a method or variable? What happens if there's a race condition in which one is defined first?

Again, I'm all in favor of allowing users to define non-alphanumeric math operators in the top level, because those can't be used as symbols for methods or variables, and we already know how to deal with different packages overriding the same method.

pzinn:
actually, it makes no difference whether the keyword is alphanumeric or not. there will be a conflict if two packages define the same keyword. Right now it's a non issue because no package does so, but eventually we need a mechanism to resolve if say two different packages want to use `×` for different purposes. (The biggest problem I can see is, we can't reconcile them if they have different `Syntax` or `Precedence`)

pzinn:
To clarify, `Symbol`s can be in different dictionaries, but the underlying `Word`s, which carry the parsing information, cannot (and shouldn't precisely because they affect parsing).

pzinn:
Let me try to summarise some of the changes that have been requested during the M2internals meeting:
- not let users define *alphanumeric* keywords. though I don't fully agree with this, allowing it does open a can of worms (as the `hi` example showed) and maybe for now it's simpler to prevent this.
- if we do that, that definitely forces another change that was not mentioned: currently math symbols are considered as alphanumeric. This is a hack I introduced a while ago due to the backwards compatibility issue mentioned above in the discussion. This will have to go, otherwise we won't be able to define any math symbol as a keyword.
- have more meaningful type names. `Keyword` is not really an appropriate name for symbols like `⊗`. One could create a new type called `Operator` (which, just like `Keyword`, would be a fake type anyway... they're all `Symbol`s internally).
- Even for these operators, there'd still be an issue if two different packages try to define the same say cow symbol. There needs to be a way of dealing with this, and this remains the most problematic issue. One could make the symbols local (right now all `Keyword` are really global symbols), but that wouldn't really the solve the parsing issue -- if for some reason one package wants to use a symbol as unary and the other as binary, it will be hard to reconcile.
- Only have makeKeyword in core, not packages: this one I strongly disagree with, and will not implement. Users are not babies. If they want they can already easily break M2 in all kinds of ways (but why would they?). Keywords (or operators, or whatever they will be called) should be definable in packages.

mahrud:
Limiting `makeKeyword` to the interpreter or the Core was mainly a solution to the problematic issue that you mentioned. Ultimately this PR introduces some useful features which can be added now, but jumping directly from idea to production for major changes is a recipe for ending up with compatibility nightmares and packages that only work with a specific version of M2 or indecipherable issues that take valuable time to resolve.

Perhaps a good place to start is a wiki page where you lay out a proposal (similar to [Python Enhancement Proposals](https://peps.python.org/pep-0001/)) with concise technical specifications of how you believe keywords/operators/etc should be handled, from parsing and binding in the interpreter all the way to the packages. This can simultaneously serve as a documentation of your contribution to this part of the interpreter. This can be discussed and amended, until there's agreement.

M2/Macaulay2/m2/expressions.m2
pzinn: for the record, undid these changes to expressions.m2

pzinn:
the checks were successful before, I don't think the ThreadedGB stuff has anything to do with this PR.

mahrud:
I agree, but I restarted that job anyway. My preference would be to merge this PR right after the release to give some time for daily testing.

mahrud:
@pzinn @d-torrance once the conflict is resolves, I think this can be merged?

mikestillman:
@pzinn and @mahrud , what is the status of this PR?  Are you both happy with it now, or do we need to look at it closer?

mahrud:
I'm waiting for @pzinn to resolve the conflcts before merging it.

mahrud:
@pzinn do you want someone else to fix the conflict in lex.d?

mahrud:
I'm having a little trouble using this, wondering what I'm doing wrong:
```m2
i1 : needsPackage "A"

o1 = A

o1 : Package

i2 : 2 ⊙  2
stdio:2:2:(3): error: undefined symbol ⊙

i7 : get A#"source file"

o7 = newPackage "A"

     new Keyword from "⊙"
     ZZ ⊙ ZZ := times
```

But if I paste the same directly in M2 it works.

mahrud:
Interesting, this also fails:
```m2
newPackage "A"

export { "⊙" }

new Keyword from "⊙"

ZZ ⊙ ZZ := times
```
With the error:
```m2
i1 : needsPackage "A"
../../Macaulay2/packages/A.m2:3:6:(2):[11]: error: invalid symbol
```
But this one works:
```m2
newPackage "B"

new Keyword from "⊙"

export { "⊙" }

ZZ ⊙ ZZ := times
```

M2/Macaulay2/packages/CpMackeyFunctors/Code/BoxProduct.m2
```diff
@@ -106,6 +106,7 @@ boxProduct (MackeyFunctorHomomorphism, CpMackeyFunctor) := MackeyFunctorHomomorp
 CpMackeyFunctor ** CpMackeyFunctor := boxProduct
 CpMackeyFunctor ** MackeyFunctorHomomorphism := boxProduct
 MackeyFunctorHomomorphism ** CpMackeyFunctor := boxProduct
+new Keyword from "⊠"
```
mahrud, line 109: It appears that `⊠` has stopped working, unless the package is debugged.

mahrud (changes requested):
All unicode keywords defined by package authors are now broken. I suggest the following:
- move the newly added stuff in `m2/methods.m2` to `m2/operators.m2` or similar
- move all calls to `new Keyword from String` to `m2/operators.m2` (i.e. not from packages)
- move back all unicode keyword exports to `m2/exports.m2` and document them in Macaulay2Doc.

I think this is better than individual packages exporting these symbols, because for instance I would like to use ⊠ in my own package, but I don't want to depend on the CpMackeyFunctors package for it, and I also don't want the two packages not to work together.

M2/Macaulay2/packages/CpMackeyFunctors/Code/BoxProduct.m2
pzinn: probably forgot to export it, that's all

pzinn:
I don't understand any of the comments. seems to be working as intended. you always need to define a keyword before using it.
> I think this is better than individual packages exporting these symbols, because for instance I would like to use ⊠ in my own package, but I don't want to depend on the CpMackeyFunctors package for it, and I also don't want the two packages not to work together.

You obviously don't need to load CpMackeyFunctors to use ⊠, you just define it in your own package, that's the whole point. Just like for ordinary symbols, you will get an annoying warning message when both are loaded. There's not much I can do about that, it's how Macaulay2 works.

pzinn:
> I was about to merge this, but I realized that all unicode keywords that used to be exported by Core and overloaded by packages are now broken. I suggest the following:

what exactly is broken? except the minor `export` oversight which I can fix.

pzinn:
The only reasonable alternative I can think of is that mathematical symbols are "auto-defined", in the same way as ordinary symbols: the first time they appear, a keyword is created. The only issue with this is that it means they can no longer be used as ordinary symbols, e.g., ⊠ can only be a `Keyword`, not a `Symbol`.

M2/Macaulay2/packages/CpMackeyFunctors/Code/BoxProduct.m2
mahrud: It used to be exported from Core.

mahrud:
> Just like for ordinary symbols, you will get an annoying warning message when both are loaded. There's not much I can do about that, it's how Macaulay2 works.

Defining common symbols in Core is the whole point of `m2/shared.m2`.

mahrud:
> I don't understand any of the comments. seems to be working as intended. you always need to define a keyword before using it.

I provided full examples, compare package A and package B [here](https://github.com/Macaulay2/M2/pull/3599#issuecomment-3095312325). Currently, the standard is that all exported symbols are declared in the beginning of the package, before they are defined. But with keywords it seems they need to be exported at the end, or all keywords need to be declared defined in the beginning before the call to `export`.

pzinn:
that is correct. keywords need to be defined before they can be exported. the problem is that `export` defines the symbol at the same time. I've already raised this issue at some point, I think `export` should really be at the end even for ordinary symbols, this way the creation of the symbol would be correctly documented where it is first used, not at the level of the `export` statement, but there's not much I can do at this stage.

mahrud:
> I think `export` should really be at the end even for ordinary symbols, this way the creation of the symbol would be correctly documented where it is first used, not at the level of the `export` statement

I don't understand this part. Try `locate symbol Proj` and the location of `Proj = ...` comes up, not anything related to export.

> the problem is that `export` defines the symbol at the same time.

Are you referring to `getGlobalSymbol(pd, nam)` in `export`? Can `new Keyword` overwrite it?

pzinn:
oh yeah I fixed the location thing a while back (now `export` creates the symbol but leaves the location undetermined).
and no, you can't overwrite the creation of a symbol. so the moment you write say `export "×"` the symbol will be created.
currently this symbol will be an ordinary `Symbol` and nothing you do after that will change that.
as I said an alternative would be to impose that any mathematical symbol is a `Keyword` and then `export` could take that into account. I'm thinking this might be the way to go.

M2/Macaulay2/d/convertr.d
```diff
@@ -189,7 +192,12 @@ export convert0(e:ParseTree):Code := (
...
 		-- TODO: is this check necessary?
 		if token.word.typecode == TCid
-		then Code(binaryCode(b.Operator.entry.binary, convert(b.lhs), rhs, pos))
+		then (
+		  f:=b.Operator.entry.binary; if f==dummyBinaryFun then
+		  Code(ternaryCode(binarymethod1,convert(b.lhs), rhs, globalSymbolClosureCode(b.Operator.entry, dummyPosition),pos))
+		  else
+		  Code(binaryCode(f, convert(b.lhs), rhs, pos))
+		)
 		else dummyCode -- should not occur
```
mahrud, line 201: Could you elaborate on the changes in this section?

PR #3602, opened 2024-11-28
changed interiorVector to work for cones with no rays, and added a test

AlheydisGeiger:
At the Macaulay2 in the Sciences we observed that the interiorVector function in the Polyhedral package does not give an point inside the cone for a cone that has no rays.


lkastner (approved):
Thank you.

PR #3606, opened 2024-12-04
Add Vector == ZZ method

d-torrance:
Allows calling "zero" to test whether a vector is the zero vector.

Also add a unit test.

### Before

```m2
i1 : zero vector {0, 0, 0}
stdio:1:4:(3): error: no method for binary operator == applied to objects:
                          3
            0 (of class ZZ )
     ==     0 (of class ZZ)
```

### After
```m2
i1 : zero vector {0, 0, 0}

o1 = true
```


M2/Macaulay2/m2/matrix.m2
```diff
@@ -119,6 +119,7 @@ Matrix == Matrix := (f,g) -> (
 Matrix == Number :=
 Matrix == RingElement := (m,f) -> m - f == 0		    -- slow!
 Matrix == ZZ := (m,i) -> if i === 0 then rawIsZero m.RawMatrix else m - i == 0
+Vector == ZZ := (v,i) -> matrix v == i
```
mahrud, line 122: 1. Did you mean to also add `ZZ == Vector`? This might be a simple way to do both:
```suggestion
Vector == ZZ := ZZ == Vector := (u, v) -> matrix u == matrix v
```
2. What you wrote also makes `vector {1} == 1` to be true, which is probably fine, but just making sure this is intended.
3. Is this automatically caught in a generic documentation node?

d-torrance: 1.  I didn't think to, but that's a good idea!  I just pushed a new version adding this, and also supporting comparison using `==` with `Number` and `RingElement`, matching the corresponding `Matrix` methods.
2.  Yes, that's intended.
3. No, but I think that's a wider issue with all the `==` methods.  None of them seem to be documented!

mahrud: While you're at it, maybe also define `Vector == Matrix`?

I don't particularly care about all of these nodes being documented, as long as it's obvious what they do. I do care about them not "missing documentation" though, so either marking as undocumented or globbing all missing nodes involving `==` seem like fine solutions to me. This would still allow documenting the ones that are noteworthy, like for modules/ideals/sheaves.

M2/Macaulay2/m2/matrix.m2
```diff
@@ -116,9 +118,12 @@ Matrix == Matrix := (f,g) -> (
            raw(super f * inducedMap(source f, source g)) === raw super g
     )
 
-Matrix == Number :=
-Matrix == RingElement := (m,f) -> m - f == 0		    -- slow!
+Matrix == Number      :=
+Matrix == RingElement :=
+Vector == Number      :=
+Vector == RingElement := (m,f) -> m - f == 0                -- slow!
```
mahrud, line 124: Does this need to be written this way? This is slow due to #3035. For instance try `v == 1` with `v` the matrix defined in that issue.

M2/Macaulay2/packages/Macaulay2Doc/doc.m2
mahrud, line 1: Instead of a long list of methods like this, which requires constant maintenance, I'd rather you glob them all like in 7bfcb62fe07d0e63965d88a8d6b798d0a9adb78b.

M2/Macaulay2/m2/matrix.m2
mahrud: Also, `Vector == Number` covers `Vector == ZZ`, no?

d-torrance: Do you have a suggestion for a better way to define `Matrix == Number` and friends?   That's probably outside the scope of this PR.

Regarding `Vector == ZZ`, we currently need it to avoid a recursion error.  Otherwise, something like `v == 0` would call `v - 0 == 0`.

M2/Macaulay2/m2/matrix.m2
d-torrance: Yeah, `==` is really bad for missing documentation nodes!  https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/__eq_eq.html

Some sort of automatic way for adding otherwise undocumented methods for common operators like `+`, `-`, and `==` to the main node for that operator sounds like great idea.  But I think it's outside the scope of this PR.

M2/Macaulay2/m2/matrix.m2
mahrud: I guess I'm wondering what's wrong with `m == matrix f`? The slight edge case is when the target of m has relations and you want to reduce f first, but maybe the user should explicitly write `m - f * id_(target m)` in that case.

M2/Macaulay2/m2/matrix.m2
mahrud: Okay, then maybe don't touch the documentation at all.

d-torrance: I figured I'd at least document the `+` and `-` methods that I added.  The spacing between all the various keys for these nodes were super messy and inconsistent, so I decided I might as well clean it up a bit.

M2/Macaulay2/m2/matrix.m2
```diff
@@ -127,6 +134,8 @@ Matrix + RingElement := (f,r) -> if r == 0 then f else f + r*id_(target f)
 RingElement + Matrix := (r,f) -> if r == 0 then f else r*id_(target f) + f
 Number + Matrix := (i,f) -> if i === 0 then f else i*id_(target f) + f
 Matrix + Number := (f,i) -> if i === 0 then f else f + i*id_(target f)
+Vector + Number := Vector + RingElement := (v,r) -> vector(matrix v + r)
+Number + Vector := RingElement + Vector := (r,v) -> vector(r + matrix v)
```
mahrud, line 138:
```m2
i1 : Number + Vector := RingElement + Vector := (r,v) -> vector(r + matrix v);

i2 : 1 + (coker matrix {{ 10 }})_0
stdio:1:58:(3):[1]: error: expected source to be free with rank 1
```

M2/Macaulay2/m2/matrix.m2
mahrud: Sure, but this doesn't apply to vectors.

M2/Macaulay2/m2/matrix.m2
mahrud: If you want to clean up the messy documentation nodes then use globbing, otherwise why touch a hundred lines just to fix the indentation?

M2/Macaulay2/m2/matrix.m2
d-torrance: Calling `matrix` on the `Number`/`RingElement` won't work for the zero vector if the rank of the module is greater than one, since then we'll be comparing matrices with different sizes:

```m2
i1 : matrix vector {0, 0} == matrix 0

o1 = false
```
I did change it to just `matrix v == f`, which works just fine:

```m2
i2 : matrix vector {0, 0} == 0

o2 = true
```
Then we don't have to worry about defining `Vector == ZZ`.

M2/Macaulay2/m2/matrix.m2
d-torrance: Fair enough.  I removed that commit and just added the new keys for now.

M2/Macaulay2/m2/matrix.m2
d-torrance: This is coming from `vector(Matrix)`:

```m2
i1 : code (vector, Matrix)

o1 = -- code for method: vector(Matrix)
     M2/Macaulay2/m2/modules.m2:66:17-69:5: --source code:
     vector Matrix := f -> (
          if not isFreeModule source f or numgens source f =!= 1 then error "expected source to be free with rank 1";
          new target f from {f}
         )
```

Should `Vector` objects be elements of free modules?  Or arbitrary modules?

mahrud: That's not the problem. Single generators of any module whatsoever are vectors. The implementation of `Number + Vector` isn't correct because it uses the identity matrix on the target.

d-torrance: Hmm, this seems to be a problem with matrix addition:

```m2
i1 : M = coker matrix 10;

i2 : A = map(M, ZZ^1, 1)

o2 = | 1 |

                    1
o2 : Matrix M <-- ZZ

i3 : 1 + A

o3 = | 2 |

o3 : Matrix M <-- M

i4 : A + 1

o4 = | 2 |

                    1
o4 : Matrix M <-- ZZ
```

Shouldn't they both have `ZZ^1` as their source?

d-torrance: I just pushed a new commit that should fix this

mahrud (commented):
Apparently I had accidentally left a partial review on this months ago.

mahrud (approved):

Issue #3612, opened 2024-12-15
Augmented Null-coalescing op and immutable hash tables


mahrud:
I don't think this should be an error:
```m2
i1 : h = new HashTable from { symbol cache => new CacheTable };

i2 : h.cache

o2 = CacheTable{}

o2 : CacheTable

i3 : try h.cache else h.cache = new CacheTable

o3 = CacheTable{}

o3 : CacheTable

i4 : h.cache ??= new CacheTable
stdio:4:8:(3): error: attempted to modify an immutable hash table
```

mahrud:
Also, I'm confused about this outcome. I think it indicates that augmented null-coalescing operator acts like `=` rather than `:=`, but then why isn't `z` globally defined?
```m2
i1 : f = () -> ( y := null; y ??= 1; z ??= 2; );

i2 : peek User#"private dictionary"

o2 = GlobalDictionary{"f" => f}
                      "z" => z

i3 : f()

i4 : y

o4 = y

o4 : Symbol

i5 : z

o5 = z

o5 : Symbol
```

d-torrance:
In general, `x ??= y` does the same thing as `x = x ?? y`.  Isn't this the expected behavior when `x` is immutable?

```m2
i1 : h = new HashTable from { symbol cache => new CacheTable };

i2 : h.cache =  h.cache ?? new CacheTable
stdio:2:8:(3): error: attempted to modify an immutable hash table
```

And in the second example, `z` isn't an error or null, so it's not assigned 2 -- it just stays as the symbol `z`.  It works if it's assigned as null:

```m2
i1 : f = () -> ( z = null; z ??= 2; );

i2 : f()

i3 : z

o3 = 2
```

mahrud:
I see your point, but it would be very convenient if that syntax for making sure a hash table (mutable or not) has a key worked. For instance of `x ??= y` was instead interpreted as `x ?? (x = y)`.

d-torrance:
What's the desired behavior?  If `h` is an immutable hash table doesn't have a `cache` key already, then we shouldn't be adding one later on.  If we want to do that, then we should have created it as a mutable hash table.

mahrud:
[...]

PR #3614, opened 2024-12-17
Update behavior of augmented null coalescing operator

d-torrance:
`x ??= y` now behaves like `x ?? (x = y)` rather than `x = x ?? y`.

These are equivalent when x is null.

When x is non-null, we avoid an unnecessary re-assignment step, which also allows for the possibility of using this operator with immutable hash tables without raising errors when keys already exist.

Closes: #3612

### DIFF
--- a/M2/Macaulay2/d/evaluate.d
+++ b/M2/Macaulay2/d/evaluate.d
@@ -1236,16 +1236,37 @@ tryEval(c:Code):Expr := (
 	else tryEvalSuccess = true);
     p);
 
+nullify(c:Code):Expr := (
+    e := tryEval(c);
+    if tryEvalSuccess
+    then (
+	when e
+	is Nothing do e
+	else (
+	    f := lookup(Class(e), QuestionQuestionS);
+	    if f == nullE then e
+	    else applyEE(f, e)))
+    else nullE);
+
+nullCoalescion(lhs:Code,rhs:Code):Expr := (
+    e := nullify(lhs);
+    when e
+    is Nothing do eval(rhs)
+    else e);
+setup(QuestionQuestionS, nullify, nullCoalescion);
+
 augmentedAssignmentFun(x:augmentedAssignmentCode):Expr := (
     when lookup(x.oper.word, augmentedAssignmentOperatorTable)
     is null do buildErrorPacket("unknown augmented assignment operator")
     is s:Symbol do (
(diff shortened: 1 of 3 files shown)


M2/Macaulay2/d/evaluate.d
```diff
@@ -1236,16 +1236,37 @@ tryEval(c:Code):Expr := (
...
     when lookup(x.oper.word, augmentedAssignmentOperatorTable)
     is null do buildErrorPacket("unknown augmented assignment operator")
     is s:Symbol do (
 	-- evaluate the left-hand side first
 	lexpr := nullE;
-	if s.word.name === "??" -- null coalescion; ignore errors
+	if s.word.name === "??" -- x ??= y is treated like x ?? (x = y)
 	then (
-	    e := tryEval(x.lhs);
-	    if tryEvalSuccess then lexpr = e)
```
mahrud, line 1248: I'm a little confused. Where is the actual assignment to `lexpr` happening now?

d-torrance: We don't need it anymore when the left-hand side is non-null since we return right away.  And when it's null, we've already assigned `lexpr` as `nullE` a couple lines earlier.

d-torrance: The assignment happens in one of several places, depending on what the code looks like.  

`lexpr` is the *evaluated* left-hand side.  But we still have the original unevaluated `Code` object, which we need to figure out what type of assignment we'll end up doing (global, local, assigning an element of a list/hash table, etc.).  Here's the basic gist:

* Go ahead and evaluate the left hand side (`lexpr`) and stuff it in an `evaluatedCode` object (`left`). We also check for errors here and whether someone has installed an augmented assignment method for this operator for the class of `lexpr`.  If so, call that and return the result.
* Now go back and look at the original unevaluated `Code` object to figure out how to assign things.  But instead of using this `Code` object, we swap it out with `left` to avoid re-evaluating the left-hand side.

d-torrance:
Ok yeah -- at the very least we should fix the `-*dummy position*-` in the message.  But I suppose it would be cool if this worked like `a = a + 1; b = b + 1`.

mahrud:
Yeah, I don't necessarily _need_ that to work (at least it's not a priority), but the dummy position surprised me.

PR #3615, opened 2024-12-17
Augmented assignment updates

d-torrance:
This fixes the weird error message @mahrud pointed out in https://github.com/Macaulay2/M2/pull/3614#issuecomment-2549027711.  Now we have:

```m2
i1 : (a,b) ??= (1,2)
stdio:1:6:(3): error: augmented assignment not implemented for this code
```

Also a few other small updates that I noticed while looking at the code.

### DIFF
--- a/M2/Macaulay2/d/convertr.d
+++ b/M2/Macaulay2/d/convertr.d
@@ -298,7 +298,7 @@ export convert0(e:ParseTree):Code := (
 	      is t:Token do Code(augmentedAssignmentCode(
 		      b.Operator.entry, convert(b.lhs), convert(b.rhs), t.entry, pos))
 	      else Code(augmentedAssignmentCode(
-		      b.Operator.entry, dummyCode, dummyCode, dummySymbol, dummyPosition)) -- CHECK
+		      b.Operator.entry, convert(b.lhs), convert(b.rhs), dummySymbol, pos))
 		      )
 	  else Code(binaryCode(b.Operator.entry.binary, convert(b.lhs), convert(b.rhs), pos))
 	  )
--- a/M2/Macaulay2/d/evaluate.d
+++ b/M2/Macaulay2/d/evaluate.d
@@ -1268,21 +1268,20 @@ augmentedAssignmentFun(x:augmentedAssignmentCode):Expr := (
 	    is Nothing do nothing
 	    else return e)
 	else lexpr = eval(x.lhs);
-	left := evaluatedCode(lexpr, dummyPosition);
-	when left.expr is e:Error do return Expr(e) else nothing;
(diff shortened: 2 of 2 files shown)


M2/Macaulay2/d/convertr.d
```diff
@@ -298,7 +298,7 @@ export convert0(e:ParseTree):Code := (
 	      is t:Token do Code(augmentedAssignmentCode(
 		      b.Operator.entry, convert(b.lhs), convert(b.rhs), t.entry, pos))
 	      else Code(augmentedAssignmentCode(
-		      b.Operator.entry, dummyCode, dummyCode, dummySymbol, dummyPosition)) -- CHECK
+		      b.Operator.entry, dummyCode, dummyCode, dummySymbol, pos)) -- CHECK
```
mahrud, line 301: Maybe also remove the `CHECK` while you're editing this line?

mahrud: Also, are the `dummyCode`s and `dummySymbol` right?

M2/Macaulay2/d/evaluate.d
```diff
@@ -1268,46 +1268,45 @@ augmentedAssignmentFun(x:augmentedAssignmentCode):Expr := (
...
+	    rexpr := eval(x.rhs);
+	    when rexpr is e:Error do return rexpr else nothing;
+	    r := applyEEE(meth, lexpr, rexpr);
 	    when r
 	    is s:SymbolClosure do (
 		if s.symbol.word.name === "Default" then nothing
 		else return r)
 	    else return r);
 	-- if not, use default behavior
+	left := evaluatedCode(lexpr, codePosition(x.lhs));
```
mahrud, line 1284: I know writing tests for errors is complicated, but did you try to reproduce an error at the errors points below?

d-torrance: Oh yeah -- they're easy to reproduce.  For example:

```m2
i1 : x = 5

o1 = 5

i2 : x += "foo"
stdio:2:2:(3): error: no method for binary operator + applied to objects:
            5 (of class ZZ)
      +     "foo" (of class String)
```

M2/Macaulay2/d/convertr.d
d-torrance: I think if we end up at this line, we're bound to get a "not implemented" error, so I don't think it really matters.  We might as well call `convert` on the two sides, I suppose.

I think `dummySymbol` probably makes the most sense for `info` in this case.  This field is kind of a hack -- it's either the variable name  we're assigning to or the operator on the left hand side, depending on the situation.

mahrud (approved):

Issue #3603, opened 2024-11-29
applyPairs for lists


mahrud:
This:
```m2
hashTable applyPairs({10,20,30}, identity)
```
should act like this:
```m2
i12 : hashTable toList apply(pairs {10,20,30}, identity)

o12 = HashTable{0 => 10}
                1 => 20
                2 => 30
```

d-torrance:
Probably the same for `selectPairs` and `scanPairs`.

Would it be better to turn this into methods, or implement the list versions in the interpreter?

mahrud:
Between methods or interpreter, I'm not sure which is better.

By the way, you could define `pairs Iterator` to give a new iterator and define `applyPairs` for an iterator.

d-torrance:
This was implemented in #3617

PR #3617, opened 2024-12-18
Add several pairs-related methods

d-torrance:
We add `pairs` for any class with an `iterator` method as well as `applyPairs`, `scanPairs`, and `selectPairs` for non-hash tables.  These are just shortcuts for `apply(pairs ...)`, `scan(pairs ...)`, and `select(pairs ...)`.

```m2
i1 : take(pairs Iterator(() -> 5), 4)

o1 = {(0, 5), (1, 5), (2, 5), (3, 5)}

o1 : List

i2 : applyPairs({2, 4, 6}, (i, x) -> x^i)

o2 = (1, 4, 36)

o2 : Sequence

i3 : scanPairs({2, 4, 6}, (i, x) -> print i^x)
0
1
64

i4 : selectPairs({2, 4, 6}, (i, x) -> even(i + x))

o4 = ((0, 2), (2, 6))

o4 : Sequence
```

`selectPairs` doesn't currently work for iterators since `select` is defined for them at top-level, and this is all done in the interpreter.

Closes: #3603

### DIFF
--- a/M2/Macaulay2/d/actors2.dd
+++ b/M2/Macaulay2/d/actors2.dd
@@ -104,7 +104,7 @@ values(e:Expr):Expr := (
      else WrongArg("a hash table or dictionary"));
 setupfun("values",values);
 
-pairs(e:Expr):Expr := (
+export pairs(e:Expr):Expr := (
      when e
      is oc:DictionaryClosure do (    -- # typical value: pairs, Dictionary, List
 	  o := oc.dictionary.symboltable;
--- a/M2/Macaulay2/d/actors2.dd
+++ b/M2/Macaulay2/d/actors2.dd
@@ -142,7 +142,12 @@ export pairs(e:Expr):Expr := (
 	    while i < length(o) do (
 	      provide Expr(Sequence(toExpr(i),o.i));
 	      i = i+1;)))
-     is o:List do pairs(Expr(o.v))   -- # typical value: pairs, BasicList, List
+     -- # typical value: pairs, BasicList, List
+     is o:List do (
+	 r := pairs(Expr(o.v));
+	 when r
+	 is s:Sequence do list(o.Class, s, o.Mutable)
+	 else buildErrorPacket("internal error; expected a sequence"))
      else applyEE(getGlobalVariable(pairsIteratorS), e)); -- # typical value: pairs, Thing, Iterator
 setupfun("pairs",pairs);
 
--- a/M2/Macaulay2/d/actors2.dd
+++ b/M2/Macaulay2/d/actors2.dd
@@ -143,7 +143,7 @@ pairs(e:Expr):Expr := (
 	      provide Expr(Sequence(toExpr(i),o.i));
 	      i = i+1;)))
      is o:List do pairs(Expr(o.v))   -- # typical value: pairs, BasicList, List
-     else WrongArg("a hash table, a sequence, a list, or a raw polynomial"));
+     else applyEE(getGlobalVariable(pairsIteratorS), e)); -- # typical value: pairs, Thing, Iterator
 setupfun("pairs",pairs);
 
 -- operators
--- a/M2/Macaulay2/d/actors3.d
+++ b/M2/Macaulay2/d/actors3.d
@@ -82,7 +82,7 @@ override(e:Expr):Expr := (
 	  if length(args) == 2 then (
 	       when args.0
 	       is h:HashTable do (
-		    if h.Mutable then WrongArg("an immutable hash table")
+		    if h.Mutable then WrongArgImmutableHashTable()
 		    else when args.1 is v:Sequence do override(h,v,numOptions(v))
 		    else override(h,Sequence(args.1),if isOption(args.1) then 1 else 0)
 		    )
--- a/M2/Macaulay2/d/actors3.d
+++ b/M2/Macaulay2/d/actors3.d
@@ -2363,6 +2363,20 @@ scan(e:Expr):Expr := (
      else WrongNumArgs(2));
 setupfun("scan",scan);
 
+-- # typical value: scanPairs, Thing, Function, Nothing
+scanPairs(e:Expr):Expr := (
+    when e
+    is a:Sequence do (
+	if length(a) == 2 then (
+	    when a.0
+	    is o:HashTable do (
+		if o.Mutable then WrongArgImmutableHashTable(1)
+		else scanpairs(a.1, o))
+	    else scan(pairs(a.0), a.1))
+	else WrongNumArgs(2))
+    else WrongNumArgs(2));
+setupfun("scanPairs", scanPairs);
+
 nextPrime(e:Expr):Expr := (
      when e
      is x:ZZcell do toExpr(nextPrime(x.v - oneZZ))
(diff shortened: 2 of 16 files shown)


M2/Macaulay2/d/actors3.d
```diff
@@ -1895,6 +1896,22 @@ map(e:Expr):Expr := (
      else WrongNumArgs(2,3));
 setupfun("apply",map);
 
+-- # typical value: applyPairs, BasicList, Function, List
+-- # typical value: applyPairs, Dictionary, Function, List
+-- # typical value: applyPairs, Thing, Function, Iterator
```
mahrud, line 1901: I think it would be better to list these closer to where they are handled (e.g. near `mappairs` or `map`. We don't currently do this, but I'd like to have `locate` also show the location of compiled functions in the interpreter (the location is stored in `tvalues.m2` as  a comment right now).

M2/Macaulay2/packages/Macaulay2Doc/doc_tables.m2
```diff
@@ -233,15 +233,16 @@ doc ///
...
  Usage
   pairs x
  Inputs
-  x:HashTable
-   or @TO Dictionary@ or @TO BasicList@
+  x:{HashTable, Dictionary, BasicList}
+   or any instance of a class with an @TO iterator@ method installed
  Outputs
-  L:List
+  L:{List, Iterator}
```
mahrud, line 245: I think it might be better to separate the documentation of the iterator form, ideally using `Synopsis`.

M2/Macaulay2/packages/Macaulay2Doc/functions/select-doc.m2
```diff
@@ -274,10 +276,10 @@ doc ///
     selectPairs(n, x, f)
   Inputs
     n:ZZ
-    x:HashTable -- must be immutable
+    x:{HashTable, BasicList}
     f:Function
   Outputs
-    :HashTable
+    :{HashTable, Sequence}
```
mahrud, line 282: Why a Sequence?

mahrud: Oh I guess this should change after your latest commit.

d-torrance: Yeah -- this was based on the current "always a sequence" behavior.  The current draft always returns a `List` object, but I like the idea of behaving like `apply` and returning an instance of the same class.

M2/Macaulay2/packages/Macaulay2Doc/doc_tables.m2
d-torrance: Cool -- I didn't know about `Synopsis`!  Definitely useful for stuff like this.

d-torrance:
I added a few more commits updating `hashTable` to allow `Sequence` arguments, as suggested by the original title of #3619.  This way, `hashTable @@ pairs` will work with `Sequence` objects.

M2/Macaulay2/d/actors2.dd
```diff
@@ -142,8 +142,13 @@ pairs(e:Expr):Expr := (
...
 	      i = i+1;)))
-     is o:List do pairs(Expr(o.v))   -- # typical value: pairs, BasicList, List
-     else WrongArg("a hash table, a sequence, a list, or a raw polynomial"));
+     -- # typical value: pairs, BasicList, List
+     is o:List do (
+	 r := pairs(Expr(o.v));
+	 when r
+	 is s:Sequence do list(o.Class, s, o.Mutable)
+	 else buildErrorPacket("internal error; expected a sequence"))
+     else applyEE(getGlobalVariable(pairsIteratorS), e)); -- # typical value: pairs, Thing, Iterator
```
mahrud, line 151: fwiw I think putting the `typical value` tag on its own line would be better.

M2/Macaulay2/d/actors2.dd
```diff
@@ -142,8 +142,13 @@ pairs(e:Expr):Expr := (
...
 	      provide Expr(Sequence(toExpr(i),o.i));
 	      i = i+1;)))
-     is o:List do pairs(Expr(o.v))   -- # typical value: pairs, BasicList, List
-     else WrongArg("a hash table, a sequence, a list, or a raw polynomial"));
+     -- # typical value: pairs, BasicList, List
+     is o:List do (
+	 r := pairs(Expr(o.v));
+	 when r
+	 is s:Sequence do list(o.Class, s, o.Mutable)
+	 else buildErrorPacket("internal error; expected a sequence"))
```
mahrud, line 150: when would `pairs(Expr(o.v))` with `o:List` fail?

M2/Macaulay2/d/actors3.d
```diff
@@ -2346,6 +2363,20 @@ scan(e:Expr):Expr := (
      else WrongNumArgs(2));
 setupfun("scan",scan);
 
+-- # typical value: scanPairs, Thing, Function, Nothing
```
mahrud, line 2366: Since this line is for readability of documentation and doesn't have to be literally true, I wonder if it would be useful to use a dummy `Iterable` class instead of `Thing` for things like this. I know that practically anything with an `iterator` method defined on it can be iterable, so it may cause some confusion.

M2/Macaulay2/d/actors4.d
```diff
@@ -200,6 +200,8 @@ select(e:Expr):Expr := (
      else WrongNumArgs(2,5));
 setupfun("select", select).Protected = false; -- will be overloaded in m2/lists.m2 and m2/regex.m2
 
+-- # typical value: selectPairs, HashTable, Function, HashTable
+-- # typical value: selectPairs, ZZ, HashTable, Function, HashTable
 selectPairs(nval:int, obj:HashTable, f:Expr):Expr := (
```
mahrud, line 205: I think at some point we need to decide on a convention about where to place these lines. I don't find placing both of these here particularly useful. To me it either means one of them should be defined on the top level, or one of them is in the wrong place.

M2/Macaulay2/d/expr.d
```diff
@@ -389,6 +389,8 @@ export WrongArgMatrix(n:int):Expr := WrongArg(n,"a raw matrix");
 export WrongArgMatrix():Expr := WrongArg("a raw matrix");
 export WrongArgHashTable():Expr := WrongArg("a hash table");
 export WrongArgHashTable(n:int):Expr := WrongArg(n, "a hash table");
+export WrongArgImmutableHashTable():Expr := WrongArg("an immutable hash table");
+export WrongArgImmutableHashTable(n:int):Expr := WrongArg(n, "an immutable hash table");
```
mahrud, line 393: Seems to me like `WrongArgHashTable` should say `immutable hash table` and we should have a `WrongArgMutableHashTable` which says `mutable hash table`.

mahrud (approved):
Looks good to me. I left some comments, but most of them are rather subjective and don't need to block merging.

M2/Macaulay2/d/actors2.dd
d-torrance: It shouldn't -- we should always get a sequence here. I don't expect we'll ever raise this error.

d-torrance: Hmm, unless `pairs` raises an error... I'll look into this more.

Issue #2534, opened 2022-06-27
Texmacs does not display outcome


Corrado-possieri:
I used to run Macaulay2 within texmacs, but it does not work anymore.
In particular, texmacs is not able to display the output of Macaulay2 commands.
If a command ends with ";" it is correctly executed, but if one wants to see the output it is stuck on busy.
I am running both Macaulay and TexMacs on an M1 MacbookPro.

DanGrayson:
I get the same behavior on my macbook:

<img width="798" alt="Screen Shot 2022-07-01 at 8 41 43 AM" src="https://user-images.githubusercontent.com/700228/176896660-a5f66909-592f-4997-88e3-1cd98df7723a.png">

I'll ask Joris.

DanGrayson:
Here's what the raw output looks like:

<img width="953" alt="Screen Shot 2022-07-01 at 8 51 17 AM" src="https://user-images.githubusercontent.com/700228/176898148-65e8c555-7d1f-4f16-af14-5d347e70d53d.png">

Corrado-possieri:
Do you have any news on this aspect?

bztd:
Ask about it in the TeXmacs forum

http://forum.texmacs.cn/t/macaulay-does-not-advance/1942

mgubi:
It seems to me that the invocation of the `separate` function is wrong:
```
i1 : s="pippo"

o1 = pippo

i2 : separate("\0",s)

o2 = {p, i, p, p, o}

o2 : List

i3 : separate("\\0",s)

o3 = {pippo}

o3 : List
```
so I think one should use `"\\0"` instead of `"\0"` as argument to separate. With this fix, TeXmacs works correctly, but there is still a problem with matrices:

<img width="786" alt="Image" src="https://github.com/user-attachments/assets/c916625b-3316-4a10-87e2-0e18e8cca4cc" />
As you can see the matrix is not typesetter correctly.

d-torrance:
Fixed in #3627

PR #3627, opened 2025-01-17
TeXmacs fixes

d-torrance:
This includes @mgubi's [proposed fix](http://forum.texmacs.cn/t/macaulay-does-not-advance/1942/12) for #2534 and also fixes the reported matrix display issue.  This was a small bug in `mathML(MatrixExpression)` where we were trying to just display the first row.

### Before
image

### After
image

@mgubi, @bztd:  Thoughts?

### DIFF
--- a/M2/Macaulay2/m2/mathml.m2
+++ b/M2/Macaulay2/m2/mathml.m2
@@ -59,7 +59,7 @@ mathML FunctionApplication := m -> (
      then concatenate (mfun, bigParenthesize margs)
      else concatenate (bigParenthesize mfun, bigParenthesize margs)
      )
-mathML MatrixExpression := x -> concatenate( "<mrow><mo>(</mo>", mtableML x#0, "<mo>)</mo></mrow>", newline )
+mathML MatrixExpression := x -> concatenate( "<mrow><mo>(</mo>", mtableML x, "<mo>)</mo></mrow>", newline )
 mathML Minus := v -> concatenate( "<mo>-</mo>", mathML v#0)
 mathML Divide := x -> concatenate("<mfrac>", mathML x#0, mathML x#1, "</mfrac>")
 mathML OneExpression := x -> "<mn>1</mn>"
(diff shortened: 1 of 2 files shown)


bztd:
I will apply the changes to the stable version and do testing.

d-torrance:
I added a couple more commits adding `AfterPrint` support for TeXmacs mode.  So for example, we get the source/target info for matrices like we do with the standard and webapp modes:

image

mgubi:
I do not know Macauly2, so I trust you :) the output seems correct. What about the matrix which includes symbolic entries and is in the TeXmacs' help for the M2 Plugin? (Help->Plugin->Macauly2).

d-torrance:
Yes, that works!

image

mgubi:
I gave a look at the web app. Since TeXmacs can read plugin output in HTML format, one could reuse the code written for the web app to produce HTML output to send to texmacs. Most of the features of the web app could be emulated in TeXmacs.

d-torrance:
The webapp mode outputs html, but the math is mostly `$`-delimited LaTeX that is rendered using Javascript (KaTeX).  For example:

```html
i2 : 2:0matrix {{1, 2}, {3, 4}}

o2 = $\left(\!\begin{array}{cc}
1&amp;2\\
3&amp;4
\end{array}\!\right)$

o2 : <span><samp class="token class-name">Matrix</samp> ${\mathbb Z}^{2}\,\longleftarrow \,{\mathbb Z}^{2}$</span>
```

Is that something that TeXmacs would support?  Otherwise, I think the current MathML-based implementation mostly works.

mgubi:
IIRC TeXmacs imports correctly Wikipedia pages with math (which should be MathJax). Alternatively one needs to switch to "latex" input in those pieces. TeXmacs supports "html", "latex" and some others.

bztd:
In my local tests M2 already works in TeXmacs.

d-torrance:
Great!  I'll go ahead and merge this

PR #3630, opened 2025-01-19
Add "sign" method

d-torrance:
I'm amazed that we don't have this already!

We add a `sign` method that wraps the corresponding functions in GMP and MPFR (and returns $\frac{z}{|z|}$ for nonzero complex $z$).  For example:

```m2
i1 : sign 5

o1 = 1

i2 : sign 0

o2 = 0

i3 : sign(-3)

o3 = -1
```


mahrud:
I'll be honest, I don't think I like `sign` being reserved for something that in principle is as simple as `x // abs x`. Sign of a permutation is probably useful as a method, but again the question is whether it's worth reserving a symbol which may be used in people's code for other things.

d-torrance:
We already have two packages (`GradedLieAlgebras` and `Permutations`) there were exporting their own `sign` methods, so I think it makes sense to move the actual method function definition to `Core`.   There were also a couple other packages that were defining `sign` internally with the same behavior that I'm proposing, and this simplifies things by moving all that code to one place.

There were a number of packages that used `sign` as a variable.  I went ahead and changed those to `sgn`.  There's certainly a chance this might break other code out in the wild, but that's a risk any time we add something new.

M2/Macaulay2/packages/Macaulay2Doc/functions/sign-doc.m2
```diff
@@ -0,0 +1,26 @@
+doc ///
+  Key
+    sign
```
mahrud, line 3: If this is a shared symbol, then it should be documented similar to other shared symbols, with links to nodes in other packages that enhance it.

M2/Macaulay2/packages/GradedLieAlgebras.m2
```diff
@@ -72,7 +72,6 @@ export {
     "mbRing",
     "minimalModel",
     "normalForm",
-    "sign",
```
mahrud, line 75: You removed it from here, but several places in the documentation link to `sign`, which now goes to an irrelevant node in Macaulay2Doc.

M2/Macaulay2/packages/Permutations.m2
```diff
@@ -33,7 +33,6 @@ export {
     "isCDG",
     "foataBijection",
     "ord",
-    "sign",
```
mahrud, line 36: Same problem here, this change mangles the documentation links to `sign`.

mahrud (changes requested):
I'm still not a fan of the places where a variable or function named `sign` is changed to `sgn`. Why not define [sgn](https://en.wikipedia.org/wiki/Sign_function) instead?

d-torrance:
I've made the requested documentation changes, I think.

Regarding `sign` v. `sgn`:  I went with the choice that was consistent with the two packages that already define this method.

seangrate:
I think `sign` makes sense to define and be moved to the Core as a method; then the `GradedLieAlgebras` and `Permutations` packages just extend that method.

Regarding taking up another symbol name: I think this is just a byproduct of Macaulay2's structure where there is a just a giant dictionary of defined symbols (I'm not sure the technical details here; I'm just a fan of a namespace with structure like `packageName.methodName`). For instance, as a user, I might want to keep track of the difference between two objects, but I can't use `difference` as the symbol name, and if I try to shorten it to `diff` to bypass this, I still can't because `diff` is also reserved. But this might be getting beside the point...

mahrud (approved):
I forgot this was still waiting.

PR #3633, opened 2025-01-22
Update mathml.m2 to avoid display errors in texmacs

Corrado-possieri:
Update of the file mathml.m2 so to avoid display errors in Texmacs showing "bad mfrac" when typing fractions with negative sign.

### DIFF
--- a/M2/Macaulay2/m2/mathml.m2
+++ b/M2/Macaulay2/m2/mathml.m2
@@ -61,7 +61,7 @@ mathML FunctionApplication := m -> (
      )
 mathML MatrixExpression := x -> concatenate( "<mrow><mo>(</mo>", mtableML x, "<mo>)</mo></mrow>", newline )
 mathML Minus := v -> concatenate( "<mo>-</mo>", mathML v#0)
-mathML Divide := x -> concatenate("<mfrac>", mathML x#0, mathML x#1, "</mfrac>")
+mathML Divide := x -> concatenate("<mfrac><mrow>", mathML x#0,"</mrow>", mathML x#1, "</mfrac>")
 mathML OneExpression := x -> "<mn>1</mn>"
 mathML ZeroExpression := x -> "<mn>0</mn>"
 mathML Sum := v -> (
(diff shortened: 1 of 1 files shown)


M2/Macaulay2/m2/mathml.m2
```diff
@@ -61,7 +61,7 @@ mathML FunctionApplication := m -> (
      )
 mathML MatrixExpression := x -> concatenate( "<mrow><mo>(</mo>", mtableML x, "<mo>)</mo></mrow>", newline )
 mathML Minus := v -> concatenate( "<mo>-</mo>", mathML v#0)
-mathML Divide := x -> concatenate("<mfrac>", mathML x#0, mathML x#1, "</mfrac>")
+mathML Divide := x -> concatenate("<mfrac><mrow>", mathML x#0,"</mrow>", mathML x#1, "</mfrac>")
```
d-torrance, line 64: Should the denominator be enclosed in `<mrow>` tags too?

This is kind of artificial since I'm not sure if it will ever happen by accident, but I get the following:

image

This is what we get with standard mode:

```m2
i1 : Divide(1, Minus 2)

      1
o1 = --
     -2

o1 : Expression of class Divide
```

Or maybe `mathML Minus` should enclose everything in `<mrow>` tags?

Corrado-possieri: I don't think that also the denominator needs to be enclosed in `<mrow>` tags (it simply put the minus in the numerator). 
Probably also the edit on the mathML Minus should work. 
I tested it and changing `mathML Minus` into
`mathML Minus := v -> concatenate( "<mrow><mo>-</mo>", mathML v#0, "</mrow>")`
by keeping `mathML Divide` unaltered works fine.

d-torrance: Yeah, I think I like updating `mathML Minus` instead.  It's not just `Divide` objects that are giving us issues when they contain `Minus` objects.  For example:
image

Corrado-possieri: Yeah, it seems better.  Thanks a lot.
Should I update the pull request?

d-torrance: Yes, please -- that would be great!

Corrado-possieri: Done.

d-torrance: Thanks!

d-torrance (approved):

PR #3641, opened 2025-02-06
Don't try to get typical values from nonexistent files

d-torrance:
This fixes a strange bug I ran into.

On my laptop, somehow I ended up with a .d file that was a symlink to a nonexistent file.  When building Macaulay2 for the first time, it tried to generate `tvalues.m2`.  By default, it calls `get` on all the .d and .dd files in the d directory and writes some things to `tvalues.m2`.  But in my case, `get` failed, and so we got an error but ended up with an empty `tvalues.m2`.

Then when loading M2 the next time, it saw that `tvalues.m2` existed (even though it was empty), and proceeded on its merry way.  But none of  the methods that should have been defined from compiled functions were defined!  This ended up leading to a bunch of errors in Macaulay2Doc for documentation keys with no corresponding methods.

A simple fix is to just skip over files where `fileExists` fails.  (I think -- is there any edge case where `fileExists` will return true but `get` fails?)

### DIFF
--- a/M2/Macaulay2/m2/typicalvalues.m2
+++ b/M2/Macaulay2/m2/typicalvalues.m2
@@ -110,8 +110,9 @@ generateTypicalValues = (srcdir) -> (
     printerr("Generating typical values in ", relativizeFilename typicalValuesSource);
     outfile := openOut typicalValuesSource;
     for file in sort ddfiles do (
+	try src := get(srcdir | file) else continue;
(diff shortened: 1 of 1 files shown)


d-torrance:
I updated this a bit so we don't have to worry about whether `fileExists` and `get` agree, and instead just use `try` with `get`.

mahrud (approved):

Issue #3643, opened 2025-02-07
Path to source ignored in core tests


antonleykin:
The attempt to specify `dir` in
`M2 --srcdir dir --check 2`
has not produced the expected outcome.
The tests were still grabbed from 
`/opt/homebrew/share/doc/Macaulay2/Core/tests/`

d-torrance:
Fixed in #3644

PR #3644, opened 2025-02-08
Prioritize running Core tests from source repository

d-torrance:
If we're running a development version of Macaulay2 from the source repository, then we potentially might want to make changes to the tests in the source repository.  So we load the Core tests from that directory rather than from the installed tests in say, /usr/share/doc/Macaulay2/Core/tests.

Closes: #3643


M2/Macaulay2/m2/testing.m2
```diff
@@ -60,9 +60,14 @@ captureTestResult := (desc, teststring, pkg, usermode) -> (
...
 loadTestDir := pkg -> (
-    -- TODO: prioritize reading the tests from topSrcdir | "Macaulay2/tests/normal" instead
-    testDir := pkg#"package prefix" |
-        replace("PKG", pkg#"pkgname", currentLayout#"packagetests");
+    testDir := (
+	-- prioritize Core tests from source repository
+	if (pkg === Core and (
+		dir := minimizeFilename(pkg#"source directory" |
+		    "../tests/normal/");
+		isDirectory dir)) then dir
```
mahrud, line 68: The correct way to check if we're running from the build directory is to check if `topSrcdir` is not null.

d-torrance: That only works if we're running from the *build* directory, though.  The use case that Anton pointed out to me yesterday was when running the regular old system copy of the M2 binary, but using the `--srcdir` command line option to grab Core from the source repository to test a new feature.

mahrud: Why would this even work?

d-torrance: That's correct -- it doesn't work a lot of the time exactly for this reason. But it's useful when it does!

mahrud: Is there a solution which involves setting `topSrcdir` in this case as well?

d-torrance: `--srcdir` really only adds a directory to the path.  *Maybe* it's the path to Core source, but not necessarily.  So I'm not sure if setting topSrcdir based on its value makes sense.

mahrud:
> `--srcdir` really only adds a directory to the path. _Maybe_ it's the path to Core source, but not necessarily. So I'm not sure if setting topSrcdir based on its value makes sense.

I've always been unclear about this, so I like the idea of taking this opportunity to clarify how `--srcdir` should be used. In particular, I think it should give an error if the output isn't one of two cases:
1. a source directory like `<git root>/M2/Macaulay2` containing `m2/Core.m2`
2. a path like `share/Macaulay2` containing `Core/Core.m2`

How does that sound? Any other directories that make sense?

M2/Macaulay2/m2/testing.m2
mahrud: Something about this still deeply bothers me:
```m2
	if (pkg === Core and (
		dir := minimizeFilename(pkg#"source directory" |
		    "../tests/normal/");
```

d-torrance:
I looked at the source more carefully and apparently `--srcdir` only does anything when its argument contains `Macaulay2/m2` and/or `Macaulay2/packages` directories and adds them to the path.  So yeah, setting `topSrcdir` in this case definitely makes sense.

antonleykin (approved):
Nice work!

PR #3645, opened 2025-02-08
Remove emacs submodule

d-torrance:
As discussed at Fields earlier, we remove the M2-emacs submodule.  This will prevent the common occurrence of accidentally updating the submodule hash in a commit.

 The sources for the generated files are restored to `M2/Macaulay2/editors/emacs`.  The generated files can then be copied to the M2-emacs repository at each release.

Another question is distribution.  If the M2-emacs files are no longer included in the main source repository, then what's the best way to get them to users?  Maybe this is more of a package maintainer issue for the various distributions (e.g., I would need to make a few updates to how the Debian/Ubuntu elpa-macaulay2 package is built.)  But I suppose `setupEmacs` could be modified to say, download a tarball of the M2-emacs repo from GitHub and install it.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -25,6 +25,3 @@
 [submodule "M2/submodules/googletest"]
 	path = M2/submodules/googletest
 	url = https://github.com/google/googletest.git
-[submodule "M2/Macaulay2/editors/emacs"]
-	path = M2/Macaulay2/editors/emacs
-	url = https://github.com/Macaulay2/M2-emacs.git
--- a/M2/Macaulay2/editors/Makefile.in
+++ b/M2/Macaulay2/editors/Makefile.in
@@ -4,10 +4,9 @@ ARGS = --script
 .NOTPARALLEL:
 
 MADEFILES = M2-symbols.el M2-emacs-help.txt M2-emacs.m2
-SRCFILES = M2-init.el M2-mode.el M2.el README.md
 
-all: @pre_emacsdir@ $(addprefix @pre_emacsdir@/, $(SRCFILES) $(MADEFILES))
-verify: $(addprefix @pre_emacsdir@/, $(SRCFILES) $(MADEFILES)) $(addprefix emacs/, $(SRCFILES) $(MADEFILES)); ls -lrt $^
+all: @pre_emacsdir@ $(addprefix @pre_emacsdir@/, $(MADEFILES))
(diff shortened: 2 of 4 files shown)


mikestillman:
For an uninitiated user, getting the emacs package from GitHub and installing it is a bit of a pain.  I would rather make that as easy as possible for users not used to dealing with emacs…. I’m not sure the best way to fix this problem.

I know I said I was annoyed by the changes to the submodule (true), but I hope there is a way they can get it with no real thought involved…

Mike

mahrud:
I disagree with this change. Now there are two copies of M2-symbols.el.in across different repositories that need to be synced.

> This will prevent the common occurrence of accidentally updating the submodule hash in a commit.

The simpler solution is to just disable building the emacs stuff by autotools by default, or alternatively entirely removing the emacs subdirectory.

mahrud:
Alternatively, you can update the autotools build to not change the emacs files directly in the submodule, but instead in the build directory.

d-torrance:
> For an uninitiated user, getting the emacs package from GitHub and installing it is a bit of a pain. I would rather make that as easy as possible for users not used to dealing with emacs…. I’m not sure the best way to fix this problem.

I think it would be pretty straightforward to have `setupEmacs` automate this.  Download the tarball from GitHub, then fire up Emacs in batch mode to run `package-install`.

> Now there are two copies of M2-symbols.el.in across different repositories that need to be synced.

I would propose removing the duplicate files from M2-emacs if this PR is accepted.  The generated files, like M2-symbols.el.gz, would be updated in M2-emacs at each new release.  Maybe the generated documentation files can be copied over, or maybe we can just remove them.

> Alternatively, you can update the autotools build to not change the emacs files directly in the submodule, but instead in the build directory.

That's already the behavior.

mahrud:
> That's already the behavior.

How is the submodule changing then?

d-torrance:
Closing for now.  It might make more sense to do something like this if we move the m2 code that generates the various syntax highlighting files from templates to a package like Style -- see https://github.com/Macaulay2/M2/pull/3298#issuecomment-2183065149.

mahrud:
We should add `ignore = all` to `.gitmodule`:
https://git-scm.com/docs/gitmodules#Documentation/gitmodules.txt-submoduleltnamegtignore

PR #3647, opened 2025-02-09
Documentation stub for the profiler

mahrud:
Follow up to #3431.


mahrud:
That PR also exported a couple of other new things: breakpoint and (a different) stacktrace, but those aren't documented yet.

Issue #3454, opened 2024-09-04
Killing unreachable threads


mahrud:
If I run a function which starts a number of threads and waits for them to finish, but then I interrupt the parent function, the orphaned threads continue to run and I think there's no way to cancel them.

It would be great to:
1. when a local dictionary closes (e.g. function ends or is interrupted and control is handed back to the user) any remaining active thread objects in the local dictionary which are now inaccessible (i.e. not returned to the user) should be cancelled.
2. there should be a way to retrieve all active threads, and a method to cancel them all.

mahrud:
New, super important wish-list item:

3. unreachable tasks should be cancelled.

That is to say, if I make a task inside a function and that function returns some value without ever doing anything with the task, the now-unreachable task shouldn't continue freely. There could maybe be a `KeepAlive` flag to `schedule` to not cancel particular tasks, but that should be an intentional decision.

mahrud:
I had some tasks running in the background and accidentally started something really long in the main thread, which I then had to abort (didn't gracefully stop), and in they chaos I saw messages like this printed:
```m2
--cancelling inaccessible task 1234 still running
```
This is printed by `taskCellFinalizer` so I tried tracing to see where was this called from, but couldn't exactly figure out where did the list of inaccessible tasks come from. It would be great if we could initiate whatever started this manually.

Issue #3439, opened 2024-08-26
Bizarre crash involving local symbol and computing basis in parallel


mahrud:
I accidentally made a typo in a random piece of code I wrote, and I can't understand the ensuing crash. To replicate, put this in a file and run from "needs" until the end.
```m2
await = method(Dispatch => Thing)
await Thing := identity
await BasicList := L -> apply(L, await)
await Task := await @@ taskResult

async = method(Dispatch => Thing)
async Function := f -> (x -> schedule(f, x))

repeat = (n, R) -> (
    if n == 1 then return n;
    r := {repeat(n//2, R), repeat(n//2, R)};
    basis(10, ideal vars R);
    t) --                                <-- here is my typo!
repeat = async repeat

end--
restart
needs "parallel.m2"

R = ZZ/101[x_0..x_10]
elapsedTime await repeat(allowableThreads*2, R)
```
Note the typo I made, the `t` should have been an `r`. Here is what I got on a `1.24.05` binary:
```m2
i1 : needs "parallel.m2"

i2 : 
     R = ZZ/101[x_0..x_10]

o2 = R

o2 : PolynomialRing

i3 : elapsedTime await repeat(allowableThreads*2, R)
 -- 0.38629 seconds elapsed

o3 = t

o3 : Symbol

i4 : /home/linuxbrew/.linuxbrew/share/Macaulay2/Core/expressions.m2:675:47:(1):[21]: error: unknown engine error
/home/linuxbrew/.linuxbrew/share/Macaulay2/Core/expressions.m2:675:47:(1):[21]: error: unknown engine error
/home/linuxbrew/.linuxbrew/share/Macaulay2/Core/expressions.m2:675:47:(1):[21]: error: unknown engine error
/home/linuxbrew/.linuxbrew/share/Macaulay2/Core/basis.m2:262:10:(1):[19]: --back trace--
/home/linuxbrew/.linuxbrew/share/Macaulay2/Core/basis.m2:262:10:(1):[19]: --back trace--
/home/linuxbrew/.linuxbrew/share/Macaulay2/Core/basis.m2:262:10:(1):[19]: --back trace--
/home/linuxbrew/.linuxbrew/share/Macaulay2/Core/methods.m2:625:51:(1):[18]: --back trace--
/home/linuxbrew/.linuxbrew/share/Macaulay2/Core/methods.m2:625:51:(1):[18]: --back trace--
[...]

PR #3648, opened 2025-02-09
async/await pattern

mahrud:
This adds a very rudimentary, top-level version of the async/await parallelization [pattern](https://en.wikipedia.org/wiki/Async/await). Here is a demo:
```m2
i1 : f = x -> (sleep 1; 2*x);

i2 : elapsedTime apply(3, f)
 -- 3.0006s elapsed

o2 = {0, 2, 4}

o2 : List

i3 : elapsedTime await apply(3, async f)
 -- 1.0005s elapsed

o3 = {0, 2, 4}

o3 : List
```
The main benefit of this pattern is simplicity, requiring adding as little as two words to make a computation run in parallel. Specifically, the advantage over `parallelApply` is that you don't need a host of new functions, instead you can turn existing functions parallel. More examples and good practices are [here](https://javascript.info/async-await).

Here are some more examples of instantly turning all/any/position parallel:
```m2
i4 : g = async f;

i5 : elapsedTime all({ f 2, f 3, f 4 }, even)
 -- 3.00085s elapsed

o5 = true

i6 : elapsedTime all({ g 2, g 3, g 4 }, even @@ await)
 -- 1.00104s elapsed

o6 = true

i7 : elapsedTime any({ f 2, f 3, f 4 }, i -> i % 3 == 0)
 -- 3.00053s elapsed

o7 = true

i8 : elapsedTime any({ g 2, g 3, g 4 }, i -> (await i) % 3 == 0)
 -- 1.00048s elapsed

o8 = true

i9 : elapsedTime position({ 2, 3, 4 }, i -> (f i) % 3 == 0)
 -- 2.00041s elapsed

o9 = 1

i10 : elapsedTime position({ g 2, g 3, g 4 }, i -> (await i) % 3 == 0)
 -- 1.00038s elapsed

o10 = 1
```

@d-torrance: I'd like to know whether `parallelApply`, which is more complicated, has specific benefits or guardrails that are missing from the first example above. I imagine if I had a massive example this might crash and burn, but short of examples like the typo in #3439, I haven't seen any drawbacks.

The next steps are:
- [ ] make sure this is stable when scaled up
- [ ] expand await to also handle other container types
- [ ] move async/await to keywords in the interpreter
- [ ] unreachable tasks should be cancelled (#3454), because currently in the all/any/position examples above, all branches are still working in the background even if very first one returned in such a way that the any/all command was done.


d-torrance:
This is great!

Is this ready to merge now, or should the TODO checklist be tacked first?

mahrud:
At the very least unreachable tasks scheduled with async should be killed. The others as much as possible.

mahrud:
Closing this since killing unreachable tasks in the current tasks architecture would require substantial rewrite.

PR #3649, opened 2025-02-09
Use lookup for print

d-torrance:
This is a slight refactoring of `print` that uses `lookup` instead of manually looping and calling `parent` at top level, which I noticed while playing around with the new profiler.

Cc: @pzinn

### DIFF
--- a/M2/Macaulay2/m2/expressions.m2
+++ b/M2/Macaulay2/m2/expressions.m2
@@ -1156,13 +1156,9 @@ texMath VectorExpression := v -> (
 
 -----------------------------------------------------------------------------
 print =  x -> (
-    c := class x;
-    while not c#?{topLevelMode,print} do (
-	if c === Thing then (<< net x << endl; return); -- default
-	c = parent c;
-	);
-    c#{topLevelMode,print} x
-    )
+    f := lookup({topLevelMode, print}, class x);
+    if f === null then << net x << endl -- default
+    else f x;)
 -----------------------------------------------------------------------------
 
 File << Thing := File => (o,x) -> printString(o,net x)


M2/Macaulay2/m2/expressions.m2
```diff
@@ -1156,13 +1156,8 @@ texMath VectorExpression := v -> (
...
 -----------------------------------------------------------------------------
 print =  x -> (
-    c := class x;
-    while not c#?{topLevelMode,print} do (
-	if c === Thing then (<< net x << endl; return); -- default
-	c = parent c;
-	);
-    c#{topLevelMode,print} x
-    )
+    if (f := lookup({topLevelMode, print}, class x)) =!= null then f x
```
mahrud, line 1159: any reason for not breaking this into two lines?

pzinn: incidentally, wasn't the operator `??` introduced for such situations?

d-torrance: Ooh, yeah -- it was!

d-torrance: I couldn't come up with a good readable version using `??`, but I did like the idea of moving then `then` statement to its own line.

M2/Macaulay2/m2/expressions.m2
```diff
@@ -1156,13 +1156,9 @@ texMath VectorExpression := v -> (
...
-    c := class x;
-    while not c#?{topLevelMode,print} do (
-	if c === Thing then (<< net x << endl; return); -- default
-	c = parent c;
-	);
-    c#{topLevelMode,print} x
-    )
+    if (f := lookup({topLevelMode, print}, class x)) =!= null
+    then f x
+    else << net x << endl;) -- default
```
mahrud, line 1161: This is what I meant about two lines:
```suggestion
    f := lookup({topLevelMode, print}, class x);
    if f === null then << net x << endl -- default
    else f x;)
```

mahrud (approved):

pzinn (approved):

PR #3650, opened 2025-02-09
[draft] ZeroMQ proof-of-concept

mahrud:
I added a proof-of-concept for using [ZeroMQ](http://wiki.zeromq.org/) in M2 using the [czmq](http://czmq.zeromq.org/) wrapper in C.

ZeroMQ is a library for messaging using tcp, inter-thread communication, inter-process communication, and "pragmatic general multicast". It supports several messaging [patterns](https://zguide.zeromq.org/docs/chapter2/#Messaging-Patterns) (push & pull, publisher & subscriber, request & reply, dealer & router, etc), all of which are **asynchronous and non-blocking**.

Here's a demo:
```m2
i1 : debug Core

i2 : -- send requests without blocking
     elapsedTime apply(5, i -> zmqopenwrite("random(ZZ^2, ZZ^2)"));
 -- .0000477s elapsed

i3 : -- instructions to read, evaluate, and reply
     worker = i -> zmqopenwrite toString(net i | " got " | net value zmqopenread(sleep i; "A"));

i4 : -- start workers without blocking
     elapsedTime apply(5, async worker);
 -- .000402361s elapsed

i5 : -- instructions to receive a reply
     reader = i -> zmqopenread("B");

i6 : -- wait to receive responses
     elapsedTime netList await apply(5, async reader)
 -- .0015996s elapsed

     +-------------+
o6 = |0 got | 1 4 ||
     |      | 3 3 ||
     +-------------+
     |0 got | 1 4 ||
     |      | 3 3 ||
     +-------------+
     |1 got | 8 8 ||
     |      | 3 7 ||
     +-------------+
     |2 got | 5 9 ||
     |      | 8 6 ||
     +-------------+
     |3 got | 9 0 ||
     |      | 9 8 ||
     +-------------+
```
(This is very rudimentary and has some race conditions in there because everything is using the same socket, but that's easy to fix. Also some weirdness is happening with two workers getting id "0" ... is the input "i" not threadlocal?!)

There are several immediate applications:
1. communicating with a Jupyter server
2. communicating with different M2 threads or processes without blocking
3. communicating with different processes that support ZMQ (e.g. GAP, python, julia, etc)

TODOs:
- [ ] make the class `file` wrap around `zsock_t` so we can use << in the top level
- [ ] handle closing/destroying/freeing zsocks and zstrs properly
- [ ] test zsock << matrix, and reading in a separate process
- [ ] based on some random crashes I saw, I think `value` is not thread safe!!

Once things are ironed out, it would be useful if you could setup threads with access to the same dictionary, then pass symbols to running threads without blocking them using zmq.


mahrud:
@d-torrance I imaging for lsp you needed a similar thing but for json-rpc, right? It would be nice to make sure both of these have the same API as `file` for open/write/read/close, but the `file` struct in the interpreter is getting massive and clunky. How did you implement yours?

d-torrance:
> How did you implement yours?

Just using `file` -- an `LSPServer` object is a hash table with a `file` as one of its values.

mahrud:
Shouldn't the file be read/written with json-rpc protocol?

PR #3654, opened 2025-02-11
Add lastError function

d-torrance:
This returns a sequence containing the location and message string for the last error (at least most of the time, e.g., not for syntax errors).

The eventual goal is to use this in the MPI package (see #2129) for handling errors gracefully, e.g., doing something like `try x else lastError()`.

### DIFF
--- a/M2/Macaulay2/d/parser.d
+++ b/M2/Macaulay2/d/parser.d
@@ -182,14 +182,13 @@ accumulate(e:ParseTree,file:TokenFile,prec:int,obeylines:bool):ParseTree := (
 	  );
      ret
      );
+makeParseError(token:Token, message:string):ParseTree := (
+     printErrorMessage(token, message);
+     errorTree);
 export errorunary(token1:Token,file:TokenFile,prec:int,obeylines:bool):ParseTree := (
-     printErrorMessage(token1,"syntax error at '" + token1.word.name + "'");
-     errorTree
-     );
+     makeParseError(token1,"syntax error at '" + token1.word.name + "'"));
 export errorbinary(lhs:ParseTree, token2:Token, file:TokenFile, prec:int,obeylines:bool):ParseTree := (
-     printErrorMessage(token2,"syntax error at '" + token2.word.name + "'");
-     errorTree
-     );
+     makeParseError(token2,"syntax error at '" + token2.word.name + "'"));
 export defaultunary(token1:Token,file:TokenFile,prec:int,obeylines:bool):ParseTree := (
      accumulate(ParseTree(token1),file,prec,obeylines)
      );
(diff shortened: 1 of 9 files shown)


M2/Macaulay2/tests/normal/error-messages.m2
```diff
@@ -1,63 +1,3 @@
...
-d(x,y,z)
-e(x,y,z)
-f(x,y,z)
-g(x,y,z)
-
-d(x,p1,h1)
-e(x,p1,h1)
-f(x,p1,h1)
-g(x,p1,h1)
+assert((last try 1/0 else lastError()) == "division by zero")
```
mahrud, line 1: Are you planning on replacing the manual tests above with automatic versions?

pzinn:
I did something similar on my dev branch. I'll compare the code.

M2/Macaulay2/d/expr.d
pzinn, line 1: not directly related, but do you know what these two lines are for:
https://github.com/Macaulay2/M2/blob/3fc8759c8a5666bce4ba6905468efa6dcc000fcd/M2/Macaulay2/d/expr.d#L307-L308

d-torrance: No idea!  They don't appeared to be used anywhere.

d-torrance:
> * I update the last error message when it's _printed_, not when it's built.

One of the big goals was to find out if there was an error without actually *raising* one, e.g., `try foo() else lastError()` would tell us what happened if `foo()` raised and error without actually raising an error.

That being said, @MichaelABurr has some code updating it during `printErrorMessage`.  See https://github.com/d-torrance/M2/pull/13.  Is that what you were doing, too?

> * I define a new global variable `lastError` (well, I didn't call it that...) rather than a function `lastError()`. On that last point, can you comment on why this is better?

That was our original idea, but we didn't realize it was possible.  Most of the global variables like this (e.g., `printWidth`) seem to be *set* at top-level and accessed in the interpreter.  Whereas this would be set in the interpreter and accessed at top-level.  We couldn't figure out an obvious way to do that, hence the function.

Is your code live on Macaulay2Web?  Or just in a branch somewhere?

pzinn:
good point. maybe I'm wrong about updating it when it's printed, as you said, the main goal is to test the error message with `try`. Yes it is live on M2Web, but it's not working any longer, I made this as the same time as my (failed) error upgrade exactly a year ago, and since then changes to the official development branch broke it; try `viewHelp "try"`, the example there doesn't work any more.

pzinn:
the relevant commit seems to be 74e022a7e5c2e541fdba21f51ddeffbd60a22027
yeah it's clumsy, it updates the error message when it's printed *or* when `try` is called (which I'm assuming is the part that got overwritten by a subsequent PR).

mahrud:
You're essentially trying to add a variable that remembers the value of `current` when an error occurs, right?

d-torrance:
I've rewritten this a bit so that `lastError` is a variable, like Paul's version, that we can clear by assigning null:

```m2
i1 : 1/0
stdio:1:1:(3): error: division by zero

i2 : lastError

o2 = (stdio:1:0-1:3, division by zero)

o2 : Sequence

i3 : lastError = null
```

M2/Macaulay2/d/parser.d
```diff
@@ -446,9 +441,8 @@ export unarytry(tryToken:Token,file:TokenFile,prec:int,obeylines:bool):ParseTree
 	  elseToken := gettoken(file,false);
 	  if elseToken == errorToken then return errorTree;
 	  if elseToken.word != elseW then (
-	       printErrorMessage(elseToken,"syntax error : expected 'else'");
-	       printErrorMessage(tryToken," ... to match this 'try'");
-	       return errorTree);
+	       makeParseError(elseToken,"syntax error : expected 'else'");
+	       return makeParseError(tryToken," ... to match this 'try'"));
```
pzinn, line 445: just curious, what's the meaning of these changes?

pzinn (approved):

MichaelABurr (approved):

PR #3655, opened 2025-02-11
Add language server package

d-torrance:
This is a (very very very) early draft of a `LanguageServer` package that implements Microsoft's [Language Server Protocol](https://microsoft.github.io/language-server-protocol/) (LSP), which is used by multiple editors (Emacs, VSCode, etc.)

### DIFF
--- a/M2/Macaulay2/packages/CMakeLists.txt
+++ b/M2/Macaulay2/packages/CMakeLists.txt
@@ -141,6 +141,7 @@ foreach(package IN LISTS PACKAGES)
     file(COPY ${package}  DESTINATION ${M2_DIST_PREFIX}/${M2_INSTALL_DATADIR})
   endif()
 
+
   ## Configuring the command strings
   _M2_STRING_CONFIGURE(M2_INSTALL_TEMPLATE M2_INSTALL_STRING TRUE)
   _M2_STRING_CONFIGURE(M2_CHECK_TEMPLATE   M2_CHECK_STRING)
@@ -252,3 +253,6 @@ foreach(package IN LISTS PACKAGES)
     DEPENDS M2-core
     )
 endforeach()
+
+file(COPY LanguageServer/M2-language-server
+     DESTINATION ${M2_DIST_PREFIX}/${M2_INSTALL_BINDIR})
--- a/M2/Macaulay2/packages/=distributed-packages
+++ b/M2/Macaulay2/packages/=distributed-packages
@@ -295,5 +295,7 @@ MRDI
 EliminationTemplates
 WittVectors
 Padic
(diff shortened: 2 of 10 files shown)


mahrud:
Is there a TODO list for this? Is the idea that LSP should fully replace the M2-mode on Emacs and extensions on Vim and VSCode? If so it would be good to have a list of M2-mode features that need to be reimplemented.

d-torrance:
A TODO list would be good.  For the time being, I was just focusing on getting the basic [lifecycle messages](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#lifeCycleMessages) working and hadn't put much thought into what actual features to implement.  So current TODO list:

- [ ] Get life cycle messages working
- [ ] Write a TODO list :)

I don't think LSP would replace M2-mode in Emacs, just supplement it.  That seems to be the case for the other major modes for languages that already have LSP support, like python-mode.

I'm also thinking of breaking the JSON-RPC stuff, which is essentially fully functional, into its own package.  It's not unique to LSP, and potentially might be useful for some other stuff.  It's apparently widely used in cryptocurrency applications, for example.

d-torrance:
I think this is ready to go.  In addition to the basic lifecycle stuff, it supports three methods:

* auto-completion
* hover documentation
* find definition

I'm proposing that a simple shell script, `M2-language-server`, which fires up an M2 process, loads this package, and starts a language server, be distrbuted alongside the `M2` binary.

I've written code for it to work in Emacs (https://github.com/Macaulay2/M2-emacs/pull/99), VS Code (https://github.com/Macaulay2/vscode-macaulay2/pull/9), Neovim (https://github.com/Macaulean/macaulay2.nvim/pull/4), and Jupyter (https://github.com/Macaulay2/Macaulay2-Jupyter-Kernel/pull/5).

## AI Disclosure

Claude wrote the first draft of the tests (which I ended up changing substantially) and pretty much all of the documentation.

d-torrance:
Discussed with Mike.  Will merge after fixing conflict

M2/Macaulay2/packages/CMakeLists.txt
```diff
@@ -141,6 +141,7 @@ foreach(package IN LISTS PACKAGES)
     file(COPY ${package}  DESTINATION ${M2_DIST_PREFIX}/${M2_INSTALL_DATADIR})
   endif()
 
+
```
mahrud, line 144: While removing the COPY call below, also delete this new line.

d-torrance: Sure thing -- not sure how this change snuck into this commit!

Issue #2882, opened 2023-06-18
Substituting matrices in polynomials


mahrud:
This issue is inspired by [this](https://math.stackexchange.com/questions/4250154/cayley-hamilton-in-macaulay2) Stack Exchange question that I randomly ran into.

Would be great if the following worked:
```m2
S = QQ[t]
m = random(ZZ^2, ZZ^2)
f = det(m - t*m^0) -- characteristic polynomial
-- By Cayley-Hamilton theorem we should have:
0 == f[m] -- error: expected a matrix with 1 row
0 == sub(f, t => m) -- error: expected substitution values to be ring elements or numbers
```
The easiest workaround I can think of is still too complicated:
```m2
0 == sum(pairs reverse flatten entries last coefficients(f, Variables => t), (i, c) -> m^i * c)
```
But more generally, if I have a polynomial in some number of variables, it shouldn't be too difficult to substitute matrices (or even ring maps) for the variables or return an error if the matrices are not compatible in some way.

Related: #2863, #2792

pzinn:
with univariate polynomials, I think this is a good idea and should be implemented.
not sure about multivariate polynomials, since the matrices you feed may not commute and then it's ambiguous how to take products.

mahrud:
I main had linear combinations of variables in mind, but you could use a multivariate free algebra instead.

PR #3657, opened 2025-02-15
Add RingElement \ VisibleList

d-torrance:
For evaluating a polynomial at multiple points at once:

```m2
i1 : R = QQ[x]

o1 = R

o1 : PolynomialRing

i2 : x^2 \ {1, 2, 3}

o2 = {1, 4, 9}

o2 : List
```

We don't define `VisibleList / RingElement` since it's already used for division.


M2/Macaulay2/m2/ringmap.m2
```diff
@@ -221,8 +221,9 @@ tensor(RingMap, Matrix) := Matrix => {} >> opts -> (f, m) -> (
     map(f ** target m, f ** source m, f cover m))
 RingMap ** Matrix := Matrix => (f, m) -> tensor(f, m)
 
-VisibleList / RingMap := VisibleList => (v,f) -> apply(v,x -> f x)
-RingMap \ VisibleList := VisibleList => (f,v) -> apply(v,x -> f x)
+VisibleList / RingMap     := VisibleList => (v,f) -> apply(v,x -> f x)
+RingMap     \ VisibleList :=
+RingElement \ VisibleList := VisibleList => (f,v) -> apply(v,x -> f x)
```
mahrud, line 226: This _looks_ like two similar methods, but really they are completely different:
1. `RingMap \ List` evaluates a single ring map with multiple inputs
2. `RingElement \ List` creates multiple ring maps and evaluates each with the same input.

Besides this conflation, on the surface this is a convenient improvement, but underneath this is bad for efficiency because creating throwaway new ring maps using `substitute` is slow and shouldn't be made more convenient.

I'm not in favor of this change.

mahrud (changes requested):

M2/Macaulay2/m2/ringmap.m2
d-torrance: Ah, good point.  Are you suggesting that we scrap this idea entirely, or just re-implement by creating a single ring map?

mahrud: I don't know how to do this with a single ring map, but `(x -> x^2) \ {1, 2, 3}` is a simple enough workaround that I would close this PR.

To be honest, I've found #2863 convenient, but also found it to be a surprising bottleneck several times and regretted it. It's plausible that we can come back to this PR when `sub` is significantly more efficient though, along with other useful features like #2882.

d-torrance: Fair enough -- I'll close this

PR #3658, opened 2025-02-15
Update copyright notices

d-torrance:
As discussed in recent M2internals meetings, we update the main copyright notices to indicate that the copyright is held by the "Macaulay2 Authors" rather than just Dan & Mike.

We also:
* Remove the old, largely out-of-date M2/Macaulay2/README.in, which generated `/usr/share/doc/Macaulay2/README`.  This file was probably mostly unnoticed.  Who would look there?
* Add a copyright notice to the main `README.md` in the root directory of the source tree.
* Markdown-ify M2/LICENSING-NOTES (now M2/LICENSING-NOTES.md) to use tables with links to each of the various libraries and programs.  We also add some that were missing and use short identifiers from the [SPDX License List](https://spdx.org/licenses/) for each of the licenses.
* Add a list of who actually comprises the "Macaulay2 Authors".  The list consists of everyone with at least one commit to the d, e, and/or m2 directories, plus a link to the "authors of Macaulay2 packages" documentation page.
* Bump the latest copyright year to 2025.


M2/LICENSING-NOTES.md
```diff
@@ -0,0 +1,122 @@
...
+| -lmpfi             | [MPFI](https://gitlab.inria.fr/mpfi/mpfi)                                    | LGPL-2.1-or-later                     |
+| -lmpfr             | [MPFR](https://www.mpfr.org/)                                                | LGPL-3.0-or-later                     |
+| -lmps              | [MPSolve](https://numpi.dm.unipi.it/scientific-computing-libraries/mpsolve/) | GPL-3.0-or-later                      |
+| -lnormaliz         | [Normaliz](https://www.normaliz.uni-osnabrueck.de/)                          | GPL-3.0-or-later                      |
+| -lntl              | [NTL](https://libntl.org/)                                                   | LGPL-2.1-or-later                     |
+| -lpython3          | [Python](https://www.python.org/)                                            | PSF-2.0                               |
+| -lreadline         | [Readline](https://www.gnu.org/software/readline)                            | GPL-3.0-or-later                      |
+| -ltbb              | [oneTBB](https://uxlfoundation.github.io/oneTBB/)                            | Apache-2.0                            |
+
+Programs we include with our distribution:
```
mahrud, line 56: Do we?

M2/LICENSING-NOTES.md
```diff
@@ -0,0 +1,122 @@
...
+| [TOPCOM](https://www.wm.uni-bayreuth.de/de/team/rambau_joerg/TOPCOM/) | GPL-2.0-or-later |
+
+
+Libraries linked with some of our programs:
+|        |                                            |                  |
+| ------ | ------------------------------------------ | ---------------- |
+| -lcdd  | [cddlib](https://github.com/cddlib/cddlib) | GPL-2.0-or-later |
+| -lglpk | [GLPK](https://www.gnu.org/software/glpk/) | GPL-3.0-or-later |
+
+Code by other authors compiled as part of M2:
```
mahrud, line 75: Why not add Zinin, Furnish, and Boyd to the list below and remove this section?

M2/LICENSING-NOTES.md
d-torrance: Potentially.  I suppose this should be something like "... we may include ..."

M2/LICENSING-NOTES.md
d-torrance: Makes sense!

M2/LICENSING-NOTES.md
mahrud: Which distribution includes these? A build target that bundles these programs is not the same as a distribution.

d-torrance: The RHEL rpms include most (maybe all) of them, since so few are available in the official repos.

d-torrance:
I've updated this based on where I think we landed after this afternoon's M2internals.  In particular:

* LICENSING-NOTES: gone
* Licenses package: gone
* List of authors: now on the [wiki](https://github.com/Macaulay2/M2/wiki/The-Macaulay2-Authors)

Cc: @mikestillman, @mahrud, @antonleykin, @pzinn, @dswinarski, @rohansjoshi: Any thoughts?

PR #3659, opened 2025-02-16
Properly display degrees in mathML(MatrixExpression)

d-torrance:
@mikestillman - this fixes the matrix display bug we were seeing in Jupyter in texmacs mode.

### Before
image

### After
image

We also replace the *'s with ·'s when displaying products:
image

### DIFF
--- a/M2/Macaulay2/m2/mathml.m2
+++ b/M2/Macaulay2/m2/mathml.m2
@@ -110,32 +110,31 @@ mathML Sum := v -> (
 		    	 else mathML v#i),
 		    n:newline),
 	       "</mrow>",newline)))
+
 mathML Product := v -> (
-     n := # v;
-     if n === 0 then "<mn>1</mn>"
-     else if n === 1 then mathML v#0
-     else (
-     	  p := precedence v;
-	  seps := newClass(MutableList, (n+1) : "<mo>*</mo>");
-	  if n>1 and isNumber v#0 and startsWithSymbol v#1 then seps#1 = "<mo></mo>";
-     	  boxes := apply(#v,
-	       i -> (
-		    term := v#i;
-		    nterm := mathML term;
-	       	    if precedence term <= p then (
-			 seps#i = seps#(i+1) = "<mo></mo>";
-			 nterm = ("<mrow><mo>(</mo>", nterm, "<mo>)</mo></mrow>");
-			 );
-		    if class term === Power
-		    and not (term#1 === 1 or term#1 === ONE)
-		    or class term === Subscript then (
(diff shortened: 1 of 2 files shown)


mikestillman:
@d-torrance Thanks for fixing this!  

In displaying polynomials, I was wondering if perhaps we would prefer to not print an asterisk or a dot?  What do you think?

Sometimes there is no dot, sometimes there is.  I think it has to do with whether there is an exponent (so something like x * y^2, but x^2 y).  I will look at the output more carefully to see what seems preferable.

mikestillman:
I can either accept this PR as is, and we can work on the spacing later, or, if you have ideas, we can make those changes now.

M2/Macaulay2/m2/mathml.m2
```diff
@@ -59,7 +59,29 @@ mathML FunctionApplication := m -> (
      then concatenate (mfun, bigParenthesize margs)
      else concatenate (bigParenthesize mfun, bigParenthesize margs)
      )
-mathML MatrixExpression := x -> concatenate( "<mrow><mo>(</mo>", mtableML x, "<mo>)</mo></mrow>", newline )
+
+-- zero-width "phantom" element to make sure that degrees line up
+mphantom := x -> concatenate(
```
mikestillman, line 64: We don't need to do this now, but fixing the spacing between terms (currently the + or - spacing is less than inside a monomial.  Also, sometimes the dot (or previously, asterisk) is included, sometimes not).  See conversation for an example...

mikestillman (commented):
I can accept this as is.  We can play with the spacing mentioned in the review either for this PR or later.

M2/Macaulay2/m2/mathml.m2
d-torrance: I think the big problem with how we represent these `Product` objects (not just by `mathML`, but in `net(Product)` too) is not knowing what the factors are.  We could totally represent `Product{x, y}` by `xy`, but we want `Product{2,3}` to be `2 * 3` and not `23`.

d-torrance:
@mikestillman -- I've updated it so we only use &middot; in `factor(ZZ)`, i.e., immediately after a `Power` object with an exponent of 1:

image

d-torrance:
I changed my mind and switched `mathML(Product)` so that it's essentially identical to `texMath(Product)` for consistency:

image

mikestillman:
Thanks, this latest version displays polynomials in texmacs mode in jupyter much better!

mikestillman (approved):
This looks good to me, thanks!

PR #3662, opened 2025-02-19
Move grammar generation to Style package

d-torrance:
As suggested by @mahrud in https://github.com/Macaulay2/M2/pull/3298#issuecomment-2183065149), we move the `generateGrammar` function from a build script to a method exported by the `Style` packagee.

### DIFF
--- a/M2/Macaulay2/packages/SimpleDoc.m2
+++ b/M2/Macaulay2/packages/SimpleDoc.m2
@@ -66,6 +66,8 @@ applySplit = (functionTable, textlines) -> apply(splitByIndent(textlines, false)
 	functionTable#key(textlines_{s+1..e}, getLinenum textlines#s)))
 
 -- Mapping tables for evaluating docstring keywords
+-- When adding a new keyword that isn't already an exported symbol,
+-- add it to the list of constants for syntax highlighting in Style.m2
 NodeFunctions = new HashTable from {
     "Node"            => (textlines, keylinenum) -> new Node from nodeCheck(applySplit(NodeFunctions, textlines), keylinenum),
     "Key"             => (textlines, keylinenum) -> Key             => getKeys(textlines, keylinenum),
--- a/M2/Macaulay2/packages/Style.m2
+++ b/M2/Macaulay2/packages/Style.m2
@@ -60,9 +60,20 @@ generateSymbols = () -> (
     cachedSymbols.Type     = first \ select(symbols, isType);
     cachedSymbols.Function = first \ select(symbols, isFunction);
     cachedSymbols.Constant = sort join(
-	first \ select(symbols, isConst),
-	{"Node", "Item", "Example", "CannedExample", "Pre", "Code", "Tree",
-	    "Synopsis"}); -- SimpleDoc words
+	first \ select(symbols, isConst), {
+	    -- SimpleDoc docstring keywords -- keep updated
+	    -- NodeFunctions
+	    "Node",
+	    "Synopsis",
+	    -- DescriptionFunctions
+	    "CannedExample",
+	    "Code",
+	    "Example",
+	    "Pre",
+	    "Tree",
+	    -- ConsequenceFunctions
+	    "Item"
+	    });
     )
 
 -------------------------------------------------------------------------------
--- a/M2/Macaulay2/packages/Style.m2
+++ b/M2/Macaulay2/packages/Style.m2
@@ -52,15 +52,6 @@ generateSymbols = () -> (
 
     if length symbols < 1500 then error "expected more entries for M2-symbols";
 
-    -- Check for invalid symbols
-    bad := select(symbols, (name, symb) -> not okay(name, symb));
-    if #bad > 0 then error(
-	"encountered symbol(s) that are not alphanumeric or have less than 2 characters:",
-	concatenate apply(bad, (name, symb) -> {
-		"\n\t",
-		-* TODO: symbolLocation name, ": here is the first use of ", *-
-		toString name}));
-
     -------------------------------------------------------------------------------
     -- Put the symbols into bins
     cachedSymbols.Symbol   = first \ select(symbols,
--- a/M2/Macaulay2/packages/Style.m2
+++ b/M2/Macaulay2/packages/Style.m2
@@ -3,16 +3,184 @@ newPackage( "Style",
      AuxiliaryFiles => true,
      Headline => "style sheets and images for the documentation",
      Keywords => {"Documentation"},
-     Version => "1.0"
+     Version => "1.1"
      )
+
+---------------------------------------------
+-- grammar generation code adapted from    --
+-- M2/Macaulay2/editors/make-M2-symbols.m2 --
+---------------------------------------------
+
+export {"generateGrammar"}
+
+importFrom(Core, "sortBy")
+
+-- TODO: Move these two elsewhere:
+Function and Function := (f, g) -> s -> f s and g s
+Function or  Function := (f, g) -> s -> f s or  g s
+
+is := X -> (name, symb) -> instance(value symb, X)
+
+isAlpha        := s -> match("^[[:alpha:]]+$", s)
+isAlphaNumeric := s -> match("^[[:alnum:]]+$", s)
+
+-- Things we want to highlight:
+isType     := is Type
+isKeyword  := is Keyword
+isFunction := is Function
+isConst    := (name, symb) -> (isAlphaNumeric name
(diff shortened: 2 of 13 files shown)


mahrud:
I probably don't have time to review this in the short term.

M2/Macaulay2/packages/Style.m2
```diff
@@ -3,16 +3,184 @@ newPackage( "Style",
...
+-- M2/Macaulay2/editors/make-M2-symbols.m2 --
+---------------------------------------------
+
+export {"generateGrammar"}
+
+importFrom(Core, "sortBy")
+
+-- TODO: Move these two elsewhere:
+Function and Function := (f, g) -> s -> f s and g s
+Function or  Function := (f, g) -> s -> f s or  g s
```
mahrud, line 20: Do you want to actually move these to Core while you're moving them? Maybe also add `not Function` and `Function xor Function`. Especially the former would be super useful for things like this:
```m2
i1 : not Function := f -> x -> not f x

i2 : select({0,1,0,2,0,3}, not zero)

o2 = {1, 2, 3}
```

d-torrance: Might as well!

M2/Macaulay2/packages/Style.m2
```diff
@@ -3,16 +3,184 @@ newPackage( "Style",
...
+    if length symbols < 1500 then error "expected more entries for M2-symbols";
+
+    -- Check for invalid symbols
+    bad := select(symbols, (name, symb) -> not okay(name, symb));
+    if #bad > 0 then error(
+	"encountered symbol(s) that are not alphanumeric or have less than 2 characters:",
+	concatenate apply(bad, (name, symb) -> {
+		"\n\t",
+		-* TODO: symbolLocation name, ": here is the first use of ", *-
+		toString name}));
```
mahrud, line 66: I recently learned something super cool about how `error` prints symbols which I didn't know when I wrote this:
```m2
i7 : error splice("encountered symbol(s) that are not alphanumeric or have less than 2 characters:",
     toSequence between_" " {symbol Module, symbol print})
stdio:13:5:(3): error: encountered symbol(s) that are not alphanumeric or have less than 2 characters:'Module' 'print'
../linuxbrew/.linuxbrew/share/Macaulay2/Core/expressions.m2:1234:42-1234:48: here is the first use of 'Module'
../linuxbrew/.linuxbrew/share/Macaulay2/Core/expressions.m2:1158:0-1158:5: here is the first use of 'print'
```

M2/Macaulay2/packages/Style.m2
```diff
@@ -3,16 +3,184 @@ newPackage( "Style",
...
+    -------------------------------------------------------------------------------
+    -- Put the symbols into bins
+    cachedSymbols.Symbol   = first \ select(symbols,
+	(name, symb) ->isAlphaNumeric name);
+    cachedSymbols.Keyword  = first \ select(symbols, isKeyword);
+    cachedSymbols.Type     = first \ select(symbols, isType);
+    cachedSymbols.Function = first \ select(symbols, isFunction);
+    cachedSymbols.Constant = sort join(
+	first \ select(symbols, isConst),
+	{"Node", "Item", "Example", "CannedExample", "Pre", "Code", "Tree",
```
mahrud, line 77: Do you mind adding a comment close to this, and also somewhere in SimpleDoc, to keep these in sync?

mahrud (approved):
I haven't checked this myself, but if the output is identical, this seems fine.
I left a few comments that would be good to implement, though I realize that those are just things you moved, not added.

M2/Macaulay2/packages/Style.m2
d-torrance: Is there a reason we hardcode these here instead of just exporting them from `SimpleDoc` and letting them get pickedup with all the other symbols?  (I suppose `Node` is a class, so it would get a different color -- is that the main reason?)

M2/Macaulay2/packages/Style.m2
d-torrance: That's neat!

I'm a little confused at this whole error message, though.  The `symbols` list appears to be made up of two things: the list of all packages and everything from the dictionaries in the pre-loaded packages.  We already call `select` with `okay` on the latter, so we know that none of those will end up in `bad`.  Why not just do the same thing on the former?

And also, these packages are always paired up with the `Core` symbol (I guess so that when we call `is` on them, we know that they're packages), so even if one of them ends up in `bad` and we modify the error message to print the symbol location, it will just be the location of `Core`.

mahrud: I'm not sure. Maybe experiment with not selecting on line 51 and see if anything breaks?

d-torrance: There are *tons* of non-alphanumeric symbols (mathematical operators, all the `Package$foo` ones) that were getting filtered out by `select`.

The only elements of this list we weren't selecting out with `okay` are the package names, and we know that they'll be alphanumeric since `newPackage` raises an error when they're not.  I just pushed a commit removing this check.

M2/Macaulay2/packages/Style.m2
d-torrance: I just pushed a commit exporting these docstring symbols from `SimpleDoc` so we can stop hard-coding them here.  I also renamed the `Node` class to `DocNode` so `Node` would still be the same color as the others.

M2/Macaulay2/packages/SimpleDoc.m2
```diff
@@ -17,16 +17,39 @@ newPackage(
...
+
+    -- docstring symbols
+    "CannedExample",
+    "Code",
+    "Example",
+    "Item",
+    "Node",
+    "Pre",
+    "Synopsis",
+    "Tree",
```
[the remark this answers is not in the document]
d-torrance: And case in point, we got the following build failure from the `Varieties::tangentSheaf(ProjectiveVariety)` example :)

```m2
i5 : Node = Proj(QQ[a,b,c]/ideal(b^2*c-a^2*(a+c)))
stdio:5:11:(3): error: assignment to protected global variable 'Node', originally defined at stdio:5:0:(3):
```
I'll restore the original behavior but add some comments.  The annoying this is that many of the SimpleDoc docstring symbols are in fact symbols defined in Core, but not these few.

d-torrance:
I've rebased this on top of `development` since there were some conflicts with the reason Macaulay2Doc updates.  I also updated the boolean operator docs.

M2/Macaulay2/packages/Style.m2
```diff
@@ -3,16 +3,182 @@ newPackage( "Style",
...
+	    -- ConsequenceFunctions
+	    "Item"
+	    });
+    )
+
+-------------------------------------------------------------------------------
+-- Generate syntax files from templates in the same directory
+
+generateGrammar = method()
+generateGrammar(String, String, Function) := (template, outfile, demarkf) -> (
```
mahrud, line 83: I wonder if it would make more sense to replace `demarkf` with `tolistf`, which also takes care of how line wrapping long lists of words and how to enclose the list (e.g. braces, brackets, or nothing at all.

M2/Macaulay2/packages/Style.m2
```diff
@@ -3,16 +3,182 @@ newPackage( "Style",
...
+    if not fileExists template
+    then (
+	printerr("warning: ", template,
+	    " does not exist; skipping generation of ", outfile);
+	return);
+    printerr("generating ", outfile);
+    directory := replace("/[^/].*$", "", outfile);
+    if not isDirectory directory then makeDirectory directory;
+    output := get template;
+    if #cachedSymbols == 0 then generateSymbols();
```
mahrud, line 93: It would be great if there was a version of `generateSymbols` and `generateGrammar` that accepted a package and listed symbols available after that package (and all its dependencies) have been loaded. This would be used for per-package syntax highlighting.

mahrud (approved):
Happy to accept this as is, but added to more suggestions.

M2/Macaulay2/packages/Style.m2
d-torrance: This is a great idea!  I think it would be good for another PR, though.  I'm also not entirely sure what the behavior should be.  If a user specifies a particular package, should we generate symbols for:

* just the package
* the package and its dependencies
* the package, its dependencies, plus Core and the pre-loaded packages

Also, if we run `generateGrammar` a couple times in success with different packages, should it keep the symbols from the earlier runs cached, or wipe the cache and start fresh?

PR #3664, opened 2025-02-20

n-m-g:
A number of improvements -- cleaned up code to make it more readable, expanded documentation to make it more useable and added some additional tests.

### DIFF
--- a/M2/Macaulay2/packages/AbstractSimplicialComplexes.m2
+++ b/M2/Macaulay2/packages/AbstractSimplicialComplexes.m2
@@ -1,6 +1,6 @@
 -- -*- coding: utf-8 -*-
---------------------------------------------------------------------------------
--- Copyright 2024  Nathan Grieve
+---------------------------------------------------------------------------------
+-- Copyright 2025  Nathan Grieve
 --
 -- This program is free software: you can redistribute it and/or modify it under
 -- the terms of the GNU General Public License as published by the Free Software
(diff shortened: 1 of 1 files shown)


M2/Macaulay2/packages/AbstractSimplicialComplexes.m2
```diff
@@ -329,117 +288,172 @@ simplicialMakeMatrix(List,List) := (kPlusOneFaces,kFaces) -> (
...
+        AbstractSimplicialComplexes
+    Headline 
+        a package for working with abstract simplicial complexes
+    Description
+        Text
+	    The package AbstractSimplicialComplexes provides a methodology for 
+            working with abstract simplicial complexes.  The starting point is to represent each
+	    abstract simplicial complex as a graded list.  In this regard our approach differs
+	    fundamentally from that of the package
+	    @HREF("https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/SimplicialComplexes/html/index.html","SimplicialComplexes.m2")@.
```
mahrud, line 397: You should link to another package with `@TO "SimplicialComplexes::SimplicialComplexes"@`.

n-m-g:
Hi,

Many thanks for this comment.  The reason for pointing to the HREF instead of TO is that the internal documentation links never seem to work on my own machine.  (Yes, this means that the M@ documentation didn’t get completely set-up correctly on my machine when installing via the home-brew method.). But, I am happy to make the change if you insist.  Please let me know.

Best,
Nathan

n-m-g:
I've updated the file and created a new pull request.  Many thanks for these comments.

PR #3665, opened 2025-02-20

n-m-g:
Display math symbols now added to the documentation.  Note, this doesn't display well in the Safari browser but it does in the Google Chrome Browser.

### DIFF
--- a/M2/Macaulay2/packages/AbstractSimplicialComplexes.m2
+++ b/M2/Macaulay2/packages/AbstractSimplicialComplexes.m2
@@ -1,6 +1,6 @@
 -- -*- coding: utf-8 -*-
---------------------------------------------------------------------------------
--- Copyright 2024  Nathan Grieve
+---------------------------------------------------------------------------------
+-- Copyright 2025  Nathan Grieve
 --
 -- This program is free software: you can redistribute it and/or modify it under
 -- the terms of the GNU General Public License as published by the Free Software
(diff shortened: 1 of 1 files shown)


d-torrance:
For future reference, there's no need to close a pull request and open a new one when adding a feature. Just push any new commits to the same branch and the pull request will be updated automatically.

M2/Macaulay2/packages/AbstractSimplicialComplexes.m2
```diff
@@ -71,236 +76,194 @@ AbstractSimplicialComplex.GlobalAssignHook = globalAssignFunction
...
-
 new AbstractSimplicialComplex := AbstractSimplicialComplex =>(cl) -> (
-    K := newClass(AbstractSimplicialComplex, new HashTable); -- sigh
-    K)
+     K := newClass(AbstractSimplicialComplex, new HashTable); 
+     K)
 
 spots AbstractSimplicialComplex := List => (
-  K -> sort select(keys K, i -> class i === ZZ))
+     K -> sort select(keys K, i -> class i === ZZ))
```
d-torrance, line 84: This call to `select` can be simplified to just `select(keys K, ZZ)`

M2/Macaulay2/packages/AbstractSimplicialComplexes.m2
```diff
@@ -71,236 +76,194 @@ AbstractSimplicialComplex.GlobalAssignHook = globalAssignFunction
...
 
--- return the p-faces of a simplicial complex
+-- Return the p-faces of a simplicial complex.
 
 AbstractSimplicialComplex _ ZZ := AbstractSimplicialComplex => (K,p) -> (
-  if K#?p then K#p 
-  )
+     if K#?p then K#p
+     else {})
```
d-torrance, line 90: This can be simplified using the "null coalescing operator" `??`, i.e., `K#p ?? {}`.

M2/Macaulay2/packages/AbstractSimplicialComplexes.m2
```diff
@@ -71,236 +76,194 @@ AbstractSimplicialComplex.GlobalAssignHook = globalAssignFunction
...
-myList := select(L,i -> isSubset(x,i));
-if #myList == 1 then
-return true
-else return false
-    )
+isMaximal := (x,L) -> (
+     myList := select(L,i -> isSubset(x,i));
+     if #myList == 1 then
+     return true
+     else return false)
```
d-torrance, line 98: Explicitly returning `true` or `false` here is redundant -- the correct one is already the value of the expression `#myList == 1`.  You also don't need to use `return` on the last line of a function.  So this coupld be simplified to just ending with `#myList == 1`.

M2/Macaulay2/packages/AbstractSimplicialComplexes.m2
```diff
@@ -71,236 +76,194 @@ AbstractSimplicialComplex.GlobalAssignHook = globalAssignFunction
...
 
 AbstractSimplicialComplex == AbstractSimplicialComplex := Boolean => (K,L) ->(
-    return (abstractSimplicialComplexFacets K) == (abstractSimplicialComplexFacets L)
-    )
-
-
---- returns the dimension of a simplicial complex
+     facetsK := abstractSimplicialComplexFacets K;
+     facetsL := abstractSimplicialComplexFacets L;
+     return facetsK == facetsL)
```
d-torrance, line 118: This `return` can be dropped (as can several others in other functions).

M2/Macaulay2/packages/AbstractSimplicialComplexes.m2
```diff
@@ -71,236 +76,194 @@ AbstractSimplicialComplex.GlobalAssignHook = globalAssignFunction
...
+-- Make a random subset of {1,...,n}.
 
 randomSubset = method()
 
 randomSubset(ZZ) := List => (n) -> (
-   setRandomSeed(currentTime());
-   k := random(1,n); 
-   sort unique (for i from 1 to k list (random(1,n)))
-    )
+     setRandomSeed(currentTime()); -- without doing this we will obtain very biased output
```
d-torrance, line 169: I wouldn't call `setRandomSeed` in a function like this.  Let the user call it prior to calling if they'd like.  Otherwise, there's no way of reproducing results.  (The same comment holds for several other functions below.)

M2/Macaulay2/packages/AbstractSimplicialComplexes.m2
```diff
@@ -71,236 +76,194 @@ AbstractSimplicialComplex.GlobalAssignHook = globalAssignFunction
...
-    setRandomSeed(currentTime());
-    n := #L;
-    k := random(0,n);
-    mySubset := subsets(L,k);
-    mySubset_(random(binomial(n,k)))
-    )
+     setRandomSeed(currentTime());  -- without doing this we will obtain very biased output
+     n := #L;
+     k := random(0,n);
+     mySubset := subsets(L,k);
```
d-torrance, line 185: This seems very inefficient, generating *all* the subsets and then randomly picking one.  You could use something like Knuth's Algorithm S instead.

d-torrance (changes requested):

n-m-g:
Hi,

Many thanks for this.  I will consider implementing these suggestions.

About the random seed, yes to allow for reproducibility, perhaps creating an option for those methods is the best thing to do.  Let me think a bit more about that.

Best,
Nathan

M2/Macaulay2/packages/AbstractSimplicialComplexes.m2
n-m-g: Hi,

About the Algorithm S, it seems what I'm doing already is actually not so different; one still needs to "make the sample space"  (compare with Algorithm S (Selection sampling technique) on p. 142 https://seriouscomputerist.atariverse.com/media/pdf/book/Art%20of%20Computer%20Programming%20-%20Volume%202%20(Seminumerical%20Algorithms).pdf).  Similar considerations might apply to my random seed comment above.  But, I'll think a bit more about this and at least it will be worthwhile to add a reference to Knuth's book.  Many thanks again for these comments.

Best,
Nathan

d-torrance:
I don't see any updates.  Did you push the changes to this branch (`patch-5`)?

n-m-g:
Hi,

Yes, I attempted to change the file directly but I didn’t issue a pull request.  Let me try that again.

Best,
Nathan

PR #3667, opened 2025-02-20

n-m-g:
Various cosmetic updates.


d-torrance:
Again, for future reference, it's unnecessary to open a new pull request.  Just push your changes to the same branch.  The first PR was on your `patch-4` branch, the next was `patch-5`, and now this is `patch-6`.  If everything had stayed on `patch-4`, then we could have kept the original PR. :)

n-m-g:
Dear Doug,

My apologies for this!  However, I am just editing the file as you suggested.  If we can undo everything and then push to the original, that would be fine.  Do keep me posted.  

Best,
Nathan

n-m-g:
Dear Doug,

I seem to have found the edit the existing pull request option again.  I just updated the file and it should install correctly.  Do let me know if there are any additional install issues!  I likely will make some additional changes to the file later today!

Best,
Nathan

n-m-g:
Dear Doug,

I have implemented options for the random seed now and have pushed these to the “patch-6-branch”.

I suppose we can reopen the “patch-4-branch” if needed, push these changes there and close the existing one if you’d like.

Apologies for all of this confusion on my end!

Best,
Nathan

n-m-g:
Dear Doug,

Many thanks again for your comments.  The current version of the package will remain unchanged for now.  I'll be in touch in case that further changes are made in due time.  Many thanks again to yourself and Mahrud for your feedback.

Best,
Nathan

d-torrance:
There are still some examples and tests that don't use `Seed => false`, so they won't be reproducible.

I still don't think that any methods should be calling `setRandomSeed` behind the scenes.  That function has a global side effect, and it should only be called when the user wants to call it.

n-m-g:
Dear Doug,

Yes, the intent was to include some random examples in the documentation.  Even with random seed => false, of course these methods still have some random features.  In either case, the user can attempt to reproduce the output as they wish.  Again, I’ll be in touch in case additional changes are made/requested.  Many thanks again for these comments.

Best,
Nathan

d-torrance:
I'll go ahead and merge this since the current behavior of the package is already to re-set the random seed when calling the various random methods.  But I'd strongly consider removing this feature in future updates.  Calling `setRandomSeed` changes things globally, and it's not good practice to do this silently inside a function.

Also, thinking out loud, `randomSubset` is a useful enough general function that it probably should be in Core.  This isn't the only package that has an implementation of it.

n-m-g:
Dear Doug,

This all sounds fine.  Yes, the random seed needs to be set each time prior to creating random numbers since otherwise one gets a biased output.  So, yes this is something that can be improved in the core.  I am creating these methods for a specific purpose and they are working as intended for that purpose.

Best,
Nathan

PR #3669, opened 2025-02-22
Add "vector" constructor methods taking ring as input

d-torrance:
Currently, we can do the following for matrices and it works as expected:

```m2
i1 : matrix(QQ, {{1, 2}, {3, 4}})

o1 = | 1 2 |
     | 3 4 |

              2       2
o1 : Matrix QQ  <-- QQ
```

But it doesn't work for vectors:

```m2
i2 : vector(QQ, {1, 2, 3})
stdio:2:6:(3): error: no method found for applying vector to:
     argument 1 :  QQ (of class Ring)
     argument 2 :  {1, 2, 3} (of class List)
```

We add this functionality.  In particular, given a `Ring` (or `RingFamily`, so we catch `RR`, `CC`, and `RRi`) and a matrix, list, number, or ring element, we construct an element of the appropriate free module over that ring:

```m2
i1 : vector(QQ, {1, 2, 3})

o1 = | 1 |
     | 2 |
     | 3 |

       3
o1 : QQ
```

### DIFF
--- a/M2/Macaulay2/m2/modules.m2
+++ b/M2/Macaulay2/m2/modules.m2
@@ -172,6 +172,14 @@ new Module from Sequence := (Module,x) -> (
 vector(Module, Matrix) := (M, f) -> vector map(M,,entries f)
 vector(Module, List)   := (M, v) -> vector map(M,,apply(splice v, i -> {i}))
 vector(Module, RingElement) := vector(Module, Number) := (M, x) -> vector(M, {x})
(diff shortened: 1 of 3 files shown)


mahrud (approved):
I had convinced myself that `vector(QQ^3, {1, 2, 3})` works and it's just as good as `vector(QQ, {1, 2, 3})`, but I see the appeal.

PR #3670, opened 2025-02-23
Update JSON package to version 0.4

d-torrance:
We ~~export~~ *remove* the `JSONEncoder` class so that developers can write their own `toJSON` methods.  Thanks @mahrud for the suggestion!

### DIFF
--- a/M2/Macaulay2/packages/JSON.m2
+++ b/M2/Macaulay2/packages/JSON.m2
@@ -365,9 +365,9 @@ outfile = openOut(outdir | "/test-parse.m2")
 outfile << commentize  copyrightBanner << endl
 for tst in sort select(tsts, f -> match("^y_", f)) do (
     outfile << endl << commentize tst << endl;
-    json = get(testdir | "/" | tst);
+    testjson = get(testdir | "/" | tst);
     outfile << "assert BinaryOperation(symbol ===, fromJSON " <<
-    format json << ", " << toExternalString fromJSON json << ")" << endl)
+    format testjson << ", " << toExternalString fromJSON testjson << ")" << endl)
 close outfile
 ///
 
--- a/M2/Macaulay2/packages/JSON.m2
+++ b/M2/Macaulay2/packages/JSON.m2
@@ -85,9 +85,13 @@ unescapedP = Parser(c -> if c === null then null else (
 	if x < 0x20 or x == 0x22 or x == 0x5c or x > 0x10ffff
 	then null
 	else terminalParser c))
-charP = unescapedP | ("\\" @
+deformat = x -> (
+    if last x === "/" then "/"
+    else value concatenate("\"", x, "\""))
+escapedP = deformat % ("\\" @
     orP("\"", "\\", "/", "b", "f", "n", "r", "t",
 	andP("u", hexDigitP, hexDigitP, hexDigitP, hexDigitP)))
+charP = unescapedP | escapedP
 stringP = ((l, x, r) -> concatenate x) % andP("\"", *charP, "\"")
 
 -- objects
--- a/M2/Macaulay2/packages/JSON.m2
+++ b/M2/Macaulay2/packages/JSON.m2
@@ -129,7 +129,9 @@ fromJSON File   := fromJSON @@ get
 -- encoding --
 --------------
 
-toJSON = method(Options => {
+toJSON = method(
+    Dispatch => Thing,
+    Options => {
 	Indent         => null,
 	IndentLevel    => 0,
 	NameSeparator  => ": ",
--- a/M2/Macaulay2/packages/JSON.m2
+++ b/M2/Macaulay2/packages/JSON.m2
@@ -1,8 +1,8 @@
 newPackage(
     "JSON",
     Headline => "JSON encoding and decoding",
-    Version => "0.3",
-    Date => "September 14, 2024",
+    Version => "0.4",
+    Date => "February 24, 2025",
     Authors => {{
 	    Name => "Doug Torrance",
 	    Email => "dtorrance@piedmont.edu",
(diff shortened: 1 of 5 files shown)


M2/Macaulay2/packages/JSON.m2
```diff
@@ -275,6 +290,12 @@ doc ///
       option.
     Example
       toJSON(hashTable{"foo" => 1, "bar" => 2, "baz" => 3}, Sort => true)
+    Text
+      Developers interested in writing a @CODE "toJSON"@ method for a new class
+      @CODE "X"@ should write a @CODE "toJSON(JSONEncoder, X)"@ method.
```
mahrud, line 295: I'm still unclear on why this is necessary. Can the content of JSONEncoder be relegated to options instead?

d-torrance: Possibly -- I'll play around with it

d-torrance: Figured it out!  This is much nicer -- thanks for the suggestion!

mahrud:
Also, take a look at various json methods I had to define to deal with serializing hypertext objects and documentation nodes [here](https://github.com/mahrud/LearnM2/blob/learn/generate-json.m2). I had to resort to a hacky middle ground which you might not like, namely having `toJSON VisibleList` return an `Array` and `toJSON HashTable` return a `HashTable`, then writing `format'` to convert to a string. This was necessary to handle nested hash tables and lists without superfluous quotations all over the place, but perhaps there is an easier solution that I missed.

I don't remember the reason for each particular entry, but once you have `toJSON Hypertext`, as a challenge try something like this with different packages:
```m2
elapsedTime pkg = loadPackage(pkgname, Reload => true, LoadDocumentation => true)

L = new HashTable from pkg#"raw documentation";
L = selectValues(L, x -> not x#?PrimaryTag and not x#?"undocumented");
elapsedTime ("static/"|pkgname|".json") << toJSON(L, Indent => 2) << flush << close
```
I still can't get `Macaulay2Doc` serialized this way, but smaller packages work fine.

d-torrance:
Cool, I'll check it out!

mahrud:
This is not valid JSON:
```m2
i14 : json { "A\nB" }

o14 = ["A
      B"]

i15 : fromJSON json { "A\nB" }
stdio:16:8:(3): error: syntax error at token '
', position 3

```

d-torrance:
I added a few new `toJSON` methods so that calling `toJSON` on a package's raw documentation appears to mostly work.

d-torrance:
> This is not valid JSON:

Indeed!  Good catch.

mahrud:
> I added a few new `toJSON` methods so that calling `toJSON` on a package's raw documentation appears to mostly work.

Now try `fromJSON` and compare with the raw documentation haha

mahrud:
I think other than the newline issue this is working great.

An annoying source of complexity for me was if I want `json TO` to print links in a different way, then I also need to reimplement `json SPAN` and any other container which may contain a `TO`. The shortcut is to just redefine `html TO`, but that would have other side-effects that break the rest of the system. I'm not sure how to handle this best.

mahrud:
This is mostly unrelated, but do you know what's happening here?
```m2
i35 : html Saturation#"raw documentation"#"annihilator"#Inputs#0

o35 = <span><span class="tt">M</span>, $\left\{\texttt{an }\texttt{DocumentTag}\left\{\mathit{Format}\ \Rightarrow \
      \texttt{Ideal},\:\mathit{Key}\ \Rightarrow \ \mathit{Ideal},\:\mathit{Package}\ \Rightarrow \
      \texttt{Macaulay2Doc},\:\texttt{RawDocumentation}\ \Rightarrow \ \texttt{null}\right\}\texttt{ideal},\:\texttt{,
      },\:\texttt{a }\texttt{DocumentTag}\left\{\mathit{Format}\ \Rightarrow \ \texttt{Module},\:\mathit{Key}\ \Rightarrow \
      \mathit{Module},\:\mathit{Package}\ \Rightarrow \ \texttt{Macaulay2Doc},\:\texttt{RawDocumentation}\ \Rightarrow \
      \texttt{null}\right\}\texttt{module},\:\texttt{, or },\:\texttt{a }\texttt{DocumentTag}\left\{\mathit{Format}\
      \Rightarrow \ \texttt{RingElement},\:\mathit{Key}\ \Rightarrow \ \mathit{RingElement},\:\mathit{Package}\ \Rightarrow \
      \texttt{Macaulay2Doc},\:\texttt{RawDocumentation}\ \Rightarrow \ \texttt{null}\right\}\texttt{ring element}\right\}$,
      </span>
```

d-torrance:
Maybe this is overkill, but I went ahead and updated `format` so that it escapes newline characters (plus a few others -- all the ones we can escape in M2 strings).  This seems like reasonable behavior based on its documentation:

> a string obtained from s by inserting escape sequences, thereby preparing it for printing in a form suitable for reading in again
 
It's still not quite right when composing with `fromJSON`, though:

```m2
i2 : fromJSON toJSON "foo\nbar"

o2 = foo\nbar

i3 : oo == "foo\nbar"

o3 = false
```

d-torrance:
I think this should be in decent shape now, and escaped characters in JSON strings are now properly de-escaped in `fromJSON`.  So for example:

```m2
i2 : toJSON "foo
     bar"

o2 = "foo\nbar"

i3 : fromJSON oo

o3 = foo
     bar
```

d-torrance:
This still needs a bit more work:

```m2
i2 : toJSON "foo\0"

o2 = "foo\0"

i3 : fromJSON oo
stdio:3:8:(3): error: syntax error at token '0', position 5
```

`toJSON "foo\0"` should probably instead return `"foo\u0000"`.  I think maybe `format` should just do that for all the control characters except the ones that JSON supports like `\n` and `\t`.  I'll take a look at some other common standards (Python strings, etc.) and try to make `format` output things that are as universal as possible.

d-torrance:
Ok -- I'll go ahead and merge this.  I'll likely play around with improving `format` so it works with these other control characters in a future PR.

PR #3672, opened 2025-02-24
Organizing Macaulay2Doc #1

mahrud:
This is a massive PR that reorganizes Macaulay2Doc by chapter, meaning that as much as possible subnodes of the same branch are moved to the same file. The first benefit is that this way it's easier to linearly proofread the documentation like a book. The second benefit is that eventually we can store several nodes in a single html file containing a whole branch, which is easier to read or search through than spread across several files.

To give you a better idea, here is the table of contents along with the file each node is stored in: https://gist.github.com/mahrud/d0a64b1513497224dd6a8e8f8d03190d

- **moved Macaulay2Doc demos**
- **moved a few Macaulay2Doc files**
- **moved isIsomorphism to shared.m2**
- **removed a couple of unused Macaulay2Doc files**
- **moved expression and value docs**
- **combined a few Macaulay2Doc files**
- **moved documentation of Fano**
- **moved docs for permutations**
- **organized several list related doc nodes**
- **organized doc nodes in doc13.m2**
- **collected documentation of ^ in caret.m2**
- **organized doc nodes in doc2.m2**
- **moved documentation of intervals**
- **collected overview docs about chain complexes**
- **reorganized overview doc nodes**
- **organized more overview nodes in order**


d-torrance (approved):
This is great!!

mahrud:
Now someone needs to _just_ read it all :)

mikestillman:
really!  I'll take a look.  First need to finish another PR ...

PR #3678, opened 2025-02-25
Load tests from "tests" subdirectory of package directory

d-torrance:
This update allows package authors to add files to the `tests` subdirectory of a package directory and have their contents automatically be loaded as package tests (like the Core tests), replacing the need to use `TEST get(...)`.  The JSON package is updated to use this feature as a proof of concept.

Surprisingly, only one package (Polyhedra) already had a `tests` subdirectory, so we shuffle around its files a bit.

Cc: @mikestillman -- we talked about this idea briefly at Fields


mikestillman (commented):
This generally looks good.  Did you mention to @lkastner about the changes to the Polyhedra package (even though they are essentially cosmetic)?  (I guess I'm alerting him now, if he wasn't aware earlier :)

I think it is a good idea.  Just to check: each file is treated as a single test object, and this object is visible with `tests("Foo")`, right?

Basically, I'm ready to approve this.  I have not downloaded it to try it, but I definitely like the idea, and the code changes seem reasonable, at least the non-autotools ones, where I can understand them...

Thanks for doing this!

d-torrance:
> I think it is a good idea. Just to check: each file is treated as a single test object, and this object is visible with `tests("Foo")`, right?

Yes, that's exactly right!

lkastner (approved):

PR #3681, opened 2025-03-04
Remove old bug files

d-torrance:
Removing some old bug files that refer to bugs that have been fixed.  Details are in the individual commit messages.


mikestillman:
@d-torrance Were there any test files you think we should create because of any of these issues?

d-torrance:
Possibly -- I'll take another look soon

d-torrance:
There were only a few bugs that really made sense to add as tests, either because tests already existed or the bug didn't really need one (licensing issues, Debian packaging issues, etc.).

I also added `vector(Vector)` as a better fix of one of the bugs.

I think this is in pretty good shape now

d-torrance:
Discussed with Mike

PR #3683, opened 2025-03-06
Update fplll (and lots of other programs/libraries)

d-torrance:
I'd never noticed this before, but we have the option of linking against [fplll](https://fplll.github.io/fplll/) to use for one of the `LLL` strategies in the `LLLBases` package.  However, we were forced to build it ourselves rather than using the library already available on the system, and the version we were building was over 10 years old!

We update the autotools build so that it also checks the system for the library, and so that when we do build it, we build the latest version.

This is a draft for now to make sure that the tests still pass when building fplll ourselves.  If that all works, then I'll remove that commit and just use the Ubuntu/brew fplll packages.


d-torrance:
Quite possibly!  I discovered not too long ago that when we built GMP it was like a 10 year old version.  It would definitely be good to go through everything and check.

d-torrance:
I audited what we're building (at least on the autotools side of things).  These are the things that are out of date:

| library | M2 version | latest verson | notes |
| --- | --- | --- | --- |
| cdd+ | 77a | 77 | are we using this? |
| factory | 4.4.0 | 4.4.1 | 4.4.0 seems to be the lowest version with its own tarball outside of singular |
| flint | 3.0.0 | 3.1.3 | |
| gc | 8.2.6 | 8.2.8 ||
| gdbm | 1.9.1 | 1.24 |
| gfan | 0.6.2 | 0.7 ||
| gftables | 4.0.1 | 4.4.1 | why isn't this just included w/ factory? |
| glpk | 4.5.9 | 5.0 ||
| gtest | 1.11.0 | 1.16.0 ||
| lapack | 3.9.0 | 3.12.1 ||
| libtool | 2.4.2 | 2.5.4 | remove? this should be available already, plus I think I removed the Makefile that built it |
| linbox | 1.7.0 | 1.7.0 | remove? we aren't using this |
| lrslib | 071a | 073 | |
| msolve | 0.7.3 | 0.7.4 ||
| nauty | 2.8.8 | 2.8.9 ||
| pari | 2.11.2 | 2.17.1 | remove -- we stopped using this |
| polymake | 2.9.8 | 4.12 | Polymake package should be updated to handle latest versions |
| tbb | 2021.13.0 | 2022.0.0 ||
| topcom | 0.17.8 | 1.1.2 ||

d-torrance:
All the tests are finally passing when we build all these libraries and programs!  https://github.com/Macaulay2/M2/actions/runs/13733104922

I'll remove the final commit now and go back to Ubuntu/Homebrew packages

M2/cmake/build-libraries.cmake
```diff
@@ -180,7 +180,7 @@ ExternalProject_Add(build-bdwgc
   CMAKE_ARGS        -DCMAKE_INSTALL_PREFIX=${M2_HOST_PREFIX}
                     -DCMAKE_SYSTEM_PREFIX_PATH=${M2_HOST_PREFIX}
                     -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
-                    -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS}
+                    -DBUILD_SHARED_LIBS=ON # for ForeignFunctions package
```
mahrud, line 183: Why is this forced on?

M2/cmake/check-libraries.cmake
```diff
@@ -178,7 +178,6 @@ endif()
 ###############################################################################
 ## TODO: Do we still want these libraries?
 #   fplll	Lattice algorithms using floating-point arithmetic	(uses gmp and mpfr)
-#   linbox	Exact computational linear algebra	(needs fflas and givaro)
```
mahrud, line 181: My understanding was that Mike was still interested in experimenting with Linbox, even though it isn't actively in use.

M2/cmake/build-libraries.cmake
d-torrance: `ForeignFunctions` has a `getMemory` function for allocating memory, which is very useful when your foreign function has a pointer as one of its arguments.  Under the hood, it calls either `GC_malloc` or `GC_malloc_atomic`.  Currently, this is implemented as a foreign function at top level, which means that we need libgc to be dynamically linked so we can use `dlsym`.

It could be implemented differently to allow for static linking by moving some of that code to ffi.d.  Would that be preferable?

M2/cmake/check-libraries.cmake
d-torrance: @mikestillman - What's the status of using linbox?

M2/cmake/build-libraries.cmake
mahrud: I think so? The ability to have a completely static build is always good (though there are currently a few obstacles using the CMake build that I don't understand).

d-torrance: FWIW, I checked and I have the same problems statically linking bdwgc in the cmake build on the `development` branch, so not it's not an issue with the newer version of bdwgc or anything.

mahrud:
This is definitely not in the scope here, but if you're looking at LLL package anyway, would be great if you could:
1. figure out what't the deal with the LLL strategy of (minimalPresentation, Module) which is disabled by `return null`?
2. document these nodes, even if just a stub:
```m2
gramm
gramm(Matrix)
LLL(...,Limit=>...)
gcdLLL(...,Strategy=>...)
kernelLLL
kernelLLL(...,Limit=>...)
kernelLLL(Matrix)
hermite
hermite(Matrix)
hermite(...,Strategy=>...)
hermite(...,ChangeMatrix=>...)
hermite(...,Limit=>...)
```

d-torrance:
Msolve 0.7.5 was just released, so I've updated this PR accordingly.

d-torrance:
@sashahbc: FYI, one of the ToricHigherDirectImages tests was failing in the builds for this PR:

```m2
i6 : HT = computeEigencharacters(phi,1,D);

i7 : chars = keys HT

o7 = {| 0  |, | 0 |}
      | -1 |  | 1 |

o7 : List

i8 : assert(chars == {matrix{{0},{1}}, matrix{{0},{-1}}})
stdio:9:6:(3): error: assertion failed
```

Since `chars` is the output of `keys` and there's not guarantee that hash table keys will be in any particular order, I went ahead and updated this check to:

```m2
assert(set chars == set {matrix{{0},{1}}, matrix{{0},{-1}}})
```

d-torrance:
@mahrud - I just realized that the gtest commit was updated to https://github.com/google/googletest/commit/e90fe2485641bab0d6af4500192dc503384950d1 in #3744, which is more recent than the v1.16 tag, which is where I'm proposing that we update it in this PR.  Was there a particular reason for using that commit?

mahrud:
It really doesn't matter.

d-torrance:
Ok cool.  I'll keep this PR the same, then.  At some point between 1.16 and the current commit, googletest removed support for C++14, which is leading to a failure in the autotools build in RHEL:

```
configure:4087: checking gtest/gtest.h usability
configure:4087: g++ -c -std=gnu++14 -g3 -O2  -Wno-mismatched-tags -w -I/__w/M2-workflows/M2-workflows/M2/M2/BUILD/build/../../include -I/__w/M2-workflows/M2-workflows/M2/M2/BUILD/build/include -I/__w/M2-workflows/M2-workflows/M2/M2/BUILD/build/usr-host/include -isystem /usr/include/libxml2    -DNDEBUG -I/usr/include/eigen3  -I/usr/include/python3.9 -I/usr/include -DBOOST_STACKTRACE_LINK conftest.cpp >&5
In file included from /__w/M2-workflows/M2-workflows/M2/M2/BUILD/build/usr-host/include/gtest/gtest-message.h:57,
                 from /__w/M2-workflows/M2-workflows/M2/M2/BUILD/build/usr-host/include/gtest/gtest-assertion-result.h:46,
                 from /__w/M2-workflows/M2-workflows/M2/M2/BUILD/build/usr-host/include/gtest/gtest.h:63,
                 from conftest.cpp:52:
/__w/M2-workflows/M2-workflows/M2/M2/BUILD/build/usr-host/include/gtest/internal/gtest-port.h:273:2: error: #error C++ versions less than C++17 are not supported.
  273 | #error C++ versions less than C++17 are not supported.
      |  ^~~~~
```

M2/cmake/check-libraries.cmake
mikestillman: I did have the plan to experiment with linbox, but I haven't done so recently.

mikestillman:
@d-torrance Is there anything specific you are worried about here that I should take a look at.  In general, upgrading packages, at least when all the tests pass, seems like a great idea...!

mikestillman:
@d-torrance Would it be possible to keep the linbox stuff for now?  They are doing some work on the program (some commits within the last few months, not huge amount at all), and they do, I believe, handle sparse matrices over finite fields, which maybe we could use...

Everything else looks great in this PR...

d-torrance:
Ok -- I'll rebase out the linbox commit and then merge

PR #3685, opened 2025-03-09
Added K[G]-Algebras package

danielk011:


mahrud (commented):
Thanks for your contribution! I left a few comments and feedback below.

There is also more information here: https://github.com/Macaulay2/M2/wiki/Package-Writing-Style-Guide

mahrud:
One more thought: I would consider naming your main type `GroupAlgebra` (or `GroupRing`) and the main constructor `groupAlgebra` (or `groupRing`). This is more closely aligned with the [naming conventions](https://github.com/Macaulay2/M2/wiki/Package-Writing-Style-Guide).

M2/Macaulay2/packages/KGAlgebras.m2
```diff
@@ -0,0 +1,210 @@
...
+    Date => "February 19th, 2025",
+    Authors => {
+	{Name => "Daniel Koizumi", Email => "daniel.koizumi@utexas.edu"},
+    {Name => "Moses Samuelson-Lynn", Email => "m.samuelsonlynn@utah.edu"},
+    {Name => "Toshi Mowery", Email => "tosmow@gmail.com"}
+    },
+    Headline => "A package for KG-algebras",
+    Keywords => {"KG-algebra"},
+    DebuggingMode => false,
+    Reload=>true,
```
d-torrance, line 13: The `Reload` option to `newPackage` should be to `false`, or even better, just omitted since that's the default value.

d-torrance (changes requested):

M2/Macaulay2/packages/KGAlgebras.m2
```diff
@@ -0,0 +1,210 @@
...
+	FA := freeAlgebra(K,groupElts);
+	use FA; -- make FA our working ring
+	local H;
+	local productInv;
+	local multTableIdealGens;
+	for i in 0 .. n do(
+		for j in 0 .. n do( 
+			productInv = position(matrixElts,M -> M==matrixElts#i * matrixElts#j);
+			H= (FA_i)*(FA_j)-(FA_productInv);
+			multTableIdealGens=append(multTableIdealGens,H);
```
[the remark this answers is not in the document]
d-torrance: I think if the mutable list starts out empty, then we'll still end up allocating a bunch of new lists along the way as it grows, right?  So it would probably be good to do something like this first since we know how long it will be:

```m2
mutlTableIdealsGens = new MutableList from n^2:null
```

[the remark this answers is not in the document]
danielk011: Thank you for the suggestion. I'm getting errors from trying to create an ideal from a mutable list. Is there a nice way to generate an ideal from a mutable list that is officially supported?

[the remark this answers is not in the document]
d-torrance: You can call `toList` on it first, e.g.:

```m2
i1 : R = QQ[x];

i2 : ideal toList new MutableList from {x}

o2 = ideal x

o2 : Ideal of R
```

M2/Macaulay2/packages/KGAlgebras.m2
```diff
@@ -3,132 +3,115 @@ newPackage(
...
+        error "KGModules: Expected second argument to be a field.";
+    );
+    matrixElts := group G; -- create set of matrix elements
+    n := #matrixElts-1; -- maximum index, since we are starting at 0
+    groupIndices := {0..n}; -- indices
+    g := symbol g;
+    groupElts := g_0 .. g_n; -- group elements
+    FA := freeAlgebra(K,groupElts);
+    use FA; -- make FA our working ring
+    multTableIdealGens := new MutableList;
```
d-torrance, line 30: See the suggestion above.  Starting with an empty mutable list isn't much better than the old `append` implementation since we'll end up having to reallocate everything every time we grow it by one more element.  But we know the final length, so we can start out with a mutable list of the correct length and not have to worry about it ever growing:

```m2
multTableIdealGens := new MutableList from (n + 1)^2:null
```

Another option would be to avoid mutable lists entirely and use `apply`:

```m2
multTableIdealGens = apply((n + 1)^2, k -> (
	(i, j) := quotientRemainder(k, n + 1);
	...
	H)
```

M2/Macaulay2/packages/KGAlgebras.m2
```diff
@@ -3,132 +3,115 @@ newPackage(
...
     {Name => "Moses Samuelson-Lynn", Email => "m.samuelsonlynn@utah.edu"},
     {Name => "Toshi Mowery", Email => "tosmow@gmail.com"}
     },
-    Headline => "A package for KG-algebras",
-    Keywords => {"KG-algebra"},
-    DebuggingMode => false,
-    Reload=>true,
+    Headline => "A package for group algebras",
+    Keywords => {"Group Theory"},
+    DebuggingMode => true,
```
d-torrance, line 12: `DebuggingMode` should either be `false` or (even better) removed, since `false` is the default.

M2/Macaulay2/packages/KGAlgebras.m2
```diff
@@ -3,132 +3,115 @@ newPackage(
...
+    use FA; -- make FA our working ring
+    multTableIdealGens := new MutableList;
+    for i in 0 .. n do(
+        for j in 0 .. n do( 
+            productInv := position(matrixElts,M -> M==matrixElts#i * matrixElts#j);
+            H := (FA_i)*(FA_j)-(FA_productInv);
+            multTableIdealGens##multTableIdealGens = H;
+        );
+    );
+    print multTableIdealGens;
```
d-torrance, line 38: Did you mean to print this?  I'd either remove it or swap it with something like `if debugLevel > something then print ...`.

M2/Macaulay2/packages/KGAlgebras.m2
```diff
@@ -3,132 +3,115 @@ newPackage(
...
+    while true do(
+        --Do the meataxe algorithm 
+        p := random S; --generate random element of KG algebra
+        L := toList coefficients p;  
+        M := (group G)#0-(group G)#0; -- zero matrix which has the same dimensions as an element of G
+        m := 0;
+        for x in (flatten entries L#0) do(
+            n := index(x);
+            M = M + (group G)#n * sub((flatten entries L#1)#m,K);
+            m = m+1
```
d-torrance, line 56: This is totally optional, but if you're interested, you can use [augmented assignment](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_augmented_spassignment.html) here:

```m2
M += (group G)#n * ...
m += 1
```

It also works for other binary operators like `*`, so line 74 could become:

```m2
Mp *= (apply(...
```

d-torrance (changes requested):

M2/Macaulay2/packages/KGAlgebras.m2
d-torrance: Oops, yeah I missed that!  `multTableIdealGens##multTableIdealGens = H` will *always* grow the list, even if we've allocated a bunch of elements at the beginning since `#multTableIdealGens` will always point one past the last index!

Instead, something like `multTableIdealGens#((n + 1)*i + j) = H` should work.

M2/Macaulay2/packages/KGAlgebras.m2
```diff
@@ -0,0 +1,192 @@
...
+            R = QQ[x1,x2,x3];
+            L = {matrix {{0,1,0},{1,0,0},{0,0,1}}, matrix {{0,0,1},{0,1,0},{1,0,0}}, matrix {{1,0,0},{0,0,1},{0,1,0}} };
+            G = finiteAction(L,R);
+            groupRing(G,QQ)
+///
+doc ///
+    Key
+        meataxe        
+    Headline
+        Boolean function implementing the meataxe algorithm determining the irreducibility of a finite group representation of a finite dimensional vector space. If the module is irreducible, returns true. Otherwise, the method returns a nontrivial submodule.
```
d-torrance, line 162: The builds are failing with the following error:

```m2
../../../../../Macaulay2/m2/document.m2:598:32:(1):[72]: error: document: documentation headlines must be less than 100 characters long:
  "Boolean function implementing the meataxe algorithm determining the irreducibility of a finite group representation of a finite dimensional vector space. If the module is irreducible, returns true. Otherwise, the method returns a nontrivial submodule."
```
In general, headlines should just be a few words.  Most of this stuff can get moved elsewhere in the documentation node.

d-torrance (changes requested):

M2/Macaulay2/packages/KGAlgebras.m2
```diff
@@ -0,0 +1,191 @@
+newPackage(
+    "KGAlgebras",
+    Version => "0.5",
+    Date => "March 11th, 2025",
+    Authors => {
+    {Name => "Daniel Koizumi", Email => "daniel.koizumi@utexas.edu"},
+    {Name => "Moses Samuelson-Lynn", Email => "m.samuelsonlynn@utah.edu"},
+    {Name => "Toshi Mowery", Email => "tosmow@gmail.com"}
+    },
+    Headline => "A package for group algebras",
```
d-torrance, line 10: The tests are failing because of the redundant "package for" in the headline:

```m2
/Users/runner/work/M2/M2/M2/Macaulay2/tests/normal/release-checklist.m2:1:1 error:
 --      --	    "^[A-Z]", -- headlines should not be capitalized, but hard to check
 --      	    "\\.$",   -- headlines should not end with a period
 --      	    "\\n",    -- headlines should not contain line breaks
 --      	    -- redundant clauses for package headlines:
 --      	    "(T|t)his (P|p)ackage",
 --      	    "(P|p)ackage (for|on|to)"}, opts.Headline));
 -- 
 -- i10 : if #pkgs > 0 then error("packages whose headlines doesn't follow guidelines:", newline,
 --           toString netList(toList \ pairs applyValues(pkgs, opts -> opts.Headline)))
 -- stdio:23:23:(3): error: packages whose headlines doesn't follow guidelines:
 -- +----------+----------------------------+
 -- |KGAlgebras|A package for group algebras|
 -- +----------+----------------------------+
 -- 
stdio:1:5:(3): error: test(s) #252 of package Core failed.
```

M2/Macaulay2/packages/KGAlgebras.m2
```diff
@@ -0,0 +1,191 @@
...
+            H := apply(group G, transpose); --Generate list of group elements' matrices but transposed
+            MPP := transpose MP;
+            NP := kernel MPP;
+            yp := (gens NP)_0;
+            M2 := matrix apply(H,z -> z*yp);
+            if debugLevel == 1 then (
+                print(rank M2);
+            );
+            if rank M2 != rank (H#0) then (
+                return image M2;
```
d-torrance, line 95: Sometimes this function returns a boolean and sometimes it returns a module.  That seems confusing for the caller -- how do they know what type to expect?

Would it make sense to return both as a sequence?  So the caller could do something like:

```m2
(isIrreducible, M) = meataxe(G,S,K)
if isIrreducible then ...
```

mahrud: I have another implementation of the meataxe for modules and we return a list of the indecomposable pieces. So if the input M is indecomposable, the ouput is just {M}. You could try this instead.

M2/Macaulay2/packages/KGAlgebras.m2
```diff
@@ -0,0 +1,191 @@
...
+    Inputs
+        G:FiniteGroupAction
+            A finite group action
+        S:FreeAlgebraQuotient
+            A KG-algebra
+        K:Ring
+            The underlying field of the other parameters.
+    Outputs
+        T:Boolean
+            Returns true if the Meataxe algorithm states that the K[G]-module is irreducible. Returns a nontrivial submodule if the module is reducible.
```
d-torrance, line 173: The documentation should be updated since this doesn't return a boolean any more.

Also, now that a module is return in both cases, how do we know if it's irreducible?

d-torrance (changes requested):

seangrate:
> One more thought: I would consider naming your main type `GroupAlgebra` (or `GroupRing`) and the main constructor `groupAlgebra` (or `groupRing`). This is more closely aligned with the [naming conventions](https://github.com/Macaulay2/M2/wiki/Package-Writing-Style-Guide).

I'd also consider renaming the package "GroupAlgebras" or something along those lines.

d-torrance:
@danielk011 - Just a friendly reminder about this PR

Issue #1690, opened 2020-12-13
Allow Sequences of specific types for NewFromMethod


mahrud:
It would be great to allow something like this:
```m2
new X from (Y, Z, W) := ...
```
With automatic type checking and lookup. The method key could be `(NewFromMethod, X, (Y, Z, W))`, or maybe the last items don't need to be in a sequence.

_Originally posted by @mahrud in https://github.com/Macaulay2/M2/pull/1385#r531725791_

This is the dual of #1375, which is asking for the output type to be a sequence with specific types.

mahrud:
Just to be a little bit more general, would be good to combine this with #1375 and allow:
```m2
new (A, B, C) from (X, Y, Z) := ...
```
The distinctive feature of `NewFromMethod` (and other `New*Method`) in my view is that the method can also be installed on the output types, rather than the input, and it kind of makes sense for a type to have pointers to its own constructors. Right?

@DanGrayson do you imagine a method like the one above should be installed on the youngest of all `A B C X Y Z` or just `A B C`?

For other types of methods, I can imagine a new syntax for `installMethod`:
```m2
installMethod( (m, A, B, C), (X, Y, Z), f )
```
@DanGrayson do you see any reasons against this syntax?
For this one it's more clear that the method should be installed under the youngest of `A B C` or in `nullaryMethods`.

DanGrayson:
> Just to be a little bit more general, would be good to combine this with #1375 and allow:
> 
> ```m2
> new (A, B, C) from (X, Y, Z) := ...
> ```
> 
> The distinctive feature of `NewFromMethod` (and other `New*Method`) in my view is that the method can also be installed on the output types, rather than the input, and it kind of makes sense for a type to have pointers to its own constructors. Right?

That might be a good idea.
> 
> @DanGrayson do you imagine a method like the one above should be installed on the youngest of all `A B C X Y Z` or just `A B C`?

On the youngest of all of them, so it gets collected as soon as it's unusable.

> 
> For other types of methods, I can imagine a new syntax for `installMethod`:
> 
> ```m2
> installMethod( (m, A, B, C), (X, Y, Z), f )
> ```
> 
> @DanGrayson do you see any reasons against this syntax?

It should be something like `installMethod( (m, A, B, C), X, Y, Z, f )` for compatibility with the current
syntax. 
> ```

> For this one it's more clear that the method should be installed under the youngest of `A B C` or in `nullaryMethods`.

[...]

PR #3686, opened 2025-03-11
Allow 2 or 3 inputs to "new ... from" methods

d-torrance:
I'm a big fan of self-initializing types since they make the syntax of constructing new objects of a given type similar to other object-oriented languages.

However, it's kind of a pain if you want to make it work with multiple arguments.  Currently, we have to do the following:

```m2
Foo = new SelfInitializingType of HashTable
new Foo from Sequence := (T, s) -> (
   -- check the length of s and the types of its elements to make sure that they're what you want
```

I figured it would be great if we could let the usual multiple dispatch we use when looking up methods do this work for us.  So we add the ability to for `new ... from` to take 2 or 3 arguments.  (We can't go up to 4 since the type counts as one of the arguments already.)

For example:

```m2
i1 : X = new SelfInitializingType of HashTable;

i2 : new X from (ZZ,ZZ)    := (T,x,y)   -> T {a => x, b => y, c => 0};

i3 : new X from (ZZ,ZZ,ZZ) := (T,x,y,z) -> T {a => x, b => y, c => z};

i4 : X(2,3)

o4 = X{a => 2}
       b => 3
       c => 0

o4 : X

i5 : X(3,4,5)

o5 = X{a => 3}
       b => 4
       c => 5

o5 : X
```

As a proof of concept, we update a couple `new .. from` methods in `Core` (for `Module` and `GroebnerBasis`) to use this proposed feature.

### DIFF
--- a/M2/Macaulay2/m2/modules.m2
+++ b/M2/Macaulay2/m2/modules.m2
@@ -156,18 +156,15 @@ isHomogeneous Vector := (v) -> isHomogeneous v#0
 Module = new Type of ImmutableType
 Module.synonym = "module"
 new Module from List := (Module,v) -> new Module of Vector from hashTable v
-new Module from Sequence := (Module,x) -> (
-     (R,rM) -> (
-	  assert instance(R,Ring);
-	  assert instance(rM,RawFreeModule);
-	  new Module of Vector from hashTable {
-     	       symbol cache => new CacheTable from { 
-		    cache => new MutableHashTable	    -- this hash table is mutable, hence has a hash number that can serve as its age
-		    },
-     	       symbol RawFreeModule => rM,
-     	       symbol ring => R,
-     	       symbol numgens => rawRank rM
-     	       })) x
+new Module from (Ring, RawFreeModule) := (Module, R, rM) -> (
+    new Module of Vector from hashTable {
+	symbol cache => new CacheTable from {
+	    cache => new MutableHashTable  -- this hash table is mutable, hence has a hash number that can serve as its age
+	    },
+	symbol RawFreeModule => rM,
+	symbol ring => R,
+	symbol numgens => rawRank rM
+	})
 
 vector(Module, Matrix) := (M, f) -> vector map(M,,entries f)
 vector(Module, List)   := (M, v) -> vector map(M,,apply(splice v, i -> {i}))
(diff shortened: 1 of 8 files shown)


mahrud:
I'm not opposed to this, but I want to suggest a generalization of this using algebraic types, e.g. see ListType in https://github.com/Macaulay2/M2/issues/1979.

d-torrance:
Yeah, that looks really cool.  But a lot of work.  :)

Clicking through the various links at that PR, I noticed that this (at least partially) resolves #1690.

d-torrance:
I added one more feature -- `new X from ()` as a synonym for `new X`.    This way, `X()` works for self-initializing types.

So for example:

```m2
i1 : X = new SelfInitializingType of BasicList

o1 = X

o1 : SelfInitializingType

i2 : new X := T -> new T from {1,2,3}

o2 = FunctionClosure[stdio:2:9-2:32]

o2 : FunctionClosure

i3 : X()

o3 = X{1, 2, 3}

o3 : X
```

d-torrance:
I'm having second thoughts about that last commit...  I think it's a breaking change for subclasses of `BasicList`, where `X()` would have previously worked just fine and given us an instance of `X` as an empty list.  I'll remove it.

M2/Macaulay2/m2/modules.m2
```diff
@@ -156,18 +156,15 @@ isHomogeneous Vector := (v) -> isHomogeneous v#0
...
-     	       })) x
+new Module from (Ring, RawFreeModule) := (Module, R, rM) -> (
+    new Module of Vector from hashTable {
+	symbol cache => new CacheTable from {
+	    cache => new MutableHashTable  -- this hash table is mutable, hence has a hash number that can serve as its age
+	    },
+	symbol RawFreeModule => rM,
+	symbol ring => R,
+	symbol numgens => rawRank rM
+	})
```
mahrud, line 167: Do you know why the previous version of this method was implemented like this?
```m2
new Module from Sequence := (Module,x) -> (  (R,rM) -> ( ... ) ) x
```

mahrud (approved):

M2/Macaulay2/m2/modules.m2
d-torrance: No idea

PR #3687, opened 2025-03-11
Add JSONRPC package

d-torrance:
We add a `JSONRPC` package implementing a basic [JSON-RPC](https://www.jsonrpc.org/) server.  This is split off from the `LanguageServer` package (see #3655) since JSON-RPC has uses beyond LSP (using Macaulay2 to mine for bitcoin?? :rofl:)

This is currently a draft on top of #3686 since one of the exported classes uses the proposed new `new ... from` behavior.

Here's a basic example of its use:
```m2
i2 : server = new JSONRPCServer

o2 = server

o2 : JSONRPCServer

i3 : registerMethod(server, "sum", plus)

i4 : setLogger(server, printerr)

i5 : handleRequest(server, makeRequest("sum", {2, 3}, 1))
 -- received request: {"method": "sum", "id": 1, "jsonrpc": "2.0", "params": [2, 3]}
 -- calling method "sum" with params [2, 3]
 -- method "sum" returned: 5
 -- sending response: {"result": 5, "jsonrpc": "2.0", "id": 1}

o5 = {"result": 5, "jsonrpc": "2.0", "id": 1}
```

### DIFF
--- a/M2/Macaulay2/packages/=distributed-packages
+++ b/M2/Macaulay2/packages/=distributed-packages
@@ -286,3 +286,4 @@ IntegerProgramming
 GameTheory
 AllMarkovBases
 CpMackeyFunctors
+JSONRPC
--- a/M2/Macaulay2/packages/JSONRPC.m2
+++ b/M2/Macaulay2/packages/JSONRPC.m2
@@ -1,0 +1,728 @@
+-- JSONRPC package for Macaulay2
+-- Copyright (C) 2025 Doug Torrance <dtorrance@piedmont.edu>
+
+-- This program is free software; you can redistribute it and/or modify
+-- it under the terms of the GNU General Public License as published by
+-- the Free Software Foundation; either version 2 of the License, or
+-- (at your option) any later version.
+
+-- This program is distributed in the hope that it will be useful,
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
+-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+-- GNU General Public License for more details.
+
+-- You should have received a copy of the GNU General Public License along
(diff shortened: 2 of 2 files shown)


mahrud:
What's the reasoning behind separating this from JSON?

d-torrance:
JSON is a lower-level package for converting back and forth between Macaulay2 objects and JSON strings, and JSONRPC is a package that uses JSON to implement a JSON-RPC server.  JSON-RPC is a protocol that just happens to be implemented using JSON.  I'd liken it to any of the various packages that import gfanInterface to make calls to gfan.

mahrud:
That comparison doesn't make sense to me. JSON has <400 lines and essentially 2 methods, gfanInterface has <5000 lines and about 60 methods! A few of those are low level translation functions and the rest are pretty high level. It makes sense to me to to separate this from LanguageServer, but I don't see a reason why it should be a separate package if JSON can reasonably absorb it.

Also consider that, for now, there's a limit to how many databases M2 can open simultaneously.

d-torrance:
Presumably there might eventually be other packages that make use of JSON.  Should they pull in the unrelated JSON-RPC stuff?  I'm of the opinion that packages should be modular and focus on one thing.

Maybe a better example would be if JSON was included in Parsing.  Why would we want to pull it in for Classic?

Are we approaching the maximum number of open databases on any modern systems?  `ulimit -n` is 1024 on my machine.

mikestillman:
I think JSON and JSONRPC should be in different packages, as JSONRPC is just using JSON, not part of the JSON protocol.

Eventually, I would like to see a different way of doing databases for doc, as currently, it is quite a few files...

This package looks good to me, although I have not gone through it line by line...  But as it is on its own, and as it will be tested further when we do the language server code, I think we should accept this PR.

PR #3688, opened 2025-03-11
Organizing Macaulay2Doc #3

mahrud:
- **fixed style issues in various doc headlines**
- **made lift/promote subnodes of ring maps**
- **made logical integer operators subnodes**
- **made topLevelMode a subnode**
- **moved various file related doc nodes**
- **added node for analytical functions**
- **added apply(Thing, Command) and VisibleList // Function for convenience**
- **updated docs for lists**
- **added various doc reorganizations**
- **made many methods docs subnodes**
- **made many doc nodes subnodes**
- **moved docs of Matrix, Vector, Symbol, Keyword**
- **moved some ChainComplex nodes**
- **added isFinitePrimeField(GaloisField)**
- **made more doc nodes subnodes**
- **removed CacheFunction and stashValue**
- **added node for caching**
- **moved docs out of doc8.m2**
- **removed null entries from Subnodes and SeeAlso lists**
- **made almost all docs subnodes, added isEmpty(Thing)**
- **added the final node as a subnode**
- **changed > to » on doc pages**
- **removed null entries in Authors lists**
- **added check that every node in Macaulay2Doc is a subnode**
- **added debug(String) and debug(LocalDictionary)**
- **documented every missing node in Macaulay2Doc**


mahrud:
Upshots:
- every node is a subnode now (though some will need to move elsewhere)
- there are no missing docs from Macaulay2Doc itself

M2/Macaulay2/m2/methods.m2
```diff
@@ -683,28 +683,10 @@ protect QuotientRingHook
 -- stashing or caching computed values for future reference in functions that take a mutable hash table as input
 -----------------------------------------------------------------------------
 
-CacheFunction = new Type of FunctionClosure
```
d-torrance, line 681: Why is CacheFunction going away?  I've never used it, but it looks like maybe a lot of this caching code can be simplified with `??=`?  What above the removing of keys?

M2/Macaulay2/m2/methods.m2
```diff
@@ -683,28 +683,10 @@ protect QuotientRingHook
...
-	  if x#?key then (
-	       val := x#key;
-	       if class val === CacheFunction then (
-		    remove(x,key);
-		    x#key = val x)
-	       else val
-	       )
-	  else x#key = f x))
+-- deprecated, but kept for backwards compatibility
+-- FIXME: https://github.com/Macaulay2/M2/issues/3612
```
d-torrance, line 687: Is there something still outstanding with #3612?  I thought it was fixed in #3614?

M2/Macaulay2/packages/Macaulay2Doc/testing.m2
```diff
@@ -0,0 +1,119 @@
+debug Core
```
d-torrance, line 1: Did you mean to commit this file?

M2/Macaulay2/m2/methods.m2
mahrud: It's not really used anywhere. `cacheValue` and `stacheValue` returned this type, but my plan is to deprecate those in favor of `??=`.

M2/Macaulay2/packages/Macaulay2Doc/testing.m2
mahrud: I guess not, but it doesn't hurt.

mahrud:
I'm made this a draft until the Varieties is merged, then I'll rebase and fix the couple of things you mentioned also.

d-torrance (approved):
Amazing!  Thanks again!!

M2/Macaulay2/m2/methods.m2
pzinn: actually, `CacheFunction` is used, for example in the package `LieTypes`, which no longer loads now.

pzinn: if `cacheValue` is to be deprecated, then some discussion is needed on what exactly it will be replaced with.

pzinn:
I see, you've changed `LieTypes`, with this commit it should work again.

PR #3690, opened 2025-03-12
Use C++ when checking for eigen (autotools)

d-torrance:
I was playing around with building Macaulay2 in [OpenIndiana](https://www.openindiana.org/) and was puzzled when it failed detecting eigen.

I checked out `config.log` and noticed that the check failed when one of the eigen headers tried to `#include <cmath>`, and it couldn't find the file.  It turns out that it was trying to compile using `gcc` instead of `g++`!

It's a trivial fix, just adding `AC_LANG([C++])`, so I'm skipping the builds.

Cc: @MichaelABurr


mikestillman (approved):
This looks fine to me!

PR #3692, opened 2025-03-13

n-m-g:
Some additional checks added and other cosmetic improvements.

### DIFF
--- a/M2/Macaulay2/packages/AbstractSimplicialComplexes.m2
+++ b/M2/Macaulay2/packages/AbstractSimplicialComplexes.m2
@@ -22,7 +22,7 @@
 newPackage(
     "AbstractSimplicialComplexes",
     Version => "1.0",
-    Date => "12 March 2025",
+    Date => "13 March 2025",
     Headline => "Abstract Simplicial Complexes",
     Authors => {{ Name => "Nathan Grieve", Email => "nathan.m.grieve@gmail.com",
 	    HomePage => "https://sites.google.com/view/nathan-grieve"}},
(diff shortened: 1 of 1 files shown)


M2/Macaulay2/packages/AbstractSimplicialComplexes.m2
```diff
@@ -241,7 +241,9 @@ randomSubSimplicialComplex(AbstractSimplicialComplex) := AbstractSimplicialCompl
 ambientAbstractSimplicialComplexSize = method() 
 
 ambientAbstractSimplicialComplexSize(AbstractSimplicialComplex) := (K) -> (
-    max flatten(K_0))
+    L := flatten(K_0);
+    if L != { } then return(max L) else return(0) -- by default {{}} has size 0
```
d-torrance, line 245: Thanks for the updates!

I'll go ahead and merge this, but a couple stylistic things for future reference:

* It isn't necessary to use `return` in the last line of a function.
* Parentheses aren't necessary for functions with a single argument.

See also the [Package Writing Style Guide](https://github.com/Macaulay2/M2/wiki/Package-Writing-Style-Guide) for these and lots of other tips.

So for example, this could be simplified to:

```m2
L := flatten K_0;
if L != { } then max L else 0
```
or even more compactly:
```m2
if (L := flatten K_0) != {} then max L else 0
```

d-torrance (approved):

n-m-g:
Hi,

Many thanks for all of this. 

Best,
Nathan

PR #3693, opened 2025-03-13
Networking updates

d-torrance:
I've been looking through our networking code while I work on the language server and have made a few changes:

* We had quite a few things hidden behind `#ifdef HAVE_function` clauses for networking functions that have been POSIX standard for over twenty years, so we remove those.  We also remove a bunch of checks for functions that generated macros we weren't actually using.
* We refactor the error handing for `getaddrinfo`, which previously used a global variable and I don't think it was thread safe and remove a few unused error handling functions.
* We loop through all the addresses returned by `getaddrinfo` and use the first one that works rather than just hoping that the very first one works.
* Add support for `openOutAppend` to sockets and pipes.  It does the exact same thing as `openOut`, but our current behavior for something like `openOutAppend "$localhost:12345" << "Hello, world!" << close` is to actually write to a file named `$localhost:12345` rather than send a message to port 12345.

This is a draft for now, as I'll probably do more with it (e.g., documentation).


mahrud:
I would ask someone else for review if you want this in this release.

d-torrance:
@jkyang92 - Would you be willing to take a look at this PR?

M2/Macaulay2/d/actors2.dd
```diff
@@ -425,96 +425,115 @@ export setStdIO   (f:file):void := ( stdIO    = f; setGlobalVariable(stdioS,  Ex
...
 	  is g:file do Expr(g)
 	  is m:errmsg do buildErrorPacket(m.message))
-     is Error do filename
      else WrongArgString());
 setupfun("openOut",openOut);
 openOutAppend(filename:Expr):Expr := (
      when filename
+     -- # typical value: openOutAppend, File, File
+     is f:file do (
+	  when openOut(f)
```
jkyang92, line 483: Does `openOutAppend` make sense for listeners? Regardless, this is non-obvious enough it should probably have a comment before the call to `openOut(f)` that mentions that the only files this accepts are listeners

M2/Macaulay2/d/scclib.c
```diff
@@ -722,38 +657,67 @@ int system_acceptNonblocking(int so) {
...
+
+   The following function determines the error message:
+*/
+
+M2_string system_netstrerror(int errcode) {
+  if (errcode == ERROR || errcode - ERROR == EAI_SYSTEM)
+    return M2_tostring(strerror(errno));
+  else
+    return M2_tostring(gai_strerror(errcode - ERROR));
+}
```
jkyang92, line 678: system_strerror tests the interrupt flag and returns a different message, should this test it as well?

M2/Macaulay2/d/hashtables.dd
```diff
@@ -416,7 +416,7 @@ installIt(numtypes:int,assgnmt:bool,h:HashTable,key:Expr,value:Expr):Expr := (
      else buildErrorPacket(messx));
 -----------------------------------------------------------------------------
 -- nullary methods
-export NullaryMethods := newHashTableWithHash(mutableHashTableClass,nothingClass);
+export NullaryMethods := newHashTableWithHash(typeClass,nothingClass);
```
jkyang92, line 440: Maybe add a comment to the effect that the NullaryMethods object is a dummy type for the nullary methods to install into?

jkyang92 (commented):
I think this is essentially fine, I added some minor notes. Sorry that I didn't really remember that you had asked me for this till now.

M2/Macaulay2/d/actors2.dd
d-torrance: Hmm, good question.  Maybe we should just raise an error in this case?

jkyang92: I'd either have `openOutAppend` not accept listeners or do what you did here, but I'm leaning towards just adding a comment. But this detail should probably be documented. In particular the documentation openListener only mentions openInOut right now.

In reading through the networking stuff, my preferred solution for all of this is for accept to be a different method, and for `openListener` to return a Listener type which probably should be a subtype of File.

Right now `openIn`/`openOut`/`openInOut` on listeners all just call accept and so at the actual socket level it doesn't make a difference. The only difference is whether the Macaulay2 file object can be read/write from.

mikestillman:
@d-torrance @jkyang92 Are you both happy with this package yet?

d-torrance:
No -- I still need to address Jay's comments

PR #3695, opened 2025-03-14

n-m-g:

### DIFF
--- a/M2/Macaulay2/packages/AbstractSimplicialComplexes.m2
+++ b/M2/Macaulay2/packages/AbstractSimplicialComplexes.m2
@@ -29,7 +29,6 @@ newPackage(
     AuxiliaryFiles => false,
     DebuggingMode => false,
     PackageImports => {"Complexes"},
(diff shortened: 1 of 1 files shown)


n-m-g:
Thanks!

d-torrance:
Great!

Yes, one of the things we do at each release is "certify" all the packages that have been published in JSAG since the last release (which includes a nifty gold star in the [packages list](https://www.macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_packages_spprovided_spwith_sp__Macaulay2.html)).  That includes updating the code to match the published version, if necessary.  This will all happen in the next few weeks, since Macaulay2 1.25.05 is scheduled to be released on May 15.

However, I don't see a link to the source code on [JSAG](https://msp.org/jsag/2025/15-1/p03.xhtml).  Usually there's a "Supplementary material" subsection with a link.  Maybe it's just been published too recently and they're still working on it?

n-m-g:
Yes, many thanks.  The article just appeared so, as you say I expect that the source code will appear along side of it in due time.

Best,
Nathan

Issue #3697, opened 2025-03-15
Flint 3.2.0


mahrud:
Apparently M2 doesn't build with the version of Flint released a few days ago: https://github.com/Macaulay2/homebrew-tap/issues/261
```cpp
==> cmake --build M2/BUILD/build-brew --target M2-core M2-emacs
Last 15 lines from /Users/ben/Library/Logs/Homebrew/macaulay2/06.cmake:
In file included from /tmp/macaulay2-20250315-29886-cyar5j/M2-release-1.24.11/M2/Macaulay2/e/aring-glue.hpp:7:
In file included from /tmp/macaulay2-20250315-29886-cyar5j/M2-release-1.24.11/M2/Macaulay2/e/aring-translate.hpp:24:
In file included from /tmp/macaulay2-20250315-29886-cyar5j/M2-release-1.24.11/M2/Macaulay2/e/aring-zzp-flint.hpp:22:
In file included from /opt/homebrew/opt/flint/include/flint/nmod.h:22:
In file included from /opt/homebrew/opt/flint/include/flint/ulong_extras.h:26:
/opt/homebrew/opt/flint/include/flint/longlong.h:71:9: warning: '__ll_highpart' macro redefined [-Wmacro-redefined]
   71 | #define __ll_highpart(t) ((ulong) (t) >> (FLINT_BITS / 2))
      |         ^
/opt/homebrew/Cellar/givaro/4.2.0_1/include/recint/reclonglong.h:48:9: note: previous definition is here
   48 | #define __ll_highpart(t) ((UWtype) (t) >> (W_TYPE_SIZE / 2))
      |         ^
3 warnings generated.
```

Moreover, it seems like Factory 4.4.0 (which is ~a year old) may also have an issue: https://github.com/Macaulay2/homebrew-tap/issues/260
```cpp
[...]

PR #3698, opened 2025-03-15
Bump factory to 4.4.1 and FLINT to 3.2.1

d-torrance:
Since there are no longer any [factory-specific tarballs](https://www.singular.uni-kl.de/ftp/pub/Math/Singular/Factory/) after version 4.4.0, ~~we switch to downloading the entire Singular source tarball, but just building factory.~~  *Edit:*  There is now a factory 4.4.1 tarball on the Macaulay2 website.

We also drop the redundant gftables "library" from the autotools build since it's included with factory and fix the broken links pointed out in https://github.com/Macaulay2/M2/issues/3697#issuecomment-2726821527.

This is a draft for now -- I'll remove the final commit if/when the GitHub builds succeed.

This ~~partially~~ addresses #3697.  *Edit:*  This PR was combined with #3699 and also bumps FLINT to 3.2.1.


d-torrance:
This has been combined with #3699, and we bump factory and FLINT together.

mahrud:
I think Dan would ask the Singular developers and they would create a tarball for factory alone. I could be wrong though.

d-torrance:
@mahrud: Any idea about this error in the macOS cmake build?

```
CMake Error at /Users/runner/work/M2/M2/M2/BUILD/build/libraries/flint/build/CMakeFiles/CMakeScratch/TryCompile-smWOJZ/CMakeLists.txt:28 (target_link_libraries):
  Target "cmTC_e9203" links to:

    gmp::gmp

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.



CMake Error at /usr/local/share/cmake/Modules/Internal/CheckSourceCompiles.cmake:108 (try_compile):
  Failed to generate test project build system.
Call Stack (most recent call first):
  /usr/local/share/cmake/Modules/CheckCSourceCompiles.cmake:58 (cmake_check_source_compiles)
  CMakeLists.txt:156 (check_c_source_compiles)
```

mahrud:
Seems like Flint's cmake build can't find gmp, but no clue why.

d-torrance:
It looks like FLINT has stopped supporting their CMake build outside of Windows (https://github.com/flintlib/flint/pull/2256).  I just pushed a commit switching our CMake build to use their autotools build.

M2/cmake/build-libraries.cmake
```diff
@@ -340,28 +340,34 @@ _ADD_COMPONENT_DEPENDENCY(libraries ntl gmp NTL_FOUND)
...
-                    -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
-                    -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
-                    -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
-                    -DCMAKE_C_FLAGS=${CFLAGS}
-                    -DCMAKE_CXX_FLAGS=${CXXFLAGS}
-                    -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS}
-                    -DIPO_SUPPORTED=OFF # TODO: because of clang; see https://github.com/wbhart/flint2/issues/644
-                    -DWITH_NTL=ON
-  INSTALL_COMMAND   ${CMAKE_COMMAND} --install . ${strip_setting}
+  CONFIGURE_COMMAND cd ${CMAKE_SOURCE_DIR}/submodules/flint &&
```
mahrud, line 350: I think the right way is to change the `BINARY_DIR` above.

d-torrance: I think I'm just going to switch it to an in-tree build.  It turns out that the out-of-tree builds are broken and aren't installing all the header files, which is why the cmake-ubuntu build was failing (factory couldn't find flint.h).  I reported it upstream, so it should be fixed next release.  I'll mention something about bootstrap.sh, too.

M2/Macaulay2/e/aring-gf-flint-big.hpp
```diff
@@ -11,6 +11,7 @@
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wconversion"
 #include <flint/flint.h>       // for flint_free, flint_rand_t, fmpz_t
+#include <gmp.h>               // required for FLINT functions that use GMP
```
mahrud, line 14: I don't think we include `gmp.h` directly, instead you should probably include `math-include.h`.

d-torrance: We do in a few places:

```m2
$ grep -r "<gmp.h>"
NCAlgebras/FreeAlgebra.hpp:#include <gmp.h>                      // for mpz_srcptr, mpq_srcptr
NCAlgebras/FreeAlgebraQuotient.hpp:#include <gmp.h>                      // for mpz_srcptr, mpq_srcptr
schreyer-resolution/res-f4-m2-interface.cpp:#include <gmp.h>                                          // for mpz_clear
```

I don't see the advantage of including another header that includes even more headers we won't be using.

mahrud: Not really: the first two include math-include.h via ringelem.hpp, the third includes it via gbring.hpp, so in all cases including gmp.h is redundant and should be removed.

The point is to make sure every time certain libraries are loaded, the same macros.

d-torrance: I've updated that commit with your suggestion, and also removed the redundant gmp.h includes from those other files.

d-torrance:
There's some kind of floating point exception in the `ARingQQFlint.arithmetic` test that will be a pain to debug...

mahrud:
The msolve error could be unrelated.

Could you fix these warnings, though?
```cpp
In file included from /home/linuxbrew/.linuxbrew/opt/flint/include/flint/flint.h:197,
                 from /home/mahrud/Projects/M2/quickfix/M2/Macaulay2/e/aring-qq-flint.hpp:13,
                 from /home/mahrud/Projects/M2/quickfix/M2/Macaulay2/e/aring-qq.hpp:6,
                 from /home/mahrud/Projects/M2/quickfix/M2/Macaulay2/e/aring-translate.hpp:18,
                 from /home/mahrud/Projects/M2/quickfix/M2/Macaulay2/e/aring-glue.hpp:7,
                 from /home/mahrud/Projects/M2/quickfix/M2/Macaulay2/e/gbring.cpp:9:
/home/linuxbrew/.linuxbrew/opt/flint/include/flint/longlong.h:74:9: warning: "__ll_B" redefined
   74 | #define __ll_B ((ulong) 1 << (FLINT_BITS / 2))
      |         ^~~~~~
In file included from /home/linuxbrew/.linuxbrew/Cellar/givaro/4.2.0_1/include/recint/recdefine.h:145,
                 from /home/linuxbrew/.linuxbrew/Cellar/givaro/4.2.0_1/include/recint/ruruint.h:43,
                 from /home/linuxbrew/.linuxbrew/Cellar/givaro/4.2.0_1/include/recint/rumanip.h:42,
                 from /home/linuxbrew/.linuxbrew/Cellar/givaro/4.2.0_1/include/recint/ruconvert.h:21,
                 from /home/linuxbrew/.linuxbrew/Cellar/givaro/4.2.0_1/include/gmp++/gmp++_int.h:26,
                 from /home/linuxbrew/.linuxbrew/Cellar/givaro/4.2.0_1/include/gmp++/gmp++.h:57,
                 from /home/linuxbrew/.linuxbrew/Cellar/givaro/4.2.0_1/include/givaro/givinteger.h:19,
                 from /home/linuxbrew/.linuxbrew/Cellar/givaro/4.2.0_1/include/givaro/givpoly1dense.h:22,
                 from /home/linuxbrew/.linuxbrew/Cellar/givaro/4.2.0_1/include/givaro/givpoly1.h:38,
                 from /home/mahrud/Projects/M2/quickfix/M2/BUILD/build/usr-host/include/fflas-ffpack/ffpack/ffpack.h:38,
                 from /home/mahrud/Projects/M2/quickfix/M2/Macaulay2/e/aring-zzp-ffpack.hpp:17,
                 from /home/mahrud/Projects/M2/quickfix/M2/Macaulay2/e/aring-translate.hpp:17:
/home/linuxbrew/.linuxbrew/Cellar/givaro/4.2.0_1/include/recint/reclonglong.h:46:9: note: this is the location of the previous definition
   46 | #define __ll_B ((UWtype) 1 << (W_TYPE_SIZE / 2))
      |         ^~~~~~
```

d-torrance:
So for the msolve errors, we first get the following:

```m2
sh: line 1: 55919 Abort trap: 6           (core dumped) /usr/local/bin/msolve --help > /dev/null 2>&1
 -- warning: could not find msolve with version at least v0.7.0
```

This looks like it's happening during:
https://github.com/Macaulay2/M2/blob/e3edfa575d7d909bd792f371f1210e582891f079/M2/Macaulay2/packages/Msolve.m2#L40-L43

This hasn't been a problem until the last day or so, so it seems likely to be related to the FLINT transition.

But the actual error is coming later:

```m2
stdio:3:30:(3): error: msolve returned an error (32512):
sh: /usr/bin/msolve: No such file or directory
```

Why is it trying to call `/usr/bin/msolve`, when brew installed it in `/usr/local/bin`?  I think it's because of the next few lines:

https://github.com/Macaulay2/M2/blob/e3edfa575d7d909bd792f371f1210e582891f079/M2/Macaulay2/packages/Msolve.m2#L45-L48

`true` *is* located in `/usr/bin`/, so that's where `findProgram` thinks it found msolve.  I don't think we want to use `true` here.  Was the goal of this piece of code to support older versions of msolve?  (IIRC, they only added the `-V` option to query the version number in a pretty recent version.) 

---

Regarding the compiler warnings, those should be fixed by using the latest givaro (see https://github.com/Macaulay2/M2/issues/3697#issuecomment-2726945669).

mahrud:
> `true` _is_ located in `/usr/bin`/, so that's where `findProgram` thinks it found msolve.

LOL I didn't realize this was the behavior. Shouldn't it check the directory where the first argument is found?

d-torrance:
> Shouldn't it check the directory where the first argument is found?

This first argument may not be an actual program -- it's just a string identifier.  This is to deal with things like 4ti2 and TOPCOM that are more like suites of programs.

d-torrance:
Yay, the builds finally worked!  https://github.com/Macaulay2/M2/actions/runs/14023102376

Now let's remove the last commit and try with the Homebrew/Ubuntu packages...

mahrud (approved):

Issue #1894, opened 2021-01-28
comments in examples


d-torrance:
Comments are sometimes stripped from examples, and sometimes they aren't.  The behavior is also different between generated exampled and canned examples.

Consider the simple package below:

```m2
newPackage("Foo", Headline => "foo")

export {"foo"}

foo = () -> print "foo"

beginDocumentation()

doc ///
 Key
   foo
 Description
   Text
     Regular example
   Example
     -- foo
     elapsedTime 2 + 2
  Text
    Canned example
  CannedExample
    i1 : -- foo
         elapsedTime 2 + 2
     -- 3.33e-6 seconds elapsed

    o1 = 4
///
```

This gives us:
```m2
i3 : help "foo"

o3 = foo
     ***

     Description
     ===========

     Regular example

     +--------------------------+
     |i1 : elapsedTime 2 + 2    |
     | -- 3.5e-7 seconds elapsed|
     |                          |
     |o1 = 4                    |
     +--------------------------+

     Canned example

     +----------------------+
     |i1 : -- foo           |
     |     elapsedTime 2 + 2|
     |                      |
     |o1 = 4                |
     +----------------------+

     For the programmer
     ==================

     The object "foo" is a function closure.

o3 : DIV
```

d-torrance:
I think what's happening is that generated examples strip comments from the original code, but leave comments generated when running the example (e.g., `elapsedTime` and `commentize`).  Canned examples strip any comments that are preceded by only whitespace.

mahrud:
Stripping of commented lines happens very early in in SimpleDoc processing:
https://github.com/Macaulay2/M2/blob/a9b2a9c3af77eec2d660de1320da32b980eac518/M2/Macaulay2/packages/SimpleDoc.m2#L51

DanGrayson:
I no longer remember why Mike and I wanted to strip comments out.  One argument in favor of doing so is that package documentation would be clearer if such comments were incorporated into the surrounding prose.  One argument against doing so is that we might be documenting how to put comments in code.

d-torrance:
Non-`SimpleDoc` examples have different behavior still.  `EXAMPLE` won't accept any comments:

```m2
document{
    Key => "bar",
    EXAMPLE lines ///
        -- bar
        2 + 2
        ///
}
```
results in:
```m2
 -- making example results for "bar"                                         -- 0.495855 seconds elapsed
stdio:15:1:(3): error: example results terminate prematurely: bar
```

[...]

PR #3702, opened 2025-03-20
Don't put "-* no-capture-flag *-" on its own line

d-torrance:
Otherwise, the documentation will be garbled.

When not-capturing an example, we remove any instances of the string `"-* no-capture-flag *-"`.  If this string is by itself on a line, then we end up with an empty line in the example input, which messes everything up.

One example is the [findProgram](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_find__Program.html) docs.   This is out of order:

```m2
     +----------------------------------------------------------------------------+
     |  i3 : findProgram "normaliz"                                               |
     |                                                                            |
     |  o3 = normaliz                                                             |
     |                                                                            |
     |  o3 : Program                                                              |
     +----------------------------------------------------------------------------+

     If cmd is not provided, then cmd is run with the common --version command
     line option.
```

After the change:

```m2
     If cmd is not provided, then name is run with the common --version command
     line option.

     +-------------------------------+
     |  i3 : findProgram "normaliz"  |
     |                               |
     |  o3 = normaliz                |
     |                               |
     |  o3 : Program                 |
     +-------------------------------+
```

It might be more elegant to also remove the newline in the special case where `-* no-capture-flag *-` is by itself on a line.  Allowing `-- no-capture-flag` would be cool too.  But that would require some fancier regexes.  But right now, this only affects Macaulay2Doc.  Maybe something to think about if/when we start capturing the other packages' examples.

### DIFF
--- a/M2/Macaulay2/packages/Macaulay2Doc/functions/debug-doc.m2
+++ b/M2/Macaulay2/packages/Macaulay2Doc/functions/debug-doc.m2
@@ -27,8 +27,7 @@ Node
     Text
       This function allows access to the local symbols of a any loaded file.
     Example
-      -* no-capture-flag *-
-      load "Macaulay2Doc/demos/demo1.m2"
+      load "Macaulay2Doc/demos/demo1.m2" -* no-capture-flag *-
       listUserSymbols
       debug "Macaulay2Doc/demos/demo1.m2"
       listUserSymbols
--- a/M2/Macaulay2/packages/Macaulay2Doc/functions/findProgram-doc.m2
+++ b/M2/Macaulay2/packages/Macaulay2Doc/functions/findProgram-doc.m2
@@ -137,11 +137,10 @@ doc ///
       file, then @TT "name"@ should coincide with the name of this
       file.
     Example
-      -* no-capture-flag *-
-      programPaths#"gfan" = "/path/to/gfan/"
+      programPaths#"gfan" = "/path/to/gfan/" -* no-capture-flag *-
       gfan = findProgram("gfan", "gfan _version --help", Verbose => true)
     Text
-      If @TT "cmd"@ is not provided, then @TT "cmd"@ is run with the common
+      If @TT "cmd"@ is not provided, then @TT "name"@ is run with the common
       @TT "--version"@ command line option.
     Example
       findProgram "normaliz"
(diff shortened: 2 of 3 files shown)


d-torrance:
I played around with removing empty lines, and it still messes up the documentation.  For example, consider the following silly documentation node:

```m2
doc ///
  Key
    "foo"
  Description
    Text
      foo
    Example
      -* no-capture-flag *-
      print "foo"
    Text
      bar
    Example
      print "bar"
///
```
If we additionally remove empty lines in `captureExampleOutput` (I added ` replace("^\n", "", inputs)` since the comment is the first line in this case), then we still get:

```m2
i3 : help "foo"

o3 = foo
     ***

     foo

     +--------------------+
     |  i1 : print "foo"  |
     |  foo               |
     +--------------------+
     |  i2 : print "bar"  |
     |  bar               |
     +--------------------+

     bar

     +---------+
     |  i3 :   |
     +---------+
```
The only difference is that without removing that empty line, we had an extra newline right after `i1:`:

```m2
      +--------------------+
      |  i1 :              |
      |       print "foo"  |
      |  foo               |
      +--------------------+
```

I think the broader issue is the way that `processExamples` and `processExamplesLoop` sprinkles the example results in with the surrounding text.  The hope is that we encounter an `ExampleItem` in `docBody` exactly where we should, but comments and/or empty lines mess that up.

In this particular case, we were expecting:

1. text
2. example
3. example
4. text
5. example

But item 2 didn't produce any output, sliding item 3 into its slot, item 5 into item 3's slot, and giving us an empty input line in item 5's slot.

Probably this whole part of the code should be re-examined (see the TODO notes at the end of `examples.m2`).

Another issue is that SimpleDoc splits up the examples into buckets based on indentation, and maybe comments shouldn't get their own bucket.

See also #1894.

But long story short, I guess I'm proposing that we fix these few doc nodes with this band-aid fix for now, and maybe re-work how examples are incorporated into the documentation at a later date.

M2/Macaulay2/packages/Macaulay2Doc/functions/examples-doc.m2
```diff
@@ -46,7 +46,8 @@ Node
     Text
       Certain examples or tests might not be compatible with @TT "capture"@. In such cases,
       use the string @CODE "-* no-capture-flag *-"@ somewhere within @TT "str"@ so that the
-      code is run in a separate Macaulay2 subprocess instead.
+      code is run in a separate Macaulay2 subprocess instead.  In examples, this string
+      should not go on its own line or the documentation will be garbled.
```
mahrud, line 50: I don't think this needs to be mentioned in the documentation. It's a caveat that will hopefully be fixed at some point, and there's a good chance we'll forget to remove this line and years later someone (likely us) will wonder why the documentation will be garbled.

mahrud (approved):

Issue #3700, opened 2025-03-19
bug in hasMinor (returns false for two isomorphic matroids)


anna-brosowsky:
The matroid K4 can be inputted in two different ways, but when using the different descriptions I'm incorrectly getting false when checking of one is a minor of the other:

```
i1 : needsPackage "Matroids";

i2 : B = {{0,1,2},{0,1,3},{0,1,4},{0,1,5},{0,2,3},{0,2,5},{0,3,4},{0,4,5},{1,2,3},{1,2,4},{1,3,5},{1,4,5},{2,3,4},{2,3,5},{2,4,5},{3,4,5}};

i3 : M = matroid(toList (0..5),B);  K4 = matroid completeGraph 4;

i4 : areIsomorphic(M,K4)

o4 = true

i5 : hasMinor(M,K4)

o5 = false

i6 : hasMinor(K4,K4)

o6 = true
```

PR #3703, opened 2025-03-20
Matroids -- fixing hasMinor

anna-brosowsky:
For matroids of the same size, hasMinor now checks whether they are isomorphic instead of literal equality.

This fixes issue #3700


d-torrance:
Thanks!

@jchen419 - Does this look okay?

jchen419 (approved):
Looks good to me!

PR #3704, opened 2025-03-23
Add support for FLINT 3.2 (master branch)

d-torrance:
This cherry-picks the `gmp.h` commit from #3698 onto the `master` branch, since the macOS builds are now failing since FLINT 3.2 is available in Homebrew.  (See https://github.com/Macaulay2/M2/actions/runs/14016356099).

### DIFF
--- a/M2/Macaulay2/m2/programs.m2
+++ b/M2/Macaulay2/m2/programs.m2
@@ -89,18 +89,21 @@ findProgram(String, List) := opts -> (name, cmds) -> (
 	instance(opts.MinimumVersion, Sequence) and
 	class \ opts.MinimumVersion === (String, String)) then
 	error "expected MinimumVersion to be a sequence of two strings";
-    pathsToTry := fixPath \ join(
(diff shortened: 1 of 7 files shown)


M2/Macaulay2/m2/programs.m2
```diff
@@ -89,18 +89,21 @@ findProgram(String, List) := opts -> (name, cmds) -> (
 	instance(opts.MinimumVersion, Sequence) and
 	class \ opts.MinimumVersion === (String, String)) then
 	error "expected MinimumVersion to be a sequence of two strings";
-    pathsToTry := fixPath \ join(
+    pathsToTry := fixPath \ nonnull splice {
```
mahrud, line 92: Why not use `flatten`, then you don't need the couple of `toSequence` calls below.

d-torrance: Oh yeah -- of course!  Didn't occur to me

mahrud (approved):

Issue #3679, opened 2025-02-25
compilation warnings in the c++ engine


mikestillman:
On MacOS, when compiling under cmake and using apple clang++ (and maybe other compilers too), we get the following redefinition warnings.  @d-torrance, maybe we should change our config file to #define these with a "1"...?

```
In file included from /Users/mike/src/M2-current-branches/M2-sparse-matrices/M2/BUILD/mike/builds.tmp/cmake-appleclang/Macaulay2/d/python-exports.h:3:
/Users/mike/src/M2-current-branches/M2-sparse-matrices/M2/BUILD/mike/builds.tmp/cmake-appleclang/include/M2/config.h:60:9: warning: 'HAVE_ALARM' macro redefined [-Wmacro-redefined]
   60 | #define HAVE_ALARM
      |         ^
/opt/homebrew/Caskroom/miniforge/base/include/python3.12/pyconfig.h:67:9: note: previous definition is here
   67 | #define HAVE_ALARM 1
      |         ^
In file included from /Users/mike/src/M2-current-branches/M2-sparse-matrices/M2/Macaulay2/d/python-c.c:2:
In file included from /Users/mike/src/M2-current-branches/M2-sparse-matrices/M2/BUILD/mike/builds.tmp/cmake-appleclang/Macaulay2/d/python-exports.h:3:
/Users/mike/src/M2-current-branches/M2-sparse-matrices/M2/BUILD/mike/builds.tmp/cmake-appleclang/include/M2/config.h:67:9: warning: 'HAVE_ALLOCA_H' macro redefined [-Wmacro-redefined]
   67 | #define HAVE_ALLOCA_H
      |         ^
/opt/homebrew/Caskroom/miniforge/base/include/python3.12/pyconfig.h:73:9: note: previous definition is here
   73 | #define HAVE_ALLOCA_H 1
      |         ^
In file included from /Users/mike/src/M2-current-branches/M2-sparse-matrices/M2/Macaulay2/d/python-c.c:2:
In file included from /Users/mike/src/M2-current-branches/M2-sparse-matrices/M2/BUILD/mike/builds.tmp/cmake-appleclang/Macaulay2/d/python-exports.h:3:
/Users/mike/src/M2-current-branches/M2-sparse-matrices/M2/BUILD/mike/builds.tmp/cmake-appleclang/include/M2/config.h:82:9: warning: 'HAVE_DLFCN_H' macro redefined [-Wmacro-redefined]
[...]

PR #3705, opened 2025-03-23
Define various HAVE_* macros to 1 in cmake build

d-torrance:
This should fix macro redefinition warnings for macros that also defined in the Python headers.

Closes: #3679

### DIFF
--- a/M2/include/M2/config.h.cmake
+++ b/M2/include/M2/config.h.cmake
@@ -57,14 +57,14 @@
 
 // TODO: only used in Macaulay2/d/interrupts.d. Still needed?
 /* Defined if you have the `alarm' function. */
-#cmakedefine HAVE_ALARM
+#cmakedefine HAVE_ALARM 1
 
 // TODO: used a few places, what is it for?
 /* Define to 1 if you have `alloca', as a function or macro. */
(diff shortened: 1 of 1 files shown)


M2/include/M2/config.h.cmake
```diff
@@ -57,14 +57,14 @@
 
 // TODO: only used in Macaulay2/d/interrupts.d. Still needed?
 /* Defined if you have the `alarm' function. */
-#cmakedefine HAVE_ALARM
+#cmakedefine HAVE_ALARM 1
```
mahrud, line 60: Just to make sure, do you want these to be `#define blah 1` if defined and `/* #undef blah */` otherwise, or do you want `#define blah 0` if it's not defined? In that case you want `#cmakedefine01 blah` here. See [here](https://cmake.org/cmake/help/latest/command/configure_file.html#transformations).

mahrud (approved):

M2/include/M2/config.h.cmake
d-torrance: They should be undefined, not 0, since they're used in `#ifdef`'s and `#if defined(...)`'s.

PR #3706, opened 2025-03-24
Incidence cohomology

EmanuelaMarangone:
This package computes the sheaf cohomology of line bundles on the incidence correspondence, 
      the torus equivariant splitting type of the bundle of principal parts for $\mathcal{O}(d)$ on $\mathbb{P}^1$,
      the multiplication in the graded Han-Monsky representation ring, and checks the Weak Lefschetz property.

### DIFF
--- a/M2/Macaulay2/packages/IncidenceCorrespondenceCohomology.m2
+++ b/M2/Macaulay2/packages/IncidenceCorrespondenceCohomology.m2
@@ -17,7 +17,7 @@ newPackage(
 		  HomePage => "https://sites.google.com/view/ethan-reed/home"}
 	    },
     	Headline => "Cohomology on the incidence correspondence, bundle of principal parts, and Lefschetz properties",
-	Keywords => {"incidence correspondence"," Lefschetz properties", "principal parts bundles", "Han--Monsky representation ring"}
+	Keywords => {"Algebraic Geometry", "Commutative Algebra", "Flag Varieties"}
 	--Certification => {
 	  --   "journal name" => "",
 	    -- "journal URI" => "",
@@ -676,25 +676,22 @@ hasWLP(ZZ, List) := opts -> (p, L) -> (
      	     return j==0;)))
 
 
+
+
 -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+
 
 -- auxiliary function
 -- Input: integers n,s 
 -- output:list of n-tuples (a_1,....,a_n) such that a_1+.......+a_n-n=s and 2<=a_1<= ....<= a_n
-   
-   nTuples = (n, s) -> (
-    tuples:= {}; -- initialize an empty list to store the tuples
-    -- base case
-    if n == 1 then (
-        tuples = append(tuples, {s+1});
-        return tuples;)
-    else( -- recursive case
-    	for i from 2 to (s+n)//n do ( -- loop over possible values of the first component (from 1 to s)
-            subTuples := select(nTuples(n - 1, s - i+1), x -> x#0>=i); -- recursively generate (n-1)-tuples that sum to (s-i +n) and that the fist component is not less then i
-            iTuples := apply(subTuples, x -> {i}|x);-- append i as the first component to each sub-tuple
-            tuples = tuples | iTuples;
-    	    );
-	return tuples;))
+
+
+nTuples = (n, s) -> (
+    if n == 1 then {{s + 1}}
+    else flatten apply(toList(2..(s+n)//n), i -> (
+	    subTuples := select(nTuples(n - 1, s - i+1), x -> x#0>=i); -- recursively generate (n-1)-tuples that sum to (s-i +n) and that the fist component is not less then i
+	    apply(subTuples, x -> {i}|x)) -- append i as the first component to each sub-tuple
+	))
 
 
 -- user function:  this function lists all the monomial CI of socle degree s that fail the WLP 
@@ -708,6 +705,9 @@ monomialCIsWithoutWLP(ZZ, ZZ, ZZ) := opts -> (p, n, s) -> (
     L:=nTuples(n,s);--list of n-tuples (a_1,....,a_n) such that a_1+.......+a_n-n=s and 2<=a_1<= ....<= a_n
     select(L, x -> not hasWLP(p, x,opts))
 )
+
+
+
 
 --------------------------------------------------------
 ---- Strong Lefschetz Property (SLP)----------------------
@@ -802,8 +802,7 @@ hasSLP(ZZ, List) := opts -> (p, L) -> (
 		    else (
 			d:=L1#0;
 			d0:=d%p;
-			if sum L -n-(d-1)<=min{p-d0,d0} then true 
-			else  false
+			sum L -n-(d-1)<=min{p-d0,d0}
 			    )
 			)
 		    )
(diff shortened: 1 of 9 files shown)


M2/Macaulay2/packages/IncidenceCorrespondenceCohomology.m2
```diff
@@ -0,0 +1,2105 @@
...
+		  Email => "emanuela.marangone@umanitoba.ca", 
+		  HomePage => "https://emanuelamarangone.wixsite.com/marangone"}, 
+	    {Name => "Claudiu Raicu", 
+		  Email => "craicu@nd.edu", 
+		  HomePage => "https://www3.nd.edu/~craicu/"}, 
+	    {Name => "Ethan Reed",
+		  Email => "ereed4@nd.edu", 
+		  HomePage => "https://sites.google.com/view/ethan-reed/home"}
+	    },
+    	Headline => "Cohomology on the incidence corresponce, bundle of principal parts, and lefschetz properties for monomial complete intersections",
```
d-torrance, line 19: The Headline should be shortened, as we're currently getting the following build error:

```m2
../../../../../Macaulay2/m2/packages.m2:262:65:(1):[85]: error: newPackage: expected Headline to be less than 100 characters
```

M2/Macaulay2/packages/IncidenceCorrespondenceCohomology.m2
```diff
@@ -0,0 +1,2105 @@
...
+	Keywords => {"Character formulas", "cohomology of line bundles", "incidence correspondence"," Lefschetz properties", "principal parts", "Han--Monsky representation ring"}
```
d-torrance, line 20: I think this is too many keywords.  The main purpose of the `Keywords` option to `newPackage` is sort packages under the various headings on the [packages provided with Macaulay2](https://www.macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_packages_spprovided_spwith_sp__Macaulay2.html) documentation page.  Ideally, the keywords should come from existing headings on that page.

M2/Macaulay2/packages/IncidenceCorrespondenceCohomology.m2
```diff
@@ -0,0 +1,2105 @@
...
+
+--Input:
+--prime p and integer n
+--Output:
+--smallest integer e such that n < p^e
+
+boundq = (p, n) ->(
+    counter := 0; q := 1;
+    while n >= q do (counter = counter + 1; q = p*q);
+    return counter
```
d-torrance, line 71: It's not necessary to use `return` on the last line of a function.  (See https://github.com/Macaulay2/M2/wiki/Package-Writing-Style-Guide for this and lots of other stylistic pointers for packages.)

M2/Macaulay2/packages/IncidenceCorrespondenceCohomology.m2
```diff
@@ -0,0 +1,2105 @@
...
+-- output:list of n-tuples (a_1,....,a_n) such that a_1+.......+a_n-n=s and 2<=a_1<= ....<= a_n
+   
+   nTuples = (n, s) -> (
+    tuples:= {}; -- initialize an empty list to store the tuples
+    -- base case
+    if n == 1 then (
+        tuples = append(tuples, {s+1});
+        return tuples;)
+    else( -- recursive case
+    	for i from 2 to s+n//n do ( -- loop over possible values of the first component (from 1 to s)
```
d-torrance, line 693: `s + n//n` would always evaluate to just `s + 1`.  Is that the intent?

M2/Macaulay2/packages/IncidenceCorrespondenceCohomology.m2
```diff
@@ -0,0 +1,2105 @@
...
+            subTuples := select(nTuples(n - 1, s - i+1), x -> x#0>=i); -- recursively generate (n-1)-tuples that sum to (s-i +n) and that the fist component is not less then i
+            iTuples := for j from 0 to #subTuples -1 list  {i}| subTuples#j;-- append i as the first component to each sub-tuple
+            tuples = tuples | iTuples;
```
d-torrance, line 696: In general, repeatedly appending to a list inside a loop is a bad idea, since it generates a brand new list at every iteration.  Is it possible to re-factor this using `apply`,`for ... list`, or maybe modifying a mutable list in-place?

M2/Macaulay2/packages/IncidenceCorrespondenceCohomology.m2
```diff
@@ -0,0 +1,2105 @@
...
+		    else false)-- by Corollary 4.8  [Cook'11] if n>2 and p=2 we don't have the SLP
+		else( L1:=select(L, x -> x>=p);-- list the a_j bigger then p
+		    if #L1!=1 then  ( 
+			if #L==2 and #L1==2 then (return auxSLP2(p, L#0, L#1);)--if this is the case when n=2, p>2 and both a_1,a_2>=p, so we use auxSLP
+			else  false)-- if n>2 it is necessary that exactly one a_j is bigger then p
+		    else (
+			d:=L1#0;
+			d0:=d%p;
+			if sum L -n-(d-1)<=min{p-d0,d0} then true 
+			else  false
```
d-torrance, line 807: In general, `if x then true else false` can be simplified down to just `x`.

M2/Macaulay2/packages/IncidenceCorrespondenceCohomology.m2
```diff
@@ -0,0 +1,2105 @@
...
+--Computations of sheaf cohomology for divided powers of the cotangent sheaf on projective space
+-----------------------------------------------------------
+-----------------------------------------------------------
+
+--Input: prime integer p, and list of non-negative integers less than p list1
+--list1 corresponds to a positive integer written in base p
+--Output: the corresponding integer now in base 10
+basep10 = (p, list1) ->(
+    if list1 == {}
+    then return 0
```
d-torrance, line 834: I don't think you need the special case for the empty list here, since the `for` loop on the next line will return an empty list in this case, and `sum {}` will evaluate to 0 already.

d-torrance (changes requested):
Thanks for the contribution!  Please see the comments.

M2/Macaulay2/packages/IncidenceCorrespondenceCohomology.m2
d-torrance: This is still an issue -- the problem is that we keep re-creating `tuples` by appending to it with `|`.  Here's an alternate implementation using `apply`:

```m2
nTuples = (n, s) -> (
    if n == 1 then {{s + 1}}
    else flatten apply(toList(2..(s+n)//n), i -> (
	    subTuples := select(nTuples(n - 1, s - i+1), x -> x#0>=i);
	    apply(subTuples, x -> {i}|x))))
```

M2/Macaulay2/packages/IncidenceCorrespondenceCohomology.m2
d-torrance: This is still an issue.  The suggestion is to replace these two lines with just:

`sum L -n-(d-1)<=min{p-d0,d0}` (i.e., remove the `if` and `then true else false`).

d-torrance (changes requested):
Thanks for the quick updates!  There are still a few comments.

M2/Macaulay2/packages/IncidenceCorrespondenceCohomology.m2
EmanuelaMarangone: Sorry, I didn’t understand what you meant previously. Yes, I can use the existing keywords—"Algebraic Geometry," "Commutative Algebra," and "Flag Varieties" fit the package. Can I use all three, or do I need to choose just one?

d-torrance: Yeah, three is probably fine.

d-torrance (approved):
Looks good -- thanks the prompt responses!

Issue #3709, opened 2025-03-26
Negative size mutable matrix crashes M2 and allows for buffer overflow


leokayser:
While debugging my code I stumbled across the following funny behavior:
```Macaulay2
mutableMatrix(ZZ, 1, 1) --fine
mutableMatrix(ZZ,-1, 1) --crashes
mutableMatrix(ZZ, 1,-1) --crashes
mutableMatrix(ZZ,-1,-1) --fine
```
The error is ` *** out of memory trying to allocate -16 bytes, exiting *** `
I believe this is caused by the missing bound / non-negativity checks in line 87 in `Macaulay2/e/dmat.hpp` in the engine
https://github.com/Macaulay2/M2/blob/master/M2/Macaulay2/e/dmat.hpp#L87

The fact that one can instanciate a -a by -b matrix with (-a)*(-b)=ab entries itself leads to interesting and potentially dangerous side-effects.
```Macaulay2
A = mutableMatrix(ZZ,-2,-2)
A_(2,0)
A_(3,0)
...
```
will often return `0` or cause a segmentation fault, but sometimes random numbers from memory.

d-torrance:
Yikes!

`mutableIdentity` has a similar issue:

```m2
i1 : mutableIdentity(ZZ, -1)
-- SIGSEGV
-* stack trace, pid: 854694
 0# profiler_stacktrace(std::ostream&, int) at /usr/src/macaulay2-1.24.11+git202503252349-0ppa202503250209~ubuntu24.04.1/M2/Macaulay2/d/main.cpp:131
 1# segv_handler at /usr/src/macaulay2-1.24.11+git202503252349-0ppa202503250209~ubuntu24.04.1/M2/Macaulay2/d/main.cpp:243
 2# 0x000071E0F7245330 in /lib/x86_64-linux-gnu/libc.so.6
 3# rawMutableMatrixHash at interface/mutable-matrix.cpp:68
 4# basic_hash at /usr/src/macaulay2-1.24.11+git202503252349-0ppa202503250209~ubuntu24.04.1/M2/Macaulay2/d/basic.d:47
 5# basic_list_1 at /usr/src/macaulay2-1.24.11+git202503252349-0ppa202503250209~ubuntu24.04.1/M2/Macaulay2/d/basic.d:115
```

PR #3713, opened 2025-03-28
Raise an error when creating mutable matrices w/ negative dimensions

d-torrance:
We've actually been trying to do this for `mutableIdentity`, but never checked to see if `IM2_MutableMatrix_identity` returned null.

### Before

```m2
i1 : mutableIdentity(ZZ, -1)
-- SIGSEGV
-* stack trace, pid: 1080639
 0# profiler_stacktrace(std::ostream&, int) at /usr/src/macaulay2-1.24.11+git202503252349-0ppa202503250209~ubuntu24.04.1/M2/Macaulay2/d/main.cpp:131
 1# segv_handler at /usr/src/macaulay2-1.24.11+git202503252349-0ppa202503250209~ubuntu24.04.1/M2/Macaulay2/d/main.cpp:243
 2# 0x0000777B07645330 in /lib/x86_64-linux-gnu/libc.so.6
 3# rawMutableMatrixHash at interface/mutable-matrix.cpp:68
 4# basic_hash at /usr/src/macaulay2-1.24.11+git202503252349-0ppa202503250209~ubuntu24.04.1/M2/Macaulay2/d/basic.d:47
 5# basic_list_1 at /usr/src/macaulay2-1.24.11+git202503252349-0ppa202503250209~
```

```m2
i1 : mutableMatrix(ZZ, -1, 1)


 *** out of memory trying to allocate -16 bytes, exiting ***
```

### After
```m2
i1 : mutableIdentity(ZZ, -1)
stdio:1:15:(3): error: expected nonnegative integer

i2 : mutableMatrix(ZZ, -1, 1)
stdio:2:13:(3): error: expected nonnegative integers
```

Cc: @leokayser 

Closes: #3709

### DIFF
--- a/M2/Macaulay2/tests/normal/mutmat.m2
+++ b/M2/Macaulay2/tests/normal/mutmat.m2
@@ -59,3 +59,7 @@ Q = mutableMatrix id_(R^2)_P
 assert(L_(0, 1) == 0)
 assert(U_(1, 0) == 0)
 assert(Q * L * U == M)
+
+-- issue #3709 (these used to crash M2)
+try mutableMatrix(ZZ, -1, 1)
+try mutableIdentity(ZZ, -1)
--- a/M2/Macaulay2/tests/normal/mutmat.m2
+++ b/M2/Macaulay2/tests/normal/mutmat.m2
@@ -61,5 +61,7 @@ assert(U_(1, 0) == 0)
 assert(Q * L * U == M)
 
 -- issue #3709 (these used to crash M2)
-try mutableMatrix(ZZ, -1, 1)
-try mutableIdentity(ZZ, -1)
+assert try ( mutableMatrix(ZZ, -1, -1); false ) else true
+assert try ( mutableMatrix(ZZ, -1,  1); false ) else true
+assert try ( mutableMatrix(ZZ,  1, -1); false ) else true
+assert try ( mutableIdentity(ZZ, -1);   false ) else true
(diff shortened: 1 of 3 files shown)


M2/Macaulay2/tests/normal/mutmat.m2
```diff
@@ -59,3 +59,7 @@ Q = mutableMatrix id_(R^2)_P
 assert(L_(0, 1) == 0)
 assert(U_(1, 0) == 0)
 assert(Q * L * U == M)
+
+-- issue #3709 (these used to crash M2)
+try mutableMatrix(ZZ, -1, 1)
+try mutableIdentity(ZZ, -1)
```
mahrud, line 65: These aren't sufficient tests, because they could return something without crashing or throwing an error and the test would pass. Try this instead:
```suggestion
assert try ( mutableMatrix(ZZ, -1, -1); false ) else true
assert try ( mutableMatrix(ZZ, -1,  1); false ) else true
assert try ( mutableMatrix(ZZ,  1, -1); false ) else true
assert try ( mutableIdentity(ZZ, -1);   false ) else true
```

d-torrance: That commit suggestion feature is so cool!  You're even a co-author on the commit -- neat!

mahrud: Huh I didn't know that!

mahrud (approved):

leokayser (approved):

PR #3718, opened 2025-03-31
Organizing Macaulay2Doc #4

mahrud:
- **deprecated (vars, QuotientRing)**
- **fixed baseName over fraction field**
- **switched to using Boost::regex as header-only library**
- **removed unused variables in packages/CMakeLists.txt**
- **updated msolve in cmake**
- **added missing install commands for msolve and normaliz**
- **added locate(Package)**
- **added about(ScriptedFunctors)**
- **added `<kbd>` to ignore list for KaTeX**
- **updated documentation of html(TEX)**
- **fixed issue with missing node ancestors**
- **fixed problem with spacing of EXAMPLE PRE objects**
- **removed 'Description' header from package docs**
- **marked some Truncations nodes as undocumented**
- **documented some options of quotient'**
- **documented some missing nodes**
- **renamed Groebner to Gröbner, except in doc keys**
- **renamed hashtable to hash table in docs**
- **added deepSelect**
- **added unique(Hypertext)**
- **updated help documentation**


M2/Macaulay2/d/CMakeLists.txt
```diff
@@ -146,7 +146,8 @@ add_library(M2-interpreter OBJECT ${CLIST} ${CXXLIST} ${TAGS}
   $<$<BOOL:${WITH_XML}>:xml-c.c xml-c.h>
   $<$<BOOL:${WITH_PYTHON}>:python-c.c>)
 
-target_link_libraries(M2-interpreter PRIVATE M2-supervisor Boost::regex
+target_link_libraries(M2-interpreter PRIVATE M2-supervisor
+  Boost::boost # boost_regex is now header-only, so we don't link Boost::regex
```
d-torrance, line 150: Will this matter when building on older systems where it's not header-only?

mahrud: regex has been header-only since Boost 1.77 (~2021), so I'm not worried about it.

M2/Macaulay2/m2/packages.m2
```diff
@@ -581,6 +581,12 @@ debug GlobalDictionary := dict -> (
     if not isMember(dict, dictionaryPath) then dictionaryPath = prepend(dict, dictionaryPath);
     checkShadow())
 
+locate Package := pkg -> NumberedVerticalList nonnull (
```
d-torrance, line 584: I'm a little torn about the return value here.  On one hand, it seems that `locate` should return a single `FilePosition` object, but on the other hand, it does make sense to list all the files.

Is there a reason for `NumberedVerticalList`?  For things like `methods` and `about` where there's a `ZZ` method we can run right afterwards, that makes sense, but I"m wondering if just `VerticalList` would suffice?

mahrud: It's not in its final form, hence why it's not really mentioned in the documentation, but I wanted to put it out there in case others want to shape it differently.

There are very few occasions when it makes sense to use VerticalList but not NumberedVerticalList, and those are when you explicitly don't want additional numbers listed on the left. This wasn't a case like that.

d-torrance (approved):
Looks good -- thanks!

PR #3722, opened 2025-04-03
ConnectionMatrices: v1.0

mahrud:
This is a pull request for the ConnectionMatrices package, which grew out of the Macaulay2 in the Sciences workshop at MPI MiS in Leipzig. The package includes methods for computing connection matrices of D-ideals and testing integrable systems. A companion preprint is available [here](https://arxiv.org/abs/2504.01362).

cc: @p-goerlach @jkoefler @alsattelberger @HendSch @nicolas-alexander-weiss and @frazaffa


mahrud:
@d-torrance I will fix the "Infering" typos, but "ALS" are initials, not a typo. Do you know how to disable codespell in one file or silence it in one line (ideally without adding "ALS" to the global ignore list)

d-torrance:
There's apparently an inline ignore option, but it's halfway new (codespell 2.3.0, released last May).  So it's not in Ubuntu 24.04.  I suppose we can install it with pip or something.  I'll play around with it

d-torrance:
So yeah, installing the latest codespell with pip is straightforward. But now it catches a ton more "errors"...

https://github.com/d-torrance/M2/actions/runs/14234556826/job/39891423019

mahrud:
Oh that's too much, I'll just expand ALS.

d-torrance (approved):
Thanks for the contribution!

PR #3723, opened 2025-04-04
Link against gtest instead of building it (autotools)

d-torrance:
The source isn't available in all platforms' googletest packages (e.g., Fedora), and the recommendation to compile unit tests using gtest-all.cc has been deprecated (or at least it doesn't seem to be mentioned in the documentation).

This brings the autotools build's behavior in line with the cmake build, which already links against googletest.

This is a draft for now, as it depends on three additional PR's:

https://github.com/Macaulay2/memtailor/pull/9
https://github.com/Macaulay2/mathic/pull/12
https://github.com/Macaulay2/mathicgb/pull/45


mikestillman (approved):
I cannot easily vet the autotools code, but overall this is good.  I was always annoyed that the recommendation was to compile with gtest...

Thanks for doing this!

PR #3724, opened 2025-04-04
Find path to Python shared library (autotools)

d-torrance:
If we can't find it using the default search path, then query the LIBDIR config variable from the sysconfig module, add it to LDFLAGS, and try again.

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -155,15 +155,12 @@ jobs:
         if: matrix.build-system == 'autotools'
         run: |
           ../../autogen.sh
-          export PYVERSION=`python3 -c "from sys import version_info; \
-            print(f'{version_info.major}.{version_info.minor}')"`
           export CPPFLAGS="-I`brew --prefix`/include \
             -I`brew --prefix libomp`/include         \
             -I`brew --prefix readline`/include"
           export  LDFLAGS="-L`brew --prefix`/lib \
             -L`brew --prefix libomp`/lib         \
-            -L`brew --prefix readline`/lib       \
-            -L/Library/Frameworks/Python.framework/Versions/${PYVERSION}/lib"
+            -L`brew --prefix readline`/lib"
           export F77=gfortran-14
           ../../configure --enable-download --with-system-gc
 
--- a/M2/configure.ac
+++ b/M2/configure.ac
@@ -1385,9 +1385,17 @@ then AC_DEFINE(WITH_PYTHON,1,[whether we are linking with the python library])
 	print(get_config_var('CONFINCLUDEPY'))"`
     CPPFLAGS="$CPPFLAGS -I$PYTHON_INCLUDE_DIR"
     AC_CHECK_HEADER(Python.h,,AC_MSG_ERROR(Python.h not found))
-    AC_SEARCH_LIBS(Py_Initialize,
-	python${PYTHON_VERSION} python${PYTHON_VERSION}m,,
-	AC_MSG_ERROR(libpython not found))
+    AC_SEARCH_LIBS([Py_Initialize],
+	[python${PYTHON_VERSION} python${PYTHON_VERSION}m],,
+	[PYTHON_LIBDIR=`$PYTHON_BIN -c \
+	    "from sysconfig import get_config_var; \
+	    print(get_config_var('LIBDIR'))"`
+	 AC_MSG_NOTICE([adding -L$PYTHON_LIBDIR to LDFLAGS and trying again...])
+	 AS_UNSET([ac_cv_search_Py_Initialize])
+	 LDFLAGS="$LDFLAGS -L$PYTHON_LIBDIR"
+	 AC_SEARCH_LIBS([Py_Initialize],
+	     [python${PYTHON_VERSION} python${PYTHON_VERSION}m],,
+	     [AC_MSG_ERROR(libpython not found)])])
 fi
 
 dnl boost 1.65 was the first release containing the stacktrace library


mikestillman (commented):
I am interested to see if this improves the situation trying to include Python on my macos autotools build.

I can either accept it, and see how it works, or try it out first.

d-torrance:
It would be great if you gave it a try -- I'm curious.  It did make it so that we don't need to include any special Python build flags for the macOS GitHub builds, so I'm hopeful!

mikestillman:
I rebased this on top of the current development, I'm on MacOS (latest 15.4.1) and here is the configure I use (autotools):
```
arm64-appleclang-python : always
	mkdir -p builds.tmp/arm64-appleclang-python
	cd builds.tmp/arm64-appleclang-python; ../../../../configure \
	--enable-download --with-system-gc \
	--with-system-memtailor --with-system-mathic \
	--with-system-mathicgb \
	LDFLAGS="-L`brew --prefix`/lib  \
		-L`brew --prefix factory`/lib \
		-L`brew --prefix libomp`/lib  \
		-L`brew --prefix readline`/lib \
		-L`brew --prefix python`/Frameworks/Python.framework/Versions/3.13/lib" \
	CPPFLAGS="-I`brew --prefix`/include   \
	    -I`brew --prefix`/include/cddlib    \
		-I`brew --prefix factory`/include   \
		-I`brew --prefix libomp`/include    \
		-I`brew --prefix readline`/include" \
	F77="gfortran-14" \
	PKG_CONFIG_PATH="`brew --prefix factory`/lib/pkgconfig"

```

When running this, I get:
```
...
checking for ffi.h... yes
checking for library containing ffi_call... -lffi
checking for ffi_get_struct_offsets... yes
checking whether nm accepts the demangle option... yes
checking whether getaddrinfo can handle numeric service (port) numbers... configure: error: test file failed to compile
make: *** [arm64-appleclang-python] Error 1
bash-3.2$ 
```

This is possibly a libffi problem?  I can't tell in `config.log` which one is really causing the problem.  @d-torrance Any suggestions?

mikestillman:
Addendum: the configure works fine if I disable pyrhon in configure line

d-torrance:
It seems like it's happening after the libffi check, but inside the getaddrinfo check.  Would you send me your config.log?

Whenever we get past that issue, try removing `-L`brew --prefix python`/Frameworks/Python.framework/Versions/3.13/lib"` from command line.  Ideally, that *should* be detected automatically after these proposed changes.

mikestillman:
I tried adding in DYLD_LIBRARY_PATH with the explicit directory for libffi.8.dylib, but get the same error.

```
arm64-appleclang-python : always
	mkdir -p builds.tmp/arm64-appleclang-python
	cd builds.tmp/arm64-appleclang-python; DYLD_LIBRARY_PATH="/opt/homebrew/Cellar/libffi/3.4.8/lib/:$DYLD_LIBRARY_PATH" ../../../../configure \
	--enable-download --with-system-gc \
	--with-system-memtailor --with-system-mathic \
	--with-system-mathicgb \
	--with-gtest-include-path="`brew --prefix googletest`/include" \
	--with-gtest-source-path="`brew --prefix googletest`/include/googletest/googletest" \
	LDFLAGS="-L`brew --prefix`/lib  \
		-L`brew --prefix factory`/lib \
		-L`brew --prefix libffi`/lib \
		-L`brew --prefix libomp`/lib  \
		-L`brew --prefix readline`/lib \
		-L`brew --prefix python`/Frameworks/Python.framework/Versions/Current/lib" \
	CPPFLAGS="-I`brew --prefix`/include   \
	    -I`brew --prefix`/include/cddlib    \
		-I`brew --prefix factory`/include   \
		-I`brew --prefix libomp`/include    \
		-I`brew --prefix readline`/include \
		-I`brew --prefix python`/Frameworks/Python.framework/Versions/Current/include" \
	F77="gfortran-14" \
	PKG_CONFIG_PATH="`brew --prefix factory`/lib/pkgconfig"
```

mikestillman:
It does work on development

mikestillman:
Here is what finally worked.  I uninstalled brew's version of miniforge.  Even after I commented it out in my shell profile file (.bash_profile, in my case), the configure script still found it.  Unfortunately, it didn't really find it (actually, I'm a little unclear where the error then comes from), and gives errors later in that it can't compile `getaddrinfo` (really, it can't find a library).

After uninstalling, using the following configure entry in BUILD/mike/Makefile (or rather, I will check this in later), the following configures, and then compiles the executable correctly.  I will approve this PR. @d-torrance Do I really need all of these items, I wonder?

```. 
arm64-appleclang-python : always
    mkdir -p builds.tmp/arm64-appleclang-python
    cd builds.tmp/arm64-appleclang-python;  ../../../../configure \
    --enable-download --with-system-gc \
    --with-system-memtailor --with-system-mathic \
    --with-system-mathicgb \
    LDFLAGS="-L`brew --prefix`/lib  \
        -L`brew --prefix factory`/lib \
        -L`brew --prefix libffi`/lib \
        -L`brew --prefix libomp`/lib  \
        -L`brew --prefix readline`/lib" \
    CPPFLAGS="-I`brew --prefix`/include   \
        -I`brew --prefix`/include/cddlib    \
        -I`brew --prefix factory`/include   \
        -I`brew --prefix libomp`/include    \
        -I`brew --prefix readline`/include" \
    F77="gfortran-14" \
    PKG_CONFIG_PATH="`brew --prefix factory`/lib/pkgconfig"
```

mikestillman (approved):
After removing miniforge from homebrew, this works.  Actually, I wonder if I should check development branch and see if it works correctly too?

mikestillman:
@d-torrance I just checked.  Without these changes, but after removing miniforge, the current development version before this code, fails to find python.  So let's merge this!

PR #3726, opened 2025-04-09
And mathML(Complex) and related updates

d-torrance:
We introduce a new type, `ComplexExpression`, to unify the printing of both the legacy `ChainComplex` and the newer `Complex` types.  In particular, the latter didn't have a `mathML` method, and so they previously weren't supported in TeXmacs or Jupyter when using TeXmacs mode:

image

Now, `net`, `texMath`, and `mathML` just call `expression` first, and so we only need to implement each once and they'll work for both types:

image

A couple other related changes:

* We add `short(Module)` for use in complex expressions.  It uses `moduleAbbrv` (which has been slightly rewritten to just return null instead of having a backup representation) to use the module's variable name when possible, unless the module is free over a ring that's been assigned to a variable, in which case we use that respresentation.
* We try  to use `substitute` when building a legacy complex using `chainComplex` if the differential maps are given over different rings.  This is to fix some issues with `value(ComplexExpression)`, where the differential maps are `MatrixExpression` objects that may have forgotten some important info like what rings they were over.  `value(ComplexExpression)` isn't perfect -- it still fails in other situtations like when we've lost degree information.


pzinn:
Looking good. I tried your code on M2web and I get:
Screenshot from 2025-04-09 15-11-10
I wonder how it will look after your PR?

d-torrance:
It *should* look the same.  `texMath(ComplexExpression)` is essentially identical to the old `texMath(ChainComplex)`.  There were some subtle differences between that and `texMath(Complex)` in how to "shorten" the differential maps.  The `ChainComplex` version used `short(Matrix)` and the `Complex` version rolled its own `texMatrixShort`, but they were pretty similar, so I just went with `short(Matrix)` for everything.

pzinn:
great. yeah `texMatrixShort` was supposed to be gone but it somehow resurfaced in `Complex`...

pzinn:
one potential issue (I'm not sure if that's the one you're facing in your example) is that you probably don't want to take the expression (`MatrixExpression`) of a huge matrix since it literally spits out the list of entries (even if eventually it shortens it as in the current display in `WebApp` mode).

d-torrance:
Lol yeah -- that's exactly what's happening:

```m2
i12 : profileSummary

o12 = #run     %time   position                                                       
      1        93.71   src/macaulay2/M2/M2/Macaulay2/m2/chaincomplexes.m2:135:4-137:51
      16       93.7    src/macaulay2/M2/M2/Macaulay2/m2/matrix.m2:217:4-217:42        
      4586444  44.17   src/macaulay2/M2/M2/Macaulay2/m2/polyrings.m2:205:61-205:73    
```

4.5 million calls to `expression 0`...

pzinn:
on `vanilla`, I resolved this by having `short` act at the level of `Matrix` directly -- but I think you've reached the same conclusion. the only problem being, `value` won't work any more -- but I think that's an acceptable price to pay.

mahrud:
What's the point of an intermediary type ComplexExpression? I want to remove SheafExpression already.

d-torrance:
> What's the point of an intermediary type ComplexExpression? I want to remove SheafExpression already.

For the time being, while we have two distinct complex types with `Complex` and `ChainComplex`, it gives us a way to write the various printing methods just once.

Also, based on the [Expression](https://www.macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/___Expression.html) docs, it seems like having expression classes for mathematical objects is part of the design of the language:

> they are an intermediate phase in the conversion of a mathematical object to a net that can be printed

What's the problem with `SheafExpression`?

pzinn:
I think the only problem with your `short(Matrix)` is that it doesn't correctly treat zero matrices? (comparing quickly with the code on vanilla which I never got around to PRing)

d-torrance:
Oh ok -- I didn't think about zero matrices.  I just used your commit instead.

pzinn:
interesting. this is a good opportunity to clean up/debug this code...

M2/Macaulay2/m2/modules.m2
```diff
@@ -249,6 +242,15 @@ toString Module := M -> toString expression M
...
 texMath Module := M -> texMath expression M
 
+-- returns null if can't shorten (free module or assigned to variable)
+moduleAbbrv = M -> (
+    if isFreeModule M and hasAttribute(ring M, ReverseDictionary)
+    then expression M
+    else if hasAttribute(M, ReverseDictionary)
+    then getAttribute(M, ReverseDictionary))
+
+short Module := M -> moduleAbbrv M ?? expression M
```
mahrud, line 252: I don't understand why you removed the second input of `moduleAbbrv`, when this could have been implemented as:
```suggestion
short Module := M -> moduleAbbrv(M, expression M)
```

d-torrance: I've always thought that `moduleAbbrv`'s two inputs were kind of clunky.  For example, in `Matrix.AfterPrint`, we don't need the second input at all -- if we can't abbreviate the module, then we just don't print anything.  I figured this was an opportunity to update it.

d-torrance:
@sashahbc: Sorry to keep bugging you with these `ToricHigherDirectImages` issues!

Test 7 is failing on the builds for this PR:

```m2
i9 : assert(presentation RD2 == matrix {{y_1,0},{0,y_1^2},{0,0}})
stdio:10:6:(3): error: assertion failed
```

The rows were permuted a bit from the expected matrix:

```m2
i55 : presentation RD2

o55 = | 0   0     |
      | y_1 0     |
      | 0   y_1^2 |

              3      2
o55 : Matrix Q  <-- Q
```

From what I can tell, `RD2` is the first homology of a complex generated by the unexported `HDIComplex` method.  And its main loop is:

```m2
    directSum for mu in keys eigenchars list (
```
Since `eigenchars` is a hash table, we're not sure what order its keys will appears in the list returned by `keys`.  Could that be the issue?

sashahbc:
Yeah that's definitely the problem. Not sure what the best solution would be. I was hoping ```components``` would detect the different summands but once we've taken homology I guess it loses track of that.

sashahbc:
It seems like a bad idea because it's a probabilistic test but could we do
```assert( isIsomorphic( RD2, coker matrix {{y_1,0},{0,y_1^2},{0,0}} )_0 )```?

**Comment by d-torrance** [comment]
I suppose that would work. Is it okay that the order of the rows isn't deterministic, though?

**Comment by sashahbc** [comment]
Yeah I am a bit uncomfortable with the order changing. Mahrud had a suggestion which I'll make a PR for in a sec.

File: M2/Macaulay2/m2/expressions.m2 [hunk 2]
```diff
@@ -550,25 +550,46 @@ toString'(Function, SparseMonomialVectorExpression) := (fmt,v) -> toString (
 -----------------------------------------------------------------------------
 MatrixExpression = new HeaderType of Expression
 MatrixExpression.synonym = "matrix expression"
-matrixOpts1 := new OptionTable from {Blocks=>null,Degrees=>null,MutableMatrix=>false};
+matrixOpts1 := new OptionTable from {
+    Blocks => null,
+    Degrees => null,
+    MutableMatrix => false,
+    zero => null};
```
pzinn, line 557: any reason `zero` is the only lowercase option? and incidentally which type is this `zero`?

d-torrance: I went with `zero` since it's already exported.  When it's non-null, it will be a sequence of two modules (the target and source).

M2/Macaulay2/m2/matrix.m2
```diff
@@ -215,6 +215,7 @@ protect Blocks
 blockMatrixForm=false;  -- governs expression Matrix inclusion of blocks
 expression Matrix := m -> (
     x := applyTable(entries m, expression);
+    if #x == 0 then x = {symbol zero => (target m, source m)};
```
pzinn, line 218: I don't understand why this line is not above the previous one (why go through all the entries if the matrix is zero anyway?)

pzinn: err, I mean, and why the test isn't simply `x==0`

d-torrance: Great point!

d-torrance: Ok, I've updated this so `expression` of a zero matrix always just uses the short form and never deals with the individual entries.

M2/Macaulay2/m2/expressions.m2
pzinn: I meant, what is the class of `zero` itself, not the option labelled zero.
usually it's nice if all options have the same type; which one is a matter of taste, it seems (all symbols seems the most common choice, though there's serious overhead in doing that; I would think all strings would actually make more sense, but it's rarely used. I digress)

d-torrance: This is an excellent question!  I thought I'd put `symbol zero` here, but I didn't.  And `zero` is a function closure, which *should* have raised an error...

```m2
i1 : new OptionTable from {zero => null}
stdio:1:0:(3): error: expected option key to be a symbol, a string, or a type
```

d-torrance: Ohhh I figured it out -- `zero` is still a symbol at this point since it's not defined as a function until integers.m2, which is later in the load sequence.  (Same with `MutableMatrix` -- it's a type defined in mutablemat.m2, but it's a symbol here).

I'll change this so they're both explicitly symbols, just in case any code gets moved around

pzinn (approved):

d-torrance:
Yeah, with both complex types moving to their own packages, I don't know if having a shared expression type in Core makes sense any more...

I think I'll trim this down to just the `MatrixExpression` changes and adding `mathML(Complex)`.

M2/Macaulay2/m2/chaincomplexes.m2
```diff
@@ -137,15 +137,15 @@ net ChainComplex := C -> (
 	  a := s#0;
 	  b := s#-1;
 	  horizontalJoin between(" <-- ", apply(a .. b,i ->
-		  stack (net moduleAbbrv(C_i, C_i)," ",net i)))))
```
mahrud, line 140: This file is moved in the Complexes PR, so you might as well rebase on top of that PR I think.

mikestillman:
This is probably not relevant to what you are doing here, but it is something that is a problem.
```
 M = random(ZZ^2000, ZZ^2000);
13 M
```

or even 
```
try(13 M; false) else true
```
This last one takes about 30 seconds on my laptop...

mikestillman (approved):
This looks good to me.  Thanks! It appears that much of the discussion in the PR is related to pre-chain complex movement to OldChainComplexes, Complexes.  Sorry about that!

@pzinn Is there a regression in the web printing of Complexes?  Are we using something that you got rid of, without our noticing?

PR #3727, opened 2025-04-10
Update M2-emacs commit w/ new M2-comint-fix-unclosed function

d-torrance:
We update the M2-emacs submodule to get the changes from https://github.com/Macaulay2/M2-emacs/pull/82, which closes any unclosed strings or comments from the previous output.


mahrud (commented):
I'd rather wait for the release in case there are more updates.

d-torrance:
Fair enough. At minimum, we'll have updated symbols and version number.

PR #3729, opened 2025-04-12
Update ToricHigherDirectImages test so that order doesn't matter

d-torrance:
See https://github.com/Macaulay2/M2/pull/3683#issuecomment-2781518437.  This failing test is also affecting #3726 now.

Cc: @sashahbc


sashahbc (commented):
Looks good to me, sorry for the confusion.

PR #3731, opened 2025-04-15
Revert to old code for computing deltaE

nilten:
The previous version of the package computed the polytope "deltaE" of weights on which all cohomology groups are supported by just taking the convex hull of the weights of generators for the bundle as we range over all charts. I don't know of any mathematical justification of this, and this implementation was likely due to a communication error.

The implementation of deltaE has been replaced by older code that computes the correct thing. In e.g. the Klyachko case, the filtrations induce a natural stratification of the cocharacter lattice on which the graded pieces of the cohomology are independent of the weight. This version of deltaE takes the convex hull of all bounded regions.

### DIFF
--- a/M2/Macaulay2/e/NCAlgebras/FreeAlgebra.hpp
+++ b/M2/Macaulay2/e/NCAlgebras/FreeAlgebra.hpp
@@ -9,7 +9,6 @@
 #include "ringelem.hpp"               // for ring_elem
 #include "style.hpp"                  // for GEOHEAP_SIZE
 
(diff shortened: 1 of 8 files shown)


d-torrance:
The builds are failing with the following:

```m2
 error: documentation key for "deltaE(ToricVectorBundle)" encountered, but no method installed
```

So it looks like the package documentation needs to be updated to account for this change.

Also for future reference, pull requests should target the `development` branch instead of `master`.  I went ahead and updated it for you.

d-torrance (approved):
Thanks!

PR #3732, opened 2025-04-15
Added sorting to eigencharacter keys

sashahbc:
Related https://github.com/Macaulay2/M2/pull/3726#issuecomment-2807418694

Per Mahrud's suggestion, I added sorting to the keys of the hashtable (which are always integer valued matrices). ```sortBy``` gets imported to make things easier. I changed the tests to reflect this.

### DIFF
--- a/M2/Macaulay2/packages/ToricHigherDirectImages.m2
+++ b/M2/Macaulay2/packages/ToricHigherDirectImages.m2
@@ -1060,7 +1060,7 @@ phi = map(Y,X,matrix{{1,0}})
 assert(isWellDefined phi)
 D = {0,-2,0,0,0,-2,1,0}
 HT = computeEigencharacters(phi,1,D);
-assert((sortBy entries) keys HT == {matrix{{0},{1}}, matrix{{0},{-1}}})
+assert(set keys HT == set {matrix{{0},{1}}, matrix{{0},{-1}}})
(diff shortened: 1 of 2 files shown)


M2/Macaulay2/packages/ToricHigherDirectImages.m2
```diff
@@ -1059,8 +1060,7 @@ phi = map(Y,X,matrix{{1,0}})
 assert(isWellDefined phi)
 D = {0,-2,0,0,0,-2,1,0}
 HT = computeEigencharacters(phi,1,D);
-chars = keys HT
-assert(set chars == set {matrix{{0},{1}}, matrix{{0},{-1}}})
+assert((sortBy entries) keys HT == {matrix{{0},{1}}, matrix{{0},{-1}}})
```
mahrud, line 1063: Did you mean to change this from checking set equality? sortBy isn't exported, so the test doesn't have access to it.

sashahbc: I did, but I didn't realize this fact. I changed it back.

mahrud:
I think something went wrong with the commits, but @d-torrance or I can squash to PR into a single commit.

d-torrance (approved):
Thanks again!

Issue #1774, opened 2021-01-02
Feature request: partial parallel assignment


mahrud:
Basically, would be good to get this to work:
```m2
i1 : (a, b, c) = (1, 2, 3)

o1 = (1, 2, 3)

o1 : Sequence

i2 : (a, , c) = (1, 2, 3)
-*dummy position*- error: syntax error: parallel assignment expected symbol
```
Usually it's just a convenient thing when you don't want to use an extra symbol, but sometimes the convenience is really worth it.

DanGrayson:
That's a good idea.

mahrud:
Same with this: (which is a strange error, by the way)
```m2
i1 : h = new MutableHashTable from {};

i2 : (h#0, h#1) = (0, 1);
stdio:2:3:(3): error: syntax error: parallel assignment expected symbol list
stdio:2:8:(3): error: syntax error: parallel assignment expected symbol
```
Also, some languages have a more syntax like this:
```m2
L = (0, 1, 2, 3, 4)
(a, b.., c, d) = L -- a=0, b=(1, 2), c=3, d=4
(a, b, c.., d) = L -- a=0, b=1, c=(2, 3), d=4
(a, b,  .., d) = L -- a=0, b=1,           d=4
```

d-torrance:
Following up the discussion from #3614, another nice parallel assignment enhancement would be *augmented* parallel assignment, e.g., it would be great if this worked:

```m2
i1 : (a, b, c) = (1, 2, 3)

o1 = (1, 2, 3)

o1 : Sequence

i2 : (a, b, c) += (4, 5, 6)

o1 = (5, 7, 9)

o1 : Sequence
```

d-torrance:
Fixed by #3748

PR #3733, opened 2025-04-16
Document that we need to use installMethod for >>

d-torrance:
`X >> Y := f` parses as `X >> (Y := f)` due to precedence.

### DIFF
--- a/M2/Macaulay2/m2/help.m2
+++ b/M2/Macaulay2/m2/help.m2
@@ -298,15 +298,16 @@ getOperator := key -> if operator#?key then (
     op := toString key;
     if match("^[[:alpha:]]*$", op) then op = " " | op | " ";
     fixup DIV (
-	if binary#?key and key =!= symbol ?? then {
-	    PARA {"This operator may be used as a binary operator in an expression like ", TT ("x" | op | "y"), ". ",
-		"The user may install ", TO "Macaulay2Doc :: binary methods", " for handling such expressions with code such as"},
-	    PRE if key === symbol SPACE
-	    then "         X Y := (x,y) -> ..."
-	    else "         X "|op|" Y := (x,y) -> ...",
-	    PARA {"where ", TT "X", " is the class of ", TT "x", " and ", TT "Y", " is the class of ", TT "y", "."}},
-	if key === symbol ?? then { -- can't install binary methods
-	    PARA {"This operator may be used as a binary operator in an expression like ", TT ("x" | op | "y"), ". "}},
+	if binary#?key then {
+	    PARA {"This operator may be used as a binary operator in an expression like ", TT ("x" | op | "y"), "."}},
+	    -- ?? binary methods can't be installed
+	    -- >> needs to use installMethod due to precedence
+	    if not isMember(key, {symbol ??, symbol >>}) then {
+		PARA {"The user may install ", TO "Macaulay2Doc :: binary methods", " for handling such expressions with code such as"},
+		PRE if key === symbol SPACE
+		then "         X Y := (x,y) -> ..."
+		else "         X "|op|" Y := (x,y) -> ...",
+		PARA {"where ", TT "X", " is the class of ", TT "x", " and ", TT "Y", " is the class of ", TT "y", "."}},
 	if prefix#?key then {
 	    PARA {"This operator may be used as a prefix unary operator in an expression like ", TT (op | "y"), ". ",
 		"The user may ", TO2{ "Macaulay2Doc :: installing methods", "install a method" }, " for handling such expressions with code such as"},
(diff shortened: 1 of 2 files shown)


d-torrance:
I bumped the precedence up, i.e.,:

```diff
--- a/M2/Macaulay2/d/binding.d
+++ b/M2/Macaulay2/d/binding.d
@@ -226,6 +226,7 @@ bumpPrecedence();
      export LeftArrowW := binaryright("<-"); export LeftArrowS := makeKeyword(LeftArrowW);
      export RightArrowW := binaryright("->",arrowop); export RightArrowS := makeKeyword(RightArrowW);
      export DoubleArrowS := makeKeyword(binaryright("=>"));
+bumpPrecedence();
      export GreaterGreaterS := makeKeyword(binaryright(">>"));
 bumpPrecedence();
      whenW = token("when"); makeKeyword(whenW);
```

And now this happens at startup:

```m2
../../Macaulay2/m2/methods.m2:107:46:(1):[21]: error: expected parenthesized argument list or symbol
../../Macaulay2/m2/startup.m2.in:65:23:(0):[21]: error: error occurred in parsing
../../Macaulay2/m2/startup.m2.in:65:23:(0): entering debugger (enter 'help' to see commands)
```

Here's the offending line:

```
functionClosure := methodFunction := opts >> o -> args -> (
```

So it's parsing it like `(opts >> o) -> args` now, but I guess we want it to still be `opts >> (o -> args)`.

mahrud:
So >> needs to be between -> and :=?

d-torrance:
If we bump up the precedence of `->`, then that messes up functions that use any of the assignment operators.  In particular, when I made this change:

```diff
--- a/M2/Macaulay2/d/binding.d
+++ b/M2/Macaulay2/d/binding.d
@@ -224,9 +224,11 @@ bumpPrecedence();
      export ColonEqualW := binaryright(":="); export ColonEqualS := makeKeyword(ColonEqualW);
      export EqualW := binaryright("="); export EqualS := makeKeyword(EqualW);
      export LeftArrowW := binaryright("<-"); export LeftArrowS := makeKeyword(LeftArrowW);
-     export RightArrowW := binaryright("->",arrowop); export RightArrowS := makeKeyword(RightArrowW);
      export DoubleArrowS := makeKeyword(binaryright("=>"));
+bumpPrecedence();
      export GreaterGreaterS := makeKeyword(binaryright(">>"));
+bumpPrecedence();
+     export RightArrowW := binaryright("->",arrowop); export RightArrowS := makeKeyword(RightArrowW);
 bumpPrecedence();
      whenW = token("when"); makeKeyword(whenW);
      ofW = token("of"); makeKeyword(ofW);
```

I got:

```m2
../../Macaulay2/m2/startup.m2.in:163:85:(0): error: left hand side of assignment inappropriate
-*dummy position*- error: error occurred in parsing
```

Here's the corresponding code:

```m2
     setAttribute = (val,attr,x) -> ( Attributes#val ??= new MutableHashTable )#attr = x;
```

So I think this is parsing `x -> y = z` as `(x -> y) = z` instead of `x -> (y = z)` as desired.

So I think we probably need all the various assignment operators, including `>>`, to stick with the same precedence.

mahrud:
I don't follow. Currently := and >> and -> all have precedence 72, as well as many others, and all we want is for := to be parsed after all of the others, no?

mahrud:
I disagree, I think we should put parenthesis around the line in system.m2!

d-torrance:
It would be a *major* breaking change.   When I put parentheses around that statement, then another pops up that needs parentheses, and then another one.  It's a big old game of whack-a-mole.  I've already added six pairs of parentheses and I'm still in Core.

mahrud:
I think it's worth discussing, because I don't think it would be a frequent issue outside of Core. I'm curious how many fixes in packages are required.

Another idea: could we move both := and => early?

d-torrance:
I tried dropping => down to the same level as :=, but it's a similar game of whack-a-mole.  There's just so much code that assumes they all have the same precedence.  After these experiments, I really don't think that we should mess with this.

Another possible solution for the installing >> methods issue is #1774.  Currently, `(X >> Y) := f` doesn't work because it thinks it's parallel assignment, and we only support parallel assignment when it's a sequence of symbols.  But if parallel assignment was generalized so that `(x_1,...,x_n) op (y_1,...,y_n)` just ran `x_i op y_i` for each `i`, then it should work.

mahrud (changes requested):
I'd rather keep this open as an issue until we have a better solution.

d-torrance:
Closing -- I'll have a PR with parallel assignment updates, including support for `(X >> Y) := f`, soon.

PR #3736, opened 2025-04-17
Added the Subring package

fragandi:
Hi everyone, here is a small package with methods for handling subrings based on the ones available in SAGBI and in InvariantRing


M2/Macaulay2/packages/Subring/Subring.m2
```diff
@@ -0,0 +1,174 @@
...
+    Version => "1.0",
+    Date => "June 6, 2023",
+    Authors => {
+	    {Name => "Francesca Gandini", Email => "fra.gandi.phd@gmail.com"},
+        {Name => "Casey Hill", Email => "casey.hill@uky.edu"},
+        {Name => "Trevor K. Karn", Email => "karnx018@umn.edu"},
+        {Name => "Miranda Moore", Email => "moor2340@umn.edu"},
+        {Name => "Christopher O'Neill", Email => "cdoneill@sdsu.edu"}},
+    Headline => "a package for subrings",
+    Keywords => {"Documentation"},
```
d-torrance, line 13: You probably don't want "Documentation" as a keyword -- that's more for packages like `Macaulay2Doc` and `SimpleDoc` that actually deal with documentation.

Try picking one (or possibly more) of the mathematical keyword(s) from the headings on the [packages provided with Macaulay2](https://www.macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_packages_spprovided_spwith_sp__Macaulay2.html) documentation page.

M2/Macaulay2/packages/Subring/Subring.m2
```diff
@@ -0,0 +1,174 @@
...
+    DebuggingMode => true
```
d-torrance, line 14: `DebuggingMode` should either be omitted or set to `false`.

d-torrance (changes requested):
Thanks for the contribution!

In addition to the other comments, there's also a typo:

```
M2/Macaulay2/packages/Subring/SubringDoc.m2:282: accessable ==> accessible
```

d-torrance:
Also, "Subring" should be added to the [=distributed-packages](https://github.com/Macaulay2/M2/blob/development/M2/Macaulay2/packages/%3Ddistributed-packages) file.

M2/Macaulay2/packages/Subring/Subring.m2
mahrud, line 1: This file is in the wrong place, no?

I would like to allow putting package files inside the package directory, but I think several things need to be changed for this first.

fragandi:
> Also, "Subring" should be added to the [=distributed-packages](https://github.com/Macaulay2/M2/blob/development/M2/Macaulay2/packages/%3Ddistributed-packages) file.

I could not find this file, can you give me more details on the path?

d-torrance:
It's at `M2/Macaulay2/packages/=distributed-packages`

fragandi:
> > Also, "Subring" should be added to the [=distributed-packages](https://github.com/Macaulay2/M2/blob/development/M2/Macaulay2/packages/%3Ddistributed-packages) file.
> 
> I could not find this file, can you give me more details on the path?

NWM I thought a file could not start with = but I guess that's ok!

M2/Macaulay2/packages/Subring/Subring.m2
fragandi: I did not know this was an issue. How should I change the paths? Or should I just merge all files in one and get rid of the directory?

d-torrance: `git mv M2/Macaulay2/packages/Subring/Subring.m2 M2/Macaulay2/packages/Subring.m2` should do the trick.  You'd also need to update the `load` to point to the new relative location of the other two files, e.g.,:

```m2
load "./Subring/SubringDoc.m2"
```

fragandi: > `git mv M2/Macaulay2/packages/Subring/Subring.m2 M2/Macaulay2/packages/Subring.m2` should do the trick. You'd also need to update the `load` calls to point to the new relative location of the other two files, e.g.,:
> 
> ```m2
> load "./Subring/SubringDoc.m2"
> ```

Can you explain what you want me do in UI terms? I do not have a terminal setup on this computer but can do "copy editor edits" with github.dev

d-torrance: I've never done this, but apparently you can move the file by just changing its filename?

https://docs.github.com/en/repositories/working-with-files/managing-files/moving-a-file-to-a-new-location

fragandi: Yep, so you just want me to 

1. move Subring.m2 outside of the Subring directory and just as a file in the packages directory
2. change the path that loads tests and documentation to reflect this directory chage

Correct?

d-torrance: That's right!

fragandi: I've done this, let me know if you want me to do something else

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
```
d-torrance, line 1: Was this file included by mistake?  Or is it something that would be useful for all VS Code users?

M2/Macaulay2/packages/Subring/SubringDoc.m2
```diff
@@ -0,0 +1,374 @@
...
+   the License, or any later version.
+*-
+
+
+beginDocumentation()
+
+doc ///
+ Node
+  Key
+   Subrings
```
d-torrance, line 17: The tests are failing because there's no `Subrings` symbol.

Both the package itself and the main subring class are called `Subring`, but the package should probably be `Subrings` instead as suggested by the documentation.

Renaming it would require:
* Renaming the main package file to `Subrings.m2`
* Updating the entry in the `=distributed-packages` file
* Updated the first argument in the call to `newPackage`

mahrud:
@fragandi is the plan to eventually make SubalgebraBases depend on this package? Because currently the `subring` conflict makes them not usable at the same time.

fragandi:
I have talked to Ollie @ollieclarke8787 before this PR and the idea is indeed to have Subrings as part of the M2 packages and then have SubalgebrasBases call on it. Maybe Ollie you can look into this conflict? Or let me know if I messed up something in the PR.

mahrud:
The current failure was because you forgot `AuxiliaryFiles => true` in the package header. The packages are fine alone, just not usable at the same time right now.

d-torrance:
@fragandi - Just a friendly reminder about this PR

fragandi:
@ollieclarke8787 and I will follow up on this around the time of the Warwick workshop (6/29/26). Converted to draft until then.

fragandi:
@galettof and @trevorkarn might also be interested in this PR. We have a check-in meeting scheduled for June 22, let me know if you would like a link!

ollieclarke8787:
Some updates on the above changes:
- The type `Subring` in the `Subrings` package is now completely interchangeable with the type `Subring` in `SubalgebraBases` so it should be a matter of importing the Subrings in SubalgebraBases and removing parts of that package.
- I currently get an error when installing the package, which suggests there may be a problem with the documentation (since `needsPackage "Subrings"` seems to be fine) - but I'm having trouble tracking this down.

M2/Macaulay2/packages/Subrings.m2
```diff
@@ -0,0 +1,204 @@
...
+
+export {"subring",
+    "subringGenerators",
+    "presentationRing",
+    "presentationMap",
+    "presentationIdeal",
+    "toQuotientRing",
+    "isSubringElement",
+    "GeneratorSymbol",
+    "flattenedRing"
```
d-torrance:
```suggestion
    "flattenedRing",
    "Subring",
```

This type needs to be exported as well.  That's why we were getting that strange error during `installPackage`.

d-torrance (changes requested):

M2/Macaulay2/packages/Subrings.m2
```diff
@@ -0,0 +1,205 @@
...
+    genMatrix.cache#Subring = S
+    )
+
+-- a method to create subrings from a `List` of generators
+subring List := opts -> genList -> (
+    subring(matrix {genList}, opts)
+    )
+
+flattenedRing = method()
+flattenedRing Subring := S -> S#"flattenedRing"
```
mahrud, line 72: Why not just install a method `flattenRing Subring`?

ollieclarke8787:
`isSubringElement` now uses hooks, so it should be possible modify `SubalgebraBases` to build on top of it. TODO: add a new hook to a function that checks if a sagbi basis (of high enough degree) is already computed and return whether the element subducts to zero against the sagbi basis.

M2/Macaulay2/packages/Subrings.m2
ollieclarke8787: `flattenedRing` was defined in `SubalgebraBases` so I pulled it across to `Subrings`, it should be possible to swap over to using `flattenRing` but the underlying meaning seems a little different: the `flattenedRing S` of `S:Subring` is equivalent to `flattenRing ambient S`. For now I've added `flattenRing` in b4777fe

ollieclarke8787:
`SubalgebraBases` seems to be functioning with the Subrings package but the docs have a lot of ambiguous reference warnings.

fragandi:
Spoke with Ollie @ollieclarke8787 today, his work addressed the comments above, let us know if there any other further suggested edits.

M2/Macaulay2/packages/Subrings.m2
```diff
@@ -0,0 +1,219 @@
...
+      {Name => "Casey Hill", Email => "casey.hill@uky.edu"},
+      {Name => "Trevor K. Karn", Email => "karnx018@umn.edu"},
+      {Name => "Miranda Moore", Email => "moor2340@umn.edu"},
+      {Name => "Christopher O'Neill", Email => "cdoneill@sdsu.edu"},
+      {Name => "Oliver Clarke", Email => "oliver.clarke@durham.ac.uk"}
+      },
+    Headline => "a package for subrings",
+    Keywords => {"Commutative Algebra"},
+    AuxiliaryFiles => true,
+    DebuggingMode => true
```
d-torrance:
```suggestion
```

`DebuggingMode` should be false (which is the default, so omitting it works too).

d-torrance (changes requested):

M2/Macaulay2/packages/Subrings.m2
```diff
@@ -0,0 +1,218 @@
...
+    Date => "July 3, 2026",
+    Authors => {
+      {Name => "Francesca Gandini", Email => "fra.gandi.phd@gmail.com"},
+      {Name => "Casey Hill", Email => "casey.hill@uky.edu"},
+      {Name => "Trevor K. Karn", Email => "karnx018@umn.edu"},
+      {Name => "Miranda Moore", Email => "moor2340@umn.edu"},
+      {Name => "Christopher O'Neill", Email => "cdoneill@sdsu.edu"},
+      {Name => "Oliver Clarke", Email => "oliver.clarke@durham.ac.uk"}
+      },
+    Headline => "a package for subrings",
```
d-torrance: The builds are failing now with the following check:

```m2
 --      --	    "^[A-Z]", -- headlines should not be capitalized, but hard to check
 --      	    "\\.$",   -- headlines should not end with a period
 --      	    "\\n",    -- headlines should not contain line breaks
 --      	    -- redundant clauses for package headlines:
 --      	    "(T|t)his (P|p)ackage",
 --      	    "(P|p)ackage (for|on|to)"}, opts.Headline));
 -- 
 -- i10 : if #pkgs > 0 then error("packages whose headlines doesn't follow guidelines:", newline,
 --           toString netList(toList \ pairs applyValues(pkgs, opts -> opts.Headline)))
 -- stdio:24:23:(3):[1]: error: packages whose headlines doesn't follow guidelines:
 -- +--------+----------------------+
 -- |Subrings|a package for subrings|
 -- +--------+----------------------+
```

So at minimum, get rid of the "a package for" here.  "subrings" by itself might be a little too terse though lol

d-torrance (changes requested):

M2/Macaulay2/packages/Subrings.m2
ollieclarke8787: Changed to `subrings of polynomial rings and quotients` [6b67548](https://github.com/Macaulay2/M2/pull/3736/commits/6b675485a05035508053881bc3e349e3af46ef19)

PR #3737, opened 2025-04-18
Document unicode math operators

d-torrance:
We document the three unicode math operators that were added back in #3584 (·, ⊠, and ⧢).

We also:

* Add support for using the augmented versions of these operators. 
* Add a `Vector · Vector` dot product method.


pzinn:
I'm still opposed to this change, cf my rejected PR #3599

mahrud:
> I'm still opposed to this change, cf my rejected PR #3599

It wasn't rejected. We requested changes to balance introducing a new, useful feature while staying clear of undefined behavior when packages define conflicting keywords, at least until the details of that are ironed out. I think with relatively minor changes, it can still make it to this upcoming release.

M2/Macaulay2/tests/normal/augmented-assignment.m2
```diff
@@ -74,6 +74,15 @@ assert Equation(x#0, 28)
 Foo @ Foo := (x, y) -> Foo(4*x#0 - 5*y#0)
 x @= Foo 7
 assert Equation(x#0, 77)
+Foo · Foo := (x,y) -> Foo(6*x#0 * 7*y#0)
+x ·= Foo 8
```
mahrud, line 78: This looks awkward, and I'm not convinced that it is useful enough to justify supporting it.

M2/Macaulay2/packages/Macaulay2Doc/operators/tensor.m2
```diff
@@ -34,3 +34,33 @@ Node
...
+    Text
+      To insert this character in Emacs, you may press @KBD "C-x 8 RET"@
+      or @KBD "M-x insert-char"@ and then enter "squared times" in the
+      minibuffer.
+
+      Alternatively, you may press @KBD "C-x RET C-\\"@ or
+      @KBD "M-x set-input-method"@ and then enter "TeX" in the minibuffer.
+      Afterwards, typing "\boxtimes" will input the ⊠ character.  You may then
+      toggle the input method using @KBD "C-\\"@  or
+      @KBD "M-x toggle-input-method"@.
```
mahrud, line 53: Since these paragraphs are mostly repeated, it may be worthwhile to have an automated section printed at the bottom of unicode operators that explains how to enter them. e.g. something like the "For the programmer" section underneath operators, which explains how to define methods on them.

Also, I still think every unicode operator should be tied to an ascii method name so it can be still used without unicode entry (e.g. `**` and `tensor` for $\otimes$).

M2/Macaulay2/tests/normal/augmented-assignment.m2
d-torrance: My original thought was that every flexible binary operator (as long as it didn't introduce ambiguity when appending an extra `=`, like `==` -> `===`) should support augmented assignment out of the box.

While it's certainly true it would probably be a bad idea to use `·=` for an augmented dot product (First `v` is a vector, but after `v ·= w`, it's an element of the ring...), we might as well make it available for some other context where it might make sense.

M2/Macaulay2/packages/Macaulay2Doc/operators/tensor.m2
d-torrance: I like the idea of an automated section -- I'll push a proposal here in a bit.

As for the ascii method name... I'm not sure if we should force a particular name for each operator, or just leave it up to package authors.  Maybe one package  has ⊠ to be a shortcut for `externalTensorProduct` and another for `strongProduct` , etc.

But yeah, we should probably provide an ascii version for each method that's actually defined.  I'll go ahead and add `dotProduct(Vector, Vector)` as a synonym for `Vector · Vector`.

d-torrance:
I ended up dropping the `Vector · Vector` method from this PR, although I'll likely revisit it later.  Various complications:

* What to do with rings over CC?  We should probably use the Hermitian transpose, and perhaps that should be implemented in the engine rather than conjugating every entry one at a time?
* If we add a `dotProduct` method function to Core to serve as an ascii synonym, then we need to move the definition away from BoijSoiderberg, which currently exports it.  But that that opens up a can of worms with the [documentation](https://www.macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/BoijSoederberg/html/_dot__Product.html), as its Inputs section defines five different inputs despite the fact that the method takes at most three inputs.

mahrud:
> * But that that opens up a can of worms with the [documentation](https://www.macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/BoijSoederberg/html/_dot__Product.html), as its Inputs section defines five different inputs despite the fact that the method takes at most three inputs.

For the record, the solution is to either use `Synopsis` or group the types together, e.g. `M => {matrix,BettiTally}` or similar.

d-torrance:
Discussed with Mike.  We may change this completely once #3599 is merged

Issue #3145, opened 2024-02-27
Case (in)sensitive files in macOS


mahrud:
M2 shouldn't confused a file in the same directory named `elimination.m2` with the package `Elimination.m2`, even if it's running on macOS. This causes too many head scratching confusions for people.

mahrud:
Here is the top link on google: https://stackoverflow.com/a/69828765/14107885

mikestillman:
This would be nice to implement.  I thought we could not find easily this information.  We are now using c++17, so we should be able to use this.

mahrud:
Even without this solution, somehow macOS can show you your files with the correct capitalization when you open a file browser or type `ls`, right? Then you can also get the filename with correct capitalization as a string and compare the strings in M2 before you decide to load the file or look in the next directory on path.

mikestillman:
One has to be a bit careful:
```
bash-3.2$ ls
lecture.m2
bash-3.2$ ls LecTure.m2
LecTure.m2
```

But it is true that emacs on MacOS is mostly case sensitive (i.e. if you do ^X^F and give a filename to open, if it is the wrong capitalization, it fixes it for you.

Devlin-Mallory:
This does appear to still be an issue...

Issue #3225, opened 2024-05-11
needsPackage should reload the package timestamp is new


mahrud:
`needs` automatically reloads the file if the modification time is more recent than the last time it was read. `needsPackage` should do the same with the timestamp of the main package file and the timestamp of the auxiliary directory, if it exists.

d-torrance:
Fixed by #3741

mahrud:
I have to say, this has been a little bit of a pain for the following reason: when the package is reloaded, all of its types and methods get redefined, therefore previous instances of those types are no longer usable. This also causes a wall of "shadow" warnings which is really not helpful. Perhaps a small warning would go a long way.

Issue #2413, opened 2022-03-12
`needs` confuses files of the same name


mahrud:
Try running:
```m2
needs "code.m2"
```
And nothing happens. Now try running:
```m2
load "code.m2"
```
and you get an error that `code.m2` was not found on path. This is because `code.m2` is a Core script, so now all future invocations of `needs` for it are ignored.

What's worse, if you're developing some code in `code.m2` in your local directory then `needs` doesn't load it at all.

DanGrayson:
This doesn't seem to be a bug -- after all, files can have the same name and be in different directories.

mahrud:
After M2 has loaded, the Core directory is _not_ on the path. `needs` should not be confused about that.

mahrud:
It would be great to fix this, perhaps by storing the full path of the file when `needs` is used, rather than just the filename.

mahrud:
Related: #3145

d-torrance:
Fixed by #3741

PR #3741, opened 2025-04-20
needs and needsPackage updates

d-torrance:
Several related changes:

* `minimizeFilename` now expands `~` before minimizing
* We update `needs` so that it checks for files in the same way as `load` and `input`, i.e., it searches through the various directories in `path` and loads the first hit, but only if it's been modified since the last load or has never been loaded.  So for example, if we call `needs "foo.m2"` and then later create another `foo.m2` somewhere earlier in `path`, the newer `foo.m2` is loaded.  Previously, we would have just checked to see if the original `foo.m2` needed reloading.  I believe this fixes [#2413](https://github.com/Macaulay2/M2/issues/2413).
* `needsPackage` now checks whether any of the files in a previously-loaded package have been updated, and if so, reloads the package, as requested in [#3225](https://github.com/Macaulay2/M2/issues/3225).

### DIFF
--- a/M2/Macaulay2/m2/packages.m2
+++ b/M2/Macaulay2/m2/packages.m2
@@ -191,7 +191,10 @@ needsPackage String  := opts -> pkgname -> (
     and (opts.FileName === null or
 	realpath opts.FileName == realpath pkg#"source file")
     and pkg.PackageIsLoaded
-    then use value PackageDictionary#pkgname
+    then (
+	if any(packageFiles pkg, file -> fileTime file > filesLoaded#file)
+	then loadPackage(pkgname, opts ++ {Reload => true})
+	else use pkg)
     else loadPackage(pkgname, opts))
 
 -- used as the default loadOptions in newPackage
@@ -581,11 +584,15 @@ debug GlobalDictionary := dict -> (
     if not isMember(dict, dictionaryPath) then dictionaryPath = prepend(dict, dictionaryPath);
     checkShadow())
 
-locate Package := pkg -> NumberedVerticalList nonnull (
-    pkgaux := if not pkg#?"auxiliary files" then {}
-    else select(values loadedFiles, match_(pkg#"auxiliary files"));
+packageFiles = pkg -> (
+    pkgaux := (
+	if not pkg#?"auxiliary files" then {}
+	else select(values loadedFiles, match_(pkg#"auxiliary files")));
+    prepend(realpath pkg#"source file", pkgaux))
+
+locate Package := pkg -> NumberedVerticalList (
     -- TODO: somehow keep track of the number of lines of each file
-    apply(prepend(pkg#"source file", pkgaux), file -> new FilePosition from (file, 0, 0)))
+    apply(packageFiles pkg, file -> new FilePosition from (file, 0, 0)))
 
 -----------------------------------------------------------------------------
 -- evaluateWithPackage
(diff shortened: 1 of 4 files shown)


M2/Macaulay2/m2/packages.m2
```diff
@@ -191,7 +191,10 @@ needsPackage String  := opts -> pkgname -> (
     and (opts.FileName === null or
 	realpath opts.FileName == realpath pkg#"source file")
     and pkg.PackageIsLoaded
-    then use value PackageDictionary#pkgname
+    then (
+	if any(packageFiles pkg, file ->fileTime file > filesLoaded#file)
+	then loadPackage(pkgname, opts ++ {Reload => true})
+	else use pkg)
```
mahrud, line 197: What's the difference between `use pkg` and `use value PackageDictionary#pkgname`?

mahrud (approved):

mahrud (commented):
Actually, one test is failing, except that seems good.

mahrud:
Also, could you test this PR on macOS and see if it resolves #3145 or not?

M2/Macaulay2/m2/packages.m2
d-torrance: Nothing -- we'd already defined `pkg` a few lines earlier, so I figured it would be good to use it rather than calling `value` again.

mahrud:
Something else that would be desirable: if I type `viewHelp PackageName` even without the quotes, I get the right thing, this is because it's easy to check that we're given a symbol not assigned to a package, turn it into a string, and load that package. I suspect doing the same with `needsPackage` is going to relieve a lot of pain!

ps: I do get an unfortunate warning that would be nice to silence by deleting the symbol from User's dictionary first:
```m2
i1 : viewHelp NormalToricVarieties
 -- warning: symbol "NormalToricVarieties" in PackageDictionary is shadowed by a symbol in User#"private dictionary"
 --   use the synonym Package$NormalToricVarieties
Opening in existing browser session.
```

mahrud:
Would be great if this PR made it in this release. What's the status, can I help in some way?

d-torrance:
Some kind of key error on the macOS build.  I'm hoping that the latest commit will shed more light on what's going on!

d-torrance:
Yay it's fixed!  I think the problem was that macOS's temporary directory is under `/var/folders`, which is really a symlink to `/private/var/folders`.  We were using `realpath` when storing filenames in the hash table of loaded files, but weren't when we were looking them up.

@mahrud -- You already approved, but just wanted to double check -- do you think this is okay to merge?

mahrud:
I think so, but I'd rather someone using macOS also tested this just to be sure. @Devlin-Mallory would you be willing to do this?

Devlin-Mallory:
> I think so, but I'd rather someone using macOS also tested this just to be sure. @Devlin-Mallory would you be willing to do this?

Yes, I can take a look tomorrow! Just try loading some packages like normal, or anything in particular to check?

mahrud:
> Yes, I can take a look tomorrow! Just try loading some packages like normal, or anything in particular to check?

Remember how `elimination.m2` was confusing your M2? First and foremost we want to avoid that kind of issue. Second, if a package has changed, we want `needsPackage` to actually reload it, so try editing something in a package and call `needsPackage` again without restarting and see if it works (note that Type will be overwritten, but that's hard to avoid).

Devlin-Mallory:
> > Yes, I can take a look tomorrow! Just try loading some packages like normal, or anything in particular to check?
> 
> Remember how `elimination.m2` was confusing your M2? First and foremost we want to avoid that kind of issue. Second, if a package has changed, we want `needsPackage` to actually reload it, so try editing something in a package and call `needsPackage` again without restarting and see if it works (note that Type will be overwritten, but that's hard to avoid).

This seems to work great for me on macOS! Really like the new behavior of needsPackage.

d-torrance:
Great!  Thanks for trying it out

PR #3742, opened 2025-04-20
Turn "intersection" into a synonym for "intersect"

d-torrance:
We've essentially already done this in Core, as `intersection(Set, Set)` and `intersection(RRi, RRi)` were identical to the corresponding `intersect` methods.

I tested a few packages (like Polyhedra) that define `intersection` methods locally, and everything seems to still work...

Closes: #2257


mahrud:
Can you think of a way to also add `intersectin(List, List)`? The fact that `intersect` is a binary method makes this complicated.

mahrud (approved):

PR #3744, opened 2025-04-21
Assorted fixes

mahrud:
- **fixed links from RationalMaps to Cremona**
- **fixed DegreeLimit => 0 or {} in Hom**
- **fixed a problem with nauty installation**
- **fixed a problem with msolve installation**
- **added msolve as a build dependency**
- **set CMake to be quiet about e-antic**
- **updated minimum gtest version**
- **updated mpfr download url in cmake**
- **changed runHooks debug info to depend on debugHooksLevel**
- **avoid duplicating strategies in HookPriority list**
- **passed // and \\\\ options to Hom**
- **added test for order of overwriting options**
- **added a cache table to GaloisField**
- **added (options, List), e.g. for 'options methods codim'**
- **switched to ??= for caching Module ^ Array**

### DIFF
--- a/M2/Macaulay2/m2/galois.m2
+++ b/M2/Macaulay2/m2/galois.m2
@@ -225,6 +225,7 @@ GF(Ring) := GaloisField => opts -> (S) -> (
 	  );
      F.use = F -> var <- F_0;
      F.use F;
+     F.cache = new CacheTable;
      F / F := (x,y) -> if y == 0 then error "division by zero" else x // y;
      F % F := (x,y) -> if y == 0 then x else 0_F;
      F)
--- a/M2/Macaulay2/m2/matrix2.m2
+++ b/M2/Macaulay2/m2/matrix2.m2
@@ -296,6 +296,7 @@ modulo(Matrix,Matrix)  := Matrix => options -> (m,n) -> (
      then map(L,source f,f)			    -- it can happen that L has a Schreier order, and we want to preserve that exactly
      else f)
 
+-- TODO: make this work between non-free modules
 quotientRemainder'(Matrix, Matrix) := Matrix => (f, g) -> (
     L := source f;
     M := target f;
@@ -351,8 +352,11 @@ quotient(Matrix, Matrix) := Matrix => opts -> (f, g) -> (
 addHook((quotient, Matrix, Matrix), Strategy => Default,
     -- Note: this strategy only works if the remainder is zero, i.e.:
     -- homomorphism' f % image Hom(source f, g) == 0
-    -- TODO: should we pass MinimalGenerators => false to Hom and homomorphism'?
-    (opts, f, g) -> map(source g, source f, homomorphism(homomorphism' f // Hom(source f, g))))
+    (opts, f, g) -> (
+	opts = new OptionTable from {
+	    DegreeLimit       => opts.DegreeLimit,
+	    MinimalGenerators => opts.MinimalGenerators };
+	map(source g, source f, homomorphism(homomorphism'(f, opts) // Hom(source f, g, opts)))))
 
 -- FIXME: this is still causing unreasonable slow downs, e.g. for (large m) // (scalar)
 addHook((quotient, Matrix, Matrix), Strategy => "Reflexive", (opts, f, g) -> if f == 0 or isFreeModule source f then (
@@ -396,8 +400,11 @@ quotient'(Matrix, Matrix) := Matrix => opts -> (f, g) -> (
 addHook((quotient', Matrix, Matrix), Strategy => Default,
     -- Note: this strategy only works if the remainder is zero, i.e.:
     -- homomorphism' f % image Hom(g, target f) == 0
-    -- TODO: should we pass MinimalGenerators => false to Hom and homomorphism'?
-    (opts, f, g) -> map(target f, target g, homomorphism(homomorphism' f // Hom(g, target f))))
+    (opts, f, g) -> (
+	opts = new OptionTable from {
+	    DegreeLimit       => opts.DegreeLimit,
+	    MinimalGenerators => opts.MinimalGenerators };
+	map(target f, target g, homomorphism(homomorphism'(f, opts) // Hom(g, target f, opts)))))
 
 addHook((quotient', Matrix, Matrix), Strategy => "Reflexive",
     (opts, f, g) -> if all({source f, target f, source g, target g}, isFreeModule) then dual quotient(dual f, dual g, opts))
--- a/M2/Macaulay2/m2/methods.m2
+++ b/M2/Macaulay2/m2/methods.m2
@@ -638,9 +638,11 @@ pushInfoLevel :=  n -> (
     infoLevel = infoLevel + n; n)
 popInfoLevel  := (n, s) -> (infoLevel = infoLevel - n; s)
 
+debugHooksLevel = debugLevel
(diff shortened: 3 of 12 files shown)


M2/cmake/check-libraries.cmake
```diff
@@ -142,7 +142,7 @@ find_package(EAntic	2.0.0 QUIET) # only needed when libnormaliz.so needs it
 find_package(MSolve	0.7.0)
 find_package(Frobby	0.9.0)
 find_package(CDDLIB)  # 0.94m?
-find_package(GTest	1.10)
+find_package(GTest	1.16)
```
d-torrance, line 145: Why such a high minimum version?  1.16 was only released this February.

M2/cmake/build-libraries.cmake
```diff
@@ -211,8 +211,8 @@ _ADD_COMPONENT_DEPENDENCY(libraries bdwgc "" BDWGC_FOUND)
 # NOTE: mpfr puts pointers to gmp numbers in thread local variables, unless
 # specially configured, so we shouldn't tell gmp to use libgc (we used to do that)
 ExternalProject_Add(build-mpfr
-  URL               https://www.mpfr.org/mpfr-current/mpfr-4.1.0.tar.xz
-  URL_HASH          SHA256=0c98a3f1732ff6ca4ea690552079da9c597872d30e96ec28414ee23c95558a7f
+  URL               https://www.mpfr.org/mpfr-current/mpfr-4.2.1.tar.xz
+  URL_HASH          SHA256=277807353a6726978996945af13e52829e3abd7a9a5b7fb2793894e18f1fcbb2
```
d-torrance, line 215: MPFR 4.2.2 was released last month.  Should we use it?  https://www.mpfr.org/mpfr-current/

M2/Macaulay2/m2/methods.m2
```diff
@@ -656,7 +658,7 @@ runHooks(Symbol,                  Thing) := true >> opts -> (key,        args) -
 runHooks(Sequence,                Thing) := true >> opts -> (key,        args) -> runHooks(getHookStore(key, false), key, args, opts)
 runHooks(MutableHashTable, Thing, Thing) := true >> opts -> (store, key, args) -> (
     store = if store#?key then store#key else (
-	if debugLevel > 1 then printerr("runHooks: no hooks installed for ", toString key); return );
+	if debugHooksLevel > 1 then printerr("runHooks: no hooks installed for ", toString key); return );
```
d-torrance, line 663: Should `debugHooksLevel` be documented anywhere?  Should it be exported?

Right now we *kind of* document `debugLevel` for hooks.  It's not mentioned explicitly, but it's set in one of the examples at [using hooks](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_using_sphooks.html).

M2/Macaulay2/m2/methods.m2
```diff
@@ -628,6 +628,7 @@ addHook(MutableHashTable, Thing, Function) := opts -> (store, key, hook) -> (
     store = store#key;
     ind := #store.HookPriority; -- index to add the hook in the list; TODO: use Priority to insert in the middle?
     alg := if opts.Strategy =!= null then opts.Strategy else ind;
+    if not store.HookAlgorithms#?alg then
```
d-torrance, line 632: What if someone wants to overwrite a strategy?

M2/cmake/check-libraries.cmake
mahrud: Because it's the only one compatible with the most recent version of cmake, and we always download gtest anyway, so might as well be the latest one.

M2/Macaulay2/m2/matrix2.m2
```diff
@@ -351,8 +352,11 @@ quotient(Matrix, Matrix) := Matrix => opts -> (f, g) -> (
 addHook((quotient, Matrix, Matrix), Strategy => Default,
     -- Note: this strategy only works if the remainder is zero, i.e.:
     -- homomorphism' f % image Hom(source f, g) == 0
-    -- TODO: should we pass MinimalGenerators => false to Hom and homomorphism'?
-    (opts, f, g) -> map(source g, source f, homomorphism(homomorphism' f // Hom(source f, g))))
+    (opts, f, g) -> (
+	opts = new OptionTable from {
+	    DegreeLimit       => opts.DegreeLimit,
+	    MinimalGenerators => opts.MinimalGenerators };
+	map(source g, source f, homomorphism(homomorphism'(f, opts) // Hom(source f, g, opts)))))
```
d-torrance, line 359: Should we add any tests for these changes?

M2/Macaulay2/m2/galois.m2
```diff
@@ -225,6 +225,7 @@ GF(Ring) := GaloisField => opts -> (S) -> (
 	  );
      F.use = F -> var <- F_0;
      F.use F;
+     F.cache = new CacheTable;
```
d-torrance, line 228: Is this just for future use?  Or are there some methods defined for ancestors that might look for it?

M2/Macaulay2/m2/methods.m2
mahrud: This indeed allows them to overwrite a strategy. This will prevent duplicating it, so the same strategy isn't attempted twice. To change the order they have to manually edit the priority list.

M2/Macaulay2/m2/matrix2.m2
mahrud: Some tests are included in a different package which may or may not make it to this release. I think existing tests are sufficient for now.

M2/cmake/build-libraries.cmake
mahrud: Maybe for the next release.

M2/Macaulay2/m2/methods.m2
mahrud: I don't think it's worth exporting. Neither is `HookAlgorithms`, `HookPriority`, etc.

M2/Macaulay2/m2/methods.m2
d-torrance: Ah, ok.  I imagined a set of parentheses that wasn't there.  :)

M2/Macaulay2/m2/galois.m2
mahrud: It's for my personal uses :)

M2/cmake/check-libraries.cmake
mahrud: `find_package` also looks in `usr-host` where it's installed in.

d-torrance: Ah, ok.  That makes sense.

mahrud:
I added an example of accessing the hook data by importing the Core symbols. I still don't think they should be exported. They are internal details, but worth having an example.

mikestillman (approved):
While I have not downloaded and tried this out, these look like reasonable changes.  I had placed myself as reviewer, not realizing that @d-torrance was already in the process of reviewing it.

Thanks for cleaning this code up!

mikestillman:
Oops, I approved it, but it looks like there is a build failure...

d-torrance:
The build failure was just during the installation of the Ubuntu prereqs. Probably just some server thing.

PR #3746, opened 2025-04-22
Prepend -lgtest to LIBS

d-torrance:
This is a quick followup to #3723 -- I broke the autotools build in Ubuntu 20.04.  (It's broken even worse in Ubuntu 18.04, but that's a different problem...)

Since gtest is statically linked (at least on Ubuntu), it needs to appear prior to any prerequisite libraries.  And on older systems (e.g., Ubuntu Focal), it requires `-lpthread`.  This already appears in `LIBS` (from our check for `pthread_mutex_trylock` earlier).

Cc: @mikestillman


mikestillman (approved):
I approve these changes.

It would be nice to be able to build M2 with as many things as possible being made static.  I will ask you about that separately...

PR #3748, opened 2025-04-23
Improvements to parallel assignment

d-torrance:
We add various parallel assignment features, e.g:

### Parallel assignment for mutable hash tables and lists
```m2
i1 : x = new MutableHashTable;

i2 : y = new MutableList;

i3 : (x.foo, y#0) = (1, 2)

o3 = (1, 2)

o3 : Sequence

i4 : peek x

o4 = MutableHashTable{foo => 1}

i5 : peek y

o5 = MutableList{2}
```

### Partial parallel assignment
```m2
i6 : (x,,y) = (1,2,3)

o6 = (1, , 3)

o6 : Sequence

i7 : x

o7 = 1

i8 : y

o8 = 3
```

### Parallel method installation

```m2
i9 : X = new Type of HashTable;

i10 : foo = method();

i11 : (foo X, X + X) := (x -> 1, (x,y) -> 2)

o11 = (FunctionClosure[stdio:11:19-11:25], FunctionClosure[stdio:11:27-11:37])

o11 : Sequence

i12 : (x, y) = (new X, new X)

o12 = (X{}, X{})

o12 : Sequence

i13 : foo x

o13 = 1

i14 : x + y

o14 = 2
```

### Parallel augmented assignment

```m2
i15 : (x,y,z) = (1,2,3)

o15 = (1, 2, 3)

o15 : Sequence

i16 : (x,y,z) += (4,5,6)

o16 = (5, 7, 9)

o16 : Sequence

i17 : x

o17 = 5

i18 : y

o18 = 7

i19 : z

o19 = 9

i20 : (x,y,z) *= 2 -- not sure if this is a good idea or not

o20 = (10, 14, 18)

o20 : Sequence

i21 : x

o21 = 10

i22 : y

o22 = 14

i23 : z

o23 = 18
```

### Fixes precedence issues with assignment
See https://github.com/Macaulay2/M2/pull/3733

```m2
i25 : (X >> X) := (x,y) -> 5

o25 = FunctionClosure[stdio:25:12-25:22]

o25 : FunctionClosure

i26 : (x,y) = (new X, new X)

o26 = (X{}, X{})

o26 : Sequence

i27 : x >> y

o27 = 5
```

TODO:
- [ ] Finish tests
- [ ] Add documentation
- [ ] (maybe?) support parallel assignment for `<-`
- [ ] (maybe?) support `(x,y,z) = 1`, which would be equivalent to `(x,y,z) = (1,1,1)`
- [ ] Update `>>` docs to mention that we need parentheses for installing  methods

Closes: #1774


mahrud:
I don't love `(foo X, X + X) := (x -> 1, (x,y) -> 2)` ...

d-torrance:
Yeah I don't expect anyone would use parallel assignment much for method installation.  It just seems like something that ought to work if we want = and := to work in general.  Also, it lets use put parentheses around a single expression for grouping, solving the `>>` method installation issue.

mahrud:
I'm gonna unsubscribe because there are a lot of force pushes but the notifications don't show what has actually changed. @ me when it is ready.

bugs/dan/0-parallel-assignment
```diff
@@ -1,37 +0,0 @@
...
-
-    i24 : (b) = a
-
-    o24 = 1 : (x)
-
-    o24 : Sequence
-
-    i25 : a
-
-    o25 = 1 : (x)			-- wanted to get x here
```
mahrud, line 35: I'm a bit confused about this comment, but I think we would want `(b) = 1:x` to set `b = x`.

d-torrance: Ok, it's restored

M2/Macaulay2/d/evaluate.d
```diff
@@ -1202,7 +1202,12 @@ parallelAssignmentFun(x:parallelAssignmentCode):Expr := (
     value := eval(x.rhs);
     when value is Error do return value else nothing;
     -- (x) = y should behave just like x = y
-    if nlhs == 1 then value = seq(value);
+    if nlhs == 1 then (
+	when value is a:Sequence do (
+	    -- unless y is a sequence of length 1, then it behaves like x = y#0
```
mahrud, line 1207: Just to make sure, `(x) = (1,2,3)` should still be an error, correct?

d-torrance: In the current draft, it's equivalent to `x = (1,2,3)`.  But for consistency, you're right -- it probably should be an error.  I'll push a new version here in a moment.

M2/Macaulay2/d/evaluate.d
```diff
@@ -1194,15 +1194,20 @@ globalAssignmentFun(x:globalAssignmentCode):Expr := (
      globalAssignment(t.frameindex,t,newvalue));
 
 ParallelAssignmentError(n:int):Expr := buildErrorPacket(
-    "parallel assignment: expected a sequence of " + tostring(n) + " values");
+    "parallel assignment: expected a sequence of " + tostring(n) + " value(s)");
```
mahrud, line 1197: Not worth a separate commit, but in the next push this could be changed to `parallel assignment expected a sequence of length N` (note I removed the colon also).

mahrud: Or perhaps `parallel assignment to N symbols encountered a sequence of length M`

mahrud:
One more thing, either for now or a future PR: would be great if this kind of parallel assignment also worked:
```m2
i1 : for (a,b) in {(1,2), (3,4)} list a
stdio:1:4:(3): error: expected symbol
```

d-torrance:
Ooh yeah that would be nice!  I played around with it for a bit and it would require a lot of pretty low-level changes to binding and whatnot.  Let's hold off for a future PR.

M2/Macaulay2/d/common.d
```diff
@@ -45,10 +45,6 @@ export codePosition(c:Code):Position := ( -- TODO retire
     is f:Error                     do f.position
     );
 
-export setuppostfix(e:SymbolClosure,fn:unop):void := (
-     unopNameList = unopNameListCell(fn,e.symbol,unopNameList);
-     e.symbol.postfix = fn;
-     );
```
mahrud, line 51: I understand what this commit does, but was it necessary? I see the simplification, but also seems like not a bad idea to be able to differentiate postfix vs unary operators in the interpreter.

M2/Macaulay2/d/parse.d
```diff
@@ -210,7 +210,6 @@ export augmentedAssignmentCode := {+
     oper:Symbol,
     lhs:Code,
     rhs:Code,
-    info:Symbol, -- variable name or operator
```
mahrud, line 214: Is there any reason to keep this for debugging purposes?

M2/Macaulay2/m2/integers.m2
```diff
@@ -116,6 +116,7 @@ scan({symbol +=, symbol -=, symbol &=, symbol |=, symbol ^^=},
 
 store = method()
 store(AtomicInt, ZZ) := atomicStore
+(<- AtomicInt) := store
```
mahrud, line 119: Why is this method defined this way?

M2/Macaulay2/packages/Macaulay2Doc/operators/assignment.m2
```diff
@@ -485,21 +487,16 @@ document {
...
 	       { "new local variables ", TT "x", ", ", TT "y", ", ", TT "z", ", ... are created" },
 	       { "the expressions c,d,e,... are assigned to the variables x,y,z,..., respectively, as above." },
-	       { "If the left hand side has more elements than the right hand side, then the extra symbols on the left side are given the value ", TO "null", "." },
-	       { "If the left hand side has fewer elements than the right hand side, then the last symbol on the left hand side is given
-     		    as value a sequence containing the trailing elements of the right hand side." 
-		    },
-	       { "If the right hand side is not a sequence, then it is assigned to the first symbol on the left, and the remaining symbols are assigned the
-		    value ", TO "null", "."
-		    }
+	       { "the number of expressions must match the number of variables." }
```
mahrud, line 490: When was this documented behavior removed? It actually seems preferable to me.

mahrud (commented):
The amount of changes is not overwhelming, but I don't feel confident in thoroughly reviewing this before the next release. I think we should merge this after the release, so any potential bugs have some time to reveal themselves.

Also, several aspects of this touch on things that @pzinn has more experience with, so his review may also be valuable.

M2/Macaulay2/packages/Macaulay2Doc/operators/assignment.m2
d-torrance: No idea.  I was surprised to learn about it!

It probably wouldn't be too much trouble to restore it.

d-torrance:
I'm totally fine with waiting for the next release.

M2/Macaulay2/packages/Macaulay2Doc/operators/assignment.m2
```diff
@@ -45,7 +45,10 @@ document {
 	       },
 	  EXAMPLE lines ///
 	       installMethod(symbol <-, String, peek)
-	  ///
+	  ///,
+	  PARA { "The following syntax may also be used.  Note that the parentheses on the left hand side are necessary because of the low precedence of the ",
+	      CODE "<-", " operator."},
+	  EXAMPLE "(<- String) := peek"
```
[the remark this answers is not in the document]
d-torrance: Fair point -- the arrow definitely seems like it's on the wrong side!  I can remove these commits

M2/Macaulay2/d/parse.d
d-torrance: This was a hack that I came up when adding augmented assignment.  There were two unrelated cases where we needed some information about a symbol that wasn't being stored in the Code object:

* The variable name for `globalMemoryReferenceCode` and `threadMemoryReferenceCode`, which currently just store the frame index of the variable.
* The binary/unary operator for `binaryCode` and `unaryCode`, which currently just store the function associated with the operator.

With this PR, the proposal is that these various Code objects now store the symbols themselves.  So the symbol that was in `info` is still there, but now inside `lhs`.

mahrud:
How much harder is it to also allow `(a,b,c) = {1,2,3}` and `(a,b,c) = [x,y,z]` etc?
(I don't think there's much point in `{a,b,c} = {1,2,3}`, other than maybe if you think it's good to return error if the types don't match?)

d-torrance:
Adding support for the various types of visible list on the right hand side would be very simple, I think.  Ensuring that the types match would be a bit trickier since we currently don't keep track of which type of parentheses we're using on the left hand side.

mahrud:
Okay, I'm totally fine with `{a,b,c} = {1,2,3}` not working. After all the output is not a list, but the individual symbols changed. However, we need to be careful of what is the result of `L = (a,b,c) = {1,2,3}`. A list or a sequence?

d-torrance:
I just pushed a proposal to allow other types of list on the right hand side.  Currently, it always returns a sequence, since the easy way to implement this is just to extract the sequence from the list and use it.

```m2
i1 : {x, y, z} = {1, 2, 3}

o1 = (1, 2, 3)

o1 : Sequence
```

mahrud:
Thinking about this more, I think we want `L = (a,b,c) = {1,2,3}` to be a list.

d-torrance:
Done!

```m2
i1 : {x, y, z} = {1, 2, 3}

o1 = {1, 2, 3}

o1 : List
```

mahrud:
@d-torrance once Paul's PR is merged, shall we merge this as well?

d-torrance:
Sounds good to me!

M2/Macaulay2/d/evaluate.d
mikestillman: @d-torrance @mahrud Is this agreed on, should I take a separate look?

mikestillman:
oops, I missed the lat part of the conversation with my last question. I should try playing with #3599, and this one #3748, myself.  Perhaps also with #3737?

d-torrance:
Discussed with Mike

PR #3749, opened 2025-04-23
Add threadLock keyword -- locks mutex, evals code, then unlocks

d-torrance:
We add a `threadLock` keyword as suggested in https://github.com/Macaulay2/M2/pull/3739#issuecomment-2816776740 to  help write thread-safe code.  For example:

```m2
i1 : x = new MutableList;

i2 : y = new MutableList;

i3 : scan(1000, i -> schedule(() -> (x##x = null; threadLock y##y = null)))

i4 : #x

o4 = 180

i5 : #y

o5 = 1000
```

### TODO:
- [ ] Add documentation
- [ ] Deal with any potential issues (e.g., what happens if we cancel a task that locked the mutex before it unlocks it?)

### DIFF
--- a/M2/Macaulay2/d/pthread.d
+++ b/M2/Macaulay2/d/pthread.d
@@ -3,6 +3,7 @@ use evaluate;
 use expr;
 
 header "#include \"../system/supervisorinterface.h\"";
+header "#include <errno.h>"; -- for EBUSY
 
 
 taskCreatePush(f:function(TaskCellBody):null,tb:TaskCellBody) ::=  Ccode(taskPointer,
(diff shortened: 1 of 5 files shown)


M2/Macaulay2/d/pthread.d
```diff
@@ -231,6 +232,20 @@ export getIOThreadMode(e:Expr):Expr := (
     else WrongArg("a file or ()"));
 setupfun("getIOThreadMode", getIOThreadMode);
 
+-- TODO: what if we cancel a task before it unlocks the mutex?
+mutex := newMutex;
+threadLock(c:Code):Expr := (
+    r := 0;
+    while (r = trylock(mutex); r == Ccode(int, "EBUSY")) do (
```
mahrud, line 239: This is not good. I think you want `pthread_cond_wait` or similar.

d-torrance: From what I've read, `pthread_mutex_trylock` is the only way to check if a mutex is locked.  `pthread_cond_wait` blocks until some condition is met, and we don't want to block since we want to check for interrupts.

mahrud: I definitely think we need pthread conditions, because you can signal only one or all threads using pthread_cond_signal and pthread_cond_broadcast that an event has changed. This is how tasks are implemented too. e.g. this is how `taskResult` blocks:
https://github.com/Macaulay2/M2/blob/2565455411d15a3386204aa62a00e20ee5c0e99f/M2/Macaulay2/system/supervisor.cpp#L210-L215
In fact, I suspect if we change `interrupt_handler` in `main.cpp` to also signal the current blocking task, it would interrupt just fine.

d-torrance:
Closing in favor of #4350

PR #3752, opened 2025-04-23
added package ExteriorExtensions

LukeOeding:
Adding package under review at JSAG.
I've corrected some spelling errors.

PR #3753, opened 2025-04-24
modified distributed-packages

LukeOeding:
added ExteriorExtensions to =distributed-packages


d-torrance:
There's no reason to open multiple pull requests -- just keep adding commits to the same branch and the first one will be updated automatically!

Closing this one in favor of #3752 so we can keep track of the comments that are already there.

Issue #3717, opened 2025-03-31
ignore submodule changes


mahrud:
We should add `ignore = all` to `.gitmodule`:
https://git-scm.com/docs/gitmodules#Documentation/gitmodules.txt-submoduleltnamegtignore
This would prevent people from accidentally adding commits to the emacs submodule, unless they do `git add -f .../emacs`.

The only downside is that you might not realize your emacs submodule is outdated, and need to rely on autotools/cmake/`git submodule update` to update them for you.

_Originally posted by @mahrud in https://github.com/Macaulay2/M2/issues/3645#issuecomment-2691268644_

mahrud:
@d-torrance should we add this to this release?

d-torrance:
Fixed in #3755

PR #3755, opened 2025-04-28
Ignore submodules

d-torrance:
This should help prevent accidentally updating submodule commits.

Closes: #3717


mahrud (approved):

mahrud:
Beginning to think this was not the right solution and should be reverted. I was really hoping that `git commit -a` would not include changes to ignored but changed submodules ...

d-torrance:
Yeah I agree -- I've accidentally almost included submodule changes in PR's a couple times since this was merged

PR #3757, opened 2025-04-29
Fixing small bug in ToricHigherDirectImages

sashahbc:
There was a missing minus sign in the method for modules, which made outputs different for sheaves vs. modules. I fixed this and also added a test to check if the outputs for different input types are the same.


d-torrance (approved):
Thanks!

PR #3761, opened 2025-05-01
Isomorphism: v2.0

mahrud:
The biggest change is that `isIsomorphic` now returns a boolean, and to access the cached isomorphism the function `isomorphism` is added.

- **moved symbol Tries to Core**
- **moved symbol Strict to Core**
- **added option Tries to isIsomorphic**
- **added caching to isIsomorphic**
- **added a few quick strategies for isIsomorphic**
- **moved a test in Isomorphism**
- **moved a test in Isomorphism**
- **fixed a test in Isomorphism**
- **fixed indentation of package Isomorphism**
- **added isomorphism(Module, Module) to Isomorphism**
- **mentioned documentation of Tries to isIsomorphic**
- **changed isIsomorphic to return Boolean**
- **changed isomorphism to give a homogeneous map if possible**
- **fixed a bug in an Isomorphism example**
- **fixed a bug in isIsomorphic for multigraded modules**
- **updated version and package info for Isomorphism**
- **updated isIsomorphic and added isomorphism in Varieties**
- **commented isIsomorphic from TateOnProducts**
- **updated isIsomorphic and isomorphism methods across several packages**
- **mentioned the change in isIsomorphic in changes.m2**

### DIFF
--- a/M2/Macaulay2/packages/EagonResolution.m2
+++ b/M2/Macaulay2/packages/EagonResolution.m2
@@ -1519,8 +1519,6 @@ assert (golodBetti (coker vars R,b) == betti H)
 -- TODO: PLACE INTO M2 CORE
 --SOME USEFUL INTERNAL FUNCTIONS
 --    "compositions"
---    "isDegreeZeroSurjection",
---    "isIsomorphic"
 -*
 compositions(ZZ,ZZ,ZZ) := (nparts, k, maxelem) -> (
     -- nparts is the number of terms
@@ -1541,32 +1539,6 @@ compositions(ZZ,ZZ,ZZ) := (nparts, k, maxelem) -> (
      compositionn = null;
      result
      );
-
-isDegreeZeroSurjection := method()
-isDegreeZeroSurjection(Module,Module) := Boolean => (A,B) -> (
-    --tests a random degree 0 map f:A --> B to see whether its a surjection, 
-    --and returns the answer. If "true" and  o.Verbose == true then returns f.
-    H := Hom(A,B);
-    B0 := basis(0,H); -- this seems to be total degree 0 in case of degreeLength>1
-    f := homomorphism(B0*random(source B0, (ring B0)^1));
-    coker f == 0
-)
-
--- the following function works for modules in the graded case.
--- and produces (only) graded isomorphisms.
-isIsomorphic = method()
-isIsomorphic(Module,Module) := (A,B) -> (
-    --tests random degree 0 maps A->B, B->A and returns true
-    --if both are surjective.
-    if not(isHomogeneous A and isHomogeneous B) then 
-	  error"not implemented for inhomogeneous modules";
-    Ap := prune A;
-    Bp := prune B;
-    dA := set flatten degrees source gens Ap;
-    dB := set flatten degrees source gens Bp;
-    if dA =!= dB then return false;
-    degreeZeroSurjection(Ap,Bp) =!= null and degreeZeroSurjection(Bp,Ap) =!= null
-    )
 *-
 
 TEST/// --test of degreeZeroSurjection
--- a/M2/Macaulay2/packages/Isomorphism.m2
+++ b/M2/Macaulay2/packages/Isomorphism.m2
@@ -171,6 +171,18 @@ isIsomorphismFree = (N, M0, o) -> (
     setIsomorphism(N, M0, map(N, M0, id_M^p2 // id_N^p1,
 	    Degree => -diffdegs), o))
 
+-- returns a minimization M1, presented as
+-- a cokernel of m, and an isomorphism back to M
+minimization = M -> (
+    k := quotient ideal vars ring M;
+    m := presentation M;
+    iso := if m ** k == 0
+    then map(M, M1 := coker m, 1)
+    else (
+	m = presentation(M1 = prune M);
+	M1.cache.pruningMap);
+    (M1, m, iso))
+
 -- TODO: hookify and split into strategies, then add strategies that:
 -- - check if pruning of the modules is cached already
 -- - check if Hilbert polynomial, regularity, dimension, etc. are known
@@ -196,26 +208,12 @@ isIsomorphic(Module, Module) := Boolean => o -> (N, M) -> (
     if tries > 1 then return any(tries,
 	i -> isIsomorphic(N, M, o, Tries => 1));
     --
-        if o.Homogeneous == true and 
-	        not (isHomogeneous M and isHomogeneous N) then 
-	        error"inputs not homogeneous";
-	resS := S/(ideal gens S);
+    if o.Homogeneous == true and not (
+	isHomogeneous M and isHomogeneous N)
+    then error "inputs not homogeneous";
 
-    	m := presentation M;
-
-	if m**resS == 0 then 
-	    (M1 := M; 
-	     m1 := map(M, coker m, 1)) else
-	     (m = presentation (M1 = prune M);
-	     m1 = M1.cache.pruningMap
-    	     );--iso from M1 to M
-	 
-    	n := presentation N;
-	if n**resS == 0 then 
-	    (N1 := N;
-	    n1 := map(N,coker n, 1)) else
-	    (n = presentation (N1 = prune N);
-	    n1 = N1.cache.pruningMap); --iso from N1 to N
+    (M1, m, isoM) := minimization M;
+    (N1, n, isoN) := minimization N;
 
 	--handle the cases where one of M,N is 0
 	isZM1 := target m ==0;
@@ -248,8 +246,8 @@ isIsomorphic(Module, Module) := Boolean => o -> (N, M) -> (
 	kmodule := coker vars S;
 	gbar := kmodule ** g;
 
-	if coker gbar != 0 or kernel g != 0 then false else (
-	    setIsomorphism(N, M, n1*g*(m1^-1), o))
+    if coker gbar != 0 or kernel g != 0 then false
+    else setIsomorphism(N, M, isoN * g * inverse isoM, o)
     )
 isIsomorphic(Matrix, Matrix) := Boolean => o -> (n, m) -> isIsomorphic(coker m, coker n, o)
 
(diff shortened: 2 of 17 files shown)


Devlin-Mallory:
Am I misunderstanding what it should do, or is `isomorphism` (and `getIsomorphism`) not working? Running the following:
`S = ZZ/2
M = S^1
N = S^1
isIsomorphic(M, N) -- true
isomorphism(M,N) -- returns nothing 
`
doesn't give me the isomorphism M->N it found in isIsomorphic. (I think the issue is the same for non-free modules.)

Devlin-Mallory:
There may also be a small bug when checking isIsomorphic over rings that are fields:
```
S=(ZZ/2) -- or QQ
isIsomorphic (coker matrix{{0,1_S},{1,0}},coker matrix{{1_S,0},{0,0}}) -- fails with "error: expected ideal of the same ring"
S=(ZZ/2)[x]
isIsomorphic (coker matrix{{0,1_S},{1,0}},coker matrix{{1_S,0},{0,0}}) -- correctly answers false
S=(GF 4)
isIsomorphic (coker matrix{{0,1_S},{1,0}},coker matrix{{1_S,0},{0,0}}) -- fails with "error: expected a polynomial ring or free algebra"
S=(GF 4)[x]
isIsomorphic (coker matrix{{0,1_S},{1,0}},coker matrix{{1_S,0},{0,0}}) -- correctly answers false
S=ZZ
isIsomorphic (coker matrix{{0,1_S},{1,0}},coker matrix{{1_S,0},{0,0}}) -- also correctly answers false
```

Devlin-Mallory:
> Isomorphism wasn't designed for vector spaces or ZZ-modules, so I'm not surprised. Try checking your examples with the old version.

Issue is the same there! Just wanted to flag it in case it should be changed. The other issue however (of not returning the isomorphism) is still the case when S=(ZZ/2)[x] (I've updated above).

mahrud:
> (Also, just wanted to note that isIsomorphic worked over ZZ, just not over fields, which I find surprising!)

It didn't:
```m2
i1 :   M = ZZ^1 ++ coker(2 * id_(ZZ^1))

o1 = cokernel | 0 |
              | 2 |

                              2
o1 : ZZ-module, quotient of ZZ

i2 :   N = coker(2 * id_(ZZ^1)) ++ ZZ^1

o2 = cokernel | 2 |
              | 0 |

                              2
o2 : ZZ-module, quotient of ZZ

i3 :   isIsomorphic(N, M)

o3 = (false, )

o3 : Sequence
```

M2/Macaulay2/packages/EagonResolution.m2
```diff
@@ -1541,32 +1539,6 @@ compositions(ZZ,ZZ,ZZ) := (nparts, k, maxelem) -> (
...
-    Ap := prune A;
-    Bp := prune B;
-    dA := set flatten degrees source gens Ap;
-    dB := set flatten degrees source gens Bp;
-    if dA =!= dB then return false;
-    degreeZeroSurjection(Ap,Bp) =!= null and degreeZeroSurjection(Bp,Ap) =!= null
-    )
 *-
 
 TEST/// --test of degreeZeroSurjection
```
eisenbud, line 1544: Why is the test still here if the function has been removed?

mahrud: I kept `degreeZeroSurjection` in place because it's used elsewhere in the package, by `homologyIsomorphism`.

M2/Macaulay2/packages/Isomorphism.m2
```diff
@@ -102,438 +133,718 @@ checkDegrees(Module, Module) := Sequence => o -> (A,B) -> (
...
+
+-- key to cache known isomorphisms
+protect Isomorphisms
+
+-- currently we cache isomorphisms ONLY in the source
+setIsomorphism = (N, M, f, o) -> (
+    f = reduceCoefficient f;
+    o = { o.Strict or all(degree f, zero),
+	o.Homogeneous or isHomogeneous f };
+    M.cache.Isomorphisms ??= new MutableHashTable;
```
eisenbud, line 153: cache in younger of M,N

M2/Macaulay2/packages/Isomorphism.m2
```diff
@@ -102,438 +133,718 @@ checkDegrees(Module, Module) := Sequence => o -> (A,B) -> (
...
+	and ( not o.Homogeneous or o.Homogeneous and o'#1 )
+	then return M.cache.Isomorphisms#isom);
+    if N.cache.?Isomorphisms then
+    for isom in keys N.cache.Isomorphisms do (
+	(M', o') := isom;
+	if M' === M
+	and ( not o.Strict      or o.Strict      and o'#0 )
+	and ( not o.Homogeneous or o.Homogeneous and o'#1 )
+	then return inverse N.cache.Isomorphisms#isom))
+
```
eisenbud, line 173: simplify b y age

M2/Macaulay2/packages/Isomorphism.m2
```diff
@@ -102,438 +133,718 @@ checkDegrees(Module, Module) := Sequence => o -> (A,B) -> (
...
+isIsomorphic = method(
+    TypicalValue => Boolean,
+    Options => {
+	Homogeneous => true,  -- if false, allow an inhomogeneous isomorphism
+	Strict      => false, -- forces strict equality of degrees
+	Verbose     => false,
+	Tries       => null,  -- number of attempts at generating a random map
+    })
+isIsomorphic(Module, Module) := Boolean => o -> (N, M) -> (
+    -- returns a pair (false, null) or (true, f), where f is an isomorphism M --> N.
```
eisenbud, line 219: does not return pair

eisenbud (approved):

M2/Macaulay2/packages/Varieties/SheafMaps.m2
```diff
@@ -191,12 +200,25 @@ isIsomorphic(CoherentSheaf, CoherentSheaf) := Sequence => o -> (F, G) -> (
...
+isomorphism(CoherentSheaf, CoherentSheaf) := SheafMap => o -> (F, G) -> (
+    if F === G then id_F else if isIsomorphic(F, G, o,
+	Strict => true, Homogeneous => true)
+    -- FIXME: this is probably not correct yet, because we may
+    -- need to compose/precompose with the pruning maps of F and G.
+    then --map(F, G,
+	--inverse (prune F).cache.pruningMap *
+	sheaf isomorphism splice(G.cache.Isomorphisms#F,
+	    Strict => true, Homogeneous => true)
+	--* (prune G).cache.pruningMap)
```
mahrud, line 218: @Devlin-Mallory the easy was I tried to get `isomorphism` to give a map with correct source and target didn't seem to work. I think there's something wrong in pruning map that should be fixed separately.

Devlin-Mallory: is there a good example where this doesn't give the right answer? (in any case, it seems fine for now)

M2/Macaulay2/packages/Varieties/SheafMaps.m2
```diff
@@ -167,17 +167,26 @@ ZZ == SheafMap := Boolean => (n, f) -> f == n
...
+    and G.cache.Isomorphisms#?F then return true;
+    (M, N) := (module prune F, module prune G);
+    if isIsomorphic(M, N, o, Strict => true, Homogeneous => true)
+    then (( G.cache.Isomorphisms ??= new MutableHashTable )#F = (M, N); true)
+    else false)
 
 -- TODO: perhaps better would be to construct random
 -- maps F --> G and check their kernel and cokernel.
-isIsomorphic(CoherentSheaf, CoherentSheaf) := Sequence => o -> (F, G) -> (
+isIsomorphic(CoherentSheaf, CoherentSheaf) := Boolean => o -> (F, G) -> F === G or (
```
Devlin-Mallory, line 186: I do want to reiterate what @kellerlv and I suggested a while back: that the default method should be to prune both sheaves, with the user able to opt out, since the current method will essentially always do a *nontrivial* truncation, even if F and G are both equal to their prunes and then check for isomorphism of the nontrivial truncations, which can be quite slow. (No need to go back and forth on this further, so feel free to disregard.)

Devlin-Mallory (approved):

M2/Macaulay2/packages/Varieties/SheafMaps.m2
mahrud: Yeah, I know. I am warming up to that idea, but just for the record, my objection is that the user can always call `isIsomorphic(prune F, prune G)` to opt into what you suggest. So the question is which should be the default, and I actually think using the current default has been helpful for discovering bugs.

That said, getting the isomorphism to have actually the right source and target may be harder in the truncation strategy, so I may end up commenting it out out of laziness ...

mahrud: I just added an unused Strategy option for isIsomorphic and isomorphism, which we may later use for this purpose.

(I also tempted fate and decided to skip the tests, since the addition was minimal and the documentation built fine ...)

mahrud: Aha! David found an example where comparing pruned sheaves is incorrect but truncation works:
```m2
  R = QQ[x,y,z]
  X = Proj R
  F = sheaf(R^1/(x,y))
  F1 = sheaf HH^0(F(>=-1))
  F2 = sheaf HH^0(F(>=-10))
  isIsomorphic(F1,F2) -- gives true, using truncation
  isIsomorphic(
      module prune F1,
      module prune F2, Strict => true) -- gives false
```
Of course, the solution would be to truncate the pruned sheaves and compare the truncations, which is my point exactly :)

M2/Macaulay2/packages/Varieties/SheafMaps.m2
mahrud: Incidentally, `isomorphism(F1,F2)` from the other thread is an example of this an isomorphism where the source and target are not set correctly.

PR #3763, opened 2025-05-01
Msolve parser fix

### LINKED ISSUE #3424: Msolve issues

antonleykin:
Fixed the parser for `msolve` output.
Note:
* we replace `int` with `mpz_t` and `mpz_class` where appropriate
* this change affects the classes used by `mathicgb` (but we have no issues so far)

### DIFF
--- a/M2/Macaulay2/e/BasicPoly.hpp
+++ b/M2/Macaulay2/e/BasicPoly.hpp
@@ -12,9 +12,12 @@
 class BasicPoly
 {
 public:
-  std::vector<int> mCoefficients;
+  std::vector<mpz_t> mCoefficients;
   std::vector<int> mComponents; // if zero length: all components are 0.
   std::vector<int> mMonomials; // a concatenated list of varpower monomials.  Each first entry is its length.
+
+  void clear(); // resets all data to represent the zero polynomial
+  ~BasicPoly() { clear(); }
   
   size_t termCount() const { return mCoefficients.size(); } 
   void debug_display(std::ostream& o) const;
--- a/M2/Macaulay2/e/BasicPoly.hpp
+++ b/M2/Macaulay2/e/BasicPoly.hpp
@@ -8,11 +8,12 @@
 #include <iosfwd>
 #include <string>
 #include <unordered_map>
-  
+#include <gmpxx.h>
+
 class BasicPoly
 {
 public:
-  std::vector<mpz_t> mCoefficients;
+  std::vector<mpz_class> mCoefficients;
   std::vector<int> mComponents; // if zero length: all components are 0.
   std::vector<int> mMonomials; // a concatenated list of varpower monomials.  Each first entry is its length.
 
--- a/M2/Macaulay2/e/BasicPolyList.cpp
+++ b/M2/Macaulay2/e/BasicPolyList.cpp
@@ -40,7 +40,7 @@ void BasicPolyListStreamCollector::appendExponent(VarIndex index, Exponent expon
   mValue[mCurrentPoly].mMonomials[mSizeEntryInMonomial] += 2;
 }
 
-void BasicPolyListStreamCollector::appendTermDone(Coefficient coefficient)
+void BasicPolyListStreamCollector::appendTermDone(const Coefficient& coefficient)
 {
   mValue[mCurrentPoly].mCoefficients[mCurrentTerm] = coefficient;
 }
--- a/M2/Macaulay2/e/BasicPolyList.hpp
+++ b/M2/Macaulay2/e/BasicPolyList.hpp
@@ -25,16 +25,17 @@ long bytesUsed(const BasicPolyList& F);
 class BasicPolyListStreamCollector
 {
 public:
-  using Coefficient = int32_t;
+  using Coefficient = mpz_class;
   using VarIndex = int32_t; // TODO: these must match BasicPoly above.
   using Exponent = int32_t;
   using Component = int32_t;
+  using ModulusType = int32_t;
 private:
   BasicPolyList mValue;
 
   // We store these, as we need to be able to respond to what they are,
   // but we don't use them here at all.
-  Coefficient mModulus;
+  ModulusType mModulus;
   VarIndex mVarCount;
   Component mComCount;
 
@@ -56,7 +57,7 @@ public:
 
   // Fields required for the general stream interface (see mathicgb::mathicgb.h)
   
-  Coefficient modulus() const { return mModulus; }
+  ModulusType modulus() const { return mModulus; }
   VarIndex varCount() const { return mVarCount; }
   Component comCount() const { return mComCount; }
 
@@ -64,7 +65,7 @@ public:
   void appendPolynomialBegin(size_t termCount);
   void appendTermBegin(Component com);
   void appendExponent(VarIndex index, Exponent exponent);
-  void appendTermDone(Coefficient coefficient);
+  void appendTermDone(const Coefficient& coefficient);
   void appendPolynomialDone();
   void idealDone();
 };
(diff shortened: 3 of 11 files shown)


mahrud:
It passes the main example that @Martin-Helmer mentioned [here](https://github.com/Macaulay2/M2/pull/3392#issuecomment-2274579351). Could you actually add this as a test?
```m2
needsPackage "Msolve";
R=QQ[x..z,t]
K=ideal(x^6+y^6+x^4*z*t+z^3,36*x^5+60*y^5+24*x^3*z*t,
	-84*x^5+10*x^4*t-56*x^3*z*t+30*z^2,-84*y^5-6*x^4*t-18*z^2,
	48*x^5+10*x^4*z+32*x^3*z*t,48*y^5-6*x^4*z,14*x^4*z+8*x^4*t+24*z^2)
W1=msolveEliminate(R_0,K)
W2=eliminate(R_0,K)
sub(W1,R)==W2 
```

M2/Macaulay2/e/BasicPolyListParser.cpp
```diff
@@ -59,6 +58,41 @@ long readInteger(const std::string_view& str, size_t& begin_loc, size_t end_loc)
   return result;
 }
 
+// allocates memory to (and inits) an __mpz_struct
+// void readInteger_mpz_t(mpz_t& result, const std::string_view& str, size_t& begin_loc, size_t end_loc)
```
mahrud, line 62: What's the difference between this and `readInteger_mpz_class`?

M2/Macaulay2/packages/Msolve/tests.m2
```diff
@@ -106,19 +106,26 @@ TEST ///
...
-  needsPackage "Msolve";
   R = QQ[x..z,t]
   K = ideal(x^6+y^6+x^4*z*t+z^3,36*x^5+60*y^5+24*x^3*z*t,
       -84*x^5+10*x^4*t-56*x^3*z*t+30*z^2,-84*y^5-6*x^4*t-18*z^2,
       48*x^5+10*x^4*z+32*x^3*z*t,48*y^5-6*x^4*z,14*x^4*z+8*x^4*t+24*z^2)
   errorDepth=2
   W1 = msolveEliminate(R_0, K, Verbosity => 1)
   W2 = eliminate(R_0, K)
-  sub(W1, R) == W2
+  assert(sub(W1, R) == W2)
```
mahrud, line 117: Oh, never mind, it's already here ..

mahrud (approved):
This looks good to me, but Mike and Martin want to take a look, too.

M2/Macaulay2/e/BasicPoly.hpp
```diff
@@ -8,13 +8,17 @@
...
 #include <string>
 #include <unordered_map>
-  
+#include <gmpxx.h>
+
 class BasicPoly
 {
 public:
-  std::vector<int> mCoefficients;
+  std::vector<mpz_class> mCoefficients;
```
mikestillman, line 16: The function `bytesUsed` will need to change.  We don't really need to do that before this PR is accepted, but it should be changed.

M2/Macaulay2/e/BasicPolyList.cpp
```diff
@@ -40,7 +40,7 @@ void BasicPolyListStreamCollector::appendExponent(VarIndex index, Exponent expon
   mValue[mCurrentPoly].mMonomials[mSizeEntryInMonomial] += 2;
 }
 
-void BasicPolyListStreamCollector::appendTermDone(Coefficient coefficient)
+void BasicPolyListStreamCollector::appendTermDone(const Coefficient& coefficient)
 {
   mValue[mCurrentPoly].mCoefficients[mCurrentTerm] = coefficient;
 }
```
mikestillman, line 46: Is this correct?  Does this correctly copy the element? I think it probably does, just want to flag that

M2/Macaulay2/e/BasicPolyList.hpp
```diff
@@ -25,16 +25,17 @@ long bytesUsed(const BasicPolyList& F);
 class BasicPolyListStreamCollector
 {
 public:
-  using Coefficient = int32_t;
+  using Coefficient = mpz_class;
   using VarIndex = int32_t; // TODO: these must match BasicPoly above.
   using Exponent = int32_t;
   using Component = int32_t;
+  using ModulusType = int32_t;
```
mikestillman, line 32: What characteristics do we want to use here?  mathic uses roughly 16 bit primes, while msolve seems to want much larger primes.  Are they still less than e.g. 2^31?

M2/Macaulay2/e/BasicPolyListParser.cpp
```diff
@@ -184,7 +217,7 @@ void parseBasicPoly(const std::string_view& str, const IdentifierHash& idenHash,
                 {
                   throw parsing_error("expected a digit at position " + std::to_string(begin_loc));
                 }
-              e = readInteger(str, begin_loc, end_loc);
+              e = readInteger_long(str, begin_loc, end_loc);
```
mikestillman, line 220: exponent vectors are ints.  Are we silently truncating numbers to int size here?

M2/Macaulay2/e/coeffrings.hpp
```diff
@@ -66,6 +66,15 @@ class CoefficientRingZZp : public M2::SimpleARing<CoefficientRingZZp>
     if (a < 0) a += p;
     result = log_table[a];
   }
+  
+  void set_from_mpz(elem &result, mpz_t a) const
+  {
+    mpz_t tmp;
+    mpz_init_set_si(tmp, p);        // Convert int p to mpz_t
+    mpz_mod(a, a, tmp);             // a = a mod p (always non-negative)
```
mikestillman, line 74: Maybe we should keep an mpz_t characteristic here too, so this doesn't need to be done each time?

mikestillman (approved):
There are a few things I will want to change in this PR, but it does fix the msolve reading large integers issue, and it doesn't mess up the mathicgb computations, so I say: let's take it.  @antonleykin and I can talk through the thoughts I have about making it more future robust.

Thanks @antonleykin for fixing that!

PR #3766, opened 2025-05-02
Set googletest to version 1.16.0

d-torrance:
We're currently using a commit a little between 1.16 and the just-released 1.17.  googletest 1.17 (and the commit we're currently using) requires compiling with C++17 support.  This is currently an issue on systems like RHEL where we need to build googletest.

Eventually, we should add support for 1.17 by compiling with C++17 support in memtailor/mathic/mathicgb, but it would probably be best to deal with this after the 1.25.05 release.

This commit is cherry-picked from #3683, but I'd like to make sure it gets into the release for the sake of the RHEL builds (see https://github.com/d-torrance/M2-workflows/actions/runs/14789884670).

### DIFF
--- a/M2/libraries/gtest/Makefile.in
+++ b/M2/libraries/gtest/Makefile.in
@@ -1,11 +1,9 @@
-# from https://github.com/google/googletest/archive/release-1.8.0.tar.gz, renamed to gtest-1.8.0.tar.gz
 SUBMODULE = true
 LIBNAME = googletest
-VERSION = 1.11.0
+VERSION = 1.16.0
 CONFIGURECMD = cmake . -DCMAKE_INSTALL_PREFIX=$(PREFIX) -DBUILD_GMOCK=OFF
 
 LICENSEFILES = LICENSE
-URL = https://macaulay2.com/Downloads/OtherSourceCode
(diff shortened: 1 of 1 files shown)


mahrud:
Ah, I think the version of googletest required by mathicgb and friends also needs to be bumped to v1.16.0, otherwise on systems with newer CMake we get an error. Could you update those as well?

d-torrance:
Isn't the particular version of googletest that's installed for mathicgb and friends controlled by the M2 repository?

mahrud:
No, their CMake files currently point to v1.11.0.

d-torrance:
Oh I see.  So that's what it downloads when it can't find googletest?  Wouldn't it be okay if 1.16 was already downloaded by the Macaulay2 build?

I guess my question is -- should we worry about updating all of those submodules prior to the release?

mahrud:
Yes, I currently can't build with cmake on my system.

d-torrance:
Ok -- I'll open some PR's here in a bit.  I have an autotools fix that I'd like to get in too -- right now it's just patches in the release branch, but it would be better to have them in the submodule I think.

PR #3767, opened 2025-05-02
Macaulay2 1.25.05

d-torrance:
*Target date:* May 15, 2025

TODO:

- [ ] Finish updating the changelog
- [ ] Update the Emacs submodule commit once the symbols have been updated
- [ ] Update prism.js with updated symbols
- [ ] Update the vim files with updated symbols

We'll rebase on top of development regularly.

See also https://github.com/Macaulay2/M2/milestone/19


M2/BUILD/docker/Makefile
```diff
@@ -22,6 +22,23 @@ build: build-image
 
 build-image:;	docker build $(BUILD_ARGS) --tag $(TAG) .
 
+## Script for building Macaulay2 (autotools)
+define M2_BUILD_SCRIPT_autotools
```
mahrud, line 26: Put this in `ubuntu/Makefile` or create `autotools/Makefile`, not here.

M2/BUILD/docker/README.md
```diff
@@ -21,6 +21,11 @@ The `Dockerfile` in this directory creates a container image based on latest Ubu
 1. Build a Macaulay2 container:
 ```
 make build
+```
+
+   Or, using the autotools build:
+```
+make build-autotools
```
mahrud, line 28: I don't think you should add this here because it's misleading. In `M2_BUILD_SCRIPT_autotools` you run `make` which does way more than just build M2-binary. The point of the docker system is that you can run M2 in such a way that packages from the clone are loaded and you can edit them and have that one run, so running the entire build process is extra and might cause confusion if files aren't loaded from the clone.

d-torrance: Ok, that makes sense I think.  I've really only played around with these Dockerfiles to try building M2 in non-Debian-based Linux distros where I *do* want to run `make`to make sure that everything works.

mahrud: Yeah, I figured this is what you wanted. I don't think there's any harm on having many different directories with Docker/Makefiles specialized to different purposes, while the top directory minimal.

PR #3768, opened 2025-05-02
Update Visualize package for leadTerm changes

d-torrance:
`leadTerm` now returns an ideal instead of a matrix

This code was missed in https://github.com/Macaulay2/M2/pull/3574, probably since the Visualize methods aren't automatically tested and need to be manually tested in the browser.

Cc: @pzinn


pzinn:
I'm not sure I understand the version bump. what's the etiquette in such cases?

d-torrance:
I'm one of the authors of this package  :)  (Not one of the the original authors, but I got their blessing to add myself a few years ago and have been the de facto maintainer ever since.)  I generally do a version bump every release if there have been changes.

Issue #1130, opened 2020-05-08
use of the master branch


DanGrayson:
It has been pointed out that the master branch should always compile for people.  Since we have overnight builds, we can have those builds track a development branch, and after success, the development branch could be pulled into the master branch.

We'd have to tell people to direct their pull requests to the development branch instead of master, since github may not do that for us.

radoslavraynov:
In the short term, you can do without forcing people to resubmit. You can select where to merge their pull request: just open the PR, then [Edit] (the the far-right side), then select `base:feature/dev` from the dropdown. But in the long term, it'd be nice if people pushed straight to the development branch, to avoid merge failures.

mahrud:
When people work on multiple PR's against feature/dev, wouldn't all of them get merge conflicts and have to rebase whenever one of them gets merged?

I see the point of having a dev branch for pushing individual commits so multiple maintainers can collaborate on fixing something while master branch still builds, but I'm confused about this: if a PR (e.g. big like eigen, or just a package update) doesn't get merged until building with it succeeds, why not merge it directly to master? Assuming the requirement for merging was that it builds!

I hope I explained the question clearly.

DanGrayson:
Well, it's same situation as now, except with "master", so I don't see the problem.

I never rebase.  If the upstream master changes, I merge it into my branch.

We don't want people pushing to Macaulay2/M2, except for Mike and me.  Others should submit pull requests instead.

The problem with merging pull requests into master is that they might break master, because we don't test every pull request by rebuilding from scratch and running all the tests.  The advantage of an overnight build on the development branch is that we can be a little more care-free about pulling and trust that in the morning, an email will arrive if something  broke.

And actually, we're going to have to be careful when merging to merge with the precise commit that was tested, for the head of the development branch might advance after commencing testing and before merging with master.

radoslavraynov:
@DanGrayson, you can carry out the nightly testing on a staging branch or lightweight tag (e.g. `staged/20200509`) to avoid conflicts with advanced `feature/dev`.

DanGrayson:
I think I'll call the branch `development`.

DanGrayson:
I think implementation of this can wait until after we pull the eigen branch request and the cmake request.  In fact, the overnight job is now running on the eigen branch, usefully.

DanGrayson:
[...]

PR #3769, opened 2025-05-02
Remove "repository code URI" key from certified packages

d-torrance:
There's no reason to hardcode the package filename since we already know it.

It's also gives us fewer references to the `master` branch, and so we wouldn't need to worry about changing anything if/when we rename it (#1130).


M2/Macaulay2/m2/help.m2
```diff
@@ -249,7 +249,9 @@ documentationValue(Symbol, Package)         := (S, pkg) -> if pkg =!= Core then
 	    cert  = new HashTable from cert;
 	    -- TODO: compare with the one in installPackage.m2
 	    star := IMG { "src" => replace("PKG", "Style",currentLayout#"package") | "GoldStar.png", "alt" => "a gold star"};
-	    commit := replace("(?<=/blob/)master", toString cert#"release at publication", cert#"repository code URI");
+	    commit := concatenate(
+		"https://github.com/Macaulay2/M2/blob/",
+		cert#"release at publication", "/M2/Macaulay2/packages/", fn);
```
mahrud, line 254: But someone might download this file thinking they can use it, and find it unusable, no?

mahrud: Ah you're right. I still think this should be changed, but that's separate from this PR.

d-torrance:
I realized that this broke the links for the three packages that have been renamed since they were published in JSAG.  So I added  "legacy name" key to the `Certification` option for each of these, which is used for building the URL when it exists.

M2/Macaulay2/m2/help.m2
```diff
@@ -251,11 +251,15 @@ documentationValue(Symbol, Package)         := (S, pkg) -> if pkg =!= Core then
 	    star := IMG { "src" => replace("PKG", "Style",currentLayout#"package") | "GoldStar.png", "alt" => "a gold star"};
 	    commit := concatenate(
 		"https://github.com/Macaulay2/M2/blob/",
-		cert#"release at publication", "/M2/Macaulay2/packages/", fn);
+		cert#"release at publication", "/M2/Macaulay2/packages/",
+		(cert#"legacy name" ?? pkg#"pkgname") | ".m2");
```
mahrud, line 255: Since we also have the published code URI, what's the point of also linking to an old version on git that may or may not work?

d-torrance: Good point.  And for that matter, do we even have a strong reason to link directly to the JSAG version of the code?  It's linked from the article page.

mahrud: I don't mind either way on that.

d-torrance: Ok -- I think I'll keep the JSAG code link so at least we have one link to the legacy code but remove the commit info.  It will simplify certifying packages -- finding the exact commit that matches the JSAG version is an annoying step in the release process that I wouldn't mind doing away with.  :)

mahrud: I would say hold off on completely deleting the hash from existing ones, just in case we had a use for it later, but feel free to not populate new ones until it's needed?

mahrud (approved):

PR #3770, opened 2025-05-02
fix set Set, add toList Tally

pzinn:
this is another (minor) fix of [#3574](https://github.com/Macaulay2/M2/pull/3574): the definition of `set Set` didn't work since `set` is defined at d level.
Instead we let `set` be equivalent to `keys` on hash tables. We also add a missing `toList Tally`.

### DIFF
--- a/M2/Macaulay2/d/sets.dd
+++ b/M2/Macaulay2/d/sets.dd
@@ -19,6 +19,7 @@ makeSet(e:Expr):Expr := (
      is v:Sequence do makeSet(v)
      is w:List do if ancestor(w.Class,visibleListClass) then makeSet(w.v)
      else WrongArg("a visible list")
+     is h:HashTable do makeSet(keys(h))
      else WrongArg("a visible list"));
 setupfun("set",makeSet);
 
--- a/M2/Macaulay2/m2/set.m2
+++ b/M2/Macaulay2/m2/set.m2
@@ -15,7 +15,7 @@ tally String      :=
 tally VisibleList := Tally => tally
 
 elements = method()
-elements Tally := x -> splice apply(pairs x, (k,v) -> v:k)
+elements Tally := toList Tally := x -> splice apply(pairs x, (k,v) -> v:k)
 
 toString VirtualTally := x -> concatenate( "new ", toString class x, " from {", demark(", ", sort apply(pairs x, (v,i) -> (toString v, " => ", toString i))), "}" )
 
@@ -85,7 +85,6 @@ Set.synonym = "set"
 -- constructors, both compiled functions defined in d/sets.dd
 set VisibleList := Set => set
 new Set from List := Set => (X,x) -> set x
-set Set := identity
 
 -- set operations
 elements Set := List => keys
--- a/M2/Macaulay2/packages/Macaulay2Doc/functions/set-doc.m2
+++ b/M2/Macaulay2/packages/Macaulay2Doc/functions/set-doc.m2
@@ -1,6 +1,5 @@
 undocumented {
     (NewFromMethod, Set, List),
-    (set, Set)
 }
 
 document {


M2/Macaulay2/d/sets.dd
```diff
@@ -19,6 +19,7 @@ makeSet(e:Expr):Expr := (
      is v:Sequence do makeSet(v)
      is w:List do if ancestor(w.Class,visibleListClass) then makeSet(w.v)
      else WrongArg("a visible list")
+     is h:HashTable do makeSet(keys(h))
```
mahrud, line 22: Could you add a typical value comment so `methods set` also lists `(set, HashTable)`?

M2/Macaulay2/m2/set.m2
```diff
@@ -15,7 +15,7 @@ tally String      :=
 tally VisibleList := Tally => tally
 
 elements = method()
-elements Tally := x -> splice apply(pairs x, (k,v) -> v:k)
+elements Tally := toList Tally := x -> splice apply(pairs x, (k,v) -> v:k)
```
mahrud, line 18: Could you document  `(toList, Tally)` (perhaps just adding it to the documentation of `(elements, Tally)`?

M2/Macaulay2/d/sets.dd
mahrud: Oh, actually, this is worse than `set Set := identity` because it allocates new memory, no?

pzinn: yeah I'll rewrite it more cleanly, more like top-level `applyValues(h,x->1)`

mahrud: Actually, an easier solution might be to make it so people can install methods on set!

pzinn: yeah I don't know about that. we've been doing that a lot (superseding the d function, e.g., `toList` has suffered this fate) but I don't think it's great, it's not very clean. I'd rather basic functions like `set` be at d level.

mahrud: What is not "clean" about this? There are lots of constructions in mathematics that produce a set from an object. It seems very reasonable to allow package writers to define a type and say how `set T` should produce a set.

pzinn: what's not clean is having a d function, then renaming it at the m2 level, then introducing a new m2 function which shares the same name as the original d function but which potentially behaves differently. it's not good practice and can make debugging confusing.

d-torrance: Another common approach is to give the d function a slightly different name, often appending "0" to the name.  So `set` could be the method function, and some of the installed methods could be the compiled function `set0`.

mahrud: Suppose you could declare methods in the interpreter and directly install new method functions on them in the top level. Would this be confusing? I agree that renaming the function and assigning a method is kind of hacky (and the `set0` trick is not much better), but the solution to that is to make it possible to declare methods in the interpreter, not prevent `set` from becoming a method in the top level!

pzinn: > Suppose you could declare methods in the interpreter and directly install new method functions on them in the top level.
yeah that'd be nice... but unless I'm mistaken this isn't currently possible. I was trying to keep this PR to a minimum.

M2/Macaulay2/m2/set.m2
pzinn: decided to add it to `toList`... but why does the doc say that the input can be an integer??
```
A, a set, a string, a basic list, *or an integer*, or an instance of a class with the iterator method installed
```

M2/Macaulay2/d/sets.dd
mahrud: Yes, it's not possible. My point was that what you called unclean is a kludge until we have this feature, not that you should implement it now :)

Anyway, I think I'd like set to become a method, either in this PR or another time.

M2/Macaulay2/m2/expressions.m2
```diff
@@ -1240,7 +1240,7 @@ texMath MapExpression := x -> texMath x#0 | "\\," | (if #x>2 then "\\xleftarrow{
 expressionValue MapExpression := x -> map toSequence apply(x,expressionValue)
 
 -- moved from set.m2 because of loadsequence order
-expression Set := x -> Adjacent {set, expression keys x}
+expression Set := x -> Adjacent {set, expression sortByName keys x}
```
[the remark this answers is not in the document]
pzinn: then why do we do it for `HashTable`?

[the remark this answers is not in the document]
mahrud: We probably shouldn't, it's misleading.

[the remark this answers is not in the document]
pzinn: so should I just revert the sorting for `Set`, or also remove the sorting for `HashTable`?

n-m-g:
I don't know too much about the background on this specific  topic but I agree that it shouldn't matter if hashTable keys are sorted or not.  (At least all checks seem to have passed so it is probably OK.)  However, a situation where this might be used (and which might cause bugs that are difficult to detect) is the case of a hash table with integer keys.  One would need to look into that further and some methods may implicitly assume the keys to be sorted.  A related topic concerns the "spots" function (which I learned from Dan Grayson's ChainComplex code many years ago) and which I think should be part of the core.  But this "spots" function and its need has caused some controversy.  The aim of the sports function is to output the sorted integer keys (if any) of a hashTable.

M2/Macaulay2/m2/expressions.m2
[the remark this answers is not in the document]
mahrud: I think just remove it here for now, but separately we should discuss whether both should be sorted or unsorted.

M2/submodules/googletest
mahrud, line 1: Submodules were inadvertently changed.

n-m-g: This is interesting (and is the kind of thing that I was worried about).  As another comment about "complexity" I'm not too worried about it in terms of "sorting" since if the output is small enough to sort, then it shouldn't take too long.  But as you say it seems "not natural" to sort per se.  I'll refrain from additional comments about this for now.

pzinn: how can I revert the submodule change? this is driving me crazy.

d-torrance: The following should get you back to where it's supposed to be (possibly replacing `origin` with whatever remote you use for the main M2 repo):

```
git checkout origin/development -- M2/submodules/googletest
```

pzinn:
Rereading the changes in this PR, was `toList ZZ` supposed to be like
```
toList ZZ := n -> toList(0..n-1)
```
that would be pretty useful

mahrud:
> Rereading the changes in this PR, was `toList ZZ` supposed to be like
> 
> ```
> toList ZZ := n -> toList(0..n-1)
> ```
> 
> that would be pretty useful

I would be worried about people using `apply(toList N, ...)` instead of just `apply(N, ...)`.

d-torrance:
These commits were included in #3929.  Closing

PR #3772, opened 2025-05-03
Error reporting fixes

mahrud:
These are a few cosmetic fixes for how errors are reported. One non-cosmetic change is that C^c C^c now interrupts `check` again.

- **replaced * with = in non-error printouts**
- **improved error reporting missing methods in documents**
- **updated check and runFile**

### DIFF
--- a/M2/Macaulay2/m2/run.m2
+++ b/M2/Macaulay2/m2/run.m2
@@ -153,7 +153,7 @@ runFile = (inf, inputhash, outf, tmpf, pkg, announcechange, usermode, examplefil
      -- print debugging info
      printerr "running failed";
      printerr("error log:  " | logLocation tmpf);
-     printerr("test input: " | testLocation inf);
+     printerr("input code: " | testLocation inf);
      if debugLevel > 0 then
      stderr << commentize "full command:" << endl << cmd << endl;
      stderr << pad("*** error: M2 " | describeReturnCode r, printWidth - 32) << flush;
--- a/M2/Macaulay2/m2/run.m2
+++ b/M2/Macaulay2/m2/run.m2
@@ -15,10 +15,24 @@ ulimit := utest "-c unlimited" | utest "-t 700" | utest "-m 850000" | utest "-s 
 M2statusRegexp := "^--status:"
 statusLines := file -> select(lines file, s -> match(M2statusRegexp,s))
 
+testLocation := inf -> (
+    infcontents := get inf;
+    -- TODO: is this feature ever used?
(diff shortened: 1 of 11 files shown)


M2/Macaulay2/m2/run.m2
```diff
@@ -128,29 +142,25 @@ runFile = (inf, inputhash, outf, tmpf, pkg, announcechange, usermode, examplefil
...
-     scan(statusLines infcontents, x -> stderr << x << endl);
-     m := regex("^--([^\n]+): location of test code", infcontents);
-     if m =!= null then stderr << (
-	 substring(m#1, infcontents)) << ": (location of test code)" << endl;
-     if # findFiles rundir == 1
-     then removeDirectory rundir
+     -- print debugging info
+     printerr "running failed, too!";
+     printerr("error log:  " | logLocation tmpf);
+     printerr("test input: " | testLocation inf);
```
d-torrance, line 156: Isn't `runFile` used for both tests and examples?  Is this too test-specific?

Perhaps there could be a test-specific method that calls `runFile`, and if there was an error, print this debugging info.  And if that method already knows the `TestInput` object, we wouldn't need to worry about also storing the file location in the test string and using a regex to extract it.

mahrud: That sounds great, but I'll be honest, I hate `runFile` and I don't want to be the one to separate it out XD I wrote this PR because I was simply _too_ frustrated that C-c C-c didn't work and that the test location wasn't printed anymore.

But I will happily change "test input" to something else in a new commit and skip CI.

d-torrance: Fair enough :)  Maybe I'll tackle `runFile` this summer.

M2/Macaulay2/packages/Macaulay2Doc/functions/help-doc.m2
```diff
@@ -396,6 +396,6 @@ Node
       TABLE {
 	  { M2CODE "loadedPackages",                            "-- a list of the currently loaded packages" },
 	  { M2CODE "help \"packages provided with Macaulay2\"", "-- a list of all the available packages" },
-	  { M2CODE "help \"initialization file \"",             "-- show documentation about the file init.m2" }
+	  { M2CODE "help \"initialization file\"",              "-- show documentation about the file init.m2" }
```
d-torrance, line 399: Oops -- that was my bad!  Glad you caught it!

d-torrance (approved):
This is great!  The debugging info is really useful.  Thanks!

PR #3773, opened 2025-05-04
Ip/game theory

iremportakal:
This is a pull request for the GameTheory package, which started at the Macaulay2 in the Sciences workshop at MPI MiS in Leipzig. The package includes methods for computing algebro-geometric and combinatorial models of Nash, correlated, dependency, and conditional independence equilibria from game theory. 

cc: @lkastner @chenyangzhaoicl @ZhuangHe-CosmicBeverages @HannahTillmannMorris @erinconnelly96 @GiacomoMaletto @elkenhs @VincenzoGalgano


M2/Macaulay2/packages/GameTheory.m2
```diff
@@ -0,0 +1,2711 @@
...
+         HomePage => "https://www.irem-portakal.de"},
+      {Name => "Hannah Tillmann-Morris",
+         Email => "tillmann@mis.mpg.de",
+         HomePage => "https://sites.google.com/view/hannah-tillmann-morris"},
+      {Name => "Chenyang Zhao",
+         Email => "cz2922@ic.ac.uk",
+         HomePage => "https://www.felixzhao.com/"}
+   },
+   Headline => "A package for computing equilibria in game theory",
+   Keywords => {"Game Theory","Equilibria","Nash","Correlated","Dependency","Spohn","Conditional Independence"},
```
mahrud, line 35: The Keyword here is meant to put the package into one or more categories from [this page](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_packages_spprovided_spwith_sp__Macaulay2.html). Is there one that you think fits best with this package? If not, we could start a new "Game Theory" category:
```suggestion
   Keywords => {"Game Theory"},
```

mahrud:
Ah, directProductList isn't exported.
```m2
i1 : Q = simplex 3;  -- a tetrahedron should just return that unchanged

i2 : assert(directProductList {Q} === Q)
stdio:3:24:(3): error: no method for adjacent objects:
            directProductList (of class Symbol)
    SPACE   {Q} (of class List)

stdio:1:5:(3): error: test(s) #18 of package GameTheory failed.
```

M2/Macaulay2/packages/GameTheory.m2
```diff
@@ -0,0 +1,2711 @@
...
+    ideal fullGeneratingSet
+    )
+
+---------------------------------------------------------------
+-- directProductList (List)
+--
+-- Given a list of polytopes, it computes their direct product.
+---------------------------------------------------------------
+
+directProductList = method()
```
d-torrance, line 431: The tests are currently failing because this method is unexported:

```m2
i1 : 
     Q = simplex 3;  -- a tetrahedron should just return that unchanged

i2 : assert(directProductList {Q} === Q)
stdio:3:24:(3): error: no method for adjacent objects:
            directProductList (of class Symbol)
    SPACE   {Q} (of class List)
```

So `directProductList` should either be expored, or `debug` called like in the previous test.

However, I'm wondering if it would be simpler to make `directProduct` in *Polyhedra* a binary method.  Then we'd get this behavior out of the box.  (If we declare that a method `f` is binary, then we compute `f(x,y,z)` using `f(f(x,y),z)` automatically without having to define a 3-argument version, and so on for 4 or more arguments.)

In particular, with one small change:

```diff
--- a/M2/Macaulay2/packages/Polyhedra/extended/directProduct.m2
+++ b/M2/Macaulay2/packages/Polyhedra/extended/directProduct.m2
@@ -1,6 +1,6 @@
 
 -- PURPOSE : Computing the direct product of two polyhedra in the direct product of their ambient spaces
-directProduct = method()
+directProduct = method(Binary => true)
 
 --   INPUT : '(P,Q)',  two polyhedra
 --  OUTPUT : A polyhedron, the direct product
```

We'd get:

```m2
i1 : Q = simplex 3;

i2 : directProduct {Q} == Q

o2 = true

i3 : dim directProduct {simplex 1, simplex 2}

o3 = 3

i4 : dim directProduct {simplex 1, simplex 2, simplex 3, simplex 4, simplex 5}

o4 = 15
```

d-torrance (changes requested):

M2/Macaulay2/packages/GameTheory.m2
chenyangzhaoicl: Hi, sorry I coded this method and forgot to export, as it wasn't a related method to the package's theme. I agree that making directProduct in Polyhedra a binary method is simpler and better as it's more related to the Polyhedra package instead of the GameTheory one, and it makes the computation of direct product easier for other people in the long run.

M2/Macaulay2/packages/GameTheory.m2
```diff
@@ -0,0 +1,2711 @@
...
+P = directProductList {P1,P2};
+assert(dim P == 3);
+
+v1 = entries transpose vertices P1;
+v2 = entries transpose vertices P2;
+expectedVerts = flatten apply(v1, u -> apply(v2, w -> u | w));
+assert(sort entries transpose vertices P == sort expectedVerts)
+///
+
+TEST ///
```
chenyangzhaoicl, line 2420:
```suggestion
TEST ///
debug needsPackage "GameTheory"
```
I added the `debug` we needed for the other test.

chenyangzhaoicl (changes requested):

M2/Macaulay2/packages/GameTheory.m2
chenyangzhaoicl: I added `debug` as what I did for the previous test. Forgot to `debug` again for the second test, sorry!

M2/Macaulay2/packages/GameTheory.m2
mahrud: This doesn't actually add the change. Either you or Irem need to push this change to Irem's repository, or Irem can click "commit suggestion" to do it automatically.

M2/Macaulay2/packages/GameTheory.m2
mahrud: @d-torrance it would be good to make the change you suggest in this release, but separately.

d-torrance: Done in https://github.com/Macaulay2/M2/pull/3775

iremportakal:
Committed the changes. I am happy to replace the directProductList method in the next version of the package.

d-torrance (approved):
Looks good -- thanks!

d-torrance:
Just noticed that this was targetting `master` instead of `development`.  There probably aren't any issues, but let's let the tests run one more time to make sure.

chenyangzhaoicl (approved):

PR #3777, opened 2025-05-06
SLP expressions abbreviate printing

antonleykin:
These are minor changes to the way SLPexpressions are displayed following a discussion with @pzinn 
 sometime ago. Now there is a heuristic guess on how complex a `Gate` is --- only if it is not too complex it is printed fully.

(I've cherry-picked a branch that later accumulated more commits that are not relevant to this.)

### DIFF
--- a/M2/Macaulay2/packages/SLPexpressions.m2
+++ b/M2/Macaulay2/packages/SLPexpressions.m2
@@ -903,12 +903,12 @@ expression DetGate := g -> (expression det) eGM g.Inputs
 printLargeGate = g -> VerticalList { class g, "depth"=>depth g, "#gates"=> countGates g }
 net Gate := g -> (
     n := net expression g;
-    if depth g > 5 and width n <= printWidth and height n <= printWidth then n
+    if depth g < 6 and width n <= printWidth and height n <= printWidth then n
     else net printLargeGate g
     )
 html Gate := g -> (
     e := expression g;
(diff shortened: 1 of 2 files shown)


M2/Macaulay2/packages/SLPexpressions.m2
```diff
@@ -884,6 +889,28 @@ FY = value(diff(Y,F),h)
...
+expression DetGate := g -> (expression det) eGM g.Inputs
+printLargeGate = g -> VerticalList { class g, "depth"=>depth g, "#gates"=> countGates g }
+net Gate := g -> (
+    n := net expression g;
+    if width n <= printWidth and height n <= printWidth then n
+    else net printLargeGate g
+    )
+html Gate := g -> (
+    e := expression g;
+    html if width net e <= printWidth then e else printLargeGate g
```
pzinn, line 911: the problem with testing `width net e` is that it forces to first produce the net output before deciding whether it's worth printing everything or not. It's usually a better idea to find a simpler criterion (number of ... whatever gates are made of) to make this choice. same remark for `net Gate`.

antonleykin: This sounds reasonable. One attribute that is computed fast (faster than generating a net) is the depth. Printing is now attempted only if the depth is small.

pzinn: I would completely remove the part about `net e` in the tests, but I won't argue.

pzinn (approved):

PR #3779, opened 2025-05-06
update package GeometricDecomposability; add package IntegerProgramming

mgacummings:
1. Minor updates to the package GeometricDecomposability, including updating references and output type hints for exported methods. Fixed an issue in the package tests by replacing `loadPackage` with `needsPackage`. Rewrote method `isLexCompatiblyGVD` using an equivalent definition that should be faster.

2. Introduces the package IntegerProgramming, a small package to solve integer linear programs using Gröbner bases (using an algorithm from Cox, Little, O'Shea's Using Algebraic Geometry). This was developed as part of Mike Stillman's course Computational Commutative Algebra and Algebraic Geometry at the Fields Institute earlier this year.

### DIFF
--- a/M2/Macaulay2/packages/=distributed-packages
+++ b/M2/Macaulay2/packages/=distributed-packages
@@ -280,3 +280,4 @@ NumericalSemigroups
 IncidenceCorrespondenceCohomology
 ToricHigherDirectImages
 Brackets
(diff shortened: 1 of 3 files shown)


d-torrance (approved):
Thank you!

PR #3788, opened 2025-05-07
fix issue 3361 (rref giving wrong answer)

mikestillman:
I was using flint matrices incorrectly: I thought the matrix was stored contiguously, in row major order. 

It is stored contiguously, but contains pointers to the first element if each row, and those can be modified during a flint computation.  However, my code taking the answer from a flint computation did not respect that data.  Additionally, this way of storing the start of each row was not well suited to my use of iterators for these matrix types.

So this PR removes these iterators, and mostly removes the access to the underlying contiguous memory.  However SLP code used this isn a way that I didn't understand, so one commit adds back a method `unsafeArray` for access to this data.  We need to check that SLP evaluation is still correct (probably true, depending on whether these matrices are constructed using flint or not).

One other thing I changed: in lapack code, there was heavy use of iterators, so I removed those, and changed the signature of some helper functions, requiring changes in a number of places in the lapack code file. I removed use of garbage collected memory there as well, while I was changing these functions.

I am making this draft, until @antonleykin can audit the SLP use mentioned above: does every use of `unsafeArray` (I think 6 calls) use the matrix in a correct way?

Sorry that one of these commits is kind of monolithic (but makes the changes in iterators and array, and fixes all uses of these).


mikestillman:
I should probably make sure tests have been added too.  I don't think I did that...

fredrik-johansson:
> I was using flint matrices incorrectly: I thought the matrix was stored contiguously, in row major order.

> It is stored contiguously, but contains pointers to the first element if each row, and those can be modified during a flint computation. However, my code taking the answer from a flint computation did not respect that data. Additionally, this way of storing the start of each row was not well suited to my use of iterators for these matrix types.

FYI, FLINT 3.3 will change the matrix format, getting rid of the array of row pointers and storing entries contiguously in row major order.

mikestillman:
> > I was using flint matrices incorrectly: I thought the matrix was stored contiguously, in row major order.
> 
> > It is stored contiguously, but contains pointers to the first element if each row, and those can be modified during a flint computation. However, my code taking the answer from a flint computation did not respect that data. Additionally, this way of storing the start of each row was not well suited to my use of iterators for these matrix types.
> 
> FYI, FLINT 3.3 will change the matrix format, getting rid of the array of row pointers and storing entries contiguously in row major order (or rather with a fixed stride between the rows, to allow window matrices).

Thanks for the heads up!  I have a couple of questions: when do you think flint 3.3 will release?  In that release, are you still using row pointers to quickly access elements, or do you have another method for doing that?

fredrik-johansson:
> > > I was using flint matrices incorrectly: I thought the matrix was stored contiguously, in row major order.
> > 
> > 
> > > It is stored contiguously, but contains pointers to the first element if each row, and those can be modified during a flint computation. However, my code taking the answer from a flint computation did not respect that data. Additionally, this way of storing the start of each row was not well suited to my use of iterators for these matrix types.
> > 
> > 
> > FYI, FLINT 3.3 will change the matrix format, getting rid of the array of row pointers and storing entries contiguously in row major order (or rather with a fixed stride between the rows, to allow window matrices).
> 
> Thanks for the heads up! I have a couple of questions: when do you think flint 3.3 will release? In that release, are you still using row pointers to quickly access elements, or do you have another method for doing that?

The row pointers have been removed entirely. Accessing an element just looks like this:

    fmpz * fmpz_mat_entry(const fmpz_mat_t mat, slong i, slong j)
    {
        return mat->entries + i * mat->stride + j;
    }

The multiplication shouldn't be slower than an extra array access on recent CPUs; I haven't noticed any slowdowns after switching to the new format anyway, and it does speed up some things.

We were already going to release 3.3 a few weeks ago but decided to hold out a bit more for a big PR. The author of that PR hasn't replied however, so maybe we'll do a release without it. "Hopefully soon". :-)

antonleykin:
> I am making this draft, until @antonleykin can audit the SLP use mentioned above: does every use of `unsafeArray` (I think 6 calls) use the matrix in a correct way?

JIT compilation of SLPs is an experimental feature (at this point it doesn't work on all platforms and is not tested by the build). It is supposed to work only over standard precision real and complex numbers. 

I've tested on Mac OS --- JIT compilation works fine for RR but segfaults over CC.  
However, this is so both with and without this PR.  I'll have to dig into this later. 
This PR seems to be good to go.

mikestillman:
I want to add some tests before we merge this.  @d-torrance Currently, where should we be putting tests?  The problem with `tests/normal` is that (I think?) these are not available at run time with an installed relase build, whereas if I put them in e.g. `Macaulay2Doc/functions/functionname-doc.m2` then they are available.  Which location should I be placing tests?

mikestillman:
I doubt that I will go back to the older iterators for DMat after flint 3.3 is available, but I do want to check to make sure the multiplication in each entry access doesn't slow linear algebra down first.  Not worrying about it is much simpler, and if it isn't slower, produces much easier to read code.

mikestillman:
@d-torrance Also, I guess, if Core is kind of an actual package soon, tests could be placed in that package somehow?

antonleykin:
> > The problem with tests/normal is that (I think?) these are not available at run time with an installed relase build
> 
> This is not true. You can always run `M2 --check 2` to run the tests from `tests/normal`, which are distributed in `share/doc/Macaulay2/Core/tests`.

You can also run them interactively: type
```
tests Core
```
to  see which tests are there and where they are: 
```
     ...
     {303 => TestInput[/opt/homebrew/share/doc/Macaulay2/Core/tests/smith.m2:1:1]                  }  
     ...
```
and then run the one you want, e.g.,
```
i2 : check_303 Core                                                                                                                                   
 -- capturing check(303, "Core")                                                                                      -- 1.55115s elapsed             
```
runs the bit that is in `tests/normal/smith.m2`.

d-torrance:
Ideally, I'd like to see pre-flint 3.3 iterators supported for a while for backwards compatibility with older systems.

mikestillman:
It sounds like the flint interface won't change, just the internal implementation.  Flint doesn't have iterators for matrices, as far as I can tell (please let me know if I missed them somehow!) instead, they use entry(mat,r,c) type functions (which give references to, and the actual element, there are several of these, eg. get_entry, set_entry).

The only difference seems to be how these are/will be handled internally.

The new DMat routines I wrote should be compatible with either, although possibly slower with the 3.3 flint version, but it sounds like it is not a significant difference.

mikestillman:
@antonleykin @mahrud Thanks for recalling for me that the tests are available for installed versions!  I guess I should add the tests to tests/normal directory then.

mikestillman:
@d-torrance If this passes all tests, it can be merged.

However, there is one thing I did that might not be great.  Over RR_53 (and CC_53), I am using now the generic LU decomposition code, rather than the (faster) LU code in lapack.  This is because the LU decomposition that lapack returns is not good enough to compute things like nullspace (one of the git issues that was fixed here).

So if anyone is doing large LU decompositions over RR, and the matrix is full rank, there might be an efficiency regression here.  I feel that this needs to be addressed hopefully in the next release, but for now, giving correct answers is better than giving fast, not correct answers.

mikestillman:
I was a little too aggressive in thinking that the "naive" version of LU would suffice.  It does not.  I'm going to revert that change, so that git issue 2941 still remains unfixed.  This is an issue that if the matrix is full rank, then lapack does a great job with it.  If it isn't, then lapack doesn't care so much about the matrix... (meaning that there is no point in continuing an LU decomposition to get row echelon form, as it probably isn't going to be used anyway.  But we want to use it...

So some more thought is needed for that.  I will check in the change in a moment, after I run some local tests

PR #3789, opened 2025-05-08
Notify when random seed changes

mahrud:
The point of this change is to inform users when the global state of randomness has been affected, and hopefully avoid confusing debugging problems as a result. It would be even better to do this in the interpreter or Engine, but this should suffice.


M2/Macaulay2/m2/system.m2
```diff
@@ -144,10 +144,12 @@ exit = Command exitMethod
 
 setRandomSeed = method(Dispatch => Thing)
 setRandomSeed ZZ := seed -> (
+    if recursionDepth() >= 3 then
```
d-torrance, line 147: Why is this 3 and not 1?

d-torrance: Oh never mind, I figured it out!

mahrud: Idk, recursion depth is not precise when using methods.

M2/Macaulay2/tests/ComputationsBook/completeIntersections/test.out.expected
```diff
@@ -1,5 +1,6 @@
 
 i1 : setRandomSeed();
+ -- initializing random seed
```
d-torrance, line 3: Should the `recursionDepth` check be preventing this?

mahrud: I thought so, but somehow it didn't and the tests still failed, so I removed the check.

d-torrance (approved):

PR #3791, opened 2025-05-08
Add package AllMarkovBases

ollieclarke8787:
This package allows for computing all minimal Markov bases of a toric ideal, counting and sampling from the set of minimal Markov bases, and computation of indispensable binomials and universal Markov basis.


M2/Macaulay2/packages/AllMarkovBases.m2
```diff
@@ -0,0 +1,1403 @@
+newPackage(
+    "AllMarkovBases",
+    Version => "1.0",
+    Date => "May 08, 2025",
+    Headline => "package for computing all minimal Markov bases
+    of a configuration matrix",
```
d-torrance, line 6: Drop the redundant "package for" and put everything on a single line, i.e.:

```suggestion
    Headline => "computing all minimal Markov bases of a configuration matrix",
```

d-torrance (changes requested):
Thanks for the contribution!  I noticed a couple small issues with the `Headline` option to `newPackage` that will cause the tests to fail.

d-torrance (approved):
Thanks!

PR #3794, opened 2025-05-08
MatrixSchubert version 1.2

aalmousa:
Fixed a bug in `isASMUnion` and typos in the documentation. This version will soon appear in the Journal for Software in Algebra and Geometry (JSAG).


d-torrance (changes requested):
All the files should be up a directory:

`M2/Macaulay2/packages/MatrixSchubert/MatrixSchubert.m2` -> `M2/Macaulay2/packages/MatrixSchubert.m2`

And all the auxiliary files:

`M2/Macaulay2/packages/MatrixSchubert/MatrixSchubert/*` -> `M2/Macaulay2/packages/MatrixSchubert/*`

Also, the emacs submodule commit was updated accidentally

mahrud:
I asked Ayah and force pushed to get rid of the submodule changes. I think the problem is that the fix for ignoring the submodules is included in development, but not yet in the main branch, if people start with the main branch and add their changes, the submodule changes get included. I think this will be resolved soon when the fixed gets merged into the main branch.

d-torrance (approved):
Thank you!

PR #3796, opened 2025-05-09
Set correct installation directory for gtest (autotools)

d-torrance:
On some systems (e.g., RHEL), cmake installs it in *usr-host/lib64* by default, but we expect to find it in *usr-host/lib*.

@mahrud - I know this is an autotools update, but googletest uses cmake for its build.  Is this the correct way to set the library installation directory?  It worked for me when I tested it.


mahrud (approved):
I haven't tested this, but yes I think this is the right way to change the lib directory. The relevant documentation is [here](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html).

mahrud:
Is this related to this PR?
https://github.com/Macaulay2/M2/actions/runs/14941402099/job/41989052502?pr=3774

d-torrance:
It's failing a teeny bit earlier in that build -- in the header file detection stage.  The newer (post-1.16) googletest headers throw errors if you're not compiling with C++17 support.  This PR was dealing with the next step -- once we've found the headers, try to link against it.

mahrud:
Are you sure? It's related to gtest, limited to autotools, and started happening last night.

d-torrance:
Pretty sure.  See https://github.com/Macaulay2/M2/pull/3774#discussion_r2083197738

PR #3797, opened 2025-05-09
Package citations

mahrud:
This is my attempt at #3715.

- **added Citation key for documentation and SimpleDoc**
- **Add syntax highlighting for BibTeX to documentation**
- **upgraded cite method to look at package documentation**
- **added example citation for TerraciniLoci**

Here's the default, if no citation is provided:
image

On the other hand, given a section like:
```m2
doc ///
  Key
    TerraciniLoci
  Headline
    package for computing Terracini loci
  Description
...
  Citation
    @article{Galuppi_2025,
      title={Geometry of First Nonempty Terracini Loci},
      ISSN={1793-6683},
      url={http://dx.doi.org/10.1142/S0219199725500531},
      DOI={10.1142/s0219199725500531},
      journal={Communications in Contemporary Mathematics},
      publisher={World Scientific Pub Co Pte Ltd},
      author={Galuppi, Francesco and Santarsiero, Pierpaola and Torrance, Douglas A. and Turatti, Ettore Teixeira},
      year={2025},
      month=apr }
///
```

You get:
image

Moreover, `cite` returns that string.

I think this is good enough for this release, if we want, but there are several features that I think would be nice:
- reorder Author and Version and Citation and Certificate, or even combine them in some way that makes sense
- allow authors to add more than just the bibtex reference, e.g. provide two references and explain in which situation which one should be cited.


mahrud:
Also, note that I didn't change the citation of M2 itself, which should still be decided I think.

mahrud:
Probably _some_ packages shouldn't have citation info, right? e.g. Style, but I'm not sure where to draw the line.

d-torrance:
Although no one will likely cite *Style*, I don't see any harm in having citation info show up on its documentation page.

mahrud:
Just feels unnecessary and crowding. I also think perhaps having citation box in too many places might make people gloss over it, whereas if only packages which the authors think may be cited include citation, then people may notice it.

d-torrance (approved):

PR #3798, opened 2025-05-10
Alternate random(QQ) proposal

### LINKED ISSUE #3782: Engine unit-test regression
I just noticed this on a few builds based on various different pull requests:
```m2
        Start 126: unit-tests:RingQQ.negate
126/459 Test #126: unit-tests:RingQQ.negate .......  Passed   12.50 sec
        Start 127: unit-tests:RingQQ.add
127/459 Test #127: unit-tests:RingQQ.add ..........  Passed  331.39 sec
        Start 128: unit-tests:RingQQ.subtract
128/459 Test #128: unit-tests:RingQQ.subtract .....  Passed  345.02 sec
        Start 129: unit-tests:RingQQ.multDivide
129/459 Test #129: unit-tests:RingQQ.multDivide ...  Passed  339.06 sec
        Start 130: unit-tests:RingQQ.axioms
130/459 Test #130: unit-tests:RingQQ.axioms .......  Passed  358.77 sec
        Start 131: unit-tests:RingQQ.power
131/459 Test #131: unit-tests:RingQQ.power ........  Passed   12.68 sec
        Start 132: unit-tests:RingQQ.syzygy
132/459 Test #132: unit-tests:RingQQ.syzygy .......  Passed  336.35 sec
```
Compare to the same tests on the main branch:
```m2
        Start 126: unit-tests:RingQQ.negate
126/459 Test #126: unit-tests:RingQQ.negate ..........  Passed    0.03 sec
        Start 127: unit-tests:RingQQ.add
127/459 Test #127: unit-tests:RingQQ.add .............  Passed    0.03 sec
        Start 128: unit-tests:RingQQ.subtract
128/459 Test #128: unit-tests:RingQQ.subtract ........  Passed    0.04 sec
        Start 129: unit-tests:RingQQ.multDivide
129/459 Test #129: unit-tests:RingQQ.multDivide ......  Passed    0.04 sec
        Start 130: unit-tests:RingQQ.axioms
130/459 Test #130: unit-tests:RingQQ.axioms ..........  Passed    0.04 sec
        Start 131: unit-tests:RingQQ.power
131/459 Test #131: unit-tests:RingQQ.power ...........  Passed    0.04 sec
        Start 132: unit-tests:RingQQ.syzygy
132/459 Test #132: unit-tests:RingQQ.syzygy ..........  Passed    0.04 sec
```

And this is not a problem on GitHub, because locally the tests take ages too.

cc: @mikestillman

d-torrance:
This reverts much of #3481, which hopefully will fix #3782.  (On my system, the engine unit tests once again pass in under a second after this change.)

The Farey approximation code is still around and is still used by the top-level `random(QQ)` method, but it's no longer used by the engine.

I figure we can explore other options for probability distributions on the rationals after the release.


mahrud:
I would rather we don't revert it, even if we have to disable those tests.

d-torrance:
I just pushed an alternate proposal.  It's essentially the same as the previous behavior, but we re-roll if the numerator and denominator share a common factor.  This has a few advantages, I think:

* The support is the same as it's always been, and in particular, the notion of *height* as a bound on both numerator and denominator is correct.
* It's a uniform distribution.
* It's much faster than the Farey approximation code.  On my system, the unit tests are finishing in under a second.

mahrud:
What's the benefit of using Farey approximation in the top level still? I'm a bit worried about confusion if two methods produce different kinds of random results because one uses engine and one top level random.

Did caching not help, by the way? Ideally we would want the top level example that Mike ran to also be fast.

d-torrance:
Yeah, I'm not convinced that keeping the Farey implementation for the top-level `random(QQ)` is the best either.  But I'm not sure how best to implement it.

It seems like `random x` should return a random rational number on the interval [0, x] with denominator bounded by the `Height` option.  But while this new proposal gives a uniform distribution on the set of rationals with a given height bound, it *doesn't* give us a uniform distribution in an interval.  For example, with the default height of 10, about half of the the time we'll get something in [0, 1] and the other half we'll get something in [1, 10].

I haven't tried a cached Farey approximation yet, but I'll play around with it and maybe open a separate PR for comparison.

d-torrance:
I wrote a little C++ program that caches all the Farey sequences up to a given height and then does a binary search to figure out the closet one to a given double.  It works great with a small height, but caching takes a while if the height is big -- 7 seconds on my system for a height of 1000.  And the height that we use in the unit tests is 100000000000 lol.  

Although Farey works great for small heights, I think it's just too slow in general.

d-torrance:
Another option might be re-implementing the current code using 64-bit ints instead of of GMP ints...  I'll play around with that

mahrud:
> I wrote a little C++ program that caches all the Farey sequences up to a given height

Why all of them? Usually people want to compute random numbers with a fixed height many many times, just compute the Farey sequence for the given height that was asked once. It takes a few seconds the first time:
```m2
i1 : elapsedTime random(QQ, Height => 100000000000)
 -- 3.21021s elapsed

     160441622955811
o1 = ---------------
           2048

o1 : QQ
```
but after this I presume it should be instantaneous.

d-torrance:
The current implementation doesn't compute the entire Farey sequence -- just the nearby rationals.

So for example, let's say we want to find the Farey approximation of $\pi$.  We first lop off the 3, save that for later, and look at 0.14159...  We check if it's to the left or right of 1/2 -- since it's to the left, then we look at the "mediant" of 0/1 and 1/2, which is (0 + 1)/(1 +2) = 1/3.  It's less than that, so we look at the next median: (0 + 1)/(1 + 3) = 1/4.  Still less than that, so we get the next mediant: (0 + 1)/(1 + 4) = 1/5.  Next is (0 + 1)/(1 + 5) = 1/6, then (0 + 1)/(1 + 6) = 1/7, and then (0 + 1)/(1 + 7) = 1/8.  It's bigger than that, so we look for the mediant of 1/7 and 1/8.  That's (1 + 1)/(7 + 8) = 2/15, but if we're using the default height of 10, then we know to stop.  We compare 0.14159 with 1/7 and 1/8, and since it's a bit closer to 1/7, we use that.  Adding the 3 back, we return 22/7.

The point being that we only computed a little bit of the Farey sequence.  If the height is really big, then I don't know how useful caching that little bit will be since chances are we won't visit that little bit again.

As I type this out, I realize that caching things using a binary tree (so basically the [Stern-Brocot tree](https://en.wikipedia.org/wiki/Stern%E2%80%93Brocot_tree)) might be more useful.  At the very least, we wouldn't be duplicating near as much data.  I'll play around some more with it.

mahrud:
If there really isn't a difference between distributions, then I say let's just accept this and think about it later.

mikestillman:
Also, if need be, we can reduce the height in those unit tests...

d-torrance:
Ok -- merging this for the release

PR #3809, opened 2025-05-13
Add randomSubset to Core

d-torrance:
There are four different packages that implement their own `randomSubset` function, so I think it makes sense to provide it in Core.

We use Knuth's Algorithm S when the desired cardinality is known, and for arbitrary subsets, first pick a cardinality using a binomial distribution so we get a uniform distribution on the power set.

```m2
i1 : tally apply(10000, i -> randomSubset 4)

o1 = Tally{{} => 681          }
           {0, 1, 2, 3} => 605
           {0, 1, 2} => 553
           {0, 1, 3} => 582
           {0, 1} => 622
           {0, 2, 3} => 593
           {0, 2} => 645
           {0, 3} => 628
           {0} => 649
           {1, 2, 3} => 668
           {1, 2} => 641
           {1, 3} => 602
           {1} => 630
           {2, 3} => 652
           {2} => 636
           {3} => 613

o1 : Tally
```

We also update `VisibleList_List` so that the output has the same type as the input.

@n-m-g - This change makes it so that the random subsets are 0-based instead of 1-based.  Is that okay, or would you prefer it to stay 1-based for AbstractSimplicialComplexes?


M2/Macaulay2/packages/RandomIdeals.m2
```diff
@@ -264,15 +264,6 @@ testNewSimplex(List, List) := (P, D) ->(
 
 intersectLists = (D',D) -> D - set(D-set D')
 
-randomSubset = (n,m) -> (
```
mahrud, line 267: I think the order of arguments here is switched.

d-torrance: My instinct was that it should be (size of subset, size of set), but all four packages used (size of set, size of subset), so I went with it.  I guess that makes sense since it lines up nicely with `binomial`.

n-m-g: I would prefer 1-based since this is how I designed the package (simplicial complexes on {1,...,n} instead of {0,...,n}).     But what would be easier for other packages?

mahrud: > My instinct was that it should be (size of subset, size of set), but all four packages used (size of set, size of subset), so I went with it. I guess that makes sense since it lines up nicely with `binomial`.

This is also how `subsets(A, n)` is defined, but disregard my comment, I looked at the code and thought in this case they meant the order to be the other way. Did you do a timing comparison between your implementation and the others?

n-m-g:
Yes, this all sounds good to me.

Best,
Nathan

M2/Macaulay2/packages/RandomIdeals.m2
d-torrance: Hmm yeah, this is actually slower than a couple of the other packages' implementations...  I'll dig into this some more

d-torrance:
I scrapped the `VisibleList_List` change since it broke a few things (like the `Grassmannian` example).  I may revisit that in another PR, since the following seems strange:

```m2
i1 : [1, 2, 3]_{0,2}

o1 = {1, 3}

o1 : List
```

Shouldn't the return value be [1, 3]?

d-torrance:
@n-m-g: The AbstractSimplicialComplexes functions are all back to 1-based indexing now.

M2/Macaulay2/packages/RandomIdeals.m2
seangrate: > I would prefer 1-based since this is how I designed the package (simplicial complexes on {1,...,n} instead of {0,...,n}). But what would be easier for other packages?

You could do something like `randomSubset(n) / (i -> i+1)` for your use case.

seangrate (approved):

n-m-g:
Great.  Thanks so much!

Best,
Nathan

M2/Macaulay2/m2/lists.m2
```diff
@@ -237,6 +237,34 @@ random List := opts -> s -> (
...
+-- Knuth Algorithm S, Art of Computer Programming, Section 3.4.2
+randomSubset(ZZ, ZZ) := (N, n) -> (
+    if n < 0 or n > N then error("expected an integer between 0 and ", N);
+    t := 0;
+    apply(n, m -> (
+	    while (N - t) * rawRandomRRUniform defaultPrecision >= n - m
+	    do t += 1;
+	    first (t, t += 1))))
+-- when size not given, use binomial distribution first to determine size
+randomSubset ZZ := N -> (
```
pzinn, line 251: for a uniform random subset, it's simpler to flip N coins...

PR #3811, opened 2025-05-13
Install exports.m2 manually since it's no longer in loadsequence

d-torrance:
The autotools build uses this file to determine which .m2 files to install


mahrud:
This is why it was moved. Don't do this; instead add it to the lists that include `Core.m2` and `startup.m2`.

Issue #527, opened 2016-10-21
kernel should allow degree limits


mikestillman:
If we know that the kernel of a ring map is generated in degree d, then we should be able to tell M2 that info.  This issue is from the ask mike session at the M2 day at Fields, Oct 21, 2016.

```
R = ZZ/32003[a..d]
S = ZZ/32003[s,t]
phi = map(S,R,{s^4, s^3*t, s*t^3, t^4})
ker(phi, DegreeLimit=>3)
```

DanGrayson:
This is probably a top level change.

mahrud:
@bkholler's MultigradedImplicitization package does this, but it should become a part of the `kernel` method.

d-torrance:
> **Written by Claude** (Claude Opus 5, via Claude Code), not by @d-torrance, whose account posted it -- please weigh it accordingly.

The feature asked for here exists — `ker` takes `DegreeLimit` and it is documented at
`[kernel, DegreeLimit]`. **Leaving this open anyway, and retitling**, because on the example
in this issue the option silently returns the wrong thing, and that turns out to be a
general property rather than a quirk of the example. The original title, "kernel should
allow degree limits", describes the part that is done.

### It works exactly as documented — when the map is homogeneous

Grading `R` so that this issue's own map is degree-preserving:

```m2
S  = ZZ/32003[s,t];
Rg = ZZ/32003[a..d, Degrees => {4,4,4,4}];
phi = map(S, Rg, {s^4, s^3*t, s*t^3, t^4});
```

```
isHomogeneous phi                       true
kernel generator degrees                {8, 12, 12, 12}

ker(phi, DegreeLimit => 4)              ideal()
ker(phi, DegreeLimit => 8)              ideal(b*c-a*d)
ker(phi, DegreeLimit => 12)             ideal(b*c-a*d, c^3-b*d^2, a*c^2-b^2*d, b^3-a^2*c)
ker(phi, DegreeLimit => 16)             (same)
```

Precisely the documented behaviour: *"stop after generators for the degree `d` part of the
kernel have been discovered."*

### The example in this issue is not homogeneous, and there it misleads

With the default grading — which is what the issue writes — `a` has degree 1 and maps to
`s^4`, so `isHomogeneous phi` is **false**. Then:

```
kernel generator degrees                {2, 3, 3, 3}

ker(phi, DegreeLimit => 2)              ideal()
ker(phi, DegreeLimit => 3)              ideal()     <-- the line in this issue
[...]

PR #3816, opened 2025-05-14
Added options and hookified kernel

mahrud:
The options aren't (currently) used for anything, but having them in there in this release would be helpful for @bkholler to implement #527 using MultigradedImplicitization package by the next release.


d-torrance (approved):
Looks good to me!

PR #3817, opened 2025-05-14
Simplify randomSubset(ZZ)

d-torrance:
We were already generating N Bernoulli random variates and summing them up before calling Algorithm S for small N.  Instead, keep track of the positions of the 1's and return that, avoiding Algorithm S entirely.  Thanks to @pzinn for the suggestion!

Also fix a silly typo in the normal approximation when N is large.

### DIFF
--- a/M2/Macaulay2/m2/lists.m2
+++ b/M2/Macaulay2/m2/lists.m2
@@ -247,19 +247,10 @@ randomSubset(ZZ, ZZ) := (N, n) -> (
 	    while (N - t) * rawRandomRRUniform defaultPrecision >= n - m
 	    do t += 1;
 	    first (t, t += 1))))
--- when size not given, use binomial distribution first to determine size
 randomSubset ZZ := N -> (
-    if N < 0 then error "expected a nonnegative integer"
-    else if N < 30 then (
-	-- for small N, sum up N Bernoulli variates
-	n := 0;
-	scan(N, i -> n += random 2))
-    else (
-	-- for large N, use normal approximation
-	n = round(n/2 + sqrt(n/4)*rawRandomRRNormal defaultPrecision);
-	-- in the vanishingly unlikely case we end up outside [0, N]:
-	n = min(N, max(0, n)));
-    randomSubset(N, n))
+    if N < 0 then error "expected a nonnegative integer";
+    r := random 2^N;
+    for i to N - 1 list if r & 2^i != 0 then i else continue)
 randomSubset(VisibleList, ZZ) := (x, n) -> x_(randomSubset(#x, n))
 randomSubset VisibleList := x -> x_(randomSubset(#x))
 randomSubset(Set, ZZ) := (x, n) -> set randomSubset(toList x, n)


M2/Macaulay2/m2/lists.m2
```diff
@@ -247,19 +247,17 @@ randomSubset(ZZ, ZZ) := (N, n) -> (
...
 	    do t += 1;
 	    first (t, t += 1))))
--- when size not given, use binomial distribution first to determine size
 randomSubset ZZ := N -> (
     if N < 0 then error "expected a nonnegative integer"
     else if N < 30 then (
-	-- for small N, sum up N Bernoulli variates
-	n := 0;
-	scan(N, i -> n += random 2))
+	for i from 0 to N - 1 list if random 2 == 1 then i else continue)
```
pzinn, line 253: I think this algorithm is faster even for N>30. Did you check?

d-torrance: Yeah, I did.  It's about the same.

pzinn: but this one is so much simpler and natural...

mahrud: I think simpler (perhaps even faster) is to generate a (64-bit) integer and use the bits to return integers. For N < 64 this should be near instantaneous. For larger N, split it into 64-bit blocks ...

d-torrance: Ooh, I like the bit idea.  And with GMP, we're not even limited to 64 bits...

d-torrance: Okay yeah, that's way faster.  Thanks!

pzinn: yeah `random 2` is slower than `random RR`. Using bits of a random integer is the way to go.

d-torrance: Here's the new timing comparison (using the random bits is the last column).  Much nicer :)

```m2
      +----+-------+--------+--------+
o20 = |30  |.10018 |.0846158|.0232665|
      +----+-------+--------+--------+
      |50  |.170638|.136678 |.0367224|
      +----+-------+--------+--------+
      |100 |.341813|.254995 |.0783433|
      +----+-------+--------+--------+
      |200 |.695711|.624376 |.144193 |
      +----+-------+--------+--------+
      |500 |1.7485 |1.42215 |.410996 |
      +----+-------+--------+--------+
      |1000|3.41022|2.79286 |.858254 |
      +----+-------+--------+--------+
```

M2/Macaulay2/m2/lists.m2
```diff
@@ -247,19 +247,10 @@ randomSubset(ZZ, ZZ) := (N, n) -> (
...
-	scan(N, i -> n += random 2))
-    else (
-	-- for large N, use normal approximation
-	n = round(n/2 + sqrt(n/4)*rawRandomRRNormal defaultPrecision);
-	-- in the vanishingly unlikely case we end up outside [0, N]:
-	n = min(N, max(0, n)));
-    randomSubset(N, n))
+    if N < 0 then error "expected a nonnegative integer";
+    r := random 2^N;
+    for i to N - 1 list if r & (k := 2^i) == k then i else continue)
```
pzinn, line 253: if you really want to optimise you probably want something like
```
if first(r&1==0,r//=2) then ...
```

pzinn: though it's probably pointless in M2...

d-torrance: I tried it out and it's actually an itty bit slower than the current implementation.  Maybe GMP optimizes computing powers of 2 or something.

pzinn: okay at least remove the needless assignment of `k`

mahrud: Agreed, `r & 2^i != 0` should suffice.

d-torrance: Ok -- I'll switch it to that.  I won't bother running the tests again.  The one failing build was an unrelated ThreadedGB example

PR #3822, opened 2025-05-15
Update MultigradedImplicitization

bkholler:
Added an option to use interpolation to compute polynomials in multigraded components and changes made by @mahrud to speed up multidegree and basis computation.

### DIFF
--- a/M2/Macaulay2/packages/MultigradedImplicitization.m2
+++ b/M2/Macaulay2/packages/MultigradedImplicitization.m2
@@ -16,7 +16,6 @@ newPackage(
   Headline => "solving implicitization problems using multigradings",
   Keywords => {"Algebraic Statistics", "Commutative Algebra"},
   PackageImports => {"gfanInterface"},
-  DebuggingMode => true
 )
 
 --------------------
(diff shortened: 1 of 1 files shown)


M2/Macaulay2/packages/MultigradedImplicitization.m2
```diff
@@ -23,10 +26,11 @@ export {
   -- Methods
   "maxGrading",
   "trimBasisInDegree",
-  "componentOfKernel",
+  "computeComponent",
+  "interpolateComponent",
   "componentsOfKernel",
   -- Options
-  "Grading", "PreviousGens", "ReturnTargetGrading", "UseMatroidSpeedup"
+  "Grading", "PreviousGens", "ReturnTargetGrading", "UseMatroid", "UseInterpolation", "Verbose"
```
d-torrance: `Verbose` is already exported by Core, so you shouldn't need to export it here

mahrud: @bkholler I will make this change and push to your repository. Since it's just an update, I think we can include this PR in the June 1st patch release.

bkholler: @mahrud Fantastic! Thanks!

PR #3824, opened 2025-05-16
Bump msolve to 0.8.0

d-torrance:
msolve 0.8.0 was just released, so we update the cmake and autotools builds to use it if it's not already available.

This is a draft for now.  I'll remove the second commit, which forces the building of msolve in the GitHub builds, if the builds are successful.


d-torrance:
The builds were all successful!
https://github.com/Macaulay2/M2/actions/runs/15067458472

mahrud (approved):
Looks good to me, but I'd wait a week to see if 8.1.0 comes out.

PR #3826, opened 2025-05-16
Uncomment "changes made for the next release" documentation node

d-torrance:
It was commented because it doesn't really make sense to have in a stable release, but we should uncomment it to start documenting changes for 1.25.11.


mahrud:
Maybe add empty stubs for things like new packages, updated features, breaking changes, etc?

mahrud (approved):

PR #3828, opened 2025-05-17
Fixed a bug in random(Module, Module)

mahrud:


M2/Macaulay2/m2/genmat.m2
eisenbud, line 1: Looking at the code I realize that I was "lucky" to hit the error so quickly. This fix (and test) takes care of what I observed.

eisenbud (approved):

PR #3823, opened 2025-05-16
Assorted documentation fixes

mahrud:
- **removed Citation for packages with no authors**
- **added release date on package documentation page**
- **fixed minor issues in NoetherNormalization package**
- **combine different Sasha Zotines**
- **combine two Timothy Duffs**

PR #3830, opened 2025-05-17
Patch release 1.25.06

d-torrance:
*Target date*: June 1, 2025

- [x] Fix for `random(Module, Module)` bug
- [x] #3822
- [x] #3823
- [x] #3831
- [x] #3840 (to enable building Fedora 42 package)
- [x] Fix for #3851
- [ ] Updated citation info?
- [ ] Anything else?


mahrud:
Definitely #3823.

mahrud:
To be clear, we don't want to include new packages, etc. correct? How about package patches? e.g. #3822.

d-torrance:
I would think no new packages or features, but I'm no too picky.  If you think it would be good to have, then I'm down!

mahrud:
By the way, one of the steps in the release should be updating https://github.com/Macaulay2/language-macaulay2, because GitHub syntax highlighting uses it.

d-torrance:
Oh cool -- I always wondered how GitHub did syntax highlighting!

PR #3833, opened 2025-05-18
Added better strategy for random(Module,Module) with free source

mahrud:
I think this is a slight improvement over computing Hom.

### DIFF
--- a/M2/Macaulay2/m2/genmat.m2
+++ b/M2/Macaulay2/m2/genmat.m2
@@ -113,8 +113,10 @@ randomMR := opts -> (F,G) -> (
      map(F,G,new Matrix from f))
 
 random(Module, Module) := Matrix => opts -> (F,G) -> (
-    if not isFreeModule F or not isFreeModule G
+    if not isFreeModule G
     then return homomorphism random(Hom(G, F, DegreeLimit => 0), opts);
+    if not isFreeModule F
+    then return map(F, G, random(cover F, G, opts));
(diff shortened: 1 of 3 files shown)


M2/Macaulay2/tests/normal/randommod.m2
Devlin-Mallory, line 1: I think we should switch the target and source of f here, since a random map S^1 -> S^1/x is always well-defined; it's the other direction that we want to make sure is working.

M2/Macaulay2/m2/genmat.m2
Devlin-Mallory, line 1: looks good!

Devlin-Mallory (commented):
looks good, only comment is the one regarding the test.

M2/Macaulay2/tests/normal/randommod.m2
mahrud: What do you mean? Both directions are tested in the file.

Devlin-Mallory: Ah I see - I missed the lines above testing the other direction. Looks great then!

mahrud:
The macOS cmake build has been taking ages, upgrading the version of Eigen to see if this is the issue.

d-torrance:
@mahrud - Should this go in the patch release?  What about the eigen change?

mahrud:
Sure, I don't think either of them hurt.

PR #3832, opened 2025-05-18
Bump 4ti2 to 1.6.11

d-torrance:
4ti2 1.6.11 was released recently.  Draft for now -- I'll remove the second commit if the builds are successful

PR #3846, opened 2025-05-23
Drop 4ti2's autoconf version down to 2.69 for older systems

d-torrance:
After #3832, all the RHEL builds started to fail because the minimum autoconf version was bumped to 2.71 in 4ti2 1.6.11.  But that doesn't seem to matter because the build still works fine using autoconf 2.69.

### DIFF
--- a/M2/BUILD/docker/rhel/Dockerfile
+++ b/M2/BUILD/docker/rhel/Dockerfile
@@ -18,9 +18,9 @@ RUN dnf install -y 'dnf-command(config-manager)' && \
 RUN dnf -y install autoconf automake bison boost-devel bzip2 cmake     \
     diffutils eigen3 flex gc-devel gcc-c++ gcc-gfortran gdbm-devel git \
     glpk-devel gmp-devel gtest-devel lapack-devel libffi-devel libtool \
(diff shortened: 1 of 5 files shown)


M2/BUILD/docker/rhel/Dockerfile
```diff
@@ -18,9 +18,9 @@ RUN dnf install -y 'dnf-command(config-manager)' && \
 RUN dnf -y install autoconf automake bison boost-devel bzip2 cmake     \
     diffutils eigen3 flex gc-devel gcc-c++ gcc-gfortran gdbm-devel git \
     glpk-devel gmp-devel gtest-devel lapack-devel libffi-devel libtool \
-    libxml2-devel make mpfr-devel ncurses-devel openblas-devel patch   \
-    python3-devel R readline-devel rpm-build tbb-devel which xz-devel  \
-    zlib-devel
+    libxml2-devel make mpfr-devel ncurses-devel ninja-build            \
```
mahrud, line 21: What's ninja used for?

mahrud: Oh I see you just want this docker to work for both cmake and autotools build, never mind.

mahrud (approved):
I didn't test this myself but seems fine!

Issue #3851, opened 2025-05-26
needsPackage bug


mahrud:
The workshop testbot seems to have broken:
```m2
2025-05-26T13:19:54.6482302Z  -- Found 1 test file(s) matching '/__w/Workshop-2024-Leipzig/Workshop-2024-Leipzig/tests/*.m2'.
2025-05-26T13:19:54.7222293Z  -- capturing check(0, "User")/usr/share/Macaulay2/Core/packages.m2:195:68:(1):[20]: error: key not found in hash table:
2025-05-26T13:19:54.7224192Z         "/usr/share/Macaulay2/Core/startup.m2" (of class String)
2025-05-26T13:19:54.7224958Z /usr/share/Macaulay2/Core/packages.m2:195:14:(1):[19]: --back trace--
2025-05-26T13:19:54.7225754Z /usr/share/Macaulay2/Core/methods.m2:154:98:(1):[18]: --back trace--
2025-05-26T13:19:54.7226558Z /usr/share/Macaulay2/Core/examples.m2:96:73:(1):[16]: --back trace--
2025-05-26T13:19:54.7232804Z /usr/share/Macaulay2/Core/methods.m2:154:98:(1):[15]: --back trace--
2025-05-26T13:19:54.7233350Z /usr/share/Macaulay2/Core/option.m2:17:14:(1):[14]: --back trace--
2025-05-26T13:19:54.7233784Z /usr/share/Macaulay2/Core/testing.m2:66:32:(1):[13]: --back trace--
2025-05-26T13:19:54.7234204Z /usr/share/Macaulay2/Core/testing.m2:121:48:(1):[12]: --back trace--
2025-05-26T13:19:54.7234515Z  -- .0718583s elapsed
2025-05-26T13:19:54.7234815Z /usr/share/Macaulay2/Core/testing.m2:107:101:(1):[12]: --back trace--
2025-05-26T13:19:54.7235265Z /usr/share/Macaulay2/Core/methods.m2:154:98:(1):[11]: --back trace--
2025-05-26T13:19:54.7235675Z /usr/share/Macaulay2/Core/option.m2:17:14:(1):[10]: --back trace--
2025-05-26T13:19:54.7236131Z /usr/share/Macaulay2/Core/testing.m2:101:37:(1):[9]: --back trace--
2025-05-26T13:19:54.7236534Z /usr/share/Macaulay2/Core/methods.m2:154:98:(1):[8]: --back trace--
[...]

PR #3483, opened 2024-09-18
Add examples(ZZ) to see examples for the doc nodes from "about"

d-torrance:
We also add some informative comments to the output.  For example:

```m2
i1 : about rank

o1 = {0 => Chordal :: rank(ChordalNetNode)                   }
     {1 => DiffAlg :: rank(DiffAlgDistribution)              }
     {2 => HyperplaneArrangements :: rank(CentralArrangement)}
     {3 => HyperplaneArrangements :: rank(Flat)              }
     {4 => InvariantRing :: rank(DiagonalAction)             }
     {5 => LieTypes :: rank(LieAlgebra)                      }
     {6 => Macaulay2Doc :: rank                              }
     {7 => Macaulay2Doc :: rank of a matrix                  }
     {8 => Macaulay2Doc :: rank(GradedModule)                }
     {9 => Macaulay2Doc :: rank(Matrix)                      }
     {10 => Macaulay2Doc :: rank(Module)                     }
     {11 => Macaulay2Doc :: rank(MutableMatrix)              }
     {12 => Matroids :: rank(Matroid)                        }
     {13 => Matroids :: rank(Matroid,List)                   }
     {14 => Matroids :: rank(Matroid,Set)                    }
     {15 => OldToricVectorBundles :: rank(ToricVectorBundle) }
     {16 => Posets :: rank(Poset)                            }
     {17 => Schubert2 :: rank(AbstractSheaf)                 }
     {18 => SparseResultants :: rank(MultidimensionalMatrix) }
     {19 => ToricVectorBundles :: rank(ToricVectorBundle)    }
     {20 => WeylGroups :: rank(DynkinDiagram)                }
     {21 => WeylGroups :: rank(RootSystem)                   }

o1 : NumberedVerticalList

i2 : examples 6

o2 = -- examples for tag: rank
     -- ../src/macaulay2/M2/M2/Macaulay2/packages/Macaulay2Doc/functions/rank-doc.m2:32:0
     R = ZZ/101[x,y,z]
     p = vars R;
     rank kernel p
     rank cokernel p
     C = res cokernel p
     rank C
```

PR #3852, opened 2025-05-26
needsPackage updates

d-torrance:
Two small updates:

* Have `packageFiles User` return the empty list, which fixes the current problem of `needsPackage User` raising a key error (startup.m2 isn't in the loaded files hash table since it's a string not a file)
* Add a warning message about recreating instances of types when reloading a package

Closes: #3851

### DIFF
--- a/M2/Macaulay2/m2/packages.m2
+++ b/M2/Macaulay2/m2/packages.m2
@@ -553,10 +553,15 @@ debug GlobalDictionary := dict -> (
     checkShadow())
 
 packageFiles = pkg -> (
-    pkgaux := (
+    srcfile := realpath pkg#"source file";
+    if (
+	srcfile == "stdio"         or
+	srcfile == "currentString" or
+	match("/startup\\.m2(?:\\.in)?$", srcfile))
+    then {}
+    else prepend(srcfile,
 	if not pkg#?"auxiliary files" then {}
-	else select(values loadedFiles, match_(pkg#"auxiliary files")));
-    prepend(realpath pkg#"source file", pkgaux))
+	else select(values loadedFiles, match_(pkg#"auxiliary files"))))
 
 locate Package := pkg -> NumberedVerticalList (
     -- TODO: somehow keep track of the number of lines of each file
--- a/M2/Macaulay2/m2/packages.m2
+++ b/M2/Macaulay2/m2/packages.m2
@@ -168,6 +168,8 @@ loadPackage String  := opts -> pkgname -> (
     if opts.Reload === true then (
 	dismiss pkgname;
 	if isPackageLoaded pkgname then (
+	    printerr("warning: reloading ", pkgname,
+		"; recreate instances of types from this package");
 	    closePackage value PackageDictionary#pkgname;
 	    -- clear out the value of the symbol
 	    PackageDictionary#pkgname <- PackageDictionary#pkgname));
--- a/M2/Macaulay2/tests/normal/testing.m2
+++ b/M2/Macaulay2/tests/normal/testing.m2
@@ -21,6 +21,9 @@ expectedCode = DIV{
 assert BinaryOperation(symbol ===, code pkgtest, expectedCode)
 assert BinaryOperation(symbol ===, code 0, expectedCode)
 
+TEST "assert Equation(1 + 1, 2)"
+check User
+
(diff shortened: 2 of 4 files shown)


M2/Macaulay2/m2/packages.m2
```diff
@@ -553,6 +555,7 @@ debug GlobalDictionary := dict -> (
     checkShadow())
 
 packageFiles = pkg -> (
+    if pkg#"pkgname" == "User" then return {};
```
mahrud: Is this the only package that is problematic?
How about a package defined directly in the interpreter?

d-torrance: Good point!

```m2
i2 : newPackage "Foo"

o2 = Foo

o2 : Package

i3 : endPackage "Foo"

o3 = Foo

o3 : Package

i4 : packageFiles oo

o4 = {/home/profzoom/src/macaulay2/M2/M2/Macaulay2/m2/stdio}

o4 : List
```

I'll play around with this some more

d-torrance: This should be addressed now

M2/Macaulay2/m2/packages.m2
```diff
@@ -553,10 +555,15 @@ debug GlobalDictionary := dict -> (
     checkShadow())
 
 packageFiles = pkg -> (
-    pkgaux := (
+    srcfile := realpath pkg#"source file";
+    if (
+	srcfile == "stdio"         or
+	srcfile == "currentString" or
+	match("/startup\\.m2(?:\\.in)?$", srcfile))
```
mahrud, line 562: why the extra `(?:\\.in)?$`?

M2/Macaulay2/tests/normal/testing.m2
```diff
@@ -21,6 +21,9 @@ expectedCode = DIV{
 assert BinaryOperation(symbol ===, code pkgtest, expectedCode)
 assert BinaryOperation(symbol ===, code 0, expectedCode)
 
+TEST "assert Equation(1 + 1, 2)"
+check User
```
mahrud, line 25: Check still calls `needsPackage`, right? does `needsPackage`/`loadPackage` do anything for User?

M2/Macaulay2/m2/packages.m2
d-torrance: The file location of `startupString` is startup.m2.in in the build directory by default.  I've patched the Debian package so that it points to the distributed copy of startup.m2 for reproducibility, so this catches both cases.

If you'd like, I could change this regex to just `/startup\\.m2\\.in$` and add a corresponding patch to the Debian package.

M2/Macaulay2/tests/normal/testing.m2
d-torrance: Good question!

It's already marked as loaded since #3483, so nothing unless we pass `Reload => true`.  Which currently raises an error.  Same with packages we create in the REPL:

```m2
i1 : loadPackage(User, Reload => true)
 -- warning: reloading User; recreate instances of types from this package
stdio:1:11:(3): error: file not found on path: "User.m2"

i2 : newPackage "Foo"; endPackage "Foo"

o3 = Foo

o3 : Package

i4 : loadPackage(oo, Reload => true)
 -- warning: reloading Foo; recreate instances of types from this package
stdio:3:11:(3): error: file not found on path: "Foo.m2"
```

I'll play around with this some more.

M2/Macaulay2/m2/packages.m2
mahrud: I'm saying either way "/startup.m2" should match, no?

d-torrance: Oh sure.  I guess I was trying to avoid the very minor edge case of an actual package file whose path contains the string "startup.m2" somewhere in the middle, say `/foo/bar/startup.m2/Foo.m2`.

M2/Macaulay2/tests/normal/testing.m2
d-torrance: Now that I think about this some more, maybe the current behavior is the correct one?  We can't reload a package if there's no file to reload it from...

mahrud (approved):

PR #3857, opened 2025-05-28
added mention of 'cite' in Macaulay2Doc

mahrud:
Adds a blurb about also citing other packages in Macaulay2Doc.


mahrud:
What it looks like:
image
Open to other options or phrasings.

d-torrance (approved):
Looks good!

PR #3858, opened 2025-05-28
Remove use of deprecated strstream header

jamesjer:
The `<strstream>` header was deprecated in C++98 and is slated for removal in C++26.  In any case, nothing in `<strstream>` is used in `f4.cpp`.


d-torrance (approved):
Thank you!

PR #3859, opened 2025-05-29
Replace egrep & fgrep with grep -E / -F

jamesjer:
The GNU grep manual says:
```
7th Edition Unix had commands egrep and fgrep that were the counterparts of the modern ‘grep -E’ and ‘grep -F’.
Although breaking up grep into three programs was perhaps useful on the small computers of the 1970s, egrep
and fgrep were deemed obsolescent by POSIX in 1992, removed from POSIX in 2001, deprecated by GNU Grep 2.5.3
in 2007, and changed to issue obsolescence warnings by GNU Grep 3.8 in 2022; eventually, they are planned to be
removed entirely.
```

This patch makes the suggested change to `grep -E` and `grep -F`, except for the example in `M2/Macaulay2/packages/Macaulay2Doc/ov_files.m2`, which I was afraid to touch.

### DIFF
--- a/M2/Macaulay2/bin/Makefile.in
+++ b/M2/Macaulay2/bin/Makefile.in
@@ -174,8 +174,8 @@ endif
 ifeq (@OS@ @SHARED@,Linux yes)
 ifneq (@BUILDLIST@,)
 ifneq (@OBJDUMP@,false)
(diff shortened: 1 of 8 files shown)


d-torrance (approved):
Thank you!

d-torrance:
@jamesjer - I took care of the other case in #3860.

PR #3862, opened 2025-06-01
Update SCMAlgebras.m2

ErnestoLax:
Added new methods in the package. It can now compute the unmixed layer of an ideal and check for unmixedness. Added a method to check for canonical Cohen-Macaulayness.


d-torrance (approved):
Thank you!

PR #3864, opened 2025-06-02
Check for file existence in packageFiles instead of matching strings

d-torrance:
This is much more robust, and also avoids an issue where `realpath` was stripping away the filename so that the "startup.m2" regex didn't match.

In particular, this was the big problem:

```m2
i1 : User#"source file"

o1 = /usr/local/share/Macaulay2/Core/startup.m2

i2 : realpath oo

o2 = /usr/local/share/Macaulay2
```

I'm not sure if this is the expected behavior of `realpath` or not, but that might be a bug too.

### DIFF
--- a/M2/Macaulay2/m2/last.m2
+++ b/M2/Macaulay2/m2/last.m2
@@ -37,6 +37,7 @@ addStartFunction(
 		   DebuggingMode  => true);
 	       User.PackageIsLoaded = true;
 	       User#"source directory" = "";
+	       User#"source file" = "stdio";
 	       path = prepend("./",path); -- now we search also the user's current directory, since our files have already been loaded
 	       path = unique apply( path, minimizeFilename);	    -- beautify
 	       allowLocalCreation User#"private dictionary";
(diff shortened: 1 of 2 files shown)


mahrud (approved):
This is good, but I'd rather fix `User#"source file"` to something more reasonable, like stdio.

mahrud:
This is still not good:
```m2
ii10 : debug Core

ii11 : packageFiles pkg

oo11 = {/home/mahrud/Projects/M2/M2/M2/Macaulay2/m2/startup.m2.in}

oo11 : List

ii12 : file

oo12 = /home/mahrud/Projects/M2/M2/M2/Macaulay2/m2/startup.m2.in

ii13 : filesLoaded#file
stdio:13:11:(3): error: key not found in hash table:
        "/home/mahrud/Projects/M2/M2/M2/Macaulay2. (of class String)

ii14 : filesLoaded#?file

oo14 = false
```

I think we need one or both of these changes:
- `needsPackage` should be nilpotent for User, Core, etc.
- source file for User needs to be stdio or currentString

mahrud (approved):

Issue #3867, opened 2025-06-03
building 1.25.06 on MacOS 15.5 on autotools is failing


mikestillman:
The build of `lrslib` can't find gmp.h.  This started happening at least a month or two ago.  @d-torrance Any suggestions for how to fix this?

```
cd build && gtar xf /Users/mike/src/M2-current-branches/M2-development/M2/BUILD/mike/builds.tmp/arm64-appleclang/../../../../BUILD/tarfiles/lrslib-073.tar.gz --gzip
cd build && for i in /Users/mike/src/M2-current-branches/M2-development/M2/BUILD/mike/builds.tmp/arm64-appleclang/../../../../libraries/lrslib/patch-073 ; do patch --batch -p0 < $i ; done
patching file 'lrslib-073/lrslib.c'
touch .patched-073
+ set +x
make: Entering directory '/Users/mike/src/M2-current-branches/M2-development/M2/BUILD/mike/builds.tmp/arm64-appleclang/libraries/lrslib/build/lrslib-073'
+ true
+ touch .configured-073
+ set +x
make: Entering directory '/Users/mike/src/M2-current-branches/M2-development/M2/BUILD/mike/builds.tmp/arm64-appleclang/libraries/lrslib/build/lrslib-073'
[...]

PR #3869, opened 2025-06-03
Update lrslib build flags (autotools)

d-torrance:
We were defining a bunch of flags that aren't used.  Also add a hacky solution to the fact that CPPFLAGS and LDFLAGS aren't honored.

This should fix builds on ARM macOS, where GMP is in /opt/homebrew/, which isn't searched by default (#3867).

I tested this using GitHub builds for macOS 13, 14, and 15:  https://github.com/d-torrance/M2/actions/runs/15425348026

Cc: @mikestillman


mikestillman (approved):
This works for me on MacOS!

Issue #3868, opened 2025-06-03
strict equality of complexes


mahrud:
I closed my earlier issue because the issue is in using `===` for complexes, not truncation itself:
```m2
i1 : R = QQ[x];

i2 : complex R === complex R

o2 = false
```
This is because Complex inherits from mutable hash table, and there isn't another `===` method installed.

So perhaps really the issue is using `===` or `=!=` for complexes, for instance in liftMapAlongQuasiIsomorphism:
https://github.com/Macaulay2/M2/blob/a816bf9a79201423e49991615b19cecfff9ed4a9/M2/Macaulay2/packages/Complexes/ChainComplexMap.m2#L1279

mahrud:
The particular issue with the truncation functions (#3865) is that they always construct a new complex, hence they will never be functorial as long as === behaves the way it currently does:
```m2
i16 : C = complex R

       1
o16 = R
       
      0

o16 : Complex

i17 : canonicalTruncation(C, (1,1)) === canonicalTruncation(C, (1,1))

o17 = false

```

n-m-g:
Yes, the use of "===" is something to be aware of.  If this is already causing bugs here then it might also be causing bugs say in the Complexes "Hom" and/or "TensorProduct" chain complexes implementation.  However, if one is willing to keep track of "prunning maps" then "==" can sometimes be used as a work around.  This might also explain why some examples are returning in error when trying to install the SpectralSequences package.  (See the PR https://github.com/Macaulay2/M2/pull/3807).

Do keep me posted.

mahrud:
I don't think these are related issues.

n-m-g:
Yes, probably not but do let me know if you encounter any additional bugs with the Complexes package.  None of us have yet looked into seriously with issues with why https://github.com/Macaulay2/M2/pull/3807 doesn't work with the latest M2 release (except looking into that code which uses "===" quite a bit may be a place to look at more seriously).

n-m-g:
As one follow-up, do you know if any of this "===" and/or the Hom/Tensor product code changed from 1.24 to 1.25?

mahrud:
I have looked into it, and it is not related to === or Hom or tensor. Let's keep this issue focused.

mahrud:
@mikestillman @ggsmith I understand the reason for making Complex inherit from mutable hash tables. I presume you don't want to change this, so should === and =!= basically never be used with them? e.g. should liftMapAlongQuasiIsomorphism use != instead, or something else?

[...]

PR #3874, opened 2025-06-05
Changed Complex strict equality internally

### LINKED ISSUE #3865: canonicalTruncation is not quite functorial
Here's an example borrowed from the documentation of [canonicalTruncation(ComplexMap,ZZ,ZZ)](https://macaulay2.com/doc/Macaulay2-1.25.05/share/doc/Macaulay2/Complexes/html/_canonical__Truncation_lp__Complex__Map_cm__Z__Z_cm__Z__Z_rp.html):
```m2
R = ZZ/101[a..d];
C = (freeResolution coker matrix{{a,b,c}})[1];
D = freeResolution coker matrix{{a*b,a*c,b*c}};
E = freeResolution coker matrix{{a^2,b^2,c*d}};
f = randomComplexMap(D, C);
```

I realized that canonicalTruncation does not commute with source/target:
```m2
A = source canonicalTruncation(f, (0, 1));
B = canonicalTruncation(source f, (0, 1));
A === B -- false
A == B  -- true

A = target canonicalTruncation(f, (0, 1));
B = canonicalTruncation(target f, (0, 1));
A === B -- false
A == B  -- true
```

This means, for instance, that things like `liftMapAlongQuasiIsomorphism` won't work as expected because the targets are not identical.

mahrud:
This changes how two complexes are compared to be identical internally, without incurring any computational cost. The point is that these basic things didn't work before: (I added these as a test)
```m2
R = QQ[x]
C = complex R
D = complex R
id_C |  id_D
id_C || id_D
id_C // id_D
```
But now complexes are compared like so:
```m2
isIdentical(Complex, Complex) := (C, D) -> C === D or C.module === D.module and C.dd.map === D.dd.map

isIdentical(ComplexMap, ComplexMap) := (f, g) -> f === g or (
    f.degree === g.degree and f.map === g.map
    and isIdentical(f.source, g.source)
    and isIdentical(f.target, g.target))
```
Fixes #3865 and #3868

### DIFF
--- a/M2/Macaulay2/packages/Complexes/ChainComplex.m2
+++ b/M2/Macaulay2/packages/Complexes/ChainComplex.m2
@@ -280,8 +280,12 @@ length Complex := (C) -> (
     hi-lo
     )
 
-Complex == Complex := (C,D) -> (
-    if C === D then return true;
+-- this method is not exported, and is only used internally instead of === and =!=,
+-- since Complex is a MutableHashTable and are never identical unless literally the same.
+isIdentical = method()
+isIdentical(Complex, Complex) := (C, D) -> C === D or C.module === D.module and C.dd.map === D.dd.map
+
+Complex == Complex := (C,D) -> isIdentical(C, D) or (
     (loC,hiC) := C.concentration;
     (loD,hiD) := D.concentration;
     if ring C =!= ring D then return false;
--- a/M2/Macaulay2/packages/Complexes/ChainComplexMap.m2
+++ b/M2/Macaulay2/packages/Complexes/ChainComplexMap.m2
@@ -110,7 +110,7 @@ map(Complex, Complex, ComplexMap) := ComplexMap => opts -> (tar, src, f) -> (
     )
 
 ComplexMap | ComplexMap := ComplexMap => (f,g) -> (
-    if target f =!= target g then error "expected targets to be the same";
+    if not isIdentical(target f, target g) then error "expected targets to be the same";
     deg := degree f;
     if deg =!= degree g then error "expected maps with the same degree";
     result := map(target f, source f ++ source g, {{f,g}}, Degree=>deg);
@@ -121,7 +121,7 @@ ComplexMap | ComplexMap := ComplexMap => (f,g) -> (
     )
 
 ComplexMap || ComplexMap := ComplexMap => (f,g) -> (
-    if source f =!= source g then error "expected sources to be the same";
+    if not isIdentical(source f, source g) then error "expected sources to be the same";
     deg := degree f;
     if deg =!= degree g then error "expected maps with the same degree";
     result := map(target f ++ target g, source f, {{f},{g}}, Degree=>deg);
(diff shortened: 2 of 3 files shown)


ggsmith:
This looks like a good idea to me.  However, should the method name be ```areIdentical```?

mahrud:
Since it's an internal method I don't mind either way, but for the record I was going with matching `isIsomorphic`. Would you like me to change it?

ggsmith:
Having investigated a bit further, I am still concerned that ```isIdentical``` is still too slow.  In particular, it appears to take essentially the same amount of time as ```==```.

mahrud:
What was the example you tried it on?

ggsmith:
@mikestillman and I create some relatively large Koszul complexes to compare ```isIdentical``` and ```==```.  We also looked at the source code for ```===``` focusing on its behaviour for matrices and modules.

mahrud:
Ah, I see. Let me explain.

What you've observed about the difference between isIdentical and === is simply the fact that === for a HashTable (which like isIdentical checks === on all content) is slower than === for a MutableHashTable (which checks two fixed hash numbers)! In other words, even if you changed Complex into a HashTable, the timing would exactly match isIdentical, no better and no worse!

The reason in your examples == and isIdentical took the same amount is that internally == for modules and matrices first checks === before doing anything more complicated, and if you have large identical koszul complexes that === check succeeds immediately, which is exactly what isIdentical does, so they take the same amount. It's just that the matrices are large so checking entry by entry is taking a while.

In fact, in all cases simply constructing the koszul complex took an order of magnitude longer than isIdentical, so I'm not sure if the cost of using isIdentical is relevant at all. In other words, isIdentical is not going to be the bottleneck in a computation if constructing the objects in the first place takes an order of magnitude longer.

Finally, to respond to your initial concern, == and isIdentical are absolutely not the same when the modules are equal but not identical, which is the case you presumably want to avoid in liftMapAlongQuasiIsomorphism and other places. Here is an example:
```m2
debug needsPackage "FGLM"
I = cyclic(5)
C = complex module I
D = complex module ideal random I_*
elapsedTime isIdentical(C, D) -- false, .00001999s  elapsed
elapsedTime C === D           -- false, .000003078s elapsed
elapsedTime C == D            -- true,  .478898s    elapsed
```
If you raise that 5 to a 6 or 7, == should take up to half an hour!

mahrud:
To recap the conversation during the meeting, the example above shows that this statement about `isIdentical` is incorrect:

> In particular, it appears to take essentially the same amount of time as `==`.

I spent quite a bit of time with `liftMapAlongQuasiIsomorphism` and I could not produce an example where switching from `===` to `isIdentical` causes a noticeable slowdown. I would like to see an example if you have one in mind.

mikestillman:
@ggsmith Should we accept this change?  We can back off from it if we find it really slows down some examples.

mahrud:
@mikestillman is this going to merge or should I close it?

PR #3876, opened 2025-06-07
Bump C++ standard to C++17 (autotools)

d-torrance:
This is necessary if we want to use googletest 1.17.0

### DIFF
--- a/M2/configure.ac
+++ b/M2/configure.ac
@@ -262,7 +262,7 @@ AC_SUBST(M2_CFLAGS)
 AC_SUBST(M2_CXXFLAGS)
 
 dnl Fix standards, apply everywhere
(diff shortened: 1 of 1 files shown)


mikestillman:
I definitely want to be on c++ version at least c++17...  So if the build is ok, definitely let's do it.  In fact, we should be trying c++20 too, but that will be another PR.  Question: which versions of ubuntu support c++17, c++20?

mikestillman (approved):
Assuming the builds finish with no errors, upgrading to c++17 is a good idea!  Thanks!

d-torrance:
According to https://gcc.gnu.org/projects/cxx-status.html:

* C++17 has been almost completely supported since GCC 7, which is available in Ubuntu 18.04
* C++20 has a been almost completely supported since GCC 11, which is available in Ubuntu 22.04

PR #3871, opened 2025-06-04
Bump normaliz to 3.10.5

d-torrance:
This was just released.  We also make `--enable-download` the default behavior in the autotools build.

Draft for now to test the GitHub builds.

PR #3880, opened 2025-06-10
Add --enable-download option to check-configure

d-torrance:
During configure, we raise an error if the input for --enable-download is "yes" but neither wget nor curl are available on the system.

Now that the default for --enable-download is "yes", we should make sure that we don't raise this error during check-configure if the user initially ran configure with "no", e.g., in Debian/Ubuntu package builds.

### DIFF
--- a/M2/check-configure/Makefile.in
+++ b/M2/check-configure/Makefile.in
@@ -29,6 +29,7 @@ check-config:
 	    --with-system-memtailor=$(SYSTEM_MEMTAILOR)				\
 	    --with-system-mathic=$(SYSTEM_MATHIC)				\
 	    --with-system-mathicgb=$(SYSTEM_MATHICGB)				\
+	    --enable-download=@DOWNLOAD@					\
 	    --disable-building
 distclean:: clean; rm -f Makefile
 Makefile: Makefile.in ; cd .. && ./config.status check-configure/Makefile


MichaelABurr:
Can you provide more clarification here?  What sequence of events and situations would cause this error to be raised in the old case?  Why shouldn't an error be raised?

d-torrance:
Up until recently, users had to pass `--enable-download` to `configure` if they wanted M2 to download and build dependencies that weren't already detected on their system. I thought this was annoying, so I changed it to be the default in https://github.com/Macaulay2/M2/pull/3871 (in `development` but not in the 1.25.06 release).

If someone runs the configure script with `--enable-download`, it checks to see if `wget` and/or `curl` are installed, otherwise an error is raised:

```
checking for wget... no
checking for curl... no
configure: error: wget or curl is required
```

So if you don't want to download anything (e.g., I do this in the Debian package where we're not allowed network access when building packages and all the dependencies are already available as system packages so there's no reason to download anything anyway), you can pass `--disable-download` and skip this check.

But currently, we don't save the value of this option for `make check-configure`.  So someone without `wget` or `curl` on their system who ran `configure --disable-download` then runs into the following:

```
$  make -C check-configure check
make: Entering directory '/root/M2/M2/BUILD/build/check-configure'
rm -rf tmp
mkdir tmp
make: [check-config] remark: the test below is important (only) for testing whether the configure script detects pre-installed libraries properly
.
.
.
checking for wget... no
checking for curl... no
configure: error: wget or curl is required
make: *** [Makefile:15: check-config] Error 1
make: Leaving directory '/root/M2/M2/BUILD/build/check-configure'
```

So all this PR does is pass the appropriate value to `--enable-download`.

d-torrance:
Discussed with Mike

PR #3883, opened 2025-06-12
Bump FLINT to 3.3.0

d-torrance:
This was just released.

Draft for now to test the GitHub builds.

### DIFF
--- a/M2/cmake/build-libraries.cmake
+++ b/M2/cmake/build-libraries.cmake
@@ -383,7 +383,7 @@ ExternalProject_Add(build-factory
   SOURCE_DIR        libraries/factory/build
   DOWNLOAD_DIR      ${CMAKE_SOURCE_DIR}/BUILD/tarfiles
   BUILD_IN_SOURCE   ON
-#  PATCH_COMMAND     patch --batch < ${CMAKE_SOURCE_DIR}/libraries/factory/patch-4.4.0
+  PATCH_COMMAND     patch --batch -p1 < ${CMAKE_SOURCE_DIR}/libraries/factory/patch-4.4.1
   CONFIGURE_COMMAND autoreconf -vif &&
                     ${CONFIGURE} --prefix=${M2_HOST_PREFIX}
                       #-C --cache-file=${CONFIGURE_CACHE}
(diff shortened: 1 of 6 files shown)


mahrud:
Seems alarming that only one build failed because of factory not being compatible with the upgraded flint.

d-torrance:
I think the other builds were just using the existing factory package.  The cmake build has a minimum version (which apparently I set back in #3286) that's higher than the Ubuntu package.

d-torrance:
@mahrud:  It looks like forcing factory to be built when flint is built fixes some of the issues.  What's the best way to do this in the cmake build?

@mikestillman:  There are a ton of EngineTests segfaults when linking against the new flint.  It looks like it's all related to LU decomposition.  You worked with this code recently, right?  Any ideas?  Here's one example:

```m2
i1 : -- test source: ../../../../Macaulay2/packages/EngineTests/LinearAlgebra.Test.FLINT.m2:139:5-144:3
     
       for i from 2 to 13 do (
           R = GF(2,i,Strategy=>"Flint");
           time testLinearAlgebra R
           )
-- SIGSEGV
-* stack trace, pid: 226359
 0# profiler_stacktrace(std::ostream&, int) at ../../../../Macaulay2/d/main.cpp:131
 1# segv_handler at ../../../../Macaulay2/d/main.cpp:243
 2# 0x00007F690BA45330 in /lib/x86_64-linux-gnu/libc.so.6
 3# _perm_parity at /home/runner/work/M2/M2/M2/BUILD/build/libraries/flint/build/flint-3.3.0/src/perm/parity.c:45
 4# MutableMat<DMat<M2::ARingGFFlint> >::LU(MutableMatrix*, MutableMatrix*) const at ../../../../Macaulay2/e/mutablemat-imp.hpp:197
 5# rawLU at ../../../../Macaulay2/e/interface/mutable-matrix.cpp:606
 6# interface_rawLU at ../../../../Macaulay2/d/interface.dd:3907
```

d-torrance:
I think I figured out the EngineTests segfaults. Flint no longer initializes the permutation vector when the matrix has zero columns, so now we do it manually.

d-torrance:
I reported the issue upstream and instead added their patch instead of dealing with it ourselves.

d-torrance:
The patch worked!

https://github.com/Macaulay2/M2/actions/runs/15655873707

mahrud (approved):
Thanks for dealing with this! Hopefully this makes patching the brew build easier for me.

PR #3888, opened 2025-06-14
Move gftables installation from datadir -> libdir (autotools)

d-torrance:
Previously, when we installed gftables instead of using an existing copy provided by a system factory package, we installed them in the architecture-independent directories, e.g., something like /usr/share/Macaulay2/Core/factory/gftables.

This causes problems for building rpm packages.  In particular, on some systems, like Fedora, a factory package is available, but on others, like RHEL, one is not.

The contents of the architecture-independent "share" directory are copied into the Macaulay2-common rpm package, which ideally should be usable by any rpm-compatible distribution.  And so in particular, we shouldn't install the gftables there, since we need them there on some systems but not on others.

Instead, we move installation to the architecture-dependendent "lib" directory, so they will be installed in the specific Macaulay2 rpm package for the target system only when needed.

### DIFF
--- a/M2/configure.ac
+++ b/M2/configure.ac
@@ -1550,7 +1550,7 @@ AC_ARG_VAR([GFTABLESDIR],
     [path to gftables directory if factory is already installed])
 if test $BUILD_factory = yes
 then
-    GFTABLESDIR="${datadir}/Macaulay2/Core/factory/"
+    GFTABLESDIR="${libdir}/factory/"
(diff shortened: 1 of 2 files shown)


d-torrance:
Wait, this won't fix the problem...  The path to the gftables is hard-coded in startup.m2.  I think the best bet is just to always build factory when preparing rpm packages for distribution.  Closing

PR #3899, opened 2025-06-30
Add support for new flint rand functions

d-torrance:
In a recent version of flint, the flint_randinit and flint_randclear functions were renamed to flint_rand_init and flint_rand_clear.  The older versions are still supported, but result in deprecation warnings and may be removed at some point.

We add FLINT_RAND_INIT and FLINT_RAND_CLEAR macros to cover both older and newer versions of flint.

TODO: cmake support (Cc: @antonleykin)


antonleykin:
The cmake solution seems to be the following:
```
check_function_exists(flint_rand_init HAVE_FLINT_RAND_INIT)

# If found, define a preprocessor macro for your target or globally
if(HAVE_FLINT_RAND_INIT)
  add_definitions(-DHAVE_FLINT_RAND_INIT)
endif()
```

mahrud:
> The cmake solution seems to be the following:

Not exactly. See how `HAVE_FLINT_NMOD_H` is set in `cmake/check-libraries.cmake` and used in `include/M2/config.h.cmake`.

mahrud (approved):
cmake parts look good.

PR #3900, opened 2025-06-30
Added a small clarifying remark about Krull dimension of a module

tbrazel:


M2/Macaulay2/packages/Macaulay2Doc/functions/dim-doc.m2
```diff
@@ -47,7 +47,7 @@ document {
 	  },
      Outputs => {ZZ
 	  },
-     "Computes the Krull dimension of the module ", TT "M",
+     "Computes the Krull dimension of the module ", TT "M", ". Recall that the Krull dimension of a module is by definition the ", TO2((dim,Ring), "dimension of the base ring"), " minus the ", TO2((codim,Module), "codimension"), " of the module.",
```
mahrud, line 50: But the documentation of `codim(Module)` says:
> Computes the codimension of the support of the module as given by dim(R) - dim(M).

Neither of these seem to match what the implementation of `codim(Module)` actually does though.

Tangential, but we really need a function `support(Module)` as well.

tbrazel: I think I'm confused on your comment? As I understand it, the implemented definition of dimension of a module is in  `hilbert.m2`:
https://github.com/Macaulay2/M2/blob/6fe351dcfe733f2a525d361d8585f9f2375e264d/M2/Macaulay2/m2/hilbert.m2#L109

So I was just adding a remark reminding the reader what dimension of a module is.

mahrud: I agree, this is a fine change. I'm just pointing out that it's a circular definition since the documentation doesn't say how the codimension of the module is computed.

tbrazel: Ah ok let me bug you about getting local build stuff set up this week then I'll add some codim documentation to this PR 🙂

mahrud:
Oh why did you close this?

PR #3903, opened 2025-07-01
promote cleanup

### LINKED ISSUE #3906: Constant times matrix fail
I thought we sorted out all these issues, but I guess not:
```
i1 : ii*matrix{{1}}
stdio:1:2:(3): error: key not found in hash table:
        RawRing (of class Symbol)
```

### LINKED ISSUE #3910: incorrect promotes
The principle of `promote` is that `promote(R,S)` exists if there is a natural map (typically, an inclusion) from `R` to `S`. In particular, the graph of `promote` has to be acyclic -- e.g., if both `promote(R,S)` and `promote(S,R)` we're in trouble because what are we supposed to do with `r+s` with `r` in `R` and `s` in `S`?
Unfortunately there are several current promotes that fail these basic premises. I intend to fix this in #3903.

There is one particularly problematic case, and since there was a debate on zulip, I want to summarise it here.

Currently `promote(RR,QQ)` exists. Now obviously there's no natural map from `RR` to `QQ` so it shouldn't.

Now the argument is that secretly `RR` is not really the real numbers but rather `ZZ[1/2]` and so it sits inside `QQ`.
First I strongly disagree with this argument, for many reasons:
- The philosophy of Macaulay2 is that objects should be as close to their true mathematical counterparts, irrespective of implementation.
- The argument could be made that `RR_*` is `ZZ[1/2]`, and similarly for say `RR_53`, but these are different.
- Even so, this is problematic, because `RR_*` is supposed to be a field, and `ZZ[1/2]` isn't.

Anyway, to be the devil's advocate, suppose `RR` is actually `ZZ[1/2]`. Then there is no map from `QQ` to `RR`, which means we should remove `promote(QQ,RR)`. Is that reasonable? obviously not.

So the only logical conclusion is that one should remove `promote(RR,QQ)`. It can be renamed to something different if people think it's useful.

I've only found one use of `promote(RR,QQ)`, which is in `Nauty` (well, two, since `NautyGraphs` is a fork of `Nauty` -- I'm a little confused why we have two packages, one a fork of the other, in the official version of M2. anyway), where a probability is first promoted to `QQ` before being rounded -- the former can be removed with no visible effect.

pzinn:
clean up promote code:
- remove `try` in promote
- remove redundant code
- fix #3906
- fix #3910
- minor improvement to `promote` that used to be in another PR of mine moved here, resolves the last issue in #3184


M2/Macaulay2/m2/reals.m2
```diff
@@ -106,10 +106,13 @@ promote(RawRingElement,RingElement) := (x,R) -> new R from x
 promote(Number,InexactNumber) := (x,RR) -> promote(x,default RR)
 promote(ZZ,RR') := 
 promote(QQ,RR') := 
+promote(RR',RR') :=
```
mahrud: A bit confused:
1. I think there's a better way to have promote behave like identity if the input ring and new ring are identical.
2. The commit that changes this file says
> fix isPromotable ring family
> This reverts commit https://github.com/Macaulay2/M2/commit/63d7fc73d45757ad1f46b63845af1d377574f934.

But that commit doesn't seem to have anything to do with this file.

M2/Macaulay2/m2/classes.m2
```diff
@@ -22,7 +22,7 @@ pluralize = s -> demark_" " append(
     drop(ws := separate_" " s, -1),
     if  plurals#?(last ws)
     then plurals#(last ws) else last ws | "s")
-pluralsynonym = T -> if T.?synonym then pluralize T.synonym else "objects of class " | toString T
+pluralsynonym = T -> pluralize lookup(symbol synonym,T) -- pluralize T.synonym ?? "objects of class " | toString T
```
mahrud: 1. Were the changes to `synonym` and `pluralsynonym` included by accident?
2. What happens if a type doesn't have a synonym? I prefer `object of class X` to just null.
3. Assuming this change is desired, `pluralsynonym = T -> pluralize synonym T` seems simpler.

M2/Macaulay2/m2/matrix.m2
```diff
@@ -3,6 +3,22 @@
...
 needs "modules.m2"
 
+notsamering := (X,Y) -> (
+     if X === Y then error("expected ",pluralsynonym X, " for the same ring")
+     else error("expected ",synonym X," and ",synonym Y," for the same ring"))
+samering = (M,N) -> if ring M === ring N then (M,N) else notsamering(class M,class N)
+nottosamering := (X,Y) -> (
+     if X === Y then error("expected ",pluralsynonym X, " for compatible rings")
+     else error("expected ",synonym X," and ",synonym Y," for compatible rings"))
+tosamering = (M,N) -> (
```
mahrud: I don't mind keeping the old names if you don't want to change names in a bunch of places, but I don't think we should actively rename something camel-cased to all lower case. Also, maybe move these functions to where `isPromotable` is defined instead.

M2/Macaulay2/m2/classes.m2
pzinn: just to clarify, this PR is very much work in progress. This will be reverted, it was just an attempt.

M2/Macaulay2/m2/reals.m2
pzinn: the revert part of the message is an accident. will fix.
the addition of `promote(XX',YY')` is an unfortunate side-effect of the messy way that inexact fields are implemented in Macaulay2 (which one day I'd love to rewrite from scratch, because it's such a mess).
They're not actual functions though because they never actually get called, really one should make them dummy functions.

M2/Macaulay2/m2/matrix.m2
pzinn: will do. not the most urgent thing right now.
once again, this is a *draft*.

M2/Macaulay2/m2/classes.m2
mahrud: Ah, sorry I missed that this was a draft.

d-torrance:
Would you fix the conflict in matrix.m2?

M2/Macaulay2/m2/reals.m2
```diff
@@ -157,6 +159,10 @@ QQ _ ComplexField :=
 RR _ ComplexField :=
 CC _ ComplexField := (x,R) -> toCC(R.precision,x)
 
+lift1 := z -> if z === 0. then 0/1 else if isFinite z then (
```
d-torrance, line 162: Can we make this `=` so we can access this using `debug Core`?

d-torrance: And maybe give it a more descriptive name?  `internalRepresentation` or something?

M2/Macaulay2/packages/Macaulay2Doc/changes.m2
```diff
@@ -2891,11 +2891,6 @@ document {
 			 "The quotient and remainder for two ring elements can now be obtained simultaneously,
 			 saving time.  See ", TO (quotientRemainder,RingElement,RingElement), "."
 			 },
-		    LI {
-			 "The binary representation of a real number is now available using ", TO (promote,RR,QQ), ".
-			 The code for ", TO (lift,RR,QQ), " has been tightened up so a rational number is provided
-			 that provides exactly the same real number when promoted."
-			 },
```
d-torrance, line 2898: Can we keep this for historical purposes?  Maybe change the `TO`'s to `TT`'s or something so there aren't broken links.  And possibly add a note that this feature was removed in a later version

d-torrance (changes requested):
Discussed with Mike.  Looks very nice -- thanks!  Just a couple comments

pzinn:
changes implemented as suggested

d-torrance (approved):
Thanks!

Issue #3848, opened 2025-05-24
"try" should suppress error processing


mahrud:
Here's a bizarre observation: I tried to profile @pzinn's examples [here](https://github.com/Macaulay2/M2/issues/3843#issuecomment-2906328204), and got this:
```m2
o5 = #run  %time   position
     5     233.8   ../linuxbrew/.linuxbrew/share/Macaulay2/LieTypes.m2:531:8-533:27
     4     233.8   ../linuxbrew/.linuxbrew/share/Macaulay2/LieTypes.m2:533:13-533:27
     3103  95.37   ../linuxbrew/.linuxbrew/Cellar/macaulay2/1.25.05/share/Macaulay2/Core/lists.m2:164:5-168:12
     3146  72.18   ../linuxbrew/.linuxbrew/Cellar/macaulay2/1.25.05/share/Macaulay2/Core/matrix.m2:179:10-205:25
     4     64.35   ../linuxbrew/.linuxbrew/share/Macaulay2/LieTypes.m2:1190:4-1190:49
     4     63.63   ../linuxbrew/.linuxbrew/share/Macaulay2/LieTypes.m2:1058:4-1060:29
     44    63.55   ../linuxbrew/.linuxbrew/Cellar/macaulay2/1.25.05/share/Macaulay2/Core/lists.m2:190:5-194:12
     44    60.8    ../linuxbrew/.linuxbrew/Cellar/macaulay2/1.25.05/share/Macaulay2/Core/lists.m2:193:10-193:46
     3146  60.19   ../linuxbrew/.linuxbrew/Cellar/macaulay2/1.25.05/share/Macaulay2/Core/matrix.m2:182:15-0:0
     3103  56.61   ../linuxbrew/.linuxbrew/Cellar/macaulay2/1.25.05/share/Macaulay2/Core/lists.m2:167:10-167:46
     4720  56.57   ../linuxbrew/.linuxbrew/Cellar/macaulay2/1.25.05/share/Macaulay2/Core/modules.m2:41:5-41:21
     1573  41.14   ../linuxbrew/.linuxbrew/Cellar/macaulay2/1.25.05/share/Macaulay2/Core/methods.m2:27:65-28:88
     3103  37.99   ../linuxbrew/.linuxbrew/Cellar/macaulay2/1.25.05/share/Macaulay2/Core/lists.m2:166:10-166:20
[...]

PR #3912, opened 2025-07-04
Revert "lastError" for now

d-torrance:
This reverts #3654, which appears to be responsible for a considerable slowdown.

### DIFF
--- a/M2/Macaulay2/d/actors5.d
+++ b/M2/Macaulay2/d/actors5.d
@@ -1814,9 +1814,6 @@ export StandardE := Expr(StandardS);
 export topLevelMode := Expr(StandardS);
 topLevelModeS := dummySymbol;
 
-threadLocal lastError := nullE;
-lastErrorS := dummySymbol;
-
 initialRandomSeed := zeroZZ;
 initialRandomHeight := toInteger(10);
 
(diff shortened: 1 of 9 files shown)


d-torrance:
@mikestillman - I think this fixed the slowdown.  The cmake-macos build finished in 4.5 hours

mahrud:
Did you revert specific commits, or did you manually undo some changes?

d-torrance:
Specific commits (in particular, the final 3 commits of the original PR that actually dealt with lastError)

pzinn:
I don't think this is a good idea. Instead one should focus on the source of the slow down.
It's most likely due to `try` being overused, cf the wise advice in the documentation:
```
users are recommended to use this function sparingly, if at all.
```
For example, I'm removing `try` in `promote` in #3903. Unfortunately this is more complicated than I thought due to issue #3910.

d-torrance:
Good point.

I did a little experimenting and I think the big slowdown is the `try` in `package(Symbol)`.  I'll push an alternate PR and see if refactoring that helps

d-torrance:
It's more than just `package(Symbol)`...  There are quite a few things that use `try` and `??=` that are slowing things down.

mahrud:
1. I'm not opposed to solutions that don't use try, for instance in promote, but try is a useful and sometimes irreplaceable feature, and that recommendation you cite is from almost 28 years ago (647f7532b), so I wouldn't call try "overused".
2. lastError isn't being used anywhere yet, so reverting this PR until it can be correctly implemented is reasonable. In particular, as I pointed out in my last comment in https://github.com/Macaulay2/M2/issues/3848#issuecomment-3039399776, preparing the error message should not happen until we actually want to print it. I gave an example where producing the error computes a Groebner basis, which definitely should not happen inside `try` if the error message won't even be printed (a basic principle in functional programming).

mahrud (approved):

mahrud:
For future reference, I think this feature could be re-implemented as a _function_ `printLastError()` which only upon being called processes the error, along with a variable `lastErrorLocation` which produces the location of the error.

pzinn:
I thought `??=` didn't cause updating `lastError`? if so it's a valid alternative in many circumstances to `try`.

d-torrance:
It does.  `try` and `??` both use `tryEval` under the hood, which evaluates the code and suppresses errors.

PR #3918, opened 2025-07-06
Removed example with long runtime

Martin-Helmer:
Removed the example in whitneyStratifyPol which was causing longer run times with the new msolve.


d-torrance (approved):
Thank you!

PR #3807, opened 2025-05-13
Update SpectralSequences.m2

n-m-g:
A few additional "legacy" ChainComplex items removed.

PR #3921, opened 2025-07-08
SpectralSequences v1.1: switched to Complexes

mahrud:
This PR updates the SpectralSequences package to use Complexes. I noticed that #3807 and previous attempts had several bugs (e.g. conversion from truncate to naiveTruncation and the indices in the y-filtration were wrong), so I started from scratch.

- **updated helpers in SpectralSequences**
  - this commit also deletes various ChainComplex methods that were implemented in SpectralSequences but are now implemented by Complexes
  - it also renames `truncate(Complex,ZZ)` to `naiveTruncation(Complex,ZZ)`.
- **updated methods in SpectralSequences**
- **fixed the filteredComplex(List) constructor**
- **fixed the filteredComplex(Complex) constructor**
- **updated docs in SpectralSequences**
- **updated tests in SpectralSequences**
- **fixed an indexing bug in SpectralSequences**
  - `min/max` behave differently for Complex and ChainComplex and are used in three different places which could cause inconsistencies if there were hidden assumptions in the past. This is worth a review.

Future things to do:
- split the file in an auxiliary directory
- trim the documentation; several nodes are 95% duplicates, and some nodes are so huge that the whole documentation is about 12MB!

### DIFF
--- a/M2/Macaulay2/packages/SpectralSequences.m2
+++ b/M2/Macaulay2/packages/SpectralSequences.m2
@@ -1435,6 +1435,173 @@ doc ///
 	       appears on the $E_1$ cancels in two steps via an $E_2$ map with $k^6$ and via an $E_3$ map with a $k^2$.
 ///
 
+
+doc ///
+    Key
+      "Identifying anti-podal points of the two sphere"
+    Description
+    	  Text
+	      In this example we compute the spectral sequence arising from
+	      the quotient map
+	      $\mathbb{S}^2 \rightarrow \mathbb{R} \mathbb{P}^2$, 
+	      given by identifying anti-podal points. 
+	      This map can be realized by a simplicial map along the lines of Exercise 27, Section 6.5 of Armstrong's
+	      book {\it Basic Topology}.
+	      In order to give a combinatorial picture of the quotient map
+	      $\mathbb{S}^2 \rightarrow \mathbb{R} \mathbb{P}^2$, 
+	      given by identifying anti-podal points, we
+ 	      first make an appropriate simplicial realization of $\mathbb{S}^2$.
+	      Note that we have added a few barycentric coordinates.
+     	  Example
+	      S = ZZ[v1,v2,v3,v4,v5,v6,v15,v12,v36,v34,v46,v25];
+	      twoSphere = simplicialComplex {v3*v4*v5, v5*v4*v15, v15*v34*v4, v15*v34*v1, v34*v1*v6, v34*v46*v6, v36*v46*v6, v3*v4*v46, v4*v46*v34, v3*v46*v36, v1*v6*v2, v6*v2*v36, v2*v36*v12,v36*v12*v3, v12*v3*v5, v12*v5*v25, v25*v5*v15, v2*v12*v25, v1*v2*v25, v1*v25*v15};	   
+	  Text
+	     We can check that the homology of the simplicial complex twoSphere agrees with that of $\mathbb{S}^2$.
+	  Example
+	      C = naiveTruncation(complex twoSphere,1,infinity)	
+	      prune HH C
+	  Text
+	      We now write down our simplicial complex whose topological realization 
+	      is $\mathbb{R} \mathbb{P}^2$.
+	  Example     
+	      R = ZZ[a,b,c,d,e,f];
+	      realProjectivePlane = simplicialComplex {a*b*c, b*c*d, c*d*e, a*e*d, e*b*a, e*f*b, d*f*b, a*f*d, c*f*e,a*f*c};
+	  Text 
+	      Again we can check that we've entered a simplicial complex
+       	      whose homology agrees with that of the real projective plane.
+	  Example
+	      B = naiveTruncation(complex realProjectivePlane, 1,infinity)	 
+	      prune HH B
+    	  Text
+	      We now compute the fibers of the anti-podal quotient map
+ 	      $\mathbb{S}^2 \rightarrow  \mathbb{R} \mathbb{P}^2$.
+	      The way this works for example is:
+	      $a = v3 ~ v1, b = v6 ~ v5, d = v36 ~ v15, c = v4 ~ v2, 
+	      e = v34 ~ v12, f = v46 ~ v25$
+
+              The fibers over the vertices of $\mathbb{R} \mathbb{P}^2$ are:
+	 Example     
+	      F0twoSphere = simplicialComplex {v1,v3,v5,v6, v4,v2, v36,v15, v34,v12, v46,v25}
+    	 Text
+	      The fibers over the edges of $\mathbb{R}\mathbb{P}^2$ are: 
+   	 Example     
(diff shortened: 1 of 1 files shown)


mikestillman:
@n-m-g This modified version of your pull request #3807, which @mahrud contributed, has all examples and tests passing, and uses Complexes.  Can you check over it to make sure you are OK with the changes.  If you are, I will accept the pull request.  It looks pretty good to me. (Thanks @mahrud!)

mikestillman:
Actually, @mahrud, it has one failing build, involving tvalues.m2?  This is probably unrelated...

d-torrance:
I think that was just due to a soversion discrepancy between some brew bottles after an NTL release back when the tests were run.

M2/Macaulay2/packages/SpectralSequences.m2
```diff
@@ -109,29 +109,27 @@ ReverseDictionary = value Core#"private dictionary"#"ReverseDictionary"
 -------------------------------------------------------------------------------------------------
 --  The writing of some methods are much improved --
 
-
-
 -- since things are mutable we don't want to cache spots
 spots = method()
+spots Complex := List => C -> sort keys C.module
```
d-torrance, line 114: Comment from Mike: `keys C.module` isn't part of preferred interface, so we should keep the original implementation

M2/Macaulay2/packages/SpectralSequences.m2
```diff
@@ -109,29 +109,27 @@ ReverseDictionary = value Core#"private dictionary"#"ReverseDictionary"
...
-    )
-
-
- 
-----------------------------------------------------------------------------------
+pushFwd(RingMap, Complex) := o -> (f, C) -> (
+    (lo, hi) := concentration C;
+    if dd^C == 0
+    then complex(for i from lo to hi list pushFwd(f, C_i, o), Base => lo)
+    else complex applyValues(C.dd.map, m -> pushFwd(f, m, o)))
```
d-torrance, line 125: This should be in *Complexes*.  And could also be simplified

M2/Macaulay2/packages/SpectralSequences.m2
```diff
@@ -182,38 +182,41 @@ filteredComplex = method(Options => {
...
+	else error "expected all maps to have the same target")
+    else error "expected a list of complex maps or simplicial complexes";
+    --
+    Z := image map(C, C, i -> 0*id_(C_i)); -- make zero subcomplex as a subcomplex of ambient complex
+    P := {};
+    myList := {};
+    for p from 0 to #maps - 1 do (
+	myList = myList |
+	{#maps - (p+1) -opts.Shift => image maps#p};
+	);
```
d-torrance, line 215: This should be an `apply` or ` for ... list`

M2/Macaulay2/packages/SpectralSequences.m2
```diff
@@ -603,7 +603,7 @@ SpectralSequence ^ InfiniteNumber:=
 	-- again trying to handle the case of the zero complex --
     if min K_(infinity) < infinity and max K_infinity > - infinity then (
 	    for p from min K to max K do (
-	  	for q from - p + min K_(infinity) to max K_(infinity) do (
+		for q from -p + min K_(infinity) to max K_(infinity) + 1 do (
```
d-torrance, line 606: What bug is this fixing?

M2/Macaulay2/packages/SpectralSequences.m2
```diff
@@ -1435,6 +1435,173 @@ doc ///
 	       appears on the $E_1$ cancels in two steps via an $E_2$ map with $k^6$ and via an $E_3$ map with a $k^2$.
 ///
 
+
+doc ///
```
d-torrance, line 1439: Why are we changing the order back?

d-torrance (changes requested):
Discussed with Mike.  We should also discuss this with the package authors

M2/Macaulay2/packages/SpectralSequences.m2
mahrud: This is the original order where my changes were made, and I didn't want to spend extra time reordering them. I do think it makes more sense (e.g. Example 1, Example 2, Example 3 are next to each other), wonder why they were reordered recently.

M2/Macaulay2/packages/SpectralSequences.m2
mahrud: I would try removing the +1 and seeing if the documentation changes maybe but I'm not sure without spending more time to dig into it.

M2/Macaulay2/packages/SpectralSequences.m2
mahrud: I believe this is identical to before, I may have indented it to clearly understand what was going on in the function in order to understand it. Try clicking the gear logo next to "Submit review" and select "Hide whitespaces".

M2/Macaulay2/packages/SpectralSequences.m2
mahrud: If it was in PushForwards it could definitely be simplified, but I don't know what simplification you have in mind either here or if it were moved to Complexes.

M2/Macaulay2/packages/SpectralSequences.m2
mahrud: I'm not sure what you mean by "preferred interface". Maybe `spots` should be defined in Complexes instead, but the output is _not_ the same as the original implementation. This implementation is much faster.

d-torrance: I think what @mikestillman meant (correct me if I'm wron) was the public interface. If the implementation of Complex were to change, then this would break.

M2/Macaulay2/packages/SpectralSequences.m2
d-torrance: I think the idea was similar to above - this is sensitive to implementation of Complexes

M2/Macaulay2/packages/SpectralSequences.m2
d-torrance: You're right - this is the same as before. But while we're modifying these lines, we might as well fix this too

M2/Macaulay2/packages/SpectralSequences.m2
d-torrance: Makes sense

M2/Macaulay2/packages/SpectralSequences.m2
mahrud: I think that's a reasonable argument for why a function like `spots` or `support` should be defined in the Complexes package.

M2/Macaulay2/packages/SpectralSequences.m2
mahrud: There are plenty of fixes like this in the follow up PR. Once this is merged I'm happy to add this when I rebase that PR, but I wrote this PR last July and would like to not spend much time on it.

M2/Macaulay2/packages/SpectralSequences.m2
mahrud: I've been implementing many operations on complexes using this interface. I'm not sure if it's reasonable to expect users to not use the hash tables `C.module` or `C.dd.map` when they enable really clean and efficient implementations. I would welcome methods like `terms(Complex)` or `maps(Complex)` to return these hash tables, but that's a suggestion for Complexes.

d-torrance:
@eloisagrifo @aboocher @n-m-g: This pull request appears to fix some issues from the current version of the package.  In particular, some of the examples currently give different output than previously, and this fixes those. Is this okay to merge?

aboocher:
@d-torrance @eloisagrifo @n-m-g This sounds good to me.  Thank you so much for your help with this.

PR #3923, opened 2025-07-09
Regenerate tvalues.m2 if d directory has been updated

d-torrance:
Now that tvalues.m2 is generated at startup instead of during the build process, it's no longer updated when there have been changes to the d files, which may potentially include new methods.  If someone adds a new method in the d directory using a "typical value" comment, then they need to manually delete tvalues.m2 so that it can be regenerated.

To simplify this, we now check the file time of the d directory and compare it to the generated file.  If the d directory is newer, then we regenerate the file.

### DIFF
--- a/M2/Macaulay2/m2/typicalvalues.m2
+++ b/M2/Macaulay2/m2/typicalvalues.m2
@@ -125,8 +125,12 @@ generateTypicalValues = (srcdir) -> (
     outfile << "-- DONE: generated based on " | version#"git description" << endl << close)
 
 -- if missing or not successfully generated, tvalues.m2 is regenerated directly
-if not fileExists typicalValuesSource or not match("-- DONE", get typicalValuesSource)
-then generateTypicalValues(currentFileDirectory | "../d/")
+ddir := currentFileDirectory | "../d/"
+if (
+    not fileExists typicalValuesSource or
+    not match("-- DONE", get typicalValuesSource) or
+    isDirectory ddir and fileTime typicalValuesSource < fileTime ddir)
+then generateTypicalValues ddir
 
 -----------------------------------------------------------------------------
 -- numerical functions that will be wrapped


M2/Macaulay2/m2/typicalvalues.m2
```diff
@@ -125,8 +125,12 @@ generateTypicalValues = (srcdir) -> (
     outfile << "-- DONE: generated based on " | version#"git description" << endl << close)
 
 -- if missing or not successfully generated, tvalues.m2 is regenerated directly
-if not fileExists typicalValuesSource or not match("-- DONE", get typicalValuesSource)
-then generateTypicalValues(currentFileDirectory | "../d/")
+ddir := currentFileDirectory | "../d/"
+if (
+    not fileExists typicalValuesSource or
+    not match("-- DONE", get typicalValuesSource) or
+    isDirectory ddir and fileTime typicalValuesSource < fileTime ddir)
```
mahrud, line 132: Does directory time update when any file inside is modified?

d-torrance: It works on my system, so I assume this is the case in general on Linux. I'd be interested if it works on macOS.

mahrud: It doesn't seem to be the case on mine; try `touch`ing a file:
```
-rw-r--r--.  1 mahrud mahrud  19892 Jul 10 16:46 parse.d
drwxr-xr-x.  2 mahrud mahrud   4096 Jul  9 18:53 .
```

mahrud: Anyway, if this resolves the problem you are having, I don't mind merging it as is, just pointing out that it might not fully eliminate that problem.

d-torrance: Yeah, just touching a file doesn't update the timestamp on my system either, but editing one does.

mahrud: I believe changing a file through git doesn't change the directory also.

d-torrance: It worked for me.  I checked out `stable`:

```m2
i1 : fileTime "~/src/macaulay2/M2/M2/Macaulay2/d"

o1 = 1752520439
```
Then checked out `development`:

```m2
i2 : fileTime "~/src/macaulay2/M2/M2/Macaulay2/d"

o2 = 1752520458
```

mahrud: Hmm I thought I tried this out earlier and it didn't work, but now can't reproduce it!

mahrud (approved):

PR #3925, opened 2025-07-10
removed duplicates from engine.h

mahrud:
Pretty self explanatory PR. There's more work in this direction, but I don't know when I'll get to it.


d-torrance:
Discussed with Mike

PR #3926, opened 2025-07-10
Fixed documentation of 'dual' from BettiCharacters

mahrud:
Documenting global symbols like `dual` in a package is not a good idea because it makes them unreachable, and links to them might break. In this case, removing just `dual` from the documentation keys led to an error because two different ways of using the method with different types of input were being documented, so I updated the documentation of `dual` in BettiCharacters to use multiple synopses.


mahrud:
@galettof have you had a chance to look at this pull request?

galettof:
This looks good, thank you! I also like the separation of the use cases.

PR #3931, opened 2025-07-15
Use Macaulay2 webpage for lrslib tarball

d-torrance:
The [lrslib webpage](https://cgm.cs.mcgill.ca/%7Eavis/C/lrslib) is giving 403 forbidden errors.

I think this should also be cherry-picked onto stable.

### DIFF
--- a/M2/cmake/build-libraries.cmake
+++ b/M2/cmake/build-libraries.cmake
@@ -864,7 +864,7 @@ _ADD_COMPONENT_DEPENDENCY(programs gfan "gmp;cddlib" GFAN)
 # http://www-cgrl.cs.mcgill.ca/~avis/C/lrs.html
 # TODO: the shared library target doesn't work on Apple
 ExternalProject_Add(build-lrslib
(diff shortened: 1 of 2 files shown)


M2/cmake/build-libraries.cmake
```diff
@@ -864,7 +864,7 @@ _ADD_COMPONENT_DEPENDENCY(programs gfan "gmp;cddlib" GFAN)
 # http://www-cgrl.cs.mcgill.ca/~avis/C/lrs.html
 # TODO: the shared library target doesn't work on Apple
 ExternalProject_Add(build-lrslib
-  URL               https://cgm.cs.mcgill.ca/~avis/C/lrslib/archive/lrslib-073.tar.gz
+  URL               https://macaulay2.com/Downloads/OtherSourceCode/lrslib-073.tar.gz
```
mikestillman, line 867: As a general rule, I prefer to have tarballs in places we control, so this is good!

mikestillman (approved):
This works for me!

PR #3934, opened 2025-07-19
Remove unused usr-build directory

d-torrance:
This is where the old Makefile would build the various autotools programs before #3520.


mahrud:
I'm not aware of how this is used in the autotools build. Feel free to merge I guess.

PR #3936, opened 2025-07-19
Bump nauty to 2.9.0

d-torrance:
This was released recently


d-torrance:
Except for the unrelated normaliz problem in the cmake-macos build, the builds all worked!
https://github.com/Macaulay2/M2/actions/runs/16393951256

mahrud:
The new nauty changes some test results in normaliz, causing failures: https://github.com/Macaulay2/homebrew-tap/actions/runs/16397880778/job/46332980368?pr=281
See https://github.com/Normaliz/Normaliz/issues/436.

d-torrance:
Oh fun...

When we build normaliz as part of the M2 build, we pass `--without-nauty`, so it shouldn't matter for that at least.

mahrud:
Still, I'd rather wait if Nauty is about to release another patch fix.

mahrud (approved):

PR #3939, opened 2025-07-22
Add package Tableaux

j-graf:
This is a standalone package for computations with (skew) Young tableaux.

This implements the classes SkewTableau and YngTableau, which support

- Skew shapes,
- Shapes that are compositions,
- Shapes with negative parts.


d-torrance:
Thanks for the contribution!  Would you click the "Resolve conflicts" button to resolve the conflicts in the `=distributed-packages` file?  It should just be a matter of removing the `<<<<`, `=====`, and `>>>>>` lines that appear.  (See https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/addressing-merge-conflicts/resolving-a-merge-conflict-on-github for more.)

j-graf:
Got it, done! And thank you for the link

M2/Macaulay2/packages/Tableaux.m2
```diff
@@ -0,0 +1,69 @@
...
+export {"SkewTableau", "skewTableau",
+        "youngDiagram", "ferrersDiagram", "drawInnerShape",
+        "skewShape", "standardize", "rowEntries", "columnEntries", "rowRange", "columnRange",
+        "isSkew", "isWeaklyDecreasing", "isNonnegative",
+        "toPosition", "toIndex", "positionList", "applyEntries", "applyPositions",
+        "verticalConcatenate", "shift", "unshift",
+        "boxContent", "hookLength",
+        }
+
+export {"YngTableau", "yngTableau",
```
mahrud: Is there a particular reason you decided against `YoungTableau` and `youngTableau`? The [Package Writing Style Guide](https://github.com/Macaulay2/M2/wiki/Package-Writing-Style-Guide) recommends to "Avoid acronyms in identifier names."

M2/Macaulay2/packages/Tableaux/SkewTableaux.m2
```diff
@@ -0,0 +1,578 @@
...
+        )
+    )
+
+-- making new tableau
+
+applyEntries = method(TypicalValue => SkewTableau)
+applyEntries (SkewTableau,Function) := (T,f) -> skewTableau(truncate skewShape T, apply(entries T, theBox -> f theBox))
+
+applyPositions = method(TypicalValue => SkewTableau)
+applyPositions (SkewTableau,Function) := (T,f) -> skewTableau(truncate skewShape T, apply(positionList T, thePosition -> f thePosition))
```
mahrud: Very minor suggestion: `apply(L, x -> f x)` can be shortened to `apply(L, f)`.
```suggestion
applyEntries = method(TypicalValue => SkewTableau)
applyEntries (SkewTableau,Function) := (T,f) -> skewTableau(truncate skewShape T, apply(entries T, f))

applyPositions = method(TypicalValue => SkewTableau)
applyPositions (SkewTableau,Function) := (T,f) -> skewTableau(truncate skewShape T, apply(positionList T, f))
```

M2/Macaulay2/packages/Tableaux/documentation.m2
```diff
@@ -0,0 +1,1193 @@
...
+Inputs
+    T:SkewTableau
+      a tableau.
+Outputs
+    b:Boolean
+      returns false if the inner shape is $0$, and true otherwise.
+Description
+  Example
+    isSkew skewTableau(new Partition from {6,6,5,3,1}, new Partition from {2,1,1})
+    isSkew skewTableau(new Partition from {6,6,5,3,1}, new Partition from {})
```
mahrud: I'm a little confused here, likely because I don't remember the definitions, but shouldn't the output of `skewTableau` always be a skew tableau? Would it make sense to have a more general `Tableau` type as the input of `isSkew`, or rename it to `isWellDefined` instead?

M2/Macaulay2/packages/Tableaux/SkewTableaux.m2
```diff
@@ -0,0 +1,578 @@
...
+
+isWeaklyDecreasing = method(TypicalValue => Boolean)
+isWeaklyDecreasing Partition := lam -> rsort toList lam == toList lam
+isWeaklyDecreasing (Partition,Partition) := (lam,mu) -> isWeaklyDecreasing lam and isWeaklyDecreasing mu
+isWeaklyDecreasing SkewTableau := T -> isWeaklyDecreasing skewShape T
+
+isNonnegative = method(TypicalValue => Boolean)
+isNonnegative Partition := lam -> all(toList lam, thePart -> thePart >= 0)
+isNonnegative (Partition,Partition) := (lam,mu) -> isNonnegative lam and isNonnegative mu
+isNonnegative SkewTableau := T -> isNonnegative standardize skewShape T
```
mahrud, line 556: If this is helpful, you can also define `SkewTableau ? ZZ` to allow the syntax `T >= 0` if you'd like.

M2/Macaulay2/packages/Tableaux/SkewTableaux.m2
```diff
@@ -0,0 +1,578 @@
...
+SkewTableau_ZZ := (T,colIndex) -> (
+    colInds := columnRange T;
+    
+    if colIndex < min(colInds#0,0) or colIndex > max(colInds#-1,0) then error "index out of bounds";
+    
+    for rowIndex from 0 to numRows T - 1 list T^rowIndex#colIndex
+    )
+
+columnEntries = method(TypicalValue => List)
+columnEntries (ZZ,SkewTableau) := (colIndex,T) -> delete(null,T_colIndex)
```
mahrud, line 192: I'm a little confused what's happening with `null` here. Is there a way to avoid null entries here?

M2/Macaulay2/packages/Tableaux/SkewTableaux.m2
```diff
@@ -0,0 +1,578 @@
...
+        );
+
+    theRowOrderedRight|theRowOrderedLeft
+    )
+
+SkewTableau_Sequence := (T,thePosition)->(
+    (rowIndex,colIndex) := thePosition;
+    (lam,mu) := standardize skewShape normalizeNegativeRows T;
+
+    if not isMember(rowIndex, rowRange T) then error("index "|toString(rowIndex)|" out of range");
```
mahrud: I think this is always faster, no?
```suggestion
    if not rowIndex < numRows T then error("index "|toString(rowIndex)|" out of range");
```

M2/Macaulay2/packages/Tableaux/SkewTableaux.m2
```diff
@@ -0,0 +1,578 @@
...
+listNegativeRows SkewTableau := T -> (
+    (lam,mu) := standardize skewShape T;
+
+    for i from 0 to #lam-1 list (mu#i > lam#i)
+    )
+
+-- retrieving rows and columns
+
+numcols SkewTableau := T -> (
+    (lam,mu) := standardize skewShape normalizeNegativeRows T;
```
mahrud, line 113: Would it make sense to use caching for `standardize skewShape normalizeNegativeRows T`? It seems to be called very frequently.

mahrud (commented):
Thanks for the contribution!

I also realized that there's a [YoungTableau](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/SpechtModule/html/___Young__Tableau.html) type in the `SpechtModule` package. Have you looked at how the implementation there differs from yours? It would be convenient if that package could load yours so the packages are interoperable.

M2/Macaulay2/packages/Tableaux.m2
```diff
@@ -0,0 +1,69 @@
+newPackage(
+    "Tableaux",
+    Version => "1.0",
+    Date => "July 22, 2025",
+    Authors => {
+	{Name => "John Graf", Email => "jrgraf@alumni.ncsu.edu", HomePage => "https://j-graf.github.io/"}},
+    Headline => "a package for constructing skew tableaux",
```
d-torrance: The "a package for" in the `Headline` option is redundant and should be removed.  This is why a couple of the test builds are failing.

Instead, it should be something like:
```suggestion
    Headline => "constructing skew tableaux",
```

d-torrance (changes requested):

M2/Macaulay2/packages/Tableaux/SkewTableaux.m2
j-graf:
```suggestion
    if rowIndex < 0 or rowIndex >= numRows T then error("index "|toString(rowIndex)|" out of range");
```
Yes definitely, although I also want a lower bound here.

M2/Macaulay2/packages/Tableaux/documentation.m2
j-graf: Yes you're right, I think the choice of name is confusing. The purpose is just to check if the inner shape is 0. This is probably only useful internally, so I just removed `isSkew` from the export list.

M2/Macaulay2/packages/Tableaux.m2
j-graf: To answer your later questions as well, I referenced `SpechtModule`'s documention to make this package compatible, where possible. In a future update I would like to add the remaining missing features, and then I can try to incorporate `Tableaux` into `SpechtModule` if that is desirable. The biggest difference is that the data is mutable in `SpechtModule`.

M2/Macaulay2/packages/Tableaux/SkewTableaux.m2
j-graf: If there is a gap between rows (i.e., the shape is not weakly decreasing), then the list returned by `T_colIndex` has `null`s wherever there is a gap, to maintain consistent indices. Nulls could be avoided here by copying and slightly changing the code from `_`. For the sake of maintainability I would prefer to just call on `_`.

M2/Macaulay2/packages/Tableaux/SkewTableaux.m2
j-graf: I am now thinking this is only really useful internally, like `isSkew`, so I think I might not export it in a future version. If I do decide to keep it, I will use this suggestion.

M2/Macaulay2/packages/Tableaux/SkewTableaux.m2
j-graf: Yes, this is definitely a good idea. I will add this to my TODO list for a future update.

M2/Macaulay2/packages/Tableaux.m2
mahrud: In that case, I think you should still use the full name "YoungTableau".

j-graf: Ok, I've updated the name to YoungTableau.

mahrud (approved):

d-torrance (approved):

PR #3940, opened 2025-07-25
Bump linbox to 1.7.1

d-torrance:
We don't actually use linbox yet (so not bothering with the GitHub builds), and we only provide the option of building it during the autotools build, but they just released a new version.


d-torrance:
Discussed with Mike

Issue #3937, opened 2025-07-21
primaryDecomposition crashing with "incorrect Hilbert function given, aborting"


Devlin-Mallory:
I'm having an issue where M2 crashes when taking the primaryDecomposition of the zero ideal of a multigraded quotient ring.

The minimal example that seems to fail is:
```p=2 -- pick your prime of choice
S=(ZZ/p)[x_1,x_2,y_1,y_2,Degrees=>{{1,0},{1,0},{0,1},{0,1}}]
I=ideal random({2,2},S)
R=S/I
primaryDecomposition ideal 0_R
```
which fails with error
```
internal error: incorrect Hilbert function given, aborting
exp[0]: 2   deg: 4
```

Any multidegree except {1,1} or {1,2} has the same behavior.

This is on the most recent development branch, and the same happens on the web interface.

mahrud:
I think this is happening during a call to `kernel`, see #1935.

Devlin-Mallory:
> I think this is happening during a call to `inverse`, see [#1935](https://github.com/Macaulay2/M2/issues/1935).

or the call to `kernel`, perhaps, as `inverse f` seems to work okay in [#1935] (somehow... not sure what it's supposed to be mathematically).

mikestillman:
I have assigned myself to this.  During the kernel command, there must be a bug in using the Hilbert function.  In any case, the program should not crash.  Same with #1935.

mikestillman:
I see the problem.  The Hilbert function is being given to the Groebner basis routines, and it was not designed for multi-graded input, but is not checking that.  (More precisely, it is assuming the heft function to be the first unit vector, which is generally no longer the case). However, I think I can make it work.

mikestillman:
In order to fix this issue, I plan to add a function that checks whether hilbert function hints work for a given ring, ideal, matrix or module.  This is basically the unexported function `checkHilbertHint` in `m2/gb.m2`, but it should be called something else, and it isn't correct yet in any case: the check that all the generators have degree > 0 is too restrictive, and also tower rings are not allowed.  So some situations where this currently fails:

- `(GF 8)[x,y]` the generator of `GF 8` is being checked for having degree > 0 (which is not needed or desired)
- multigraded, homogeneous input
- tower rings, e.g. `(ZZ/101[a,b])[x,y, Join => false]`, where there is a homogeneous input in that ring.

[...]

Issue #1935, opened 2021-02-12
internal error crash when computing kernel (incorrect Hilbert function)


mahrud:
> Related crash:
> 
> ```m2
> R = QQ[x,y,z,w,u]/(u^4 - x*y*z*w)
> S = R/(x^3 - y^2*z)
> T = S/(y*w^2 - z*x^2)
> f = map(T, S)
> kernel inverse f -- internal error: incorrect Hilbert function given, aborting
> ```

In addition to giving the correct Hilbert function, it would be good to fix the engine code so when an incorrect Hilbert function is given at top level, it is a simple error, not a crash.  That might be worth a separate issue.

_Originally posted by @DanGrayson in https://github.com/Macaulay2/M2/issues/1522#issuecomment-778180240_

mahrud:
@sashahbc and @eisenbud were running into this again, via `preimage`.

mikestillman:
I see the crash, and I can fix it, giving an error.  The base problem here is that the ring map `inverse f` is not well defined, and the code for using it is assuming it is well defined, and ends up using the wrong Hilbert function.  At the moment, my error message (not checked in yet) is a bit inscrutable for the end user (it says: incorrect Hilbert function, but they know nothing about this intermal detail (the hilbert function even being used), and it doesn't address the actual issue).

d-torrance:
Fixed in #3942

Issue #1347, opened 2020-06-26
Thirteen Ways of Looking at a GröbnerBasis


mahrud:
This is a tad overwhelming:
```m2
i1 : about "bner"

o1 =
     {31 => Macaulay2Doc::computing Groebner bases                                    }
     {32 => Macaulay2Doc::Elementary uses of Groebner bases I. Math 634 Fall 2005     }
     {33 => Macaulay2Doc::fine control of a Groebner basis computation                }
     {36 => Macaulay2Doc::Gröbner bases                                              }
     {37 => Macaulay2Doc::Groebner basis examples and applications                    }
     {38 => Macaulay2Doc::GroebnerBasis                                               }
     {39 => Macaulay2Doc::groebnerBasis                                               }
     {40 => Macaulay2Doc::groebnerBasis(..., Strategy => ...)                         }
     {44 => Macaulay2Doc::GroebnerBasisOptions                                        }
     {58 => Macaulay2Doc::simple Groebner basis computations over various rings       }
     {64 => Macaulay2Doc::Tutorial: Elementary uses of Groebner bases                 }
     {65 => Macaulay2Doc::what is a Groebner basis?                                   }
...
     {14 => Macaulay2Doc::gb                                       }
```

mahrud:
PS:
```m2
      {2 => Macaulay2Doc::definition of product (block) orders            }
      {3 => Macaulay2Doc::examples of specifying alternate monomial orders}
      {4 => Macaulay2Doc::monomial orderings                              }
      {5 => Macaulay2Doc::monomial orders for free modules                }
      {6 => Macaulay2Doc::obtaining the monomial order of a ring          }
      {7 => Macaulay2Doc::order                                           }
      {8 => Macaulay2Doc::Schreyer orders                                 }
```

mahrud:
Continuing in this trend:
``m2
i1 : about "modules"
...
     {3 => Macaulay2Doc::constructing maps between modules      }
     {4 => Macaulay2Doc::equality and containment of modules    }
     {5 => Macaulay2Doc::free modules                           }
     {6 => Macaulay2Doc::free resolutions of modules            }
     {7 => Macaulay2Doc::generators of ideals and modules       }
     {8 => Macaulay2Doc::graded modules                         }
     {9 => Macaulay2Doc::homomorphisms (maps) between modules   }
     {10 => Macaulay2Doc::ideals to and from modules            }
     {11 => Macaulay2Doc::information about a map of modules    }
[...]

PR #3942, opened 2025-07-30
Git issues 3937 1935

mikestillman:
This PR should fix issues #3937, #1935.  It removes the crash when groebner bases are given a Hilbert function "hint", and the data is not correct (so likely gives errors, but still, it should not crash).

Additionally:

- A function `canUseHilbertHint` has been added, documented, which allows one (and the GB code) to decide if in a given situation, the Hilbert hint is usable.
- Tests have been added
- The abort when discovering a bad Hilbert function is removed.  Instead, an exception is issued.
- Note: multi-graded rings will not use any Hilbert hint.  If desired, we can likely add this, but it will take some effort and time.  Singly graded rings work, as long as all of the variables have positive degree.


M2/Macaulay2/e/hilb.cpp
```diff
@@ -709,33 +709,26 @@ int hilb_comp::coeff_of(const RingElement *h, int deg)
   // exp[0]=deg.
   const PolynomialRing *P = h->get_ring()->cast_to_PolynomialRing();
 
-  exponents_t exp = newarray_atomic(int, P->n_vars());
+  exponents_t exp = new int[P->n_vars()];
```
mahrud, line 712: What's the reason behind this change? If this is correct in this case I wonder about tons of other `exponents_t` initializations which use `newarray_atomic` (only three exceptions, and this makes it four). The comment before `newarray_atomic` in `newdelete.hpp` says:
> // this replaces all uses of the construction "new T[n]", with T containing NO pointers.

M2/Macaulay2/packages/Macaulay2Doc/functions/canUseHilbertHint-doc.m2
```diff
@@ -0,0 +1,61 @@
+doc ///
+  Key
+    canUseHilbertHint
+    (canUseHilbertHint, Ring)
+    (canUseHilbertHint, Ideal)
+    (canUseHilbertHint, Module)
+    (canUseHilbertHint, Matrix)
+  Headline
+    whether certain Groebner computations can make use of the Hilbert function
```
mahrud, line 9: I suspect nobody will actually find this node. How about this: add the example of passing the `Hilbert` option to one of the [many](https://github.com/Macaulay2/M2/issues/1347) nodes about Grobner bases instead, or add a separate node for the option `[gb, Hilbert]` and link to it from the documentation of `gb`. I would imagine `canUseHilbertHint` can stay internal (except imported when needed by MinimalPrimes) in that case.

d-torrance:
Should `checkHilbertHint` be exported?  We're getting this error:

```m2
ii22 : assert checkHilbertHint generators graphIdeal f
../../../../../Macaulay2/tests/normal/ker.m2:24:23:(3):[5]: error: no method for adjacent objects:
            checkHilbertHint (of class Symbol)
    SPACE   | -p_0^4+p_1 -p_0^5+p_2 -p_0^6+p_3 | (of class Matrix)
```

M2/Macaulay2/e/hilb.cpp
mikestillman: Local array of ints do not need to be garbage collected, and actually freeing them causes slowdowns in the garbage collector, I believe.  This could be changed in many places in the c++ code.

mikestillman:
This PR fixes issues #3937, #1935, #4355.

mahrud:
> * Note: multi-graded rings will not use any Hilbert hint.  If desired, we can likely add this, but it will take some effort and time.  Singly graded rings work, as long as all of the variables have positive degree.

Yes, absolutely desired :)

pzinn:
does it fix https://github.com/Macaulay2/M2/issues/732 too?

mikestillman:
@d-torrance This one is possibly ready to go, or more review perhaps?

d-torrance:
I checked and we haven't moved `hilb.cpp` in the Great Sort PR, so this should be good!

pzinn:
looks like it does *not* fix #732.

PR #3948, opened 2025-08-13
MatrixFactorizations Package

kellerlv:
A new matrix factorizations package, a package for computing with length d matrix factorizations. Authors David Favero, Sasha Pevzner, Tim Tribone, and Keller VandeBogert.


M2/Macaulay2/packages/MatrixFactorizations.m2
```diff
@@ -0,0 +1,136 @@
...
+	    Email => "pevzn002@umn.edu",
+	    HomePage => ""},
+	{Name => "Timothy Tribone",
+	    Email => "tim.tribone@utah.edu",
+	    HomePage => ""},
+        {Name => "Keller VandeBogert", 
+            Email => "kvandebo@nd.edu", 
+            HomePage => "https://sites.google.com/view/kellervandebogert/home"}
+    },
+    Headline => "a package for computing with length d matrix factorizations",
```
d-torrance: "a package for" here is redundant (and will cause one of the Core tests to fail).  Also, what's d?  Maybe "matrix factorizations of a particular length" or something like that?

M2/Macaulay2/packages/MatrixFactorizations.m2
```diff
@@ -0,0 +1,136 @@
...
+        "Complexes",
+	"CompleteIntersectionResolutions",
+	"TensorComplexes"
+    },
+    PackageImports => {
+        "Complexes",
+	"CompleteIntersectionResolutions",
+	"TensorComplexes"
+    },
+    DebuggingMode => true
```
d-torrance: `DebuggingMode` should be false, or even better, removed entirely since false is the default

d-torrance: `newPackage` should also have a `Keywords` option.  The value is a list of strings, ideally a small subset of the existing keywords, which are the headings on the [packages provided with Macaulay2](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_packages_spprovided_spwith_sp__Macaulay2.html) documentation page

d-torrance (changes requested):
Thanks for the contribution!

Without diving into the code too much yet, here are a couple quick comments.  Also, there are a couple typos:
```
M2/Macaulay2/packages/MatrixFactorizations/KellerTests.m2:671: documnetation ==> documentation
M2/Macaulay2/packages/MatrixFactorizations/MatrixFactorizationsDOC.m2:707: followig ==> following
```

M2/Macaulay2/packages/MatrixFactorizations.m2
kellerlv: Sorry about that. I've added those things now and fixed the typos.

M2/Macaulay2/packages/MatrixFactorizations.m2
```diff
@@ -0,0 +1,136 @@
...
+    PackageExports => {
+        "Complexes",
+	"CompleteIntersectionResolutions",
+	"TensorComplexes"
+    },
+    PackageImports => {
+        "Complexes",
+	"CompleteIntersectionResolutions",
+	"TensorComplexes"
+    }
```
d-torrance, line 32: There's no reason to duplicate the packages in `PackageExports` and `PackageImports`.  Do you want to export those packages' symbols to the user, or just use them internally?

M2/Macaulay2/packages/MatrixFactorizations/CIResCompatibility.m2
```diff
@@ -0,0 +1,240 @@
+needsPackage "CompleteIntersectionResolutions"
+needsPackage "PushForward"
```
d-torrance: `CompleteIntersectionResolutions` will already be pulled in from `PackageImports` or `PackageExports`.  I'd also move `PushForward` to one of those.

M2/Macaulay2/packages/MatrixFactorizations/CIResCompatibility.m2
```diff
@@ -0,0 +1,240 @@
...
+    t := local t;
+    S := Q[t_1..t_(length L)];
+    f := sum(1..length L,i->L_(i-1)*t_i);
+    C = C**S;
+    H := makeHomotopies(matrix{{f}}, (C));
+    Ln := apply(keys H,i->(i_1+2*(i_0)_0-1,i_1));
+    Hn := new HashTable from for i from 0 to length Ln-1 list Ln_i => H#((keys H)#i);
+    (lo,hi) := concentration C;
+    Lo := select(toList(lo-1..hi+1),i->odd(i));
+    Le := select(toList(lo-1..hi+1),i->even(i));
```
d-torrance: Minor quibble: `i -> odd i` and `i -> even i` can be replaced with just `odd` and `even`, resp.

M2/Macaulay2/packages/MatrixFactorizations/KellerTests.m2
```diff
@@ -0,0 +1,964 @@
...
+
+
+TEST ///
+--playing around with pruning maps/cached data of pruning maps (pruning 0 maps)
+
+///
+
+TEST ///
+--playing around with adjoining roots of unity
+///
```
d-torrance, line 153: Were there intended to be tests here?

M2/Macaulay2/packages/MatrixFactorizations/MF_functions.m2
```diff
@@ -0,0 +1,1152 @@
+
+--currently not working, but will put it back: needs "ZZdFactorizationsCODE.m2"
+--take this one out when putting above back in:
+--needs "ZZdFactorizations.m2"
+needsPackage "TensorComplexes" --Koszul factorization method needs this
+needsPackage "Cyclotomic" --for adjoining roots of unity
```
d-torrance: These should also be moved to `PackageImports` or `PackageExports`.

M2/Macaulay2/packages/MatrixFactorizations/MF_functions.m2
```diff
@@ -0,0 +1,1152 @@
...
+	);
+    )
+
+hhh(ZZdFactorization) := ZZ => F -> (p := period F;
+    if not(p==2) then error "expected 2-periodic factorization";
+    hhh(0,F) + hhh(1,F)
+    )
+
+--euler characteristic of a ZZdfactorization. Caveart: in order to be well-defined, the input should be a complex
+eulerChi = method();
```
d-torrance: `euler` already exists in Core for the Euler characteristic of various classes.  You could overload that instead of defining a new method function

M2/Macaulay2/packages/MatrixFactorizations/ZZdFactorizations.m2
```diff
@@ -0,0 +1,1754 @@
+needsPackage "Complexes"
```
d-torrance: This also should be in `PackageImports` or `PackageExports`

M2/Macaulay2/packages/MatrixFactorizations/ZZdFactorizations.m2
```diff
@@ -0,0 +1,1754 @@
...
+
+Symbol ^ ZZdFactorization := ZZdFactorizationMap => (sym, C) -> (
+    if sym === dd then C.dd
+    else error "expected symbol to be 'dd'"
+    )
+
+net ZZdFactorization := C -> (
+     (lo,hi) := (0,C.period);
+     if lo > hi then 
+         error "In a complex, lo <= hi should always hold in the concentration"
```
d-torrance: Error messages usually begin with a lowercase letter

d-torrance (changes requested):
Thanks for the quick response to the previous points!  I dug through the code a bit more and have a few more suggestions

M2/Macaulay2/packages/MatrixFactorizations.m2
kellerlv: When this was initially started, the "Complexes" package was not the default. So it's probably unnecessary to load that now. And yes the other two packages are only called internally -- I was loading them by default for my own testing purposes, but I don't think it's necessary to export their symbols automatically for users.

M2/Macaulay2/packages/MatrixFactorizations/CIResCompatibility.m2
kellerlv: Yes PushForward should be moved to just the package imports again.

M2/Macaulay2/packages/MatrixFactorizations/CIResCompatibility.m2
kellerlv: Oh yeah I'm not sure why I wrote it like that.

M2/Macaulay2/packages/MatrixFactorizations/KellerTests.m2
kellerlv: I have a memory of writing a bunch of tests for this. I need to look into if I accidentally pushed an older version of the testing file.

M2/Macaulay2/packages/MatrixFactorizations/ZZdFactorizations.m2
kellerlv: Actually that's a typo on my part anyway, it should say "factorization" not "complex". I'm going to wait and see if this final set of checks goes through successfully and then I'll implement these changes along with fixing any errors that might pop up if it fails.

M2/Macaulay2/packages/MatrixFactorizations.m2
mahrud: It is still necessary to load or export Complexes if the package uses it.

d-torrance (approved):
Thanks!

PR #3952, opened 2025-08-19
Capture package examples by default

d-torrance:
This should significantly speed up building since we won't need to fire up a ton of new M2 processes for each package.  This is a draft for now -- we should ensure that the documentation stays unchanged, so I'll compare the build artifacts.

### DIFF
--- a/M2/Macaulay2/packages/LatticePolytopes.m2
+++ b/M2/Macaulay2/packages/LatticePolytopes.m2
@@ -211,7 +211,7 @@ toricBlowUp(Polyhedron,Polyhedron,ZZ) := (P,F,k) -> (
             and contains(e,(facesAsPolyhedra((dim F),F))_0)
         ));
     pt := k * matrix{(vertices Edges_i)_0+(vertices Edges_i)_1-2*(vertices F)_0}*(1/(#latticePoints(Edges_i)-1))+matrix((vertices F)_0);
-    return intersection(A||E,b||(E*pt)));
+    polyhedronFromHData(A||E,b||(E*pt)));
 
 
 
@@ -685,9 +685,9 @@ iskCayleykEdges(Polyhedron):=P->(
 	 if length(L)==2 then(
 	     k:=(L_1-L_0)/gcd(flatten entries(A^{i}));
 	     if k==minLen then(
-		 H0:=intersection(A^{i}||-A^{i},matrix{{L_0},{-L_0}});
+		 H0:=polyhedronFromHData(A^{i}||-A^{i},matrix{{L_0},{-L_0}});
 		 P0=intersection(P,H0);
-		 H1:=intersection(A^{i}||-A^{i},matrix{{L_1},{-L_1}});
+		 H1:=polyhedronFromHData(A^{i}||-A^{i},matrix{{L_1},{-L_1}});
 		 P1=intersection(P,H1);
 		 boolean=false;
 	     );
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -142,8 +142,13 @@ jobs:
       - name: Build Macaulay2 using Ninja
         if: matrix.build-system == 'cmake'
         run: |
-          cmake --build . --target M2-core M2-emacs install-packages
-          if [[ "${{ runner.os }}" == "Linux" ]] && [[ "${{ matrix.build-system }}" == "cmake" ]]; then
+          cmake --build . --target M2-core M2-emacs
+
+      - name: Install packages using Ninja
+        if: matrix.build-system == 'cmake'
+        run: |
+          cmake --build . --target install-packages
+          if [[ "${{ runner.os }}" == "Linux" ]]; then
               sudo apt-get install -y -q --no-install-recommends dpkg-dev
               echo "GIT_COMMIT=`git describe --dirty --always --match HEAD`" >> $GITHUB_ENV
               cpack -G DEB
@@ -168,6 +173,11 @@ jobs:
           ../../configure --with-system-gc --with-fplll
 
       - name: Build Macaulay2 using Make
+        if: matrix.build-system == 'autotools'
+        run: |
+          make -j$(nproc 2>/dev/null || sysctl -n hw.logicalcpu) PACKAGES=
+
+      - name: Install packages using Make
         if: matrix.build-system == 'autotools'
         run: |
           make -j$(nproc 2>/dev/null || sysctl -n hw.logicalcpu)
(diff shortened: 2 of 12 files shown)


M2/Macaulay2/packages/SimpleDoc/doc.txt
```diff
@@ -86,7 +86,7 @@ Node
    Text
      The @TO check@ method executes all package tests defined this way.
    Example
-     check SimpleDoc
+     check SimpleDoc -* no-capture-flag *-
```
mahrud: "check" should probably be in the list of forbidden words.

M2/Macaulay2/packages/LatticePolytopes.m2
mahrud, line 1: I suspect these are because examples are run by capture in a way that a previous example loading a package has overwritten `intersection`? Would this be resolved if examples from each package had their own process and all used capture inside that one process? This might be a good idea to do anyway, since the different processes could run in parallel.

d-torrance: I think that's what we basically do anyway -- start up a new M2 process and run `installPackage` inside it for each package.

This particular commit isn't really fixing a `capture` issue, but one I noticed when comparing the capture v. non-capture diffs.  We currently have this annoying deprecation warning in the [docs](https://www.macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/LatticePolytopes/html/_toric__Blow__Up.html)  (which `capture` misses), and this change fixes that:

```m2
i4 : toricBlowUp(P,Q,1)
Warning: This method is deprecated and will be removed in version 1.11 of Polyhedra. Please consider using polyhedronFromHData instead.

o4 = Polyhedron{...1...}

o4 : Polyhedron
```

d-torrance:
Something that's confusing me is the non-captured v. captured diff of some *LocalRings* examples, e.g.,:

```m2
--- old/usr/share/doc/Macaulay2/LocalRings/example-output/_local__Resolution.out
+++ new/usr/share/doc/Macaulay2/LocalRings/example-output/_local__Resolution.out
@@ -74,11 +74,11 @@
 i11 : C = resolution coker m

 

        1      3      3      1

-o11 = R  <-- R  <-- R  <-- R

-                            

-      0      1      2      3

+o11 = R  <-- R  <-- R  <-- R  <-- 0

+                                   

+      0      1      2      3      4

 

-o11 : Complex

+o11 : ChainComplex
```

Without `capture`, `resolution` uses the new *Complexes* version, but with it, we use the legacy version.  I don't see where any examples are modifying any global variables or anything.

The package does have *OldChainComplexes* in `PackageImports` and *Complexes* in `PackageExports`, which seems odd.  Why use both?

d-torrance:
I think the issue is that *LocalRings* exports *Complexes*, but calling `capture` loads the pre-loaded packages, so it pulls in `OldChainComplexes` again:

```m2
i1 : locate(res, Module)

o1 = /usr/share/Macaulay2/OldChainComplexes/res.m2:284:37-311:90

o1 : FilePosition

i2 : needsPackage "LocalRings"
 -- warning: symbol "GradedModuleMap" in OldChainComplexes.Dictionary is shadowed by a symbol in Complexes.Dictionary
 --   use the synonym OldChainComplexes$GradedModuleMap
 -- warning: symbol "GradedModule" in OldChainComplexes.Dictionary is shadowed by a symbol in Complexes.Dictionary
 --   use the synonym OldChainComplexes$GradedModule
 -- warning: symbol "res" in OldChainComplexes.Dictionary is shadowed by a symbol in Complexes.Dictionary
 --   use one of the synonyms OldChainComplexes$res, OldChainComplexes$resolution
 -- warning: symbol "resolution" in OldChainComplexes.Dictionary is shadowed by a symbol in Complexes.Dictionary
 --   use one of the synonyms OldChainComplexes$res, OldChainComplexes$resolution
 -- warning: symbol "res" in Complexes.Dictionary is shadowed by a symbol in OldChainComplexes.Dictionary
 --   use one of the synonyms Complexes$freeResolution, Complexes$res, Complexes$resolution, freeResolution
 -- warning: symbol "GradedModuleMap" in Complexes.Dictionary is shadowed by a symbol in OldChainComplexes.Dictionary
 --   use one of the synonyms Complexes$ComplexMap, Complexes$GradedModuleMap, ComplexMap
 -- warning: symbol "resolution" in Complexes.Dictionary is shadowed by a symbol in OldChainComplexes.Dictionary
 --   use one of the synonyms Complexes$freeResolution, Complexes$res, Complexes$resolution, freeResolution
 -- warning: symbol "GradedModule" in Complexes.Dictionary is shadowed by a symbol in OldChainComplexes.Dictionary
 --   use one of the synonyms Complex, Complexes$Complex, Complexes$GradedModule

o2 = LocalRings

o2 : Package

i3 : locate(res, Module)

o3 = /usr/share/Macaulay2/Complexes/FreeResolutions.m2:55:36-132:5

o3 : FilePosition

i4 : capture("locate(res, Module)", UserMode => false)

o4 = (false,                                                                 )
             i1 : locate(res, Module)

             o1 = /usr/share/Macaulay2/OldChainComplexes/res.m2:284:37-311:90

             o1 : FilePosition

             i2 : 


o4 : Sequence
```

Since I believe we're switching to *Complexes* in the next release anyway, this is probably a non-issue.

d-torrance:
I think this is in pretty good shape now.  The diff in the example output seems to be just things like timings, temporary file names, and results of shuffling the order of keys in hash tables:

https://gist.github.com/d-torrance/43282670914f1a937e7b6f2faf2d63a0

mahrud:
> Since I believe we're switching to Complexes in the next release anyway, this is probably a non-issue.

I'm pessimistic about this. Lots of issues remain unsolved and several people I've talked to decided to keep using the old version until those are fixed, which means there may be more undiscovered bugs.

d-torrance:
It seems like having a package like *LocalRings* silently load another package that overrides the behavior of `res` isn't great...

Should *Complexes* and *OldChainComplexes* each check the value of `HomologicalAlgebraPackage` and only export `res` if it's them?

mahrud:
> Should _Complexes_ and _OldChainComplexes_ each check the value of `HomologicalAlgebraPackage` and only export `res` if it's them?

Well, I think the idea was that the user loading one or the other should replace what `res` points to, and therefore also whatever package the user loads, but it's not perfect.

antonleykin (changes requested):
I'm not sure what to do with the Complex vs. ChainComplex issue but the rest looks fine. 
The build times are impressively shorter. I would like to suggest to single out the install-packages target (in `test_build.yml`) so that the timing for that is shown separately in every build.

d-torrance:
The package installation is now in its own step for both the cmake and autotools GitHub builds.

antonleykin (approved):

PR #3956, opened 2025-08-25
Some typos

fchapoton:


M2/Macaulay2/editors/emacs
mahrud, line 1: You seem to have changed the emacs hash by accident.

fchapoton:
sorry. Should be better now

d-torrance (approved):
Thank you!!

Issue #3960, opened 2025-08-28
tensoring subquotient modules with ring maps


Devlin-Mallory:
I think that `tensor(RingMap, Module)` is returning the wrong answer when given a subquotient (or image) module. 

An example:
```
S = QQ[x]
M = image matrix{{x}} -- abstractly, M \cong S^1
R = S/x
tensor(map(R, S), M) == 0 -- incorrectly says map(R,S)**M is zero
```
(Note that `tensor(R,M)` gives the correct answer, i.e., a copy of `R^1`.) 

I think it's easy to see why the current code is giving the wrong answer: the code on line 215 of ringmap.m2 for `tensor(QuotientRing, Module)` is applying the ring map (in this case, killing x) to the matrix of generators of M (in this case, x), which is incorrect since tensor product is not left exact. 

Does it even make sense to expect to get back a subquotient module after tensoring? Or should `tensor(f, M)` be a synonym for `tensor(f, coker presentation M)`?

mahrud:
Didn't we decide in #3917 that `tensor(R,M)` should be an error?

Devlin-Mallory:
I think tensor(R,M) (with M an S-module ) could be a useful shorthand for tensor(f, M) when there's a natural map f: S -> R, so maybe just for R = S/I... Actually, I wasn't able to see where tensor(R,M) is defined, so it's not clear to me what is intended. (If I type `methods tensor` or `methods symbol **` there's no method for `QuotientRing, Module`.) 

In this case though, the mathematically well-defined tensor(f, M) is what's going wrong...

mahrud:
`tensor(Ring, Module)` is defined as `(symbol**, Ring, Module)`. Also, I'll point out that the reason `Hom(Module, Module)` over different rings isn't working correctly is that down the line it calls
https://github.com/Macaulay2/M2/blob/2cb101f52dfdd3abe788f2d5e1d0248adeb10b32/M2/Macaulay2/m2/Hom.m2#L44
and tensor of modules over different rings behaves unexpectedly.

PR #3963, opened 2025-08-29
tensor(RingMap, Module) fix

Devlin-Mallory:
This reverts commit 4433d30, which introduced the error in #3960 for `tensor(RingMap, Module)`, and adds a test to ensure tensor products with submodules give the correct answer.


Devlin-Mallory:
Not sure why the macos cmake build is failing...

mahrud (approved):

Issue #3914, opened 2025-07-05
msolve killed when generating whitneyStratifyPol example


d-torrance:
This has been happening pretty frequently recently in the GitHub builds, I think after msolve 0.9.0 was released.

From https://github.com/Macaulay2/M2/actions/runs/16087262096/job/45399922108:

```m2
-- making example results for "whitneyStratifyPol"                          -- running failed
 -- error log:  /home/runner/work/M2/M2/M2/BUILD/build/usr-dist/common/share/doc/Macaulay2/WhitneyStratifications/example-output/_whitney__Stratify__Pol.errors:0:1:(3):[14]: 
 -- input code: /tmp/M2-149705-0/0_whitney__Stratify__Pol.m2:0:1
*** error: M2 exited with status code 1 -- 418.481s elapsed

 -- Summary: 1 example(s) failed in package WhitneyStratifications:
 -- whitneyStratifyPol  ../../../../Macaulay2/packages/WhitneyStratifications.m2:1670:0
../../../../Macaulay2/m2/debugging.m2:18:13:(1):[10]: error: installPackage: 1 error(s) occurred in running examples for package WhitneyStratifications
../../../../Macaulay2/m2/installPackage.m2:787:59:(1):[8]: --back trace--
../../../../Macaulay2/m2/methods.m2:154:98:(1):[7]: --back trace--
../../../../Macaulay2/m2/option.m2:17:14:(1):[6]: --back trace--
../../../../Macaulay2/m2/installPackage.m2:672:18:(1):[5]: --back trace--
../../../../Macaulay2/m2/methods.m2:154:98:(1):[4]: --back trace--
../../../../Macaulay2/m2/option.m2:17:14:(1):[3]: --back trace--
currentString:1:14:(3):[2]: --back trace--
```

Here's a bit more information about the failed example from the build artifacts:

```m2
i16 : elapsedTime W=whitneyStratifyPol(I,Algorithm=>"msolve")
stdio:16:32:(3): error: msolve returned an error (9):
Killed
```

I think this example is too big.  Here's the corresponding part of the documentation:

```m2
     On harder examples the whitneyStratifyPol can be much faster than other
     methods. The following example, while taking some seconds below, takes many
     minutes or hours with other methods. In the example below the output is in
     fact the minimal Whitney stratification (though more often the output
     Whitney stratification returned by whitneyStratifyPol is not minimal).

     +--------------------------------------------------------------------------------------+
[...]

PR #3965, opened 2025-09-01
Fix to a bug in Msolve package RUR function

Martin-Helmer:
@antonleykin : This should fix the variable order bug you noticed in Msolve RUR. I now reorder the RUR numerator based on the variable order returned by Msolve. 

This example now works as expected:
needsPackage "Msolve"
QQ[x,y,z]
I = ideal {(x^3-z)*(x^2-y)}
S = ideal {z-2,y-1}
rur = msolveRUR(I+S, Verbosity=>2)
factor rur#"findRootsUniPoly"
-- point corresponding to T=1 or T=-1                                                                                                                                       
p = matrix{{1}}
p = matrix{{-1}}
apply(rur#"numerator", n->sub(-n,p)) / sub(rur#"denominator",p)

### DIFF
--- a/M2/Macaulay2/packages/WhitneyStratifications.m2
+++ b/M2/Macaulay2/packages/WhitneyStratifications.m2
@@ -1341,12 +1341,7 @@ Node
 	    I=ideal(y^2+x^3-x^2*z^2)
 	    WS3=whitneyStratifyPol(I,Algorithm=>"M2F4")
 	Text 
-	    On harder examples the whitneyStratifyPol can be much faster than other methods. The following example, while taking some seconds below, takes many minutes or hours with other methods. In the example below the output is in fact the minimal Whitney stratification (though more often the output Whitney stratification returned by whitneyStratifyPol is not minimal). 
-	Example 
-	    R=QQ[x..z,t]
-	    I=ideal(z^3+t*y^3*z+y^4*x+x^9)
-	    elapsedTime W=whitneyStratifyPol(I,Algorithm=>"msolve")
-	    peek W
+	    On harder examples the whitneyStratifyPol can be much faster than other methods and often the option Algorithm=>"msolve" will be the fastest (the msolve option is not currently set as default since the msolve package is a new addition to M2). 
 	    
 	    
 Node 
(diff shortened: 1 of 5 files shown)


mahrud:
Could you add the example as a test?

Martin-Helmer:
I added the RUR example it as a test to the tests.m2 file in the Msolve folder.

M2/Macaulay2/packages/WhitneyStratifications.m2
```diff
@@ -1341,12 +1341,7 @@ Node
 	    I=ideal(y^2+x^3-x^2*z^2)
 	    WS3=whitneyStratifyPol(I,Algorithm=>"M2F4")
 	Text 
-	    On harder examples the whitneyStratifyPol can be much faster than other methods. The following example, while taking some seconds below, takes many minutes or hours with other methods. In the example below the output is in fact the minimal Whitney stratification (though more often the output Whitney stratification returned by whitneyStratifyPol is not minimal). 
-	Example 
-	    R=QQ[x..z,t]
-	    I=ideal(z^3+t*y^3*z+y^4*x+x^9)
-	    elapsedTime W=whitneyStratifyPol(I,Algorithm=>"msolve")
```
mahrud, line 1348: Out of curiosity, is there a reason this example is removed?

Martin-Helmer: This was removed at Doug's request since it was taking too long to run the example during M2 builds:

https://github.com/Macaulay2/M2/issues/3914

I am unsure why this is showing up as a change in this pull request as I had thought this was merged into development back in July:
https://github.com/Macaulay2/M2/pull/3918#pullrequestreview-3002634347

d-torrance: It looks like you're using the same branch as that earlier PR, and since the histories are slightly different, that commit is still showing up in the list of commits in this PR since we "rebase and merged" that PR instead of just merging it.  But I think it should be rebased away when this PR is merged.

Martin-Helmer: Thanks Doug! That makes sense. Let me know if there is anything else I should do on my end.

antonleykin (approved):
I think the bug has been fixed.

PR #3968, opened 2025-09-01
CellularResolutions.m2 pull request for JSAG

jkyang92:
This is the pull request for the CellularResolutions package for JSAG. The changes are only the author emails since the package was already previously included in the Macualay2 distribution.

See also, these previous pull requests: #2833, #2844 

CC: @OlaSobieska

### DIFF
--- a/M2/Macaulay2/packages/CellularResolutions.m2
+++ b/M2/Macaulay2/packages/CellularResolutions.m2
@@ -5,8 +5,8 @@ newPackage(
     Version => "1.0",
     Date => "May 17, 2023",
     Authors => {
-        {Name => "Jay Yang", Email => "jayy@wustl.edu"},
-        {Name => "Aleksandra Sobieska", Email => "asobieska@math.wisc.edu"}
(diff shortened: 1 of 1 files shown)


d-torrance (approved):
Thanks!

I'll cancel the builds since the change are so minor

jkyang92:
Were you supposed to merge this yet? I'm not 100% sure what JSAG is trying to do with their review process but I though that there was some attempt to do the technical review via pull requests?

d-torrance:
Ohhhh whoops!  I remember there being some discussion about moving things to PR's for the JSAG review process but it didn't dawn on me that this was it!

Since it's only been 30 min or so, I went ahead and force-pushed `development` back to where it was before your PR so you can try again (or re-open this, if possible).

mahrud:
> Were you supposed to merge this yet? I'm not 100% sure what JSAG is trying to do with their review process but I though that there was some attempt to do the technical review via pull requests?

But this isn't going to work for this purpose because GitHub doesn't allow commenting anywhere that hasn't been edited. You would have to delete the package and make a new PR that adds it again for that purpose.

PR #3975, opened 2025-09-08
Raise error if duplicate symbols are given in function parameters

d-torrance:
I noticed this after fat-fingering some code:

```m2
i1 : f = (x, x) -> x;

i2 : f(3, 4)

o2 = 4
```

It doesn't seem like we should allow duplicate symbols like this.  For contrast, here's Python:

```python
>>> def f(x, x):
...     return x
... 
  File "<stdin>", line 1
SyntaxError: duplicate argument 'x' in function definition
```

So I propose we raise an error, too:

```m2
i1 : f = (x, x) -> x;
stdio:1:8:(3): error: duplicate symbol in parameter list: x
```

I had to tweak some `submatrix` methods that were passing two different `null`'s as parameters to avoid raising this error after the change.


d-torrance:
@pzinn: Would you be willing to take a look at this when you get a chance?

pzinn (approved):

PR #3977, opened 2025-09-09
Remove Debian packaging directory

d-torrance:
This has been unmaintained and unused for a couple years now that Macaulay2 is in the Debian/Ubuntu repositories and our PPA.

The Debian packaging code is available at https://salsa.debian.org/math-team/macaulay2


d-torrance:
I don't understand why the builds are failing in macOS -- brew must be raising an error at some point installing the dependencies, but it's not clear what it is.  The last message is just the usual warning about libffi being keg-only...

d-torrance:
Discussed with Mike

PR #3978, opened 2025-09-09
Only update submodules when we need to (autotools)

d-torrance:
Currently, when we call `make fetch` inside a library directory when the library uses a git submodule, we keep updating the submodule unnecessarily:

```
profzoom@peg-amy:~/src/macaulay2/M2/M2/BUILD/build$ make -C libraries/flint fetch
make: Entering directory '/home/profzoom/src/macaulay2/M2/M2/BUILD/build/libraries/flint'
git submodule update --init /home/profzoom/src/macaulay2/M2/M2/BUILD/build/../../submodules/flint
Submodule 'M2/submodules/flint' (https://github.com/flintlib/flint) registered for path '../../../../submodules/flint'
Submodule path '../../../../submodules/flint': checked out '17950040404e6ed797a4becd8a866fb3f62b5c5e'
make: Leaving directory '/home/profzoom/src/macaulay2/M2/M2/BUILD/build/libraries/flint'
profzoom@peg-amy:~/src/macaulay2/M2/M2/BUILD/build$ make -C libraries/flint fetch
make: Entering directory '/home/profzoom/src/macaulay2/M2/M2/BUILD/build/libraries/flint'
git submodule update --init /home/profzoom/src/macaulay2/M2/M2/BUILD/build/../../submodules/flint
make: Leaving directory '/home/profzoom/src/macaulay2/M2/M2/BUILD/build/libraries/flint'
```

Now we add a little file (`.submodule-updated`) when we do so and only update it again if it's missing.  We remove it (and deinit the submodule) during `make clean`.

```
profzoom@peg-amy:~/src/macaulay2/M2/M2/BUILD/build$ make -C libraries/flint fetch
make: Entering directory '/home/profzoom/src/macaulay2/M2/M2/BUILD/build/libraries/flint'
make update-submodule
make[1]: Entering directory '/home/profzoom/src/macaulay2/M2/M2/BUILD/build/libraries/flint'
git submodule update --init /home/profzoom/src/macaulay2/M2/M2/BUILD/build/../../submodules/flint
Submodule 'M2/submodules/flint' (https://github.com/flintlib/flint) registered for path '../../../../submodules/flint'
Submodule path '../../../../submodules/flint': checked out '17950040404e6ed797a4becd8a866fb3f62b5c5e'
make[1]: Leaving directory '/home/profzoom/src/macaulay2/M2/M2/BUILD/build/libraries/flint'
touch .submodule-updated
[...]

PR #3983, opened 2025-09-15
Update rpm package creation script

d-torrance:
We move the rpm package creation script from the `distributions` directory to the new `BUILD/rpm` directory (see https://github.com/Macaulay2/M2/issues/1091#issuecomment-2409536016).  Furthermore, we flip things around so that rather than calling `rpmbuild` after building M2 to create our rpm, we build M2 inside `rpmbuild`, just like the packages in the official Fedora/RHEL repositories.  Each package is build inside a docker image for the appropriate distribution.   We also add a GitHub workflow to build rpm's for various Fedora and RHEL distributions to simplify the process at each new M2 release.

The commits from this branch (cherry-picked onto `stable`) were used to generate the shiny new M2 1.24.06 RHEL rpm's available on the website (see https://github.com/Macaulay2/M2/wiki/Installing-Macaulay2-in-Linux#red-hat-enterprise-linux).

A few additional changes:

* This is on top of #3978.  Since `rpmbuild` uses an unpacked tarball for the source instead of a git repository, we need need to trick it into thinking that we've already obtained the submodules of the dependencies that use them.
* Some dependency updates to the Fedora and RHEL dockerfiles so we have the things we need to build the packages.  In particular, we drop the EPEL as a dependency for the RHEL packages.  This means we have to build a few more things, but it should simplify things for the end user since it's one less thing to worry about.
* We remove the option to build for encap support.  As far as I can tell, encap has been abandoned for 10+ years.  This simplifies the following:
* After installation, we fix the rpaths of any libraries and programs distributed with Macaulay2.  In particular, they're hardcoded by autotools & libtool during build with the build directory, and so we change them to the installation directory using the `chrpath` utility.  This has probably been a big bug for a while -- for example, the `FourTiTwo` tests were broken because running a 4ti2 binary resulted in linking errors because it couldn't find the 4ti2 libraries since the rpath pointed to a nonexistent directory!  (Note that this is a Linux-only solution.  I suppose it's potentially a problem in macOS, but I don't think it would matter for most cases since all these packages are available via Homebrew.  It's really only a problem in distros like RHEL that don't provide very many packages for things.)
* Modify the gfan, lrslib, and normaliz builds a bit to account for various flags that `rpmbuild` (in particular in RHEL 10) were throwing at it.


antonleykin (changes requested):
RPMs are probably all fine as well as the instructions
https://github.com/Macaulay2/M2/wiki/Installing-Macaulay2-in-Linux#red-hat-enterprise-linux
for those with root access.

However, on an RHEL 9.6 system with no root, I succeeded only after massaging instructions at https://macaulay2.com/Downloads/GNU-Linux/Red%20Hat%20Enterprise,%20CentOS,%20Scientific%20Linux/index.html
picking a relevant rpm from
https://macaulay2.com/Repositories/Scientific/

The instructions on macaulay2.com need to change... Perhaps,
* just point macaulay2.com to the wiki 
* add rootless instructions to the wiki

d-torrance:
I've updated the webpage to point to the wiki for both [Fedora](https://macaulay2.com/Downloads/GNU-Linux/Fedora/index.html) and [RHEL](https://macaulay2.com/Downloads/GNU-Linux/Red%20Hat%20Enterprise,%20CentOS,%20Scientific%20Linux/index.html).

For the rootless installation, I didn't have as much luck.  I successfully unpacked the M2 rpm into my home directory using `rpm2cpio`, but running it resulted in linking errors due to missing system libraries (BLAS in particular).  I suppose we could play the same game, download a BLAS rpm, unpack it with `rpm2cpio`, etc, and rinse and repeat until we have all the dependencies.  The `M2` script should set all the right environment variables to avoid all the linking errors, but it sounds like a mess!

A better solution, I think, for installing M2 on a system w/o root access would be spack.  I have a very early draft of a spack package that I'm hoping to finish before the November release.

antonleykin:
> For the rootless installation, I didn't have as much luck. I successfully unpacked the M2 rpm into my home directory using `rpm2cpio`, but running it resulted in linking errors due to missing system libraries (BLAS in particular). I suppose we could play the same game, download a BLAS rpm, unpack it with `rpm2cpio`, etc, and rinse and repeat until we have all the dependencies. The `M2` script should set all the right environment variables to avoid all the linking errors, but it sounds like a mess!

Perhaps, for now, it is still a good idea to mention on the wiki a (partial) recipe with`rpm2cpio` --- it worked for me on one computational server --- and also mention the caveats (one need luck with the libraries).

d-torrance:
Sounds good!  I've added the non-root instructions to the wiki: https://github.com/Macaulay2/M2/wiki/Installing-Macaulay2-in-Linux#installing-rpm-files-without-root-access

antonleykin (approved):

PR #3984, opened 2025-09-22
Bump macOS runners to sonoma (ventura bottles no longer built)

d-torrance:
I think the reason that the macOS builds have started failing is that they've stopped building bottles for macos-13 (https://github.com/Homebrew/brew/pull/20672).  So we bump the runners to macos-14.


d-torrance:
I'm going to go ahead and merge this (and cherry-pick onto `stable`) so we don't have so many failing builds

PR #3986, opened 2025-09-24
Look for for homebrew packages in the right places

d-torrance:
We add various homebrew package directories to the appropriate flags so that we can detect them at the beginning of the build without needing to set any variables ahead of time.  The cmake build was already doing a pretty good job of this (it was just missing openmp), and we bring autotools in line with it.

This is on top of #3984 so that the macOS builds will actually work.

Cc: @mikestillman

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -119,8 +119,6 @@ jobs:
                echo `brew --prefix make  `/libexec/gnubin >> $GITHUB_PATH
                echo `brew --prefix libtool`/libexec/gnubin >> $GITHUB_PATH
                echo `brew --prefix bison`/bin             >> $GITHUB_PATH
-               echo "CPPFLAGS=-I$(brew --prefix)/include -I$(brew --prefix libomp)/include -I$(brew --prefix readline)/include" >> $GITHUB_ENV
-               echo "LDFLAGS= -L$(brew --prefix)/lib     -L$(brew --prefix libomp)/lib     -L$(brew --prefix readline)/lib"     >> $GITHUB_ENV
           fi
 
       - uses: actions/cache@v5
--- a/M2/configure.ac
+++ b/M2/configure.ac
@@ -358,6 +358,31 @@ AC_SUBST(SIZEOF_INT_P,$ac_cv_sizeof_int_p)
 AC_CHECK_SIZEOF([long])
 AC_SUBST(SIZEOF_LONG,$ac_cv_sizeof_long)
 
+dnl Check for Homebrew packages
+
+AC_CHECK_PROGS([BREW], [brew], [false])
+
+dnl BREW_ADD_FLAG([FORMULA], [DIR], [FLAG], [PREFIX], [SEPARATOR])
+dnl adds PREFIX$(brew --prefix FORMULA)/DIR (if it exists) to FLAG,
+dnl separated by SEPARATOR
+AC_DEFUN([BREW_ADD_FLAG],
+    [AS_IF([test "x$BREW" = xbrew -a -d $($BREW --prefix $1)/$2],
+         [$3="${$3+$$3$5}$4$($BREW --prefix $1)/$2"])])
+
+dnl BREW_ADD_FLAGS([FORMULA])
+dnl adds:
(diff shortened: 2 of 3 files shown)


M2/cmake/check-libraries.cmake
```diff
@@ -58,6 +58,7 @@ check_include_files(boost/math/tools/atomic.hpp
 # TODO: replace gdbm, see https://github.com/Macaulay2/M2/issues/594
 find_package(GDBM	REQUIRED QUIET) # See FindGDBM.cmake
 
+list(APPEND CMAKE_PREFIX_PATH "$(brew --prefix libomp)")
```
mahrud: I didn't read what you're doing in the autotools build but this should certainly not be hardcoded. It's up to the user to add brew directories to the path or not, and further there's a reason keg-only bottles are keg-only! e.g this may break depending on whether you build with GCC or clang.

d-torrance: My goal was to make it simpler for the user with a reasonable set of default search paths so that the build should work out of the box most of the time on common systems without fiddling with flags.  And we're already hardcoding several other brew paths in the cmake build: https://github.com/Macaulay2/M2/commit/50ea57438295eba391c876da60be83a3e7cbe3ce

mahrud: Those are hints for standard libraries without multiple implementations, which are used only if a library wasn't found elsewhere on the path. Appending the prefix directory of libomp can determine whether the compiler uses libomp or libgomp.

d-torrance: Ok, that makes sense -- I'll remove the cmake changes

d-torrance: Side note -- do we even need to deal with OpenMP on most builds?  According to a comment in configure.ac:

> This is required for building the library csdp and good for building the library normaliz

Most of the time, we'll be using existing system packages for these.

mahrud: fflas-ffpack, which is a header-only library, uses it, and if it's not found correctly you get a warning when building M2:
> #warning "openmp was not detected correctly at configure time, please report this bug"

Fixing this on all homebrew builds was quite a headache, and I think at some point all cmake builds here were also working correctly but I see that the ubuntu one is failing now.

dimpase:
the stock AC_OPENMP does not work on maOS, as it doesn't test for the needed flag combo `-Xpreprocessor -fopenmp`. I have a prototype.

Fixing it alone is not sufficient - one needs a fixed libtool (2.6.1, the latest, does contain a fix I submitted a while ago)

d-torrance:
We do have our own version of `AC_OPENMP` -- see https://github.com/Macaulay2/M2/blob/stable/M2/m4/openmp.m4

dimpase:
that's a bit of misnaming, IMHO, to start your macro names with AC_

And no GNU libtool issues, as it's not used?

d-torrance:
> that's a bit of misnaming, IMHO, to start your macro names with AC_

Good point.  I just added a commit updating it to `M2_OPENMP`.

> And no GNU libtool issues, as it's not used?

Not that I've seen.  We don't use libtool to build M2 itself, but we do for many of its dependencies.  There might be some edge case where we build one of the dependencies that needs both libtool and openmp in macOS using some older libtool that doesn't have your patch.  But most of the time, we don't need to build dependencies in macOS anyway since they're available via Homebrew.  (The dependency builds are mostly for RHEL, where not much is packaged.)

dimpase (approved):
OK

antonleykin:
> The cmake build was already doing a pretty good job of this (it was just missing openmp)...

Is cmake still missing openmp?

d-torrance:
> Is cmake still missing openmp?

Out of the box on macOS, yes. It needs the correct incantation of flags to find the brew libomp

dimpase:
libomp lives in `prefix=HOMEBREW_PREFIX/opt/libopm`, so one needs to add $prefix/include to CPATH, 
$prefix/lib to LIBRARY_PATH, and link against libopm (i.e. -lomp)
And add `-Xpreprocessor -fopenmp` to CFLAGS and CXXFLAGS

antonleykin (approved):
This is exactly an alteration I needed to do on my mac machines :)

PR #3990, opened 2025-09-30
Add support for negative indices for ring variables & submatrices

d-torrance:
We count from the right when the index is negative, just like lists.

For example:

```m2
i1 : R = ZZ/32003[a..z];

i2 : R_-3

o2 = x

o2 : R

i3 : A = genericMatrix(R, 4, 4)

o3 = | a e i m |
     | b f j n |
     | c g k o |
     | d h l p |

             4      4
o3 : Matrix R  <-- R

i4 : A_{-3, -1}^{-2}

o4 = | g o |

             1      2
o4 : Matrix R  <-- R
```


d-torrance:
Discussed with Mike

Issue #3941, opened 2025-07-29
sqrt doesn't work over finite fields


mahrud:
```m2
i1 : sqrt(4_(ZZ/7))
stdio:1:4:(3): error: no method found for applying sqrt to:
                                ZZ
     argument   :  -3 (of class --)
                                 7
```
Note that flint even has [fmpz_mod_poly_sqrt](https://flintlib.org/doc/fmpz_mod_poly.html#c._fmpz_mod_poly_sqrt) for polynomials over integers mod n.

d-torrance:
Feature added in #3992

PR #3992, opened 2025-10-02
Add sqrt(ZZ, ZZ) method (and sqrt for ZZ/p and GF p)

d-torrance:
We use the [Tonelli-Shanks algorithm](https://en.wikipedia.org/wiki/Tonelli%E2%80%93Shanks_algorithm) to find solutions to $x^2\equiv n\pmod p$ for prime $p$.

```m2
i1 : sqrt(5, 41)

o1 = 13

i2 : sqrt(5_(ZZ/41))

o2 = 13

     ZZ
o2 : --
     41

i3 : sqrt(5_(GF 41))

o3 = 13

o3 : GF 41
```

This at least partially addresses #3941.  It's all implemented at top-level and doesn't use flint.  It also doesn't work modulo a composite number yet.


d-torrance:
Discussed with Mike

PR #3995, opened 2025-10-03
Add IFRAME class for <iframe> html tags

d-torrance:
<iframe> tags are used for inlining html documents inside html documents.  One use case is for using JavaScript in webapp mode.  For example, running the following will print "Hello, world!" to the Developer Tools console:

```m2
IFRAME("srcdoc" => "<script>console.log('Hello, world!')</script>")
```

I think this could open up lots of exciting possibilities for visualizing things in Macaulay2Web and Jupyter.

Cc: @pzinn

### DIFF
--- a/M2/Macaulay2/packages/Text.m2
+++ b/M2/Macaulay2/packages/Text.m2
@@ -9,7 +9,7 @@ newPackage("Text",
 
 exportFrom_Core {
      "ANCHOR", "BLOCKQUOTE", "BODY", "BOLD", "BR", "CDATA", "CODE", "COMMENT", "DD", "DIV", "DL", "DT", "EM", "ExampleItem", "HEAD", "HEADER1",
-     "HEADER2", "HEADER3", "HEADER4", "HEADER5", "HEADER6", "HR", "HREF", "HTML", "Hypertext", "HypertextContainer", "HypertextParagraph", "HypertextVoid", "IMG", "INDENT", "ITALIC", "KBD",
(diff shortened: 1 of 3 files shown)


M2/Macaulay2/packages/Text.m2
```diff
@@ -855,6 +855,17 @@ document {
...
+document {
+    Key => {IFRAME},
+    Headline => "HTML inline frame (iframe) element",
+    Usage => "IFRAME x",
+    Inputs => {"x" => List => "specifying the iframe attributes"},
+    PARA {
+	"An ", M2CODE "IFRAME", " object represents an inline frame in HTML, ",
+	"which allows embedding an independent browsing context (another HTML ",
+	"document) within the current one."},
+    EXAMPLE "html IFRAME {\"src\" => \"https://example.com\"}"}
```
pzinn: are you sure you want to include `html` in the example?

d-torrance: My idea was that this would at least show something interesting in standard mode.  Without `html`, we get nothing:

```m2
i2 : IFRAME {"src" => "https://example.com"}

o2 = 
```

Adding `html`:

```m2
i3 : html oo

o3 = <iframe src="https://example.com">
     </iframe>
```

But yeah -- in webapp mode, we get an honest-to-goodness iframe:

image

Most of the *Text* package docs don't have examples.  Some of the newer ones include both the `html` version and (outside an example, so it works in standard mode, too) the rendered version, like [KBD](https://www.macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Text/html/___K__B__D.html).  Maybe that would work?

d-torrance: I went ahead and updated it to include an actual IFRAME example:

image

pzinn (approved):

PR #3845, opened 2025-05-23
Various fixes, with a view toward the vscode extension

pzinn:
This branch has two purposes:
- allow any user to run Macaulay2 in Macaulay2Web in local mode (`npm start local`) or more importantly to use Macaulay2 in the newly developed [vscode extension](https://github.com/Macaulay2/vscode-Macaulay2/tree/experimental), with the same quality of rendering as with the forked branch (vanilla) used on the official Macaulay2Web server.
- in the process, I've found quite a few commits that fell through the cracks, so I'm putting them here in the hope to have the vanilla branch and the main M2 branch converge in the near future.

PR #3997, opened 2025-10-06
Add texMath(Type) to avoid recursion error

d-torrance:
This fixes a recursion error which I think may have been introduced in #3845 (when `texMath(Ring)` was split off from `texMath(HashTable)`):

```m2
i1 : texMath HashTable
stdio:2:7:(3): error: recursion limit of 300 exceeded
```

The issue is that types are mutable hash tables, and the `texMath` method for mutable hash tables calls `texMath` on the class, which is itself a mutable hash table, causing the infinite recursion.

### DIFF
--- a/M2/Macaulay2/m2/latex.m2
+++ b/M2/Macaulay2/m2/latex.m2
@@ -197,16 +197,14 @@ texMathMutable :=
 texMath MutableList  := L -> concatenate(texMath class L, "\\left\\{", if #L > 0 then "\\ldots "|#L|"\\ldots" else "\\,", "\\right\\}")
 
 texMath HashTable := H -> (
-    if isMutable H then texMathMutable H
-    else texMath class H | texMath apply(sortByName pairs H, (k, v) -> k => v)
+    texMath class H | texMath apply(sortByName pairs H, (k, v) -> k => v)
     )
 
-texMath RingFamily :=
-texMath Ring := R -> (
-    if R.?texMath then R.texMath
-    else if hasAttribute(R, ReverseDictionary) then texMath toString getAttribute(R, ReverseDictionary)
-    else (lookup(texMath,HashTable)) R -- should never happen
-    )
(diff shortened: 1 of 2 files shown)


pzinn:
ah yes this is a strange bug. on my branch I have
```
texMath HashTable := H -> if H.?texMath then H.texMath else (
    if hasAttribute(H, ReverseDictionary) then texMath toString getAttribute(H, ReverseDictionary)
    else if isMutable H then texMathMutable H
    else texMath class H | texMath apply(sortByName pairs H, (k, v) -> k => v))
```
note the extra 2nd line.
not sure why this didn't make it in #3845.
I think this is a better fix...

pzinn:
more precisely, here's the diff:
```
199,209c223,226
< texMath HashTable := H -> (
<     if isMutable H then texMathMutable H
<     else texMath class H | texMath apply(sortByName pairs H, (k, v) -> k => v)
<     )
< 
< texMath RingFamily :=
< texMath Ring := R -> (
<     if R.?texMath then R.texMath
<     else if hasAttribute(R, ReverseDictionary) then texMath toString getAttribute(R, ReverseDictionary)
<     else (lookup(texMath,HashTable)) R -- should never happen
<     )
---
> texMath HashTable := H -> if H.?texMath then H.texMath else (
>     if hasAttribute(H, ReverseDictionary) then texMath toString getAttribute(H, ReverseDictionary)
>     else if isMutable H then texMathMutable H
>     else texMath class H | texMath apply(sortByName pairs H, (k, v) -> k => v))
```
not sure why I decided to split `Ring/RingFamily` from `HashTable`... I think it's better to reunite them.

d-torrance:
I've pushed a new proposal, removing `texMath(Ring)` and replacing it with `texMath(MutableHashTable)`.  This also lets us simplify `texMath(HashTable)` since we can remove the `isMutable` check.

pzinn (approved):

PR #3999, opened 2025-10-09
Bump msolve to v0.9.2

d-torrance:
msolve 0.9.2 was released earlier today.  Draft for now to test the builds

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -177,6 +177,11 @@ jobs:
           export FC=gfortran-14
           ../../configure --with-system-gc --with-fplll
 
+      - name: Build libraries using Make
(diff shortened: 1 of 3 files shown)


d-torrance:
~~Why aren't the builds running?  :confused:~~

Now they are -- I'm just impatient, I guess!

d-torrance:
The builds all passed!

https://github.com/Macaulay2/M2/actions/runs/18395355407

I also took the opportunity the split "building libraries" into its own step in the autotools build just like we do with the cmake build.

antonleykin (approved):
My understanding that having libraries built in a separate step come out of a necessity for `cmake`. 
Nevertheless, I think, in general, it's good to have the two build processes be as similar as they can.

PR #4001, opened 2025-10-10
Add "pfaffian" method

d-torrance:
We already have `pfaffians` for computing an ideal containing all the sub-Pfaffians of a particular size, but we may just want the Pfaffian of a skew-symmetric matrix as a number/polynomial (similar to `determinant` v. `minors`).

```m2
i1 : R = QQ[a..f];

i2 : genericSkewMatrix(R, 4)

o2 = | 0  a  b  c |
     | -a 0  d  e |
     | -b -d 0  f |
     | -c -e -f 0 |

             4      4
o2 : Matrix R  <-- R

i3 : pfaffian oo

o3 = c*d - b*e + a*f

o3 : R
```


d-torrance:
Discussed with Mike

Issue #3905, opened 2025-07-01
Make Macaulay2Doc install target optional for all packages


kagurazaka-ayano:
I'm packaging M2 for nixos and I get it to build with CMake. But it takes forever to install Macaulay2Doc. I tried to build each package separately but some packages still installs target Macaulay2Doc.

I looked into [this issue](https://github.com/Macaulay2/M2/issues/3207) and took some of the works. But it still takes a lot of time.

Is there a way to not let packages build Macaulay2Doc? And also I want to build all package except that one so hardcoding the package list in the .nix file might not be a good option

Work so far:

```nix
let
  pkgs = import <nixpkgs> { };

  nativeBuildInputs = with pkgs; [
    singular
    gcc
    gnumake
    autoconf
    automake
    libtool
    yasm
    cmake
    ninja
    pkgconf
    blas
    boost
    gdbm
    tbb
    libffi
    readline
    gmp
    libxml2
    python313
    bison
    eigen
    texliveBasic
    gtest
    mpfr
    nauty
    ntl
    glpk
    mpsolve
    mpfi
    normaliz
    boehmgc
    lrs
    msolve
    csdp
    cddlib
    _4ti2
    gfan
    polymake
    flint
    fflas-ffpack
    givaro
  ];

  package_blacklist = builtins.toString [
    # documentation takes forever to build so we skip it
    "Macaulay2Doc"

  ];

  src = builtins.fetchGit {
    url = "git@github.com:Macaulay2/M2.git";
    rev = "6fe351dcfe733f2a525d361d8585f9f2375e264d";
    submodules = true;
  };
  cohomCalg = pkgs.stdenv.mkDerivation {
    pname = "cohomCalg";
    version = "0.32";
    nativeBuildInputs = with pkgs; [ gcc gnumake tree ];
    src = builtins.fetchTarball {
      url =
        "https://github.com/BenjaminJurke/cohomCalg/archive/refs/tags/v0.32.tar.gz";
    };
    buildPhase = "make -j20";
    installPhase = ''
      mkdir -p $out
      cp -r bin $out
    '';
  };
  TOPCOM = pkgs.stdenv.mkDerivation rec {
    version = "0.17.8";
    name = "topcom-${version}";
    builder = "${pkgs.bash}/bin/bash";
    args = [ build-script ];

    src = builtins.fetchTarball {
      url =
[...]

PR #4010, opened 2025-10-28
Build distribution tarballs

### LINKED ISSUE #644: Release tarball from github does not compile
Hi,
I've downloaded the realease tarball from here:
https://github.com/Macaulay2/M2/releases

During compile I run into:
make[1]: Leaving directory `/Users/tom/temp/buildM2/M2-release-1.10-stable-20170616/M2/BUILD/tom/libraries'
cd ../../.. && git submodule update --init M2/submodules/memtailor
fatal: Not a git repository (or any of the parent directories): .git
make: *** [../../submodules/memtailor/configure.ac] Error 128

I think this is because the build system expects some git information in its subdirectories, but github strips all .git directories when building the release tarball? (I could be wrong here...)

d-torrance:
This PR adds two new make targets, `make dist` and `make doc-dist`.

`make dist` builds an old-school source tarball from the M2 and M2-emacs git repos, and also includes a few generated autotools files so users can immediately run `configure` after unpacking without having to autoconf installed, similar to the `make dist` target in automake projects.  This should be useful for creating deb and rpm packages.  (I've been doing something similar on the Debian side for years.)

We also provide a backup for dealing with submodules when building from an unpacked tarball rather than a git repo.  This will simplify building rpm packages, where we currently have to trick the build system into thinking that there are submodules.

Also, the `make doc-dist` target creates a tarball containing all of the built documentation (html, info, and "dumps" of the gdbm databases).  Then, if someone runs `./configure --enable-documentation=download`, the build system will download this tarball from the webpage, unpack it, and use the prebuilt documentation rather than running all the examples from all the packages.  This will cut the M2 build time from hours down to minutes.  This should simplify the package installation process for users using "build from source" package managers (spack, gentoo, nix, etc.).  See #3905.

There is also a GitHub workflow to generate both tarballs.

### DIFF
--- a/.github/workflows/make-dist.yml
+++ b/.github/workflows/make-dist.yml
@@ -1,0 +1,30 @@
+name: Build distribution tarballs
+on: workflow_dispatch
+
+jobs:
+  make-dist:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+      - name: Install build dependencies
+        run: |
+          sudo add-apt-repository -y ppa:macaulay2/macaulay2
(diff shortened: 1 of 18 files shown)


d-torrance:
I'm going to go ahead and merge this so that it's in the release branch.  I need to make couple updates to the rpm builds to take advantage of some of the new features

Issue #997, opened 2019-09-04
submatrix keeps hash unchanged


mahrud:
Despite the fact that submatrix makes a copy of mutable matrices, hash of the two objects stays the same.

```
i1 : A = mutableMatrix {{10}}
o1 = | 10 |
o1 : MutableMatrix

i2 : B = submatrix(A,{0},{0})
o2 = | 10 |
o2 : MutableMatrix

i3 : A_(0,0) = 11
o3 = 11

i4 : B_(0,0) = 12
o4 = 12

i5 : A
o5 = | 11 |
o5 : MutableMatrix

i6 : B
o6 = | 12 |
o6 : MutableMatrix

i7 : hash A == hash B
o7 = true

i8 : hash A
o8 = 1527355229
```

DanGrayson:
It should be easy to make the submatrix routine produce a new hash code, the next one in the sequence.  It's probably doing a low-level copy operation that simply copies the hash code, too.

mahrud:
Does hash also depend on the location of the object in ram?

I wanted to use it to distinguish pointers to the same object. Is there a way to do that? (I understand that the location of an object can change, but it can be used to detect when two mutable objects are pointing to the same location)

DanGrayson:
> Does hash also depend on the location of the object in ram?

No.  One design principle used is that Macaulay2 should give the same answers on all machines.

> 
> I wanted to use it to distinguish pointers to the same object. Is there a way to do that? (I understand that the location of an object can change, but it can be used to detect when two mutable objects are pointing to the same location)

No, there is no way to do that.  It doesn't make much sense to want that either -- pointers to the same object have exactly the same behavior, and it's the object they point to that is what's important for us.

d-torrance:
Fixed by #4013

PR #4013, opened 2025-11-04
Use makeZeroMatrix, not copy constructor, for mutable submatrices

d-torrance:
This way, we end up calling the default constructor, which creates a new hash code.  Previously, we used the same hash code as the original matrix.

Also add a test to ensure that the hash changes.

Closes: #997 

### Before
```m2
i1 : A = mutableMatrix {{10}}

o1 = | 10 |

o1 : MutableMatrix

i2 : B = submatrix(A, {0}, {0})

o2 = | 10 |

o2 : MutableMatrix

i3 : (hash A, hash B)

o3 = (1845190296659672920, 1845190296659672920)

o3 : Sequence
```

### After
```m2
i1 : A = mutableMatrix {{10}}

o1 = | 10 |

o1 : MutableMatrix

i2 : B = submatrix(A, {0}, {0})

o2 = | 10 |

o2 : MutableMatrix

i3 : (hash A, hash B)

o3 = (2554347540320848225, 2672540414264377250)
```

### DIFF
--- a/M2/Macaulay2/e/mutablemat-defs.hpp
+++ b/M2/Macaulay2/e/mutablemat-defs.hpp
@@ -518,8 +518,7 @@ class MutableMat : public MutableMatrix
                 n_cols() - 1);
           return 0;
         }
-    MutableMat* result =
-        new MutableMat(*this);  // zero matrix, over the same ring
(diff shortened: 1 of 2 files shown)


d-torrance:
Discussed with Mike -- will merge after fixing conflict

Issue #3928, opened 2025-07-11
augmented assignment operators should lock the left hand side


mahrud:
Currently `H.cache.blah ??= foo` works as follows:
1. lock `H.cache`
2. attempt to read `H.cache.blah`
3. unlock `H.cache`
4. if successful, return value read in step 2
5. if it failed:
6. evaluate foo
7. lock `H.cache`
8. write `H.cache.blah = foo`
9. unlock `H.cache`

Problem: if `H.cache` is updated between steps 3 and 7, then the value is corrupted. This is also the case for any other augmented assignment operator, e.g. `+=`. Probably the solution is to lock the left hand side (which isn't always a hash table), but given the code it's unclear how to do this.

I mentioned this [here](https://github.com/Macaulay2/M2/issues/175#issuecomment-3037627180), but it's becoming very critical, so wanted to make it its own issue.

mahrud:
@d-torrance is this fixable in the short term, even if the solution is not ideal?

d-torrance:
Implemented in #4017

PR #4017, opened 2025-11-07
Lock hash tables during augmented assignment

d-torrance:
This is the beginnings of an attempt at fixing #3928.

Here's the idea:  We'd like something like `x.a += 1` to be atomic when `x` is a mutable hash table.  So at the beginning of augmented assignment, we look to see if the left hand side is binary code with `.` or `#` as its binary operator, and then check if the left hand side of *that* is a hash table.  If so, go ahead and lock it, and stuff the hash table in a variable so we know how to unlock it when we're done.

The big problem, however, is that we currently implement hash table mutexes with a rwlock, which isn't recursive.  In particular, if we lock the mutex, we can't lock it again in the same thread or we deadlock.  But we're definitely going to lock it at least one more time when we call `lookup` to get the value at the given key.

So in the current draft, M2 immediately aborts because it runs into this problem on startup.

A possible idea would be to switch from rwlock's to recursive mutexes for hash tables.  But hash tables are so fundamental in M2 that I'd rather wait until after the release to try this to give us plenty of time to test any repercussions.

Cc: @mahrud

### DIFF
--- a/M2/Macaulay2/d/evaluate.d
+++ b/M2/Macaulay2/d/evaluate.d
@@ -1401,6 +1401,29 @@ augmentedParallelAssignmentFun(oper:Symbol, lhs:CodeSequence, rhs:Code):Expr:= (
 	    else if n == 1 then result.0 else Expr(result)))
     else ParallelAssignmentError(n));
 
+HashTableOrNull := HashTable or null;
+
+-- check if code is x#k or x.k (where x is a hash table)
+-- if so, lock it and return (so we can unlock it later)
+maybeLock(c:Code):HashTableOrNull := (
+    when c
+    is x:binaryCode do (
+	if x.f == DotS.symbol.binary || x.f == SharpS.symbol.binary
+	then (
+	    y := eval(x.lhs);
+	    when y is z:HashTable do (
+		if !z.beingInitialized then lockWrite(z.mutex);
+		HashTableOrNull(z))
+	    else HashTableOrNull(null()))
+	else HashTableOrNull(null()))
+    else HashTableOrNull(null()));
+
+maybeUnlock(x:HashTableOrNull):void := (
+    when x
+    is y:HashTable do (
+	if !y.beingInitialized then unlock(y.mutex))
+    else nothing);
+
 augmentedAssignmentFun(x:augmentedAssignmentCode):Expr := (
     when lookup(x.oper.word, augmentedAssignmentOperatorTable)
     is null do buildErrorPacket("unknown augmented assignment operator")
@@ -1416,6 +1439,8 @@ augmentedAssignmentFun(x:augmentedAssignmentCode):Expr := (
 	is y:angleBarListCode do (
 	    return augmentedParallelAssignmentFun(x.oper, y.t, x.rhs))
 	else nothing;
+	-- check if we're modifying a hash table and lock if so
+	table := maybeLock(x.lhs);
 	-- evaluate the left-hand side first
 	lexpr := nullE;
 	if s.word.name === "??" -- x ??= y is treated like x ?? (x = y)
@@ -1423,20 +1448,32 @@ augmentedAssignmentFun(x:augmentedAssignmentCode):Expr := (
 	    e := nullify(x.lhs);
 	    when e
 	    is Nothing do nothing
-	    else return e)
+	    else (
+		maybeUnlock(table);
+		return e))
 	else lexpr = eval(x.lhs);
-	when lexpr is e:Error do return lexpr else nothing;
+	when lexpr is e:Error do (
+	    maybeUnlock(table);
+	    return lexpr)
+	else nothing;
 	-- check if user-defined method exists
 	meth := lookup(Class(lexpr), Expr(SymbolClosure(globalFrame, x.oper)));
 	if meth != nullE then (
 	    rexpr := eval(x.rhs);
-	    when rexpr is e:Error do return rexpr else nothing;
+	    when rexpr is e:Error do (
+		maybeUnlock(table);
+		return rexpr)
+	    else nothing;
 	    r := applyEEE(meth, lexpr, rexpr);
 	    when r
 	    is s:SymbolClosure do (
 		if s.symbol.word.name === "Default" then nothing
-		else return r)
-	    else return r);
+		else (
+		    maybeUnlock(table);
+		    return r))
+	    else (
+		maybeUnlock(table);
+		return r));
 	-- if not, use default behavior
 	c := (
 	    if s.word.name === "??" then x.rhs
@@ -1458,7 +1495,10 @@ augmentedAssignmentFun(x:augmentedAssignmentCode):Expr := (
 	    else globalAssignment(y.var.frameindex, y.var, r))
 	is y:binaryCode do (
 	    if y.oper == DotS.symbol || y.oper == SharpS.symbol
-	    then AssignElemFun(y.lhs, y.rhs, c)
+	    then (
+		z := AssignElemFun(y.lhs, y.rhs, r);
+		maybeUnlock(table);
+		z)
 	    else InstallValueFun(CodeSequence(
 		    convertGlobalOperator(y.oper), y.lhs, y.rhs, c)))
 	is y:adjacentCode do (
--- a/M2/Macaulay2/d/evaluate.d
+++ b/M2/Macaulay2/d/evaluate.d
@@ -1424,6 +1424,55 @@ maybeUnlock(x:HashTableOrNull):void := (
 	if !y.beingInitialized then unlock(y.mutex))
     else nothing);
 
+--what about mutable lists?
+--assumes that the table has been locked on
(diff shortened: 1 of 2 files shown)


jkyang92:
I feel like it would be cleaner to handle `x#k op= y` and `x.k op= y` separately since we're already matching on the structure of the left hand side. Then we can have a function that looks something like `ModifyAndAssign(x,k,op,y)` that does the actual operation on a hash table. This function can lock the table once, lookup the value and modify it. That way we aren't depending on the call to `eval(x.lhs)` to do the lookup.

The think I can't remember is if we could have a circumstance where "#" or "." is overridden with something weird.

mahrud:
I don't think # or . can be overridden at top-level (yet!) but I consider this to be higher priority than allowing top-level to override. This is extremely troublesome for me right now so I'll take any solution that works!

d-torrance:
. and # can't be overridden. I used to not like that, but now I kind of do. We have other operators (_, and I use @@ in the Python package for .) that we can use, keeping these two nice and low level for getting at the underlying list or hash table.

I like Jay's idea. I'll play around with it soon

mahrud:
(I deeply dislike @@ btw, you should just move the relevant types to the interpreter and use . I think)

jkyang92:
Also, I didn't mention it earlier, you have to make sure to evaluate the right hand side before locking on the table. This is a bit subtle about order of operations, but if the right hand side also uses the same hash table (which isn't crazy) then you would have the same recursive lock issue.

I think this makes `x += y` slightly different semantically from `x = x + y`. In particular if both the expressions x and y produce errors, which error do we get? Right now if x is a mutable hash table and 1 is not a key, then `x#1 += error "test"` produces `error: key not found in hash table`, but using the suggestion it would produce `error: test`

This is a compromise that I think is fine, but I want to point this out.

In theory a custom operator can make this even worse if it looks into the table somehow, but that shouldn't happen with sensible operators.

mahrud:
@d-torrance I can't help until after the semester but any luck with this?

d-torrance:
I haven't taken a stab at it yet, but hopefully soon!

mahrud:
@jkyang92 in principle I agree with everything in your comment, but I struggle to see a clean way to implement it. Do you have a any suggestions?

jkyang92:
@mahrud 

Inside of the `s:Symbol do (...)` block inside of `augmentedAssignmentFun`, instead of first evaluating the left hand side, you can check if `x.lhs` is a reference into a hash table (or list) (`binaryCode` with a operator of `DotS` or `SharpS`), if it is, call a separate function that handles the case of hash tables, otherwise continue with the current code (not the code in the pull request). This check is essentially what `maybeLock` in this pull request is currently doing, but instead of conditionally locking, we just take a different code path.

In the hash table case, you can then:
1. Evaluate the right hand side
2. Lock the hash table
3. Get the value from the hash table
4. Apply the operator to the entry
5. Write the entry back to the hash table
6. Unlock the hash table

The only danger with this is if the operator has a user defined method, it could in theory relock the hash table in step 4.

You could also add a bit of code before step 1 to take a read lock on the hash table and check if the entry even exists before proceeding, I don't know that this expense is worth it.

jkyang92:
Another approach that would have fewer worries about deadlock/relocking would be to "black hole" the entry in the hash table when we are modifying it. This entails putting a placeholder entry into the hash table that indicates "this entry is being modified". The black hole then should indicate which thread is currently modifying it, and have appropriate condition variables/mutexes such that other threads can wait on it. (If creating one per black hole seems expensive you can do one per thread). It's then an error (which can be sent back to the user) if you try to modify (but not read?) an entry which the current thread has black holed, but any other thread will just wait on the black hole. To avoid deadlock, you'd need special handling of waiting on black holed entries while having created black holed entries, but as long as you allow reads through the black hole, this is not a huge issue. The advantages is that this allows not holding the lock while doing the actual operation and only when actually writing the entry back to the hash table (you still have to take a write lock to add the black hole).

There are complicated questions about what sequencing of operations this corresponds to, and annoyingly, it requires a check for black holes at every write into the hash table.

mahrud:
This is a funny compiler warning:
```cpp
[45/59] Building C object Macaulay2/bin/CMakeFiles/M2-binary.dir/startup.c.o
/home/mahrud/Projects/M2/research/M2/BUILD/build/Macaulay2/bin/startup.c:196:58: warning: trigraph ‘??=’ ignored, use ‘-trigraphs’ to enable [-Wtrigraphs]
  196 |    "     setAttribute = (val,attr,x) -> ( Attributes#val ??= new MutableHashTable )#attr = x;\n"
```
Shouldn't the compiler know to ignore a hardcoded string?

ps: I actually don't think it's related to this PR but somehow this is the first I'm seeing this.

jkyang92:
@mahrud It occurs very early in the processing, even before tokenization. so the compiler can't tell if it's in a string. If we want to suppress the warning, we can split the string into two (using the fact that neighboring string literals get concatenated). Also, I checked and trigraphs have been removed in C++17 and C23.

mikestillman:
We are compiling with c++17, I thought?

jkyang92:
It's because its a C file and we're certainly not compiling with C23.

mahrud:
This seemed to solve most of my problems. Could we merge this before the release?

d-torrance:
Hmm, something's really slowing down the Ubuntu builds.  For example, from the `cmake-ubuntu-24.04-default` build on this PR:

```m2
 -- making example results for "detectCongruence(GushelMukaiFourfold,ZZ)"    -- 38.1809s elapsed
```

But from the same build for #4212:

```m2
-- making example results for "detectCongruence(GushelMukaiFourfold,ZZ)"    -- .635978s elapsed
```

The captured example results seem to take roughly the same amount of time, so my guess is that the holdup is in firing up a new M2 process.  Possibly all the extra locking and unlocking of mutexes whenever `??=` is called at startup?  We had a similar slowdown with `lastError` last year.

mahrud:
Currently the cmake-ubuntu build is stuck nearly at the end, after `[292/293] Installing package GameTheory`, and the autotools build is also stuck after GameTheory.

jkyang92:
I wouldn't be surprised if the current implementation is slow. I can try to actually implement a slightly less hacked together implementation (as described in the second paragraph here: https://github.com/d-torrance/M2/pull/21).

mahrud:
I actually can't replicate the slowdown, and have done far more extensive parallel computations.

d-torrance:
And the macOS builds were nice and quick, too.  I wonder what's different about the Ubuntu runners?

d-torrance:
Awesome -- this works!  Thanks, @jkyang92!

### Before

```m2
i1 : x = new MutableHashTable; x.a = 0

o2 = 0

i3 : scan(10000, i -> schedule(() -> x.a += 1))

i4 : x.a

o4 = 6826
```

### After

```m2
i1 : x = new MutableHashTable; x.a = 0

o2 = 0

i3 : scan(10000, i -> schedule(() -> x.a += 1))

i4 : x.a

o4 = 10000
```

d-torrance:
Hmm, ran into a problem trying to access the hash table we're modifying on the RHS of the assignment:

```m2
i1 : x = new MutableHashTable; x.a = 0; x.b = 1

o3 = 1

i4 : scan(10000, i -> schedule(() -> x.a += x.b))

i5 : x.a

o5 = 10000

i6 : scan(10000, i -> schedule(() -> x.a += x.b))
stdio:4:36:(3):[1]: error: interrupted
Aborted (core dumped)
```

mahrud:
Seems to happen after a few runs:
```m2
i5 : x = new MutableHashTable; x.a = 0; x.b = 1; scan(10000, i -> schedule(() -> x.a += x.b))

i9 : x = new MutableHashTable; x.a = 0; x.b = 1; scan(10000, i -> schedule(() -> x.a += x.b))

i13 : x = new MutableHashTable; x.a = 0; x.b = 1; scan(10000, i -> schedule(() -> x.a += x.b))
Aborted                    (core dumped) ./M2
```
Or even in one run for large enough counts:
```m2
i1 : x = new MutableHashTable; x.a = 0; x.b = 1; scan(100000, i -> schedule(() -> x.a += x.b))
Aborted                    (core dumped) ./M2
```

I really want to be able to debug issues like this with gdb/lldb (or their vscode frontends), but the tasks system makes it really convoluted. @jkyang92 @d-torrance do either of you have nice ways of doing this?

jkyang92:
I'm a bit confused what's happening. I can fix the abort, and there was an edge case where the code wasn't unlocking the hash table correctly. But that was due to not handling the case of an error correctly. But the error we are getting is that `eval` was interrupted (`error: interrupted`), which I don't know why that would be happening.

jkyang92:
I figured it out. If you don't keep a reference to the task, the the garbage collector eventually collects the task, and on collection the task is canceled, causing the interrupt. This works great if the output of the task is what matters, it's not so great when the task has side-effects that matter. I sent @d-torrance a pull request with the fix of the error edge case

PR #4018, opened 2025-11-07
add aliases for methods with long names

mgacummings:
Added `link` and `deletion` as aliases for `oneStepGVDCyI` and `oneStepGVDNyI` [for Patricia Klein]

### DIFF
--- a/M2/Macaulay2/packages/GeometricDecomposability.m2
+++ b/M2/Macaulay2/packages/GeometricDecomposability.m2
@@ -2,8 +2,8 @@
 
 newPackage(
         "GeometricDecomposability",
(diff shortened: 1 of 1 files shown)


d-torrance:
A simpler option would be to pass options to `export`, i.e., something like:

```m2
export {
    ...
    "oneStepGVD",
    "link" => "oneStepGVD",
    "oneStepGVDNyI",
    "deletion" => "oneStepGVDNyI",
    ...
}   
```

mgacummings:
That's very slick, thanks!

d-torrance (approved):
Thanks!

PR #4019, opened 2025-11-08
update ToricTopology, add moment-angle complex

feynhat:
This PR adds the following to the ToricTopology package

- a new class MomentAngleComplex,
- associated methods to compute
- - equivariant cohomology
- - Betti numbers
- - Euler characteristic
- documentation and examples illustrating these features.

Minor fixes to the rest of the ToricTopology package:
- replaces the variable name 'chi' with 'lambda' as it conflicts with the 'chi' method function defined in core (moreover, λ is more commonly used to denote the characteristic map in toric topology)
- fix bettiQTM so that it doesn't skip the 0th Betti number
- remove trailing whitespaces

### DIFF
--- a/M2/Macaulay2/packages/ToricTopology.m2
+++ b/M2/Macaulay2/packages/ToricTopology.m2
@@ -469,7 +469,7 @@ doc ///
 			products of disks and circles: $$\mathcal{Z_K} = \bigcup_{\sigma \in K}
 			\left( (D^2)^\sigma \times (S^1)^{[m] \setminus \sigma} \right).$$ These
 			spaces admit a natural action of the torus $T^m = (S^1)^m$. Non-singular
-			toric varities (not necessarily complete) are homotopy equivalent to
+			toric varieties (not necessarily complete) are homotopy equivalent to
 			partial quotients of moment-angle complexes by freely acting subtori of
 			$T^m$. Thus, moment-angle complexes are an important class of spaces
 			studied in Toric Topology. Their topological properties can be determined
@@ -587,7 +587,7 @@ doc ///
 	Description
 		Text
 			Compute the equivariant cohomology of a moment-angle complex over the
-			polynomial ring k[x_1, ..., x_m] where k is the coeffcient ring of the
+			polynomial ring k[x_1, ..., x_m] where k is the coefficient ring of the
 			polynomial ring over which the underlying simplicial complex was
 			created.
 		Text
--- a/M2/Macaulay2/packages/ToricTopology.m2
+++ b/M2/Macaulay2/packages/ToricTopology.m2
@@ -21,16 +21,17 @@
 
 newPackage(
 	"ToricTopology",
-    	Version => "1.0",
-    	Date => "September 1, 2015",
-    	Authors => {
-    		{Name => "Alvise Trevisan", Email => "a.trevisan@enpicom.com", HomePage => "http://www.enpicom.com"},
-    		{Name => "Alexander I. Suciu", Email => "a.suciu@neu.edu"}
-		},
-        Keywords => {"Toric Geometry"},
-     	PackageImports => { "OldChainComplexes", "SimplicialComplexes" },
-    	Headline => "toric topology"
-    	)
+	Version => "1.1",
+	Date => "November 7, 2025",
+	Authors => {
+		{Name => "Alvise Trevisan", Email => "a.trevisan@enpicom.com", HomePage => "http://www.enpicom.com"},
+		{Name => "Alexander I. Suciu", Email => "a.suciu@neu.edu"},
+		{Name => "Kumar Sannidhya Shukla", Email => "kshukla5@uwo.ca"}
+	},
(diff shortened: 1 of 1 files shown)


M2/Macaulay2/packages/ToricTopology.m2
```diff
@@ -336,52 +347,97 @@ permutahedronDual = (n) -> (
...
 	);
 	simplicialIntToMon(simplices)
 )
 
+-- methods involving moment-angle complexes
+
+-- equivariant cohomology module of the moment-angle complex wrt. T^m-action
+equivariantCohomology = method()
+equivariantCohomology(MomentAngleComplex) := Module => (mac) -> (
+	return coker gens monomialIdeal mac.MACSimplicialComplex;
```
d-torrance: `return` isn't necessary on the last line of a function, provided there's no semicolon.

M2/Macaulay2/packages/ToricTopology.m2
```diff
@@ -336,52 +347,97 @@ permutahedronDual = (n) -> (
...
+			b += btally#key;
+		);
+	);
+	b
+)
+
+-- all the betti numbers up to 2m of a MAC over a complex with m vertices
+bettiMAC(MomentAngleComplex) := List => (mac) -> (
+	b := {};
+	for i in 0..(2*#vertices mac.MACSimplicialComplex) do b=append(b, bettiMAC(i,mac));
```
d-torrance: Calling `append` inside a loop like this isn't efficient as we create a brand new list each time.  This could be simplified using `for ... list` or `apply`, e.g., something like:

```m2
apply(2*#vertices mac.MACSimplicialComplex + 1, i -> bettiMAC(i, mac))
```

M2/Macaulay2/packages/ToricTopology.m2
```diff
@@ -21,12 +21,12 @@
 
 newPackage(
 	"ToricTopology",
-    	Version => "1.0", 
+    	Version => "1.0",
```
d-torrance: It would be good to increment the `Version`, update the `Date`, and add yourself to `Authors`.

This package hasn't seen any significant updates in 10 years, but tagging the original author (@alvisetrevisan) for comments.

d-torrance (changes requested):
Thanks for the contribution!  Please see the comments.

d-torrance:
I'm confused by the macOS test failures.  They certainly don't have anything to do with your changes, but they're not showing up in any of the other PR's...

Also, it would be good to add some tests to the package.  See https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/___T__E__S__T.html.

d-torrance:
Ah, I think I figured out the macOS test failures:

```
dyld[13575]: Library not loaded: /opt/homebrew/opt/ntl/lib/libntl.44.dylib
  Referenced from: <1F5C7121-7AB3-33D2-8325-99AA86941686> /opt/homebrew/Cellar/factory/4.4.1/lib/libfactory-4.4.1.dylib
```

We probably need to rebuild our factory bottle against the new version of ntl.  Again, totally unrelated to this PR.

M2/Macaulay2/packages/ToricTopology.m2
feynhat: Thanks for the review. I have now updated the code.

M2/Macaulay2/packages/ToricTopology.m2
feynhat: This syntax was already used in several parts of the older code. I’ve fixed all the new instances I added and a few of the old ones too. I left the rest unchanged since I’m not yet familiar enough with those sections.

M2/Macaulay2/packages/ToricTopology.m2
feynhat: Done. Thank you for the suggestion.

feynhat:
> Also, it would be good to add some tests to the package. See https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/___T__E__S__T.html.

@d-torrance I am writing tests and will push them in a separate commit. Can I load other packages in the tests? This package is heavily dependent on the SimplicialComplexes package, I am not sure if it is possible to do a lot of testing without it. I've tried putting `needsPackage` in my `TEST ///` blocks but that fails.

d-torrance:
You should be able to call `needsPackage` in tests.  What failure are you getting?

Alternatively, it might make sense to move SimplicialComplexes to a `PackageExports` option in `newPackage`.  Then it will be available automatically without having to load it.

M2/Macaulay2/packages/ToricTopology.m2
feynhat: I changed all such occurrences.

feynhat:
I have added the tests now. Thanks.

d-torrance (approved):
Looks good -- thank you!

PR #4025, opened 2025-11-09
Bump Python package to v1.0

d-torrance:
Significant updates to the Python package.  Including:

* New `PythonContext` class replacing the undocumented Context class.
* `getattr`, `hasattr`, and `settattr` removed (use `@@` and `@@?` now)
* `getitem` and `setitem` removed (use _ now)
* NumPy support added
* New `pythonRunScript` for running multiline Python code
* Add support for working with virtual environments and installing modules using pip.
* Significant documentation improvements, including two tutorials
* Stop shipping cached examples since Python is supported by default now


d-torrance:
Discussed with Anton and Mike

PR #4026, opened 2025-11-09
Bump eigen to v5.0.1

d-torrance:
Eigen 5.0.1 was just released.

~~We also drop the required version when checking for eigen in cmake.  The problem is that Eigen 5's cmake config files see a single version number and expect it to match exactly.  Using a range would fix this for Eigen 5, but Eigen 3's cmake config files expect the major version number to match when a range is used, so we can't use something like "3.3...5".~~

### DIFF
--- a/M2/cmake/build-libraries.cmake
+++ b/M2/cmake/build-libraries.cmake
@@ -147,8 +147,8 @@ else()
   set(EIGEN_BUILD_TYPE Release)
 endif()
 ExternalProject_Add(build-eigen
-  URL               https://gitlab.com/libeigen/eigen/-/archive/5.0.0/eigen-5.0.0.tar.bz2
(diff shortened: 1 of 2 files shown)


d-torrance:
Builds passed!  https://github.com/Macaulay2/M2/actions/runs/19208646503

d-torrance:
Hmm, `find_package` still isn't finding Eigen 5.0.0 even after I dropped the version number:

```
## External components
     Need to build:
       Libraries       = eigen
       Programs        = N/A
     Already built:
       Libraries       = N/A
       Programs        = N/A
```

I think I'll drop that commit for now and deal with that issue in another PR

antonleykin (commented):
Looks fine. Let's come back to modifying cmake later.

PR #4033, opened 2025-11-17
Use macos-15 for GitHub builds

d-torrance:
The macos-latest runners migrated to sequoia a few months ago (https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/).  Now that tahoe has been released, we're currently testing two versions ago.


antonleykin (commented):
All seems fine.

PR #4034, opened 2025-11-17
Install scip during GitHub builds

d-torrance:
macOS only -- available in brew but not apt

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -65,7 +65,7 @@ jobs:
           brew update
           brew tap macaulay2/tap
           brew install --overwrite python
-          brew install automake bison boost libtool tbb ccache ctags llvm make ninja yasm libffi msolve googletest fplll eigen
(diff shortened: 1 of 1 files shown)


d-torrance:
@mikestillman: It looks like the problem is that we're assuming that scip supports the zimpl reader:

```
no reader for input file </tmp/M2-10537-0/31/codim.zpl> available
```

But the [brew scip formula](https://github.com/Homebrew/homebrew-core/blob/13f56926e27c4a1a11295bd5b70dc3798f151561/Formula/s/scip.rb) turns off zimpl support:
```ruby
 system "cmake", "-S", ".", "-B", "build", "-DZIMPL=OFF", *std_cmake_args
```

So I guess we need to use a different format if we want the brew scip to work?

mikestillman:
I'll take a look and see if changing the format will be easy.  It would be good to use built-in binaries...

d-torrance:
I'm going to close the PR (which was just an experiment) and open a corresponding issue

PR #4036, opened 2025-11-18
Update testbot dockerfile

d-torrance:
Two changes:
* bookworm -> trixie (released 8/25)
* Use new M2 Debian repo

[ci skip]

### DIFF
--- a/M2/BUILD/docker/testbot/Dockerfile
+++ b/M2/BUILD/docker/testbot/Dockerfile
@@ -1,16 +1,16 @@
 # Time usage: <5min
 # Net usage:  <200MB
-# Disk usage: <800MB
+# Disk usage: <850MB
 
-FROM debian:bookworm
+FROM debian:trixie
 
 LABEL org.opencontainers.image.source=https://github.com/Macaulay2/M2
 
 # Setup the system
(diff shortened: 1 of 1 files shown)


M2/BUILD/docker/testbot/Dockerfile
```diff
@@ -2,15 +2,15 @@
 # Net usage:  <200MB
 # Disk usage: <800MB
```
mahrud: Are these stats roughly the same?

d-torrance: Disk usage has increased a bit:

```
$ docker image ls
REPOSITORY                                   TAG            IMAGE ID       CREATED         SIZE
macaulay2                                    latest         72d02acc9e80   4 minutes ago   846MB
```

What's the best way to determine the net usage?

mahrud: I just compare RX bytes in ifconfig before and after ¯\_ (ツ)_/¯

d-torrance: Ok, cool.  It was pretty much the same (177 MB).  I'll update the disk usage since it went up a bit

mahrud (approved):

PR #4039, opened 2025-11-24
Allow Normaliz to work with Rational Polytopes

ollieclarke8787:
Modified the Normaliz package to work with rational coordinates - Macaulay2 can concatenate String | ZZ but cannot concatenate String | QQ


d-torrance:
Thanks!  Would you add a test with a rational polytope that would have failed before the fix?

ollieclarke8787:
Sure. I've added a non-integral rational example as a test (which doesn't work with the current Normaliz) and changed the documentation in `writeNmzData` to reflect that rational polytopes also work.

d-torrance (approved):
Thank you!

PR #4045, opened 2025-12-01
Fixed Documentation of 'InverseSystems' and 'inverseSystem' from InverseSystem

### LINKED ISSUE #4038: Documentation of inverseSystem
There is some conflicting information in the documentation of the package `InverseSystems`. In the documentation of the method function `inverseSystem`, there is this caveat:

```
 Caveat
       ======

       Because inverseSystem involves a conversion between the bases of the dual, it should not be used in the
       default mode unless the characteristic is greater than the highest degree to which a variable appears. To make
       $x^a$ represent $x^(a)$, for example in small characteristics use

       inverseSystem(Matrix, DividedPowers=>false)
       (which was the default behavior of the old script "fromDual").
```
I believe the correct option is setting `DividedPowers=>true` instead, according to the main documentation of the package. There are also some minor LaTeX typos on this page.

Jazengm:
Resolves #4038. The pseudocode starting with `contract` was also fixed to better match M2 syntax.

### DIFF
--- a/M2/Macaulay2/packages/InverseSystems.m2
+++ b/M2/Macaulay2/packages/InverseSystems.m2
@@ -379,7 +379,7 @@ Description
   Method: To represent finitely generated S-submodule of $D'$ 
   as an S-module we use the map of modules
   S/(x_1^d,\dots, x_n^d) -> D' sending $x^a$ to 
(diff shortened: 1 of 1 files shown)


d-torrance (approved):
Thank you so much!

PR #4046, opened 2025-12-01
Add polylog function

d-torrance:
This is a draft for now -- we should wait for #4040 and add CC/CCi versions, too.

~~Question:  Should we use `Li` instead of `polylog`?  That would match the common notation, but is it too short?~~  *Edit:* `Li` is a synonym for `polylog`.

Cc: @mikestillman


d-torrance:
Discussed with Mike

mahrud:
Please remove `Li` as a synonym for `polylog`, this is too disruptive for seemingly little benefit. There are many other common 2 character notations in math, like Im, Re, etc but we don't define any of those, and for good reason.

PR #4047, opened 2025-12-02
Add caughtError variable

d-torrance:
This is a simple alternative to `lastError` (see #3654).  Rather than setting a variable every single time we raise an error (which ended up slowing down Macaulay2 considerably, hence #3912), we set it only during the `else` clause of a `try` statement.

For example:
```m2
i1 : try 1/0 else caughtError

o1 = (stdio:1:4-1:7, division by zero)

o1 : Sequence
```

As a proof of concept, we use it in the *JSONRPC* package for improved error messages:

```m2
i1 : needsPackage "JSONRPC";

i2 : server = new JSONRPCServer;

i3 : registerMethod(server, "divide", (x,y) -> x/y);

i4 : handleRequest(server, makeRequest("divide", {1, 0}, 1))

o4 = {"error": {"code": -32603, "message": "Internal error: division by zero"},
     "jsonrpc": "2.0", "id": 1}
```

### DIFF
--- a/M2/Macaulay2/d/evaluate.d
+++ b/M2/Macaulay2/d/evaluate.d
@@ -36,6 +36,8 @@ export applyIteratorS := setupvar("applyIterator", nullE);
 export joinIteratorsS := setupvar("joinIterators", nullE);
 export pairsIteratorS := setupvar("pairsIterator", nullE);
 
+caughtErrorS := setupvarThread("caughtError", nullE);
+
 eval(c:Code):Expr;
 applyEE(f:Expr,e:Expr):Expr;
 export evalAllButTail(c:Code):Code := while true do c = (
@@ -1481,7 +1483,14 @@ export evalraw(c:Code):Expr := (
 	      if tryEvalSuccess then
 	      when ret is Error do ret
 	      else if c.thenClause == NullCode then ret   else eval(c.thenClause)
-	      else if c.elseClause == NullCode then nullE else eval(c.elseClause))
+	      else if c.elseClause == NullCode then nullE else (
+		  when ret
+		  is err:Error do setGlobalVariable(caughtErrorS,
+		      seq(locate(err.position), toExpr(err.message)))
+		  else nothing; -- shouldn't happen
+		  p := eval(c.elseClause);
+		  setGlobalVariable(caughtErrorS, nullE);
+		  p))
 	  is c:catchCode do (
 	       p := eval(c.code);
 	       when p is err:Error do if err.message == throwMessage then err.value else p
--- a/M2/Macaulay2/m2/exports.m2
+++ b/M2/Macaulay2/m2/exports.m2
@@ -532,6 +532,7 @@ export {
 	"cancelTask",
 	"capture",
 	"catch",
+	"caughtError",
 	"ceiling",
 	"centerString",
 	"changeBase",
--- a/M2/Macaulay2/packages/JSONRPC.m2
+++ b/M2/Macaulay2/packages/JSONRPC.m2
@@ -16,14 +16,28 @@
 
 newPackage("JSONRPC",
     Headline => "JSON-RPC server",
-    Version => "0.1",
-    Date => "May 19, 2025",
+    Version => "0.2",
+    Date => "December 1, 2025",
     Authors => {{
 	    Name => "Doug Torrance",
 	    Email => "dtorrance@piedmont.edu",
 	    HomePage => "https://webwork.piedmont.edu/~dtorrance"}},
     Keywords => {"System"},
     PackageImports => {"JSON"})
+
+---------------
+-- ChangeLog --
+---------------
+
+-*
+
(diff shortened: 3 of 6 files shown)


mahrud:
Does this prevent the example from https://github.com/Macaulay2/M2/issues/3848#issuecomment-3039399776? I doubt it does, but I haven't tested. Why are [`catch/throw`](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_throw.html) not sufficient for your application?

d-torrance:
No, we still process the errors as before.  The only change is assigning the location and message to this variable before evaluating the `else` clause.

How would `catch` and `throw` work to get at information about an error?  I thought they were just for things like this:

```m2
i1 : catch scan(100, i -> if i == 67 then throw i)

o1 = 67
```

mahrud:
Sorry, what does before mean? Do you mean "no it doesn't prevent it" or "no we don't process the error"?

You can combine catch/throw with try:
```m2
i1 : err = catch scan({1,2,0,3}, i -> try i^-1 else throw "can't divide by zero")

o1 = can't divide by zero
```
and then check if err is null (the go style).

d-torrance:
"no it doesn't prevent it" -- sorry I wasn't clear!

In your example we still need to provide a separate error message.  For the JSONRPC package (and I think MPI will be similar), we want to run arbitrary code that potentially might fail and if it does, then get the error message.

mahrud:
> "no it doesn't prevent it" -- sorry I wasn't clear!

Then I disagree with this change.

seangrate:
> Sorry, what does before mean? Do you mean "no it doesn't prevent it" or "no we don't process the error"?
> 
> You can combine catch/throw with try:
> 
> ```m2
> i1 : err = catch scan({1,2,0,3}, i -> try i^-1 else throw "can't divide by zero")
> 
> o1 = can't divide by zero
> ```
> 
> and then check if err is null (the go style).

Here there is only one error type you are expecting, but if I have the potential for multiple error types, then what would I throw?

mahrud:
> Here there is only one error type you are expecting, but if I have the potential for multiple error types, then what would I throw?

In Go-style of error handling, essentially every function call returns two things: an output and an error, so it would look like:
```m2
i1 : err = catch scan({1,2,0,3}, i -> (
    (inv, err) := inverse i;
    if err =!= null then inv else throw err))

o1 = can't divide by zero
```
That is to say, if every function also returns an err object (which may be null) then there's no need for a global variable like lastError or coughtError to hold it. I think it's clear that this style isn't right for math, e.g. here one should check that i is nonzero first and in general I prefer mathematical methods to check the input first and from there run completely error free.

However, for something like RPC or MPI I would definitely prefer this style to something global like proposed in this PR.

d-torrance:
That sounds great, but changing every function to return a pair would be a massive breaking change... 

In the current proposal `caughtError` is implemented as a (thread local) global variable, but it behaves like a local one.  It's set to a two-element sequence (w/ the location and error message) only inside the `else` clause, and then set back to null afterwards.

I also considered making it a true local variable, maybe behaving a bit like `for` where the user can provide a variable name.  Something like `try ... else ... as e`, where `e` inside the `else` clause would contain the error information.  But adding a whole new keyword and dealing with local frames, etc, would be a lot more work.

mahrud:
> That sounds great, but changing every function to return a pair would be a massive breaking change...

Is it massive? If you only use this style of error handling in JSON and MPI related packages it probably wouldn't break anyone else's code. This is the kind of change that is easier to do sooner rather than later, and unless the problem in https://github.com/Macaulay2/M2/issues/3848#issuecomment-3039399776 is mitigated, that's the only option I can think of.

pzinn:
I think your problem in #3848 can be fixed in unrelated ways: by making sure `error` doesn't try to do useless computations when creating the error message, and even if it does by strictly enforcing a time limit on the error output routiine.

mahrud:
Those are not sufficient workarounds. I have `try .. else ..` clauses in code that runs many many times in hopes of opportunistic shortcuts, so a small time limit on each `else` adds up, and it's not avoidable like `promote`.

I think it's good to put a "functional" hat on and ask if an error that is not going to be printed needs to be processed, and if the answer is no then a variable `caughtError` is not the answer. Maybe the answer is another keyword like `try ... catch e -> (..)`, or maybe instead of a `caughtError` variable you want to have a code that can be evaluated as necessary.

pzinn:
> Those are not sufficient workarounds. I have `try .. else ..` clauses in code that runs many many times

why would you do that? as the doc says, "users are recommended to use this function [try] sparingly, if at all."

mahrud:
I believe we had this conversation before. I don't see how that recommendation is an argument in favor of making `try` _slower_ than it is. I think my position here is clear so I'll avoid further back and forth.

d-torrance:
See #4051 for another proposal

d-torrance:
Closing this in favor of #4051

PR #4048, opened 2025-12-02
Use an atomic hash counter

d-torrance:
Previously, it was thread-local, which meant that mutable objects created in threads might not be "younger" than older objects created in other threads.

This also has the added advantage of avoiding the hack from #3359 we needed for 32-bit systems.

Closes: #3358

### Before:
```m2
i1 : Foo = taskResult schedule(() -> new Type of HashTable);

i2 : youngest(Foo, Matrix)

o2 = Matrix
```

### After:
```m2
i1 : Foo = taskResult schedule(() -> new Type of HashTable);

i2 : youngest(Foo, Matrix)

o2 = Foo
```

### DIFF
--- a/M2/Macaulay2/d/expr.d
+++ b/M2/Macaulay2/d/expr.d
@@ -22,20 +22,13 @@ threadLocal export recursionDepth := 0;
 --Maximum function depth before triggering errors
 threadLocal export recursionLimit := 300;
 
-
-threadCounter := 0;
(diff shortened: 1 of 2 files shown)


mahrud:
Oh I didn't realize hash counters were thread-local, this might be the source of some of my parallel headaches.

antonleykin (approved):
This seems good to me. 
Also, fewer hacks is always better.

PR #4049, opened 2025-12-03
better WebApp mode support for Python

pzinn:
this is work in progress to improve the display of the `Python` package in `WebApp` mode:
- slight improvement to existing output routines
- new backend for matplotlib, automatically loaded when package is loaded in `WebApp` mode
- uncanned an example, though it may need to be cached (uses matplotlib)


d-torrance (approved):
Thanks!

Issue #2596, opened 2022-09-05
Keyword shield doesn't work


mahrud:
The [documentation](http://www2.macaulay2.com/Macaulay2/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_shield.html) claims:
> shield x -- executes the expression x, temporarily ignoring interrupts.

Unless I'm misunderstanding the page, which doesn't include any examples, `shield` doesn't seem to be working:
```m2
i1 : f = t -> shield sleep t;

i2 : f 10
^Cstdio:2:1:(3): error: interrupted

i3 : f = t -> ( alarm 1; shield sleep t );

i4 : f 5
stdio:4:1:(3): error: alarm occurred
```

DanGrayson:
Yup, it seems that that doesn't work.  Good catch.

mahrud:
I doubt this will be easy to do for a volunteer. The relevant code is in `evaluate.d` and the most recent changes to it are from 13 years ago by you (there are no other contributors).

watsonkm:
If this is still available, I'd love to take a swing at it!

DanGrayson:
I've assigned you, thanks for volunteering!

watsonkm:
Of course, thanks so much!

watsonkm:
Finally got to figuring out the issue here, my apologies for the delay! 

This is what I've been able to surmise is happening during e.g. `shield sleep 5`:

- `shieldfun` sets the `interruptShield` flag and executes the `sleep 5` expression
- the `sleep 5` expression in turn calls sleep(2) which suspends the thread
- the SIGINT is received, waking the prompt thread and passing execution into `interrupt_handler`
- `interrupt_handler` sees that `interruptShield` is set, and correspondingly sets `interruptPending`
- control then passes immediately back to `sleep(2)`, which passes control flow back to `shieldfun`
- `shieldfun` turns off `interruptShield` and stores `interruptPending` into `interruptFlag`
- `evalraw` sees `interruptFlag` is set and gives the exception

The "core" issue here seems to just be that the sleep concludes immediately when SIGINT is received. There's two possible fixes I thought of--if either of them seem useful, I can write them up real quick and share a PR:

1. Make both the `sleep` and `nanosleep` keywords both use nanosleep(2) directly, saving the return value (a `timespec` with the remaining amount of sleep time before the interrupt) somewhere `interrupt_handler` can access it, and having `interrupt_handler` resume sleep if it sees that value set.
[...]

Issue #2568, opened 2022-08-17
Spurious layers when debugging


mahrud:
When debugging an error in a method call, particular if the method had options, I often have to repeatedly enter `end` several times, going through the same exact lines that are unrelated to the error:
```m2

ii31 : end
../../Macaulay2/m2/option.m2:17:8:(1):[21]: --entering debugger (type help to see debugger commands)
../../Macaulay2/m2/option.m2:17:8-19:34: --source code:
       uncurry(f, override (opts,args))
       )
  )

ii32 : end
../../Macaulay2/m2/methods.m2:80:40:(1):[19]: --entering debugger (type help to see debugger commands)
../../Macaulay2/m2/methods.m2:80:40-80:51: --source code:
    if (f := lookup key) =!= null then dispatcher f else error noMethod(M, args, outputs))

ii33 : end
../../Macaulay2/m2/methods.m2:117:35:(1):[18]: --entering debugger (type help to see debugger commands)
../../Macaulay2/m2/methods.m2:117:35-117:146: --source code:
        binaryLookup := (x, y) -> binaryCaller'(methodFunction, (methodFunction, dispatchBy x, dispatchBy y), (x, y), outputs, dispatcher(o, (x, y)));

ii34 : end
../../Macaulay2/m2/methods.m2:72:40:(1):[17]: --entering debugger (type help to see debugger commands)
../../Macaulay2/m2/methods.m2:72:40-72:45: --source code:
foldL = (f, x, L) -> (scan(L, y -> x = f(x, y)); x)

ii35 : end
../../Macaulay2/m2/methods.m2:72:23:(1):[16]: --entering debugger (type help to see debugger commands)
../../Macaulay2/m2/methods.m2:72:23-72:45: --source code:
foldL = (f, x, L) -> (scan(L, y -> x = f(x, y)); x)

ii36 : end
../../Macaulay2/m2/methods.m2:118:9:(1):[15]: --entering debugger (type help to see debugger commands)
../../Macaulay2/m2/methods.m2:118:9-118:48: --source code:
        foldL(binaryLookup, args#0, drop(args, 1)));

ii37 : end
../../Macaulay2/m2/option.m2:40:8:(1):[14]: --entering debugger (type help to see debugger commands)
../../Macaulay2/m2/option.m2:40:8-42:30: --source code:
       uncurry(f, override (,args))
       )
  )
```
[...]

PR #4051, opened 2025-12-04
Add "trap" keyword for Go-like error handling

d-torrance:
This is an alternative to #4047 that may address @mahrud's concerns.  Rather than touching `try`, we add a new keyword `attempt` that returns a pair `(val, err)` (much like Go), where `val` is the value of the code and `err` is null if no error occurred, and otherwise `val` is null and `err` is an instance of a new `Error` type with information about the error that occurred .

```m2
i1 : attempt 5

o1 = (5, )

o1 : Sequence

i2 : attempt 1/0

o2 = (, Error{"message" => division by zero})
              "position" => stdio:2:8-2:11

o2 : Sequence
```


M2/Macaulay2/packages/Cremona.m2
```diff
@@ -1690,9 +1690,9 @@ point (Ideal,Boolean) := (I,b) -> (  -- see also: code(randomKRationalPoint,Idea
    local par;
    if c == 1 then (
        L := {}; 
-       maxAttempts := 20; attempt := 0;
-       while #L == 0 and attempt < maxAttempts do (attempt = attempt+1; par = parametrize randomLinearSubspace(R,1); L = select(decompose par^* I,q -> dim q == 1 and degree q == 1));
-       if #L == 0 and attempt >= maxAttempts then error("reached maximum number of "|toString(maxAttempts)|" attempts to find point");
+       maxAttempts := 20; att := 0;
```
mahrud: `tries` (and `maxTries`) would be better I think, let's not butcher their attempt at being readable (no pun intended lol)

mahrud:
I really like this solution!

seangrate:
This seems like a good approach. Hopefully at some point, we can also build in some sort of error types so that we can check for particular errors by their *types* as opposed to their *messages*.

d-torrance:
That sounds like a good idea!

pzinn:
what I don't understand is, the output of `attempt` is a pair, but either the first or the second element of the pair is `null`? what's the point? one might as well test if output is an instance of `Error`

pzinn:
the location of the error is different than in the usual error message. is that intentional?
```
i1 : 1/0
stdio:1:1:(3): error: division by zero

i2 : attempt(1/0)

o2 = (, Error{"message" => division by zero})
              "position" => stdio:2:8-2:11

o2 : Sequence
```
(it may not be obvious, but in the example the first error point to `/` whereas the second points to `1/0`)

pzinn:
This location doesn't seem right either:
```
i1 : attempt(1/a)

o1 = (, Error{"message" => no method for binary operator / applied to objects:})
                                       1 (of class ZZ)
                                 /     a (of class Symbol)
              "position" => ../../Macaulay2/m2/robust.m2:101:25-101:66

o1 : Sequence
```

d-torrance:
> what I don't understand is, the output of `attempt` is a pair, but either the first or the second element of the pair is `null`? what's the point? one might as well test if output is an instance of `Error`

That's a good point.  I was basically just copying the Go syntax that Mahrud pointed out in the other PR.  Also, in an earlier draft, the second element was a string, not an `Error` object, so it was necessary then.

d-torrance:
As for the file positions, as I recall, errors originally all start out with `dummyPosition` and then get updated in various places, right?  I think `attempt` is grabbing it a little later than it would have gotten printed if printing wasn't suppressed by `tryEval`, so maybe the position was updated a bit more in the meantime?

mahrud:
> what I don't understand is, the output of `attempt` is a pair, but either the first or the second element of the pair is `null`? what's the point? one might as well test if output is an instance of `Error`

Strictly speaking, in Go the return value is always of the expected type, so `attempt 1/0` should return `(0, Error{...})` or something like that, but currently there's just no output. But even for us, because of type consistency this is more readable:
```m2
(inv, err) := attempt inverse n
if err =!= null then error("oops: ", err#"message")
```
than a situation where the return value has different types in each branch:
```m2
ret := attempt inverse n
if not instance(ret, Error) then inv := ret else error("oops: ", ret#"message")
```

Also, what if you're writing an error-handling related function that needs to return an error? You need to separate the legitimate output Error object from the error that may have occurred in the process :)

pzinn:
I stand by my original point.

M2/Macaulay2/d/evaluate.d
```diff
@@ -1670,6 +1670,22 @@ testfun(c:Code):Expr := (
...
+    storeInHashTable(h, toExpr("message"), toExpr(err.message));
+    Expr(sethash(h, false)));
+
+attemptfun(c:Code):Expr := (
+    r := tryEval(c);
+    if tryEvalSuccess then seq(r, nullE)
+    else (
+	when r
+	is err:Error do seq(nullE, toExpr(err))
+	else seq(r, nullE))); -- shouldn't happen
```
antonleykin: This shouldn't happen, unless the code fabricates it's own `Error`. I don't think we should let `attempt` output user-fabricated stuff.

M2/Macaulay2/packages/Macaulay2Doc/ov_debugging.m2
```diff
@@ -327,6 +327,43 @@ document {
...
+doc ///
+  Key
+    symbol attempt
+  Headline
+    attempt to evaluate code and get information about any errors that occur
+  Usage
+    attempt c
+  Description
+    Text
+      The code @VAR "c"@ is evaluated and a sequence containing two elements
```
antonleykin, line 375: Emphasize that this is a keywork in the language, not a method. One could mentally parse `(attempt 1)/0`. 
Perhaps give more examples: what would `attempt 1/2; 1/0` give?

antonleykin (changes requested):
The two ideas are good. 
The second (proposed by @seangrate) to have an `Error` type.  
The original idea to "capture message of error" (due to @d-torrance) via `attempt` (returning it, not storing, proposed by @mahrud).  

Returning one thing---either "result" or `Error`---would be more natural, since our language is not statically typed (e.g., the type of the "result" is unknown anyway).

M2/Macaulay2/d/evaluate.d
d-torrance: Makes sense.  I've replaced this with an error.

M2/Macaulay2/packages/Macaulay2Doc/ov_debugging.m2
d-torrance: Documentation expanded

d-torrance:
@antonleykin - I think I've addressed your comments.  `attempt` now returns a single element, either the value of the evaluated code or an `Error` object.

mahrud:
I would encourage you to reevaluate this decision. Here are a few dangers of this paradigm:
1. If a helper function `constructMyError` legitimately returns an `Error` object (we have LOTS of these helpers in Core), it's impossible to detect from `err := attempt constructMyError(...)` if `err` is a legitimate output or if `constructMyError` itself failed. We need Core to be robust, so I wouldn't use `attempt` there with this implementation.
2. Since error checking is less explicit, it's easy for users to forget to check for errors and only later end up with confusing type errors like `dim myFunc()` giving `error: no method found for applying dim to: argument : Error {...} (of class Error)`
3. Even worse, since `Error` inherits from hash tables, a caller can accidentally treat the output as input to a function that accepts a hash table and consequently mask the error. Here is an explicit example of this:
```m2
f := M -> attempt apply(codim M, i -> i^2);
# f(ZZ^100)      -- gives #{}         which is 0
# f((ZZ[x])^100) -- gives #Error{...} which is 2
```

If you're going through the trouble of adding a new keyword, I'd like it to be robust and future proof, especially for MPI or RPC type applications.

pzinn:
> I would encourage you to reevaluate this decision. Here are a few dangers of this paradigm:
> 
> 1. If a helper function `constructMyError` legitimately returns an `Error` object (we have LOTS of these helpers in Core), it's impossible to detect from `err := attempt constructMyError(...)` if `err` is a legitimate output or if `constructMyError` itself failed. We need Core to be robust, so I wouldn't use `attempt` there with this implementation.

this I don't understand at all. There seems to be a basic confusion here. These `Error` types are produced by the interpreter itself or at the level M2 level by calling `error`, *not* by creating an error manually. If we were to do what you suggest that would be a totally different usage which would require a separate discussion -- and a complete rewrite of this PR anyway.
> 3. Even worse, since `Error` inherits from hash tables, a caller can accidentally treat the output as input to a function that accepts a hash table and consequently mask the error. Here is an explicit example of this:
> 
> ```m2
> f := M -> attempt apply(codim M, i -> i^2);
> # f(ZZ^100)      -- gives #{}         which is 0
> # f((ZZ[x])^100) -- gives #Error{...} which is 2
> ```
well, if you're going to write `attempt`, that means you're expecting an error, so you should test for it, otherwise you probably shouldn't use `attempt` in the first place.
That being said, I agree that `Error` probably shouldn't inherit from `HashTable`; instead, just like `Dictionary`, it should be its own type.

mahrud:
> There seems to be a basic confusion here. These `Error` types are produced by the interpreter itself or at the level M2 level by calling `error`, _not_ by creating an error manually. If we were to do what you suggest that would be a totally different usage which would require a separate discussion

Why shouldn't error objects be produced at the top-level? Is something preventing it in this proposal? There is a huge upside, as @seangrate suggested above, in allowing custom subtypes inheriting from Error so that error checking looks more like `if instance(err, RingError) ...` rather than `if err#"message" == "incorrect ring" ...`. Otherwise why bother with a new type at all?

We currently don't have an Error type, but once it exists I guarantee this PR isn't going to be the end of it. Soon we'll have extensions like Python error types and Flint error types and use it with hooks and threads and loads more that we can't imagine now. (Side note: @d-torrance if `err` is an instance of `Error` then probably `error err` should produce an error with that location and message)

> well, if you're going to write attempt, that means you're expecting an error, so you should test for it, otherwise you probably shouldn't use attempt in the first place.

Mistakes like this will sneak in, either by people treating `attempt` like `try` or just mistakes in the programming where there are multiple branches and you forget to check if the output is an error in one branch.

More generally, I'm just surprised that we're going with a syntax like this:
```m2
c := attempt codim M;
if instance(c, Error) then
if c#"message" == "codim: expected an affine ring (consider Generic=>true to work over QQ"
then c = planZZ(M)
else error "unexpected module"
```

d-torrance:
I'm convinced by @mahrud's arguments and prefer the Go-style two-element output.

pzinn:
> 
> Why shouldn't error objects be produced at the top-level? Is something preventing it in this proposal? There is a huge upside, as @seangrate suggested above, in allowing custom subtypes inheriting from Error so that error checking looks more like `if instance(err, RingError) ...` rather than `if err#"message" == "incorrect ring" ...`. Otherwise why bother with a new type at all?

because that simply wouldn't work in the current setting. what would happen if you called your function that outputs an error your way without `attempt`? in fact, what's the role of `attempt` then? As I said, if you want to go this route, then we should scrap this PR and start from scratch with a completely different design.

There absolutely should be subtypes, but this is irrelevant to the point I'm making.
I'm simply suggesting `Error` be its own "root" type.

d-torrance:
I went back to the Go-style output and also added some basic support for subclassing of `Error`:

```m2
i1 : MyError = new SelfInitializingType of Error;

i2 : MyError "foo"

o2 = MyError{"message" => foo}

o2 : MyError

i3 : error o2
stdio:3:5:(3): error: foo

i4 : attempt error o2

o4 = (, MyError{"message" => foo                            })
                "position" => ../m2/debugging.m2:21:22-21:75

o4 : Sequence
```

mahrud:
Looks like the error
> "position" => ../m2/debugging.m2:21:22-21:75

is pointing to this line:
```m2
error Error := err -> olderror new class err from processArgs err#"message"
```
No obligation to fix this in this PR, but obviously the location should point somewhere more appropriate. There are probably multiple ways to fix it (e.g. maybe moving `error(Error)` to the interpreter?), but one of my hopes from the `attempt` keyword is to fix these spurious error layers: https://github.com/Macaulay2/M2/issues/2568.

pzinn:
I understand less and less the logic. Now one can create `Error` objects at the M2 level, which are then fed to `error`, which at the d level proceeds to discard that `Error` object, only keeping track of the message, and creates a new `Error` object out of it... ??

mahrud:
The point of object oriented error handling is that instead of interrupting the computation, an error object is created to help the subsequent code resolve it and proceed without interruption. But there may be errors objects which can't be handled, so there should be a way to force an interrupt so that the user can intervene. This is where `error(Error)` comes in, to simplify this:
```m2
(c, err) := attempt codim M;
if instance(err, Error) then
if instance(err, RingError)
then c = planZZ M
else error("unexpected error: " | err#"message" | " at " | err#"position")
```
into this:
```m2
(c, err) := attempt codim M;
if instance(err, Error) then
if instance(err, RingError)
then c = planZZ M
else error(err) -- the original message and position are returned
```

d-torrance:
I made some changes that may address some of @pzinn's concerns.

`Error` is now a basic type rather than a type of hash table.  Under the hood, a top-level `Error` object is a `SpecialExpr` that contains the lower-level `Error` object.  So `error(Error)` just extracts that object rather than creating a new one.

Rather than getting the message and location by looking up keys in a hash table, we can use `toString` and `locate`, resp.

antonleykin:
I think we are getting caught in the weeds and should tread carefully as the discussion diverged
1. from a simple (but elegant) solution to having a construct that 
  - handles an exception/interrupt like `try` (not terminating the program)
  - but instead of `else` returning an `Error` object
2. to an extensive rethinking of how errors are done in Macaulay2

I would like the *language* modifications to be minimal unless we are constructing a new language (Macaulay3 or GoMacaulay). I believe we should encourage users to, as someone wise said,
"prefer mathematical methods to check the input first and from there run completely error free" or write a simple statement   `error "something is wrong"` when something unexpected happen. (In particular, we'd want people to not use `try`, `catch`, etc.)

Let's try to think about minimal changes of language could be done to accomplish 1. 
(The following is not a proposition to act upon. It has downsides.) 
Could `catch` be modified so that 
```
i2 : catch apply({-1,0,3}, i->1/i)                                                                             
stdio:2:26:(3):[1]: error: division by zero 
```
returns either a string that is printed or an `Error` object?

d-torrance:
I think it would be straightforward to use `catch` instead of `attempt` for this.  It would be a breaking change, though, as right now if you use `catch` on some code that raises an error, that error is raised.  So I'd prefer to stick with `attempt` (or some other new keyword).

I don't think we should return just a string, as otherwise `attempt "foo"` and `attempt error "foo"` would be indistinguishable.  And returning just the `Error` object is a problem too, as `attempt error "foo"` and `attempt new Error from "foo"` would also be indistinguishable.

@antonleykin - Is the "extensive rethinking of how errors are done in Macaulay2" the idea of subclassing `Error`?  That's something we get for essentially free as soon as we expose that class at top level.

antonleykin:
> I think it would be straightforward to use `catch` instead of `attempt` for this. It would be a breaking change, though, as right now if you use `catch` on some code that raises an error, that error is raised. 

@mikestillman  and I discussed this earlier today. In .m2 files, the keywords `catch` and `throw` appear to be used... 0 times. Perhaps we have a chance to repurpose (dead wood)?
   
> I don't think we should return just a string, as otherwise `attempt "foo"` and `attempt error "foo"` would be indistinguishable. And returning just the `Error` object is a problem too, as `attempt error "foo"` and `attempt new Error from "foo"` would also be indistinguishable.

This is a deficiency, but also seem to be a hypothetical edge case.
Suppose we are redesigning `catch`:  one can make 
* `throw` return an error object of one type (inheriting from say, `AbstractError`) 
* any error resulting from `error` return an error object of a distinct type
* force the code that attempts to return an error object to fail (interrupt the program) 

This would make returning a pair unnecessary. Returning a pair is, of course, not a big deal, but I think we'd be just paying tribute to the edge case (e.g., hypothetical users who would use nested `catch`es or micromanaging error objects at the top level). 

> @antonleykin - Is the "extensive rethinking of how errors are done in Macaulay2" the idea of subclassing `Error`? That's something we get for essentially free as soon as we expose that class at top level.

My point: we can export `Error` to top level and subclass it anyway we want, but I would stay away from advertising this to users, and also using it ourselves (at top level) unless absolutely necessary.

mahrud:
I have an alternative name suggestion: `trap` (because errors can't _escape_ lol)

> I would like the _language_ modifications to be minimal unless we are constructing a new language (Macaulay3 or GoMacaulay). I believe we should encourage users to, as someone wise said,
> "prefer mathematical methods to check the input first and from there run completely error free" or write a simple statement `error "something is wrong"` when something unexpected happen. (In particular, we'd want people to not use `try`, `catch`, etc.)

The emphasis there was on **mathematical methods**. Literally in the next sentence I said that RPC and MPI are robust protocols and need a robust implementation that covers every hypothetical edge case (and honestly I don't think the examples that Doug and I provided are hypothetical at all).

I wouldn't recommend a casual user to use `attempt` either, just like I wouldn't recommend they use `capture`, but both are robust methods which have an important place in the Core (like robust.m2 and debugging.m2) and specific other situations (like RPC/MPI/Python but also external software interfaces).

Also, I don't think anyone is suggesting an extensive rewrite of how errors are done in M2 today. I think this PR is ready to be merged right now. However, if we're introducing a new keyword, I'd like it to allow future extensions so we avoid a future breaking change.

ps: if you want to redefine `catch`, my vote is something more intuitive:
```m2
try (...)
catch   RingError do (...) -- do this if the ring is bad
catch ModuleError do (...) -- do this if the module is bad
else (...) -- do this if there's another error
```

M2/Macaulay2/d/classes.dd
```diff
@@ -116,6 +117,13 @@ export Parent(e:Expr):HashTable := (
 export parentfun(e:Expr):Expr := Expr(Parent(e));
 setupfun("parent",parentfun);
 
+export Class(err:Error):HashTable := (
+    when err.value
+    is x:HashTable do (
+	if ancestor(x, errorClass) then x
+	else errorClass)
+    else errorClass);
+
```
pzinn, line 126: can you explain this bit?

pzinn:
just trying to understand the latest changes. now this works:
```
new Error from 42
```
but what happens under the hood?
also why the need for a special `new Error from String`?

M2/Macaulay2/tests/normal/error-messages.m2
mahrud, line 1: Might be worth adding a test for nested attempts, e.g. `attempt attempts error "foo"` gives `((, foo), )`

M2/Macaulay2/m2/debugging.m2
```diff
@@ -13,11 +13,14 @@ processArgs := args -> concatenate (
      apply(args, x -> if class x === Symbol then ("\n", toString locate x, ": here is the first use of '",toString x, "'") else "")
      )
 olderror := error
-error = args -> (
-     -- this is the body of the "error" function, which prints out error messages
-     olderror processArgs args)
+error = method(Dispatch => Thing)
+error String :=
+error Error  := olderror
+error Thing  := olderror @@ processArgs
```
mahrud: One of these is different lol
```m2
i1 : errorDepth=1;

i2 : error new Error from "hi"
stdio:2:5:(3): error: hi

i3 : error "hi"
stdio:3:5:(3): error: hi

i4 : error 12
Projects/M2/engine/M2/Macaulay2/m2/classes.m2:114:51:(1):[2]: error: 12
Projects/M2/engine/M2/Macaulay2/m2/classes.m2:114:51:(1): entering debugger (enter 'help' to see commands)
Projects/M2/engine/M2/Macaulay2/m2/classes.m2:114:50-114:55: --source code:
Function @@ Function := Function => (f,g) -> x -> f g x
```

Note that this shows up in unexpected places too:
```m2
i10 : ZZ^{1}
Projects/M2/engine/M2/Macaulay2/m2/classes.m2:114:51:(1):[3]: error: expected each multidegree to be of length 0
Projects/M2/engine/M2/Macaulay2/m2/classes.m2:114:51:(1): entering debugger (enter 'help' to see commands)
Projects/M2/engine/M2/Macaulay2/m2/classes.m2:114:50-114:55: --source code:
Function @@ Function := Function => (f,g) -> x -> f g x
```
At the very least, change `error(Thing)` to just `x -> olderror processArgs x`, then the interpreter stops at that line. But if you can make it so the interpreter never stops here that'd be super wonderful.

M2/Macaulay2/d/evaluate.d
```diff
@@ -1670,6 +1670,18 @@ testfun(c:Code):Expr := (
...
+
+attemptfun(c:Code):Expr := (
+    r := tryEval(c);
+    if tryEvalSuccess then seq(r, nullE)
+    else (
+	when r
+	is err:Error do seq(nullE, toExpr(err))
+	-- shouldn't happen since tryEvalSuccess only false when r is an Error
+	else buildErrorPacket("unable to catch error")));
+setupop(attemptS, attemptfun);
```
mahrud: Now that I look at how simple this is, could you remind me what's the issue or difficulty with this syntax?
```m2
try (...) catch err do (...) -- assigns the error object to `err`.
```
Importantly, based on the fact that `toExpr(err)` happens pretty late, I presume if we could use `catch` like this, normal `try .. else ..` would not be slowed down like the previous proposal. This might satisfy all parties (though part of me thinks a `trap` with a second error output would still be useful.

(If this doesn't work, I think renaming to `trap` will be good, so the error can say `unable to trap the error`. Also, the indentation looks strange. You could do `else when r is ...`)

M2/Macaulay2/d/actors4.d
```diff
@@ -663,10 +663,13 @@ stringcatfun(e:Expr):Expr := (
 setupfun("concatenate",stringcatfun);
 
 errorfun(e:Expr):Expr := (
-     e = stringcatfun(e);
      when e
      is s:stringCell do buildErrorPacket(s.v)
-     else buildErrorPacket("expects a string or sequence of strings as its argument"));
+     is err:Error do (
+	 err.printed = false;
+	 e)
```
mahrud: Sorry to be nitpicky but the indentation looks strange.
```suggestion
     is err:Error do ( err.printed = false; e )
```

M2/Macaulay2/d/classes.dd
d-torrance: This is kind of a hack, but this let us subclass `Error`.  We store the class in two places:

* In the `SpecialExpr` object
* In the `value` member of the low-level `Error` object, which is currently only used for things like `break`, `continue`, and `throw` that we wouldn't catch with `attempt` (since if they aren't handled yet, we'd get an "unhandled" error instead).

The reason we need it in both places is so that we don't lose the class information when we do `error(Error)`, which extracts the low-level `Error` object from the `SpecialExpr`.  So for example:

```m2
i1 : MyError = new SelfInitializingType of Error;

i2 : last attempt error MyError "foo"

o2 = foo

o2 : MyError
```

If we didn't store `MyError` inside the low-level `Error` object so that `attempt` can find it when it calls `toExpr(Error)`, then we'd lose the subclass information.

d-torrance:
> but what happens under the hood? also why the need for a special `new Error from String`?

`new Error from String` is the low-level d constructor that constructs a new low-level `Error` object from the given message and stuffs it in a `SpecialExpr` to make the top-level `Error` object.

`new Error from Thing` is a top-level wrapper around this that calls the helper function `processArgs` on its input (which is also what we use when we call `error`) to construct a string that gets sent to the low-level method.

M2/Macaulay2/m2/debugging.m2
d-torrance: Done!

```m2
i1 : errorDepth = 1;

i2 : error 12
../m2/debugging.m2:19:29:(1):[2]: error: 12
../m2/debugging.m2:19:29:(1): entering debugger (enter 'help' to see commands)
../m2/debugging.m2:19:21-19:43: --source code:
error Thing  := x -> olderror processArgs x

ii3 : ZZ^{1}
../m2/debugging.m2:19:29:(1):[3]: error: expected each multidegree to be of length 0
../m2/modules.m2:288:18:(1):[1]: --back trace--
```

d-torrance: Hmm, the `ZZ^{1}` error isn't quite as useful if we break out of the debugger:

```m2
ii4 : break

i5 : ZZ^{1}
../m2/debugging.m2:19:29:(1):[3]: error: expected each multidegree to be of length 0
../m2/debugging.m2:19:29:(1): entering debugger (enter 'help' to see commands)
../m2/debugging.m2:19:21-19:43: --source code:
error Thing  := x -> olderror processArgs x
```

M2/Macaulay2/d/actors4.d
d-torrance: Fair enough.  I generally like to stick with one statement per line but `e` by itself is a little odd.  :) Fixed

M2/Macaulay2/d/evaluate.d
d-torrance: That's probably doable.  We'd probably need to introduce a new `TryCatchDo` member of the `ParseTree` union, but that shouldn't be too bad.  Would that be the only combination, or are you envisioning `else` and/or `then` showing up with `catch` and `do` as well? 

I am a little worried that `catch` may have too high of a precedence -- we'd probably want whatever keyword went here to have a low one matching `do`, `else`, and `then`.  Maybe `try ... trap ... do`?

d-torrance:
I went ahead and added a commit implementing a `try ... trap ... do` statement.  If this works well, then we can remove `attempt`.

```m2
i1 : try 1/0 trap err do err

o1 = division by zero

o1 : Error
```

M2/Macaulay2/d/evaluate.d
mahrud: I mean, as Anton pointed out catch isn't really used anywhere, so I think `try .. catch err do ..` makes more sense, but if you don't want to mess with precedences maybe `try .. except err do ..` instead? `trap` doesn't really make sense here.

mahrud: I think we should limit the scope for now, but something like this could make sense in the future:
```m2
try ... then ...
catch   MyError as err do ...
catch RingError as err do ...
else ... -- for other errors
```

d-torrance: I've switched it to `try ... except err do ...`

mahrud:
Here's a suggestion to get things moving: let's add both `try ... except err do ...` and also `(out, err) = trap ...` (so you can undo renaming `attempt` in a few packages) in this PR, try both for a few months, and decide before the next release whether we want to keep one or both or make any improvements.

M2/Macaulay2/m2/debugging.m2
mahrud: This is an epsilon better than the current behavior, which doesn't even fit the full function:
```m2
..../debugging.m2:18:5-18:30: --source code:
     olderror processArgs args)
```
But this annoyance is exactly why I suggested moving `error(Thing)` to the interpreter (or perhaps `startup.m2`? I will try this out at some point). I would love a general solution for #2568 that tells M2 "never interrupt at this point, it's irrelevant!" e.g. I don't want to step through methods and hooks and options while trying to figure out why my math function has an error. It's possible that `shield` was supposed to do this, but that hasn't worked for quite a while: #2596

M2/Macaulay2/d/parse.d
```diff
@@ -197,6 +198,7 @@ export globalAssignmentCode := {+
      };
 export ifCode := {+ predicate:Code, thenClause:Code, elseClause:Code, position:Position };
 export tryCode := {+ code:Code, thenClause:Code, elseClause:Code, position:Position };
+export tryExceptDoCode := {+ code:Code, doClause:Code, frameID:int, framesize:int, position:Position };
```
mahrud: Is there a reason this can't fall under the same `tryCode`? (for now we could flag an error if both `except` and `else` are given)

M2/Macaulay2/d/convertr.d
```diff
@@ -333,6 +333,7 @@ export convert0(e:ParseTree):Code := (
     is i:TryThen     do Code(tryCode(convert(i.primary), convert(i.sequel), NullCode,             pos))
     is i:TryThenElse do Code(tryCode(convert(i.primary), convert(i.sequel), convert(i.alternate), pos))
     is i:TryElse     do Code(tryCode(convert(i.primary), NullCode,          convert(i.alternate), pos))
+    is i:TryExceptDo do Code(tryExceptDoCode(convert(i.primary), convert(i.doClause), i.dictionary.frameID, i.dictionary.framesize, pos))
```
mahrud: To elaborate on my previous comment, this could be something like:
```suggestion
    is i:TryExceptDo do Code(tryCode(convert(i.primary), NullCode, NullCode, convert(i.doClause), i.dictionary.frameID, i.dictionary.framesize, pos))
```
(And the other lines need to be adjusted for the new inputs too)

M2/Macaulay2/d/parse.d
d-torrance: I thought about that, but I think this keeps the code in `evaluate.d` a bit cleaner, where we don't have to worry about all the various cases of which clauses we have and don't have.

M2/Macaulay2/m2/debugging.m2
d-torrance: I went ahead and moved `error` completely to the interpreter:

```m2
i1 : errorDepth = 1

o1 = 1

i2 : ZZ^{1}
../m2/modules.m2:288:18:(1):[1]: error: expected each multidegree to be of length 0 (of class ZZ)
../m2/modules.m2:288:18:(1): entering debugger (enter 'help' to see commands)
../m2/modules.m2:288:13-288:71: --source code:
        then error("expected each multidegree to be of length ", degrk))
```

d-torrance: Hmm, the `(of class ZZ)` is silly.  I used the same robust printing method that we were using for hash table key errors, where the class makes sense.  Here it doesn't.

mahrud: Ha that can be confusing .. This is great though, thank you!

M2/Macaulay2/d/parse.d
mahrud: What about something like this? I feel like it's a pretty clean addition:
```m2
	  is c:tryCode do (
	      ret := tryEval(c.code);
	      if tryEvalSuccess then
	      -- what to do in case of success
	      when ret is Error do ret -- TODO: what kind of error leaves tryEvalSuccess=true?
	      else if c.thenClause == NullCode then ret   else eval(c.thenClause)
	      -- what to do in case of failure
	      else if c.doClause   != NullCode then evalExceptDoCode(c.doClause)
	      else if c.elseClause == NullCode then nullE else eval(c.elseClause))
```
where `evalExceptDoCode` is similar to `evalTryCode`.

mahrud: Actually it should be `evalExceptDoCode(c.doClause, ret)`.

Do you know why only 6 types of errors are escape try but some don't? e.g.
```m2
i13 : step
stdio:13:0:(3): error: unhandled step command

i14 : break
stdio:14:0:(3): error: unhandled break command

i15 : try (1+1; step) else 3

o15 = 3

i16 : try (1+1; break) else 3
stdio:16:10:(3): error: unhandled break command
```
I gather for some types of errors we want an actual interruption (e.g. literal ctrl-c interrupt should not be captured by the else clause in a try) but idk why `step` is okay but `break` is bad.

d-torrance: Good question -- I'm not sure.  `step` is only for the debugger, right?  Whereas `break` is used in both the debugger and the regular language, so maybe that's why?

Your `tryCode` suggestion seems reasonable -- I'll play around with it.

d-torrance:
I've made a few updates:

 * `attempt` is now `trap`
 * `tryExceptDoCode` has been removed and its functionality merged into `tryCode`
 * `RobustPrintMethod` has been split into two methods: `RobustPrintStringMethod` for printing arbitrary things passed to `error` and `RobustPrintNetMethod` for "key not found" errors.

M2/Macaulay2/packages/Macaulay2Doc/ov_language.m2
```diff
@@ -704,37 +704,70 @@ document {
...
-     PARA{},
-     "The behavior of interrupts (other than alarms) is unaffected.",
-     EXAMPLE "apply(-3..3,i->try 1/i else infinity)",
-     Caveat => "We will change the behavior of this function soon so that it will be possible to catch errors of a particular type.  Meanwhile, users are
-     recommended to use this function sparingly, if at all."
-     }
+doc ///
+  Key
+    symbol try
+    "except" -- TODO: 'symbol except' here raises "global symbols inadvertently defined" error.  why?
```
mahrud: Because it needs to be exported in `exports.m2`.

M2/Macaulay2/d/convertr.d
```diff
@@ -329,10 +329,11 @@ export convert0(e:ParseTree):Code := (
 	  Code(localSymbolClosureCode(nestingDepth(sym.frameID, token.dictionary), sym, pos)))
     is i:IfThen      do Code(ifCode(convert(i.predicate), convert(i.thenClause), NullCode,              pos))
     is i:IfThenElse  do Code(ifCode(convert(i.predicate), convert(i.thenClause), convert(i.elseClause), pos))
-    is i:Try         do Code(tryCode(convert(i.primary), NullCode,          NullCode,             pos))
-    is i:TryThen     do Code(tryCode(convert(i.primary), convert(i.sequel), NullCode,             pos))
-    is i:TryThenElse do Code(tryCode(convert(i.primary), convert(i.sequel), convert(i.alternate), pos))
-    is i:TryElse     do Code(tryCode(convert(i.primary), NullCode,          convert(i.alternate), pos))
+    is i:Try         do Code(tryCode(convert(i.primary), NullCode,          NullCode,             NullCode, 0, 0, pos))
```
mahrud, line 332: I think this is fine, but putting `frameID` 0 here isn't going to cause null pointer errors or weird access to other frames, right?

M2/Macaulay2/packages/Macaulay2Doc/ov_language.m2
d-torrance: D'oh -- thanks!

M2/Macaulay2/d/convertr.d
d-torrance: We only ever use this value in the `try ... except ... do ... case` (or when just printing it in `pseudocode`/`disassemble`), so it should be fine

M2/Macaulay2/d/evaluate.d
```diff
@@ -1478,10 +1480,26 @@ export evalraw(c:Code):Expr := (
...
-	      else if c.thenClause == NullCode then ret   else eval(c.thenClause)
-	      else if c.elseClause == NullCode then nullE else eval(c.elseClause))
+	      if tryEvalSuccess then (
+		  when ret is Error do ret
+		  else if c.thenClause == NullCode then ret
+		  else eval(c.thenClause))
+	      else (
+		  if c.elseClause != NullCode then eval(c.elseClause)
+		  else if c.doClause == NullCode then nullE
+		  else ( -- try .. except .. do
```
mahrud: Just pointing out that this is close to the suggestion I had, but different in one way: in my example, "except" overrides "else" but I think here "else" overrides "except". It may be a distinction without a difference in behavior, but worth be aware of.

d-torrance: The parser will raise a syntax error if both are given, so I don't think it matters:

```m2
i1 : try 1/0 except err do err then "foo" else "bar"
stdio:1:26:(3): error: syntax error at 'then'

i1 : try 1/0 then "foo" else "bar" except err do err
stdio:2:30:(3): error: syntax error at 'except'
```

M2/Macaulay2/d/evaluate.d
```diff
@@ -1670,6 +1688,16 @@ testfun(c:Code):Expr := (
     when r is Error do r else nullE);
 setupop(TestS, testfun);
 
+trapfun(c:Code):Expr := (
+    r := tryEval(c);
+    if tryEvalSuccess then seq(r, nullE)
```
mahrud: Keep in mind `tryEvalSucess` can be true with certain errors. e.g.
```m2
i6 : class \ trap (1 + 1; return)

o6 = (Error, Nothing)

o6 : Sequence

i7 : class \ trap (1 + 1; step)

o7 = (Nothing, Error)

o7 : Sequence

i8 : class \ trap (1 + 1; continue)

o8 = (Error, Nothing)

o8 : Sequence

i9 : class \ trap (1 + 1; break)

o9 = (Error, Nothing)

o9 : Sequence

i10 : trap (1 + 1; break)

--error in conversion of output to string
stdio:10:0:(3): error: at print: no method found for applying between to:
     argument 1 :  "," (of class String)
     argument 2 :  null (of class Nothing)
```
Perhaps we should rename this variable to avoid confusion?

mahrud: Maybe it should be `tryCaughtError` which is true if a catchable error is caught and false if there's either no error or the error is not catchable (e.g. interrupts, syntax errors, etc).

d-torrance: I like it!

mahrud: I think the new commit still has this behavior, no?
```m2
i6 : class \ trap (1 + 1; return)

o6 = (Error, Nothing)
```

mahrud: I think in this case we should directly return `err`, meaning something like:
```m2
i1 : class \ trap (1 + 1; return)
stdio:1:12:(3): error: unhandled return command
```
(i.e. I believe `trap` shouldn't catch syntax errors, just signaled errors)

d-torrance: Oops, yeah.  ~~I blindly just swapped true and false without thinking!!~~ Actually, it never did this.  But it should!

M2/Macaulay2/d/actors4.d
```diff
@@ -992,10 +985,13 @@ tostringfun(e:Expr):Expr := (
...
      is z:CCcell do toExpr(tostringCC(z.v))
-     is Error do toExpr("<<an error message>>")
+     is err:Error do toExpr(err.message)
      is Sequence do toExpr("<<a sequence>>")
      is HashTable do toExpr("<<a hash table>>")
-     is List do toExpr("<<a list>>")
+     is x:List do (
+	 if ancestor(x.Class, filePositionClass)
+	 then toExpr(tostringFilePosition(e))
+	 else toExpr("<<a list>>"))
```
mahrud, line 994: Not to add to your work but would it be simpler if FilePosition was its own interpreter-level class?

d-torrance: That makes a lot of sense -- we could probably just add `Position` to the `Expr` union and use that.  Maybe in another PR, though.

mahrud:
I think this PR is good to go. While testing I found some pre-existing issues with robust error printing, but I'll make those as separate issues.

mahrud:
@d-torrance could you find an example of an error that is time limited by the new robust printing code? I just want to make sure everything is working as intended.

d-torrance:
It works!

```m2
i1 : Foo = new Type of HashTable;

i2 : net Foo := toExternalString Foo := x -> (sleep 10; "foo");

i3 : x = new HashTable; x#(new Foo)
stdio:3:20:(3): error: key not found in hash table:
        new Foo from {} (of class Foo)

i5 : error new Foo
stdio:4:5:(3): error: new Foo from {}
```

mahrud:
Why does the first one take 6 seconds?
```m2
i1 : Foo = new Type of HashTable;

i2 : net Foo := toExternalString Foo := x -> (sleep 10; "foo");

i3 : x = new HashTable;

i4 : elapsedTime x#(new Foo)
stdio:4:13:(3): error: key not found in hash table:
        new Foo from {} (of class Foo)
 -- 6.00091s elapsed

i5 : elapsedTime error new Foo
stdio:5:17:(3): error: new Foo from {}
 -- 3.00045s elapsed
```

mahrud (approved):

d-torrance:
`silentRobustNet` tries both `net` and `toExternalString` before it resorts to `simpleToString`.  In this silly example, we're sleeping 3 seconds for each one

mahrud:
Maybe I've misunderstood the robust printing things, but shouldn't `toString` also be capped at 3s?
```m2
i9 : apply({net, toString, toExternalString}, m -> m Foo := x -> elapsedTime (print toString m; sleep 10; "foo"));

i10 : elapsedTime x#(new Foo)
net
 -- 3.00035s elapsed
toExternalString
 -- 3.0002s elapsed
toString
 -- 10.0005s elapsed
stdio:10:13:(3): error: key not found in hash table:
        foo (of class Foo)
 -- 16.0019s elapsed
```

d-torrance:
Where is it calling `toString`?  That's weird...

pzinn:
can we have a summary of what this PR currently does? it seems to have diverged quite a bit from the original one.

d-torrance:
> can we have a summary of what this PR currently does? it seems to have diverged quite a bit from the original one.

Great idea!  Here's what we've got:

* A new top-level (and subclass-able) `Error` type, implemented as a `SpecialExpr` containing an interpreter-level `Error`.  Access the message and position using `toString` and `locate`, resp.
* As the PR title says, a new `trap` keyword for Go-style error handling, i.e., one that returns a two-element sequence `(val, err)`, where `val` is the return value and `err` is null unless an error occured, in which case `val` is null and `err` is an `Error` object.
* New `try ... except ... do` and `try ... then ... except ... do` statements.  Inside the `do` clause, we can refer to the symbol specified by `except` to get the `Error` object corresponding to an error that occurrred, e.g. `try 1/0 except err do err` will return a "division by zero" `Error` object.
* Updates to `error`, including:
  * We can call `error` on an `Error` object to raise the corresponding error.
  * It is now completely defined in the interpreter, which should improve stack traces.
  * The helper functions used by `error` (`processArgs` and robust printing of objects) have also been moved to the interpreter

pzinn:
OK, my only worry now is that it conflicts with my tentative error upgrade (which is partially implemented on the branch running on the Macaulay2Web server). So, since we're doing fairly extensive changes, let me propose mine:
can we make errors be arbitrary expressions, rather than strings? at the end of the day, one can still use `toString` or `net` or whatever to output it, but it leaves the possibility to use other means of output for non Standard modes.

d-torrance:
Rather than storing the message as an `Expr` and post-processing it into a string based on the top level mode, wouldn't it be equivalent to keep it as a `string` but pre-process the input to `error`?  This is essentially what we're doing now with the `processArgs` helper function, but we could make this mode based.  And we wouldn't need to touch all the low-level error-printing code that doesn't know what an `Expr` is yet.

So maybe `error` could look up something like `(class x)#{topLevelMode, PrintError}` and use it to convert its input into a string to store as the message.

d-torrance:
I played around with the idea of a mode-based `ErrorPrint` method a bit on another branch: https://github.com/d-torrance/M2/commit/9ae759d0e888d769f73bbd37b1a5fc73f6f87a58

I think it's a bit cleaner, as we don't have to worry about another robust printing method in the interpreter.

*Edit:*  New version with an `AfterErrorPrint` method so we can handle the "here is the first use of..." messages for symbols more cleanly: https://github.com/d-torrance/M2/commit/b18f6e470b3b4d9c57edaf8f29528a9d1c9aff4d

mahrud:
> can we make errors be arbitrary expressions, rather than strings?

Unless I'm misunderstanding, I think this is already possible with this PR, eg:
```m2
i2 : new Error from PARA { "oops" }

o2 = PARA{"oops"}

o2 : Error
```
I guess the remaining part is adjusting `processErrorArgs0` in `debugging.dd` to not call `robustPrintString`, but something like `robustPrintError` which calls something like `Thing.RobustPrintErrorMethod` defined in `robust.m2` and can be adjusted based on mode.

I suspect this should be straightforward to do on top of this PR, but my preference is to finalize this PR first.

pzinn:
> Rather than storing the message as an `Expr` and post-processing it into a string based on the top level mode, wouldn't it be equivalent to keep it as a `string` but pre-process the input to `error`? This is essentially what we're doing now with the `processArgs` helper function, but we could make this mode based. And we wouldn't need to touch all the low-level error-printing code that doesn't know what an `Expr` is yet.
> 
> So maybe `error` could look up something like `(class x)#{topLevelMode, PrintError}` and use it to convert its input into a string to store as the message.

I think that would not be so useful -- most error messages come from the interpreter, not from explicit calls to `error`.

mahrud:
I presume this is just a rebase?

d-torrance:
Yup -- fixing the conflicct

mahrud:
@antonleykin @pzinn @mikestillman any objections to merging this for now, with the expectation that there may be future changes as Paul indicated? (And, of course, the possibility of reverting one or both new features before the release if we discover major problems)

pzinn:
fine with me

mikestillman:
I am looking at the final PR, and will review it too, and I will have comments (already do, but going back over the long discussion to make sure I don't re-ask too many questions.  I think I like this version now, but I want to make sure it doesn't slow things down, and that the doc matches the actual behavior (e.g. doc mentions `try code except code else code` but this isn't allowed?)

d-torrance:
Oops -- `try x except y else z` was definitely a typo!  I've fixed it, and also documented the `try/then/except/do` case that Mahrud added.

mikestillman:
@d-torrance Thanks!  Perhaps we should also add to the "error handling" node "stopIfError" (a mutable variable), and a sentence about `trap`?

It seems like other things that I thought were a problem have somehow been fixed since I tried this yesterday? Not sure... (I needed to do `new SelfInitializingType of Error` rather than `new Type of Error`.  Both seem to work now...)

d-torrance:
Sounds good!  I've updated "error handling" as requested.

The only difference between `SelfInializingType` and `Type`, I think, is that the former lets you do `MyError "foo"` but the latter needs `new MyError from "foo"`.

mikestillman:
It seems like we have 5 different forms for try blocks.  Perhaps we should deprecate a couple versions, or give guidance about which should be used?  Which ones are currently used? (That question is for myself, I'll take a look!)

I do like the trap function/keyword though!

mikestillman:
I found the problem: The extended markdown package (well, one of 3 different packages I just disabled) was grabbing shift-enter...  Disabling them, and restarting makes it work.

d-torrance:
There are six forms:
```m2
try x
try x else z
try x then y
try x then y else z
try x except err do z
try x then y except err do z
```

They're all documented in this PR.   I don't see a problem with having this many -- it doesn't hold a candle to all the different forms of `for` statements, for example!

mahrud:
I agree with Doug about different forms of try.

> The extended markdown package

There's a markdown package?!

mahrud:
@mikestillman is this good to merge?

d-torrance:
Discussed with Mike

PR #4053, opened 2025-12-08
Bump Eigen to versions 3.4.0 through 5.0.0

jkyang92:
Eigen changed how their versioning works, so the new version is version 5.0.0. They still use Eigen3 as the name in cmake it seems. See https://libeigen.gitlab.io/news/eigen_5.0.0_released/ for the release, it doesn't seem they made any notable breaking changes, the main change being a bump in the required c++ version which we also satisfy.

A draft pull request for now since I've only looked at the cmake build and I haven't tried testing it with older Eigen versions to make sure that the compile still works. We might also have to update the github actions

### DIFF
--- a/M2/Macaulay2/d/CMakeLists.txt
+++ b/M2/Macaulay2/d/CMakeLists.txt
@@ -166,7 +166,6 @@ target_include_directories(M2-interpreter
     $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/Macaulay2/e/mathic>
     $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/Macaulay2/e/memtailor>
     $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/Macaulay2/e/mathicgb>
-    $<BUILD_INTERFACE:$<TARGET_PROPERTY:Eigen3::Eigen,INTERFACE_INCLUDE_DIRECTORIES>>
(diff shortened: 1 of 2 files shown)


jkyang92:
The last commit was an attempt at a temporary fix, hopefully the github actions all succeed now. There are a couple interconnected issues

- Ubuntu only seems to have Eigen 3.4.0 and not 3.4.1 and in 3.4.1 they they started making changes to how FindEigen3 works.
- Versions below 3.4.1 don't support version ranges for find_package.
- In new versions FindEigen3 no longer defines `EIGEN3_INCLUDE_DIR` so I added the whole library to `target_link_libraries` for M2-interpreter so that it can find the header. This is all just to let `version.d` work

We probably want to give up on older versions of Eigen at some point, but we probably need to get an updated version in Ubuntu first.

As for the last point, I wonder if we're doing `version.d` wrong. At the very least it seems very awkward to have all this work to just include a header to get a version that cmake already theoretically knows. It seems like we should be letting cmake/autotools fill in the versions it found. If we are somehow compiling with a different version than the one we found, we have bigger issues.

d-torrance:
Thanks for looking at this!  This stumped me back in #4026.

> I've only looked at the cmake build

The autotools build already supports the various eigen versions (it just checks for the existence of a header file), so you don't need to worry about it.

d-torrance (approved):
Looks good -- thanks, Jay!

PR #4054, opened 2025-12-09
Bump TBB to 2022.3.0

d-torrance:
Autotools only, since we currently don't have the option of building TBB in the cmake build.

Draft for now to test the GitHub builds.


d-torrance:
Builds passed!

https://github.com/Macaulay2/M2/actions/runs/20050834536

antonleykin (approved):

PR #4056, opened 2025-12-10
MacaulayPosets package for JSAG

aseceleanu:
Introducing package MacaulayPosets which checks properties of posets related to shadow size.


d-torrance:
Converted to a draft so we don't accidentally merge this too soon (which I've done before!)

M2/Macaulay2/packages/MacaulayPosets.m2
```diff
@@ -0,0 +1,1956 @@
...
+    "ringFiberProduct",
+    "ringConnectedSum",
+    "MaxDegree",
+    "MaxTime",
+    "TikZ",
+    "Visual",
+    "AllOrders"
+}
+needsPackage "Posets";
+needsPackage "Visualize";
```
d-torrance: `needsPackage "Posets"` is redundant since it's already in `PackageExports` above, and it would be good to add `Visualize` there instead (or `PackageImports`, depending on whether you want users to also have access to its functionality or not).

M2/Macaulay2/packages/MacaulayPosets.m2
```diff
@@ -0,0 +1,1956 @@
...
+
+PosetMap = new Type of HashTable;
+PosetMap.synonym = "map of posets";
+
+source(PosetMap) := Poset => f -> f.source;
+target(PosetMap) := Poset => f -> f.target;
+
+map(Poset, Poset, HashTable) := PosetMap => opts -> (Q,P,f) -> (
+    -- check that the domain of f is P
+    if not set keys f == set vertices P then error "The keys must be vertices of the source Poset.";
```
d-torrance: Error message in M2 are traditionally lower case.

M2/Macaulay2/packages/MacaulayPosets.m2
```diff
@@ -0,0 +1,1956 @@
...
+    );
+    
+    -- check monotonicity
+    scan(coveringRelations P, c -> if not compare(Q, g(c#0), g(c#1)) then error "The resulting PosetMap would not be monotone.");
+    
+    g
+);
+map(Poset, Poset, List) := PosetMap => opts -> (Q,P,L) -> (
+    -- Either everything in L a vertex of Q,
+    if instance(scan(L, i -> if not isMember(i, vertices Q) then break false), Nothing) then (
```
d-torrance: This can be simplified to `isSubset(L, vertices Q)`

M2/Macaulay2/packages/MacaulayPosets.m2
```diff
@@ -0,0 +1,1956 @@
...
+    M := {};
+    if not isField baseRing R then return "The base ring of the polynomial ring must be a field.";
+    if not isHomogeneousWrtVars I then return "The given ring must be homogeneous.";
+    try R/I then S := R/I else return "The first input needs to be a polynomial ring R, second input needs to be an ideal of R";
+    
+    try super basis(S) then ( -- If the quotient ring has a finite number of elements (ie. the super basis returns a finite number of monomials): 
+        M = super basis(S);
+        M = flatten entries M;
+    ) else ( -- If the quotient ring doesn't have a finite number of elements:
+        M = new MutableList;
```
d-torrance: Do you know what the length of the list will be ahead of time?  The way this is currently implemented (calling `M#(#M)` a bunch), we'll be re-sizing `M` and creating a brand new list every time.

M2/Macaulay2/packages/MacaulayPosets.m2
```diff
@@ -0,0 +1,1956 @@
...
+                );
+            );
+        M = new List from M;
+        );
+    M
+    );
+
+getMons( Ideal  ) := options -> (I) -> ( -- If the input is an ideal.
+    R := ring I;
+    if not isPolynomialRing R then return "The given ideal must be an ideal of a polynomial ring.";
```
d-torrance: This should probably be `error` instead of `return`

M2/Macaulay2/packages/MacaulayPosets.m2
```diff
@@ -0,0 +1,1956 @@
...
+--------------------------------------------------
+
+checkUniqueMin = (poset) -> (
+    flag := true;
+    if ( #(minimalElements poset) == 1 ) then (
+        flag = true; 
+        ) else (
+        flag = false;
+        );
+    flag
```
d-torrance: This function can be simplified to one line: `#(minimalElements poset) == 1`

M2/Macaulay2/packages/MacaulayPosets.m2
```diff
@@ -0,0 +1,1956 @@
...
+    );
+
+getUniqueMin = (poset) -> (
+    if (  checkUniqueMin(poset) == true ) then (
```
d-torrance: The `== true` here is redundant since `checkUniqueMin` returns a boolean

M2/Macaulay2/packages/MacaulayPosets.m2
```diff
@@ -0,0 +1,1956 @@
...
+        el := minimalElements poset
+        );
+    el
```
d-torrance: What happens if `checkUniqueMin poset` returns false?  I suppose there should be an error

M2/Macaulay2/packages/MacaulayPosets.m2
mahrud: You should list these under `PackageImports` in the preamble instead.

M2/Macaulay2/packages/MacaulayPosets.m2
```diff
@@ -0,0 +1,1956 @@
...
+    
+    g
+);
+map(Poset, Poset, List) := PosetMap => opts -> (Q,P,L) -> (
+    -- Either everything in L a vertex of Q,
+    if instance(scan(L, i -> if not isMember(i, vertices Q) then break false), Nothing) then (
+        return( map(Q, P, new HashTable from apply(#P_*, i->P_*#i=>L#i)) );
+    );
+    -- or everything in L is an Option and not everything in L is a vertex of Q
+    if instance(scan(L, i -> if not instance(i, Option) then break false), Nothing) then (
```
d-torrance: This can be simplified to `all(L, i -> instance(i, Option))`

M2/Macaulay2/packages/MacaulayPosets.m2
d-torrance: And if you don't know the length, then `while ... list` can be useful.  It expands the list as needed, but not at every step.  (I think it does something like double the size when it's about to run out of space.)

M2/Macaulay2/packages/MacaulayPosets.m2
```diff
@@ -0,0 +1,1956 @@
...
+        return "The given set is not a subset of the given poset";
+    );
+
+    shadow := {};
+
+    -- For each covering relation k={k#0,k#1} with k#0 in S, the element k#1 is supposed to be in the upper shadow of S.
+    -- For each covering relation k={k#0,k#1} with k#1 in S, the element k#0 is supposed to be in the lower shadow of S.
+    for k in coveringRelations(P) do (
+        if any(S, s->s===k#p) then (
+            shadow = append(shadow, k#(p-1));
```
d-torrance: Using `append` inside a loop like this isn't efficient.  You could do something like:

```m2
for k in coveringRelations P list if any(S, s -> s === k#p) then k#(p - 1) else continue
```

d-torrance (changes requested):
Thanks for the contribution!  I have a number of suggestions for the code.

aseceleanu:
Implemented changes to MacaulayPosets package per Doug's suggestions.

M2/Macaulay2/packages/MacaulayPosets.m2
d-torrance: This `needsPackage "Visualize"` should also be moved to `PackageExports` (or `PackageImports`).

M2/Macaulay2/packages/MacaulayPosets.m2
```diff
@@ -0,0 +1,1913 @@
...
+        
+        posetRanks := {};
+    checkPoset := null;
+    
+    for i from 0 to (#posets - 1) do (
+        checkPoset = posets_i;
+        if (checkUniqueMin(checkPoset) == false or checkUniqueMax(checkPoset) == false ) then (
+            error "all posets must have a unique minimal element and a unique maximal element.";
+            );
+        posetRanks = append(posetRanks, #( rankPoset checkPoset  ) - 1 ); -- Gets the maximum rank of all the given lists.
```
d-torrance: Here's another `append` inside a loop, which isn't efficient since we create another brand new list each time.  This can be simplified down to something like:

```m2
posetRanks := for i from 0 to #posets - 1 list ( ....; #( rankPoset checkPoset ) - 1);
```


There are some other spots like this -- do a search for `append`.  Basically, any time you see something like:

```m2
x = {}
for i in foo do (...; x = append(x, bar))
```

It's better to rewrite it as:

```m2
x = for i in foo list (...; bar)
```

or

```m2
x = apply(foo, i -> (...; bar))
```

d-torrance (changes requested):

aseceleanu:
removed inefficient appends and moved Visualize to PackageImports

aseceleanu:
I am not sure if any more changes are requested and what they are.

d-torrance:
I don't have any additional suggestions at the moment.  Stay tuned for more info from JSAG

antonleykin:
> I am not sure if any more changes are requested and what they are.

We are waiting for more comments. They will appear either here or in a formal (blind) technical review.

M2/Macaulay2/packages/MacaulayPosets.m2
```diff
@@ -0,0 +1,1892 @@
...
+    "ringConnectedSum",
+    "MaxDegree",
+    "MaxTime",
+    "TikZ",
+    "Visual",
+    "AllOrders"
+}
+
+--Some methods use the following variable to avoid taking too long. They first set timeout to several seconds after currentTime(). When currentTime() reaches timeout, macaulayHelper (used by macaulayOrders and isMacaulay) and isAdditive return a String. For isAdditive, time will first be spent in macaulayOrders, and then more time may be spent in the body of isAdditive after macaulayOrders is used.
+timeout=0;
```
seangrate: I would recommend removing the timeout functionality from the methods. This should simplify some of the methods so that, for instance, return types don't have to constantly be checked, and removes some global scope problems. The user could wrap the `isMacaulay(P)` computation in an `alarm` block if they want some timeout functionality.

M2/Macaulay2/packages/MacaulayPosets.m2
```diff
@@ -0,0 +1,1892 @@
...
+macaulayDegree=0;
+
+--------------------------------------------------
+--- Poset maps
+--------------------------------------------------
+
+PosetMap = new Type of HashTable;
+PosetMap.synonym = "map of posets";
+
+source(PosetMap) := Poset => f -> f.source;
```
seangrate: Generally, methods with one parameter don't need parentheses.

M2/Macaulay2/packages/MacaulayPosets.m2
```diff
@@ -0,0 +1,1892 @@
...
+--------------------------------------------------
+
+posetWedgeProduct = Posets -> (
+    posets := splice {Posets};
+    for i from 0 to (#posets - 1) do (
+        if not ( instance(posets_i, Poset)  ) then (
+            error concatenate("all elements in the list must be posets, element number ", toString(i+1), " is not a Poset.");
+            );
+        );
+    for i from 0 to (#posets - 1) do ( -- Checks whether all of the posets have a unique element or not.
```
seangrate: Instead of iterating over the list of posets twice, you can combine these input validations into a single loop.

M2/Macaulay2/packages/MacaulayPosets.m2
```diff
@@ -0,0 +1,1892 @@
...
+        {posetiGroundSet, posetiCoverRels}
+        ));
+    
+    allPosetEls = flatten append(allPosetEls, for i in posetElsAndCovers list i#0);
+    allPosetCovers := for i in posetElsAndCovers list i#1;
+    
+    fiberGroundSet := flatten allPosetEls;
+    fiberCoverRelations := unique flatten allPosetCovers;
+    
+    fiberPoset := poset( fiberGroundSet, fiberCoverRelations );
```
seangrate: Since `fiberPoset` is returned on the next line, you don't have to define it; you can just end the method with the line `poset(fiberGroundSet, fiverCoverRelations)`.

M2/Macaulay2/packages/MacaulayPosets.m2
```diff
@@ -0,0 +1,1892 @@
...
+--- Ring Operations
+--------------------------------------------------
+
+ringFiberProduct = method(Binary=>true);
+ringConnectedSum = method(Binary=>true);
+
+ringProductHelper = (I1, I2, connected) -> (
+    R1 := ring I1;
+    R2 := ring I2;
+    if not ( isPolynomialRing R1 and isPolynomialRing R2 ) then error "both quotient rings must be the quotient of a polynomial ring R and an ideal of R.";
```
seangrate: The convention for M2 errors is for them _not_ to end with punctuation.

M2/Macaulay2/packages/MacaulayPosets.m2
```diff
@@ -0,0 +1,1892 @@
...
+            -- Suppose d1orders is the list of possible orderings of shadows#(s-1). Then, replace each ordering with its possible extensions to an ordering of shadows#s.
+            if true then (
+            extendedOrders := flatten for o in d1orders list (
+                for p in permutations(shadows#s - set shadows#(s-1)) list (
+                    if timeout <= currentTime() then return "timeout";
+                    
+                    oPossible := true;
+                    
+                    -- The initial segment of size s in level d should have an upper shadow no larger than the upper shadow of any other subset of size s in level d.
+                    for A in subsets(levels#d, s) do (
```
seangrate: This is quite a bit of nested logic. I'd recommend trying to refactor this so that, for instance, maybe there is a helper function verifying that the condition is true for each $s$-subset of the $d$-th level, and so on.

M2/Macaulay2/packages/MacaulayPosets.m2
```diff
@@ -0,0 +1,1892 @@
...
+        shadows := flatten append({{{}}}, for s from 1 to #(levels#d) list (
+            -- Let shadows#s be the upper shadow of the initial segment of size s in level d.
+            upperShadow(P,initialSegment(P,O,d,s))
+        ));
+        -- There might also be elements of levels#(d+1) which are not in the upper shadow of levels#d
+        shadows = append(shadows, levels#(d+1));
+        
+        for s from 1 to 1+#(levels#d) do (
+            -- Suppose d1orders is the list of possible orderings of shadows#(s-1). Then, replace each ordering with its possible extensions to an ordering of shadows#s.
+            if true then (
```
seangrate: This always evaluates to `true`, so it is not needed.

M2/Macaulay2/packages/MacaulayPosets.m2
```diff
@@ -0,0 +1,1892 @@
...
+                    while not member(order#j, level) do j = j + 1;
+                    
+                    -- insert a node for this element
+                    TikzPicture = concatenate(TikzPicture, "\tdraw \\node at (", toString(k - (#level-1)/2), ", ", toString(i), ") (", toString(j), ") {", tex order#j, "};\n");
+                );
+            );
+            
+            -- draw cover relations
+            elementToInt := new MutableHashTable; -- As maps, the inverse of order
+            for i from 0 to (#order)-1 do (
```
seangrate: This could be something like
```
elementToInt := hashTable apply(0 ..< #order, i -> order#i => i);
```

M2/Macaulay2/packages/MacaulayPosets.m2
```diff
@@ -0,0 +1,1892 @@
...
+                    newShadow := shadows#(take(level, {i,j})) - set shadows#(take(level, {j+1,#level-1}));
+                    finalNewShadow := shadows#(take(level, {0, j-i})) - set shadows#(take(level, {j-i+1,#level-1}));
+                    
+                    maybeAdditive = maybeAdditive and #initialNewShadow >= #newShadow and #newShadow >= #finalNewShadow;
+                    
+                    if not maybeAdditive then break;
+                );
+                if not maybeAdditive then break;
+            );
+            if not maybeAdditive then break;
```
seangrate: With all the `if not maybeAdditive then break`, I'd try to refactor this using some `all` or `any` functions.

M2/Macaulay2/packages/MacaulayPosets.m2
```diff
@@ -0,0 +1,1892 @@
...
+    Usage
+        posetFiberProduct(L)
+    Inputs
+        L:Sequence
+            a sequence of rank-preserving injective @TO PosetMap@s.
+    Outputs
+        :Poset
+    Description
+        Text
+            In this construction, several posets are glued together along a common subposet. Suppose $(A, \leq_A), (B, \leq_B), (C,\leq C)$ are posets and $\iota_A: C\rightarrow A$ and $\iota_B: C\rightarrow B$ are rank-preserving, injective, and monotone. Then, we have a fiber product \[A\times_C B = (A\setminus\iota_A(C))\sqcup(B\setminus\iota_B(C))\sqcup C\] with $a\geq c$ iff $a\geq_A \iota_A(c)$ for $a\in A$ and $c\in C$, and $b\geq c$ iff $b\geq_B \iota_A(c)$ for $b\in A$ and $c\in C$.
```
seangrate: Some of the documentation lines are quite long. I'd break them apart for readability.

M2/Macaulay2/packages/MacaulayPosets.m2
```diff
@@ -0,0 +1,1892 @@
...
+    y = S_1
+    
+    assert(set getMons(R, I) === set {1_S, x, y, x*y})
+    
+    assert( 31 == #getMons( QQ[x, y], ideal(x^2 - y^2), MaxDegree => 15 ) )
+    
+    x = R_0
+    y = R_1
+    
+    assert (  areIsomorphic( getPoset( R, ideal(x*y - y^2, x^4, x^3*y, x^2*y^2)  ), poset( {1, x, x^2, x^3, y, y^2, y^3}, { {1, x}, {1, y}, {x, x^2}, {x, y^2}, {x^2, x^3}, {x^2, y^3}, {y, y^2}, {y^2, y^3} }  ) ) )
```
seangrate: Having so many constructors and logic on one line is a bit hard to parse. I'd break this into multiple lines so that if this test fails for some reason, then it is a little easier to debug.

M2/Macaulay2/packages/MacaulayPosets.m2
```diff
@@ -0,0 +1,1892 @@
...
+
+-- ringConnectedSum
+assert areIsomorphic( poset({0,1,2,3,4,5},{{0,1},{0,2},{1,3},{2,4},{3,5},{4,5}}), getPoset ringConnectedSum( ZZ/2[x]/monomialIdeal(x^4), ZZ/2[x]/monomialIdeal(x^4) ) );
+assert areIsomorphic( getPoset ringConnectedSum( ZZ/2[x,y]/monomialIdeal(x^3,y^3), ZZ/2[x,y]/monomialIdeal(x^2,y^4) ), posetClosedProduct(product(chain 3, chain 3), product(chain 2, chain 4)));
+///
+
+-- Some wedge and closed product tests.
+TEST ///
+-- Example 2.5 
+assert(   isMacaulay(   posetClosedProduct( getPoset(QQ[x, y, z]/ideal(x*y, x*z, y^2 - y*z, y^2 - z^2, x^3 - y^3  )  ), getPoset(QQ[x, y, z]/ideal(x*y, x*z, y^2 - y*z, y^2 - z^2, x^3 - y^3  )  ))    )     )
```
seangrate: Using this whitespace around the functions is probably an indicator to break the code into a few lines to make it easier to parse.

M2/Macaulay2/packages/MacaulayPosets.m2
```diff
@@ -0,0 +1,1892 @@
...
+assert(   isMacaulay(   posetWedgeProduct( getPoset(QQ[x, y, z]/ideal(x*y, x*z, y^2 - y*z, y^2 - z^2, x^3 - y^3  )  ), getPoset(QQ[x, y, z]/ideal(x*y, x*z, y^2 - y*z, y^2 - z^2, x^3 - y^3  )  )    )    ) == false    )
+///
+
+
+TEST ///
+-- Theorem 2.18
```
seangrate: It would be good to include a reference to which Theorem 2.18 this is referring to.

seangrate (commented):
This looks like it will be a good addition to the Macaulay2 ecosystem. I've left some comments on some style conventions or some suggestions which either make the code more self-consistent or readable.

aseceleanu:
Thank you for the comments. A revised version has been merged above.

d-torrance:
Sean confirmed over Zulip that these changes look good.  Thanks!

PR #4058, opened 2025-12-11
Standalone engine

mikestillman:
This PR is to tease apart the c++ engine from the interpreter code.


mahrud:
Is there a sample c++ code which uses the engine directly?
Also, the `m2-` prefix in the new files in `e/interface` seems redundant, no?

mikestillman:
@d-torrance Should we move this to non-draft?  Any suggestions for renaming the files that @mahrud mentioned? (m2-mem, m2-types?). m2types we could change to interface-types, maybe also mem --> interface-mem.*.  Or we keep them what they are, and the c++ refactoring group at GT in May can perhaps change them...  Thoughts?

d-torrance:
Sure -- I just un-drafted it.

I have no strong opinions on the header names.  I'm fine with letting the refactoring group bike-shed it.  :)

mahrud:
How about my other question? Is there a sample of how to use the engine alone, or is it not there yet?

d-torrance:
The unit tests link against the engine

M2/Macaulay2/d/gmp1.d
```diff
@@ -210,46 +208,24 @@ export format(
      l:int,					   -- max number leading zeroes
      t:int,				    -- max number extra trailing digits
      sep:string,		     -- separator between mantissa and exponent
-     abb:bool,				  -- whether to abbreviate "*ii" to "i"
-     paren:bool,    -- whether to parenthesize a sum and prepend a possible '-'
```
mikestillman, line 214: Were these removed because they were not used?

d-torrance: Correct -- they were only used by the engine, and those routines are now in the engine

M2/Macaulay2/e/interface/m2-types.cpp
```diff
@@ -0,0 +1,61 @@
...
+    return p;
+}
+
+char newline[] = "\n";
+
+int M2_numTBBThreads = 0;
+int M2_gbTrace = 0;
+int M2_numericalAlgebraicGeometryTrace = 0;
+
+struct FUNCTION_CELL *thread_prepare_list;
```
mikestillman, line 55: Maybe we should initialize this to nullptr...

M2/Macaulay2/e/unit-tests/M2mem-replacement.c
```diff
@@ -1,6 +1,6 @@
 #include <stdio.h>
 #include <M2/gc-include.h>
-#include "../d/types.h"
+//#include "../d/types.h"
 
 #include "M2mem-replacement.h"
```
mikestillman: Do we even need this file any longer?

d-torrance: We still include it in a few places in the d directory

M2/Macaulay2/e/interface/m2-types.cpp
d-torrance: Do we even need it?  I don't see where it's ever used

mahrud:
> The unit tests link against the engine

I guess I was hoping for something more realistic, e.g. running an average gb/basis/res computation 100 times, and compare with the same computation in the top level to find the average overhead of the interpreter.

M2/Macaulay2/e/interface/m2-types.cpp
mikestillman: is thread_prepare_list no longer in use?

mikestillman:
@mahrud My main interest at the moment is in the unit tests, and keeping all non-engine code away from that.  Doing what you suggest is be a good thing to do, but I would rather that be done separately later.  But I do agree that your suggestion is a good idea.

edit: we need to clean up the c++ interface before that is easy and intuitive to do...
edit#2: I'm working on this, but in a branch on top of standalone-engine

mikestillman:
@d-torrance I just made some cleanups to unit test cmake and autotools builds.

PR #4059, opened 2025-12-11
Bump versions of mathic, etc.

d-torrance:
I'm in the process of tagging new releases for several of our dependencies for which we're upstream:

- [x] memtailor
- [x] mathic
- [x] mathicgb
- [x] frobby


M2/cmake/build-libraries.cmake
```diff
@@ -39,10 +39,10 @@ endforeach()
 string(REPLACE ";" " " COMPILEFLAGS "${COMPILE_OPTIONS}")
 
 # C compiler flags
-set(CFLAGS   "${COMPILEFLAGS} -w -Wimplicit -Werror")
+set(CFLAGS   "${COMPILEFLAGS} -w -Wimplicit -Werror ${CMAKE_C_FLAGS}")
 
 # C++ compiler flags
-set(CXXFLAGS "${COMPILEFLAGS} -std=gnu++11 -w -Wno-mismatched-tags -Wno-deprecated-register")
+set(CXXFLAGS "${COMPILEFLAGS} -std=gnu++11 -w -Wno-mismatched-tags -Wno-deprecated-register ${CMAKE_CXX_FLAGS}")
```
d-torrance: @mahrud -- Is there a better way to get `CXXFLAGS` from the environment and pass them to the various library builds?  (We need them for frobby, and I imagine some other C++ dependencies, when building with brew clang or we get a linking error.)

Also, do we still need the various hard-coded flags?

d-torrance:
Frobby build passed: https://github.com/Macaulay2/M2/actions/runs/21659331145

d-torrance:
Discussed with Mike

PR #4063, opened 2025-12-15
Update Python int <-> ZZ conversion for Python 3.14 from gmpy2

d-torrance:
`_PyLongNew` (which we were using for our Python `int` <-> `ZZ` conversion) was deprecated in Python 3.14 and is being replaced with new import/export functions from [PEP 747](https://peps.python.org/pep-0757/).  This results in the following compiler warnings:

```c
CC python-c.c
../../../../Macaulay2/d/python-c.c:157:18: warning: '_PyLong_New' is deprecated [-Wdeprecated-declarations]
  157 |   if (!(result = _PyLong_New(size))) {
      |                  ^
/opt/homebrew/opt/python@3.14/Frameworks/Python.framework/Versions/3.14/include/python3.14/cpython/longintrepr.h:103:1: note: '_PyLong_New' has been explicitly marked deprecated here
  103 | Py_DEPRECATED(3.14) PyAPI_FUNC(PyLongObject*) _PyLong_New(Py_ssize_t);
      | ^
/opt/homebrew/opt/python@3.14/Frameworks/Python.framework/Versions/3.14/include/python3.14/pyport.h:281:54: note: expanded from macro 'Py_DEPRECATED'
  281 | #define Py_DEPRECATED(VERSION_UNUSED) __attribute__((__deprecated__))
      |                                                      ^
1 warning generated.
```

gmpy2 solved this problem in https://github.com/gmpy2/gmpy2/pull/495, so we adapt their code.  We also add a new header file from https://github.com/python/pythoncapi-compat to support older versions of Python.  It even works with Python 3.6 in Ubuntu 18.04!

### DIFF
--- a/M2/Macaulay2/d/python-c.c
+++ b/M2/Macaulay2/d/python-c.c
@@ -1,5 +1,6 @@
 #include <Python.h>
 #include "python-exports.h"
+#include "pythoncapi_compat.h"
 
 #include <gmp.h>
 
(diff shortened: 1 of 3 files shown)


antonleykin (approved):
It seems fine... although I'm not a Python expert.

Issue #4016, opened 2025-11-07
complexity of `entries` grows quadratically in number of rows


mahrud:
I did not expect this:
```m2
i1 : elapsedTime H = random(ZZ^200000, ZZ^10);
 -- 1.75006s elapsed

              200000       10
o1 : Matrix ZZ       <-- ZZ

i5 : elapsedTime entries H^{0..10000};
 -- 1.87336s elapsed

i6 : elapsedTime entries H^{0..20000};
 -- 7.24654s elapsed

i7 : elapsedTime entries H^{0..30000};
 -- 41.3092s elapsed

i8 : elapsedTime flatten apply(3, i -> entries H^{i*10000..(i+1)*10000-1});
 -- 7.09806s elapsed
```
In particular, this seems to be about number of rows only:
```m2
i14 : elapsedTime H = transpose H;
 -- .277403s elapsed

               10       200000
o14 : Matrix ZZ   <-- ZZ

i15 : elapsedTime entries H;
 -- 2.09016s elapsed
```

mahrud:
XD
```m2
i1 : elapsedTime H = random(ZZ^200000, ZZ^10);
 -- 1.59819s elapsed

              200000       10
o1 : Matrix ZZ       <-- ZZ

i2 : elapsedTime transpose entries transpose H;
 -- 2.64812s elapsed
```

jkyang92:
You might already know this, but I looked at the code, the implementation of entries is silly. 

In particular, the code is as follows:

```
entries RawMutableMatrix := entries RawMatrix := m -> table(rawNumberOfRows m,rawNumberOfColumns m,(i,j)->rawMatrixEntry(m,i,j))1
```

But each call to `rawMatrixEntry` calls `IM2_Matrix_get_entry`. For mutable matrices this is fine since they are dense, but for a normal matrix, the sparse representation means that we have to walk through each column repeatedly to find the relevant entry. The complexity is something like O(n^2) where n is the maximum number of entries per column.

The right thing to do is to have the engine populate the array directly. I can give this a try if you're not already working on it.

mahrud:
All yours!

PR #4064, opened 2025-12-15
Improve the implementation of entries

jkyang92:
This implements functions `IM2_Matrix_get_entries` and `IM2_MutableMatrix_get_entries` that gives an array with all of the entries and uses these to implement entries. This fixes the quadratic in the number of rows behavior seen in #4016. Strictly this change is only needed for `Matrix` and not `MutableMatrix`, and I have not yet tested whether there is an improvement for `MutableMatrix`, since this version does involve an extra copy. Additionally, for `MutableMatrix` we are still calling `get_entry` entry by entry on the matrix itself when we should in theory be able to loop over an array somewhere.

There is a slight design compromise in this, as top level expects a list of lists, but it is not easy to return a list of lists from the interface level, we return just a single sequence and let the top level code rearrange that into a list of lists. This rearrangement could in theory happen at the "d" level, but it's not clear if that's faster.


mahrud:
How does this interact with Mike's PR for separating engine from interpreter?

jkyang92:
I hadn't looked at it, but the short answer is it's a bit annoying, but it's just one more type to add. I feel like the interface folder should be treated differently than the rest of the engine though. I'll look at that pull request a bit more and make some comments.

mahrud:
How so? I imagine the point is that a standalone program would only include some headers from the interface directory and link with the engine library in order to use the engine.

jkyang92:
I assume you're replying to the comment about the interface directory, I would rather the engine have a proper c++ api, rather than the pieced together API that is the interface directory. To me the interface directory exists mostly to translate the engine code to an API compatible with the interpreter. In particular I don't like the interfaces/m2-types.h file, but I don't have a better solution right now.

The actual changes in relation to mike's pull request is just to add one more type to interface/m2-types.h.

jkyang92:
I think this is ready now. This version unlike the initial version for this pull request does return an array of arrays of raw ring elements from the engine. This necessitated the definition of a new type in d, `RawRingElementArrayArray`. This is because arrays in the d language only use a 32bit length field, even on 64bit systems. But it's not too hard to create an matrix with more than 2^32 total elements which would cause incorrect return values from entries.

It might be worth considering making the length field of the d arrays a `size_t`. However, this is quite involved as there are many places in the engine that expect it to be an int.

Also, as @mahrud brought up, this has a small effect on #4058 because it adds one more type that the engine and the interpreter both need to know about.

M2/Macaulay2/m2/engine.m2
```diff
@@ -353,7 +353,10 @@ target RawMatrix := o -> rawTarget o
 source RawMatrix := o -> rawSource o
 transposeSequence := t -> pack(#t, mingle t)
 isHomogeneous RawMatrix := rawIsHomogeneous
-entries RawMutableMatrix := entries RawMatrix := m -> table(rawNumberOfRows m,rawNumberOfColumns m,(i,j)->rawMatrixEntry(m,i,j))
+entries RawMutableMatrix :=
+entries RawMatrix := m -> (
+    toList(apply(rawMatrixEntries m, row -> toList row))
```
d-torrance: Would it make sense to build these as lists at the d level, i.e., have `rawMatrixEntries` return a list of lists instead of a sequence of sequences?

jkyang92: Done, it's a tiny bit awkward because of how toExpr for a RawRingElementArray works.

d-torrance (approved):
This is great!

### Before

```m2
i1 : scan({1,2,3}, i -> (H := random(ZZ^(i*10000), ZZ^10); elapsedTime entries H))
 -- 1.30147s elapsed
 -- 11.2444s elapsed
 -- 65.9728s elapsed
```

### After
```m2
i1 : scan({1,2,3}, i -> (H := random(ZZ^(i*10000), ZZ^10); elapsedTime entries H))
 -- .0862296s elapsed
 -- .166589s elapsed
 -- .210105s elapsed
```

Thanks, Jay!

mikestillman:
This is awesome, thanks Jay!

PR #4065, opened 2025-12-17
Use elements of base ring for coefficients in random(Module, Module)

d-torrance:
Currently, the code for `random(Module, Module)` in the case where the differences in degrees aren't consistent (which has been untouched since 1999!) assumes we're working over the integers, regardless of the base ring, and thus spits out matrices with integer coefficients:

```m2
i1 : R = QQ[x];

i2 : random(R^2, R^{-1, -2})

o2 = | 6x 7x2 |
     | 6x 7x2 |

             2      2
o2 : Matrix R  <-- R

i3 : S = CC[x];

i4 : random(S^2, S^{-1, -2})
-- warning: experimental computation over inexact field begun

o4 = | 8x 2x2 |
     | 6x 4x2 |

             2      2
o4 : Matrix S  <-- S
```

We update it to use random elements of the base ring as coefficients:

```m2
i1 : R = QQ[x];

i2 : random(R^2, R^{-1, -2})

o2 = | 1/7x 1/10x2 |
     | 3/8x 3/5x2  |

             2      2
o2 : Matrix R  <-- R

i3 : S = CC[x];

i4 : random(S^2, S^{-1, -2})
-- warning: experimental computation over inexact field begun

o4 = | (.478476+.798072ii)x (.550045+.323718ii)x2 |
     | (.994119+.640657ii)x (.193998+.214085ii)x2 |

             2      2
o4 : Matrix S  <-- S
```

Note that this currently breaks this method for modules over `RRi` since `random RRi` isn't defined, but that will be fixed by #4040.

### DIFF
--- a/M2/Macaulay2/m2/genmat.m2
+++ b/M2/Macaulay2/m2/genmat.m2
@@ -120,8 +120,6 @@ random(Module, Module) := Matrix => opts -> (F,G) -> (
      R := ring F;
      if R =!= ring G then error "modules over different rings";
      if opts.MaximalRank then return (randomMR opts)(F,G);
-     p := char R;
-     if p === 0 then p = opts.Height;
      degreesTable := table(degrees F, degrees G, 
 	  (i,j) -> toList apply(j,i,difference));
      degreesTally := tally flatten degreesTable;
@@ -144,7 +142,7 @@ random(Module, Module) := Matrix => opts -> (F,G) -> (
 	       deg -> (
 		    numused := 0;
 		    if deg === 0 then (
-			 n := apply(degreesTally#deg, x -> random p);
+			 n := apply(degreesTally#deg, x -> random(R.BaseRing, opts));
 			 () -> (
 			      r := n#numused;
 			      numused = numused + 1;
@@ -158,7 +156,7 @@ random(Module, Module) := Matrix => opts -> (F,G) -> (
 			      n = first entries (
 				   m * matrix (R, table(
 					     k, degreesTally#deg, 
-					     (i,j)->random p)));
+					     (i,j)->random(R.BaseRing, opts))));
 			      () -> (
 				   r := n#numused;
 				   numused = numused + 1;


mahrud:
This caught my eye because @Devlin-Mallory and I have had lots of trouble with `random(Module, Module)`. Two things to note:
1. I doubt it, but it's worth checking that this doesn't slow down `random` significantly.
2. I can see a reasonable argument that we basically only ever want integer coefficients, because `random(M, N) * random F` can easily give you denominators if you really want them, but if you don't want denominators clearing them is a pain.

Is there a specific reason to change this behavior?

d-torrance:
1.  All the `random` methods for the base rings call the engine directly after an initial lookup, so they're pretty quick:

```m2
i1 : code (random, Type)

o1 = -- code for method: random(Type)
     /usr/share/Macaulay2/Core/methods.m2:309:15-312:16: --source code:
     random Type := opts -> (X) -> (
          m := lookup(symbol random,X);
          if m === null then error "no method found";
          (m opts) X)

i2 : code ZZ.random

o2 = /usr/share/Macaulay2/Core/integers.m2:43:12-43:49: --source code:
     ZZ.random = opts -> ZZ -> rawRandomZZ opts.Height

i3 : code QQ.random

o3 = /usr/share/Macaulay2/Core/rationals.m2:32:12-32:49: --source code:
     QQ.random = opts -> QQ -> rawRandomQQ opts.Height

i4 : code RR'.random

o4 = /usr/share/Macaulay2/Core/reals.m2:260:13-260:56: --source code:
     RR'.random = opts -> R -> rawRandomRRUniform R.precision

i5 : code CC'.random

o5 = /usr/share/Macaulay2/Core/reals.m2:261:13-261:49: --source code:
     CC'.random = opts -> C -> rawRandomCC C.precision
```

The time difference is negligible:

### Before
```m2
i2 : elapsedTime scan(1000, i -> random(R^2, R^{-1, -2}))
 -- 1.95864s elapsed
```

### After
```m2
i2 : elapsedTime scan(1000, i -> random(R^2, R^{-1, -2}))
 -- 1.86777s elapsed
```

2.  If that's true, then we should probably switch the behavior for `random(Module, Module)` when the degree differences *are* consistent, because they currently don't give integer coefficients.  For example, here's the current behavior:

```m2
i1 : R = QQ[x];

i2 : random(R^2, R^{2:-1})

o2 = | 8/9x  1/7x |
     | 10/7x 5/8x |

             2      2
o2 : Matrix R  <-- R

i3 : S = CC[x];

i4 : random(S^2, S^{2:-1})
-- warning: experimental computation over inexact field begun

o4 = | (.391977+.78413ii)x  (.343914+.525869ii)x |
     | (.232219+.747771ii)x (.172887+.438251ii)x |

             2      2
o4 : Matrix S  <-- S
```

The rationale was consistency between the two cases.

mahrud:
I agree that they should be consistent. Moreover, your example over C shows that the current behavior is not redeemable with a simple scalar multiplication, and it's much worse than I thought:
```m2
i27 : S = (GF 16)[x];

i28 : tally apply(100, i -> random(S^2, S^{-1,-2}))

o28 = Tally{| 0 0  | => 6}
            | x x2 |
            | 0 0  | => 7
            | 0 x2 |
            | 0 0 | => 9
            | x 0 |
            | 0 x2 | => 3
            | 0 x2 |
            | 0 x2 | => 6
            | x 0  |
            | 0 x2 | => 6
            | x x2 |
            | 0 x2 | => 8
            | 0 0  |
            | x 0  | => 7
            | 0 x2 |
            | x 0  | => 9
            | x x2 |
            | x 0 | => 7
            | x 0 |
            | x 0 | => 8
            | 0 0 |
            | x x2 | => 3
            | x 0  |
            | x x2 | => 5
            | 0 0  |
            | x x2 | => 6
            | x x2 |
            | x x2 | => 7
            | 0 x2 |
            0 => 3

o28 : Tally
```

mahrud (approved):

antonleykin (approved):
Looks logical to me.

PR #4070, opened 2025-12-18
Unify `Position` and `FilePosition`

d-torrance:
We currently move back and forth between the low-level `Position` type in the interpreter, which is a struct containing filename, line, and column info, and the top-level `FilePosition` one, which is a basic list containing the same information.

We unify these two, so that each `FilePosition` object just contains a  corresponding `Position` object under the hood.

This does break some code, as we can no longer use `#` to access information from a `FilePosition` object.  But we can use `toSequence` or `toList` and get the information from their output.

This is a draft for now, as there are some conflicts with #4051.


pzinn:
I'm a little confused. what is `toString FilePosition` supposed to call?

d-torrance:
> I'm a little confused. what is `toString FilePosition` supposed to call?

It calls `toSimpleString`, which calls `tostring(Position)` in `stdiop.d`.  This has the same behavior as the current top-level `toString(FilePosition)`.

d-torrance:
Closing -- there are a bunch of conflicts now that I'd rather not deal with.  It may return in another form at some point...

PR #4072, opened 2025-12-21
Fix robust printing of subclasses of String

### LINKED ISSUE #4069: robustness of key not found error for subclasses
While testing #4051, I noticed a few issues:

First, I can't explain the difference between these two error outputs:
```m2
i1 : T = new Type of String;

i2 : ZZ#(concatenate(100:"x"))
stdio:2:2:(3): error: key not found in hash table:
        "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx. (of class String)

i3 : ZZ#(new T from concatenate(100:"x"))
stdio:3:2:(3): error: key not found in hash table
```
I would imagine they should be identical except the type should say "T" in the latter. I did discover the following workaround, but I think it should be unnecessary:
```m2
i4 : toString T := x -> (print "toString"; net "HO");

i5 : ZZ#(new T from concatenate(100:"x"))
toString
stdio:5:2:(3): error: key not found in hash table:
        HO (of class T)
```

Second, I think this error printing is escaping the robust error printing time limits:
```m2
i9 : elapsedTime ZZ#(new T from concatenate(100:"x"))
toString
stdio:9:14:(3): error: key not found in hash table:
        HO (of class T)
 -- 10.0006s elapsed
```

d-torrance:
Both `net` (which calls `separate`) and `format` were broken because the associated compiled functions didn't deal with the `SpecialExpr` case, which is what we have when `String` is subclassed.

This should at least partially fix #4069

### Before
```m2
i1 : T = new SelfInitializingType of String;

i2 : T "foo"

--error or time limit reached in applying Wrap method to output; type 'debugError()' to see it


o2 = stdio:2:1:(3): error: at print: expected argument 2 to be a sequence or list of strings, nets, or nulls

i3 : x = new HashTable; x#(T "foo")
stdio:3:20:(3): error: key not found in hash table
```

### After
```m2
i1 : T = new SelfInitializingType of String;

i2 : T "foo"

o2 = foo

i3 : x = new HashTable; x#(T "foo")
stdio:4:20:(3): error: key not found in hash table:
        "foo" (of class T)
```

### DIFF
--- a/M2/Macaulay2/d/actors4.d
+++ b/M2/Macaulay2/d/actors4.d
@@ -1095,6 +1095,7 @@ setupfun("connectionCount", connectionCount);
 format(e:Expr):Expr := (
      when e
      is s:stringCell do toExpr("\"" + present(s.v) + "\"")
+     is s:SpecialExpr do format(s.e)
      is ZZcell do e
      is QQcell do e
      is RRcell do format(Expr(Sequence(e)))
--- a/M2/Macaulay2/d/regex.dd
+++ b/M2/Macaulay2/d/regex.dd
@@ -230,6 +230,8 @@ regexSeparate(e:Expr):Expr := (
 	    toExpr(r))
 	else WrongArgZZ(4)
 	else WrongArgZZ(3)
+	-- for subclasses of string
+	is text:SpecialExpr do regexSeparate(Expr(Sequence(s.0, text.e, s.2, s.3)))
 	else WrongArgString(2)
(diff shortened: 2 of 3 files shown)


mahrud (approved):
Good catch, thank you!
Just a thought: would any of the additions need typical value comments so they're listed as methods?

d-torrance:
I don't think so -- both are overwritten with actual method functions at top level

PR #4078, opened 2026-01-02
Update A1BrouwerDegrees with citation information

wgabrielong:
Added citation information following #3797

### DIFF
--- a/M2/Macaulay2/packages/A1BrouwerDegrees.m2
+++ b/M2/Macaulay2/packages/A1BrouwerDegrees.m2
@@ -256,7 +256,7 @@ Node
             getSumDecomposition beta
             gamma = makeGWuClass(matrix(QQ, {{11, 0},{0,22}}), 8)
             isIsomorphicForm(beta, gamma)
-	Citation
+    Citation
 	@article{zbMATH08005952,
 	 author = {Borisov, Nikita and Brazelton, Thomas and Espino, Frenly and Hagedorn, Thomas and Han, Zhaobo and Lopez Garcia, Jordy and Louwsma, Joel and Ong, Wern Juin Gabriel and Tawfeek, Andrew R.},
 	 title = {{{\(\mathbb{A}^1\)}}-Brouwer degrees in {Macaulay2}},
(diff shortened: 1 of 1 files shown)


M2/Macaulay2/packages/A1BrouwerDegrees.m2
```diff
@@ -256,6 +256,21 @@ Node
             getSumDecomposition beta
             gamma = makeGWuClass(matrix(QQ, {{11, 0},{0,22}}), 8)
             isIsomorphicForm(beta, gamma)
+		Citation
```
d-torrance: The indentation here is off -- it thinks it's part of the example code.  `Citation` should be at the same level as `Key`, `Headline`, etc.

d-torrance (changes requested):

M2/Macaulay2/packages/A1BrouwerDegrees.m2
```diff
@@ -256,6 +256,21 @@ Node
             getSumDecomposition beta
             gamma = makeGWuClass(matrix(QQ, {{11, 0},{0,22}}), 8)
             isIsomorphicForm(beta, gamma)
+	Citation
```
d-torrance: It's still a bit off.  Right now it's inside `Description`, lining up with `Example` and `Text`.  But it should a bit sooner, lining up with `Description`, etc.

d-torrance (changes requested):

M2/Macaulay2/packages/A1BrouwerDegrees.m2
wgabrielong: On my screen it is one back from "example" which is in the same line as key/headline. Is it that it should be moved up?

d-torrance: It's a tabs v. spaces issue.  There's one tab character before `Citation`, and on your system tabs must be set to take up 4 spaces, which would make it look like it lines up correctly.  But M2 must be interpreting that tab as 8 spaces, resulting in the wrong indentation level.

Here's what it looks like for me (I have tabs set up to be a slightly different color in Emacs):

image


A simple fix would be to delete the tab character in front of `Citation` and put in 4 spaces.

d-torrance (approved):
Looks great -- thanks!

wgabrielong:
Thanks, @d-torrance.

PR #4080, opened 2026-01-06
Fix garbage collection of foreign gmp/mpfr objects

d-torrance:
The GC finalizer callback function expects two parameters: a pointer to the GC memory we're finalizing and a pointer to some client data. We were previously passing the *first* object to `mpz_clear` and `mpfr_clear`, which was incorrect.  (This is why the mpfr code was commented out, as otherwise we were segfaulting trying to call `mpfr_clear` on our pointer object rather than on the `mpfr_ptr` it's pointing to!.  I'm surprised the gmp code wasn't segfaulting too!)

### DIFF
--- a/M2/Macaulay2/d/ffi.d
+++ b/M2/Macaulay2/d/ffi.d
@@ -250,6 +250,10 @@ ffiIntegerType(e:Expr):Expr := (
     else WrongNumArgs(2));
 setupfun("ffiIntegerType", ffiIntegerType);
 
+gmpFinalizer(obj:voidPointer, clientData:voidPointer):void := (
+    Ccode(void, "(void)", obj);
+    Ccode(void, "mpz_clear((mpz_ptr)", clientData, ")"));
+
 -- returns pointer to integer object with given value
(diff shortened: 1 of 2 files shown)


mahrud:
I'm a bit confused, did mpfr just not finalize at all before? Is it possible to add some sort of test or memory benchmark to compare with previous performance?

d-torrance:
That's correct -- I guess I gave up too soon trying to figure out why it was segfaulting...

I just added a test that creates a bunch of these objects and then collects the garbage.  Is that the kind of thing you were asking for?  I'm not sure of a good way to check if we actually finalized the things we wanted to.

jkyang92:
This applies to ffi related objects only right? Everything else the engine does that I know of reallocates the limbs into gc memory.

d-torrance:
Yes, that's correct.  Only objects of class `mpzT` and `mpfrT` constructed using the `ForeignFunctions` package.

d-torrance:
Discussed with Mike

PR #4081, opened 2026-01-07
Fix mathML(Type) -- previously caused a recursion error

d-torrance:
Currently, we have:

```m2
i1 : mathML Matrix
stdio:1:6:(3): error: recursion limit of 300 exceeded
```

What's going on?

```m2
i2 : code (mathML, Type)

o2 = -- code for method: mathML(Type)
     /usr/share/Macaulay2/Core/mathml.m2:167:24-167:80: --source code:
     mathML ImmutableType := R -> if R.?mathML then R.mathML else mathML expression R

i3 : peek expression Matrix

o3 = Holder{Matrix}

i4 : code (mathML, Holder)

o4 = -- code for method: mathML(Holder)
     /usr/share/Macaulay2/Core/mathml.m2:40:17-40:32: --source code:
     mathML Holder := v -> mathML v#0
```

Since there's no `Matrix.mathML`, we just bounce back and forth between `mathML(Type)` and `mathML(Holder)`.

We fix this by just removing `mathML(Type)` and using inheritance so it calls `mathML(MutableHashTable)` ~~(and also setting this to be `mathML(ImmutableType)`)~~, which does the expected thing:

```m2
i1 : mathML Matrix

o1 = <mi>Matrix</mi>
```

### DIFF
--- a/M2/Macaulay2/m2/enginering.m2
+++ b/M2/Macaulay2/m2/enginering.m2
@@ -274,6 +274,7 @@ expression EngineRing := R -> if hasAttribute(R,ReverseDictionary) then expressi
 toString EngineRing := toString @@ expression
 net      EngineRing :=      net @@ expression
 texMath  EngineRing :=  texMath @@ expression
+mathML   EngineRing :=   mathML @@ expression
 
 ZZ _ EngineRing := 
 RR _ EngineRing :=
--- a/M2/Macaulay2/m2/mathml.m2
+++ b/M2/Macaulay2/m2/mathml.m2
@@ -86,6 +86,12 @@ mathML Minus := v -> concatenate( "<mrow><mo>-</mo>", mathML v#0, "</mrow>")
 mathML Divide := x -> concatenate("<mfrac>", mathML x#0, mathML x#1, "</mfrac>")
 mathML OneExpression := x -> "<mn>1</mn>"
 mathML ZeroExpression := x -> "<mn>0</mn>"
+mathML BinaryOperation := m -> (
+    x := mathML m#1;
+    y := mathML m#2;
+    if rightPrecedence m#1 < lprec m#0 then x = mathMLparen x;
+    if precedence m#2 <= rprec m#0 then y = mathMLparen y;
+    mrow(x | mo m#0 | y))
 mathML Sum := v -> (
      n := # v;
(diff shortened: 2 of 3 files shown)


d-torrance:
This (kind of) broke `mathML` for polynomial rings.  This is the current behavior:

```m2
i1 : mathML(QQ[a,c,e])

o1 = <mi>ℚ</mi> <mrow><mo>[</mo><mrow><mi>a</mi><mo>,</mo><mi>c</mi><mo>,</mo><
     mi>e</mi></mrow><mo>]</mo></mrow>
```

Which renders as:

<math>
<mi>ℚ</mi> <mrow><mo>[</mo><mrow><mi>a</mi><mo>,</mo><mi>c</mi><mo>,</mo><mi>e</mi></mrow><mo>]</mo></mrow>
</math>

(Yay GitHub can render mathml in comments!)

But notice that I used variable names that aren't adjacent alphabetically.  When we do that:

```m2
i2 : mathML(QQ[a,b,c])
stdio:2:6:(3): error: mathML conversion for expression class BinaryOperation not implemented yet
```

That's because `expression` inserts a `..` operator between `a` and `c`, and `mathML` chokes on that.

With just the first commit from this PR, we get the following silly thing:

```m2
i1 : mathML(QQ[a,b,c])

o1 = <mrow><mi>PolynomialRing</mi><mtext>{...55...}</mtext></mrow>
```

So we add `mathML(EngineRing)` and `mathML(BinaryOperation)`, giving us the expected result:

```m2
i1 : mathML(QQ[a,b,c])

o1 = <mi>ℚ</mi> <mrow><mo>[</mo><mrow><mrow><mi>a</mi><mo>..</mo><mi>c</mi></
     mrow></mrow><mo>]</mo></mrow>
```

This renders as:

<math><mi>ℚ</mi> <mrow><mo>[</mo><mrow><mrow><mi>a</mi><mo>..</mo><mi>c</mi></mrow></mrow><mo>]</mo></mrow></math>

d-torrance:
I ended up just dropping `mathML(ImmutableType)` completely.  It was redundant since `Module` is the only child of `ImmutableType` and `mathML(Module)` is defined elsewhere.  None of the other text formatting methods are installed for `ImmutableType` either.

d-torrance:
Discussed with Mike

PR #3567, opened 2024-11-01
Varieties v0.3

mahrud:
This is a repeat of #3517, which is on top of #3550 and shares a exact few commits with #3526, but I don't believe there should be a conflict.

PR #4084, opened 2026-01-08
Switch ~ from postfix -> prefix/binary

d-torrance:
This has been discussed before (e.g., https://github.com/Macaulay2/M2/pull/3183#issuecomment-2059599972).  The original reason for it being a postfix operator was sheafification, but that has used `^~` since #3567.

As a prefix unary operator, it makes much more sense for the remaining methods (bitwise negation of integers).  We also add it as a binary operator for potential other uses such as equivalence relations (hence the choice to use the same precedence as `==` and friends) or building formulas in the *RInterface* package.

Draft for now for discussion since this is a breaking change.


d-torrance:
Discussed with Mike

Issue #3930, opened 2025-07-15
Use generateGrammar to generate Macaulay2Web symbols


d-torrance:
Now that the `generateGrammar` function is available in the *Style* package, we could use it to generate `M2-symbols.ts` for Macaulay2Web in the Macaulay2Web repository rather than in the `editors` directory.  See #3825.

d-torrance:
Fixed by pzinn/Macaulay2Web#41 and #4085

PR #4085, opened 2026-01-10
Stop generating Macaulay2Web autocompletion symbols

d-torrance:
This is now handled in the Macaulay2Web repository (see https://github.com/pzinn/Macaulay2Web/pull/41).

Closes: #3930


mahrud (approved):
I skipped the builds since nothing would be tested anyway.

PR #4086, opened 2026-01-11
Improve thread safety of value and readeval

### LINKED ISSUE #3927: value is not thread-safe
After lots of confusion I think I tracked down the issue that makes lots of things not thread safe (including basis and truncate, which is possibly related to #3439). `value` can't handle symbols in parallel:
```m2
i1 : R = QQ[x];

i2 : taskResult \ apply(5, i -> schedule(() -> value "x"))
../linuxbrew/.linuxbrew/Cellar/macaulay2/1.25.06_2/share/Macaulay2/Core/debugging.m2:18:13:(1):[5]: error: mutable unexported unset symbol(s) in package User: 'T'
null: here is the first use of 'T'
../linuxbrew/.linuxbrew/Cellar/macaulay2/1.25.06_2/share/Macaulay2/Core/packages.m2:491:15:(1):[4]: --back trace--
../linuxbrew/.linuxbrew/Cellar/macaulay2/1.25.06_2/share/Macaulay2/Core/packages.m2:289:27:(1):[3]: --back trace--
../linuxbrew/.linuxbrew/Cellar/macaulay2/1.25.06_2/share/Macaulay2/Core/methods.m2:458:27:(1):[2]: --back trace--

o2 = {x, , x, x, x}

o2 : List
```

Strangely, `value` is simply a wrapper around the compiled function `value'`, and using `value'` above we also get the same crash, so this isn't even a top-level issue.

jkyang92:
This makes several variables thread local so that value and readeval in interp.dd is more thread safe. This seems to fix #3927.

There are a couple more variables that are worth investigating:

- `previousLineNumber` in interp.dd I think this is only used by the interactive prompt, and so it is probably fine either as thread local or global
- `intepreterDepth`, `errorDepth`, and `loadDepth`. These are somewhat related, but currently only `loadDepth` is thread local
- `current`, this seems to be only used by the debugger. Similarly to `previousLineNumber` it's probably fine for it to be either thread local or global, but we should think about it.

### DIFF
--- a/M2/Macaulay2/d/interp.dd
+++ b/M2/Macaulay2/d/interp.dd
@@ -16,10 +16,10 @@ export setupargv():void := (
      setupconst("environment",toExpr(envp));
      );
 
-fileExitHooks := setupvar("fileExitHooks", Expr(emptyList));
-currentFileName := setupvar("currentFileName", nullE);
-currentPosFile  := dummyPosFile;
-currentFileDirectory := setupvar("currentFileDirectory", toExpr("./"));
+fileExitHooks := setupvarThread("fileExitHooks", Expr(emptyList));
+currentFileName := setupvarThread("currentFileName", nullE);
+threadLocal currentPosFile  := dummyPosFile;
+currentFileDirectory := setupvarThread("currentFileDirectory", toExpr("./"));
 update(err:Error,prefix:string,f:Code):Error := (
      if !err.printed
      then printErrorMessage0(f,prefix + ": " + err.message)
@@ -514,7 +514,7 @@ debugger(f:Frame,c:Code):Expr := (
      ret);
 debuggerpointer = debugger;
 
-currentString := setupvar("currentString", nullE);
+currentString := setupvarThread("currentString", nullE);
 value(e:Expr):Expr := (
      when e
      is q:SymbolClosure do (


mahrud:
Thanks for this! I definitely prefer it over my solution. Do you mind adding this as a test somewhere?
```m2
R = QQ[x];
assert same(taskResult \ apply(100, i -> schedule(() -> value "x")))
```
Also, if you get a chance, I would appreciate help with #3928/#4017 :)

jkyang92:
The semantics of the `currentFileName` and `currentFileDirectory` are hard to get right. My understanding is they should only change when a file is loaded, but then we have to set the correct answer when we call schedule. In particular, it's not immediately clear what the correct output of `schedule(() -> (print currentFileName))` should be. The obvious answer is whatever `currentFileName` and `currentFileDirectory` were when `schedule` was called, but that requires storing them when calling schedule.

We could also have these variables only be modified by the main thread. This would probably break loadPackage in threads. Whether calling `loadPackage` in threads is a good idea another question.

mahrud:
@d-torrance I think this is good to merge! I've been using it for a while with no issues.

d-torrance:
@jkyang92 - This is still marked as a draft.  What do you think?  Is it ready?

jkyang92:
I'll try to get the issue with `currentFileName` fixed first. It's going to have to a bit of a hacky fix but I can probably find some time to do that during the internals work this week. If I don't get to it by the end of the week, I can un-draft this and it mostly works.

jkyang92:
After poking around at the code again, I've come to the conclusion that it's fine as it is. Having `currentFileName` and `currentFileDirectory` be null in threads isn't great, but it actually doesn't seem to break anything important.

PR #4090, opened 2026-01-13
Update setupEmacs() so that it builds a package

d-torrance:
We now build a tarball and install it using `package-install-file`. This way, the package files will be byte-compiled and it will appear with the list of other packages that users install from melpa, etc.

### DIFF
--- a/M2/Macaulay2/m2/files.m2
+++ b/M2/Macaulay2/m2/files.m2
@@ -301,13 +301,13 @@ supplantFileFile = (tmp,filename,backupq) -> (
      else "file created for initialization text";
      linkFile(tmp,filename);
      removeFile tmp;
-     stderr << "--" << msg << ": " << filename << endl;
+     printerr(msg, ": ", filename);
      )
 
 supplantStringFile = (text,filename,backupq) -> (
      text = replace("/PREFIX/",prefixDirectory,text);
      if fileExists filename and text === get filename then (
-	  stderr << "--initialization text already in file: " << filename << endl;
+	  printerr("initialization text already in file: ", filename);
 	  return;
 	  );
      tmp := filename | ".Macaulay2.tmp";
@@ -336,7 +336,7 @@ mungeFile = (filename, headerline, trailerline, text) -> (
 		    );
 	       newcontents := replace(regexp, insert, oldcontents);
 	       if oldcontents == newcontents then (
-		    stderr << "--initialization text already in file, no changes needed: " << filename << endl;
+		    printerr("initialization text already in file, no changes needed: ", filename);
 		    return false;
 		    );
 	       )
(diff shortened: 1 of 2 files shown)


mahrud:
What's the advantage here, and are the tests testing anything?

A big disadvantage seems to be that it doesn't work if the user has a variant of emacs installed or if the path to emacs is not known by M2.

d-torrance:
Several advantages:

* The .el files are byte-compiled when the package is installed, which will improve performance.
* It integrates the M2 package with the rest of the emacs package ecosystem.  For example, it will appear when running `package-list-packages`.
* Consistency with the Debian/Ubuntu package, which has been installing the Emacs package like this for years.

Good point regarding M2 not knowing how to find emacs.  Perhaps we could add `setupEmacs(String)` that the user could use if just `emacs` doesn't work?

It would definitely be good to test this with several Emacs variants.

M2/Macaulay2/m2/files.m2
```diff
@@ -447,7 +447,7 @@ runAndEcho = s -> (
     printerr("running: ", cmd);
     run cmd)
 
-makeEmacsPackage = () -> (
+makeEmacsPackage = emacs -> (
     if run "command -v emacs > /dev/null" != 0 then (
```
mahrud: `command -v emacs`

M2/Macaulay2/m2/files.m2
```diff
@@ -508,7 +508,13 @@ prelim := () -> (
      promptUser = true;
      if prefixDirectory === null then error "can't determine Macaulay 2 prefix (prefixDirectory not set)";
      )
-installMethod(setupEmacs, () -> ( prelim(); mungeEmacs(); makeEmacsPackage() ))
+
+setupEmacs String := emacs -> (
+    prelim();
+    mungeEmacs();
+    makeEmacsPackage emacs)
+installMethod(setupEmacs, () -> setupEmacs "emacs")
```
mahrud: By the way, `setupEmacs() := () -> setupEmacs "emacs"` should really be allowed ...

M2/Macaulay2/m2/files.m2
d-torrance: D'oh - thanks!

M2/Macaulay2/packages/Macaulay2Doc/ov_editors_emacs.m2
```diff
@@ -54,6 +53,25 @@ Node
       ; use C-u f11 to temporarily edit the M2 executable
       (global-set-key [ f11 ] 'M2-send-to-program)
     Text
+      @HEADER3 {"Method 3: using Doom Emacs"}@
```
mahrud: Maybe just tell them to use vim instead?! XD

mahrud: I am only half joking: we should not bother specifically including info for people who choose to use an obscure emacs. It complicates the docs for the other, normal 99.99% percent.

mahrud:
I don't know what's the idea with the last few commits but I'm highly opposed. M2 documentation doesn't need lisp syntax highlighting ... you can put extra stuff on the wiki if absolutely necessary, but even then I think it's unnecessary.

d-torrance:
Fair enough -- I guess I did go a bit overboard!  I'll remove them.

d-torrance:
I think this is ready now.  Now, the code in `.emacs-Macaulay2` first checks to see if the `M2` function is defined (which it should be if package installation was successful).  But if not, it falls back to the old behavior of adding the Emacs directory to `load-path` and loading `M2-init.el`.

d-torrance:
Closing in favor of the plan discussed in #4692

PR #4091, opened 2026-01-13
Add syntactic sugar for installing nullary methods

d-torrance:
As proof of concept, use it for all the nullary methods defined in `Core`.

For example:

```m2
i1 : foo = method()

o1 = foo

o1 : MethodFunction

i2 : foo () := () -> print "Hello, world!"

o2 = FunctionClosure[stdio:2:10-2:37]

o2 : FunctionClosure

i3 : foo()
Hello, world!
```
Inspired by https://github.com/Macaulay2/M2/pull/4090#discussion_r2688055461.  I'm not sure why this wasn't implemented before.  It's mentioned as a possibility in the docs, and it's literally just one line of code!

### DIFF
--- a/M2/Macaulay2/m2/debugging.m2
+++ b/M2/Macaulay2/m2/debugging.m2
@@ -166,14 +166,7 @@ localSymbols Pseudocode :=
 localSymbols Symbol :=
 localSymbols Dictionary :=
 localSymbols Function := ls
-
--- make this work eventually:
--- localSymbols() := () -> if current === null then ls() else ls current
--- meanwhile: (see also method123())
--- nullaryMethods # (1 : localSymbols) = () -> if current =!= null then ls current else error "not in debugger (i.e., current not set)"
--- also meanwhile:
-installMethod(localSymbols, () -> if current =!= null then ls current else error "not in debugger (i.e., current not set)")
-
+localSymbols() := () -> if current =!= null then ls current else error "not in debugger (i.e., current not set)"
 localSymbols(Type,Symbol) :=
 localSymbols(Type,Dictionary) :=
 localSymbols(Type,Function) :=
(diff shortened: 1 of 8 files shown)


M2/Macaulay2/m2/debugging.m2
```diff
@@ -166,14 +166,7 @@ localSymbols Pseudocode :=
 localSymbols Symbol :=
 localSymbols Dictionary :=
 localSymbols Function := ls
-
--- make this work eventually:
--- localSymbols() := () -> if current === null then ls() else ls current
```
mahrud, line 171: hmm `ls()` seems like to work fine, no?

M2/Macaulay2/m2/integers.m2
```diff
@@ -44,7 +44,7 @@ ZZ.random = opts -> ZZ -> rawRandomZZ opts.Height
 texMath ZZ := toString
 
 gcd = method(Binary => true)
-installMethod(gcd, () -> 0)
+gcd () := () -> 0
```
mahrud: I think I would prefer no space before `(`:
```suggestion
gcd()      := ZZ => () -> 0
```

M2/Macaulay2/m2/debugging.m2
d-torrance: It's a little goofy.  If we use the commented-out code, then we get:

```m2
i1 : debug Core

i2 : localSymbols()

o2 = {f, ls, show1, processArgs, olderror, callCount, allThingsWithNames,
     justTypes, select2}

o2 : List
```

`f` is the parameter in `ls`, and the others are just local things defined in `debugging.m2`.  They're local to the scope where `localDictionaries` is called, but that's not what we want.

mahrud: Oh I see.

mahrud (approved):

PR #4092, opened 2026-01-14
Update Tableaux Package

j-graf:
This updates the package Tableaux to v0.6. This simplifies the package by merging the classes `SkewTableau` and `YoungTableau` into just `YoungTableau`. This update also brings several features from the package `SpechtModule`, so that most of its features are implemented. 

TODO: Finish implementing features from `SpechtModule`, and optimize code and algorithms more (especially cacheing).

A more complete changelog is below:
- Refactor code
  - Remove subclass `YoungTableau`, and rename `SkewTableau` to `YoungTableau`
  - Rename `truncate` to `trim` (c.f. package `Permutations`)
- Feature match `SpechtModule`
  - Subclass `MutableYoungTableau`
  - Algorithms
    - `allStandardTableaux`
    - `numStandardTableaux`/`hookLength`
    - `allTabloids`
    - `numTabloids`
  - Statistics
    - `rowStabilizer`
    - `columnStabilizer`
    - `readingWord`
  - Note: will not impliment (at lease for now)
    - `TableauList`
      - `addTableau`
      - `toListOfTableaux`
    - `firstRowDescent`
    - `indexTableau`
    - `listToTableau`
    - `rowPermutationTableaux`
    - `tableauToList`
- More features
  - Updated `net`, so that any filling can be drawn properly (e.g., if the entries are tableaux themselves)
  - Subclass `Tabloid`
  - Export method `isSkew`
  - Export methods `outerShape` and `innerShape`
  - Export method `shape` as alias for `outerShape`, raise error if `isSkew`
  - More Booleans
    - `isSemistandard`
    - `isStandard`
    - `isRowStrict` and `isColumnStrict`
  - Random tableaux
    - `randomSemistandardTableau`
    - `randomStandardTableau`
    - `randomTabloid`
  - Added `toPartitionChain`
  - Useful partition methods
    - `allSubPartitions`
    - `isNonnegative`
    - `isWeaklyDecreasing`
  - Concatenating nets: `horizonalNet` and `verticalNet`
- Bug fixes/changes


d-torrance (changes requested):
Thanks for the contribution!

There are some typos:

```
M2/Macaulay2/packages/Tableaux/documentation.m2:538: concatentates ==> concatenates
M2/Macaulay2/packages/Tableaux/documentation.m2:1488: concatentates ==> concatenates
M2/Macaulay2/packages/Tableaux/documentation.m2:1801: equivelence ==> equivalence
```

M2/Macaulay2/packages/Tableaux/algorithms.m2
```diff
@@ -142,4 +153,181 @@ numSemistandardTableaux (Partition,ZZ) := (lam,n) -> (
...
+        cornerList := positionList(youngTableau mu,isCorner);
+        randomIndex := random(#cornerList);
+        thePosition := cornerList#randomIndex;
+
+        (rowIndex,colIndex) := thePosition;
+        mu = trim new Partition from for i from 0 to #mu-1 list (
+            if i == rowIndex then mu#i-1 else mu#i
+            );
+        
+        indexList = indexList | {toIndex(thePosition,T)};
```
d-torrance: Appending to a list inside a `for` loop like this isn't efficient since you build a brand new list at each step.  I'd refactor this to use `apply` or `for ... list`.

d-torrance: The same thing is happening in `allSubPartitions` and `toPartitionChain`

d-torrance (changes requested):

M2/Macaulay2/packages/Tableaux/algorithms.m2
j-graf: In `randomStandardTableau` , `allSubPartitions`, and `toPartitionChain`, each iteration in these loops depends on the output of the previous steps, so they are actually recursive algorithms. Is there a way to use `apply` or `for ... list` in this situation?

d-torrance: Here, we should be able to do:

```m2
indexList = apply(toList(0..n), i -> (
    ...
    toIndex(thePosition,T))
```
or

```m2
indexList = for i to n list (
    ...
    toIndex(thePosition,T))
```

j-graf: Right, but my issue here is that `thePosition` in iteration `i` depends on `indexList#{0..(i-1)}`, but I can't access any `indexList#k` until the `apply` or the `for` loop has finished.

I have wondered about this before, so in general, let's replace the `...` with `someCalculation` which takes `i` inputs (or a list of length `i` as an input). Then it seems like your suggestion is
```
outputList = apply(toList(0..n), i -> (
     ithOutput := someCalculation(outputList#{0..(i-1)});
     ithOutput
     ))
```
Clearly this wouldn't work since `outputList#{0..(i-1)}` isn't initialized yet, but is there some way to rewrite it? Or maybe the inner loop actually could access `outputList#{0..(i-1)}` in _Macaulay2_?

A special case of this that would work is if `someCalculation` only required one input. In this case, one could use a temporary variable, such as
```
tempVar = initialValue
outputList = apply(toList(0..n), i -> (
     tempVar = someCalculation(tempVar);
     tempVar
     ))
```
This should work for one of my algorithms. But if I need _all_ previous results, then I need to make `tempVar` into a list and keep appending to it anyways, making something like `apply` redundant, as far as I can tell. Would making `tempVar` into a `MutableHashTable` work efficiently here?

d-torrance: Yeah, mutable objects would work.  Probably a mutable list, I think, since you can control how big it when you construct it and don't need to worry about non-integer keys.

So maybe something like:

```m2
indexList = new MutableList from apply(n, i -> null)
for i to n do (
   ...
   indexList#i = ...
)
```

j-graf: Thank you for the help, I've made the changes. I will just add that for `allSubPartitions`, there is an unknown number of outputs for each iteration of the loop, so I can't preallocate the size of the MutableList.

d-torrance (approved):
Looks good -- thanks!

PR #4097, opened 2026-01-17
Add "new MutableList from ZZ"

d-torrance:
We add a new constructor for `MutableList` suggested in an old `bugs/dan` file:

```m2
i1 : new MutableList from 6

o1 = MutableList{...6...}

o1 : MutableList

i2 : peek oo

o2 = MutableList{null, null, null, null, null, null}
```

We also improve the `MutableList` docs a bit, mentioning a couple things suggested in another old `bugs/dan` file.


mahrud:
I would prefer a different notation here. Maybe something involving `allocate` or `size` rather than squeezing into NewFromMethod, which is meant to be for subclasses only.

d-torrance:
Discussed with Mike

Issue #659, opened 2017-07-27
growth of mutable lists


DanGrayson:
It looks like mutable lists grow their length by only one at a time.  Perhaps
the length should double when it needs to increase.  Or just warn users to
amortize the cost themselves.
```
    i36 : x = new MutableList 

    o36 = MutableList{}

    o36 : MutableList

    i37 : x#4=1

    o37 = 1

    i38 : peek x

    o38 = MutableList{null, null, null, null, 1}

    i39 : x#5=1

    o39 = 1

    i40 : peek x

    o40 = MutableList{null, null, null, null, 1, 1}
```

from evaluate.d:
```
assignvector(m:List,i:Code,rhs:Code):Expr := (
     x := m.v;
     ival := eval(i);
     when ival
     is j:ZZcell do (
	  if isInt(j)
	  then (
	       k := toInt(j);
	       if k < 0 then k = k + length(x);
	       if k < 0 then return printErrorMessageE(i,"negative subscript out of bounds 0 .. "+tostring(length(x)-1));
	       val := eval(rhs);
	       when val is Error do return val else (
		    if k >= length(x) then (
			 x = new Sequence len k+1 do (
			      foreach t in x do provide t;
			      while true do provide nullE;
			      );
			 m.v = x; );
	       	    x.k = val;
	       	    val))
	  else printErrorMessageE(i,"expected small integer"))
     is Error do ival
     else printErrorMessageE(i,"index not an integer")
     );
```

d-torrance:
> **Written by Claude** (Claude Opus 5, via Claude Code), not by @d-torrance, whose account posted it -- please weigh it accordingly.

Numbers for the growth policy described here, from the pre-GitHub `bugs/` tree.
Dan Grayson filed them as `bugs/dan/0-mutable-lists`, under "This should not
slow down by a factor of 100!":

```
i39 : time (x=new MutableList ; for i to 1000 do x#i = i ;  )
     -- used 0.002677 seconds

i40 : time (x=new MutableList ; for i to 10000 do x#i = i ;  )
     -- used 2.56211 seconds
```

Still current on 1.26.06-8-g34d5846039, and the shape is unchanged -- growing one
element at a time makes filling a list of length n cost O(n²):

```
  n=1000    .0013 s
  n=5000    .0568 s
  n=10000   .3664 s
  n=20000  1.3507 s
```

Ten times the elements costs about 280 times the work, and each doubling of n
[...]

PR #4103, opened 2026-01-23
Update mutable list reallocation

d-torrance:
Currently, when we add an element to a mutable list, we reallocate exactly enough memory to fit the new element.  So doing something like `for i to n - 1 do x##x = i` constructs n brand new mutable lists.

Instead, we double the length of the underlying sequence if needed.  (And also halve it if it gets small from repeated calls to `remove`.)

So each list `x` now has two lengths:

* `length(x.v)` -- the "visible" length at top level
* `x.capacity` -- the actual length allocated for the underlying sequence

*Edit:  Originally, I had these swapped, but that broke pretty much everything, since so much code depends on `x.v` having exactly the same elements as `x`.*

For immutable lists, these two numbers should always be the same.

I changed some pretty low-level stuff, so chances are there are bugs... Draft for now

### Before

```m2
i1 : x = new MutableList

o1 = MutableList{}

o1 : MutableList

i2 : elapsedTime for i to 99999 do x##x = i
 -- 10.8305s elapsed
```

### After

```m2
i2 : elapsedTime for i to 99999 do x##x = i
 -- .0493995s elapsed
```

Closes: #659

### DIFF
--- a/M2/Macaulay2/d/parse.d
+++ b/M2/Macaulay2/d/parse.d
@@ -321,7 +321,8 @@ export List := {+
      Class:HashTable,
      v:Sequence,
      hash:hash_t,
-     Mutable:bool
+     Mutable:bool,
+     capacity:int -- amount of allocated memory (for mutable lists)
      };
 
 export Error := {+
--- a/M2/Macaulay2/d/actors2.dd
+++ b/M2/Macaulay2/d/actors2.dd
@@ -58,11 +58,7 @@ toList(e:Expr):Expr := (
      is a:Sequence do list(a)
      is b:List do (
 	  if b.Class == listClass then e
-	  else Expr(
-	       sethash(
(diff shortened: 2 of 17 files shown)


M2/Macaulay2/d/parse.d
```diff
@@ -312,7 +312,8 @@ export List := {+
      Class:HashTable,
      v:Sequence,
      hash:hash_t,
-     Mutable:bool
+     Mutable:bool,
+     capacity:int -- amount of allocated memory (for mutable lists)
```
antonleykin, line 325: Does it add one word to every list (mutable or not)?

d-torrance: Yes, it does.  It's redundant for immutable lists since this will always agree with the length of the underlying sequence.

Issue #4104, opened 2026-01-26
Errors during SDP optimization with "M2" solver


LegionMammal978:
Calling `optimize()` with an SDP problem from the SemidefiniteProgramming package can throw errors even when the instance is well-formed and feasible. For instance, the script:
```
loadPackage "SemidefiniteProgramming";
optimize(sdp(matrix{{0, 0}, {0, 1}}, 1:matrix{{1, 0}, {0, 0}}, matrix{{1}}), Solver=>"M2");
```
throws the error:
```
/usr/share/Macaulay2/Core/robust.m2:101:30:(1):[22]: error: no method for binary operator * applied to objects:
            null (of class Nothing)
      *     | 1 0 | (of class Matrix)
            | 0 0 |
/usr/share/Macaulay2/SemidefiniteProgramming.m2:531:48:(2):[21]: --back trace--
/usr/share/Macaulay2/Core/lists.m2:326:62:(1):[20]: --back trace--
/usr/share/Macaulay2/Core/lists.m2:326:49:(1):[19]: --back trace--
/usr/share/Macaulay2/Core/lists.m2:326:32:(1):[18]: --back trace--
/usr/share/Macaulay2/Core/matrix1.m2:30:18:(1):[17]: --back trace--
/usr/share/Macaulay2/Core/methods.m2:154:98:(1):[15]: --back trace--
/usr/share/Macaulay2/Core/option.m2:17:14:(1):[14]: --back trace--
/usr/share/Macaulay2/SemidefiniteProgramming.m2:531:20:(2):[13]: --back trace--
/usr/share/Macaulay2/Core/option.m2:17:14:(1):[12]: --back trace--
/usr/share/Macaulay2/SemidefiniteProgramming.m2:494:42:(2):[11]: --back trace--
/usr/share/Macaulay2/Core/option.m2:17:14:(1):[10]: --back trace--
/usr/share/Macaulay2/SemidefiniteProgramming.m2:392:37:(2):[9]: --back trace--
/usr/share/Macaulay2/Core/methods.m2:154:98:(1):[8]: --back trace--
example.m2:2:8:(3):[6]: --back trace--
/usr/share/Macaulay2/Core/startup.m2:65:23:(0):[6]: --back trace--
/usr/share/Macaulay2/Core/Core.m2:131:28:(1):[5]: --back trace--
/usr/share/Macaulay2/Core/Core.m2:127:13:(1):[3]: --back trace--
[...]

PR #4105, opened 2026-01-27
Fix check for result of "solve" in SemidefiniteProgramming

d-torrance:
When `solve` fails, it returns null rather than raising an error.  So `try` wasn't helping us at all.  Furthermore, if `solve` failed, then `Sinv` was assigned to null, giving us a "no method found" error a few lines later when we try to multiply it by a matrix.

See #4104

Cc: @LegionMammal978, @tom111


tom111 (approved):
Looks good to me. Thanks for fixing this.

Issue #3553, opened 2024-10-29
GraphicalModels: multiple uses of `gaussianRing`


taboege:
Multiple uses of `gaussianRing Graph` from the `GraphicalModels` package do not work as expected. The code below constructs rings for two different DAGs, but they turn out to be the same (they should already differ in their generators).  It looks like the `.graph` field in both rings is equal to the first DAG.

This impairs all downstream functionality that requires a `gaussianRing` constructed from a specific graph, like `gaussianVanishingIdeal`. Multiple people I talked to, including @bkholler, have noticed this and don't know any workaround.

``` M2
needsPackage "GraphicalModels";
DAGs = {
  digraph({1,2,3},{{1,2}}),
  digraph({1,2,3},{{1,2},{2,3}})
};
L = DAGs / gaussianRing;
print (L#0).graph;
print (L#1).graph;
```

It seems that `gaussianRing` maintains a cache of rings it has constructed before. Perhaps the indexing of that cache does not work as intended?

d-torrance:
It looks like the rings are cached in a hash table.  There's a comment in the code:
https://github.com/Macaulay2/M2/blob/1fd238210e9a37ddaec3b700ccd22f06ff5f28f5/M2/Macaulay2/packages/GraphicalModels.m2#L689

Is this correct? `kk` is the coefficient ring, `s`, `l`, `p`, and `k` are variable names, and `vv` is a sorted list of vertices.  But shouldn't we also keep track of the directed edges to determine the subscripts of the variables in the ring?

Cc: @lukeamendola, @luisgarciapuente, @roserhp, @olgakuznetsova, @harshitmotwani2015, @sonjapetrovic, @mikestillman

bkholler:
@lukeamendola and I will be leading a group on graphical models at the upcoming M2 conference in Leipzig. We can make the change Luis suggests or some other fix then.

lukeamendola:
Good catch! Yes, we will fix it in two weeks during the M2 Workshop in Leipzig!

taboege:
I ran into this bug for the (n+1)st time yesterday and decided to fix it (see #4106). The documentation of `markovRing` at least mentions that the resulting ring will be cached, so for `gaussianRing` I went with the fix that records a sufficient amount of data in the cache key to make it work instead of removing the caching altogether.

d-torrance:
Fixed in #4106

PR #4106, opened 2026-01-27
GraphicalModels: Fix caching of `gaussianRing MixedGraph`

taboege:
This fixes GitHub issue #3553. The gaussianRing function caches its return value. However, for a MixedGraph input, the key only records the vertex set and none of the graph structure which leads to wrong rings being returned.


d-torrance:
Discussed with Mike

PR #4110, opened 2026-01-30
Fix a bunch of compiler warnings

d-torrance:
We fix a bunch of compiler warnings -- mostly "unused parameter".


d-torrance:
Discussed with Mike

PR #4108, opened 2026-01-27
Add "parse" function for getting concrete syntax tree of M2 code

d-torrance:
This adds a parsing-level analog of `pseudocode` & `disassemble`.  It gives us a breakdown of the d-level `ParseTree` object corresponding to a string containing M2 code:

```m2
i1 : parse ///R = QQ[x,y,z,w]
     I = monomialCurveIdeal(R, {1,2,3})///

o1 = {Binary, {Binary, {Token, R}, {Token, =}, {Adjacent, {Token, QQ}, {Parentheses,
     ---------------------------------------------------------------------------------
     {Token, [}, {Binary, {Binary, {Binary, {Token, x}, {Token, ,}, {Token, y}},
     ---------------------------------------------------------------------------------
     {Token, ,}, {Token, z}}, {Token, ,}, {Token, w}}, {Token, ]}}}}, {Token,
     ---------------------------------------------------------------------------------
     -*newline*-}, {Binary, {Token, I}, {Token, =}, {Adjacent, {Token,
     ---------------------------------------------------------------------------------
     monomialCurveIdeal}, {Parentheses, {Token, (}, {Binary, {Token, R}, {Token, ,},
     ---------------------------------------------------------------------------------
     {Parentheses, {Token, {}, {Binary, {Binary, {Token, 1}, {Token, ,}, {Token, 2}},
     ---------------------------------------------------------------------------------
     {Token, ,}, {Token, 3}}, {Token, }}}}, {Token, )}}}}}
```

Draft for now, as there are a couple weird things I'd like to figure out:

* Syntax errors print, but we still get a return value:
   ```m2
   i1 : parse ";"
   currentString:1:0:(3): error: syntax error at ';'

   o1 = {dummy}

   o1 : List
   ```
* Why is this a syntax error?
   ```m2
   i1 : parse ""
   currentString:1:0:(3): error: syntax error at '-*end of file*-'

   o1 = {dummy}

   o1 : List
```

PR #4111, opened 2026-01-31
Remove unnecessary tokens from some ParseTree members

d-torrance:
I noticed this while working on #4108...

Many of the `ParseTree` members lead with a token (`if`, `while`, etc.), and we store them in the corresponding object so we determine the position of the code for `locate`.

However, we don't need the internal tokens (`then`, `do`, etc.) for anything, so there's really no reason to carry them around after parsing.


d-torrance:
Discussed with Mike

PR #4115, opened 2026-02-01
fix #4114

### LINKED ISSUE #4114: (matrix{{x}})^2 != matrix{{x^2}}
As discussed on zulip, right now we have
```
i1 : R=QQ[x]

o1 = R

o1 : PolynomialRing

i2 : (matrix{{x}})^2 == matrix{{x^2}}

o2 = false
```
which is borderline absurd.

I've always found this way of dealing with degrees incorrect. If we decide that `m=matrix{{x}}` is a map (of degree 0) from `R^{-1}` to `R^{0}`, which is the current convention, then we should be consistent and also apply it to products and powers, i.e., `m^2` should be a map (also of degree 0) from `R^{-2}` to `R^{0}`.

I believe the culprit is this section of `Matrix * Matrix` in `matrix.m2`:
```
	  dif := degrees P - degrees Q;
	  deg := (
	       if #dif === 0
	       then degree m + degree n
	       else if same dif
	       then degree m + degree n + dif#0
 	       else toList (degreeLength ring m:0)
	       );
```
the degree of `m*n` should *always* be `degree m+degree n` (in particular, preserving the degree 0 case, i.e., preserving homogeneity of maps). So instead of the code above, one should degree-shift the source and target of `n`  to align the target of `n` and the source of `m`.

Incidentally this would also resolve another annoying issue:
```
i3 : (matrix{{x}})^0
stdio:3:13:(3): error: expected source and target to agree
```

pzinn:
this is a draft for now, curious to see if this change has unexpected consequences.

### DIFF
--- a/M2/Macaulay2/m2/matrix.m2
+++ b/M2/Macaulay2/m2/matrix.m2
@@ -190,13 +190,8 @@ Matrix * Matrix := Matrix => (m,n) -> (
 	  if not isFreeModule P or not isFreeModule Q or rank P =!= rank Q
 	  then error "maps not composable";
 	  dif := degrees P - degrees Q;
-	  deg := (
-	       if #dif === 0
-	       then degree m + degree n
-	       else if same dif
-	       then degree m + degree n + dif#0
- 	       else toList (degreeLength ring m:0)
-	       );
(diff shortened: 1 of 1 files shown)


pzinn:
wait, that doesn't work. rethinking.

PR #4116, opened 2026-02-01
fix #4114

### LINKED ISSUE #4114: (matrix{{x}})^2 != matrix{{x^2}}

pzinn:
This PR fixes #4114 in a clean, simple, manner.

### DIFF
--- a/M2/Macaulay2/m2/matrix.m2
+++ b/M2/Macaulay2/m2/matrix.m2
@@ -179,8 +179,8 @@ Matrix * Matrix := Matrix => (m,n) -> (
      	  R := ring m;
 	  S := ring target n;
 	  if R =!= S then ( -- use toSameRing?
-	       try m = promote(m,S) else
-	       try n = promote(n,R) else
+	       try (m = promote(m,S); R=S) else
+	       try (n = promote(n,R); S=R) else
 	       error "maps over incompatible rings";
 	       );
 	  M = target m;
@@ -190,13 +190,8 @@ Matrix * Matrix := Matrix => (m,n) -> (
 	  if not isFreeModule P or not isFreeModule Q or rank P =!= rank Q
 	  then error "maps not composable";
 	  dif := degrees P - degrees Q;
-	  deg := (
-	       if #dif === 0
-	       then degree m + degree n
-	       else if same dif
-	       then degree m + degree n + dif#0
- 	       else toList (degreeLength ring m:0)
-	       );
+	  deg := degree m + degree n;
+	  if #dif > 0 then if same dif then N = N ** S^{-dif#0} else deg = toList (degreeLength S:0);
 	  f := m.RawMatrix * n.RawMatrix;
 	  f = rawMatrixRemake2(rawTarget f, rawSource f, deg, f, 0);
 	  f = reduce(M,f);
@@ -205,8 +200,11 @@ Matrix * Matrix := Matrix => (m,n) -> (
 	  else map(M,N,f)))
 
 Matrix#1 = f -> (
-    if source f =!= target f then error "expected source and target to agree"
-    else id_(target f))
+    M := target f;
+    N := source f;
+    if (M =!= N) and (not isFreeModule M or not isFreeModule N or rank M =!= rank N)
+    then error "expected source and target to agree"
+    else id_M)
 Matrix ^ ZZ := Matrix => BinaryPowerMethod
 
 transpose Matrix := Matrix => (cacheValue symbol transpose) (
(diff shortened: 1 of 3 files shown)


M2/Macaulay2/m2/modules.m2
```diff
@@ -281,7 +281,7 @@ Ring ^ ZZ   := Module => (R, n) -> (
 Ring ^ List := Module => (R, degs) -> (
     if not R.?RawRing then error "non-engine free modules with degrees not implemented yet";
     -- check the args
-    degs = - splice degs;
+    degs = - splice\splice degs;
```
mahrud, line 284: `deepSplice`?

pzinn: I don't think so. `deepSplice` undoes nested sequences, here we have a list of lists. the issue is that most of the time, we want to splice the innermost level (as I discovered while writing this PR), and this is not what the original line did.

M2/Macaulay2/m2/matrix.m2
```diff
@@ -205,8 +200,11 @@ Matrix * Matrix := Matrix => (m,n) -> (
 	  else map(M,N,f)))
 
 Matrix#1 = f -> (
-    if source f =!= target f then error "expected source and target to agree"
-    else id_(target f))
+    M := target f;
+    N := source f;
+    if (M =!= N) and (not isFreeModule M or not isFreeModule N or rank M =!= rank N)
```
mahrud, line 205: So you want for instance `(random(R^{1}, R^{0}))^0` to give the identity on target, but not `(random(coker vars R, R^{0}))^0`?

pzinn: you mean you'd like to go even further and allow non free modules? I was trying to introduce the minimal changes that would fix the most blatant problems. This PR already introduces some backwards incompatibility. (of course in an ideal world people wouldn't try to multiply or take power of matrices when source and target don't agree...)

mahrud: I'd rather not change it at all, but trying to see what the problem is from your perspective.

pzinn: I think I made my point pretty clear in #4114 
the current system to deal with multiplying maps whose inner degrees disagree is illogical and inconsistent. why should
- `(matrix {{x}})^2 != matrix {{x^2}`
- `(matrix{{x}})^0` raise an error and
- `(matrix{{x}}^n` be of degree n-1 when n>0

The last point has been annoying me for the past 20 years.
the good news, which only occurred to me recently, is that there is a clean, simple, fix.
This is what this PR is dong.

mahrud: Idk if I agree that it's illogical; there are two choices and ^ and ^** each take one approach to provide users with choice. Anyway, I responded to your examples directly in the issue so it doesn't get lost.

pzinn: `^` and `^**` do different things, so no, that doesn't work.

pzinn:
This is a very mild breaking change -- it only affects situations where incorrect degrees are fed to `Matrix * Matrix`, so shouldn't (and doesn't) affect any calculations in the Core or packages. And if the user was OK before with the nonsensical answer that the product gave when there is a degrees mismatch, then surely they will be OK with the more reasonable answer it now produces.

pzinn:
In view of the discussion in #4162, I should also point out a great improvement provided by this simple PR: the product of two degree zero maps is now degree zero... so "naive" (degree 0) homogeneity is now preserved.

mikestillman:
Discussed this with Doug.

PR #4119, opened 2026-02-02
Stop trying to install python & ninja in GitHub macOS builds

d-torrance:
They're already installed and we just get warnings.

For example, from https://github.com/Macaulay2/M2/actions/runs/21576025985:

image


d-torrance:
Discussed with Mike

PR #4123, opened 2026-02-04
Bump gdbm to 1.26

d-torrance:
Autotools only, as cmake doesn't have the option of building it.


d-torrance:
I also added the `--without-readline` option.  Since we're just using it as a library and not the `gdbmtool` REPL, there's no reason to  use readline.

d-torrance:
The builds all passed, but I'm going to wait for #4124

d-torrance:
Actually, I'll just close this and bump gdbm in #4124

PR #4124, opened 2026-02-04
Check sha256sum of downloaded tarballs (autotools)

d-torrance:
This is a standard check for security reasons to make sure we're installing what we expect to install.  We already do this in the cmake build, and it's also done for common package managers that download tarballs during build (e.g., homebrew and spack).

So for example:

```
4ti2-1.6.14.tar.gz  100%[===================>]   5.62M  --.-KB/s    in 0.09s   

2026-02-05 16:50:51 (65.5 MB/s) - ‘4ti2-1.6.14.tar.gz’ saved [5889819/5889819]

 ✓ SHA256 ok (1bc340173f93ca4abd30ea962118bd5057fdedf7e79c71d2a0c4cc9569f8b0b1)
```

But if we have the wrong checksum:

```
4ti2-1.6.14.tar.gz  100%[===================>]   5.62M  --.-KB/s    in 0.08s   

2026-02-05 16:51:53 (73.9 MB/s) - ‘4ti2-1.6.14.tar.gz’ saved [5889819/5889819]

error: *** SHA256 hash of 4ti2-1.6.14.tar.gz does not match the expected value
  expected: 1bc340173f93ca4abd30ea962118bd5057fdedf7e79c71d2a0c4cc9569f8b0b
    actual: 1bc340173f93ca4abd30ea962118bd5057fdedf7e79c71d2a0c4cc9569f8b0b1
```

Draft for now to test the builds.  (There's a chance some of the builds need tweaks -- it's probably been a while since we've tested them!)

*Edit:*  Various tweaks:

* Bump gdbm (formerly in #4123), ntl, mpfr, and readline to latest versions.
* Add a patch from FreeBSD to fix the gfan build in macOS
* Use the `SHARED` variable to determine whether to build mpfr and fplll as static or shared libraries.
* Build readline w/ curses support (otherwise we break `awk` in Ubuntu)
* Remove cddplus (it's unmaintained, doesn't build, and we don't use it anyway)
* Drop `-ansi` from the csdp build flags to fix a compile error (which we're already doing in our homebrew tap)
* Update a *ForeignFunctions* test so it doesn't require mpfi (this has been a long-standing issue)
* Update a probabilistic NAG example to not use `assert`
* Enable TLS support when building mpfr (avoids some segfaults when testing threads)


d-torrance:
Apparently building readline from source on Ubuntu breaks awk:

```
awk: symbol lookup error: /home/dtorrance/src/macaulay2/M2/M2/BUILD/build/usr-host/lib/libreadline.so.8: undefined symbol: UP
```

I'm not sure if this is worth opening an issue -- if we're in Ubuntu, then we'll almost certainly just install `libreadline-dev` and not bother building it.  So I just removed readline from the list of packages to build in the test commit for this PR.  The sha256sum check worked, though!

*Edit:*  Fixed it!

d-torrance:
Builds passed!

https://github.com/Macaulay2/M2/actions/runs/21737229800

d-torrance:
Discussed with Mike

PR #4130, opened 2026-02-15
Update algorithm for "hilbertFunction Module"

btotaro:
We compute hilbertFunction(d, M) for a subquotient module M as the difference of two vector-space dimensions. Each vector-space dimension is computed as "rank source basis(d, module)", rather than using hilbertSeries. There is also an option to use the earlier algorithm.


d-torrance:
A couple of the *NoetherianOperators* examples are failing:

```m2
-- -*- M2-comint -*- hash: 10390241182418781559

i1 : R = CC[x,y];

i2 : I = ideal{x^2,x*y}

             2
o2 = ideal (x , x*y)

o2 : Ideal of R

i3 : d = localHilbertRegularity(origin R, I)
-- warning: experimental computation over inexact field begun

o3 = 2

i4 : D = truncatedDual(origin R, I, 3)

o4 = | 1 y x .5y2 .166667y3 |

o4 : DualSpace

i5 : L = hilbertFunction({0,1,2,3}, D)
stdio:5:19:(3): error: expected 2 arguments but got 1
-- -*- M2-comint -*- hash: 9482620399911031897

i1 : R = CC[x,y];

i2 : I = (ideal {x,y})^4;

o2 : Ideal of R

i3 : D = zeroDimensionalDual(origin R, I)

o3 = | 1 y x .5y2 xy .5x2 .166667y3 .5xy2 .5x2y .166667x3 |

o3 : DualSpace

i4 : hilbertFunction D
stdio:4:15:(3): error: key not found in hash table:
        Space (of class Symbol)
```

The `hilbertFunction` methods from this package probably just need `opts ->` added.

btotaro:
Got it. Thanks for your help.
________________________________

M2/Macaulay2/m2/hilbert.m2
```diff
@@ -373,31 +373,45 @@ hilbertSeries ProjectiveHilbertPolynomial := opts -> P -> (
 -- hilbertFunction
 -----------------------------------------------------------------------------
 
-hilbertFunction = method()
+protect symbol UseHilbertSeries
```
mahrud: The symbol `UseHilbertSeries` needs to be exported for this strategy to be usable.

mahrud (commented):
I have a couple of concerns here:
1. A main benefit of the current algorithm is that Hilbert series are progressively cached, so computing Hilbert functions get easier and easier, but the proposed algorithm doesn't cache any information.
2. More importantly, both `minimalPresentation` and `basis` are huge time and space bottlenecks as degree gets high. Here's a simple example that essentially takes the whole memory and never finishes using the proposed algorithm:
```m2
elapsedTime hilbertFunction(5, QQ[x_0..x_100])
```

Are there examples that are faster using the proposed method?

btotaro:
Thanks for looking at this. To clarify one thing: you're right that "minimalPresentation" can be slow, but the algorithm I'm proposing only uses a minimal presentation if one has already been computed.

There are definitely some cases where the method I'm proposing for hilbertFunction is faster. To take a somewhat random example:
R0=QQ; R1=R0[x0,x1,x2,x3,x4,x5]; f1=x0^8+x1^8+x2^8+x3^8+x4^8; R2=R1/(ideal f1);
M2=module ideal(random(2,R2),random(2,R2),random(3,R2));
hilbertFunction(5,M2);
On my PC, this took 145s CPU time with the current algorithm, and 11s CPU time with the one I'm proposing.

On the other hand, you're right that there are cases where the current algorithm for hilbertFunction is faster than the new one. And I agree that caching works better for the current algorithm. Since it is not clear when to prefer the new algorithm, I think I will change my pull request to keep the current algorithm as the default, and add the new one as an option.

One reason I want to add the new algorithm at least as an option is that it is definitely faster than "rank source basis(d, M)" when M is a subquotient module. (It expresses that dimension as a difference of two dimensions which are easier to compute.) And several Macaulay2 functions use "rank source basis(d,M)". (Indeed, the latter method is praised on the documentation page for "hilbertFunction".) It should be an improvement to replace that by the new option in essentially all cases.

mahrud:
It seems like `basis` actually computes the Hilbert series along the way. Compare the timing of computing the Hilbert function, then basis:
```m2
i1 : R0=QQ; R1=R0[x0,x1,x2,x3,x4,x5]; f1=x0^8+x1^8+x2^8+x3^8+x4^8; R2=R1/(ideal f1);

i5 : M2=module ideal(random(2,R2),random(2,R2),random(3,R2));

i6 : elapsedTime hilbertFunction(5, M2);
 -- 60.5051s elapsed

i7 : elapsedTime rank source basis(5, M2);
 -- .020636s elapsed
```
with computing basis, then the Hilbert function:
```m2
i1 : R0=QQ; R1=R0[x0,x1,x2,x3,x4,x5]; f1=x0^8+x1^8+x2^8+x3^8+x4^8; R2=R1/(ideal f1);

i5 : M2=module ideal(random(2,R2),random(2,R2),random(3,R2));

i6 : elapsedTime rank source basis(5, M2);
 -- 58.2869s elapsed

i7 : elapsedTime hilbertFunction(5, M2);
 -- .00184891s elapsed
```
Try computing `rank source basis` on your system before computing Hilbert function/series, is it still faster?

btotaro:
That's interesting. If you look at the function "basis" in basis.m2, its main step is to find a presentation of the module (and then a Grobner basis of that). In other words, if the module was given as a subquotient, "basis" converts it into a quotient module. That is often most of the work needed to compute the Hilbert series, as your examples show. But that's not an objection to the new option I proposed. For a subquotient module M, the point of the new option is _not_ to directly compute basis(d,M); rather, it computes that dimension as the difference of the dimensions of two quotient modules. In the example you mention, running new examples of the randomly generated module M2 each time, I find that hilbertFunction(5,M2) takes 96s of CPU time, rank source basis(5,M2) takes 92s, and the new algorithm takes 4s. (This is related to a similar improvement of "poincare Module" that you approved recently. It may be that the new option for hilbertFunction will have about the same speed as the old one after the new "poincare Module" is installed, since that is used by hilbertSeries and hence by hilbertFunction.)

mahrud:
Oh sorry I wasn't raising an objection. I was mainly trying to figure out if there's a central place to fix this. In particular, if I'm not mistaken, we could get the same improvements if we instead change the `hilbertSeries` function to use the same trick for subquotient modules, without dealing with the memory bottleneck in the basis computation.

For instance:
```m2
i4 : M2 = module ideal(random(2, R2), random(2, R2), random(3, R2));

i5 : M3 = cokernel(gens M2 | relations M2);

i6 : A = elapsedTime hilbertSeries super M2;
 -- .00093556s elapsed

i7 : B = elapsedTime hilbertSeries M3;
 -- 2.58696s elapsed

i8 : elapsedTime hilbertFunction(5, super M2) - hilbertFunction(5, M3)
 -- .00155601s elapsed

o8 = 125
```
I'll make a pull request with this method shortly.

mahrud:
Wait, sorry, I'm confused. The Hilbert series function uses `poincare`, so by your previous PR, already benefits from this trick for subquotient modules. If I run your example using only your changes from the previous PR, I already get a fast answer:
```m2
i1 : R0=QQ; R1=R0[x0,x1,x2,x3,x4,x5]; f1=x0^8+x1^8+x2^8+x3^8+x4^8; R2=R1/(ideal f1);

i5 : M2=module ideal(random(2,R2),random(2,R2),random(3,R2));

i6 : elapsedTime hilbertFunction(5,M2);
 -- 3.2945s elapsed
```
So it would be redundant to do this from scratch in `hilbertFunction`, no?

btotaro:
Sure, it's not absolutely necessary to include this new option. But why not include it? At least it seems about the same speed as the default method (using the update to poincare Module, as you say). Using "basis(d,M)" for a quotient module M does not do exactly the same thing as calling "hilbertSeries", although they're related; so the new option might be faster in some cases. And maybe someone will improve "basis(d, M)" (used in this option) at some point, for a quotient module M: it's clearly not necessary to compute a Grobner basis for M in all degrees (as "basis" now does), only in degrees up to d. So to make room for that kind of future improvement, I'd like to include this option. Thanks for looking at this.

mahrud:
Oh I'm not opposed to adding it as an opt-in strategy, especially in anticipation of better basis algorithms as you said. I'm just curious about the boundary of when different algorithms outperform, e.g. in this case when is using basis faster than a Hilbert series computation with the same trick.

@mikestillman does computing a Grobner basis (even up to degree d) internally compute the poincare polynomial first? I recall this is the case at least for free resolutions, but I'm not sure about gb.

M2/Macaulay2/m2/hilbert.m2
```diff
@@ -373,31 +373,48 @@ hilbertSeries ProjectiveHilbertPolynomial := opts -> P -> (
...
-hilbertFunction(ZZ, Module) := (d, M) -> hilbertFunction({d}, M)
+hilbertFunction(ZZ, Module) := opts -> (d, M) -> hilbertFunction({d}, M, opts)
 
-hilbertFunction(List, Ring)   := (L, R) -> hilbertFunction(L, module R)
-hilbertFunction(List, Ideal)  :=
-hilbertFunction(List, Module) := (L, M) -> (
-    -- computes the Hilbert series to a sufficiently high order and
-    -- returns the desired coefficient, thus it is cached by hilbertSeries
+hilbertFunction(List, Ring)   := opts -> (L, R) -> hilbertFunction(L, module R, opts)
+hilbertFunction(List, Ideal)  := opts -> (L, I) -> hilbertFunction(L, cokernel generators I, opts)
```
mahrud:
```suggestion
hilbertFunction(List, Ideal)  := opts -> (L, I) -> hilbertFunction(L, comodule I, opts)
```

M2/Macaulay2/m2/hilbert.m2
```diff
@@ -373,31 +373,48 @@ hilbertSeries ProjectiveHilbertPolynomial := opts -> P -> (
...
+-- Also, we do this by finding a basis for F/N2 and F/N1 in degree d, rather than computing the whole Hilbert series.
+-- This may or may not be faster than the default strategy, but it should be at least as fast as "rank source basis(d, M)"
+-- in essentially all cases, and faster than that when M was defined as a subquotient module.
+-- If a presentation or minimal presentation for M has already been computed, we use that.
+addHook((hilbertFunction, List, Module), Strategy => Base, (opts, L, M) -> (
+	if any(select(keys M.cache, Option), o -> o#0 === symbol minimalPresentation) then
+	rank source basis(L, minimalPresentation M)
+	else if not M.?generators then rank source basis(L, M)
+	else if M.cache.?presentation then rank source basis(L, cokernel presentation M)
+	else (rank source basis(L, cokernel relations M)) - (rank source basis(L, cokernel (generators M|relations M)))))
```
mahrud: Also, once #4134 is merged we can use:
```suggestion
	if hasMinPres M then numColumns basis(L, minimalPresentation M)
	else if not M.?generators then numColumns basis(L, M)
	else if M.cache.?presentation then numColumns basis(L, cokernel presentation M)
	else (
	    numColumns basis(L, super M) -
	    numColumns basis(L, cokernel fullgens M))))
```

The changes are in `hasMinPres M`, `super M`, and `fullgens M`.

mahrud (approved):

PR #4132, opened 2026-02-15
Update SchurComplexes.m2

michaelPerlman:
Fixed a bug regarding how the function schurComplex keeps track of homological degree. Also, updated schurComplex to output a graded complex, based on the grading on the input complex.


M2/Macaulay2/packages/SchurComplexes.m2
mahrud, line 1: I think the type of the file changed into DOS format, where the line breaks are different than Mac/Linux formats, hence it looks like everything in the file has changed. You could fix this with `dos2unix` I think, or in the lower left corner of emacs by clicking where it says "DOS".

michaelPerlman: I will try that. What should I do after? Close this pull request and initiate a new one? Or is there a way to edit a pull request?

d-torrance: You can modify an existing pull request by just pushing new commits to the same branch.  (Or depending on your comfort level with git, amending the existing commits and force pushing.)

d-torrance (approved):
Thanks!

M2/Macaulay2/packages/SchurComplexes.m2
michaelPerlman: Thanks @mahrud and @d-torrance !

Was my second attempt correct? I just want to make sure I do it right for next time. I clicked “DOS” in eMacs to change the file per Mahrud’s suggestion and then I re-committed.

d-torrance: Yup -- it worked!  I "squashed" the two commits when I merged so we won't see the Unix -> DOS -> Unix changes in the git history.

PR #4133, opened 2026-02-18
Use Jansson for JSON parsing

d-torrance:
Currently, our JSON parser is implemented at top level using the *Parsing* package.  This is fine for small things, but not so great for larger ones.

As we have several JSON-reliant projects on the horizon (like MRDI serialization and LSP support), it would be good to improve our JSON support.

We swap our JSON parser for [Jansson](https://github.com/akheron/jansson), a C library that was recently used by Emacs for JSON support (they've since rolled their own). It's widely available as a package on most systems, so the additional dependency shouldn't be too much of a burden.

The improvement is staggering -- a 1.2MB MRDI file containing a big polynomial that used to take ~5 min to parse now takes about a second!

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -64,7 +64,7 @@ jobs:
           brew config
           brew update
           brew tap macaulay2/tap
-          brew install automake bison boost libtool tbb ccache ctags llvm make yasm libffi msolve googletest fplll eigen
+          brew install automake bison boost libtool tbb ccache ctags llvm make yasm libffi msolve googletest fplll eigen jansson
           brew install texinfo || true # sometimes post-install step fails
           brew install --only-dependencies macaulay2/tap/M2
           brew link factory --force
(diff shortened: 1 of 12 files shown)


d-torrance:
Discussed with Mike -- will merge after fixing conflict

mahrud:
This is yet another dependency that isn't easily available or installable on the HPCs I use, and it's really time consuming having to deal with this over and over (there is a collection of several servers and I have to do the same thing on whichever is available for use). Can we please make this optional or role it back or maybe hardcode the library in the repository?

d-torrance:
Sure -- it would be straightforward to make it optional

mahrud:
Does this library do everything you need? 
https://github.com/DaveGamble/cJSON
Maybe we could add a directory for external libraries that are bundled in with the repository, like mpreal.h.

d-torrance:
It probably would work, but I just finished putting together #4150 making Jansson optional.  :)

Also, putting my Debian developer hat on, I'd prefer to keep vendored libraries to a minimum.

PR #4134, opened 2026-02-18
Quick fixes involving gb and prune

mahrud:
- **added cached helper function 'fullgens'**
- **added helper function 'hasMinPres'**
- **added two timing tests for hilbertFunction**


M2/Macaulay2/tests/normal/hilbertFunction.m2
```diff
@@ -1,3 +1,12 @@
+-- see https://github.com/Macaulay2/M2/pull/4130
+(t, n) = toSequence timing hilbertFunction(5, QQ[x_0..x_100])
+assert(t < 1 and n == 96560646)
+
+R = QQ[x_0..x_5]/(x_0^8+x_1^8+x_2^8+x_3^8+x_4^8)
+M = module ideal random({2,2,3}, ideal vars R);
+(t, n) = toSequence timing hilbertFunction(100, M)
+assert(t < 10 and n == 32105454)
```
d-torrance: I'm always wary of timing tests because they can fail on slower systems.  This takes about 8s on my system, and so it would almost certainly fail on some of the Debian build machines, for example.

mahrud: Shouldn't `timing` and `time` measure cpu time? Is there a way to basically count how many operations were used? (e.g. allocation/deallocations, or function enter/exits)

d-torrance: Yeah, it's cpu time, but it's still time.  As far as counting the number of operations, maybe we can use `profile`?

d-torrance (approved):

M2/Macaulay2/tests/normal/hilbertFunction.m2
d-torrance: As, I feared, in one of the 32-bit PPA builds last night:

```m2
oo6 = (41.0934, 32105454)

oo6 : Sequence

ii7 : assert(t < 30 and n == 32105454)
hilbertFunction.m2:8:6:(3):[5]: error: assertion failed
```

I'm just going to comment out that particular assertion in the Debian package rather than playing whack-a-mole trying to find a sufficiently large time for all systems.

mahrud: Ah, unfortunate. Would be great to have a normalized elapsedTime method.

PR #4136, opened 2026-02-20
Add libxml2 to version hash table

d-torrance:
I think it's the only library we directly link against that wasn't included

```m2
i1 : version

o1 = HashTable{"architecture" => x86_64             }
               .
               .
               .
               "libxml2 version" => 2.9.14
               .
               .
               .
```


d-torrance:
Discussed with Mike

PR #4139, opened 2026-02-20
remove bugs directory

antonleykin:
This removes all bugs :smile:
---
@mikestillman 
Here is what I did to copy `bugs/` to my private repo 
```
git clone git@github.com:Macaulay2/M2.git temp
cd temp
git filter-repo --subdirectory-filter bugs/ --
cd ~/R/Macaulay2-bugs/
git remote add source-clone ~/M2repos/Macaulay2/temp/
git fetch source-clone
git merge source-clone/stable --allow-unrelated-histories
git commit -m "merge: planting all bugs"
``` 
Above:
* `temp` is a throw-away clone that is at `~/M2repos/Macaulay2/temp/` for me.
* `Macaulay2-bugs` is my private repo at `~/R/Macaulay2-bugs/`


d-torrance:
Should we turn the old bugs into issues?  I think we could probably do it programmatically using `gh`.

antonleykin:
> Should we turn the old bugs into issues?  I think we could probably do it programmatically using `gh`.

I don't think so. Most of these files are outdated. The best thing we can do is clean up the repo, and then perhaps Mike and I could create issues out of a handful of bugs that persisted to these days.

mikestillman:
I will make Macaulay2-mike-bugs repo per roughly your instructions (private?  not sure that is needed).  And then I will also remove those bug files (well, add them to a PR, possibly this one?)

mikestillman:
git doesn't have a command `git filter-repo`.  It seems this is a third party thing.

It can be installed in homebrew via
```bash
brew install git-filter-repo
```

@antonleykin Is this what I should use in order to follow your directions above?

mikestillman:
@antonleykin did you use the commands you gave above?  I also get an error in the `git filter-repo` line you have, after installing git-filter-repo.

mikestillman:
Should we make a repo for each of anton, mike, dan?  (gfurnish doesn't have anything in it).  What about LAcore?  who will take Dan's? @d-torrance ?

antonleykin:
> @antonleykin did you use the commands you gave above?  I also get an error in the `git filter-repo` line you have, after installing git-filter-repo.

I've missed `cd temp`.

antonleykin:
> Should we make a repo for each of anton, mike, dan?  (gfurnish doesn't have anything in it).  What about LAcore?  who will take Dan's? @d-torrance ?

In fact, I'm not sure we even need a separate repo. All the commits that introduced the files in `bugs/` will still be on the main repo. Hence, for example, in a hypothetical future micro-workshop devoted to sifting through all issues we can always examine what is in there.

I just did this exercise of copying subdirectory with history because I wanted a separate (private) repo to go through some of my bugs... but one can do it on a fork... and, say, go back in time to the moment when a particular bug-file was committed.

mahrud:
It would be good to keep track of which ones have been dealt with and which ones need more attention. Just a snapshot in git history doesn't capture this.

It might require a little bit of playing with the GitHub API but I think creating a project under the M2 org and adding all bugs as TODO tasks (with labels like Mike/dan/etc) might be a good alternative.

d-torrance:
Discussed with Mike

mahrud:
Here's a project with all bugs: https://github.com/orgs/Macaulay2/projects/46
Try different view tabs. Can easily search or swipe them into in progress or ready or just delete whatever isn't relevant.

For reference, here is what the API calls looked like:
```
addItem() {
    GH_TOKEN=[token] gh api graphql -f query="
  mutation {
    addProjectV2DraftIssue(input: {projectId: \"PVT_kwDOAC6Xfc4BQEgX\" title: \"$1\" body: $2}) {
      projectItem {
        id
      }
    }
  }"
    }

for f in `find dan/ | grep -v "/$"`; do
    content=`cat bugs/$f | jq -sR .`
    echo $f
    addItem $f "$content";
done
```

PR #4140, opened 2026-02-23
Don't create single-letter symbols for "export" error messages

d-torrance:
There's no reason to call `getGlobalSymbol` since we're not going to end up exporting a symbol.

Plus the error message was strange:

```m2
i1 : export "x"
stdio:1:6:(3): error: cannot export single-letter symbol 'x'
-* invalid file position *-: here is the first use of 'x'
```
This gets rid of the second part.


d-torrance:
Discussed with Mike

PR #4141, opened 2026-02-24
Double length of time allowed for release-checklist test (autotools)

d-torrance:
It's been timing out a lot recently in the GitHub builds because we cap the `tests/normal` tests at 90s.


d-torrance:
Discussed with Mike

PR #4145, opened 2026-02-25
Allow abortions

### LINKED ISSUE #1099: Return to interpreter after interrupt
Seems like https://github.com/Macaulay2/M2/commit/5fb48f38762fa6c929fdfe1502c753c3ac070ea4 disabled the option to jump back to the interpreter after interrupting a command *in 2006*. There is still dead code (in the sense that it never runs) in M2lib.c related to this.

Questions:
- Do we still want this feature?
- Does it work with multithreading?

mahrud:
Fixes #1099.

### DIFF
--- a/M2/Macaulay2/bin/main.cpp
+++ b/M2/Macaulay2/bin/main.cpp
@@ -272,6 +272,7 @@ void interrupt_handler(int sig) {
 	      fprintf(stderr,"returning to top level\n");
 	      fflush(stderr);
 
+	      interp_setInterpreterDepth(0);
 	      interrupts_clearAlarmedFlag();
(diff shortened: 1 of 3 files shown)


mahrud:
@d-torrance could you add a note to the release notes to check that changes to startup.m2 don't break this in the future?

d-torrance:
This is great!  I just tested this and aborting actually works now!

I'm going to go ahead and merge this, but I'm not quite sure why it worked -- how did moving around the `firstTime` stuff fix things?

mahrud:
Great question ... I actually don't remember because I wrote this months ago and haven't had any issues with it since XD

PR #4147, opened 2026-02-25
Improvements in WeylAlgebras

mahrud:
- **fixed documentation of WeylAlgebras**
- **added minor improvements in WeylAlgebras**
- **cached holonomic basis in holonomicRank**

### DIFF
--- a/M2/Macaulay2/packages/WeylAlgebras/Dbasic.m2
+++ b/M2/Macaulay2/packages/WeylAlgebras/Dbasic.m2
@@ -323,8 +323,9 @@ holonomicRank Module := M -> (
      ltM := leadTerm gens gb WtotempW presM;
      -- compute the rank
      redI := cokernel tempWtoR ltM;
-     -- Note: we don't compute dim redI, because coefficientRing W might nonzero dimension
-     B := try M.cache#"basis" = sub(cover basis(redI, Variables => gens R), matrix {W.dpairVars#1});
+     -- the standard monomials {\partial^\alpha} form a basis for differential operators modulo the system,
+     -- and their number, if finite, determines the holonomic rank by Lemma 1.4.11 and Algorithm 1.4.17 in SST
+     B := try M.cache#"standard monomials" = sub(cover basis(redI, Variables => gens R), matrix {W.dpairVars#1});
      if 0 == redI  then 0 else
      if B === null then infinity
      else numgens source B)
--- a/M2/Macaulay2/packages/WeylAlgebras/Dbasic.m2
+++ b/M2/Macaulay2/packages/WeylAlgebras/Dbasic.m2
@@ -297,12 +297,8 @@ isHolonomic Module := M -> (
      )
 
 -- This routine computes the rank of a D-module
--- QUESTION: this changes the current ring?
 holonomicRank = method()
-holonomicRank Ideal := I -> (
-     holonomicRank ((ring I)^1/I)
-     )
-
+holonomicRank Ideal  := I -> holonomicRank comodule I
 holonomicRank Module := M -> (
      W := ring M;
      createDpairs W;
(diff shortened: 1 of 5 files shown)


M2/Macaulay2/packages/WeylAlgebras/Dbasic.m2
```diff
@@ -308,29 +304,27 @@ holonomicRank Module := M -> (
...
      -- computing GB with respect to new order
      ltM := leadTerm gens gb WtotempW presM;
      -- compute the rank
      redI := cokernel tempWtoR ltM;
-     if dim redI > 0 then holRank := infinity
-     else if redI == 0 then holRank = 0
-     else holRank = numgens source basis redI;
-     holRank
-     )
+     -- Note: we don't compute dim redI, because coefficientRing W might nonzero dimension
```
antonleykin, line 323: What is your definition of the holonomic rank when the coefficient ring is not a field?
Can you give a use example?

antonleykin (changes requested):
Most of the cnages are straightforward and good.
There is an issue in `holonomicRank` that should be addressed  but I'm not sure the current solution is the right one.

M2/Macaulay2/packages/WeylAlgebras/Dbasic.m2
mahrud: Good question! I haven't thought about the case when the coefficient ring is not a field, but as for use case, this is the backbone of the ConnectionMatrices packages, where a modified version of this method is currently being used:
https://github.com/Macaulay2/M2/blob/stable/M2/Macaulay2/packages/ConnectionMatrices/holonomic.m2

mahrud: I guess really the question is what does "basis" do when the coefficient ring is not a field, and I think it just gives a generating set? I'm not an expert, is holonomic rank over integers well defined?

antonleykin: I'm not sure how to resolve this. 
@mahrud, can you figure out what ConnectionMatrices authors think?
@michaelPerlman @lorinczandras @christineBerkesch , any opinion on generalization of holonomic rank?

mahrud: Sorry, I think I'm confused. I thought you were asking out of curiosity. Is anything else in Dmodules packages designed with non-field coefficient rings in mind? The package tutorial explicitly says:
> This package is mostly concerned with computations in the Weyl algebra, the ring of differential operators over affine space (over a field of characteristic zero).

I also can't find any algorithms in the SST book that don't assume working over a field either.

antonleykin: Please revert changes in `holonomicRank` that attempted to deal with non-field coefficients case.
Then we merge the PR.
(Then @michaelPerlman  and @lorinczandras make this method and others foolproof.)

mahrud: > Please revert changes in `holonomicRank` that attempted to deal with non-field coefficients case.

I still don't understand. I don't think any changes in this PR deal with non-field coefficients case. Am I missing something?

The primary reason for this PR is to cache the basis (over a field) in the finite holonomic rank case, and the rest are cosmetic changes. Do you prefer a different implementation?

antonleykin:
I read too quickly... Please replace 
```
-- Note: we don't compute dim redI, because coefficientRing W might nonzero dimension
```
with, say
```
-- TODO: check coefficientRing is a field 
```

antonleykin:
Also, `cache#"basis"` is not used in the trio of D-modules packages... so I'd not cache anything, or `cache.holonomicRank`. 

(`Connection...` can cache the exact thing it desires since it rewrites the method.)

antonleykin:
For readibilty, I suggest to revert to just 
```
numgens source basis redI
```

mahrud:
Will replace the comment soon.

> Also, `cache#"basis"` is not used in the trio of D-modules packages... so I'd not cache anything, or `cache.holonomicRank`.
>
> (`Connection...` can cache the exact thing it desires since it rewrites the method.)

My hope was to remove the duplicate version in ConnectionMatrices once holonomicRank caches the basis. I imagine there might be other applications where having the basis is useful as well. Happy to cache it under a different name or document it so it's an official feature if you prefer.

mahrud:
I expanded the comment to elaborate on why the standard monomials are being cached.

mahrud:
Also added an example for it in the documentation.

mahrud:
This is passing tests now.

antonleykin:
Several notes for @michaelPerlman :
* I see several $\CC$ fixed in the docs in this PR  --- you'll have to go carefully through the conflicts with your Dmodules branch 
* Something like `holonomicBasis` (perhaps there is a better name) makes sense to create in the same package as `holonomicRank`.

antonleykin (approved):

PR #4150, opened 2026-02-28
Make Jansson optional

d-torrance:
If we build with `--without-janson` (autotools) or `-DWITH_JANSSON=OFF` (cmake), then we fall back on the top-level JSON parsing using the *Parsing* package.

We also fix our handling of [UTF-16 surrogate pairs](https://en.wikipedia.org/wiki/UTF-16#U+D800_to_U+DFFF_(surrogates)) (which I noticed when making sure the Jansson and top-level behaviors agreed for the JSON package tests):

### Before
```m2
1 : "\ud834\udd1e"

o1 = ������
```

### After
```m2
i1 : "\ud834\udd1e"

o1 = 𝄞
```

Draft for now to test the GitHub builds with Jansson disabled.

### DIFF
--- a/M2/Macaulay2/d/CMakeLists.txt
+++ b/M2/Macaulay2/d/CMakeLists.txt
@@ -114,6 +114,9 @@ endif()
 if(NOT WITH_FFI)
   list(REMOVE_ITEM DLIST ffi.d)
 endif()
+if(NOT WITH_JANSSON)
+  list(REMOVE_ITEM DLIST json.d)
+endif()
 
 ###############################################################################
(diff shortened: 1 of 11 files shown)


d-torrance:
Builds w/o Jansson passed!

https://github.com/Macaulay2/M2/actions/runs/22521873617?pr=4150

mahrud (commented):
I can't evaluate any of the json changes, but of course I'm glad that this is optional now (though would still prefer a permanent solution like bundling or what Emacs did).

d-torrance:
Discussed with Mike

PR #4153, opened 2026-03-01
Run autotools build in parallel

d-torrance:
We remove most of the `.NOTPARALLEL:` targets in the Makefiles and add some prereqs so we can run `make -jX` and run the autotools build in parallel.

The d directory can't be built in parallel at the moment since because we need the have already built the .sig files for each file's dependencies to generate its .dep file -- but we need the .dep file to know the dependencies...


d-torrance:
Discussed with Mike

PR #4167, opened 2026-03-19
Load package's PackageExports when capturing example output

### LINKED ISSUE #4163: Loading PackageExports when capturing examples
I think there's something missing in #3952 because `PackageExports` don't seem to be loaded when capturing the examples, thus `capture` is failing many more times than it should.

d-torrance:
Fixes #4163


mahrud (approved):
Ah, I forgot that I put this option explicitly for this purpose XD Thanks!

Issue #4172, opened 2026-03-26
Feature Request: Barycenter of polytope


anna-brosowsky:
I would love it if there were a function (say, in the [Polyhedra](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Polyhedra/html/index.html) package) that would compute the barycenter aka centroid aka center of mass of a polytope. I imagine some of the code already written for [barycentricTriangulation](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Polyhedra/html/_barycentric__Triangulation.htmll) would be helpful?

I am happy to work on this! I would probably be doing this via the algorithm of: triangulate into simplices, where it is easy to compute the centroid, then take the weighted averages of these centroids to get my centroid of the original polytope.

d-torrance:
Added in #4173

PR #4173, opened 2026-03-26
Added centroid method

seangrate:
Added a method to compute the centroid of a polyhedron. This implements the feature requested in #4172.

### DIFF
--- a/M2/Macaulay2/packages/Polyhedra/documentation/documentation.m2
+++ b/M2/Macaulay2/packages/Polyhedra/documentation/documentation.m2
@@ -108,24 +108,23 @@ doc ///
       centroid
       (centroid,Polyhedron)
    Headline
-      Computes the centroid or barycenter of a polyhedron.
+      computes the centroid or barycenter of a polyhedron
    Usage
       centroid P
    Inputs
       P:Polyhedron
+         a compact polyhedron
    Outputs
        :Matrix
    Description
       Text
-         The centroid a polyhedron is the center of mass of the polyhedron.
+         The centroid of a polyhedron is the center of mass of the polyhedron.
          This is computed by first triangulating the polyhedron, and then taking
          a weighted sum of the barycenters of the simplices in the triangulation.
       Example
-         centroid simplex 3
+         centroid stdSimplex 3
       Example
          centroid hypercube 2
-   Caveat
-      The polyhedron must be compact.
    SeeAlso
       barycentricTriangulation
 ///
--- a/M2/Macaulay2/packages/Polyhedra/documentation/documentation.m2
+++ b/M2/Macaulay2/packages/Polyhedra/documentation/documentation.m2
@@ -105,6 +105,33 @@ doc ///
 
 doc ///
    Key
+      centroid
+      (centroid,Polyhedron)
+   Headline
+      Computes the centroid or barycenter of a polyhedron.
+   Usage
+      centroid P
+   Inputs
+      P:Polyhedron
+   Outputs
+       :Matrix
+   Description
+      Text
+         The centroid a polyhedron is the center of mass of the polyhedron.
+         This is computed by first triangulating the polyhedron, and then taking
+         a weighted sum of the barycenters of the simplices in the triangulation.
+      Example
+         centroid simplex 3
+      Example
+         centroid hypercube 2
+   Caveat
+      The polyhedron must be compact.
+   SeeAlso
+      barycentricTriangulation
+///
+
+doc ///
+   Key
       latticeVolume
       (latticeVolume,Polyhedron)
    Headline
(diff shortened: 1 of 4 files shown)


M2/Macaulay2/packages/Polyhedra/exports.m2
```diff
@@ -113,6 +113,7 @@ export {
    "barycentricTriangulation",
    "regularSubdivision",
    "minimalNonFaces",
-   "stanleyReisnerRing"
+   "stanleyReisnerRing",
+   "centroid"
```
mahrud: Sidenote: this way next time someone adds a function they don't edit two lines, which can sometimes avoid merge conflicts.
```suggestion
   "centroid",
```

M2/Macaulay2/packages/Polyhedra/documentation/documentation.m2
```diff
@@ -103,6 +103,33 @@ doc ///
       Deprecated version of @TO "linealitySpace"@
 ///
 
+doc ///
+   Key
+      centroid
+      (centroid,Polyhedron)
+   Headline
+      Computes the centroid or barycenter of a polyhedron.
```
mahrud: Following the convention in the rest of the package:
```suggestion
      computes the centroid or barycenter of a polyhedron
```

M2/Macaulay2/packages/Polyhedra/documentation/documentation.m2
```diff
@@ -103,6 +103,33 @@ doc ///
...
+   Usage
+      centroid P
+   Inputs
+      P:Polyhedron
+   Outputs
+       :Matrix
+   Description
+      Text
+         The centroid a polyhedron is the center of mass of the polyhedron.
```
mahrud:
```suggestion
         The centroid of a polyhedron is the center of mass of the polyhedron.
```

M2/Macaulay2/packages/Polyhedra/documentation/documentation.m2
```diff
@@ -103,6 +103,33 @@ doc ///
...
+         This is computed by first triangulating the polyhedron, and then taking
+         a weighted sum of the barycenters of the simplices in the triangulation.
+      Example
+         centroid simplex 3
+      Example
+         centroid hypercube 2
+   Caveat
+      The polyhedron must be compact.
```
mahrud: Others can correct me if I'm mistaken, but I believe this is meant for the caveats of the algorithm or implementation, not assumptions of the definition. You could mention in the `Inputs` section that P must be compact.

M2/Macaulay2/packages/Polyhedra/extended/polyhedron/methods.m2
```diff
@@ -303,6 +303,20 @@ interiorPoint Polyhedron := P -> (
...
+--   INPUT : 'P',  a Polyhedron
+--  OUTPUT : 'p',  a point given as a matrix
+centroid = method(TypicalValue => Matrix)
+centroid Polyhedron := Matrix => P -> (
+    if not isCompact P then error "the polyhedron must be compact";
+    totalVolume := volume P;
+    sum for delta in barycentricTriangulation P list (
+        barycenter := (sum delta) / #delta;
+        (volume convexHull delta / totalVolume) * barycenter
+    )
```
mahrud: This is totally fine as as, but FYI, this also works:
```suggestion
    sum(barycentricTriangulation P, delta -> (
            barycenter := (sum delta) / #delta;
            (volume convexHull delta / totalVolume) * barycenter
        )
    )
```
The advantage (which is negligible here) is that unlike this way we don't need to construct a list and then sum it, we just sum as we go.

mahrud (commented):
Thanks! This is great, I left a few comments ranging from typos to just some FYIs in case you're interested.

M2/Macaulay2/packages/Polyhedra/extended/polyhedron/methods.m2
seangrate: Didn't realize `sum` could do this, thanks!

M2/Macaulay2/packages/Polyhedra/documentation/documentation.m2
seangrate: At least in `Polyhedra/documentation/documentation.m2`, it looks like the main convention is a capitalized first letter and the phrase ends with a period.

mahrud: Huh, that's definitely not the suggestion of the style guide, but from the [toc](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Polyhedra/html/toc.html) it seems only a few nodes are like that.

mahrud (approved):

PR #4181, opened 2026-04-07
Update SpecialFanoFourfolds from 2.7.1 to 2.8

giovannistagliano:
Added support for the DoubleSpecialCubicFourfold class (cubic fourfolds in the intersection of two Hassett divisors). Might follow up with some minor updates in the coming days.


mahrud:
Totally up to you, but you might want to consider breaking up the main file into smaller files in an auxiliary directory, e.g. docs.m2, tests.m2, or even separating by class like DoubleSpecialCubicFourfold.m2, etc. There are only 3 packages longer than 6k lines, and imho those should be split as well.

giovannistagliano:
> Totally up to you, but you might want to consider breaking up the main file into smaller files in an auxiliary directory, e.g. docs.m2, tests.m2, or even separating by class like DoubleSpecialCubicFourfold.m2, etc. There are only 3 packages longer than 6k lines, and imho those should be split as well.

Thanks for the feedback. Since the new additions are quite separate, I am considering moving them into a new package.

Between refactoring the current package into multiple files or creating this new one, which approach do you think is better for M2? I would personally prefer the second option to make future updates easier to manage.

mahrud:
If it's still about fano fourfolds, or if the intersection of tools and methods is high enough, then multiple files in the same package makes more sense to me.

giovannistagliano:
That makes sense. I'll break up the main file into several smaller ones and update the PR once it's ready. Thanks!

d-torrance:
@giovannistagliano - are there any more updates coming, or is this good to go?

giovannistagliano:
I still need to make a few small updates, almost done...

giovannistagliano:
I believe this is ready on my side

d-torrance (approved):
Thanks!

PR #4186, opened 2026-04-08
Add support for "try f" (when f is a function)

### LINKED ISSUE #4184: try modifier in apply,select,etc
Not long ago we added support for something like:
```m2
i6 : select(-5..5, even and not zero)

o6 = (-4, -2, 2, 4)
```
Which is pretty nice. Basically `not Function` and `Function and Function` return a function that negate the output or and the outputs.

Would be neat if this was also supported:
```m2
apply(10, try myfunc)
```

Maybe this is not feasible because there's an actual double meaning here, in which case case I'm happy to close this.

d-torrance:
This returns a new compiled function closure that is equivalent to `x -> try f x`.

Closes #4184

### DIFF
--- a/M2/Macaulay2/d/evaluate.d
+++ b/M2/Macaulay2/d/evaluate.d
@@ -1323,10 +1323,34 @@ tryEval(c:Code):Expr := (
 	else tryCaughtError = false);
     p);
 
+-- env.0 = function
+-- env.1 = try code
+tryFunction(x:Expr, env:Sequence):Expr := (
+    if length(env) == 2 then (
+        when env.1
+        is c:Pseudocode do (
+            when c.code
+            is t:tryCode do (
+                pos := codePosition(t.code);
+                eval(Code(tryCode(
+                            Code(adjacentCode(
+                                    evaluatedCode(env.0, pos),
+                                    evaluatedCode(x, pos),
+                                    pos)),
+                            t.thenClause, t.elseClause, t.doClause,
+                            t.frameID, t.framesize, t.position))))
+            -- none of the following should happen
+            else WrongArg("environment to contain a try code"))
+        else WrongArg("environment to contain a pseudcode"))
+    else WrongArg("environment to have length 2"));
+
 evalTryCode(c:tryCode):Expr := (
     ret := tryEval(c.code);
     -- certain errors should not be caught, see above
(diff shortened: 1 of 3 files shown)


mahrud:
I guess in the current behavior, something like `try(x -> x/0)` never fails until you actually run it, or unless there's a syntax error in which case `try` doesn't prevent the error either, so maybe this is fine?

Does this also allow `select(5, try (x -> (x-1)^1) else false)` for instance? (meaning, try that function, and if it fails instead of null return false). I imagine we can combine it with other `try` syntax as well.

Also, this is tangential, but I've always wanted this to work:
```m2
i5 : apply(5, x -> if even x then continue else x)
stdio:5:29:(3): error: unhandled continue command
```
It's really annoying that I have to use `return` instead and then remove `null` entries.

d-torrance:
> Does this also allow `select(5, try (x -> (x-1)^1) else false)` for instance?

Not currently -- only `try f` is supported.  I suppose implementing all the other variations of `try` wouldn't be too bad.  I can give it a shot.

d-torrance:
I think this is ready.  ~~One thing I'm not 100% sure about, though.  If `f` is a function, then should `try f then 2` return a function that always return 2 unless there was an error, or should it just return 2?  (I'm leaning towards the former for consistency.)~~

*Edit:*  Of course it should return a function.  The more I thought about it, the more I realized that it was dumb that I wan't sure!  That's what it does now.

mahrud:
Yeah, I think of it like evaluating a polynomial `x^2+2*x+3` on a matrix, then `3` should become 3 times the identity matrix. We're essentially implementing algebraic operations on functions, so the output has to be a function too, e.g. can we define:
```m2
f = n -> n/2
g = n -> 3*n+1
h = if even then f else g
```
Is it mathematically or computationally useful? I'm not sure! But it's convenient in something like this:
```m2
apply(100, if even then f else g)
```
We already have `Function [bool-op] Function` like:
```m2
select(10, zero or not even)
```
Then the question is whether `if/try/while/etc` should behave the same.

mahrud (approved):

d-torrance:
Hmm, I'm not sure if the current proposed behavior is correct either.  Right now `(try f else g) x` will return `f x` when there's no error, but just the function object `g` if there is.  So it probably should return `g x`, right?

PR #4188, opened 2026-04-09
fix incorrect stderrorS in webapp mode

pzinn:
In webapp mode, `stdError` is redefined to be equal to `stdIO` (to avoid annoying out-of-sync output).
Sadly, the M2 level stderr (`stderrorS`) is not updated when this redefinition takes place, causing subtle bugs (mostly due to `capture` changing the value of `stderrorS`). -> one line fix.


d-torrance (approved):
Thanks!

PR #4196, opened 2026-04-17
SchurRings extension and overhaul

kellerlv:
This overhaul of SchurRings was done in preparation for a forthcoming package on symmetric functions and character computations, which needed the `SchurRing` type to carry several character theories beyond the general linear group. The main extension is in three new character rings, selected via the `GroupActing` option on `schurRing`:

- `"Sp"` — character ring of the symplectic group; multiplication is the Littlewood–Richardson product on the GL side composed with the Koike–Terada type-C modification rule.
- `"O"` — the analogous orthogonal case via Koike–Terada modification rules of types B and D, with `OddOrEven` distinguishing the two.
- `"RatGL"` — Koike's universal rational-GL character ring on bi-partitions (λ, μ), specializing to GL_n for arbitrary n.

The new ring types carry the full API of the original `SchurRing`: multiplication, plethysm, basis conversions (`toS`/`toE`/`toH`/`toP`), dimension, and specialization. Explicit converters between the ring types (`toGL`, `toSp`, `toO`, `toSn`, …) are provided; for restriction from GL to Sp and to O these implement the King–Littlewood branching formulas. Each variant has a stable (infinitely many variables) form, with `specialize` carrying out the branching to finite rank.

Alongside the Schur basis, a monomial-basis variant of `SchurRing` is added via the new `Basis => "Monomial"` option. Multiplication in this ring is carried out through the Kostka matrix between the Schur and monomial bases; `toM` and `kostkaNumber` are exposed directly.

The existing functionality of SchurRings is unchanged aside from a few efficiency improvements.

### DIFF
--- a/M2/Macaulay2/packages/SchurRings.m2
+++ b/M2/Macaulay2/packages/SchurRings.m2
@@ -18,29 +18,35 @@
 
 newPackage(
 	"SchurRings",
-    	Version => "1.1", 
-    	Date => "August 24, 2011",
+    	Version => "2.0",
+    	Date => "April 17, 2026",
     	Authors => {
 	     {Name => "Michael Stillman", Email => "mike@math.cornell.edu", HomePage => "http://www.math.cornell.edu/~mike/"},
 	     {Name => "Hal Schenck"},
-	     {Name => "Claudiu Raicu", Email => "claudiu@math.berkeley.edu", HomePage => "http://math.berkeley.edu/~claudiu/"}
+	     {Name => "Claudiu Raicu", Email => "craicu@nd.edu", HomePage => "http://math.berkeley.edu/~claudiu/"},
+	     {Name => "Keller VandeBogert", Email => "keller.v@uky.edu", HomePage => "https://sites.google.com/view/kellervandebogert/home"}
 	     },
 	Keywords => {"Representation Theory"},
-    	Headline => "representation rings of general linear groups and of symmetric groups"
---	AuxiliaryFiles => true
+    	Headline => "representation rings of general linear groups and of symmetric groups",
+	DebuggingMode => false
     	)
 
 export {"schurRing", "SchurRing", "symmetricRing",
-     "toS", "toE", "toP", "toH",
+     "toS", "toE", "toP", "toH", "toM", "toSp", "toO", "specialize",
+     "kostkaNumber",
      "jacobiTrudi", "plethysm",
-     "centralizerSize", "classFunction", "symmetricFunction", 
+     "centralizerSize", "classFunction", "symmetricFunction",
      "scalarProduct", "internalProduct",
      "SchurRingIndexedVariableTable", "EHPVariables", "SVariable",
      "ClassFunction", "schurLevel",
      "schurResolution",
-     "SchurRingElement",     
-     "Memoize", "Schur", "EorH", "GroupActing",
-     "eVariable", "pVariable", "hVariable"
+     "SchurRingElement",
+     "Memoize", "Schur", "EorH", "GroupActing", "Basis", "OddOrEven",
+     "eVariable", "pVariable", "hVariable",
+     "modificationRule",
+     "branch",
+     "toRatGL",
+     "toSymm", "toGL", "toSn", "convert"
      }
 
 debug Core
@@ -62,6 +68,31 @@ protect symbol HtoPTable
 --protect symbol plethysmMaps
 protect symbol mapFromE
 protect symbol sFunction
+-- dispatch attributes for variant support (GL/Sn/Sp/O, Monomial, etc.)
+protect symbol multiplySchurLevel1
+protect symbol highLevelCombine
+protect symbol plethysmFcn
+protect symbol recTransOp
+protect symbol kostkaCache
+protect symbol monomialBasisRing
+protect symbol symplecticBasisRing
+protect symbol orthogonalBasisRing
+protect symbol schurBasisRing
+protect symbol stableSchurHelper
+protect symbol plethysmHelpers
+protect symbol specializedSpRings
+protect symbol specializedORings
+protect symbol specializedGLRings
+protect symbol specializedSLRings
+protect symbol specializedRatGLRings
+protect symbol specializedSnRings
+protect symbol associatedRatGLRing
+protect symbol ratNegRing
+protect symbol ratNegSym
+protect symbol ratPosSym
+protect symbol ratOuterRing
+protect symbol ratRank
+protect symbol ratStableHelper
 
 
 SchurRing = new Type of EngineRing
@@ -78,20 +109,40 @@ expression SchurRing := S -> (
     else new FunctionApplication from unhold describe S)
 undocumented (expression, SchurRing)
 
+-----------------------------------------------------------------------------
+-- Engine boundary: SchurRing plumbing.
+-- This section wraps the C++-side `rawSchurRing1` engine (which implements
+-- Littlewood-Richardson multiplication on Schur-partition monomials) as an
+-- M2 `SchurRing` type.  The engine stores elements as polynomials in
+-- monomials-that-are-partitions; the helpers below convert to/from that
+-- representation and install element-access sugar (s_lambda, etc.) on top
+-- of the raw engine ring.
+-----------------------------------------------------------------------------
+
 rawmonom2partition = (m) -> (
+     -- Decode an engine Schur monomial (= a raw monomial representing s_lambda)
+     -- into its partition.  `rawSparseListFormMonomial m` returns pairs
+     -- (x,e) = (row-length, multiplicity); `e:x` builds an e-tuple of x's,
+     -- `splice` flattens the list of tuples into the multiset of row lengths,
+     -- and `reverse` puts them in weakly-decreasing (partition) order.
      reverse splice apply(rawSparseListFormMonomial m, (x,e) -> e:x)
      )
 
 --various ways of addressing elements of a Schur ring
+-- s_{lambda_1, lambda_2, ...} from an explicit partition list
 SchurRing _ List := (SR, L) -> new SR from rawSchurFromPartition(raw SR, L)
+-- s_(lambda_1, lambda_2, ...) from a partition given as a sequence
 SchurRing _ Sequence := (SR, L) -> new SR from rawSchurFromPartition(raw SR, L)
+-- s_L for a single-row partition L; `1:L` is M2 syntax for the length-1 tuple (L)
 SchurRing _ ZZ := (SR, L) -> new SR from rawSchurFromPartition(raw SR, 1:L)
 --
 coefficientRing SchurRing := Ring => R -> last R.baseRings
 numgens SchurRing := Ring => R -> R.numgens
 
---n = schurLevel R is the number of iterations of the schurRing/symmetricRing function
---used in the construction of R
+-- `schurLevel R` = nesting depth: the number of times schurRing/symmetricRing
+-- was applied to build R.  Plain rings (e.g. QQ) have level 0; a single
+-- schurRing(QQ, s, n) has level 1; schurRing(schurRing(QQ, t, m), s, n) has
+-- level 2; each wrap increments the stored value by 1 (see newSchur2 below).
 schurLevel = method()
 schurLevel (Ring) := R -> if R.?schurLevel then R.schurLevel else 0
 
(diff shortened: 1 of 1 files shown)


pzinn:
I wonder how this compares to `character` from the package `LieAlgebraRepresentations`, which computes characters of representations of arbitrary semi-simple Lie algebras. (the package also computes tensor products, etc)

kellerlv:
> I wonder how this compares to `character` from the package `LieAlgebraRepresentations`, which computes characters of representations of arbitrary semi-simple Lie algebras. (the package also computes tensor products, etc)

I guess it depends on how you want to compare. As I see it, this is a completely different utility, though I didn't benchmark the Newell--Littlewood method here compared with the Racah--Speiser method. But having the characters actually treated formally as ring-theoretic objects is extremely convenient for my purposes (as well as the flexibility of converting between monomial/general linear/symplect/orthogonal bases). Likewise, the whole specialization phenomenon that deals with the failure of characters to behave "stably" in low dimensions is a phenomenon that Lie theory does not see. The rational character ring has no real Lie theoretic analogue since the sl-theory sees determinant twists as trivial representations.

That being said, if you are suggesting that some portion of LieAlgebraRepresentations contains any of the utilities here, I'd be interested in seeing how.

M2/Macaulay2/packages/SchurRings.m2
```diff
@@ -130,16 +198,50 @@ newSchur2(Ring,Symbol,ZZ) := (A,p,n) -> (
...
+     for term in listForm f do (
+	  lam := term#0;
+	  c := term#1;
+	  if #lam > n then continue;  -- engine usually prevents this; be safe
+	  lamList := toList lam | toList((n - #lam) : 0);
+	  k := lamList#(n-1);
+	  lamNew := (
+	       if k == 0 then toList lam
+	       else (
+		    res := for i from 0 to n-1 list lamList#i - k;
```
d-torrance: I'm kind of surprise this didn't raise an error, because `res` is a protected symbol (synonym for `resolution`).  I'd probably choose a different variable name to avoid confusion

M2/Macaulay2/packages/SchurRings.m2
```diff
@@ -156,78 +258,407 @@ dim(List,SchurRingElement) := (lis,s) -> dimSchur(lis, s);
...
+     if #alpha + #beta > n then {}
+     else {(alpha, beta, 1)}
+     )
+
+-- Enumerate all partitions contained componentwise in the given bound.
+-- Result includes the empty partition and the bound itself; trailing zeros
+-- are stripped.
+allSubpartitionsBoundedBy = (bound) -> (
+     b := toList bound;
+     while #b > 0 and b#(-1) == 0 do b = drop(b, -1);
```
d-torrance: `drop` creates a brand-new list every time it's called.  I'd figure out the position where the 0's start and do a single call to `drop` at the end, similar to what's happening with `mu` in the final `apply` at the end of this function.  In fact, it may make sense to just write a "drop trailing zeros" helper function and call it in both places

M2/Macaulay2/packages/SchurRings.m2
```diff
@@ -156,78 +258,407 @@ dim(List,SchurRingElement) := (lis,s) -> dimSchur(lis, s);
...
+     if #b == 0 then return {{}};
+     aux := (idx, prevMax) -> (
+	  if idx >= #b then return {{}};
+	  upper := min(b#idx, prevMax);
+	  result := new MutableList;
+	  v := 0;
+	  while v <= upper do (
+	       for t in aux(idx+1, v) do result#(#result) = prepend(v, t);
```
d-torrance: This isn't efficient because we're allocating a brand new list at every iteration.

Do you know how long `result` will be ahead of time?  (It's hard to tell with the recursion...)  If so, then you could construct `result` at the very beginning with the required length. 

(This will become a non-issue if/when #4103 is merged and mutable lists are allocated more efficiently.)

M2/Macaulay2/packages/SchurRings.m2
```diff
@@ -156,78 +258,407 @@ dim(List,SchurRingElement) := (lis,s) -> dimSchur(lis, s);
...
+     -- (4) Bipartition-syntax override via IndexedVariableTable
+     --     Override so that user-facing bipartition syntax
+     --       r_{{a1,a2,...}, {b1,b2,...}}   ->   r_{(alpha, beta)}
+     --     builds the basis element directly.  At finite rank, if
+     --     #alpha + #beta > n the character reduces via Koike-Terada;
+     --     we apply that modification rule here.
+     ------------------------------------------------------------------
+     t := value p;
+     t#symbol _ = a -> (
+	  isListish := x -> (class x === List or class x === Sequence);
```
d-torrance: Do you want any sort of list-ish thing here?  Would `Array` be "list-ish"?, for example?  It may make sense to use `instance(x, VisibleList)` instead

M2/Macaulay2/packages/SchurRings.m2
```diff
@@ -156,78 +258,407 @@ dim(List,SchurRingElement) := (lis,s) -> dimSchur(lis, s);
...
+	  if isListish a and #a == 2 and isListish (a#0) and isListish (a#1) then (
+	       alpha := toList a#0;
+	       beta := toList a#1;
+	       isFin := not (n === infinity or (class n === InfiniteNumber) or n < 0);
```
d-torrance: This is redundant since the class of `infinity` is `InfiniteNumber` already

M2/Macaulay2/packages/SchurRings.m2
```diff
@@ -623,9 +1337,49 @@ toSymm(RingElement) := (ps) ->
...
+		    ));
+	  if hasNeg then error("toSymm: RatGL element has nonzero negative "
+	       | "(beta) components and is not a polynomial character; no "
+	       | "canonical map to a symmetricRing exists.  Use "
+	       | "specialize(f, n) to evaluate at GL(n) instead.");
+	  if #polyTerms == 0 then return 0;
+	  -- maximum #alpha determines the minimum symmetric-ring dim.
+	  maxLen := max apply(toList polyTerms, t -> #(t#0));
+	  nT := numgens S;
+	  targetDim := if nT === infinity or class nT === InfiniteNumber
```
d-torrance: Another redundant `infinity`

M2/Macaulay2/packages/SchurRings.m2
```diff
@@ -696,2444 +1483,7792 @@ toH = method()
...
+-- Complexity per call: O(|lambda|^{ell(mu)} * ell(mu)) in the worst
+-- case, but memoization on (lambda, mu) makes repeat calls O(1) and
+-- makes the full Kostka matrix of degree d an O(p(d)^2) computation.
+--
+-- The memo table kostkaMemo is process-local and keyed on normalized
+-- (stripped) partitions, so repeated calls with equivalent inputs hit
+-- the cache.
+
+kostkaMemo := new MutableHashTable
+stripTrailingZeros = lst -> (
```
d-torrance: Ooh, here's a strip trailing zeros helper function that I suggested you might write above!  Call this in those places

M2/Macaulay2/packages/SchurRings.m2
```diff
@@ -696,2444 +1483,7792 @@ toH = method()
...
+     key := (lam, m);
+     if kostkaMemo#?key then return kostkaMemo#key;
+     -- Peel off the last (smallest) part of m.
+     mLast   := m#(#m - 1);
+     mRest   := drop(m, -1);
+     -- Sum over all sub-partitions nu of lam obtained by removing a
+     -- horizontal strip of size mLast.
+     total := 0;
+     for nu in horizontalStripComplements(lam, mLast) do
+	  total = total + kostkaNumber(nu, mRest);
```
d-torrance: This line can be simplified to use `sum`

M2/Macaulay2/packages/SchurRings.m2
```diff
@@ -696,2444 +1483,7792 @@ toH = method()
...
+-- plethysm can share it.
+skewMemo := new MutableHashTable;
+
+skewSchurExpansion = method()
+skewSchurExpansion(BasicList, BasicList) := (lambda, mu) -> (
+     lam := toList lambda;
+     m := toList mu;
+     -- Strip trailing zeros (normalize so cache keys collide)
+     while #lam > 0 and lam#(-1) == 0 do lam = drop(lam, -1);
+     while #m > 0 and m#(-1) == 0 do m = drop(m, -1);
```
d-torrance: Use the helper function here too

M2/Macaulay2/packages/SchurRings.m2
```diff
@@ -696,2444 +1483,7792 @@ toH = method()
...
+	  if k < 0 then M_(i,j) = 0_R
+	  else if k == 0 then M_(i,j) = 1_R
+	  else M_(i,j) = R.hVariable(k);
+	  );
+     dd := det(matrix M, Strategy => Cofactor);
+     sExp := toS dd;
+     -- listForm of a SchurRingElement is list of (partition, coeff) pairs
+     result := apply(listForm sExp, t -> (t#0, lift(t#1, ZZ)));
+     skewMemo#key = result;
+     result
```
d-torrance: This last line isn't necessary since the previous line (if you remove the semicolon) also evaluates to `result`

M2/Macaulay2/packages/SchurRings.m2
```diff
@@ -696,2444 +1483,7792 @@ toH = method()
...
-     n := R.dim;
-     A := R;
-     e2h := prepend(1_A, for i from 1 to n list (-1)^(i+1)*A_(2*n+i-1));
-     R.EtoHTable = if n == 0 then {1_A} else {1_A} | convolve(e2h,0);
+-- Enumerate partitions mu, contained in lambda componentwise, with all parts even.
+-- Each mu is returned as a list of positive parts (trailing zeros stripped).
+evenRowsSubpartitionsOf = method()
+evenRowsSubpartitionsOf(BasicList) := (lambda) -> (
+     lam := toList lambda;
+     while #lam > 0 and lam#(-1) == 0 do lam = drop(lam, -1);
```
d-torrance: Helper function again (also in the `apply` at the end of this function)

M2/Macaulay2/packages/SchurRings.m2
```diff
@@ -696,2444 +1483,7792 @@ toH = method()
...
+	       ))
      )
 
+-- Enumerate partitions mu, contained in lambda componentwise, with all columns
+-- of even length (equivalently, conjugate mu' has all even parts -- parts of mu
+-- come in equal pairs).
+evenColsSubpartitionsOf = method()
+evenColsSubpartitionsOf(BasicList) := (lambda) -> (
+     lam := toList lambda;
+     while #lam > 0 and lam#(-1) == 0 do lam = drop(lam, -1);
```
d-torrance: helper function again

M2/Macaulay2/packages/SchurRings.m2
```diff
@@ -696,2444 +1483,7792 @@ toH = method()
...
+--   Remove the hook of cell (k, 1) from partition lam.
+--   Returns (newLam, cR) where cR = # columns occupied by the strip.
+removeBorderStripAtFirstColumn = (lam, k) -> (
+     r := #lam;
+     cR := lam#(k-1);  -- c(R) = lam_k (# columns occupied by the strip)
+     newLam := join(
+	  take(lam, k-1),
+	  for i from k+1 to r list lam#(i-1) - 1
+	  );
+     while #newLam > 0 and newLam#(-1) == 0 do newLam = drop(newLam, -1);
```
d-torrance: helper function again

M2/Macaulay2/packages/SchurRings.m2
```diff
@@ -696,2444 +1483,7792 @@ toH = method()
...
+     toList conjugate new Partition from newLamT
+     )
 
-     schurRes(rep,M,new List from plets,DegreeLimit => d,SyzygyLimit => c)
+-- sswTypeC(lambda, n):
+--   SSW modification rule for Sp(2n).
+--   Returns (tau, sign) or null (meaning zero).
+sswTypeC = (lambda, n) -> (
+     lam := toList lambda;
+     while #lam > 0 and lam#(-1) == 0 do lam = drop(lam, -1);
```
d-torrance: helper function

M2/Macaulay2/packages/SchurRings.m2
```diff
@@ -696,2444 +1483,7792 @@ toH = method()
...
+--   SSW modification rule for O(m) (type B if m = 2n+1, type D if m = 2n).
+--   Returns (tau, sign) or null.  Here tau must satisfy the O-admissibility
+--   condition lam^T_1 + lam^T_2 <= m; the iteration stops whenever this
+--   holds.  Differences from type C (see [SSW 4.4]):
+--     (D1) strip length L = 2 * ell(lambda) - m,
+--     (D2) sign contribution uses c(R) - 1 instead of c(R),
+--     (D3) if an odd total number of strips was removed, apply sigma to tau.
+sswTypeBD = (lambda, m) -> (
+     lam := toList lambda;
+     while #lam > 0 and lam#(-1) == 0 do lam = drop(lam, -1);
```
d-torrance: helper function

M2/Macaulay2/packages/SchurRings.m2
```diff
@@ -696,2444 +1483,7792 @@ toH = method()
...
+
+-- Cache: partition lambda (list) -> list of (mu, integer) pairs representing
+-- sp_lambda = sum_mu (coef) s_mu in the stable universal character ring.
+stableSpInSchurCache := new MutableHashTable
+
+-- Return list of (mu, coef) with coef nonzero integer, such that
+--   sp_lambda  ==  sum_mu coef * s_mu.
+stableSpInSchur = (lambda) -> (
+     lam := toList lambda;
+     while #lam > 0 and lam#(-1) == 0 do lam = drop(lam, -1);
```
d-torrance: helper function

M2/Macaulay2/packages/SchurRings.m2
```diff
@@ -696,2444 +1483,7792 @@ toH = method()
...
+     stableSpInSchurCache#lam = ans;
+     ans
+     )
 
+-- Same, but for orthogonal characters (Littlewood uses nu with all even rows).
+stableOInSchurCache := new MutableHashTable
+
+stableOInSchur = (lambda) -> (
+     lam := toList lambda;
+     while #lam > 0 and lam#(-1) == 0 do lam = drop(lam, -1);
```
d-torrance: helper function (I'm going to stop making this comment now when I see future "strip trailing zeros" code, but I expect there are more!)

M2/Macaulay2/packages/SchurRings.m2
```diff
@@ -696,2444 +1483,7792 @@ toH = method()
...
+-- target ranks from the topmost layer inward; `infinity` (or any
+-- InfiniteNumber) leaves that layer untouched.  Example: for a ring S of
+-- schurLevel 2 whose outer layer is a stable Sp ring and inner layer is a
+-- stable GL ring, `specialize(f, {4, 3})` produces an element of Sp(8) over
+-- GL(3).  A shorter list only specializes the outermost layers.
+specialize(RingElement, List) := opts -> (f, ranks) -> (
+     R := ring f;
+     if #ranks == 0 then return f;
+     nTop := ranks#0;
+     g := if nTop === infinity or (class nTop === InfiniteNumber) then f
```
d-torrance: redundant `infinity` again

M2/Macaulay2/packages/SchurRings.m2
```diff
@@ -696,2444 +1483,7792 @@ toH = method()
...
+	  error "toRatGL: target ring must have GroupActing => \"RatGL\"";
+     R := ring f;
+     rawRes := raw(0_T);
+     if class R === SchurRing and R.?GroupActing and R.GroupActing == "RatGL" then (
+	  -- Already rational: re-embed (alpha, beta).  If target rank is finite
+	  -- and smaller than the bipartition's "length" ell(alpha)+ell(beta),
+	  -- apply the Koike-Terada modification.
+	  iterateRatGLTerms(f, (alpha, beta, scalar) -> (
+		    nT := numgens T;
+		    if nT === infinity or (class nT === InfiniteNumber) then (
```
d-torrance: redundant `infinity` again

d-torrance (changes requested):
I just had time for a quick scan of roughly the first 1/3 of the changes.  A broad comment: there's quite a bit of duplicate code that can probably be simplified by writing helper functions, etc.

M2/Macaulay2/packages/SchurRings.m2
mahrud: It didn't raise an error because `:=` assigns only locally. This is always possible (e.g. you can define `f = res -> 2*res` and this also works fine)

d-torrance (approved):
Thanks!

mahrud:
@kellerlv could you break up SchurRings.m2 into smaller files in an auxiliary directory, e.g. by splitting of the documentation and tests? At 10603 lines, this is now the longest package, which causes annoying delays on some filesystems, even when only calling `readPackage`.

mahrud:
Alternatively, do you mind if Doug or I do this, in a style matching your other packages?

kellerlv:
> Alternatively, do you mind if Doug or I do this, in a style matching your other packages?

That's fine with me. If you have problems let me know and I can get around to it (though probably not as quickly)

PR #4203, opened 2026-04-20
Remove "Li" synonym for polylog

d-torrance:
In response to https://github.com/Macaulay2/M2/pull/4046#issuecomment-4278364172


mahrud (approved):
Thanks!

PR #4204, opened 2026-04-20
Some updates to the Valuations package

d-torrance:
A few updates to *Valuations*:

* The target of `padicValuation` is now `ZZ` instead of `QQ`
* The source of `localRingValuation` is now the fraction field of the local ring instead of just the local ring itself.
* Allow `RingFamily` as a source for valuations.  This way, we can define valuations on `RR`, `CC`, etc.
* Refactor evaluation of valuations.  In particular, we don't actually do any promotion of the input to the source, but just check to see if it's possible as a type check. 
* Drop the redundant calls to `numerator` and `denominator` in `padicValuation` since that's already taken care of for all valuations.

Also a few related changes that popped up while I worked on this:
* We import *Classic* in *Binomials* since it calls `matrix "0"`, which will fail if *Classic* isn't loaded.
* Add promotion from local rings to fraction fields in *LocalRings*.

### DIFF
--- a/M2/Macaulay2/packages/LocalRings/localring.m2
+++ b/M2/Macaulay2/packages/LocalRings/localring.m2
@@ -67,6 +67,9 @@ localRing(EngineRing, Ideal) := (R, P) ->
         RP.localRing    = RP;
         RP.maxIdeal     =  P;
         commonEngineRingInitializations RP;
+	setupPromote(
+	    f -> numerator f / denominator f,
+	    RP, frac R);
 	RP.residueMap   = map(frac(R/P), RP, vars R % P);
          expression RP := r -> expression numerator r / expression denominator r;
            toString RP := r -> toString expression r;
--- a/M2/Macaulay2/packages/Binomials.m2
+++ b/M2/Macaulay2/packages/Binomials.m2
@@ -30,7 +30,7 @@ newPackage(
 		  HomePage => "http://www.thomas-kahle.de"}},
     	Headline => "specialized routines for binomial ideals",
 	Keywords => {"Commutative Algebra"},
-	PackageImports => {"FourTiTwo", "Cyclotomic", "LLLBases", "MinimalPrimes", "Elimination"},
+	PackageImports => {"FourTiTwo", "Cyclotomic", "LLLBases", "MinimalPrimes", "Elimination", "Classic"},
 	Certification => {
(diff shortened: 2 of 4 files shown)


MichaelABurr (approved):
This looks ok.  It took a minute to figure out what was happening with the localRingValuation code.  Someone from the LocalRings package should also take a look (it looked ok to me, but I'm not a maintainer of that code).

d-torrance:
@mahrud @mikestillman - Does https://github.com/Macaulay2/M2/pull/4204/changes/407a9f26ae9dc3384defff038077f95b8be3b26b look okay for promoting elements of local rings to fraction fields?

d-torrance:
Discussed with Mike.  TODO: lift from fraction field to local ring

M2/Macaulay2/packages/LocalRings/localring.m2
```diff
@@ -67,6 +67,9 @@ localRing(EngineRing, Ideal) := (R, P) ->
         RP.localRing    = RP;
         RP.maxIdeal     =  P;
         commonEngineRingInitializations RP;
+	setupPromote(
+	    f -> numerator f / denominator f,
+	    RP, frac R);
```
mahrud, line 72: The internal data is identical iirc, so why not just this?
```suggestion
	setupPromote(f -> new frac R from f, RP, frac R);
```

mahrud: I also want to say defining this as a function seems silly and should be done elsewhere.

d-torrance: For some reason, using `new` doesn't work right:

```m2
i2 : R = QQ[x];

i3 : p = ideal x;

o3 : Ideal of R

i4 : S = R_p;

i5 : F = frac R;

i6 : promote(S, F) := (f, K) -> new K from f;

i7 : promote(x_S, F)

o7 = x

o7 : F
```

Seems good so far.  But check this out:

```m2
i8 : oo == x_F

o8 = false
```

Very strange!

But the technique from this PR works as expected:  :confused:

```m2
i9 : promote(S, F) := (f, K) -> numerator f / denominator f;

i10 : promote(x_S, F)

o10 = x

o10 : F

i11 : oo == x_F

o11 = true
```

Issue #4210, opened 2026-04-21
Parallel assignment of local variables in packages can fail


d-torrance:
I just discovered a bug with the new parallel assignment feature.  If we try to assign variables alongside other things (i.e., not everything is a local variable) using parallel assignment *inside a package*, then we get a "mutable unexported symbol" error.

For example, in the package *Foo*, if we have:

```m2
f = () -> (
    (,x) := (1, 2);
    x)
```

then we get:

```m2
i1 : loadPackage "Foo"
stdio:1:11:(3):[1]: error: mutable unexported unset symbol(s) in package Foo: 'x'
Foo.m2:4:6-4:7: here is the first use of 'x'
```

d-torrance:
Fixed in #4215

PR #4215, opened 2026-04-24
Bind unary commas in parallel assignment

d-torrance:
Otherwise, we get "unset symbol" errors on things like `(,x) = (1,2)`

Closes: #4210


d-torrance:
Discussed with Mike

PR #4216, opened 2026-04-24
Bump FLINT to v3.5.0

d-torrance:
FLINT 3.5.0 was just released.  Draft for now to test the builds

### DIFF
--- a/M2/cmake/check-libraries.cmake
+++ b/M2/cmake/check-libraries.cmake
@@ -267,6 +267,7 @@ foreach(_library IN LISTS LIBRARY_OPTIONS)
       unset(${_name}_INCLUDEDIR CACHE)
       unset(${_name}_INCLUDE_DIR CACHE)
       unset(${_name}_INCLUDE_DIRS CACHE)
+      unset(${_name}_ROOT)
       # for GTest:
       unset(${_name}_MAIN_LIBRARY CACHE)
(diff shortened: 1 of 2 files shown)


d-torrance:
Hmm, getting a factory error now:

```m2
 -- capturing example results for "extendIdealByNonZeroMinor"               error: convertFacCF2Fq_nmod_t: element is not reduced
FLINTconvert.cc:463: In function `void convertFacCF2Fq_nmod_t(nmod_poly_struct*, const CanonicalForm&, const fq_nmod_ctx_struct*)':
failed assertion `i.exp() <= fq_nmod_ctx_degree(ctx)'
Aborted (core dumped)
```

d-torrance:
I think the problem is that M2 is linking against the FLINT 3.5.0 we just built but factory is linking against whatever old version of FLINT ships with Ubuntu 24.04.

There's a bug (which I hopefully just fixed...) in the cmake build that was causing this.  It wasn't a problem in the macOS build because FLINT 3.5.0 is already bottled in Homebrew, so we were just linking against two different copies of the same library.

d-torrance:
Yay, the cmake fix worked!  https://github.com/Macaulay2/M2/actions/runs/25007214596?pr=4216

d-torrance:
Discussed with Mike

Issue #4206, opened 2026-04-20
issue with norm


LukeOeding:
I'm having trouble getting the expected behavior from "norm".
Here's what I tried:
```
v = random(RR^2,RR^1)
norm(v)
norm_2 v
norm(2, v)
norm(2_RR, v)
norm(infinity, v)
```

Only the first one and last one execute, and both give the L^infinity norm. Of course I can implement my own 2-norm, but I would have expected this to work with any integer, certainly 2, or positive real number since the documentation says that I can do that.

d-torrance:
I'm not sure what we want when there are at least two rows and at least two columns.  For example, should `norm_2` give the Frobenius norm or the spectral norm?

LukeOeding:
I would guess that norm_2 would be the Frobenius norm, however wikipedia seems to disagree. Probably better to use the operator norm on matrices. If someone wants the vector norm on matrices, they can flatten first?

By the way, I forgot to put in the results I got:
```
Macaulay2, version 1.25.11
Type "help" to see useful commands

i1 : v = random(RR^2,RR^1)

o1 = | .159707 |
     | .11876  |

                2         1
o1 : Matrix RR    <-- RR
              53        53

i2 : norm(v)

o2 = .1597070215596925

o2 : RR (of precision 53)

i3 : norm_2 v
stdio:3:6:(3): error: no method found for applying norm to:
     argument 1 :  2 (of class ZZ)
     argument 2 :  | .159707 | (of class Matrix)
                   | .11876  |

i4 : norm(2, v)
stdio:4:4:(3): error: no method found for applying norm to:
     argument 1 :  2 (of class ZZ)
     argument 2 :  | .159707 | (of class Matrix)
                   | .11876  |

i5 : norm(2_RR, v)
stdio:5:4:(3): error: Lp norm only implemented for p = infinity

i6 : norm(infinity, v)

o6 = .1597070215596925

o6 : RR (of precision 53)

i7 : v = random(RR^2)

o7 = | .321267 |
     |  .59296 |

         2
o7 : RR
       53

i8 : norm(v)

o8 = .5929596345035341

o8 : RR (of precision 53)

i9 : norm_2 v
stdio:9:6:(3): error: no method found for applying norm to:
     argument 1 :  2 (of class ZZ)
                                             2
     argument 2 :  | .321267 | (of class RR   )
                   |  .59296 |             53

i10 : norm(2, v)
stdio:10:4:(3): error: no method found for applying norm to:
[...]

PR #4217, opened 2026-04-24
Add l^p norms for finite p

d-torrance:
For example:

```m2
i1 : norm vector {3, 4}

o1 = 4

i2 : norm_2 vector {3, 4}

o2 = 5

o2 : RR (of precision 53)

i3 : norm_1 vector {3, 4}

o3 = 7
```

Fixes: #4206


seangrate (approved):
Looks good to me. The extra documentation and tests are always appreciated.

PR #4220, opened 2026-04-24
Fixing a bug with join in SimplicialComplexes

sashahbc:
Nathan Ilten pointed out this bug to me. I'm not sure if this is meant to be intended behavior for self-tensor products, but this simple line of code fixes it in my situation.

Here's a simpler example illustrating the bug.
```
R = ZZ[x_1,x_2]
S = R
RxR = R ** S
phi1 = map(RxR,R)
phi2 = map(RxR,S)
```
Because R === S in memory, phi2 === phi1. But I would have expected phi1 to be the inclusion into the first factor and phi2 the inclusion into the second factor.


d-torrance (approved):
Thanks!

PR #4221, opened 2026-04-25
Support 2-argument "map" to/from RingFamily objects (RR/CC/RRi)

d-torrance:
### Before

```m2
i1 : f = map(CC, RR)
stdio:1:7:(3):[1]: error: no method found for applying map to:
     argument 1 :  CC (of class InexactFieldFamily)
     argument 2 :  RR (of class InexactFieldFamily)
```

### After

```m2
i1 : f = map(CC, RR)

o1 = map (CC  , RR  , {})
            53    53

o1 : RingMap CC   <-- RR
               53       53

i2 : f 2

o2 = 2

o2 : CC (of precision 53)
```


d-torrance:
Discussed with Mike

PR #4222, opened 2026-04-25
Add support for "kernel" for maps from ZZ

d-torrance:
This adds a simple strategy in the case where the source is ℤ -- the kernel is just the ideal generated by the characteristic of the target.

This fixes a 30-year old feature request!


d-torrance:
Discussed with Mike

PR #4223, opened 2026-04-25
Changed findProgram call for msolve to "msolve -h"

coolcuber:
`msolve` didn't support a `--help` option until [`v0.9.5`](https://github.com/algebraic-solving/msolve/blob/v0.9.5/src/msolve/main.c#L73).  As of `v0.9.2`, there is no error code thrown, but `msolve` does complain  to STDOUT when you call it with `--help`.  Since the minimum `msolve` version required by Macaulay2 is `v0.7.0` and `-h` works in `v0.9.5`, it might make more sense to use `-h` for now.


d-torrance (approved):
First contribution!!  :tada: 🥳 

Thank you!

Issue #4161, opened 2026-03-14
random vs shuffle for a list


mahrud:
I regularly mistake `random(List)` to mean a random element of the list, until I remember it's actually shuffling the list.

It would be a breaking change, but I wish we had `shuffle(List)` do this job and `random(List)` just give a random element of a list.

PR #4224, opened 2026-04-26
Rename random(List) to shuffle(List) so that random(List) returns a single element

d-torrance:
This addresses #4161.

I've also found the current behavior confusing.  Broadly speaking, the other `random` methods take a "set" and return a random element of that set (whether that set is a ring, a number representing an upper bound of an interval, or what have you -- it's the same basic idea).  But `random(List)` doesn't do that. 

So we change its behavior so that it does, and introduce `shuffle(List)` to take on the previous behavior.

```m2
i1 : x = toList(1..20)

o1 = {1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20}

o1 : List

i2 : random x

o2 = 6

i3 : shuffle x

o3 = {11, 10, 18, 12, 16, 20, 7, 15, 6, 14, 4, 9, 17, 8, 13, 1, 2, 19, 3, 5}

o3 : List
```

This was a pretty big change and I had to fix a bunch of packages that this change broke.  Maybe we can discuss at the next M2internals for a consensus on whether it would be worth merging?


M2/Macaulay2/packages/PathSignatures/algebra.m2
mahrud, line 1: Probably need to remove shuffle from export list of this PathSignatures (and perhaps others)

M2/Macaulay2/packages/SchurComplexes.m2
```diff
@@ -312,7 +312,7 @@ permutedTableau= (T,X,row1, row2, col1,lengthcol1,L)-> (
 
 -- Inputs: T=Tableau, vio = pair (column, row) where violation occurs, downCol2 is the number of times the entry in the second column repeats. 
 ---Outputs: Linear combination represented as a hashtable with tableau keys.
-shuffle = (T, vio, downCol2, lambdaprime) -> (
+shuffle(HashTable, Sequence, ZZ, Partition) := (T, vio, downCol2, lambdaprime) -> (
```
mahrud: Since these aren't exported, switching them to methods causes the documentation to list them as undocumented. Might be better to just rename to `shuffle'` or similar.

M2/Macaulay2/packages/SchurFunctors.m2
```diff
@@ -130,7 +130,7 @@ exchange(Filling, ZZ, ZZ, List) := (T, col1, col2, s) -> (
 	  ), sgn*sgnb)
      )
 
-shuffle = (T, nrows, col1, col2) -> (
+shuffle(Filling, ZZ, ZZ, ZZ) := (T, nrows, col1, col2) -> (
```
mahrud: Same here.

M2/Macaulay2/packages/RationalPoints2.m2
```diff
@@ -73,7 +73,7 @@ pow = (n, lst) -> (
 -- Utility function that shuffles two list u and v according to a list ind
 -- of indices in 0 (for u) and 1 (for v)
 --
-shuffle = (ind, u, v) -> (
+shuffle(List, List, List) := (ind, u, v) -> (
```
mahrud: Same here.

M2/Macaulay2/packages/SchurComplexes.m2
d-torrance: Ah good call

M2/Macaulay2/packages/PieriMaps.m2
```diff
@@ -90,7 +90,7 @@ isStandard = T -> (
 -- entries (ignoring that some entries might be equal). The output is given in 
 -- the form a hash table where the keys are the resulting tableau and the 
 -- values are -1
-shuffle = (T, col, row1, row2) -> (
+shuffle(List, ZZ, ZZ, ZZ) := (T, col, row1, row2) -> (
```
mahrud: Same here.

M2/Macaulay2/packages/PathSignatures/algebra.m2
mahrud: Ah, you've done this already. The main file was listed further down.

M2/Macaulay2/packages/GKMVarieties/Tests_GKMVarieties.m2
```diff
@@ -8,7 +8,7 @@ Tests for GKMVarieties.m2
 -- flagGeomTuttePolynomial test
 --------------------------------
 TEST ///
-ML = (random drop(drop(allMatroids 4,1),-1))_{0,1}
+ML = (shuffle drop(drop(allMatroids 4,1),-1))_{0,1}
```
mahrud: Wondering if it would be worth to switch to `randomSubset` for examples like this, or further down for `take(shuffle .., n)`.

d-torrance: Yeah I did that for a few cases where they also called `sort`. But `randomSubset` preserves order, so its not quite the same. May not matter for the particular algorithms but I didnt feel like making that call

M2/Macaulay2/packages/CodingTheory.m2
```diff
@@ -1790,7 +1790,7 @@ C := linearCode G;
 for i from 1 to 1 do(
     message := transpose matrix {(for n from 1 to numgens target G list(random(R)))};
     codeword := (transpose G)*message;
-    errors := sum take(random entries basis target codeword, 3);
+    errors := sum take(shuffle entries basis target codeword, 3);
```
mahrud: Alternative suggestion: adding `random(ZZ, List)` to take 3 random (unordered) elements from a list. (I don't understand why it's not `randomSubset(ZZ, List)`; typically simpler arguments should come first)

d-torrance: Yeah, I remember going back and forth on what to do with `randomSubset`.  I think I chose `(List, ZZ)` for consistency with `subsets`.

d-torrance: I'm not sure about the `random(ZZ, List)` idea -- the other `random(ZZ, ...)` methods all mean "return a random object with this degree", not "return this many random objects".

d-torrance: Maybe `shuffle(ZZ, List)`?

d-torrance: I went ahead and added `shuffle(List, ZZ)` instead, again for consistency with `randomSubset`, `subsets`, `take`, etc.  It's just `shuffle @@ randomSubset`.

M2/Macaulay2/m2/lists.m2
```diff
@@ -256,6 +248,15 @@ randomSubset VisibleList := x -> x_(randomSubset(#x))
 randomSubset(Set, ZZ) := (x, n) -> set randomSubset(toList x, n)
 randomSubset Set := x -> set randomSubset toList x
 
+shuffle = method()
+shuffle MutableList := s -> (
```
mahrud, line 266: Appreciate in-place functionality for mutable objects.

d-torrance:
I've implemented the suggestions from this afternoon's M2internals:

```m2
i1 : x = toList(0..12)

o1 = {0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12}

o1 : List

i2 : random x
 -- the behavior of random(List) will change soon; use shuffle(List) instead

o2 = {0, 7, 9, 4, 5, 3, 8, 12, 10, 11, 6, 2, 1}

o2 : List

i3 : random x

o3 = {5, 9, 1, 7, 12, 8, 10, 0, 6, 3, 4, 11, 2}

o3 : List

i4 : randomElement x

o4 = 0

i5 : shuffle x

o5 = {1, 4, 0, 9, 8, 3, 2, 5, 7, 10, 11, 6, 12}

o5 : List
```

M2/Macaulay2/m2/lists.m2
```diff
@@ -228,15 +228,21 @@ flexibleOperators = sort flexibleOperators
...
-     s = new MutableList from s;
-     for i from 1 to n-1 do (
-	  j := random (i+1);
-	  t := s#i ; s#i = s#j ; s#j = t;
-	  );
-     new List from s)
+randomElement = method()
+randomElement List := x -> (
+    if #x == 0 then error "expected a nonempty list";
+    x#(random(#x)))
```
mahrud, line 234: Not important for this PR, but related to #4230, I've noticed that something like the following is sometimes a little faster. I'd like to understand which interpreter optimizations come into play here better.
```suggestion
randomElement List := x -> if #x > 0 then x#(random(#x)) else error "expected a nonempty list"
```
(I forget, it's possible the then and else clauses need to be swapped for best performance)

d-torrance:
Discussed with Mike

mahrud:
Did you check if there are any `the behavior of...` strings printed in the package examples?

d-torrance:
There shouldn't be.  In earlier commits prior to the M2internals discussion, I'd changed the behavior of `random(List)` already, so everything that called it has been changed to `shuffle` or `randomSubset` so the builds would pass.

PR #4225, opened 2026-04-27
Change /bin/foo program calls

coolcuber:
While `/bin/echo` accommodates OSX users and their abnormal `echo` situation, it can break in sandboxed builds.  I reason `env echo` should make everyone happy.  The same applies for `/bin/pwd` (I'm not sure if this was the original concern, but `env pwd` still avoids accidentally using a shell builtin).  See [this PR](https://github.com/Macaulay2/frobby/pull/34#issue-4332074986) on `frobby` for an elaboration.


d-torrance (approved):
Thanks!

PR #4234, opened 2026-04-29
Use etags to generate m2 directory TAGS file (autotools)

### LINKED ISSUE #4233: Argument list too long making TAGS file
The Ubuntu autotools builds have started failing with the following, which I've been able to reproduce locally:

```sh
$ make -C Macaulay2/m2 TAGS
make: Entering directory '/home/dtorrance/src/macaulay2/M2/M2/BUILD/build/Macaulay2/m2'
: making ../../../../Macaulay2/m2/TAGS
make: /bin/sh: Argument list too long
make: *** [Makefile:60: ../../../../Macaulay2/m2/TAGS] Error 127
make: Leaving directory '/home/dtorrance/src/macaulay2/M2/M2/BUILD/build/Macaulay2/m2'
```
I think maybe we've hit the critical mass of numbers of packages where we can't list them all out at once on the command line on some systems...

d-torrance:
~~It's as pretty pointless file, as no functions are defined, and beginning in April 2026, we started having so many package files that the build failed with "Argument list too long" errors~~

I've replaced this with another solution -- generate the `TAGS` file the same way we do in cmake, by running `etags` on the Core .m2 files.  Not only should this fix the failing autotools builds, but it also gives us a useful `M-.` in the M2 directory!

Should fix #4233

### DIFF
--- a/M2/Macaulay2/m2/Makefile.in
+++ b/M2/Macaulay2/m2/Makefile.in
@@ -36,31 +36,9 @@ clean::;rm -f @pre_packagesdir@/Core/tvalues.m2
 all: tags
 tags: Makefile @srcdir@/TAGS @srcdir@/TAGS.doc
 clean::; rm -f TAGS @srcdir@/TAGS @srcdir@/TAGS.doc
-@srcdir@/TAGS: \
-		$(DUMPEDM2SRCFILES) \
-		$(shell @FIND@ @srcdir@/../packages -name "*.m2" -a -not -name .\#\* | sort) \
-		../editors/emacs/make-M2-emacs-help.m2 \
-		../editors/make-M2-symbols.m2 \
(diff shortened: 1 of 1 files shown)


mahrud:
> It's as pretty pointless file, as no functions are defined

Uhh it's almost 3700 lines for me and sometimes useful!
I don't mind you changing autotools to not generate it though.

d-torrance:
I see the point of the TAGS file in the d and e directories since they keep track of function definitions, but the m2 one just has filenames.  What's it useful for?

d-torrance:
Ah -- just looked the cmake recipe -- it's completely different!  It actually runs `etags` on the Core files instead of just throwing in a bunch of filenames.  I'll do that instead

d-torrance:
I'm going to go ahead and merge this so the autotools-ubuntu builds stop failing

PR #4238, opened 2026-04-30
Bump tbb to v2023.0.0

d-torrance:
Draft for now to test the builds

### DIFF
--- a/M2/libraries/tbb/Makefile.in
+++ b/M2/libraries/tbb/Makefile.in
@@ -1,6 +1,6 @@
 HOMEPAGE = https://oneapi-src.github.io/oneTBB/
-VERSION = 2022.3.0
-SHA256SUM = 01598a46c1162c27253a0de0236f520fd8ee8166e9ebb84a4243574f88e6e50a
+VERSION = 2023.0.0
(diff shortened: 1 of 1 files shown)


mahrud:
I am _very_ hesitant about touching the tbb stuff if it involves any change to the code, since it's so annoying to get it on servers.

d-torrance:
So far, it's looking like we're in the clear, as the builds are all chugging along without issue!

And I'm totally on board with your point.  If we do need to make any changes to the code for TBB (or any other library for that matter), then I strongly think that we should make those changes backwards compatible to support older versions.

mahrud:
I guess I'm wondering if we should even keep using the 2021 version (which is widely available as far as I can tell) so that new code is forced to remain compatible with it, until there are specific new tbb features that are desired, then we can upgrade with backwards compatibility (that is, avoid unintentionally using features incompatible with tbb 2021)

d-torrance:
Right now we're even still compatible with tbb 2020.  IIRC 2020 -> 2021 was the version bump with the most breaking changes.  The interface has been pretty stable since then.

d-torrance:
Builds passed!  https://github.com/Macaulay2/M2/actions/runs/25173391336

d-torrance:
Discussed with Mike

Issue #4164, opened 2026-03-17
Debug build fails


mahrud:
After the standalone engine PR I can't compile in debug build because of the following error:
```
In file included from /home/mahrud/Projects/M2/research/M2/Macaulay2/d/M2mem.c:5:
/home/mahrud/Projects/M2/research/M2/Macaulay2/d/debug.h:18:40: error: redefinition of ‘debug_version’
   18 |   static __attribute__ ((unused)) void debug_version() {}
      |                                        ^~~~~~~~~~~~~
compilation terminated due to -Wfatal-errors.
```
the solution might be:
```patch
diff --git a/M2/Macaulay2/e/interface/m2-mem.h b/M2/Macaulay2/e/interface/m2-mem.h
index b46ed5242a..d9825a1a4d 100644
--- a/M2/Macaulay2/e/interface/m2-mem.h
+++ b/M2/Macaulay2/e/interface/m2-mem.h
@@ -4,6 +4,9 @@
 #include <stdlib.h>

 // d/debug.h
+#ifndef _DEBUG_H_
+#define _DEBUG_H_ 1
+
 #ifndef NDEBUG

 #if defined(__cplusplus)
@@ -32,10 +35,12 @@ extern "C" {
 #define TRAPCHK(p)
 #define TRAPCHK_SIZE(n)

+#endif
 #endif


```

But this still fails
```cpp
/home/linuxbrew/.linuxbrew/bin/ld: Macaulay2/e/libM2-engine.a(m2-mem.cpp.o):/home/mahrud/Projects/M2/research/M2/Macaulay2/e/interface/m2-mem.cpp:32: multiple definition of `trapaddr'; Macaulay2/d/CMakeFiles/M2-interpreter.dir/debug.c.o:/home/mahrud/Projects/M2/research/M2/Macaulay2/d/debug.c:16: first defined here
/home/linuxbrew/.linuxbrew/bin/ld: Macaulay2/e/libM2-engine.a(m2-mem.cpp.o):/home/mahrud/Projects/M2/research/M2/Macaulay2/e/interface/m2-mem.cpp:33: multiple definition of `trapcount'; Macaulay2/d/CMakeFiles/M2-interpreter.dir/debug.c.o:/home/mahrud/Projects/M2/research/M2/Macaulay2/d/debug.c:17: first defined here
[...]

Issue #3938, opened 2025-07-22
fatalarrayindex in ctype_isvalidsymbol


d-torrance:
On autotools build, development branch, I get the following error in debug mode, which looks related to this.
```
/opt/homebrew/bin/ginstall -c M2 /Users/mike/src/M2-current-branches/M2-development/M2/BUILD/mike/builds.tmp/debug-arm64-appleclang/usr-dist/aarch64-Darwin-MacOS-15.5/bin/M2
: : : : : : : : : : : : : : : : : : : : : : : : : : : : : : : :
: Macaulay2 executable location:
:    "/Users/mike/src/M2-current-branches/M2-development/M2/BUILD/mike/builds.tmp/debug-arm64-appleclang/usr-dist/aarch64-Darwin-MacOS-15.5/bin/M2-binary"
: : : : : : : : : : : : : : : : : : : : : : : : : : : : : : : :
gmake[2]: Leaving directory '/Users/mike/src/M2-current-branches/M2-development/M2/BUILD/mike/builds.tmp/debug-arm64-appleclang/Macaulay2/bin'
gmake -C m2 all 
gmake[2]: Entering directory '/Users/mike/src/M2-current-branches/M2-development/M2/BUILD/mike/builds.tmp/debug-arm64-appleclang/Macaulay2/m2'
echo '# -*- sh -*-' >.gdb-directories
/bin/echo 'echo -- loading .gdb-directories\n' >>.gdb-directories
for i in \
	../../libraries/*/build/*/. \
	../../libraries/ntl/build/*/src/. \
	../../libraries/libfac/build/*/factor/. \
	../../libraries/libfac/build/*/charset/. \
	;\
do if [ -d $i ] ;\
   then echo directory $i >>.gdb-directories ;\
   fi ;\
done
/bin/echo 'echo -- loaded .gdb-directories\n' >>.gdb-directories
cp ../../../../../../Macaulay2/m2/.gdbinit.mike .gdbinit
echo '# -*- sh -*-' >.gdb-files
/bin/echo 'echo -- loading .gdb-files\n' >>.gdb-files
echo 'set environment LD_LIBRARY_PATH /Users/mike/src/M2-current-branches/M2-development/M2/BUILD/mike/builds.tmp/debug-arm64-appleclang/usr-host/lib:/Users/mike/src/M2-current-branches/M2-development/M2/BUILD/mike/builds.tmp/cmake-appleclang/installed/lib' >>.gdb-files
[...]

Issue #1539, opened 2020-10-09
Out of memory running SpecialFanoFourfolds example on i386


d-torrance:
On a build in an i386 chroot, I get the following when running the `detectCongruence` example:

```M2
i7 : -- A general GM fourfold of discriminant 20
     X = specialGushelMukaiFourfold("surface of degree 9 and genus 2",ZZ/33331);

o7 : SpecialGushelMukaiFourfold (Gushel-Mukai fourfold containing a surface of degree 9 and sectional genus 2)

i8 : describe X

o8 = Special Gushel-Mukai fourfold of discriminant 20
     containing a surface in PP^8 of degree 9 and sectional genus 2
     cut out by 19 hypersurfaces of degree 2
     and with class in G(1,4) given by 6*s_(3,1)+3*s_(2,2)
     Type: ordinary
     (case 17 of Table 1 in arXiv:2002.07026)

i9 : time f = detectCongruence X;


 *** out of memory trying to allocate 131108 bytes, exiting ***
```

giovannistagliano:
Maybe, in this package it's better to enable the options:  
`AuxiliaryFiles=>true, 
CacheExampleOutput => true` 
Let me know if I need to make a pull request with the changes.

DanGrayson:
No, we like running examples to make sure everything is working.  Let's just make the example less demanding.

giovannistagliano:
OK, I'll think about how to simplify the example. However, on my laptop that code is not so hard. It takes less than 10 seconds and uses about 200 MB of memory.

DanGrayson:
Is your laptop 64 bit or 32 bit?

DanGrayson:
(Maybe it does point to a real problem elsewhere in the system.)

giovannistagliano:
> Is your laptop 64 bit or 32 bit?

It's 64 bit with Ubuntu 20.04 and M2 1.16.

DanGrayson:
Maybe there's some sort of problem that causes it to fail on a 32 bit machine.  It's hard to imagine what it might be, though.  I'll set up a 32-bit debian vm and we'll see.

DanGrayson:
Okay, I've reproduced the problem here.

d-torrance:
As with #1429, I'm fixing this temporarily in the Debian package by using a canned example:
https://salsa.debian.org/science-team/macaulay2/-/blob/master/debian/patches/skip-special-fano-fourfolds-example.patch

Would this be worth submitting upstream?

[...]

Issue #1064, opened 2020-01-08
MinimalPrimes crashes occasionally


DanGrayson:
About once per month MinimalPrimes crashes M2 during a test, as it tries the SplitTower method:

<img width="788" alt="Screen Shot 2020-01-08 at 5 18 24 PM" src="https://user-images.githubusercontent.com/700228/72024454-18dfc980-323b-11ea-8987-344ea1963f2b.png">

DanGrayson:
That's under Ubuntu 16.04 on habanero.

fhinkel:
Unassigning myself. Realistically, I don't have time to helpf with this issue. Sorry 😞

d-torrance:
Just saw this for `check(13, "MinimalPrimes")` in an [armhf PPA build on Ubuntu 20.04](https://launchpadlibrarian.net/527821423/buildlog_ubuntu-focal-armhf.macaulay2_1.17.2.1+git202103122347-0ppa202103131942~ubuntu20.04.1_BUILDING.txt.gz):

```m2
 -- capturing check(13, "MinimalPrimes")                                    -- SIGSEGV
-* stack trace, pid: 20485
 0# stack_trace(std::ostream&, bool) at ./M2/Macaulay2/d/main.cpp:124
 1# segv_handler at ./M2/Macaulay2/d/main.cpp:240
 2# 0xF6A57900 in /lib/arm-linux-gnueabihf/libc.so.6
 3# FractionField::mult(ring_elem, ring_elem) const at ./M2/Macaulay2/e/frac.cpp:516
 4# GBRingPoly::mult_by_term1(FreeModule const*, gbvector const*, ring_elem, int const*, int) at ./M2/Macaulay2/e/gbring.cpp:116
 5# GBRing::mult_by_term(FreeModule const*, gbvector const*, ring_elem, int const*, int) at ./M2/Macaulay2/e/gbring.cpp:839
 6# GBRing::gbvector_reduce_lead_term(FreeModule const*, FreeModule const*, gbvector*, gbvector*&, gbvector*&, gbvector const*, gbvector const*, bool, ring_elem&) at ./M2/Macaulay2/e/gbring.cpp:967
 7# ReducedGB_Field::remainder(POLY&, bool, ring_elem&) at ./M2/Macaulay2/e/reducedgb-field.cpp:124
[...]

PR #4241, opened 2026-05-01
Use engine for all debug build code

d-torrance:
This way we have it for the standalone engine but don't get multiple definition errors when we link with the interpreter.

We also fix a couple out of bounds errors in the interpreter that fired after I got the debug build working. (#3938)

Closes: #4164

Draft for now to test the GitHub debug builds.


d-torrance:
Even though only 25% of the builds passed, I think this is in good shape as we got a functioning debug version of M2 in each case, and each of them chugged along for over 2 hours before running into a known memory issue (#1064, #1539, #2162): https://github.com/Macaulay2/M2/actions/runs/25199464796

mahrud:
I'm good as long as the binary actually compiled.

d-torrance:
Discussed with Mike

PR #4243, opened 2026-05-01
Minor updates to LieAlgebraRepresentations. 

dswinarski:
1. Added function to compute dual of a LieAlgebraRepresentation. 
2. Fixed bug in spinRepresentationMatrices and halfspinRepresentationMatrices (the basis of the Cartan subalgebra did not match that produced by lieAlgebraBasis). 
3. Updated documentation for gtInvariantInVtensorVdual to note that a conjectural formula is now proven..


M2/Macaulay2/packages/LieAlgebraRepresentations/test dual.m2
```diff
@@ -0,0 +1,6 @@
+	    sl6 = simpleLieAlgebra("A",5)
```
d-torrance: Would you rename this file to `test-dual.m2` or something without a space in it?  The system thinks there are two files here, `M2/Macaulay2/packages/LieAlgebraRepresentations/test` and  `dual.m2`.

d-torrance (changes requested):
Thanks for the contribution!  Only one comment.

dswinarski:
Thanks! I meant to delete the file 'test dual.m2' altogether.

M2/Macaulay2/packages/LieAlgebraRepresentations/test dual.m2
mahrud: This should be considered a bug, no? I agree spaces in filenames are strange but maybe that should be an error not a crash.

d-torrance: Yeah, I agree. It works fine with the cmake build already

d-torrance (approved):
Thanks!

PR #4244, opened 2026-05-01
Return the minimal primes of the zero ideal

### LINKED ISSUE #3505: isPrime in ZZ/n
`isPrime` doesn't work for ideals in `ZZ/n`, and we get different errors depending on whether `n` is prime or not:

```m2
i1 : isPrime ideal 0_(ZZ/2)
stdio:1:7:(3): error: no method found for applying presentation to:
     argument   :  ZZ (of class Ring)

i2 : isPrime ideal 0_(ZZ/4)
stdio:2:7:(3): error: no applicable strategy for (minimalPrimes,Ideal)
```

*Related:*  Should we check that a ring is an integral domain before creating a fraction field?   We'll likely get an error eventually, but it might be good to check up front:

```m2
i1 : R = ZZ/4

o1 = R

o1 : QuotientRing

i2 : 1/2_R

     -1
o2 = --
      2

o2 : frac R

i3 : oo^2
stdio:3:2:(3): error: a non unit was found in a ring declared to be a field

```
It is already mentioned that the ring should be an integral domain in the "Caveat" section in the [documentation for `frac`](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_frac.html), so maybe that's enough to avoid the extra overhead of checking?   Maybe we should have an `isWellDefined(FractionField)` that checks this?

d-torrance:
Previously, we only checked if the output of flattenRing was the zero ideal, but that may not be the case, e.g., for (0) in ZZ/n, this gives us (n).

This fixes #3505.

### Before
```m2
i1 : minimalPrimes ideal 0_(ZZ/2)
stdio:1:13:(3):[1]: error: no method found for applying presentation to:
     argument   :  ZZ (of class Ring)

i2 : minimalPrimes ideal 0_(ZZ/4)
stdio:2:13:(3):[1]: error: no applicable strategy for (minimalPrimes,Ideal)
```

### After
```m2
i1 : minimalPrimes ideal 0_(ZZ/2)

o1 = {ideal 0}

o1 : List

i2 : minimalPrimes ideal 0_(ZZ/4)

o2 = {ideal 0}

o2 : List
```

### DIFF
--- a/M2/Macaulay2/packages/MinimalPrimes.m2
+++ b/M2/Macaulay2/packages/MinimalPrimes.m2
@@ -218,8 +218,7 @@ updateComputation(MinimalPrimesComputation, List) := List => options minimalPrim
 -- Helper for minimalPrimes and decompose
 minprimesHelper = (I, key, opts) -> (
     if I == 1 then return {};
-    J := first flattenRing I;
-    if J == 0 then return {I};
+    if I == 0 or (J := first flattenRing I) == 0 then return {I};
     -- TODO: make presentation work for ZZ, then move this line
     if ring I === ZZ then return ideal \ first \ toList factor (trim I)_0;
     S := ring presentation ring J;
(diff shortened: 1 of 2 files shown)


d-torrance:
Never mind -- this was a dumb idea.  Lots of rings aren't integral domains!

PR #4246, opened 2026-05-02
PieriMaps: 2.0 overhaul

kellerlv:
Adds projection-direction maps (dualPieri, dualPieriColumn, dualLR),
point evaluators (applyPieri, applyPieriColumn, applyLR, applyDualPieri,
applyDualLR, applyDualPieriColumn), three basis conventions
(Row / Filling / Weyl) with conversion utilities, GL_n equivariance
and well-definedness checkers (verifyEquivariant, verifyWellDefined),
and symbolic display of any constructed map (symbolicForm).

### DIFF
--- a/M2/Macaulay2/packages/PieriMaps.m2
+++ b/M2/Macaulay2/packages/PieriMaps.m2
@@ -313,26 +313,25 @@ schurRank(ZZ, List) := (n, mu) -> (
 standardTableauxCache = new MutableHashTable
 
 standardTableaux = method()
-standardTableaux(ZZ, List) := (dim, mu) -> (
-     key := (dim, mu);
-     if standardTableauxCache#?key then return standardTableauxCache#key;
-     if #mu == 0 then (standardTableauxCache#key = {{}}; return {{}});
-     output := {};
-     otherrows := standardTableaux(dim, drop(mu, 1));
-     firstrow := rsort compositions(dim, mu#0);
-     for i in firstrow do (
-     	  temp := {};
-     	  for j from 0 to #i-1 do
-     	  for k from 1 to i#j when true do
-     	  temp = append(temp,j);
-     	  for j in otherrows do (
-	       temp2 := prepend(temp,j);
-     	       if isStandardPM(temp2) === null then
-     	       output = append(output, temp2);
+standardTableaux(ZZ, List) := (dim, mu) -> standardTableauxCache#(dim, mu) ??= (
+     if #mu == 0 then {{}}
+     else (
+	  output := {};
+	  otherrows := standardTableaux(dim, drop(mu, 1));
+	  firstrow := rsort compositions(dim, mu#0);
+	  for i in firstrow do (
+	       temp := {};
+	       for j from 0 to #i-1 do
+	       for k from 1 to i#j when true do
+	       temp = append(temp,j);
+	       for j in otherrows do (
+		    temp2 := prepend(temp,j);
+		    if isStandardPM(temp2) === null then
+		    output = append(output, temp2);
+		    );
 	       );
-     	  );
-     standardTableauxCache#key = output;
-     output
+	  output
+	  )
      )
 
 -- Input:
@@ -428,8 +427,7 @@ pmToFillingExpansion = T -> (
           if degenerate then return;
           F := new sfFillingType from sortedCols;
           contribution := (totalSign * 1_QQ) / denom;
-          if accum#?F then accum#F = accum#F + contribution
-          else accum#F = contribution;
+          accum#F = (accum#F ?? 0) + contribution;
           ));
      new HashTable from for k in keys accum list if accum#k != 0 then k => accum#k else continue
      )
@@ -447,7 +445,7 @@ pmToFilling = T -> (
 	  decomp := sfStraightenPM F;
 	  for k in keys decomp do (
 	       v := c * decomp#k;
-	       if out#?k then out#k = out#k + v else out#k = v;
+	       out#k = (out#k ?? 0) + v;
 	       );
 	  );
      new HashTable from for k in keys out list if out#k != 0 then k => out#k else continue
@@ -485,8 +483,7 @@ fillingToPMExpansion = F -> (
                for j from 0 to lambda#i - 1 list (orderedCols#j)#i;
           sortedRows := apply(rows, sort);
           contribution := (totalSign * 1_QQ) / denom;
-          if accum#?sortedRows then accum#sortedRows = accum#sortedRows + contribution
-          else accum#sortedRows = contribution;
+          accum#sortedRows = (accum#sortedRows ?? 0) + contribution;
           ));
      new HashTable from for k in keys accum list if accum#k != 0 then k => accum#k else continue
      )
@@ -502,7 +499,7 @@ fillingToPM = F -> (
           decomp := memo#(apply(K, sort));
           for std in keys decomp do (
                cc := c * decomp#std;
-               if out#?std then out#std = out#std + cc else out#std = cc;
+               out#std = (out#std ?? 0) + cc;
                );
           );
      new HashTable from for k in keys out list if out#k != 0 then k => out#k else continue
@@ -528,10 +525,7 @@ fillingToWeyl = F -> (
 fillingToPMMatrixCache = new MutableHashTable
 pmToFillingMatrixCache = new MutableHashTable
 
-fillingToPMMatrix = (mu, n) -> (
-     mm := trimPartPM mu;
-     key := (mm, n);
-     if fillingToPMMatrixCache#?key then return fillingToPMMatrixCache#key;
+fillingToPMMatrix = (mu, n) -> fillingToPMMatrixCache#(mm := trimPartPM mu, n) ??= (
      pmS := standardTableaux(n, mm);
      muTcols := conjPartPM mm;
      ensureSchurFn();
@@ -544,19 +538,15 @@ fillingToPMMatrix = (mu, n) -> (
           for sj from 0 to #sfS - 1 list (
                decomp := decomps#sj;
                U := pmS#ti;
-               if decomp#?U then decomp#U else 0_QQ
+               decomp#U ?? 0_QQ
                )
           );
      -- Source labels = standard Fillings; target labels = standard PM tableaux.
      attachBasisLabels(result, sfS, pmS);
-     fillingToPMMatrixCache#key = result;
      result
      )
 
-pmToFillingMatrix = (mu, n) -> (
-     mm := trimPartPM mu;
-     key := (mm, n);
-     if pmToFillingMatrixCache#?key then return pmToFillingMatrixCache#key;
+pmToFillingMatrix = (mu, n) -> pmToFillingMatrixCache#(mm := trimPartPM mu, n) ??= (
      pmS := standardTableaux(n, mm);
      muTcols := conjPartPM mm;
      ensureSchurFn();
@@ -566,12 +556,11 @@ pmToFillingMatrix = (mu, n) -> (
           for pj from 0 to #pmS - 1 list (
                decomp := decomps#pj;
                F := sfS#fi;
-               if decomp#?F then decomp#F else 0_QQ
+               decomp#F ?? 0_QQ
                )
           );
      -- Source labels = standard PM tableaux; target labels = standard Fillings.
      attachBasisLabels(result, pmS, sfS);
-     pmToFillingMatrixCache#key = result;
      result
      )
 
@@ -634,10 +623,9 @@ pieriHelper(List, ZZ, PolynomialRing) := (mu, k, P) -> (
 	       remove(H, T);
 	       straighten(U, memo);
 	       for i in keys memo#U do
-	       if H #? i then H#i = H#i + coeff * (memo#U)#i * X_((T_0)_0)
-	       else H#i = coeff * (memo#U)#i * X_((T_0)_0);
+	       H#i = (H#i ?? 0) + coeff * (memo#U)#i * X_((T_0)_0);
 	       );
-     	  for t in Tbasis do if H #? t then row = append(row, H#t) else row = append(row, 0);
+     	  for t in Tbasis do row = append(row, H#t ?? 0);
 	  output = append(output, row);
 	  );
      return map(P^(#Tbasis), P^{#Sbasis:-1}, transpose output);
@@ -750,14 +738,14 @@ pieriColumnHelper = (mu, k, P) -> (
 			      U#(J#i) = prepend(dropped, U#(J#i));
 			      sgn := if b % 2 == 0 then 1 else -1;
 			      key := new List from U;
-			      cur := if temp#?key then temp#key else 0;
+			      cur := temp#key ?? 0;
 			      temp#key = cur + sgn * coeff;
 			      );
 			 );
 		    h = temp;
 		    );
 	       for K in keys h do (
-		    cur := if H#?K then H#K else 0;
+		    cur := H#K ?? 0;
 		    H#K = cur + h#K;
 		    );
 	       );
@@ -771,10 +759,9 @@ pieriColumnHelper = (mu, k, P) -> (
 	       while #U > 0 and #(U#(#U - 1)) == 0 do U = drop(U, -1);
 	       UF := new sfFillingType from U;
 	       UFkey := toList U;
-	       if not memo#?UFkey then memo#UFkey = sfStraightenPM UF;
-	       decomp := memo#UFkey;
+	       decomp := memo#UFkey ??= sfStraightenPM UF;
 	       for std in keys decomp do (
-		    idx := if Tidx#?(toList std) then Tidx#(toList std) else null;
+		    idx := Tidx#(toList std) ?? null;
 		    if idx === null then continue;
 		    rowVec#idx = rowVec#idx + coeff * (decomp#std) * X#poppedEntry;
 		    );
@@ -1038,7 +1025,7 @@ lrMap(Sequence, List, ZZ) := opts -> (shapes, Q, n) -> (
 			 flatten for k from 0 to n - 1 list
 			      toList(expv#(a*n + k) : k);
 		    straighten(tvec, straightenH);
-		    decomp := if straightenH#?tvec then straightenH#tvec else new HashTable from {};
+		    decomp := straightenH#tvec ?? new HashTable from {};
 		    for ssyt in keys decomp do (
 			 if not nuIdx#?ssyt then continue;
 			 dc := decomp#ssyt;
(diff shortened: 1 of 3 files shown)


kellerlv:
One comment about this: since version 1.0 was verified as a JSAG paper, the citation generator was automatically adding me as an author of the actual JSAG article, which isn't correct. So I manually removed the JSAG certification but kept a line in the documentation stating that 1.0 was certified. However, it seems like a small oversight that any authors added after JSAG certification are automatically added as authors of the actual JSAG article.

d-torrance:
Yeah -- that's definitely a minor bug with the *PackageCitations* package.  But on the other hand, keeping track of two separate sets of authors in the package metadata doesn't seem great, either.  The solution is to use a `Citation` key in the main doc node of the package, e.g.,:

https://github.com/Macaulay2/M2/blob/661767fb25505390010482cb453d5d140804acdf/M2/Macaulay2/packages/TerraciniLoci.m2#L137-L147

M2/Macaulay2/packages/PieriMaps.m2
```diff
@@ -10,32 +10,95 @@
 
 newPackage(
     	  "PieriMaps",
-   	  Version => "1.0",
-	  Date => "July 3, 2009",
-	  Certification => {
```
d-torrance, line 15: Let's keep the `Certification` option in here for historical purposes

M2/Macaulay2/packages/PieriMaps/doc.m2
```diff
@@ -0,0 +1,1429 @@
...
+     " changes the bases of the free modules in a way which is not invertible
+     (over ZZ) when the ground field has positive characteristic.",
+     PARA{},
+     "Version 1.0 of this package was accepted for publication in ",
+     HREF{"https://msp.org/jsag/2009/1-1/", "volume 1"}, " of ",
+     HREF{"https://msp.org/jsag/", "The Journal of Software for Algebra and Geometry: Macaulay2"},
+     " on 2009-06-27, in the article ",
+     HREF{"https://msp.org/jsag/2009/1-1/p02.xhtml", "Computing inclusions of Schur modules"},
+     " (DOI: ", HREF{"https://doi.org/10.2140/jsag.2009.1.5", "10.2140/jsag.2009.1.5"},
+     ").",
```
d-torrance: This part is automatically generated when a `Certification` option is present in `newPackage`, so let's drop this after adding `Certification` back (and adding a `Citation` key to the main doc node to avoid the "Keller didn't co-author the JSAG article" issue).

M2/Macaulay2/packages/PieriMaps.m2
```diff
@@ -308,11 +662,155 @@ pieri(List, List, PolynomialRing) := (mu, boxes, P) -> (
...
+--    c'_J = prod_{q intermediate} (mu'_k - mu'_(J_q) + J_q - k)
+-- (the negative of the literal transpose of the row coefficient, accounting
+-- for the wedge anticommutativity).
+-------------------------------------------------------------------------------
+
+pieriColumnHelper = (mu, k, P) -> (
+     ensureSchurFn();
+     n := numgens P;
+     X := gens P;
+     for i from 0 to #mu - 2 do if mu#i < mu#(i+1) then error ("not a partition: " | toString mu);
```
d-torrance: Very minor suggestion -- you don't need to worry about calling `toString` when passing things to `error`, as it already does that for you.  So this could be simplified to `error("not a partition: ", mu)`.

M2/Macaulay2/packages/PieriMaps.m2
```diff
@@ -213,32 +337,240 @@ straighten(List) := t -> (
...
+               for i from 0 to lambdaT#j - 1 list (sigmas#i)#j;
+          totalSign := 1;
+          sortedCols := {};
+          degenerate := false;
+          for c in cols do (
+               sgn := sortWithSignPM c;
+               if sgn === null then ( degenerate = true; break );
+               totalSign = totalSign * (sgn#0);
+               sortedCols = append(sortedCols, sgn#1);
+               );
```
d-torrance, line 410: Rather than calling `append` inside a loop like this and creating a new list on each iteration, it's more efficient to use `for ... list` or `apply`, i.e.,

```m2
sortedCols := for c in cols list ( ... ; sgn#1)
```
or
```m2
sortedCols := apply(cols, c -> ( ... ; sgn#1))
```

M2/Macaulay2/packages/PieriMaps.m2
```diff
@@ -213,32 +337,240 @@ straighten(List) := t -> (
...
+               )
+          );
+     denom := product apply(lambdaT, k -> factorialPM k);
+     accum := new MutableHashTable;
+     cartesianForEachPM(colChoices, choice -> (
+          totalSign := 1;
+          orderedCols := {};
+          for c in choice do (
+               totalSign = totalSign * (c#0);
+               orderedCols = append(orderedCols, c#1);
```
d-torrance: See `append` comment above

M2/Macaulay2/packages/PieriMaps.m2
```diff
@@ -374,240 +872,1450 @@ pureFree(List, PolynomialRing) := (d, P) -> (
...
-     	  pieri({3,1}, {1}, QQ[a,b,c]) -- removes the last box from row 1 of the partition {3,1}
-	  res coker oo -- resolve this map
-	  betti oo -- check that the resolution is pure
-	  ///
-     }
+-- Strip rows (1-indexed) for label a (0-indexed) inside an LR filling Q.
+stripRowsForLabel = (Q, a) -> (
+     rows := {};
+     for i from 0 to #Q - 1 do
+     for lab in Q#i do if lab == a then rows = append(rows, i + 1);
```
d-torrance: See `append` comment above.  This one's a bit less straightforward since you're only calling `append` sometimes and it's a nested for loop. 

Is there a chance that `lab == a` is true more than once for a particular `i`?  If not, then I think this is a one-liner -- something like:
```m2
apply(select(#Q, i -> isMember(a, Q#i), i -> i + 1)
```

If `a` can appear more than once, then it's a bit trickier...  Probably `while ... list` would help.

M2/Macaulay2/packages/PieriMaps.m2
```diff
@@ -374,240 +872,1450 @@ pureFree(List, PolynomialRing) := (d, P) -> (
...
+     mm  := padPart(mu, L);
+     for i from 0 to L - 1 do if lam#i < mm#i then return {};
+     r := #nu;
+     rowLens := for i from 0 to L - 1 list lam#i - mm#i;
+     -- Cells in reverse-reading-word order.
+     cells := {};
+     for i from 0 to L - 1 do (
+	  rs := mm#i;
+	  len := rowLens#i;
+	  for t from 0 to len - 1 do cells = append(cells, (i, rs + len - 1 - t));
```
d-torrance: See `append` comment above

M2/Macaulay2/packages/PieriMaps.m2
```diff
@@ -374,240 +872,1450 @@ pureFree(List, PolynomialRing) := (d, P) -> (
...
+	  for i from 0 to #Nmu - 1 do (
+	       f := composite_(i, j);
+	       if f == 0 then continue;
+	       for pair in listForm f do (
+		    expv := pair#0;
+		    c := pair#1;
+		    if c == 0 then continue;
+		    tvec := for a from 0 to r - 1 list (
+			 ms := {};
+			 for k from 0 to n - 1 do for kk from 1 to expv#(a*n + k) do ms = append(ms, k);
```
d-torrance: See `append` comment above

M2/Macaulay2/packages/PieriMaps.m2
```diff
@@ -374,240 +872,1450 @@ pureFree(List, PolynomialRing) := (d, P) -> (
...
+	       Fl := apply(toList F, toList);
+	       for c from 0 to #Fl - 1 do (
+		    if #(Fl#c) >= 2 then (
+			 swapped := for j from 0 to #(Fl#c) - 1 list (
+			      if j == 0 then (Fl#c)#1
+			      else if j == 1 then (Fl#c)#0
+			      else (Fl#c)#j
+			      );
+			 newF := for cc from 0 to #Fl - 1 list (if cc == c then swapped else Fl#cc);
+			 out = append(out, new sfFillingType from newF);
```
d-torrance: See `append` comment above.  This one's trickier -- maybe use a mutable hash table and store the `sfFillinType` objects as keys?  The values could be anything -- `null` or `true` or whatever -- can then you can call `keys` on the mutable hash table when it's time to return from the function.

M2/Macaulay2/packages/PieriMaps.m2
```diff
@@ -374,240 +872,1450 @@ pureFree(List, PolynomialRing) := (d, P) -> (
...
+     for comp in compositions(n, d) do (
+	  muNew := for i from 0 to n-1 list lamPad_i + comp_i;
+	  if not all(0..n-2, i -> muNew_i >= muNew_(i+1)) then continue;
+	  -- horizontal strip: each column has at most one new cell, i.e.
+	  -- muNew_(i+1) <= lamPad_i for all i.
+	  if not all(0..n-2, i -> muNew_(i+1) <= lamPad_i) then continue;
+	  muTrim := muNew;
+	  while #muTrim > 0 and muTrim#-1 == 0 do muTrim = drop(muTrim, -1);
+	  boxes := flatten for i from 0 to n-1 list toList((comp_i):(i+1));
+	  out = append(out, (muTrim, boxes));
```
d-torrance: See `append` comment above

M2/Macaulay2/packages/PieriMaps.m2
```diff
@@ -374,240 +872,1450 @@ pureFree(List, PolynomialRing) := (d, P) -> (
...
+	  if not all(0..n-2, i -> muNew_i >= muNew_(i+1)) then continue;
+	  muTrim := muNew;
+	  while #muTrim > 0 and muTrim#-1 == 0 do muTrim = drop(muTrim, -1);
+	  -- For each row r in sub, the new box is at column lamPad_r + 1.
+	  -- Sort by column descending (with ties broken by lower row first) so
+	  -- iterated pieriColumn produces valid intermediate partitions.
+	  pairs := sort apply(sub, r -> (lamPad_r + 1, r));
+	  pairs = reverse pairs;
+	  cols := apply(pairs, p -> p#0);
+	  out = append(out, (muTrim, cols));
```
d-torrance: See `append` comment above

M2/Macaulay2/packages/PieriMaps.m2
```diff
@@ -374,240 +872,1450 @@ pureFree(List, PolynomialRing) := (d, P) -> (
...
+     totalCells := sum mu + sum nu;
+     cands := apply(pmPartitionsAtMost(totalCells, n), p -> toList p);
+     blocks := {};
+     blockInfo := {};
+     for lamP in cands do (
+	  lrTabs := lrTableaux(lamP, mu, nu);
+	  for Qp in lrTabs do (
+	       blocks = append(blocks,
+		    lrMap((lamP, mu, nu), Qp, n, Convention => conv));
+	       blockInfo = append(blockInfo, (lamP, Qp));
```
d-torrance: See `append` comment above.  Since you're constructing two lists and are doing it inside a nested for loop, then maybe a mutable list would work well.  Since you know the lengths ahead of time, you can pre-allocate them.  Maybe something like:

```m2
blocks = new MutableList from #cands * #lrTabs
blockInfo = new MutableList from #cands * #lrTabs
i = 0
for lamP in cands do (
   ...
   for Qp in lrTabs do (
      blocks#i = ...
      blockInfo#i = ...
      i += 1))
```

d-torrance (changes requested):
Thanks for the contribution!  I have a few comments

M2/Macaulay2/packages/PieriMaps.m2
kellerlv: I never really understood the difference between this and using append, but after benchmarking I am very surprised how much slower calling append in a loop can be.

M2/Macaulay2/packages/PieriMaps.m2
```diff
@@ -178,9 +306,17 @@ schurRank(ZZ, List) := (n, mu) -> (
...
 -- Output: 
 -- All standard tableaux of shape mu and with labels from 0..dim-1
+-- Memoize standardTableaux: it's called many times per pieri / lrMap with
+-- the same (dim, mu) (e.g., for source and target shapes, repeated across
+-- pieri compositions in lrMap, and between successive iterations of
+-- benchmarks/loops).  The cache is purely a function of (dim, mu).
+standardTableauxCache = new MutableHashTable
+
 standardTableaux = method()
 standardTableaux(ZZ, List) := (dim, mu) -> (
```
mahrud: I think something this might be easier and keep the main code cleaner, though up to you:
```suggestion
standardTableaux(ZZ, List) := (dim, mu) -> standardTableauxCache#(dim, mu) ??= if #mu == 0 then {{}} else (
```

M2/Macaulay2/packages/PieriMaps.m2
```diff
@@ -213,32 +350,238 @@ straighten(List) := t -> (
...
+               if decomp#?U then decomp#U else 0_QQ
+               )
+          );
+     -- Source labels = standard Fillings; target labels = standard PM tableaux.
+     attachBasisLabels(result, sfS, pmS);
+     fillingToPMMatrixCache#key = result;
+     result
+     )
+
+pmToFillingMatrix = (mu, n) -> (
```
mahrud: Similarly here, even simpler:
```suggestion
pmToFillingMatrix = (mu, n) -> pmToFillingMatrixCache#(mm := trimPartPM mu, n) ??= (
```

M2/Macaulay2/packages/PieriMaps.m2
```diff
@@ -374,240 +883,1434 @@ pureFree(List, PolynomialRing) := (d, P) -> (
...
+-- Caching A by (kind, lambda, d, n, K, conv) (resp. (mu, nu, n, K, conv)
+-- for LR) lets sibling calls share the heavy block construction, and
+-- using `solve` instead of `A^{-1}` extracts only the rows we want
+-- (cost O(N^2 * blockSize) instead of O(N^3) for a full inverse).
+--
+-- A typical workflow projects onto many sister summands of the same
+-- ambient tensor product; the cache turns those into nearly-free calls.
+-- ====================================================================
+
+pmDualBlockCache = new MutableHashTable;
```
mahrud, line 1442: And in the next few functions dealing with `pmDualBlockCache`, and further down in `pmCLambdaCache`, but I won't repeat myself.

M2/Macaulay2/packages/PieriMaps.m2
```diff
@@ -374,240 +883,1434 @@ pureFree(List, PolynomialRing) := (d, P) -> (
...
+     else lab
+     );
+
+symbolicForm = method()
+symbolicForm Matrix := M -> (
+     if not M.cache#?"sourceBasis" then
+	  M.cache#"sourceBasis" = toList(1..rank source M);
+     if not M.cache#?"targetBasis" then
+	  M.cache#"targetBasis" = toList(1..rank target M);
+     if not M.cache#?"rule" then (
```
mahrud: Oh but here I will point out:
```suggestion
     M.cache#"sourceBasis" ??= toList(1..rank source M);
     M.cache#"targetBasis" ??= toList(1..rank target M);
     M.cache#"rule" ??= (
```
Very tangential, but is 1-indexing the bases necessary? Usually 0-indexing pays off later on. Also, I believe rank of a module is automatically cached.

kellerlv: 1-indexing is a personal preference, but, technically is not compatible with this package's conventions anyway (tableaux are 0-indexed by default), so I changed it to 0 indexing.

M2/Macaulay2/packages/PieriMaps.m2
d-torrance: Yeah, each call to `append` creates a brand new list and loops through the original list copying elements over before copying the new element in the last spot.  So it's a great way to turn an O(n) algorithm into an O(n^2) one!

We're planning on eventually having it do the sensible thing for *mutable* lists and append in-place, growing the list as needed like Python does, but that won't be until at least the June release, if not later.

M2/Macaulay2/packages/PieriMaps.m2
```diff
@@ -178,24 +306,32 @@ schurRank(ZZ, List) := (n, mu) -> (
...
+	  firstrow := rsort compositions(dim, mu#0);
+	  for i in firstrow do (
+	       temp := {};
+	       for j from 0 to #i-1 do
+	       for k from 1 to i#j when true do
+	       temp = append(temp,j);
+	       for j in otherrows do (
+		    temp2 := prepend(temp,j);
+		    if isStandardPM(temp2) === null then
+		    output = append(output, temp2);
```
d-torrance: Found another spot with a couple of nested appends.   I think this can be simplified by swapping the order of loops, so something like:

```m2
output = apply(otherrows, j -> apply(firstrow, i -> ...))
```

Also, `when true` is redundant

d-torrance (changes requested):
Thanks for the updates!  Found one more `append` issue

mahrud:
@d-torrance this is a good suggestion to support: accumulate-or-init `H#k = (H#k ?? 0) + new`. Maybe `H#k += new` should do this automatically even if `H#k` doesn't exist yet?

d-torrance (approved):
Thanks again!  Tentatively approving pending the Ubuntu builds finally working.  (We're still recovering from a cyberattack on Canonical the other day, and fetching information from their servers keeps timing out...)

d-torrance:
> this is a good suggestion to support: accumulate-or-init `H#k = (H#k ?? 0) + new`. Maybe `H#k += new` should do this automatically even if `H#k` doesn't exist yet?

This reminds me of Python's `defaultdict` class:

```python3
>>> import collections
>>> x = collections.defaultdict(lambda:0)
>>> x[0]
0
>>> x[1] += 1
>>> x[1]
1
```
We could do it all at top level if we use `_` instead of `#`:

```m2
i1 : DefaultDict = new SelfInitializingType of MutableHashTable;

i2 : new DefaultDict from Function := (T, f) -> (
         d := new DefaultDict;
         d.DefaultFactory = f;
         d);

i3 : DefaultDict_Thing := (d, k) -> (
         if d#?k then d#k
         else d#k = d.DefaultFactory());

i4 : DefaultDict_Thing = (d, k, e) -> d#k = e;

i5 : x = DefaultDict(() -> 0)

o5 = DefaultDict{...1...}

o5 : DefaultDict

i6 : x_0

o6 = 0

i7 : x_1 += 1

o7 = 1

i8 : x_1

o8 = 1
```

mahrud:
That's way too complicated. I would prefer something simple and intuitive, e.g. maybe `H#k op= x` initialize `H#k` with the unit of `op` (so 0 for += and -= but 1 for *= and /=). Anyway, let's continue elsewhere.

PR #4249, opened 2026-05-03
Add minimalPrimes strategy for ideals in ZZ and ZZ/n

### LINKED ISSUE #3505: isPrime in ZZ/n

d-torrance:
This closes #3505:

```m2
i1 : isPrime ideal 0_(ZZ/2)

o1 = true

i2 : isPrime ideal 0_(ZZ/4)

o2 = false
```

We also add a few related updates I encountered along the way:

* We add `module(RingFamily)` so `module RR` and `module CC` work.
* We add `presentation(Ring)` (and `presentation(RingFamily)` so `presentation ZZ` works.  It calls `presentation(Module)`, so the result gets cached.
* We add `RingElement^Ring` as syntactic sugar for `lift` (matching `Number^Ring`)
* We can now find the characterisic of tower rings over `ZZ`.

### DIFF
--- a/M2/Macaulay2/m2/quotring.m2
+++ b/M2/Macaulay2/m2/quotring.m2
@@ -232,8 +232,15 @@ Ring / List := Ring / Sequence := QuotientRing => (R,f) -> R / promote(ideal f, 
 
 -----------------------------------------------------------------------------
 
-presentation PolynomialRing := Matrix => R -> map(R^1, R^0, 0)
-presentation QuotientRing   := Matrix => R -> R.presentation ??= (
+ZZ.presentation =
+QQ.presentation = R -> presentation module R
+presentation Ring := Matrix => R -> (
+    if R.?presentation then R.presentation R
+    else notImplemented "presentation for this ring")
+presentation InexactFieldFamily := Matrix => R -> presentation default R
+presentation PolynomialRing :=
+presentation InexactField   := Matrix => R -> presentation module R
+presentation QuotientRing   := Matrix => R -> R.cache.presentation ??= (
 	  S := ambient R;
 	  f := generators ideal R;
 	  while class S === QuotientRing do (		    -- untested code
--- a/M2/Macaulay2/m2/quotring.m2
+++ b/M2/Macaulay2/m2/quotring.m2
@@ -283,7 +283,7 @@ char QuotientRing := (stashValue symbol char) ((S) -> (
      if isPrime p or isMember(QQ,S.baseRings) then return if S == 0 then 1 else p;
      relns := presentation S;
      if relns == 0 then return char ring relns;
-     if coefficientRing S =!= ZZ then notImplemented();
+     if ultimate(coefficientRing, S) =!= ZZ then notImplemented();
      g := generators gb relns;
      if g == 0 then return char ring g;
      m := g_(0,0);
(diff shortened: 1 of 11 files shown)


M2/Macaulay2/m2/quotring.m2
```diff
@@ -232,7 +232,8 @@ Ring / List := Ring / Sequence := QuotientRing => (R,f) -> R / promote(ideal f,
 
 -----------------------------------------------------------------------------
 
-presentation PolynomialRing := Matrix => R -> map(R^1, R^0, 0)
+presentation Ring           :=
+presentation RingFamily     := Matrix => presentation @@ module
```
mahrud: `@@` is unnecessarily opaque, as it makes it hard to see the code.
Also, I'm worried about someone defining some other type of non-polynomial ring, forgetting to define `presentation` for it, and because of this line always getting that the presentation of their ring is trivial.

mahrud: Yeah, this seems to be a problem with at least SheafOfRings:
```m2
i40 : (showStructure())#Thing#HashTable#MutableHashTable#Type#Ring

o40 = EngineRing : FractionField
                   FreeAlgebra
                   GaloisField
                   InexactField : ComplexField
                                  RealField
                   LocalRing
                   PolynomialRing
                   QuotientRing : FreeAlgebraQuotient
                   SchurRing
      NCRing : NCPolynomialRing
               NCQuotientRing
      SheafOfRings
```

d-torrance: Good point.  I've updated it so that `(presentation, Ring)` first looks for the presence of a `presentation` key in the ring object (which is only installed for `ZZ` and `QQ` and w/o using `@@`), runs it if it's there, and raises an error otherwise:

```m2
i1 : presentation ZZ

o1 = 0

              1
o1 : Matrix ZZ  <-- 0

i2 : presentation QQ

o2 = 0

              1
o2 : Matrix QQ  <-- 0

i3 : presentation new Ring
stdio:3:12:(3):[1]: error: presentation for this ring not implemented yet
```

d-torrance:
Discussed with Anton and Mike

PR #4251, opened 2026-05-03
Update hash for ZZ/RR to be platform-independent and use 64 bits

d-torrance:
Previously, we used (mostly) 32-bit integer arithmetic for computing hashes for `RR` objects and large `ZZ` objects that don't fit in a C `int`, and our hashing function was platform-dependent.  For example:

```m2
i1 : version#"pointer size"

o1 = 4

i2 : hash 1.23p200

o2 = 18446744073069319069

i3 : hash 2^100

o3 = 59792
```

vs.

```m2
i1 : version#"pointer size"

o1 = 8

i2 : hash 1.23p200

o2 = 18446744072207201388

i3 : hash 2^100

o3 = 0
```

Also, check this out!  We're in collison town!

```m2
i1 : tally apply(100, hash @@ numeric)

o1 = Tally{842 => 2 }
           843 => 2
           844 => 4
           845 => 8
           846 => 16
           847 => 32
           848 => 36

o1 : Tally
```

We switch to a new algorithm uses 64-bit arithmetic throughout and uses `mpz_export` to guarantee we get the same result regardless of pointer size and endianness.  After the update:

```m2
i1 : version#"pointer size"

o1 = 4

i2 : hash 1.23p200

o2 = 10359908877735906505

i3 : hash 2^100

o3 = 3405506911546692669
```
and

```m2
i1 : version#"pointer size"

o1 = 8

i2 : hash 1.23p200

o2 = 10359908877735906505

i3 : hash 2^100

o3 = 3405506911546692669
```

And no more collisions :)

```m2
i5 : #tally apply(100, hash @@ numeric)

o5 = 100
```

## AI Disclosure

Claude Code wrote the first draft of this code, and in particular was responsible for the suggestion to use `mpz_export` to achieve platform independence.  I heavily modified the code that was generated.

### DIFF
--- a/M2/Macaulay2/d/gmp.d
+++ b/M2/Macaulay2/d/gmp.d
@@ -1406,17 +1406,9 @@ export intersectRRi (x:RRi, y:RRi, prec:ulong):RRi := (
      Ccode( void, "mpfi_intersect(", z, ",",  x, ",",  y, ")" );
      moveToRRiandclear(z));
 
-export hash(x:RR):hash_t := hash_t(precision0(x)) + Ccode(hash_t,
-     "mpfr_hash(",					    -- see gmp_aux.c for this function
-          x, 
-     ")"
-    );
+export hash(x:RR):hash_t := Ccode(hash_t, "mpfr_hash(", x, ")"); -- see gmp_aux.c
 
-export hash(x:RRi):hash_t := hash_t(precision0(x)) + Ccode(hash_t,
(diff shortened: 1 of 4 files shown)


mahrud:
Shocked at `apply(100, hash @@ numeric)`

d-torrance:
Discussed with Anton and Mike

Question:  Why does ZZ have a bucket with 5 things in it?

PR #4252, opened 2026-05-03
Update exterior algebra docs

joel-dodge:
Updates to the docs page on exterior algebras.

Clarifies the note on left / right module structure over exterior algebras and provides a link to the general docs on modules in macaulay. Improve the example there to give a better indication of what is being communicated. Add a warning example that quotients of exterior algebras by non-two-sided ideals are not well-defined.

edit: updated screenshot to reflect current state of changes.
_usr_local_share_doc_Macaulay2_Macaulay2Doc_html__exterior_spalgebras html (2)

### DIFF
--- a/M2/Macaulay2/packages/Macaulay2Doc/ov_rings.m2
+++ b/M2/Macaulay2/packages/Macaulay2Doc/ov_rings.m2
@@ -1190,50 +1190,62 @@ document {
      }
 
 document {
-     Key => "exterior algebras",
-     -- making one, making quotients,
-     -- using it.
-     -- modules are right-modules, example of multiplication.
-     "An exterior algebra is a polynomial ring where multiplication is
-     mildly non-commutative, in that, for every x and y in the ring,
-     y*x = (-1)^(deg(x) deg(y)) x*y, and that for every x of odd degree,
-     x*x = 0.",
-     "In Macaulay2, deg(x) is the degree of x, or the first degree of x, in case 
-     a multi-graded ring is being used.  The default degree for each variable is 1, so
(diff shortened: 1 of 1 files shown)


joel-dodge:
I need to update the development branch of my fork.

mahrud:
Thanks! If you're going through the trouble of rewriting the node, it would be preferable to write it in SimpleDoc format instead of the old style.

joel-dodge:
Oh great. I followed my nose to make these changes and it did not lead me to discover that there was a new format. I'll poke around and update this to the new style!

joel-dodge:
For documentaiton packages that are not tied to a specific package, is there a quick way to generate the current state of the docs (with perhaps in-progress edits) in the browser? I see the suggestionn on how to do so for package documentation here: https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_writing_spdocumentation.html

I am iterating by rebuilding macaulay2 to see updates but doing so takes a few minutes when there are docs changes and if there is a faster way I'd love to know about it!

d-torrance:
The Core documentation *is* a package -- *Macaulay2Doc*.

Rather than installing it (which takes quite a while), here's my usual workflow for viewing changes that I've made to the documentation:

```m2
loadPackage("Macaulay2Doc", LoadDocumentation => true, Reload => true, FileName => "/path/to/Macaulay2Doc.m2")
-- wait for a few seconds -- Macaulay2Doc is pretty big!
help "doc node i just edited"
show oo -- to view the html version in the browser
```

d-torrance (approved):
Thanks!

PR #4253, opened 2026-05-03
fix stderr in openFiles in WebApp mode

pzinn:
This is a followup to #4188 -- in WebApp mode we currently have `toString first openFiles() == "stderr"` but `first openFiles() =!= stderr`, i.e., `openFiles()` isn't correctly updated to the new `stderr`. This fixes it.


d-torrance (approved):
Thanks!

Issue #667, opened 2017-07-28
hash codes of function closures


DanGrayson:
This identity of hash codes is counterintuitive and may slow things down, if functions are used as keys in hash tables:
```
i48 : g = i -> () -> i

o48 = g

o48 : FunctionClosure

i49 : hash g 2

o49 = 1204238

i50 : hash g 3

o50 = 1204238
```

The same remark applies to symbols returned by functions.

mahrud:
Related conversation: #1567

d-torrance:
This was fixed in #2934

mahrud:
It certainly chamged, though I do wish `hash g 3 == hash g 3` was true.

d-torrance:
I'll go ahead and close this (I meant to close it when adding https://github.com/Macaulay2/M2/issues/667#issuecomment-4365270454, but apparently clicked the wrong button...).  Since function closures are mutable, I think using generating the hash from the address like we're doing now is probably as good as we can do.

Issue #2591, opened 2022-09-02
Broken hash functions


mahrud:
The function `hash` seems to be broken for certain classes of objects:
1. MutableMatrix #997
2. MethodFunctionWithOptions
```m2
i44 : L = select(value \ values Core.Dictionary, x -> class x === MethodFunctionWithOptions);

i45 : unique \\ hash \ L

o45 = {242339741, 242343940}

o45 : List

i46 : #L

o46 = 212
```
3. Composed functions
```m2
Related:
```m2
i1 : hash(toList @@ hilbertFunction)

o1 = 1091563

i2 : hash(expression @@ net)

o2 = 1091563
```
4. This random collection of `MethodFunctionSingle`s:
```m2
document, SYNOPSIS, makeDocumentTag, monoid, hooks
```
5. Most `MethodFunctionBinary`s
```m2
104 : L = select(value \ values Core.Dictionary, x -> class x === MethodFunctionBinary);

i105 : unique \\ hash \ L

o105 = {243950607, 242343937, 243946629}

o105 : List

i106 : #L

o106 = 8
```
There are more collisions that are likely just coincidence, but the above are likely bugs.

DanGrayson:
These collisions are important to prevent only if the functions are used as keys in hashtables.  I think they never are.

mahrud:
Literally all of the methods above are used as keys in hash tables 🙃

DanGrayson:
Which hash tables?

mahrud:
Types.

DanGrayson:
Ah, right, I was focusing on the functions in your item #3 above, which are not.

A particularly egregious example is `(buckets Module)#149`, which has 29 entries in a linked list.  Making the hash codes different might be a good way to avoid searching that list every time one of those methods is called.  And a good way to do that would be to add a integer field to the line `export FunctionClosure := {+ frame:Frame, model:functionCode };`, initializing it from the value of `nextHash()`.

DanGrayson:
Good catch!

mahrud:
> A particularly egregious example is `(buckets Module)#149`, which has 29 entries in a linked list.

That isn't the worst by far:
```m2
i92 : max \\ length \ buckets typicalValues

o92 = 40

i93 : max \\ length \ buckets Attributes

o93 = 157
```

Even for types alone, there are 32 keys with the same hash in Matrix and 49 on Ideal:
```m2

i1 : c = first commonest tally apply(toList methods Ideal, hash)

[...]

Issue #1567, opened 2020-10-30
Lisp equality of functions


mahrud:
Why is === defined the way it is for functions? I imagine == is rather difficult for functions, but shouldn't === just check that the lisp version of both functions are the same? I think that would make sense, but I'm actually surprised about this:

```m2
i1 : f = x -> y -> 2*x;

i2 : (f 2) == (f 2)
stdio:2:7:(3): error: no method for binary operator == applied to objects:
--            -*Function[stdio:1:12-1:17]*- (of class FunctionClosure)
--     ==     -*Function[stdio:1:12-1:17]*- (of class FunctionClosure)

i3 : (f 2) === (f 2)

o3 = false -- WHAT??

i4 : code (f 2) == code (f 2)

o4 = true

i5 : code (f 2) === code (f 2)

o5 = true

i6 : code (f 2) == code (f 3)

o6 = false

i7 : code (f 2) === code (f 3)

o7 = false

i8 : disassemble (f 2) == disassemble (f 2)

o8 = true

i9 : disassemble (f 2) === disassemble (f 2)

o9 = true

i10 : disassemble (f 2) == disassemble (f 3)

o10 = true -- WHAT??

i11 : disassemble (f 2) === disassemble (f 3)

o11 = true -- WHAT??

```

_Originally posted by @mahrud in https://github.com/Macaulay2/M2/pull/1385#issuecomment-719007041_

mahrud:
>> Why is === defined the way it is for functions? I imagine == is rather difficult for functions, but shouldn't === just check that the lisp version of both functions are the same? I think that would make sense, but I'm actually surprised about this:
>
>The code for `f===g` is designed to never go into an infinite loop, and part of the way of ensuring that is that cycles in the graph of pointers are not traversed.  That's why mutable objects are compared, not by comparing their contents, but by comparing the pointers to them.  Functions are mutable objects, because the values stored as the values of local variables can be changed.
>
>Anyway, just don't use functions as keys -- it's easy to avoid.

_Originally posted by @DanGrayson in https://github.com/Macaulay2/M2/pull/1385#issuecomment-719018348_

mahrud:
Despite the argument above, I think === equality of functions should be based on comparison of the "disassembled" version of it. Specifically, this reasoning doesn't make sense to me:

> Functions are mutable objects, because the values stored as the values of local variables can be changed.

[...]

PR #4254, opened 2026-05-04
Assign magic numbers for hashing sequences to variables

d-torrance:
*Update:*  This PR originally involved hashing of functions and pseudocode objects, but I dropped pretty much everything and kept just a single slightly-related commit: using variables instead of magic numbers for computing hash codes of sequences.  We do this in two places: in `hash` for sequences and also in various `lookup` methods where it's quicker to just compute what the hash code would be if we had a sequence object but without actually constructing it. 

~~We add hash codes to all members of the `Code` union in the interpreter, which we then use to compute proper hashes of function and pseudocode objects.~~

### Before

```m2
i1 : hash (() -> print "Hello, world!")

o1 = 6474929089309016109

i2 : hash (() -> print "Hello, world!")

o2 = 14328623249302651656
```

### After
```m2
i1 : hash(() -> print "Hello, world!")

o1 = 9105046539553097294

i2 : hash(() -> print "Hello, world!")

o2 = 9105046539553097294
```

~~Closes: #667~~

## AI Disclosure

This was almost entirely vibe-coded by Claude Code, but with serious hand-holding.

### DIFF
--- a/M2/Macaulay2/d/basic.d
+++ b/M2/Macaulay2/d/basic.d
@@ -4,6 +4,10 @@ use expr;
 
 header "#include <engine.h>"; -- required for raw hash functions
 
+-- used to hash sequences below and for quick method lookup in hashtables.dd
+export seqHashSeed := hash_t(27449);
+export seqHashMult := hash_t(27457);
+
 export hash(e:Expr):hash_t := (
      when e
      is x:HashTable do x.hash
@@ -26,9 +30,8 @@ export hash(e:Expr):hash_t := (
      is x:RRicell do hash(x.v)
      is x:CCcell do hash(x.v)
      is x:Sequence do (
-	  -- the numbers here are the same as in binary lookup() in objects.d!!
-	  h := hash_t(27449);
-	  foreach y in x do h = h * 27457 + hash(y);
+	  h := seqHashSeed;
+	  foreach y in x do h = h * seqHashMult + hash(y);
 	  h)
      is x:stringCell do hash(x.v)				    -- for strings, keep internal and external hash the same
      is n:Net do hash(n)
--- a/M2/Macaulay2/d/hashtables.dd
+++ b/M2/Macaulay2/d/hashtables.dd
@@ -475,8 +475,7 @@ export lookupUnaryValue(s:HashTable,meth:Expr,methhash:hash_t):Expr := (
      key1 := Key1;
      key1.0 = Expr(s);
      key1.1 = meth;
-     -- the big numbers here are the same as in hash() for sequences in structure.d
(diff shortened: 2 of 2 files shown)


pzinn:
Are you saying Claude Code can program in "d", despite the lack of any meaningful documentation?

d-torrance:
Yup! It was basically just copying and modifying the hashing function for the Expr union, but it did a decent job!

pzinn:
Interesting. I for one would be happy to never have to write another line of "d".
BTW:
Maybe that's something you guys can talk about at the upcoming M2 meeting: how to get Claude Code to entirely rewrite the whole d code into a more modern language...

M2/Macaulay2/d/basic.d
```diff
@@ -120,7 +120,7 @@ export hash(e:Expr):hash_t := (
      is x:RRicell do hash(x.v)
      is x:CCcell do hash(x.v)
      is x:Sequence do (
-	  -- the numbers here are the same as in binary lookup() in objects.d!!
+	  -- these constants must match the manual hash computations in hashtables.dd
```
mahrud: Would it be feasible to have a single source of truth for all magic numbers?

d-torrance: I just pushed a commit assigning them to variables.  From what I can understand, it's written this way for quick method lookup so we don't have to construct a brand new sequence and hash that.

d-torrance:
Currently, this is changing some behavior, e.g.,, if we have `f = i -> () -> i` (as in `testing.m2`), then `hash f i` will be the same regardless of the value of `i`.  I'm not sure if that's ideal or not.  We could try to hash the frame values in some way, but they're mutable.

mahrud:
I believe there are a few issues where Dan explained to me that this is exactly why function closures have distinct hashes :)

d-torrance:
D'oh -- now `res` and `syz` have the same hash code again (since they just use the same `uncurry` code).  I think I'm going to drop everything but the sequence hash stuff.

mahrud:
> D'oh -- now `res` and `syz` have the same hash code again (since they just use the same `uncurry` code). I think I'm going to drop everything but the sequence hash stuff.

Yeah, #1567 has good context about this also. Might be nice to add a few tests to make sure this and the problem in #2591 don't unintentionally happen again.

d-torrance:
Discussed with Anton and Mike

mahrud:
A suggestion that Dan had at some point: we should occasionally perturb these magic number (maybe just in a debug build, not even necessarily in a release), just to catch tests that fail because the ordering of keys in hash tables changes (which would indicate a false assumption).

d-torrance:
Ooh that's a cool idea.  Maybe I'll do that right now...

PR #4255, opened 2026-05-04
BettiCharacters v2.6 (with degree fix)

galettof:
BettiCharacters has been updated to version 2.6. Changes include significant optimizations of the core algorithm, a new option for semidirect products of finite groups acting on tori, and methods for hyperoctahedral groups. This version introduces breaking changes to the Character type, making it incompatible with previous versions.

### DIFF
--- a/M2/Macaulay2/packages/BettiCharacters.m2
+++ b/M2/Macaulay2/packages/BettiCharacters.m2
@@ -36,7 +36,6 @@ newPackage(
 	 "published article URI" => "https://msp.org/jsag/2023/13-1/p04.xhtml",
 	 "published article DOI" => "10.2140/jsag.2023.13.45",
 	 "published code URI" => "https://msp.org/jsag/2023/13-1/jsag-v13-n1-x04-BettiCharacters.m2",
-	 "repository code URI" => "https://github.com/Macaulay2/M2/blob/master/M2/Macaulay2/packages/BettiCharacters.m2",
 	 "release at publication" => "a446af4424af33c06ab97694761a4d5bbc4d535f",
 	 "version at publication" => "2.1",
 	 "volume number" => "13",
--- a/M2/Macaulay2/packages/BettiCharacters.m2
+++ b/M2/Macaulay2/packages/BettiCharacters.m2
@@ -1,5 +1,5 @@
 --------------------------------------------------------------------------------
--- Copyright 2021-2023  Federico Galetto
+-- Copyright 2021-2026  Federico Galetto
 --
 -- This program is free software: you can redistribute it and/or modify it under
 -- the terms of the GNU General Public License as published by the Free Software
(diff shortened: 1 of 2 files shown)


M2/Macaulay2/packages/BettiCharacters.m2
```diff
@@ -36,6 +36,7 @@ newPackage(
 	 "published article URI" => "https://msp.org/jsag/2023/13-1/p04.xhtml",
 	 "published article DOI" => "10.2140/jsag.2023.13.45",
 	 "published code URI" => "https://msp.org/jsag/2023/13-1/jsag-v13-n1-x04-BettiCharacters.m2",
+	 "repository code URI" => "https://github.com/Macaulay2/M2/blob/master/M2/Macaulay2/packages/BettiCharacters.m2",
```
d-torrance: You can drop this "repository code URI" option -- it's not used any more (plus it refers to the old "master" branch we've since renamed to "stable")

M2/Macaulay2/packages/BettiCharacters.m2
```diff
@@ -897,59 +1243,324 @@ symmetricGroupActors PolynomialRing := R -> (
...
+    rsort while #used < #sigma list (
+	u := min toList( set(sigma) - set(used) );
+	cycle := {};
+	while not isMember(u,cycle) do (
+	    cycle = append(cycle,u);
+	    u = sigma_u;
+	    );
+	used = used | cycle;
+	#cycle
+	)
```
d-torrance, line 1281: This algorithm involves a bunch of appends and joins inside loops, which isn't very efficient.   It would be great to refactor this to use mutable lists or hash tables instead.  (I think it's nontrivial to do, so this is just a suggestion not a blocker for merging.)

Another option:  the *Permutations* package has a `cycleType` method you could call here if you construct a `Permutation` object from `sigma`.

d-torrance (changes requested):
Thanks!  Just a couple comments

M2/Macaulay2/packages/BettiCharacters.m2
galettof: Okay thanks! I will remove that line.

M2/Macaulay2/packages/BettiCharacters.m2
```diff
@@ -524,91 +797,117 @@ numActors(Action) := ZZ => A -> A.numActors
...
-    )
-
 -- returns various group actors
 actors = method(TypicalValue=>List)
 
 -- returns actors on resolution in a given homological degree
--- if homological degree is not the one passed by user,
+-- if homological degree is not the one passed by user upon construction,
 -- the actors are computed and stored
 actors(ActionOnComplex,ZZ) := List => (A,i) -> (
```
mahrud, line 812: Same comment as before about caching simplifications:
```suggestion
actors(ActionOnComplex,ZZ) := List => (A,i) -> A.cache#(symbol actors, i) ??= (
```

galettof: Maybe I can have this done in the next couple of weeks and have it included in 1.26.06?

M2/Macaulay2/packages/BettiCharacters.m2
galettof: This is used to compute character tables of hyperoctahedral groups H_n. For values of n used in practice, this is reasonably quick. For larger n, I agree this can be improved but that point the other algorithms in the package will become the bottleneck.

I first wrote this code when Permutations was not included and, though I considered using it later, I like the idea of minimizing dependencies, especially for small portions of code like this one.

In summary: if it's okay, let's keep this as is. I will make a mental note to use mutable lists or hash tables for future code.

d-torrance: Sounds good!

M2/Macaulay2/packages/BettiCharacters.m2
d-torrance: That works!

d-torrance (approved):
Will merge pending the builds passing

PR #4256, opened 2026-05-04
WittVectors package (V1)

Devlin-Mallory:
This is a first draft of the `WittVectors` package, arising from @kschwede's RTG 2023--2024 Macaulay2 seminar. This implements (truncated) Witt vectors of positive-characteristic rings, including addition, multiplication, etc., explicit presentations of (truncated) Witt vectors, and related methods for lifts of Frobenius and quasi-F-splitting height. 

This is joint work with Anne Fayolle, Abhay Goel, Eamon Quinlan-Gallego, and Teppei Takamatsu. Comments and suggestions very welcome!


M2/Macaulay2/packages/WittVectors/FrobeniusLiftings.m2
```diff
@@ -0,0 +1,117 @@
...
+    J := sub(ideal for j in constraint_* list (inverse Cflatmap)( ideal last coefficients Cflatmap sub(j, Cflat)), Cflat);
+    elimList := (for i in gens Bcc list sub(i, Cflat)) | (for i in gens S list sub(i, Cflat));
+    sub(eliminate(elimList, J), (ZZ/p)[gens T])
+)
+
+
+evalMap = method()
+evalMap(List, Ideal, Ring):= (L, I, T) -> (
+    S := ring I;
+    if length L != numgens S then return "error: needed a list with as many entries as the variables of S";
```
d-torrance: We should probably actually raise an error (i.e., `error "needed a list ..."`) here rather than return a string

M2/Macaulay2/packages/WittVectors/QuasiFSplittings.m2
```diff
@@ -0,0 +1,103 @@
...
+I#4 = x^4 + y^4 + z^4 + w^4 + x^2*z^2 + x*y*z^2;
+I#5 = x^4 + y^4 + z^4 + w^4 + x^3*z + z^3*w + y*z^2*w + y*z*w^2;
+I#6 = x^4 + y^4 + z^4 + w^4 + x^2*z^2 + x^2*y*z;
+I#7 = x^4 + y^4 + z^4 + w^4 + x*y^2*z + x*z^2*w + y*z*w^2 + y^2*z*w;
+I#8 = x^4 + x^2*y*z + x^2*y*w + 2*x^2*z^2 + x*y*w^2 + 2*y^4 + y^3*w + z^4 + w^4;
+I#9 = x^4 + y^4 + z^4 + w^4 + x*y^3 + y^3*w + z^2*w^2 + 2*x*y*z^2 + y*z*w^2;
+I#10 = x^4 + 2*x^2*y*z + x^2*y*w + x*y^2*w + y^4 + y^3*w + y^2*z^2 +2*y^2*z*w + y^2*w^2 + y*z^3 + y*z^2*w + y*z*w^2 + z^4 + z*w^3;
+I#11 = x^4 + y^4 + z^4 + w^4;
+if n>10 then return I#11 else return I#n
+)
```
d-torrance: It would be more efficient to construct the polynomials ahead of time rather than constructing all 11 of them on every call to this function.

Also, is this function even necessary?  It's not exported, and the only place I see it called is in one of the tests.

M2/Macaulay2/packages/WittVectors/WittConstructor.m2
```diff
@@ -0,0 +1,404 @@
...
+--WittIdeal display
+
+addCommas := LL -> (
+    nn := #LL;
+    out := ();
+    for jj in 0..(2*nn-2) do(
+	if jj % 2 == 0 then(
+	    out = append(out, LL#(jj//2))
+	    ) else (
+	    out = append(out, ", ");
```
d-torrance: In general, using `append` like this inside a loop isn't efficient because it constructs a brand new list on each iteration.  But for this particular function, I think you can just drop it entirely and use [`demark`](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_demark.html).

M2/Macaulay2/packages/WittVectors/WittConstructor.m2
```diff
@@ -0,0 +1,404 @@
...
+    w1 == n*witt({1_(unWitt ring w1)} | for i from 0 to wittLength ring w1 - 2 list 0)
+)
+
+verschiebung(WittRingElement) := ww -> (
+    R := (ring ww).unWitt;
+    witt({0_R}|ww.tuple)
+)
+
+truncate(ZZ, WittRingElement) :=  {} >> opts -> (n, w) -> (
+    if length w<n then error "Can't truncate to something longer";
```
d-torrance: I'd use a lowercase "c" here for consistency with most other M2 error messages

M2/Macaulay2/packages/WittVectors.m2
```diff
@@ -0,0 +1,73 @@
...
+	{Name => "Eamon Quinlan-Gallego", Email => "eamon.quinlan@utah.edu", HomePage => "https://eamonqg.github.io/"},
+    	{Name => "Teppei Takamatsu", Email => "teppeitakamatsu.math@gmail.com", HomePage => "https://sites.google.com/view/teppei-takamatsu/home"}
+    },
+    PackageImports => {
+        "TestIdeals",
+        "Polyhedra",
+        "PushForward",
+        "Elimination",
+    },
+    Headline => "A Macaulay2 package for Witt vectors",
```
d-torrance: The "A Macaulay2 package for" here is redundant.  Ideally, a package heading completes the phrase "a package for ..."

M2/Macaulay2/packages/WittVectors.m2
```diff
@@ -0,0 +1,73 @@
...
+    Keywords => {"Witt Vectors"},
```
d-torrance: The "Keywords" option is used for determining which heading(s) each package appears under on the [packages provided with Macaulay2](https://www.macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_packages_spprovided_spwith_sp__Macaulay2.html) page.  I'd suggest picking one or two of the existing headings unless you feel strongly that "Witt Vectors" should be a new heading on this page.

d-torrance (changes requested):
Thanks for the contribution!  I have a few comments

d-torrance (changes requested):
Thanks for the contribution!  I have a few comments

M2/Macaulay2/packages/WittVectors/Documentation.m2
```diff
@@ -0,0 +1,1128 @@
...
+	 Example 
+		S = ZZ/5[x,y,z]
+		I = ideal(y*x,z*x)
+	 	fSplittingHeight I
+	 Text
+	    We get an error since the ideal is not generated by a regular sequence
+	 Example
+		S = GF(4)[x,y,z]
+		I=ideal(x)
+		fSplittingHeight I
```
d-torrance, line 406: Raising an error in an example causes the build to fail.  To avoid this, add `stopIfError = false` first

d-torrance (changes requested):

M2/Macaulay2/packages/WittVectors/FrobeniusLiftings.m2
Devlin-Mallory: done!

M2/Macaulay2/packages/WittVectors/QuasiFSplittings.m2
Devlin-Mallory: good point - this was an older function that we had left in for testing purposes. I've moved the relevant polynomials directly to the test file.

M2/Macaulay2/packages/WittVectors/WittConstructor.m2
Devlin-Mallory: I reworked `net(WittIdeal)` in a different way, but still avoiding append.

M2/Macaulay2/packages/WittVectors/WittConstructor.m2
Devlin-Mallory: done!

M2/Macaulay2/packages/WittVectors.m2
Devlin-Mallory: ah, had no idea! fixed now.

M2/Macaulay2/packages/WittVectors/Documentation.m2
Devlin-Mallory: thank you, added that now!

Devlin-Mallory:
> Thanks for the contribution! I have a few comments

Thanks so much for the quick review! I think I've implemented all the suggestions you made so far.

d-torrance (changes requested):
One additional comment:  We're missing a top-level documentation node.  There should be a node like the following that serves as an introduction to the package.

```m2
doc ///
  Key
    WittVectors
  ...
```

d-torrance (approved):
Thank you!

PR #4258, opened 2026-05-05
Expand setupPromote docs

### LINKED ISSUE #4208: Improve setupPromote documentation
The `setupPromote` docs currently only mention `setupPromote(RingMap)`, but there are 5 more methods to mention!

d-torrance:
This fixes #4208.

## AI Disclosure

I had Claude Code examine the M2 source to determine when we might want to use the optional fourth argument for setting up a degree map and suggest a good example and accompanying text.


pzinn:
This is fine, but the example Claude Code created is a bit silly -- why not just define a ring map to do the job, rather than a function that calls `sub` (i.e., creates a ring map anyway, but with the wrong degree mapping, apparently).

PR #4260, opened 2026-05-05
Allow lifting from fraction field -> local ring

d-torrance:
This is a quick followup to #4204 addressing https://github.com/Macaulay2/M2/pull/4204#issuecomment-4381090881.

For example:

```m2
i2 : R = QQ[x];

i3 : p = ideal x;

o3 : Ideal of R

i4 : S = R_p;

i5 : lift(1/(x - 1), S)

       1
o5 = -----
     x - 1

o5 : S

i6 : lift(1/x, S)
stdio:6:4:(3):[1]: error: expected a denominator outside the maximal ideal
```

### DIFF
--- a/M2/Macaulay2/packages/LocalRings/localring.m2
+++ b/M2/Macaulay2/packages/LocalRings/localring.m2
@@ -67,9 +67,6 @@ localRing(EngineRing, Ideal) := (R, P) ->
         RP.localRing    = RP;
         RP.maxIdeal     =  P;
         commonEngineRingInitializations RP;
-	setupPromote(
-	    f -> numerator f / denominator f,
-	    RP, frac R);
 	RP.residueMap   = map(frac(R/P), RP, vars R % P);
          expression RP := r -> expression numerator r / expression denominator r;
            toString RP := r -> toString expression r;
@@ -94,5 +91,14 @@ localRing(EngineRing, Ideal) := (R, P) ->
         if R.?generatorExpressions then RP.generatorExpressions = R.generatorExpressions;
         if R.?indexSymbols then RP.indexSymbols = applyValues(R.indexSymbols, r -> promote(r,RP));
         if R.?indexStrings then RP.indexStrings = applyValues(R.indexStrings, r -> promote(r,RP));
+	setupPromote(
+	    f -> numerator f / denominator f,
+	    RP, frac R);
+	liftFromFractionFieldMap := map(RP, frac R);
+	setupLift(f -> (
+		if isMember(denominator f, RP.maxIdeal)
+		then error "expected a denominator outside the maximal ideal"
+		else liftFromFractionFieldMap f),
+	    frac R, RP);
         RP
         )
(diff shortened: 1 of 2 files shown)


M2/Macaulay2/packages/LocalRings/localring.m2
```diff
@@ -70,6 +70,11 @@ localRing(EngineRing, Ideal) := (R, P) ->
 	setupPromote(
 	    f -> numerator f / denominator f,
 	    RP, frac R);
+	setupLift(f -> (
+		if isMember(denominator f, RP.maxIdeal)
+		then error "expected a denominator outside the maximal ideal"
+		else (map(RP, frac R)) f),
```
mahrud: This map should be cached.

d-torrance: Like cached in a cache table for the local ring, or assigned to a local variable before the call to `setupLift`?

mahrud: It shouldn't be created every time promote happens. Creating ring maps is bizarrely slow.

d-torrance: I agree, but which of the following are you suggesting?

```m2
RP.cache.liftFromFractionFieldMap = map(RP, frac R);
setupLift( ..., RP.cache.liftFromFractionFieldMap f ...)
```

or

```m2
liftFromFractionFieldMap := map(RP, frac R);
setupLift( ..., liftFromFractionFieldMap f ...)
```

mahrud: Probably the latter. Ideally `map(Ring, Ring)` should do this automatically, maybe this is not so hard actually.

d-torrance: Hmm, there's something strange going on...  If I construct the map ahead of time using either the cache table or a local variable, then it just sends $x$ to 0...

```m2
i6 : lift(x_F, S)

o6 = 0

o6 : S

i7 : lift(1/(x - 10), S)

     -1
o7 = --
     10

o7 : S
```

d-torrance: Never mind -- I was defining the map too early, before the local ring was properly set up.

M2/Macaulay2/packages/LocalRings/tests.m2
```diff
@@ -625,12 +625,14 @@ end--
 ///
 
 TEST ///
--- promoting to fraction field
+-- promoting/lifting to/from fraction field
 R = QQ[x]
 p = ideal x
 S = R_p
```
mahrud: Sorry this is nitpicky but could you rename the rings? Perhaps S for polynomial ring (symmetric algebra) and R for the local ring.

d-torrance: Done

mahrud (approved):

PR #4261, opened 2026-05-06
Make static linking against Boost optional

jkyang92:
Gentoo doesn't have the static libraries for Boost installed, so setting `Boost_USE_STATIC_LIBS ON` breaks the build. This lets us select if we want to force linking against the static version of Boost. I set it to on by default since that's what was already there and it's probably a better to link statically when possible since Boost updates often break dynamic linking.

Related, but not in this change: the `cmake_policy(SET CMP0167 OLD)` line shouldn't be necessary for any Boost version after 1.70. Are there any systems where this matters still?


d-torrance (approved):
Thanks!

Issue #62, opened 2013-11-19
overriding methods implemented internally


antonleykin:
It appears that one can override methods implemented internally, but can't use them.

```
toList Matrix := M -> flatten entries M
M = matrix{{1}}
class M
toList M

stdio:4:1:(3): error: expected list, sequence, or set
```

mahrud:
Why is this not easily doable?
```m2
i1 : toList Matrix := M -> flatten entries M

o1 = -*Function[stdio:1:20-1:39]*-

o1 : FunctionClosure

i2 : lookup(toList, Matrix)

o2 = -*Function[stdio:1:20-1:39]*-

o2 : FunctionClosure

i3 : toList matrix{{0}}
stdio:4:1:(3): error: expected list, sequence, or set
```
Maybe there can be something like
```m2
toList = method(InternalFunction => toList)
```
which calls `methods toList` (which are "fake" methods from `typicalvalues.m2`) and replaces each of those with an actual method function.

kellerlv:
This appears to be fixed? Tested on **M2 1.26.06**, running the reproduction from the original report verbatim:

```
i1 : toList Matrix := M -> flatten entries M

o1 = FunctionClosure[stdio:1:17-1:39]

o1 : FunctionClosure

i2 : M = matrix{{1}}

o2 = | 1 |

              1       1
o2 : Matrix ZZ  <-- ZZ

i3 : class M

o3 = Matrix

o3 : Type

i4 : toList M

o4 = {1}

o4 : List
```

Rather than erroring, `toList M` now returns `{1}`.

To confirm this reflects the user-installed method actually being dispatched, here are two checks:

**Without the override**, `toList` still rejects a matrix, so there is no native handling:

```
i1 : M = matrix{{1}}

i2 : toList M
stdio:2:6:(3):[1]: error: expected a list, sequence, or iterable object
```

**With an override returning a distinctive value**, that value comes back, so the method table is being consulted:

```
i1 : toList Matrix := M -> "OVERRIDE-WAS-USED"

i2 : M = matrix{{1}}

i3 : toList M

o3 = OVERRIDE-WAS-USED
```

So overriding a method implemented internally now works as the report expected. I wasn't able to identify the PR that fixed it; it predates 1.26.05 as far as I can tell.

d-torrance:
It's `toList` specifically that was turned into a method at some point - compiled functions in general still can't be overridden

PR #4262, opened 2026-05-07
Fix / and \ methods for Command objects

d-torrance:
We were trying to run `apply` directly, but `Command` objects are really lists containing functions, not functions themselves, so that won't work out of the box.

This fixes an issue mentioned in Zulip by @pzinn.

### DIFF
--- a/M2/Macaulay2/m2/classes.m2
+++ b/M2/Macaulay2/m2/classes.m2
@@ -86,17 +86,17 @@ Command Thing := (x,y) -> x#0 y
 -- Now some extra stuff:
 
 apply(Thing, Command)   := VisibleList => (v,f) -> apply(v, i -> f i)
-Command   \ VisibleList :=
+Command   \ VisibleList := VisibleList => (f,v) -> apply(v,f#0)
 Function  \ VisibleList := VisibleList => (f,v) -> apply(v,f)
-Command   \ String      :=
+Command   \ String      := Sequence    => (f,s) -> apply(s,f#0)
 Function  \ String      := Sequence    => (f,s) -> apply(s,f)
-Command  \\ Thing       := 
+Command  \\ Thing       :=
 Function \\ Thing       := VisibleList => (f,v) -> f v
-       List /  Command  :=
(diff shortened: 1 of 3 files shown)


M2/Macaulay2/m2/classes.m2
```diff
@@ -86,17 +86,17 @@ Command Thing := (x,y) -> x#0 y
 -- Now some extra stuff:
 
 apply(Thing, Command)   := VisibleList => (v,f) -> apply(v, i -> f i)
```
pzinn: what's this line about, then?

pzinn: oh. answering myself. `apply` is not actually a method (ugh) so this line doesn't work.

d-torrance: Ooh, nice catch.  `apply` is a compiled function, so while the interpreter let's us "install" methods for it, they don't actually work.  (See #62).

All that happens is we have a new key/value pair on the youngest of the types, and maybe that's useful for documentation and calling `methods`.  But in this case, it should probably just be removed.

pzinn: wow... an open issue from 2013...

pzinn (commented):
assuming the tests pass, all good

M2/Macaulay2/m2/classes.m2
```diff
@@ -85,18 +85,17 @@ Command Thing := (x,y) -> x#0 y
 
 -- Now some extra stuff:
 
-apply(Thing, Command)   := VisibleList => (v,f) -> apply(v, i -> f i)
```
mahrud, line 88: Why remove this?

d-torrance: It doesn't work:

```m2
i1 : apply({rank, ideal}, help)
stdio:1:5:(3):[2]: error: expected argument 2 to be a function
```

mahrud: Oh because `apply` is a compiled function, gotcha. One day ...

PR #4239, opened 2026-04-30
Bump RInterface to v0.2

d-torrance:
Changes include:

* Release under GPL
* Add hash table <-> environment conversion
    ```m2
    i3 : RObject hashTable {"x" => 6, "y" => 7}

    o3 = <environment: 0x78f17a21eef0>

    o3 : RObject of type environment

    i4 : value oo

    o4 = HashTable{"x" => 6}
                   "y" => 7

    o4 : HashTable
    ```
* Add `RValue` method for running R code given as a string.
  ```m2
   i2 : RValue "pnorm(1.96)"

   o2 = [1] 0.9750021

   o2 : RObject of type double
   ```
* Add `RContext` class for running R code given as a string, but using a common environment (based on the `PythonContext` class in the *Python* package)
    ```m2
    i11 : ctx = RContext "x = 5"
    
    o11 = ctx
    
    o11 : RContext
    
    i12 : ctx "x + 2"
    
    o12 = [1] 7
    
    o12 : RObject of type double
    
    i13 : listSymbols ctx
    
    o13 = symbol  class   value
          ------  -----   -----
          x       double  [1] 5
    
    i14 : use ctx
    
    i15 : x
    
    o15 = [1] 5
    
    o15 : RObject of type double
    ```
* Add support for ~ as a binary operator for constructing formulas
    ```m2
    i22 : RQuote "y" ~ RQuote "x"
    
    o22 = .Primitive("~")(y, x)
    
    o22 : RObject of type language
    ```
* Improved documentation and tests
* Set up R in the GitHub builds so the package is properly tested.

### AI Disclosure

Claude Code did a first pass of the updates to the documentation and tests.

PR #4263, opened 2026-05-07
Fix a couple RInterface bugs

d-torrance:
This is a quick follow up to #4239:

* Fix `net(RObject)` for R symbols (we weren't calling `value`, so we were getting `ForeignObject` of type `char*` rather than an M2 string...)
* Use the R top-level `new.env` rather than the C API `R_NewEnv` for constructing environments.  The latter isn't available on older versions of R, and so the PPA builds were failing in Ubuntu 18.04 and 20.04.

*TODO*:  `net(RObject)` is broken in Ubuntu 18.04, but the tests all pass, and the likelihood that someone is going to use this package on such an old system is vanishly small, so I'm not going to worry about it before the release


d-torrance:
Discussed with Anton and Mike

Issue #4232, opened 2026-04-29
three bugs in html(TO2)


mahrud:
Here is the extraction of the two bugs related to the conversation on Zulip.

1. Location of package nodes should be trivially resolved, without loading the package or database.
```m2
i1 : debug Core; notify = true; html TO "Probability :: Probability"
 -- loading Probability.m2
 -- opening database ../linuxbrew/.linuxbrew/Cellar/macaulay2/1.25.11_4/lib/Macaulay2/Probability/cache/rawdocumentation-dcba-8.db
 -- beginDocumentation: using documentation database, skipping the rest of /home/linuxbrew/.linuxbrew/Cellar/macaulay2/1.25.11_4/share/Macaulay2/Probability.m2
 -- package "Probability" loaded
 -- loaded /home/linuxbrew/.linuxbrew/Cellar/macaulay2/1.25.11_4/share/Macaulay2/Probability.m2

o3 = <a title="basic probability functions" href="/home/linuxbrew/.linuxbrew/share/doc/Macaulay2/Probability/html/index.html">Probability</a>
```

2. Location of other nodes should be resolved using the dictionary alone, and only load the package documentation if the dictionary is not present.
```m2
i1 : debug Core; notify = true; html TO "Probability :: density"
 -- loading Probability.m2
 -- opening database ../linuxbrew/.linuxbrew/Cellar/macaulay2/1.25.11_4/lib/Macaulay2/Probability/cache/rawdocumentation-dcba-8.db
 -- beginDocumentation: using documentation database, skipping the rest of /home/linuxbrew/.linuxbrew/Cellar/macaulay2/1.25.11_4/share/Macaulay2/Probability.m2
 -- package "Probability" loaded
 -- loaded /home/linuxbrew/.linuxbrew/Cellar/macaulay2/1.25.11_4/share/Macaulay2/Probability.m2

o3 = <a title="probability density (or mass) function"
     href="/home/linuxbrew/.linuxbrew/share/doc/Macaulay2/Probability/html/_
     density.html">density</a>
```

3. After the above, User's private dictionary should not be polluted:
```m2
i4 : peek User#"private dictionary"

o4 = GlobalDictionary{"density" => density}
```
(I'm pretty sure this isn't the fault of Probability here, it's just an example)

The first two are basically #786, which should alleviate #1643 once fixed, but they're not totally related. #4228 is also related, but independent.

cc: @pzinn

mahrud:
This looks good now:
```m2
[...]

PR #4266, opened 2026-05-07
fixed package node links and headline to not load packages unnecessarily

mahrud:
Fixes the first two parts of #4232.


mahrud:
@d-torrance is there an easy way to tell why there are two broken links without building the entire thing with autotools?
```
./Macaulay2Doc/html/___Package.html:187:84: error: broken link: ../../User/html/index.html
./Macaulay2Doc/html/_get__Symbol.html:71:290: error: broken link: ../../User/html/index.html
```

mahrud:
Oh, never mind, it's the `User (missing documentation)` errors, which are now links to nowhere. Now sure what should be done about those.

mahrud:
@d-torrance I think this is working, but it's worth checking for unintended consequences just in case.

d-torrance (approved):
This is great!  This cut `elapsedTime html help "packages provided with Macaulay2"` on my system from 66.6296s down to .0876638s!

mahrud:
This was the whole point of adding `readPackage` to get the headline for that page without loading literally every single package.

PR #4268, opened 2026-05-08
gfan-0.8beta

antonleykin:
We build gfan-0.8 with
* [x] cmake --- on MacOS we force`gcc` instead of `clang`
* [x] autotools

Lesson learned: autotools and cmake builds use different philosophy (at least in the aspect of building **programs**): 
* autotools wants to build the correct versions of programs if old versions are found on the system
* cmake simply passes this on to checks in packages (so that the build fails when packages discover that versions are too old)


M2/cmake/check-libraries.cmake
```diff
@@ -228,6 +228,15 @@ find_program(4TI2	NAMES	circuits 4ti2-circuits 4ti2_circuits)
...
 find_program(GFAN	NAMES	gfan)
+if(GFAN)
+  execute_process(COMMAND ${GFAN} _version OUTPUT_VARIABLE GFAN_VERSION_OUTPUT ERROR_QUIET)
+  string(REGEX MATCH "gfan([0-9]+\\.[0-9]+[a-z]*)" _gfan_match "${GFAN_VERSION_OUTPUT}")
+  set(GFAN_VERSION "${CMAKE_MATCH_1}")
+  if(GFAN_VERSION VERSION_LESS "0.8")
+    message(STATUS "Found gfan version ${GFAN_VERSION}, but >= 0.8 is required; will build from source")
+    unset(GFAN CACHE)
+  endif()
+endif()
```
mahrud: Why check the version here, as opposed to in `gfanInterface` using the `MinimumVersion` option of `findProgram`, as we currently do?

M2/cmake/build-libraries.cmake
```diff
@@ -837,22 +837,27 @@ _ADD_COMPONENT_DEPENDENCY(programs cohomcalg "" COHOMCALG)
 # https://users-math.au.dk/~jensen/software/gfan/gfan.html
 # gfan needs cddlib and is used by the packages gfanInterface and StatePolytopes
 # TODO: would gfan benefit from enabling the USEFACTORY option?
+# gfan 0.8beta requires GCC (does not build with clang)
+find_program(GFAN_CXX NAMES g++-15 g++-14 g++-13 g++-12
+  HINTS /opt/homebrew/bin /usr/local/bin)
+if(NOT GFAN_CXX)
+  message(FATAL_ERROR "GCC (g++) is required to build gfan but was not found")
+endif()
```
mahrud: Am I misreading this, or does this force the entire build process to fail if `g++` of one of these hardcoded versions isn't found?! I'm strongly opposed to this.

M2/cmake/check-libraries.cmake
d-torrance: I think a version check at the beginning is good so the build doesn't fail an hour later when we get to installing the package

d-torrance: That being said, I'm not convinced we should require 0.8 -- everything in M2 still works just fine with gfan 0.6

M2/cmake/build-libraries.cmake
d-torrance: I agree -- this check should only fire if we're definitely building gfan.  If we found it already on the system and don't need to build it, then who cares what versions of GCC we have.

M2/cmake/check-libraries.cmake
mahrud: Are we going to do this with all other software that requires a particular version? Even if so, it's not supposed to be done like this. CMake has caching machinery for dependencies so you don't have to hardcode a message like `Found gfan version ${GFAN_VERSION}, but >= 0.8 is required; will build from source`

antonleykin: Doug is right in that we want to treat gfan as a required component. The 0.6 would be insufficient very soon, also my understanding is that 0.8 is much more efficient.

As to a better cmake solution, Mahrud, please suggest. (We do need to fail at build, not at package level.)

M2/cmake/build-libraries.cmake
antonleykin: This has been made conditional. 
Mahrud, do you write "am I misreading this" in AI prompts as well, or this is reserved for humans? :smile:

M2/cmake/check-libraries.cmake
d-torrance: I'd rather wait on requiring 0.8 until we actually need it.  It's only in beta right now and won't be available in package managers.

mahrud: > I think a version check at the beginning is good so the build doesn't fail an hour later when we get to installing the package

I often build M2 just to get to the binary, because none of the external programs are necessary for my needs. Why should the build fail for me?

You can have a single M2 script that runs when tvalues.m2 is generated that checks all versions using findProgram and prints a warning if you'd like, but I see no reason to check before then, let alone a build failure.

mahrud: By the way, feel free to adjust the autotools build however you'd like, but the cmake build is geared towards engine development and gfan is not a required component of the engine, hence why you don't get an error for any missing library or program except for really low level things like blas or gmp.

mahrud:
gfan-0.8beta is now available through homebrew: https://github.com/Macaulay2/homebrew-tap/pull/338

M2/cmake/check-libraries.cmake
mahrud: The changes in `build-libraries.cmake` look good to me now, but I believe the changes here to `check-libraries.cmake` should be reverted, since `gfanInterface` can check version if necessary.

mahrud (changes requested):

antonleykin:
After disucssions with @mahrud and @d-torrance, we came to conclusion that a clean way going forward is to _not_ patch `gfan` and let it build itself (both on ubuntu and MacOS) if a sufficient version is not installed. In most cases, `gfan` is installed (e.g., via homebrew on a Mac).

M2/cmake/check-libraries.cmake
mahrud: @antonleykin this was my review and requested changes ^^

.github/workflows/test_build.yml
```diff
@@ -175,7 +175,7 @@ jobs:
           ../../autogen.sh
           export F77=gfortran-14
           export FC=gfortran-14
-          ../../configure --with-system-gc --with-fplll
+          ../../configure --with-system-gc --with-fplll --enable-build-libraries=gfan
```
d-torrance: It would be good if this line is removed before we merge, but I need to backport the Ubuntu gfan package first...

d-torrance: Done - the Github builds should now grab the latest gfan from the PPA. Since its also on the homebrew tap, there's no reason to build it ourselves on every build

M2/Macaulay2/packages/gfanInterface.m2
```diff
@@ -1013,9 +1013,9 @@ runGfanCommandCaptureBoth = (cmd, opts, data) -> (
 	if gfanProgram === null then
 	    gfanProgram = findProgram("gfan", "gfan --help",
 		Verbose => gfanVerbose,
-		-- version 0.6 is necessary for gfanMixedVolume
-		-- https://github.com/Macaulay2/M2/issues/1962
-		MinimumVersion => ("0.6",
+		-- version 0.8 is required (the 0.6.2 series predates several
+		-- features used here and is no longer supported upstream)
+		MinimumVersion => ("0.8",
```
mahrud, line 1018: FYI Fedora ships with gfan 0.7. Is this also insufficient? (I understand that there are new features in 0.8, but those aren't being used in gfanInterface yet, right?)

mahrud:
Just to double check: is the current plan squash-merging this commit once it doesn't force build gfan during github builds? I recall discussing at some point that we didn't want just-@claude-authored commits.

M2/Macaulay2/packages/gfanInterface.m2
d-torrance: For our Fedora packages, 0.8 will be included.  And I imagine that the Fedora maintainer will likely package gfan 0.8 soon for the official Fedora packages

mahrud: I don't think they ship beta packages, but idk.

d-torrance:
Yes, the Claude commits were removed

Issue #4267, opened 2026-05-07
Equality of visible lists


d-torrance:
Right now we have:

```m2
i1 : {1, 2, 3} == (1, 2, 3)
stdio:1:10:(3):[1]: error: expected argument 2 to be a visible list

i2 : {1, 2, 3} == [1, 2, 3]

o2 = false
```
After discussing this while looking at #4242 with Anton and Mike, they should both be errors.

d-torrance:
> **Written by Claude** (Claude Opus 5, via Claude Code), not by @d-torrance, whose account posted it -- please weigh it accordingly.

Done, in [#4269](https://github.com/Macaulay2/M2/pull/4269) — commit
[`75771ac107`](https://github.com/Macaulay2/M2/commit/75771ac107), *"Raise an error when comparing
lists of different types"*, merged 2026-05-26. It never said `Closes #4267`, which is the only
reason this is still open.

Both cases now behave the way Anton and Mike wanted, with one message rather than the old
asymmetry:

```m2
i1 : {1, 2, 3} == (1, 2, 3)
stdio:1:10:(3):[1]: error: expected lists of the same class

i2 : {1, 2, 3} == [1, 2, 3]
stdio:2:10:(3):[1]: error: expected lists of the same class
```

The check is in `m2/lists.m2:60` and `d/actors3.d:194-208`. Closing.

<sub>Triaged in a sweep of the open issues; source of truth: [`bug-triage/issues.tsv`](https://github.com/d-torrance/M2/blob/bug-triage/bug-triage/issues.tsv).</sub>

PR #4269, opened 2026-05-08
Raise an error when comparing lists of different types

d-torrance:
This is a response to a now-removed commit from #4242.  In it, I had proposed that `==` between lists of different types should return false, but after discussion with others, the consensus was that it should raise an error.

Currently, the behavior is mixed and the error messages are cryptic:

```m2
i1 : (1,2,3) == {1,2,3}
stdio:1:8:(3):[1]: error: expected argument 1 to be a visible list

i2 : {1,2,3} == (1,2,3)
stdio:2:8:(3):[1]: error: expected argument 2 to be a visible list

i3 : {1,2,3} == [1,2,3]

o3 = false
```

Here's the proposed behavior:

```m2
i1 : (1,2,3) == {1,2,3}
stdio:1:8:(3):[1]: error: expected lists of the same class

i2 : {1,2,3} == (1,2,3)
stdio:2:8:(3):[1]: error: expected lists of the same class

i3 : {1,2,3} == [1,2,3]
stdio:3:8:(3):[1]: error: expected lists of the same class
```

Closes: #4267

### DIFF
--- a/M2/Macaulay2/d/actors3.d
+++ b/M2/Macaulay2/d/actors3.d
@@ -96,6 +96,17 @@ override(e:Expr):Expr := (
      else WrongNumArgs(2));
 setupfun("override",override);
 -----------------------------------------------------------------------------
+EqualEqualfunpointer := dummyEE;
+listComparison(s:Sequence, t:Sequence):Expr := (
+    n := length(s);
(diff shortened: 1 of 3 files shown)


d-torrance:
Discussed with Mike

Issue #3896, opened 2025-06-26
Lifting from fraction field failure


pzinn:
this fails even though it shouldn't:
```
i1 : R=QQ[x];

i2 : F=frac R;

i3 : lift(1/123_F,R)
stdio:3:4:(3): error: cannot lift given ring element

i4 : lift(1/123_F,QQ)
stdio:4:4:(3): error: cannot lift given ring element
```
This can be traced back to `frac.cpp`:
```
bool FractionField::lift(const Ring *Rg,
                         const ring_elem f,
                         ring_elem &result) const
{
  // Rg = R ---> frac R
  // f is an element of frac R.

  ring_elem
      hdenom;  // used in the case when the denominator can be a unit, but not 1
               // e.g. when this = frac (QQ[x,y,z]).  Is an element of
  if (Rg == R_)
    {
      frac_elem *h = FRAC_VAL(f);
      if (R_->is_equal(h->denom, R_->one()))
        {
          result = R_->copy(h->numer);
          return true;
        }
      else
        {
          if (R_->is_field())
            {
              // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
              // try to lift denominator.  If can, can lift, otherwise not.
              if (R_->lift(R_, h->denom, hdenom))
                {
                  ring_elem hinv = R_->invert(hdenom);
                  result = R_->mult(hinv, h->numer);
                  return true;
                }
            }
        }
    }
  return false;
}
```
I suspect the offending line is ` if (R_->is_field())` which is always false (if R_ is a field, then it's its own fraction field).
Removing it seems to fix the issue, though I'm not sure exactly what's going on. (it should probably test for invertibility of the denominator)

d-torrance:
Fixed by #4270

PR #4270, opened 2026-05-08
fix lift from fraction field to base/coefficient ring (#3896)

pzinn:
FractionField::lift was gated on R_->is_field(), which is always false for the polynomial ring underlying a fraction field.  Replace the guard with R_->is_unit(h->denom), and chain through R_->lift so e.g. lift(1/123_(frac QQ[x]), QQ) also works.

# AI Disclosure
Claude Code helped me write this.


pzinn:
this is ready to be merged

mahrud:
Will it conflict with the monster engine PR?

pzinn:
how would I know? I submitted this before any of this

mahrud:
I just meant is it changing the same files as that PR? It's okay if you don't know, it's a reminder for whoever is going to merge to check first.

d-torrance:
Pretty much every file in the engine is getting moved :)  I think the plan is to merge the "great sort" PR first (#4347) and then rebase the others (including this one) after that.

M2/Macaulay2/e/rings/frac.cpp
```diff
@@ -326,36 +326,30 @@ bool FractionField::lift(const Ring *Rg,
...
-                  result = R_->mult(hinv, h->numer);
-                  return true;
-                }
-            }
-        }
+      result = g;
+      return true;
     }
-  return false;
+  return R_->lift(Rg, g, result);
```
d-torrance, line 352: This file was moved to `M2/Macaulay2/e/rings/frac.cpp`

d-torrance (changes requested):

d-torrance (approved):
Reviewed with @local-ring.  Looks good!

PR #4272, opened 2026-05-08
fix net BR, HR

### LINKED ISSUE #2754: <br/> not respected within certain HTML elements in conversion to Net

pzinn:
(this is *not* urgent; feel free to ignore for upcoming release)
This fixes #2754, in particular.
Before:
the first two have incorrect `net`:
Screenshot from 2026-05-09 01-49-30
here it's the `html` of the first two that's wrong:
Screenshot from 2026-05-09 02-02-48

After:
Screenshot from 2026-05-09 01-49-47
Screenshot from 2026-05-09 02-03-54

### DIFF
--- a/M2/Macaulay2/m2/format.m2
+++ b/M2/Macaulay2/m2/format.m2
@@ -31,7 +31,16 @@ setupRenderer = (parser, joiner, T) -> (
 -- Default joiners: (TODO: move to string.m2?)
 -- concatenate
 -- horizontalJoin
-wrapHorizontalJoin := x -> wrap horizontalJoin x
+net BR := info BR := x -> stack()
+net HR := info HR := x -> concatenate(printWidth:"-")
+
+horizontalJoin' := x -> ( -- horizontalJoin except for BRs and HRs
+    netBR := net BR{}; netHR := net HR{};
+    x' := sublists(toList x, y -> y=!=netBR and y=!=netHR, toList, y -> {y});
+    stack(horizontalJoin\x')
+    )
+
+wrapHorizontalJoin := x -> wrap horizontalJoin' x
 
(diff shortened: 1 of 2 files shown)


d-torrance (approved):

M2/Macaulay2/m2/format.m2
```diff
@@ -31,7 +31,16 @@ setupRenderer = (parser, joiner, T) -> (
 -- Default joiners: (TODO: move to string.m2?)
 -- concatenate
 -- horizontalJoin
-wrapHorizontalJoin := x -> wrap horizontalJoin x
+net BR := info BR := x -> stack()
+net HR := info HR := x -> concatenate(printWidth:"-")
```
mahrud, line 35: Did you only move this line up?

pzinn: `net BR` was undefined, added it. I moved `net HR` up for clarity

pzinn:
BTW -- at some point in the future I want to rewrite `format.m2` completely and simplify massively the way `net Hypertext` works. This PR isn't it -- it's just a quick fix for a problem that's been annoying me for a long time and been blocking me in various ways.

PR #4277, opened 2026-05-09
Added progress bars

seangrate:
This implements [`tqdm`](https://tqdm.github.io/)-style progress bars to display progress for common iterables such as iterators, lists, and sets. In this style, as an iterable is exhausted, visual progress is shown (if the length is known) and some time statistics are also calculated.

bars-demo

For lists and sets, the total number of iterations is known. For Iterator objects, this is no longer the case, so a `TotalIterations` optional argument can be passed so that accurate time statistics can be displayed like normal.
```
for throwaway in progressBar(someIterable, TotalIterations=>100) do (
    -- Some computation...
)
```

Other customization is available as well, namely in a description for the progress bar and in changing the bar itself.

bars-demo-v2

I've put the code and documentation with the `Iterator` stuff as it seemed fitting. 

**Caveat**: At the moment, the progress bars really only work with the `for ... in ...` construction, so if there is a way to have them be compatible with `apply` (for instance, as `progressBar apply(someIterable, someFunction)` or `apply(progressBar someIterable, someFunction)`, that would be awesome.


mikestillman:
I would love to be able to do something like this for a GB computation...

seangrate:
I think I'm a bit stuck in trying to fix these errors and I feel like this is stemming from some sort of symbol-exporting  issue (how exactly Macaulay2 handles symbols and variables hasn't quite clicked with me yet). Any suggestions?

M2/Macaulay2/packages/Macaulay2Doc/doc_iterators.m2
```diff
@@ -155,3 +155,101 @@ doc ///
...
+      to track progress accurately. However, for an @TO Iterator@ object, its
+      length is not necessarily known. In this case, the progress bar's visual
+      bar will not update, though the elapsed time will still update.
+    Example
+      for throwaway in progressBar iterator(1..25) do (
+        -- A nontrivial computation
+          for i from 0 to 1000000 do 223131321321*223131321324
+      )
+    Text
+      To bypass this, the @TT TotalIterations@ optional argument is available
```
joel-dodge: I dug into this a little bit because I am learning my way around the various parts of macaulay2 and I also love the user experience of progress bars :D

I believe the errors in your build are coming from lines like this where you pass a `Symbol` directly to `TT` which expects to be passed a string (or possible some other type of gadget that knows how to be rendered as hypertext but I think here you probably intend for it to be a string). 

On the other hand, if you want to have a link and there is a doc node  for the key `TotalIterations` then you can use `@TO TotalIterations@` to achieve this.

Whatever is inside the `@...@` signs is macaulay2 code that is executed to produce a string.

joel-dodge: My understanding of the use of `@...@` in docs comes from reading https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/SimpleDoc/html/_doc.html

seangrate: Ahhh I deliberately wanted to use `@TT ... @` as opposed to linking; I just mindlessly forgot that it has to be enclosed with quotes. Thanks!

M2/Macaulay2/packages/Macaulay2Doc/doc_iterators.m2
```diff
@@ -155,3 +155,101 @@ doc ///
...
+          -- A nontrivial computation
+          for i from 0 to 100 do 22313*22131
+      )
+    Text
+      For lists and sets, the length of the iterable is known and can be used
+      to track progress accurately. However, for an @TO Iterator@ object, its
+      length is not necessarily known. In this case, the progress bar's visual
+      bar will not update, though the elapsed time will still update.
+    Example
+      for throwaway in progressBar iterator(1..25) do (
```
d-torrance, line 225: When I run this example locally, this one hangs.  I added some print statements and `currentIndex` never changes from 0

M2/Macaulay2/m2/iterators.m2
```diff
@@ -73,3 +73,185 @@ pairsIterator = x -> Iterator (
 	y := next iter;
 	if y === StopIteration then StopIteration
 	else (i, (i += 1; y))))
+
+ProgressBar = new Type of MutableHashTable
+ProgressBar.synonym = "progressBar"
```
d-torrance, line 78: I think "progress bar" would be more consistent with other synonyms

d-torrance (changes requested):

Issue #921, opened 2019-01-29
lift breaks


DanGrayson:
see https://groups.google.com/forum/#!topic/macaulay2/QowiaTEHIqo

radoslavraynov:
I see at least two separate problems. The first is that `R=CC[]; lift(1_R,QQ)` breaks because `RR` is on the `R.baseRings` path, not `default RR` (or better, the one with the precision of CC). The second is that `R=RR[]; lift(1_R,QQ)` fails because it calls repeatedly `r = rawLift(B, r)` for `B` in the reversed base ring list, but the corresponding case in `ConcreteRing<RingType>::lift()` in *e/aring-glue.hpp* isn't implemented. Instead `lift(1.0, QQ)` works because it's implemented directly in *m2/reals.m2*.

Any suggestions what the path of least resistance is? Thanks in advance!

d-torrance:
Fixed in #4280

Issue #2509, opened 2022-05-19
Crash when lifting from RR to ZZ


pzinn:
one-line crash:
```
i1 : lift( matrix {{1.}}, ZZ)
error!! lift called with no ZZ lifting method
-- SIGSEGV
-* stack trace, pid: 632942
 0# std::vector<boost::stacktrace::frame, std::allocator<boost::stacktrace::frame> >::size() const at /usr/include/c++/11/bits/stl_vector.h:919
 1# segv_handler at /home/pzinn/M2/M2/Macaulay2/bin/main.cpp:244
 2# 0x00007FCE3FFF4750 in /lib64/libc.so.6
 3# RingZZ::is_zero(ring_elem) const at /home/pzinn/M2/M2/Macaulay2/e/ZZ.cpp:149
 4# Ring::set_entry(vecterm*&, int, ring_elem) const at /home/pzinn/M2/M2/Macaulay2/e/ring-vecs.cpp:605
 5# Matrix::iterator::next() at /home/pzinn/M2/M2/Macaulay2/e/matrix.hpp:254
 6# interface_rawLift at /home/pzinn/M2/M2/Macaulay2/d/interface.dd:1461
 7# evaluate_evalraw at /home/pzinn/M2/M2/Macaulay2/d/evaluate.d:1303
 8# evaluate_evalraw at /home/pzinn/M2/M2/Macaulay2/d/evaluate.d:1263
 9# evaluate_evalraw at /home/pzinn/M2/M2/Macaulay2/d/evaluate.d:1263 (discriminator 2)
10# evaluate_applyFCCS at /home/pzinn/M2/M2/Macaulay2/d/evaluate.d:839 (discriminator 5)
11# evaluate_evalraw at /home/pzinn/M2/M2/Macaulay2/d/evaluate.d:1299
12# evaluate_applyFCS at /home/pzinn/M2/M2/Macaulay2/d/evaluate.d:529 (discriminator 1)
13# evaluate_evalraw at /home/pzinn/M2/M2/Macaulay2/d/evaluate.d:1307
14# evaluate_applyFCS at /home/pzinn/M2/M2/Macaulay2/d/evaluate.d:472 (discriminator 1)
15# iteratedApply at /home/pzinn/M2/M2/Macaulay2/d/actors3.d:2247 (discriminator 6)
16# evaluate_evalraw at /home/pzinn/M2/M2/Macaulay2/d/evaluate.d:1303
[...]

Issue #1812, opened 2021-01-12
Lifting scalars from polynomial rings


jchen419:
Often it is desirable to view a scalar in a polynomial ring as an actual scalar (e.g. in conjunction with `coefficients`), but this can sometimes be difficult to accomplish for the finite precision field families:
```
i1 : R = RR[x]; lift(1_R, QQ)
stdio:1:12:(3): error: cannot lift given ring element

i3 : R = CC[x]; lift(1_R, QQ)
stdio:2:12:(3): error: no method found for applying raw to:
     argument   :  RR (of class InexactFieldFamily)
```
One workaround is to first lift to the `coefficientRing` of the polynomial ring, and then do a further lift (to `QQ` here), but it would be good to have the displayed commands just work (not to mention that the error messages are not so helpful).

d-torrance:
Fixed in #4280

PR #4245, opened 2026-05-01
InvariantRing v2.4

galettof:
`InvariantRing` has been updated to version 2.4, which includes a new algorithm for invariants of elementary abelian $p$-groups, as well as bugfixes and documentation improvements. The `permutationMatrix` method now takes arrays as inputs.

PR #4280, opened 2026-05-10
Improve lifting from RR to QQ

d-torrance:
Currently, we implement lifting `RR` from `QQ` using continued fraction approximation.  It works only for lifting `RR` objects to `QQ` -- no matrices or polynomials, and it's slow because it's written at top-level.

We move this code to the engine, speeding things up and giving us lifting of polynomials and matrices for free.  We also get lifting to `ZZ` for free for most things (not matrices...), and lifting from `RRi` and `CC` to `QQ` for free.  (Not `CCi` though ... not sure why.)

Lifting matrices to `ZZ` used to segfault.  It still doesn't work, but now it doesn't segfault!

### Before

```m2
i1 : lift(1_(RR[x]), QQ)
stdio:1:4:(3):[1]: error: cannot lift given ring element

i2 : lift(1_(CC[x]), ZZ)
stdio:2:4:(3):[1]: error: cannot lift given ring element

i3 : lift(matrix(RR, 1), QQ)
stdio:3:4:(3):[1]: error: cannot lift given matrix

i4 : lift(matrix(RR, 1), ZZ)
error!! lift called with no ZZ lifting method
-- SIGSEGV
-* stack trace, pid: 23779
 0# profiler_stacktrace(std::ostream&, int) at /usr/src/macaulay2-1.25.11+git202604300057-0ppa202604290114~ubuntu24.04.1/M2/Macaulay2/d/main.cpp:132
 1# segv_handler at /usr/src/macaulay2-1.25.11+git202604300057-0ppa202604290114~ubuntu24.04.1/M2/Macaulay2/d/main.cpp:248
 2# 0x0000749623045330 in /lib/x86_64-linux-gnu/libc.so.6
 3# RingZZ::is_zero(ring_elem) const at /usr/src/macaulay2-1.25.11+git202604300057-0ppa202604290114~ubuntu24.04.1/M2/Macaulay2/e/ZZ.cpp:158
 4# Ring::set_entry(vecterm*&, int, ring_elem) const at /usr/src/macaulay2-1.25.11+git202604300057-0ppa202604290114~ubuntu24.04.1/M2/Macaulay2/e/ring-vecs.cpp:606
 5# IM2_Matrix_lift at interface/matrix.cpp:804
```

### After

```m2
i1 : lift(1_(RR[x]), QQ)

o1 = 1

o1 : QQ

i2 : lift(1_(CC[x]), ZZ)

o2 = 1

i3 : lift(matrix(RR, 1), QQ)

o3 = | 1 |

              1       1
o3 : Matrix QQ  <-- QQ

i4 : lift(matrix(RR, 1), ZZ)
stdio:4:4:(3):[1]: error: cannot lift given matrix
```

Closes: #921, #1812

Partial band-aid fix for #2509

### AI Disclosure

Most of the code was written by Claude Code, with significant hand-holding from me.

### DIFF
--- a/M2/Macaulay2/e/basic-rings/aring-QQ-gmp.hpp
+++ b/M2/Macaulay2/e/basic-rings/aring-QQ-gmp.hpp
@@ -9,6 +9,7 @@
 #include "basic-rings/aring.hpp"
 #include "buffer.hpp"
 #include "rings/ringelem.hpp"
+#include <cmath>
 #include <iosfwd>
 #include "exceptions.hpp"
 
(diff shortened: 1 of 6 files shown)


pzinn:
why can't we lift matrices to `ZZ`?

d-torrance:
No one's written the code yet :)

d-torrance:
@galettof: The builds for this PR started failing because of #1773.  The offending test had been skipped for a few years because this issue kept coming up, but it looks like it got un-skipped in #4245. So I pushed a commit to start skipping it again until the underlying issue is fixed.

galettof:
When I pushed the update to InvariantRing 2.4, I started from the most recent version in my repo. I wonder if that version did not have the previous fix to the issue mentioned here. I can check and report back.

galettof:
A quick update: test 9 was commented out in the previous release, so that's why it didn't fail. :sweat_smile: 
The PR for version 2.4 uncommented it, hence the new failures. You are welcome to comment it out again. If you do, please let me know so I can do the same in my repo.

Having reviewed the discussion of #1773, I am going to suggest the issue originates outside of InvariantRing. Something about hash codes depending on the architecture? If so, this is beyond my ability to resolve. If there is a way to change the code in the package to fix the issue, maybe we can work on it at the next workshop, but that wouldn't stop the issue appearing again elsewhere.

d-torrance:
The proposed band-aid fix to the package that you may want to commit locally is here: https://github.com/Macaulay2/M2/pull/4280/changes/1a55fb7869eb6fa8fc3a815a4a48212ca2d94fc7.  Rather than commenting out the entire test, it just adds a `no-check-flag` comment to the test to tell `check` to skip it.

The issue in #1773 is certainly related to hash codes, but the bug is definitely with the algorithm in the package.  We have a mutable hash table `U`, we then call `select(keys U, ...)`, and then iterate through that.  There's no guarantee that `keys U` will have any particular order, and with different orders, we get different results.  See https://github.com/Macaulay2/M2/issues/1773#issuecomment-777611786.

(The fact that that there's no guarantee on the order of the output of `keys` is where hash codes come in, and why this bug crops up when seemingly unrelated changes are made to M2.)

galettof:
Can you run code in an instance of M2 running on the offending architecture? If so, can you insert a line `print peek nonemptyU` after `nonemptyU` is first created, then run the code in the test? I am curious to see the order of keys that is causing this problem and compare with the one I get.

d-torrance:
Now that #1773 is fixed, I've removed the commit that skipped `check(9, "InvariantRing")`.

d-torrance:
Discussed with Mike.  Will merge post-Great Sort

PR #4285, opened 2026-05-10
Update copyright years for pygments lexer

d-torrance:
This way, we match the file in the pygments repo.  Rather than updating each year, they've just made the last year "present".


d-torrance:
Discussed with Mike

PR #4287, opened 2026-05-11
Update Debian package workflow

d-torrance:
Build a source package first and use that.  Also don't do a full build for the x86 package -- we just need the binary packages, and the .changes file was expecting there to be source package files to sign that weren't included in the artifacts.

Basically, I didn't actually check to see if I'd be able to upload the Debian packages this workflow created until after the release ... and I wasn't!  This fixes that, and there are now nice fresh 1.26.05 .deb files on macaulay2.com after running this updated workflow on my fork (https://github.com/d-torrance/M2/actions/runs/25637548581).

### DIFF
--- a/.github/workflows/build-deb.yml
+++ b/.github/workflows/build-deb.yml
@@ -80,14 +80,20 @@ jobs:
           cd macaulay2
           git checkout debian/${{ matrix.release }}
       - name: Generate tarball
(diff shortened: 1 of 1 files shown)


d-torrance:
Discussed with Mike

d-torrance:
Ugh -- I merged this into `stable`!

Issue #3330, opened 2024-06-25
Syntax highlighting in packages


mahrud:
A pretty glaring caveat of syntax highlighting is that it doesn't work for packages. Here's an [example](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/NormalToricVarieties/html/_make__Smooth_lp__Normal__Toric__Variety_rp.html) with only symbols from Core being highligted, which looks really odd.

A pretty simple solution would be to generate a prism grammar for each package in `installPackage` and load it in the header of html files from that package.

It would be great if the same worked for emacs.

mahrud:
There are also a few related suggestions in .https://github.com/Macaulay2/M2/pull/3298.

pzinn:
(copied from zulip) a tangentially related issue. Say one defines a new function `myfunc`, then neither in emacs nor in the webapp will the autocompletion work if one types say `myf` followed by tab. However it does work in the terminal using readline. Why can't we have something similar without readline? it can't be that hard...

pzinn:
@d-torrance's answer:
> I've played around with this a bit. We could potentially do something like call apropos to get a list of possible autocompletions, but output it in such a way that the editor doesn't display the output to the user. This is basically how autocompletion works in the Python Emacs mode
originally I was hoping we could reuse the code made for `readline` but I guess most of the work is done by `readline` itself...

pzinn:
continuing my digression, I'm assuming this is the relevant part of `expr.d`:
```
export completions(s:string):array(string) := (
     n := length(s);
     if n == 0 then return array(string)();		    -- don't complete the null string
     v := newvarstringarray(6);
     d := globalDictionary;
     while (
	  lockRead(d.symboltable.mutex);
	  foreach bucket in d.symboltable.buckets do (
	       b := bucket;
	       while true do when b
	       is null do break
	       is q:SymbolListCell do (
		    t := q.word.name;
		    if isalnum(t.0) && n <= length(t) && 0 == strncmp(s,t,n) then append(v,t);
		    b = q.next; ));
	  unlock(d.symboltable.mutex);
	  d != d.outerDictionary) do d = d.outerDictionary;
     extract(v));
```

mahrud:
It is not a digression. Syntax highlighting and autocompletion have the same purpose.

pzinn:
... progress ...

[...]

PR #4288, opened 2026-05-11
Add completion control requests

pzinn:
This is a first attempt at getting dynamic autocomplete even when `readline` is not used (e.g., in emacs or in the WebApp). cf discussion in #3330.

### DIFF
--- a/M2/Macaulay2/d/expr.d
+++ b/M2/Macaulay2/d/expr.d
@@ -108,6 +108,48 @@ export completions(s:string):array(string) := (
 	  unlock(d.symboltable.mutex);
 	  d != d.outerDictionary) do d = d.outerDictionary;
      extract(v));
+
+-- This assumes x is a valid symbol name; arbitrary strings would need format from stdio.d.
+completionJSONQuote(x:string):string := "\"" + x + "\"";
+
+completionKind(e:Expr):string := when e
+     is FunctionClosure do "function"
+     is CompiledFunction do "function"
+     is CompiledFunctionClosure do "function"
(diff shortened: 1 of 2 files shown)


pzinn:
590094210-f4b79335-d6de-4eb3-869d-28816655a74c

d-torrance:
This looks cool!  What's the interface that the editor would use to alert M2 that it wants some completions?  It looks like maybe it sends a special string to the interpreter?  And how are the possible completions returned?

mahrud:
Is this using an existing protocol or is it specific to M2?

pzinn:
I don't know of any existing protocol to interact at the d level.
An obvious idea to implement this is to do it by exposing `completions` at m2-level and then having the interface (emacs, client in browser) call it each time Tab is pressed. This is a terrible idea since it will mess up everything (output formatting, positioning, multi-line input, etc). That's what Claude Code immediately suggested, took 20 mins to implement with 100s of lines of code, and made a huge buggy mess.
Interestingly, when I asked codex (same prompt), it pointed out that this was *not* the way to go and instead suggested the current procedure that I ended up implementing. This is more in line with what `readline` does -- it intercepts at the d-level the completion requests. It's a bit more code at the level of M2 itself, but makes it a lot easier for the UI.
The details are subject to change, but right now the response is produced by `sendCompletionControlResponse`, so some control code, a request id to avoid issues with multiple request/delays, then a `\t`-separated list of possible completions, finally another control code.

pzinn:
the response is now JSON with some metadata

Issue #4286, opened 2026-05-10
segfault in res14.m2


d-torrance:
I'm getting the following segfault on a number of Debian/Ubuntu builds of the 1.26.05 package, mostly on 32-bit systems:

```m2
testing: res14.m2
-- SIGSEGV
-* stack trace, pid: 56061
 0# profiler_stacktrace(std::ostream&, int) at ./M2/Macaulay2/d/main.cpp:131
 1# segv_handler at ./M2/Macaulay2/d/main.cpp:246
 2# 0x000005A0 in linux-gate.so.1
 3# __gmpn_copyi_k7_mmx in /usr/lib/i386-linux-gnu/libgmp.so.10
-- end stack trace *-
Command exited with non-zero status 1
```

This is happening on every 32-bit build (armhf and i386) in Debian sid and all versions of Ubuntu since 20.04.

The Ubuntu 18.04 builds are failing on *all* architectures.  The 64-bit ones give us a bit more information about where in the code it's happening:

```m2
testing: res14.m2
          R = kk[x,y,z];
          I = ideal(x^2-y*z, x^4);
          C = res I;
          assert try ( res(ideal I_*, Strategy => Nonminimal); false ) else true;
          assert try ( res(ideal I_*, Strategy => NonminimalWithGB); false ) else true;
      	);
M2-binary: malloc.c:2868: mremap_chunk: Assertion `((size + offset) & (GLRO (dl_pagesize) - 1)) == 0' failed.
```

d-torrance:
I was able to reproduce it locally in an Debian trixie i386 schroot and got a stacktrace from gdb:

```m2
i1 : R = QQ[x,y,z]

o1 = R

o1 : PolynomialRing

i2 : I = ideal(x^2 - y*z, x^4)

             2         4
o2 = ideal (x  - y*z, x )

o2 : Ideal of R

i3 : res(I, Strategy => Nonminimal)

Thread 24 "M2-binary" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0xdc365b40 (LWP 1196444)]
0xf57a63d9 in __gmpn_copyi_k7_mmx () from /lib/i386-linux-gnu/libgmp.so.10
(gdb) bt
#0  0xf57a63d9 in __gmpn_copyi_k7_mmx () from /lib/i386-linux-gnu/libgmp.so.10
#1  0xf57a1143 in __gmpq_set () from /lib/i386-linux-gnu/libgmp.so.10
[...]

PR #4289, opened 2026-05-11
Fix segfault/error in Nonminimal res over QQ on 32-bit/older systems

d-torrance:
When res(I, Strategy => Nonminimal) is called over QQ, the computation should immediately raise an error: the strategy relies on rank computation over a finite field (rankUsingDenseMatrix constructs an ARingZZpFFPACK or ARingZZpFlint from the ring's characteristic), and characteristic 0 is invalid for that purpose.

Previously, the error was caught lazily in F4ResComputation::get_matrix, which meant res() itself would succeed and only fail when the caller tried to access a differential.  On 32-bit and older 64-bit systems (e.g. Ubuntu 18.04), a segfault in from_M2_vec fired even earlier: from_ring_elem for ARingQQGMP called mpq_set with an mpz_srcptr disguised as mpq_srcptr, causing mpq_set to read the denominator field past the end of the mpz_t object and crash.

Move the QQ check to choose_res, before createF4Res is called, so that res() raises the error immediately on all platforms.  Remove the now- unreachable guards from get_matrix, including a stray debug std::cout.

Closes: #4286 

## AI Disclosure

This one was all Claude.  I even let it write the commit message

### DIFF
--- a/M2/Macaulay2/e/resolution/comp-res.cpp
+++ b/M2/Macaulay2/e/resolution/comp-res.cpp
@@ -134,6 +134,11 @@ ResolutionComputation *ResolutionComputation::choose_res(
                 "polynomial rings");
             return nullptr;
           }
+        if (K->is_QQ())
+          {
+            ERROR("Nonminimal resolution strategy not available over QQ");
+            return nullptr;
+          }
         if (M2_gbTrace > 0) emit_line("resolution Strategy=>4 (res-f4)");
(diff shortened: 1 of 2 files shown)


d-torrance:
discussed with Mike.  Will rebase on top of the the Great Sort PR once it is merged

mikestillman (approved):
I want to fix the underlying problem here, but this is a good change, that we should or could accept before "the great sort" PR.

mikestillman:
Maybe after that PR...

PR #4291, opened 2026-05-11
Catch stray symbols in evaluateWithPackage

### LINKED ISSUE #4290: load example results from documentation database without loading package
Follow up on #4232. There are two issues in this:
```m2
i1 : debug Core; notify = true; help "Probability :: density";
 -- opening database usr-dist/x86_64-Linux-Fedora-43/lib64/Macaulay2/Probability/cache/rawdocumentation-dcba-8.db
 -- loading Probability.m2
 -- beginDocumentation: using documentation database, skipping the rest of /home/mahrud/Projects/M2/quickfix/M2/Macaulay2/packages/Probability.m2
 -- warning: symbol "density" in User#"private dictionary" is shadowed by a symbol in Probability.Dictionary
 --   use the synonym density$0
 -- package "Probability" loaded
 -- loaded /home/mahrud/Projects/M2/quickfix/M2/Macaulay2/packages/Probability.m2

i4 : peek User#"private dictionary"

o4 = GlobalDictionary{"density" => density  }
                      "density$0" => density
```

1. The package is being needlessly loaded, but everything we need is in the database file.
2. For some reason `density` ends up in user's private dictionary.

The first issue is because the package is being loaded in line:
https://github.com/Macaulay2/M2/blob/c6aff10750a8b8a7e221f04c72d3d8b90c902ea9/M2/Macaulay2/m2/help.m2#L447-L450
This is not strictly necessary, but we need the package prefix here:
https://github.com/Macaulay2/M2/blob/c6aff10750a8b8a7e221f04c72d3d8b90c902ea9/M2/Macaulay2/m2/examples.m2#L211-L216

The latter is because we call `value` on the a string containing a raw documentation hash table in the database, which may contain something like `Key => density`, hence `density` ends up a symbol in user's dictionary.

mahrud:
Closes #4290.
- **removed fetchRawDocumentationNoLoad**
- **added safety net for stray symbols**

### DIFF
--- a/M2/Macaulay2/m2/document.m2
+++ b/M2/Macaulay2/m2/document.m2
@@ -363,30 +363,22 @@ storeRawDocumentation := (tag, rawdoc) -> (
     currentPackage#rawKey#fkey = rawdoc)
 
 -----------------------------------------------------------------------------
--- fetchRawDocumentation, fetchRawDocumentationNoLoad
+-- fetchRawDocumentation
 -----------------------------------------------------------------------------
--- if a package object is given, try pkg#rawKey first, then pkg#rawKeyDB
--- if a package name is given, try its database first, then try loading the package
-getpkgdocs = pkgname -> getpkgNoLoad pkgname ?? openPackageDatabase pkgname ?? getpkg pkgname
+-- if a package is previously loaded, try pkg#rawKey first, then pkg#rawKeyDB
+getpkgdocs = pkgname -> getpkgNoLoad pkgname ?? openPackageDatabase pkgname
 
-fetchRawDocumentation = method()
-fetchRawDocumentation DocumentTag        :=  tag        -> fetchRawDocumentation(tag.Package, tag.Format)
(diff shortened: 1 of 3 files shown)


d-torrance (approved):
Thanks!

This definitely seems to fix the dictionary issue, but it looks like the package is still loading, just after the database instead of before:

```m2
i1 : debug Core; notify = true; help "Probability :: density";
 -- opening database /usr/lib/x86_64-linux-gnu/Macaulay2/Probability/cache/rawdocumentation-dcba-8.db
 -- loading Probability.m2
 -- beginDocumentation: using documentation database, skipping the rest of /home/profzoom/src/macaulay2/M2/M2/Macaulay2/packages/Probability.m2
 -- package "Probability" loaded
 -- loaded /home/profzoom/src/macaulay2/M2/M2/Macaulay2/packages/Probability.m2
```

mahrud:
Yes, this fixes part 2 but not part 1 of that issue. Part 1 requires rethinking how `processExamples` works and I need to think about it a little more.

PR #4293, opened 2026-05-11
Add (ring, ComplexIntervalField)

d-torrance:
Fixes an issue mentioned on Zulip by @klee669, for which we add a test.


d-torrance:
Discussed with Mike

PR #4300, opened 2026-05-14
Fix a bug in projdim, add functionality to verify not regularInCodimension

kschwede:
In the FastMinors package.

This fixes a bug in projDim that I had intended to do a pull request on nearly two years ago.  

This also adds functionality to verify a variety is not regular in codimension n (via the VerifyNonRegular => true option).  Appropriate documentation and tests are also added.

### DIFF
--- a/M2/Macaulay2/packages/FastMinors.m2
+++ b/M2/Macaulay2/packages/FastMinors.m2
@@ -1,5 +1,5 @@
 newPackage( "FastMinors",
-Version => "1.2.6", Date => "May 15th, 2023", Authors => {
+Version => "1.3.1", Date => "May 13th, 2026", Authors => {
     {Name => "Boyana Martinova",
     Email=> "martinova@wisc.edu",
     HomePage=> "https://sites.google.com/view/bmartinova"
@@ -18,7 +18,7 @@ Version => "1.2.6", Date => "May 15th, 2023", Authors => {
 }, --this file is in the public domain
     Headline => "faster linear algebra operations",
     PackageImports => {"Complexes"},
-    PackageExports => {"RandomPoints"},
+    PackageExports => {"RandomPoints",  "PrimaryDecomposition"},
     DebuggingMode => false, Reload=>false,
 Keywords => {"Linear Algebra"},
 Certification => {
@@ -87,7 +87,8 @@ export{
   "PointOptions", --options to be based to the RandomPoints package
   "UseOnlyFastCodim",
   "RegularInCodimensionTutorial", --help file
-  "FastMinorsStrategyTutorial"
+  "FastMinorsStrategyTutorial",
+  "VerifyNonRegular"
 }
 
 protect MutableSmallest;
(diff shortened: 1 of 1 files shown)


M2/Macaulay2/packages/FastMinors.m2
```diff
@@ -847,7 +849,9 @@ regularInCodimension(ZZ, Ring) := opts -> (n1, R1) -> (
         R1a = ambR/Id;
     );
 
-    if not (isField coefficientRing ambR) then return "Ambient ring is not field";
+    if not (isField coefficientRing ambR) then return "Ambient coefficient ring is not field";
+    if (not instance(ambR, PolynomialRing)) then return "Ambient ring is not a polynomial ring";
+    if (Id == 0) then return true;
```
mahrud: Two things:
1. were these supposed to be errors? Having a function that sometimes returns boolean and sometimes a string is an anti-pattern.
2. the parentheses aren't necessary.

M2/Macaulay2/packages/FastMinors.m2
```diff
@@ -921,6 +930,21 @@ regularInCodimension(ZZ, Ring) := opts -> (n1, R1) -> (
             quotient1 = ambR/(Id+sumMinors);
             d = dim(ideal quotient1);
         );
+        if (opts.VerifyNonRegular) and (r-d <= n1) then (--if we should try to check if the ring is not regular
+            if (opts.Verbose or debugLevel > 0) then print "regularInCodimension: verify nonregularity";
```
mahrud: I would suggest using `printerr` instead of `print` for info logs like this.

M2/Macaulay2/packages/FastMinors.m2
```diff
@@ -921,6 +930,21 @@ regularInCodimension(ZZ, Ring) := opts -> (n1, R1) -> (
...
+            decompList = minprimes(Id + sumMinors);
+            dimList = apply(decompList, jj -> dim jj);            
+            ij = 0;
```
mahrud, line 942: For variables used inside the loop, you could just define them using `:=` and avoid the `local ...` declarations above.

M2/Macaulay2/packages/FastMinors.m2
kschwede: Yes, they were supposed to be errors.  I will update and fix.

M2/Macaulay2/packages/FastMinors.m2
kschwede: Done, and changed throughout the file.

M2/Macaulay2/packages/FastMinors.m2
kschwede: I'd rather not in this case as long as anyone doesn't mind.

kschwede:
I've pushed the changes @mahrud Mahrud suggested, and I pushed some additional changes suggested by ChatGPT 5.5 Extended Pro  (no code was written by chatgpt, but I reviewed the suggested changes and implemented those that I agreed with).

mahrud:
All looks good to me! Thanks for the AI attribution.

PR #4301, opened 2026-05-14
improved caching syntax in InvariantRings

galettof:
used ??= when possible to improve syntax for caching as suggested by @mahrud

### DIFF
--- a/M2/Macaulay2/packages/BettiCharacters.m2
+++ b/M2/Macaulay2/packages/BettiCharacters.m2
@@ -809,98 +809,87 @@ actors = method(TypicalValue=>List)
 -- returns actors on resolution in a given homological degree
 -- if homological degree is not the one passed by user upon construction,
 -- the actors are computed and stored
-actors(ActionOnComplex,ZZ) := List => (A,i) -> (
-    -- if not cached, compute
-    if not A.cache#?(symbol actors,i) then (
-	-- homological degrees where action is already cached
-	places := apply(select(keys A.cache, k -> instance(k,Sequence) and k#0 == symbol actors), k -> k#1);
-	-- get the complex
-	C := A.target;
-	-- if zero in that hom degree, return zeros
-	if zero(C_i) then return toList(A.numActors:map(C_i));
-	-- if hom degree is to the right of previously computed
-	if i > max places then (
-	    -- compute GB of differential but only up to min gens
-	    -- NOTE: does not work if ChangeMatrix=>false (which is default)
-	    GB := gb(C.dd_i,StopWithMinimalGenerators=>true,ChangeMatrix=>true);
-	    A.cache#(symbol actors,i) =
-	    apply(A.ringActors, actors(A,i-1),
-		-- given a map of free modules C.dd_i : F <-- F',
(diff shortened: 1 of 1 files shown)


M2/Macaulay2/packages/BettiCharacters.m2
```diff
@@ -809,98 +809,87 @@ actors = method(TypicalValue=>List)
...
-	    );
-	-- cache character
-	A.cache#(symbol character,i) = 	new Character from {
+character(ActionOnComplex,ZZ) := Character => op -> (A,i) -> A.cache#(symbol character,i) ??= (
+    F := coefficientRing A.ring;
+    DR := A.degreesRing;
+    n := A.numActors;
+    -- if complex is zero in hom degree i, return empty character, don't cache
+    if zero (A.target)_i then (
+	return new Character from {
```
mahrud, line 864: Small note: calling `return` in the middle of the function actually skips storing it in `A.cache#(symbol character,i)`, so you might want to do this instead:
```suggestion
	return A.cache#(symbol character,i) = new Character from {
```

galettof: Thanks! I prefer not to store zero actors/characters, so this is actually good for me.

M2/Macaulay2/packages/BettiCharacters.m2
```diff
@@ -1045,17 +1033,16 @@ actors(ActionOnGradedModule,List) := List => (A,d) -> (
...
 		--result must be reduced against module relations
 		--then factored by original basis to get action matrix
 		(g,g0) -> (sub(g0*b,g) % A.relations) // GB
+		);
 	    );
-	);
-    );
-    -- return cached value
-    A.cache#(symbol actors,degRep)
+	A.cache#(symbol actors,degRep)
```
mahrud, line 1044: This end part is unnecessary. Your function should look like this:
```m2
actors(ActionOnGradedModule,List) := List => (A,d) -> (
    -- ensure function is computed with rep of degree orbit
    degRep := A.degreeRepresentative d;
    A.cache#(symbol actors,degRep) ??= (
...
       ) -- notice no semicolon, so A.cache#(symbol actors,degRep) is returned
    )
```

M2/Macaulay2/packages/BettiCharacters.m2
mahrud: Makes sense! Just wanted to point out. (I sometimes leave a comment to the effect that "we return here to avoid caching")

M2/Macaulay2/packages/BettiCharacters.m2
galettof: I agree. I would also have to remove the semicolon in the if statement though. I think my decision to keep the end part here was driven by the preference to see what's returned in a single line at the very end of the function.

mahrud: That's fine, the only thing you would miss out is "tail-call optimization", but that's probably minor in this case.

d-torrance (approved):
Thanks!

Issue #3421, opened 2024-08-21
building M2 on Apple silicon MacOS Sonoma


mikestillman:
This issue will contain my experiments for compiling M2 on Apple silicon, on MacOS Sonoma, and with different compilers, cmake/autotools, etc.  I have found a number of issues, and now I want to be systematic about this.

My use of this as an issue is experimental (for me!)

Here is the first list to try

- [ ] cmake build, apple clang
  - [x] configure
          earlier, I needed to install brew packages msolve and texinfo (for install-info)
          I also need to add in `-DBUILD_NATIVE=off`, perhaps this should be done automatically?
          ```-DCMAKE_PREFIX_PATH="`brew --prefix tbb@2020`;`brew --prefix`"```
  - [x] executable `ninja M2-core`
  - [x] examples `ninja install-packages`
      Question: How do I tell what the default target for `ninja` is?
  - [ ] package tests
        One test in RunExternalM2 fails.  Also, with one package left, the process appears to hang (it is either done with SpecialFanoFourfolds, or hanging during that).
  - [ ] M2 tests
- [ ] cmake build, llvm clang
- [ ] cmake build, gcc14

mikestillman:
@d-torrance @mahrud When compiling on appleclang, on MacOS Sonoma (maybe others too), I get the following problem during ctest.  This same problem happens with every test in the `normal` directory.

Try

```
ctest -R normal/000-core.m2 -V
```

Running this gives on my apple M1 mac:
```
4155: Test command: /Users/mike/src/M2-current-branches/M2-development/M2/BUILD/mike/builds.tmp/cmake-appleclang/usr-dist/arm64-Darwin-macOS-14.6.1/bin/M2 "-q" "--stop" "--silent" "--no-debug" "--no-randomize" "-e" "errorDepth=1" "-e" "debugLevel=0" "-e" "gbTrace=" "-e" "path=join(path,{"/Users/mike/src/M2-current-branches/M2-development/M2/Macaulay2/tests/","/Users/mike/src/M2-current-branches/M2-development/M2/packages/"})" "-e" "input("/Users/mike/src/M2-current-branches/M2-development/M2/Macaulay2/tests/normal/000-core.m2")" "-e" "exit 0"
4155: Working Directory: /Users/mike/src/M2-current-branches/M2-development/M2/Macaulay2/tests/normal
4155: Environment variables: 
4155:  GC_MAXIMUM_HEAP_SIZE=400M
[...]

PR #4303, opened 2026-05-14
Set default `gbTrace` value for CMake build-system

alejo7797:
If `gbTrace` is not passed either as an environment variable or using `-DgbTrace=foo` then the command line produced by `ctest` for most tests is incorrect, e.g. as seen here: https://github.com/Macaulay2/M2/issues/3421#issuecomment-4453102652

### DIFF
--- a/M2/Macaulay2/packages/CMakeLists.txt
+++ b/M2/Macaulay2/packages/CMakeLists.txt
@@ -33,6 +33,7 @@ set(PACKAGES_DEVEL       "EngineTests" CACHE INTERNAL "the list of packages to o
 ## Arguments to M2
 set(errorDepth                  "1" CACHE STRING "set the error printing depth")
 set(debugLevel                  "0" CACHE STRING "set the debugging level")
+set(gbTrace                     "0" CACHE STRING "set the Groebner basis trace level")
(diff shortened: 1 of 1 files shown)


M2/Macaulay2/packages/CMakeLists.txt
```diff
@@ -33,6 +33,7 @@ set(PACKAGES_DEVEL       "EngineTests" CACHE INTERNAL "the list of packages to o
 ## Arguments to M2
 set(errorDepth                  "1" CACHE STRING "set the error printing depth")
 set(debugLevel                  "0" CACHE STRING "set the debugging level")
+set(gbTrace                     "0" CACHE STRING "set the engine tracing level")
```
mahrud: Thanks! Apparently I removed this in eea3c6e552fb3215b1e8b36925a30f45a5fa16a8 since it wasn't used in this directory, not realizing at it's still used in other directories. I would just change the description:
```suggestion
set(gbTrace                     "0" CACHE STRING "set the Groebner basis trace level")
```
(Feel free to push a new commit with `[skip ci]` somewhere in the message to skip the builds)

d-torrance (approved):
Thanks!

Issue #261, opened 2015-05-29
reloadPackage would be useful


zteitler:
For convenience (especially for new users) it would be nice to have a command ```reloadPackage``` that would reload an already-loaded package. Perhaps something like this would work:
```
reloadPackage = method();
reloadPackage (Package) := Package => o -> P -> loadPackage(toString P, Reload => true, o)
```

zteitler:
I couldn't get the title via ```P.title``` or similar things.

DanGrayson:
Why not just make loadPackage(Package) a method?

And add a method for reloadPackage(String)?

And submit a pull request for us implementing this?

PR #4312, opened 2026-05-18
Added check for getconf

coolcuber:
This is the cause of a subtle problem @wegank encountered while drafting [#519802](https://github.com/NixOS/nixpkgs/pull/519802) in nixpkgs.  Surprisingly, if `getconf` isn't on `$PATH`, Core check `#261` can fail, for reasons detailed in [this comment](https://github.com/NixOS/nixpkgs/pull/519802#issuecomment-4472656747) (also included below).  This is a bit of a nitpick because `getconf` can be found on any system that uses `glibc`, and some that don't, but is not a coreutil, so a check before the rest of the build couldn't hurt.  It certainly would've saved me a day's worth of going through log files.
>I think [this line](https://github.com/coolcuber/nixpkgs-review-gha/actions/runs/25978294393/job/76362348774#step:6:2146) reveals the issue.  Because `getconf` couldn't be found, the documentation database files went to `/lib/share/Macaulay2/*/cache/rawdocumentation-dcba-.db` and `check_261` tries to load them from `/lib/share/Macaulay2/*/cache/rawdocumentation-dcba-8.db` (the correct path), but it didn't find them there, so it ran out of memory trying to load `Macaulay2Doc` without them (see [this comment](https://salsa.debian.org/math-team/macaulay2/-/blob/debian/latest/debian/patches/skip-failing-core-tests.patch?ref_type=heads#L35) in the Debian package).  I think we missed this because there isn't a check for `getconf` in the configuration script and on Linux `getconf` is in `glibc-bin`.

### DIFF
--- a/M2/configure.ac
+++ b/M2/configure.ac
@@ -231,6 +231,8 @@ then DOCUMENTATION="html info"
 elif test "$DOCUMENTATION" = download
 then DOCUMENTATION=""
      BUILDLIST="$BUILDLIST Macaulay2-docs"
+     AC_CHECK_PROGS(GETCONF,getconf,false)
+     AS_IF([test $GETCONF = false], AC_MSG_ERROR([getconf is required]))
 else for DOCTYPE in $DOCUMENTATION
     do case $DOCTYPE in
             html|pdf)


d-torrance (changes requested):
Thanks!

The error from the build log is occuring during the build of the `Macaulay2-docs` "library", which will only happen for builds that call `configure` with the `--enable-documentation=download` option.  (Indeed, `getconf` is called in the `Macaulay2-docs` Makefile to determine the pointer size of the system so we  know the filenames of the gdbm dumps of the documentation database.)  So I think this check should probably be conditional on that.

Would you move it down here?

https://github.com/Macaulay2/M2/blob/c6aff10750a8b8a7e221f04c72d3d8b90c902ea9/M2/configure.ac#L231-L233

coolcuber:
[This](https://github.com/Macaulay2/M2/actions/runs/26033763605/job/76526299146?pr=4312#step:16:5318) really perplexes me (the only change was in `configure.ac`).  Is this a known issue with the CMake build?

d-torrance:
Not a cmake issue, but a race condition in the ThreadedGB package.  We get the same thing on the autotools build every once in a while, too

d-torrance (approved):
Thanks!

PR #4313, opened 2026-05-18
Allow gcd in fields that are also quotient rings

joel-dodge:
I ran into this bug when computing pushFwd along some ring maps. The logic change seems crisp and clear enough so I am pushing the bare change as a PR but would be happy to reproduce an example that led to me noticing this and capturing it in a test case if requested.

This is in effect just a swap of L24 <> L25 so that the early return comes before the QuotientRing error check. Everything else is just standardized formatting so I could read the code to make sure I understood it while making this change.

### DIFF
--- a/M2/Macaulay2/m2/factor.m2
+++ b/M2/Macaulay2/m2/factor.m2
@@ -15,28 +15,36 @@ monic := t -> (
 gcd(ZZ,RingElement) := (r,s) -> gcd(promote(r,ring s),s)
 gcd(RingElement,ZZ) := (r,s) -> gcd(promote(s,ring r),r)
 
-gcd(RingElement,RingElement) := RingElement => (r,s) -> (
-     R := ring r;
-     if ring s =!= R then error "gcd: expected elements in the same ring";
-     if instance(R,QuotientRing) then error "gcd: unimplemented for this ring";
-     if isField R then if r == 0 and s == 0 then 0_R else 1_R
-     else if factoryAlmostGood R then (
-	  if (options R).Inverses then (r,s) = (numerator r, numerator s);
- 	  new ring r from rawGCD(raw r, raw s))
-     else (
-	 (R1,f,g) := flattenRing(R,Result=>3);
-	 if factoryAlmostGood R1 then g gcd(f r,f s)
-     else if instance(R,PolynomialRing) and numgens R == 1 and isField coefficientRing R then monic (
-	  -- does this depend on the monomial order in R, too?
-	  -- would this code work for more than one variable?
-	  if gbTrace >= 3 then << "gcd via syzygies" << endl;
-	  if r == 0 then s
-	  else if s == 0 then r
-	  else (
-	       a := (syz( matrix{{r,s}}, SyzygyLimit => 1 ))_(0,0);
-	       if s%a != 0 then error "can't find gcd in this ring";
-	       s // a))
-     else notImplemented()))
+gcd(RingElement, RingElement) := RingElement => (r, s) -> (
+    R := ring r;
+    if ring s =!= R then error "gcd: expected elements in the same ring";
+
+    -- GaloisFields are QuotientRings so this line has to go before the
+    -- QuotientRing check
+    if isField R then return if r == 0 and s == 0 then 0_R else 1_R;
+    if instance(R, QuotientRing) then error "gcd: unimplemented for this ring";
+
+    if factoryAlmostGood R then (
+        if (options R).Inverses then (r, s) = (numerator r, numerator s);
+        new ring r from rawGCD(raw r, raw s)
+    ) else (
+        (R1, f, g) := flattenRing(R, Result => 3);
+        if factoryAlmostGood R1 then g gcd(f r,f s)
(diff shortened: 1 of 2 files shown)


d-torrance (changes requested):
Thanks!

Would you add a test (in an appropriate file under `M2/Macaulay2/tests/normal`) that would have failed before this change of taking the gcd of two elements in a finite field?

Also, it's pretty stupid in this case, but it might be cool if `gcdCoefficients` also worked over finite fields.

M2/Macaulay2/m2/factor.m2
```diff
@@ -15,28 +15,36 @@ monic := t -> (
...
-	       s // a))
-     else notImplemented()))
+gcd(RingElement, RingElement) := RingElement => (r, s) -> (
+    R := ring r;
+    if ring s =!= R then error "gcd: expected elements in the same ring";
+
+    -- GaloisFields are QuotientRings so this line has to go before the
+    -- QuotientRing check
+    if isField R then return if r == 0 and s == 0 then 0_R else 1_R;
+    if instance(R, QuotientRing) then error "gcd: unimplemented for this ring";
```
mahrud: I know this is not something you changed, but `unimplemented` seems incorrect. Also we have `notImplemented("blah")`.

M2/Macaulay2/m2/factor.m2
```diff
@@ -15,28 +15,36 @@ monic := t -> (
...
-	       a := (syz( matrix{{r,s}}, SyzygyLimit => 1 ))_(0,0);
-	       if s%a != 0 then error "can't find gcd in this ring";
-	       s // a))
-     else notImplemented()))
+gcd(RingElement, RingElement) := RingElement => (r, s) -> (
+    R := ring r;
+    if ring s =!= R then error "gcd: expected elements in the same ring";
+
+    -- GaloisFields are QuotientRings so this line has to go before the
+    -- QuotientRing check
```
mahrud: nitpick: maybe make this one line?

mahrud:
This is a broader comment about gcd: perhaps each ring should have its own gcd function so we don't have to have a long gcd method on RingElement that checks a bunch of unrelated cases. This could be accomplished similar to `ReduceHook`, for instance.

M2/Macaulay2/m2/factor.m2
joel-dodge: If you suggest an appropriate error message I'll update it!

mahrud:
```m2
i2 : notImplemented("gcd for quotient rings is")
stdio:2:14:(3):[1]: error: gcd for quotient rings is not implemented yet
```

joel-dodge:
> Also, it's pretty stupid in this case, but it might be cool if `gcdCoefficients` also worked over finite fields.


Just to clarify, by this do you mean:

```
R = GF(9);
i = random(0, 8); j = random(0, 8);
gcdCoefficients(R_i, R_j)
```

or do you mean:

```
R = GF(9)[x];
gcdCoefficients(x + a, x*(x+a)*(x-a)^2)
```

IIUC the former is not supported (and I think doesn't make sense to be supported given what gcdCoefficients says on the tin) and the latter is supported and works as expected currently.

d-torrance:
I was referring to the former.  It's silly, but perhaps `gcdCoefficients(a, b)` should return something like `(1_R, 1/a, 0_R)` if `a` is nonzero (and similar if `a=0` but `b` is nonzero, and `(0_R,0_R,0_R)` if they're both zero).

M2/Macaulay2/m2/factor.m2
```diff
@@ -15,28 +15,35 @@ monic := t -> (
...
-	       if s%a != 0 then error "can't find gcd in this ring";
-	       s // a))
-     else notImplemented()))
+gcd(RingElement, RingElement) := RingElement => (r, s) -> (
+    R := ring r;
+    if ring s =!= R then error "gcd: expected elements in the same ring";
+
+    -- GaloisFields are QuotientRings so this line has to go before the QuotientRing check
+    if isField R then return if r == 0 and s == 0 then 0_R else 1_R;
+    if instance(R, QuotientRing) then notImplemented("gcd for quotient rings is")
```
d-torrance: (Trying out this suggestion thing...)

```suggestion
    if instance(R, QuotientRing) then notImplemented("gcd for quotient rings is");
```

d-torrance (changes requested):

M2/Macaulay2/m2/factor.m2
joel-dodge: and ill try out commiting them. I usually hate doing this since then i have to pull the changes back down if i want to make any local changes on top of them but this is for science. :D

joel-dodge:
> I was referring to the former. It's silly, but perhaps `gcdCoefficients(a, b)` should return something like `(1_R, 1/a, 0_R)` if `a` is nonzero (and similar if `a=0` but `b` is nonzero, and `(0_R,0_R,0_R)` if they're both zero).

oooh yes makes perfect sense.

d-torrance (approved):
Thanks!

PR #4314, opened 2026-05-18
added the ability to resetProfileTable

LukeOeding:
resetProfileTable lets the user reset this variable so that they can profile different functions in the same M2 session and not have the results just be cumulative to what has already been profiled.

### DIFF
--- a/M2/Macaulay2/d/profiler.dd
+++ b/M2/Macaulay2/d/profiler.dd
@@ -1,4 +1,4 @@
--- Copyright 2024 by Mahrud Sayrafi
+-- Copyright 2024 by Mahrud Sayrafi (and Luke and Rosie and Jay, but mostly Jay)
 
 use binding;  -- for symbols
 use evaluate; -- for eval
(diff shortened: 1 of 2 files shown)


LukeOeding:
this is a simple comment

M2/Macaulay2/d/profiler.dd
```diff
@@ -40,7 +40,14 @@ setupfun("stacktrace", stacktrace).Protected = false; -- will be overloaded in m
...
 ProfileTable.beingInitialized = false; -- mutex locks will occur
-setupconst("ProfileTable", Expr(ProfileTable));
+profileTableS:= setupvar("ProfileTable", Expr(ProfileTable));
+
+resetProfileTable(e:Expr):Expr := (
+    ProfileTable = newHashTable(Tally, nothingClass);
+    ProfileTable.beingInitialized = false; -- for technical reasons
+    setGlobalVariable(profileTableS,Expr(ProfileTable));
+    nullE);
+setupfun("resetProfileTable", resetProfileTable);
```
d-torrance, line 50: Would you add documentation for this new function?

d-torrance (changes requested):

mahrud:
@d-torrance  We decided to combine various changes into a single PR later.

PR #4318, opened 2026-05-18
format (kernel, RingMap) hooks code for readability

joel-dodge:
I had to reformat the code to understand it for the purposes of writing documentation so pushing the reformatting as it seems net positive and improves adherance to macaulay2 code formatting standards. It is expected that this is a white space only change.

### DIFF
--- a/M2/Macaulay2/m2/ringmap.m2
+++ b/M2/Macaulay2/m2/ringmap.m2
@@ -239,109 +239,122 @@ kernel RingMap := Ideal => opts -> (cacheValue (symbol kernel => opts)) (f -> (
 algorithms := new MutableHashTable from {}
 algorithms#(kernel, RingMap) = new MutableHashTable from {
     FractionField => (opts, f) -> (
-	(F, R) := (target f, source f);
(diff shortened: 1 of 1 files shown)


mahrud:
I appreciate the clarification that this is a white space only change. The reason it's like this is that at some point in the past these were separate functions, then I changed them into hooks, but in order to preserve git history (e.g. blame) I didn't re-indent the body of the algorithms and only changed the scaffolding.  Because of this, I'm ambivalent-to-opposed towards merging this PR.

joel-dodge:
I am surprised at wanting to trade off current human readability for ease of git history inspection in this way. Certainly reading the code to understand what it does without caring who write it is the more most common case, right? I would also say that after a refactoring like this, even a mostly formal one, the refactorer IS the author of these lines now! Doesn't it make sense to prioritize readability of the code and let anyone who wants to know the evolution in git history use a more complicated workaround (eg checkout the last edits parent, git blame there, wash rinse repeat)?

d-torrance (approved):
Thanks!

PR #4176, opened 2026-03-30
Some contributions to the PushForward package

joel-dodge:
This PR touches nearly every line of the PushForward package. It offers new functionality, bug-fixes and pure refactoring for readability / maintainability but leaves the algorithmic core of the package mostly unchanged. Itemized changes include:
* extends `isModuleFinite` and `pushFwd` methods to cover the case of skew symmetric polynomial rings and to better handle relations coming from towers of rings.
* implements isomorphisms in both directions between a module and it's push-forward available via the `pushforward` and `pushforward'` methods.
* introduces auxiliary files and separates tests and documentation nodes to their own files.
* renames some internal methods and refactors them to share code where possible.
* implements caching scheme to support `pushforward`, `pushforward'` methods and to cache computations of push-forward modules.
* adds test cases to cover all new functionality.

It is recommended to review this PR in github commit by commit although even then the changes are not all easy to read. Moving the auxiliary files in particular seems to have confused git into producing a difficult to read diff. I also recommend turning off whitespace changes in the viewing settings for this PR.

PR #4322, opened 2026-05-18
Atlanta workshop test-suite overhaul 

kellerlv:
PR to run Macaulay2's CI against the accumulated test-suite overhauls from the Atlanta workshop branch.

We are open to suggestions for ways to split this up in a way that is easier to digest for the reviewers. 

AI (Claude) was used for multiple tasks such as reformatting, detecting broken/commented out tests, helping trace bugs, and other tedious yet straightforward tasks.


M2/Macaulay2/packages/PushForward.m2
mahrud, line 1: This one will conflict with #4176.

M2/Macaulay2/packages/DirectSummands/tests.m2
mahrud, line 1: Somewhat confused about these added tests. Are these things that weren't already checked by existing tests?

M2/Macaulay2/packages/PushForward.m2
kellerlv: Got it, will take a look

M2/Macaulay2/packages/DirectSummands/tests.m2
kellerlv: Some of these functions were legitimately untested and we've just been going on the principle that any untested exported function should have some test, however trivial. But some of these are indeed indirectly tested, so they can easily be pruned down.

mahrud:
@kellerlv or @tbrysiewicz, when this is ready, could one of you rebase this PR and force push to "linearize" the commit history? (i.e. get rid of merge commits)

kellerlv:
> @kellerlv or @tbrysiewicz, when this is ready, could one of you rebase this PR and force push to "linearize" the commit history? (i.e. get rid of merge commits)

Yup, we can do that

mahrud:
Beautiful rebasing, thank you! Could you disclose AI use in the top PR message?

M2/Macaulay2/m2/testaudit.m2
mahrud, line 1: I presume this is redundant with the TestAudit package PR? Which one should be merged first?

M2/Macaulay2/packages/DirectSummands/tests.m2
```diff
@@ -464,6 +464,97 @@ TEST ///
...
+  assert not isDirectSummand(S^3, M)
+  -- non-free L: a genuine summand of N ++ N is found
+  N = module ideal(x,y,z)
+  assert not isFreeModule N
+  assert isDirectSummand(N, N ++ N)
+  -- non-free L: an unrelated module is not a summand
+  assert not isDirectSummand(module ideal(x,y), N ++ N)
+  -- error: the two modules must be defined over the same ring
+  S2 = ZZ/5[a,b,c]
+  assert(try (isDirectSummand(S2^1, M); false) else true)
```
Devlin-Mallory: This test passes if `isDirectSummand(S2^1, M)` fails for any reason, not just for being defined over the same ring. That's probably fine, but we could also try to test against the error message from `isDirectSummand` ("expected objects over the same ring").

M2/Macaulay2/packages/DirectSummands/tests.m2
```diff
@@ -464,6 +464,97 @@ TEST ///
...
+  -- run: the projectors drive the decomposition into three summands
+  L = summandsFromProjectors(M, P)
+  assert(3 == #L)
+  assert isIsomorphic(directSum L, M)
+  -- error: findProjectors rejects a non-homogeneous module
+  Minhom = coker matrix{{x, y + 1}, {y, x}}
+  assert not isHomogeneous Minhom
+  assert(try (findProjectors Minhom; false) else true)
+  -- error: an indecomposable module (here R^1) yields no projectors
+  assert(try (findProjectors R^1; false) else true)
```
Devlin-Mallory: Similarly to my previous comment: maybe we should have tests that pass on `findProjectors` returning `null` look different than tests that pass when an error is raised.

M2/Macaulay2/packages/DirectSummands/tests.m2
```diff
@@ -464,6 +464,97 @@ TEST ///
...
+  -- property: each returned endomorphism is (weakly) idempotent
+  assert all(I, isWeakIdempotent)
+  -- run: the idempotents drive the decomposition into three summands
+  L = summandsFromIdempotents(M, I)
+  assert(3 == #L)
+  assert isIsomorphic(directSum L, M)
+  -- findIdem is an exported synonym for findIdempotents
+  assert(findIdem === findIdempotents)
+  -- error: an indecomposable module (here R^1) yields no idempotents
+  assert(try (findIdempotents R^1; false) else true)
```
Devlin-Mallory: Likewise here

M2/Macaulay2/packages/DirectSummands/tests.m2
```diff
@@ -464,6 +464,97 @@ TEST ///
...
+  -- the 0-th Frobenius pullback is the identity
+  assert(frobeniusPullback(0, R^{-1}) == R^{-1})
+  -- on a cokernel, the presentation entries are raised to the p^e power
+  assert(frobeniusPullback(1, coker matrix{{x}}) == coker matrix{{x^3}})
+  assert(frobeniusPullback(2, coker matrix{{x}}) == coker matrix{{x^9}})
+  -- the matrix version raises every entry to the p^e power
+  assert(frobeniusPullback(1, vars R) == map(R^1, R^{-3}, {{x^3}}))
+  -- the Ring and Ideal inputs reduce to the module case
+  assert(frobeniusPullback(1, R) == R^1)
+  assert instance(frobeniusPullback(1, ideal x), Module)
```
Devlin-Mallory, line 570: These are great, and perhaps we're happy just having coverage of each function, but it might be worth adding tests that address two of the more subtle settings for `frobeniusPullback`:
1. When R is defined over GF p^e for e > 1, `frobeniusPullback` also involves a twist of the underlying ring. So maybe a test for the behavior in this setting (e.g., if R = (GF 4)[x]/(x+a), then the first Frobenius pullback of R^1 should be ((GF 4)[x]/(x^2+x+a), etc).
2. Maybe a test for `frobeniusPullback(ZZ, CoherentSheaf)` would be good - e.g., check `frobeniusPullback(1,frobeniusPushforward(1,OO_X))` is trivial when `X` is an elliptic curve. 

(I feel bad asking you to write tests for things we didn't!)

Devlin-Mallory:
Left a few local comments on the DirectSummands tests, but overall I'm happy with the added tests - thanks all!

M2/Macaulay2/m2/testaudit.m2
kellerlv: Yes (edited because I misread this originally)

M2/Macaulay2/packages/DirectSummands/tests.m2
kellerlv: No worries asking for tests, we are happy to add suggestions from the package authors.

kellerlv: For 1. : I'm a little confused. Where did x^2 + x + a come from? I was expecting the quotient to become (GF 4)[x] / (x + a^2). The current code is giving the quotient by x + a + 1, which I guess makes sense since a^2 = a+1. 

For 2: I tried a characteristic 5 example and one subtlety is that we actually get a rank 5 bundle, but this technically makes sense since pushing forward O_X should technically decompose it as a rank 5 bundle, then pulling back each of those summands gives a line bundle, so the output should be rank 5, no? This is also what the code gave when I ran your suggestion.

Devlin-Mallory: For 2: sorry, by trivial I just meant direct sum of OO_X, so exactly as you say! (Or maybe there's good tests for pullbacks of the tangent sheaf of P^n, though I don't know what one would hope to be true.)

For 1: You're correct, I intended x + a^2 = x + a + 1 as the denominator.

M2/Macaulay2/packages/ConnectionMatrices/normalForm.m2
```diff
@@ -73,6 +73,16 @@ inferWeylAlgebra = F -> (
...
+--
+-- TODO: scope gotcha worth smoothing over. R is `(baseFractionField D)[d_1..d_n]`, so its
+-- coefficient ring F is a commutative fraction field whose generators have the SAME names
+-- as D's coordinate variables (x, y, ...). However, `use R` only exposes R's own
+-- generators (the d_i); the bare names x, y, ... continue to point to D's non-commutative
+-- Weyl-algebra elements. As a result, expressions like `x/y` evaluated in test/user code
+-- error with "fraction field of non-commutative ring requested". Workaround for callers:
+-- access the coefficient-ring vars positionally, e.g. `F = coefficientRing R; xF = F_0;`,
+-- and build R-expressions from xF, yF, R_0, R_1, .... A future improvement would be to
+-- provide a helper that rebinds the names to the F-generators on entry to a scope.
```
mahrud: I'm really curious, was this AI generated?!

If so, it's a surprisingly deep observation, but misses the fact that `rationalWeylAlgebra` is intentionally not exported because it's not even really the rational Weyl algebra (which M2 doesn't currently support), but only it's graded associative ring (see the comment above).

The intended usage is something like this:
```m2
i1 : debug needsPackage "ConnectionMatrices"

i2 : D = makeWeylAlgebra(QQ[x,y], w = {1,2});

i3 : F = baseFractionField D;

i4 : R = rationalWeylAlgebra D;

i5 : ring \ {x,y,dx,dy}

o5 = {F, F, R, R}

o5 : List

i6 : x/y

     x
o6 = -
     y

o6 : F

i7 : x/y * dx * dy

     x
o7 = -*dx*dy
     y

o7 : R
```

kellerlv: This was hybrid AI and myself -- I distinctly remember this back and forth and probably would have just moved on for lack of mental energy if I hadn't been able to tell Claude to investigate this closer.

mahrud: Interesting. Anyway, the merge conflict over ConnectionMatrices was my fault (well, we wanted to update the package) so I just resolved it.

M2/Macaulay2/packages/Saturation/saturate-test.m2
```diff
@@ -235,3 +235,30 @@ TEST ///
...
+  assert(isSubset(I, saturate(I, x)))
+  -- saturate stabilizes the quotient sequence:
+  -- saturate(I, f) equals quotient(I, f^N) once N is large enough
+  assert(saturate(I, x) == quotient(I, x^10))
+  -- isSupportedInZeroLocus boundary cases
+  S = QQ[u,v]
+  m = ideal vars S
+  assert(isSupportedInZeroLocus_m (S^0) == true)
+  assert(isSupportedInZeroLocus_m (S^1/m^2) == true)
+  assert(isSupportedInZeroLocus_m (S^1) == false)
```
mahrud: Just an observation (not worth changing):
Several places above I saw assertions like:
```m2
  assert(isSubset(I, saturate(I, x)))
```
Which I would write without the outermost parentheses `assert isSubset(I, saturate(I, x))`, but it's really strange to test with `== true/false` rather than this:

```suggestion
  assert isSupportedInZeroLocus_m (S^0)
  assert isSupportedInZeroLocus_m (S^1/m^2)
  assert not isSupportedInZeroLocus_m (S^1)
```

M2/Macaulay2/packages/Isomorphism.m2
```diff
@@ -867,6 +867,35 @@ TEST ///
...
+  -- The cache stores the isomorphism under key (N, M, {true, false})
+  -- rather than {true, true} (which is what the silenced assert at
+  -- :752 expects).
+  assert(N.cache.cache.Isomorphisms#?(N, M, {true, false}))
+  assert(not N.cache.cache.Isomorphisms#?(N, M, {true, true}))
+  -- A single isomorphism is currently cached.
+  assert(1 == #N.cache.cache.Isomorphisms)
+  -- And nothing is cached on M (only the larger of the two stores).
+  assert(not M.cache.cache.?Isomorphisms)
+///
```
mahrud, line 897: Huh this is almost identical to the test starting at line 745, except that the variables are defined with := instead of = and it quotes my comments there rather than copying them, and importantly "pins the current behavior" even though it is incorrect.

Isn't this counterproductive, in that it gives the impression that certain code paths are being tested, even though previously they were not being tested because they were incorrect?

kellerlv: Oh yeah -- this was inspired by some discussions about a wished-for "BrokenTest" type or something like that, where one can put known bugs in the broken tests and it'll actually give an error if you do something that fixes the test (then it can be easily converted into a proper TEST). My reasoning here is that sometimes you don't know what a given change will fix and it's useful to have things formally recorded that DO fail so it's easy to tell if you've fixed something, so this is a hack-y version of trying to accomplish this.

M2/Macaulay2/packages/TestAudit.m2
```diff
@@ -0,0 +1,523 @@
+newPackage(
+          "TestAudit",
```
d-torrance: Since this package was merged separately in #4327, there's a conflict now.  So this file (and `test-audit.m2`) should be removed

d-torrance (changes requested):

kellerlv:
The StatePolytope failure on autotools-ubuntu looks like a gfan-version bump in CI, not from this PR. Between #4332 (May 20, green) and now, the Macaulay2 apt repo started shipping gfan 0.8~beta-2 which overrides the stock Ubuntu gfan 0.6.2. Test #0 (pre-existing, not from this PR) and test #3 (ours) both pin gfan output via `set X === set Y`, which breaks if the new gfan normalizes monomials any differently. Both pass locally with Homebrew's gfan 0.8beta.

Is the gfan bump intentional? If so I can soften test #3 here, but test #0 is a wider issue.

antonleykin (approved):
I've looked at the tests in some packages I was involved in. There are no obvious blunders. My opinion: unless anyone notices anything major that needs immediate correction, let's merge this PR. 

We should alert the community (package developers) that their packages test suites have been revised and ask to correct (add/remove/replace) if necessary.

PR #4326, opened 2026-05-19
Use FLINT for sqrt(CC) 

d-torrance:
This way, we can remove complex.c and complex.h from the e directory.


d-torrance:
Closing -- will open a PR against Michael's fork instead

PR #4327, opened 2026-05-19
Test audit package

tbrysiewicz:
**PR Summary**

This branch adds a new distributed package, `TestAudit`, for producing lightweight reports about package test coverage and test organization.

The new `testAudit` command accepts either a `Package` or package name and reports:

- number and source locations of package tests
- test organization style: `no tests`, `auxiliary`, `interspersed`, or `together`
- exported functions and options not mentioned in tests
- silenced tests
- FIXME/TODO markers in source comments

It also adds optional report sections:

- `CommentReport => true`: includes comments attached to or inside tests
- `SpeedReport => true`: times tests via `check`
- `TestScore => true`: produces a heuristic score out of 100, based on test coverage, silenced tests, and FIXME/TODO markers

The package includes documentation and package-level tests for the main command and each option. It is also added to `=distributed-packages`.

There is one small follow-up TODO added in `installPackage.m2` noting a future `TestAudit => true` option, so `installPackage` could reuse test results instead of making `testAudit(..., SpeedReport => true)` rerun tests.


M2/Macaulay2/m2/installPackage.m2
```diff
@@ -661,6 +661,9 @@ installPackage = method(
 	Verbose                => false
 	})
 
+--TODO: implement option TestAudit=>true so testAudit can make use of the fact
+-- that tests are run on installPackage (otherwise testAudit(pkg, SpeedTest=>true)
+-- would have to run them a second time)
```
mahrud: Did you mean examples? `installPackage` doesn't run the tests, `check` does.

M2/Macaulay2/packages/TestAudit.m2
```diff
@@ -0,0 +1,501 @@
...
+      SeeAlso
+        CommentReport
+        SpeedReport
+        TestScore
+      ///
+
+      doc ///
+      Key
+        CommentReport
+        [testAudit, CommentReport]
```
mahrud: Perhaps it was a conscious choice, but just FYI you can document optional arguments in the documentation of the method `testAudit`, so that users can see all relevant info in one page rather than 4 different pages.

M2/Macaulay2/m2/installPackage.m2
tbrysiewicz: Oops! good point. Nevermind.

M2/Macaulay2/packages/TestAudit.m2
```diff
@@ -0,0 +1,523 @@
...
+
+-- Load documentation so package TEST blocks are available.
+testScore String := pkgname -> testScore needsPackage(pkgname, LoadDocumentation => false)
+
+
+
+      -* Documentation section *-
+      beginDocumentation()
+      doc ///
+      Key
```
mahrud: I think something strange happened with the indentation here? An easy fix, if you don't want to figure out how to unindent lots of documentation sections manually, is this:

```suggestion
-* Documentation section *-
beginDocumentation()
doc ///
Node
      Key
```
(in other words, keep the documentation body at the same level, but add `Node` as a parent keyword to all of each node.

M2/Macaulay2/packages/TestAudit.m2
```diff
@@ -0,0 +1,523 @@
...
+testAudit = method(Options => {CommentReport => false, SpeedReport => false, ScoreReport => false})
+testScore = method()
+
+--------------------------------------------------------------------------------
+-- Helpers
+--------------------------------------------------------------------------------
+
+testInputs = method()
+testInputs Package := pkg -> tests pkg
+testInputs String := pkgname -> testInputs needsPackage(pkgname, LoadDocumentation => false)
```
d-torrance: I don't think this method is necessary as it's essentially just the same thing as `tests`.

Indeed, the only difference between `tests(String)` and `testInputs(String)` is the value of `LoadDocumentation`.  But once we load the package and call `tests(Package)` on it, then we'll just call `loadPackage` again with `LoadDocumentation => true`:

```m2
i1 : code (tests, Package)

o1 = -- code for method: tests(Package)
     /usr/share/Macaulay2/Core/testing.m2:93:17-100:5: --source code:
     tests Package := pkg -> (
         if pkg#?"documentation not loaded"
         then pkg = loadPackage(pkg,
             FileName => pkg#"source file",
             LoadDocumentation => true);
         if not pkg#?"test directory loaded" then loadTestDir pkg;
         previousMethodsFound = new NumberedVerticalList from pkg#"test inputs"
         )
```

M2/Macaulay2/packages/TestAudit.m2
```diff
@@ -0,0 +1,523 @@
...
+
+
+-- The coverage checks below are textual heuristics, not dispatch tracing.
+testCodeString = method()
+testCodeStringFromInputs := inputs -> (
+    demark(newline, apply(toList(0..#inputs-1), i -> inputs#i#"code"))
+)
+
+testCodeString Package := pkg -> testCodeStringFromInputs testInputs pkg
+testCodeString String := pkgname -> testCodeString needsPackage(pkgname, LoadDocumentation => false)
```
d-torrance: We don't need `LoadDocumentation => false` here -- it's redundant since that's the default and as above, we'll end up loading it anyway when we call `tests` on the `Package` object.

M2/Macaulay2/packages/TestAudit.m2
```diff
@@ -0,0 +1,523 @@
...
+
+
+exportedSymbols = method()
+exportedSymbols Package := pkg -> sort toList pkg#"exported symbols"
+exportedSymbols String := pkgname -> exportedSymbols needsPackage(pkgname, LoadDocumentation => false)
+
+
+-- These helpers classify exported symbols by the class of their current value.
+-- Anything not recognized as function-like or a Type is reported as "other".
+exportedValue := s -> try value s else null
```
d-torrance: Is `try` necessary here?  I think `s` is also symbol, so `value` should always work.  (And even if it is necessary, `else null` is redundant because `try` returns null anyway if an error occurs.

M2/Macaulay2/packages/TestAudit.m2
```diff
@@ -0,0 +1,523 @@
...
+className := x -> toString class x
+
+isExportedFunction := s -> (
```
d-torrance: I think `instance(exportedValue s, Function)` would be much simpler here (or maybe even just `instance(value s, Function)` if `try` isn't necessary -- see comment above)

M2/Macaulay2/packages/TestAudit.m2
```diff
@@ -0,0 +1,523 @@
...
+    v := exportedValue s;
+    member(className v, {"MethodFunction", "MethodFunctionWithOptions", "FunctionClosure", "Function"})
+)
+
+isExportedType := s -> (
```
d-torrance: Same ideas as above -- we can just use `instance` here

d-torrance (changes requested):
Thanks for this awesome package!  I have a few comments from a quick scan (possibly more to come later...)

M2/Macaulay2/packages/TestAudit.m2
```diff
@@ -0,0 +1,523 @@
+newPackage(
+          "TestAudit",
+          Version => "0.1",
+          Date => "May 19, 2026",
+          Headline => "Provides test audit functionality",
```
mahrud: Nitpicky suggestion for consistency with other package headlines
```suggestion
          Headline => "test audit functionality",
```
Also, I think the same issue as `packageTemplate` with the indentation happened here.

M2/Macaulay2/packages/TestAudit.m2
```diff
@@ -0,0 +1,616 @@
...
+    "testScore",
+    "CommentReport",
+    "SpeedReport",
+    "ScoreReport",
+    "Experimental"
+    }
+
+CommentReport = symbol CommentReport
+SpeedReport = symbol SpeedReport
+ScoreReport = symbol ScoreReport
```
mahrud: I think now that this is a package these lines aren't necessary, though probably harmless.

M2/Macaulay2/packages/TestAudit.m2
```diff
@@ -0,0 +1,616 @@
...
+
+
+exportedSymbols = method()
+exportedSymbols Package := pkg -> sort toList pkg#"exported symbols"
+exportedSymbols String := pkgname -> exportedSymbols needsPackage pkgname
+
+
+-- These helpers classify exported symbols by the value attached to the symbol.
+-- Anything that is not a Function or a Type is reported as "other".
+isExportedFunction := s -> instance(value s, Function)
```
mahrud, line 57: I think this was mentioned at some point earlier, but currently methods defined on existing symbols are not counted this way. I believe the only way to get those is like this:
```m2
i1 : methods needsPackage "Pullback"

o1 = {0 => (pullback, RingMap, RingMap)      }
     {1 => (internalUseDirectSum, Ring, Ring)}
```
(The first time you run this it's slow, but after the first time lots of stuff is cached so after that it's actually pretty fast considering what it does)

d-torrance (changes requested):
Would click on "Resolve conflicts" and fix the conflict in `=distributed-packages`?  Should just be a matter of removing `<<<`', `====`', and `>>>>`.

For the most part, I think this looks like it's in pretty good shape.  Mahrud's `methods` suggestion would be a good fix.

Also, in Atlanta I believe there was a suggestion for the report to use html (a la the various types exported by the [Text](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Text/html/index.html) package) so the output will display nicely in webapp mode.  But maybe that's a stretch goal for a future release.

d-torrance (approved):
Thanks!

PR #4332, opened 2026-05-20
InvariantRing v2.5

### LINKED ISSUE #1773: check_9 "InvariantRing" sometimes fails

galettof:
InvariantRing has been updated to version 2.5. The algorithm for elementary abelian $p$-groups is now much faster, and is used by default when possible. Bug fixes.

@d-torrance: I tried to implement many of the changes suggested in the previous PR to improve the new code, especially where it was recreating lists inside loops. I also think I fixed #1773: is there a way to run Test 9 on ARM a bunch of times to see if the issue is triggered again?

### DIFF
--- a/M2/Macaulay2/packages/InvariantRing.m2
+++ b/M2/Macaulay2/packages/InvariantRing.m2
@@ -9,8 +9,8 @@
 
 newPackage(
         "InvariantRing",
-        Version => "2.4", 
-        Date => "May 1, 2026",
+        Version => "2.5", 
+        Date => "May 25, 2026",
         Authors => {
 	    {Name => "Luigi Ferraro", 
 		 Email => "luigi.ferraro@utrgv.edu", 
@@ -55,9 +55,8 @@ newPackage(
 	     },
 	AuxiliaryFiles => true,
         DebuggingMode => false,
-	PackageExports => {
-	    "Graphs"
-	    }
+	PackageImports => {"Elimination","Normaliz","Polyhedra"},
+	PackageExports => {"Graphs"}
         )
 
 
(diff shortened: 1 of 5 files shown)


d-torrance:
Awesome!  Since you're not calling `keys` anywhere now, the hash code issue shouldn't be a problem at all.

d-torrance (approved):
Thanks!

PR #4333, opened 2026-05-20
fix comment syntax highlighting

pzinn:
This fixes prism's handling of comments. silly examples of noncomments:
```m2
Symbol * Symbol := toString; symbol <-*symbol a
ZZ |- ZZ := plus; 3 |--1
```
I also removed not highlighting `:--` : I can't think of any circumstance where this would occur, but in any case I think this is a valid comment?
someone should check my regex...

### DIFF
--- a/M2/Macaulay2/editors/prism/macaulay2.js.in
+++ b/M2/Macaulay2/editors/prism/macaulay2.js.in
@@ -4,11 +4,11 @@ Prism.languages.m2 =
 Prism.languages.macaulay2 = {
     'comment': [
 	{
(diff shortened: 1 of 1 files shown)


mahrud:
Oh good catch! By the way, @YueRen fixed `A <-- B` (and longer arrows) being classified as comments by Emacs in https://github.com/Macaulay2/M2-emacs/pull/96. Might be worth also fixing them in prism.

Also, FYI, if you include `[skip ci]` somewhere in the commit message, the checks won't run.

d-torrance (approved):
Thanks!

PR #4334, opened 2026-05-20
Added exception for Darwin builds that aren't using Homebrew

coolcuber:
If a Darwin build wasn't built with Homebrew, the `dlopen'` function might lead to undesirable behavior.  If Homebrew is innstalled, it might call the wrong libraries, and if it isn't, it would lead to an error (from trying to call `!brew --prefix` without error handling).

This is to address one of the issues from [#519802](https://github.com/NixOS/nixpkgs/pull/519802#discussion_r3266523386) in Nixpkgs.  I was originally going to carve out an exception for Nix builds, but thought that this is better as a general fix.


d-torrance (approved):
Thanks!

PR #4340, opened 2026-05-21
Add accompanying paper to the GameTheory package

iremportakal:
Version 1.0 of the package has been reviewed and is on track for an acceptance recommendation in the Journal of Software for Algebra and Geometry.  As a new part of their technical review process (May, 2026), they ask that we submit a new pull request.  The [accompanying paper](https://arxiv.org/abs/2507.16755) to the package is added.

### DIFF
--- a/M2/Macaulay2/packages/GameTheory.m2
+++ b/M2/Macaulay2/packages/GameTheory.m2
@@ -69,7 +69,7 @@ export {
    "spohnCI"
 }
 
-
+-- The version 1.0 of the package has been reviewed and on the track for an acceptance recommendation in Journal of Software for Algebra and Geometry. 
 --***************************************--
 --   METHODS FOR CORRELATED EQUILIBRIA   --
 --***************************************--
@@ -817,7 +817,7 @@ doc ///
     Text
      {\bf Game Theory} is a package for several equilibrium concepts in game theory. It constructs the algebro-geometric and
      combinatorial models for Nash, correlated, dependency, and conditional independence equilibria. The latter three notions of
-     equilibria are all generalizations of Nash equilibria. An $n$-player game in normal form is defined by $n$-tensors of format
+     equilibria are all generalizations of Nash equilibria. The accompanying paper for this package is [@HREF("https://arxiv.org/abs/2507.16755","The GameTheory package for Macaulay2")@]. An $n$-player game in normal form is defined by $n$-tensors of format
      $d_1 \times d_2 \times \cdots \times d_n$, where $d_i$ is the number of pure strategies of player $i$.
      The entry $\{j_1, j_2, \cdots, j_n\}$ of the $i$-th (payoff) tensor for player $i$ is the payoff when player $1$ chooses
      strategy $j_1$, player $2$ chooses strategy $j_2$, and so on. One can define a specific game or a random game, e.g.,
--- a/M2/Macaulay2/packages/GameTheory.m2
+++ b/M2/Macaulay2/packages/GameTheory.m2
@@ -69,7 +69,7 @@ export {
    "spohnCI"
 }
 
--- The version 1.0 of the package has been reviewed and on the track for an acceptance recommendation in Journal of Software for Algebra and Geometry. 
+-- Version 1.0 of the package has been reviewed and is on track for an acceptance recommendation in the Journal of Software for Algebra and Geometry. 
 --***************************************--
 --   METHODS FOR CORRELATED EQUILIBRIA   --
 --***************************************--


iremportakal:
Regarding the failing check, it seems to be regarding another package, namely the SubalgebraBases package on macOS. Is there anything I should do?

d-torrance:
> Regarding the failing check, it seems to be regarding another package, namely the SubalgebraBases package on macOS. Is there anything I should do?

Nope - it's totally unrelated

M2/Macaulay2/packages/GameTheory.m2
d-torrance, line 37: We don't need *Polyhedra* in both `PackageImports` and `PackageExports`.  Just pick one - do we want to export its symbols or not?

M2/Macaulay2/packages/GameTheory.m2
```diff
@@ -69,7 +69,7 @@ export {
    "spohnCI"
 }
 
-
+-- Version 1.0 of the package has been reviewed and is on track for an acceptance recommendation in the Journal of Software for Algebra and Geometry.
```
d-torrance, line 72: This comment should probably be dropped since it would be out of date as soon as the package appears in JSAG.

M2/Macaulay2/packages/GameTheory.m2
d-torrance, line 100: In general, `append` inside a for loop isn't efficient since we create a brand new list at every iteration.  Try to refactor this to use `apply` or `for ... list` instead.

d-torrance, line 108: Is there a good reason to use mutable hash tables?  For example, this means we can't compare Tensor objects:

```m2
i24 : z = zeroTensor {2,2}

o24 = z

o24 : Tensor

i25 : w = zeroTensor {2,2}

o25 = w

o25 : Tensor

i26 : z === w

o26 = false
```

If they were regular immutable hash tables, then `===` would just compare the key/value pairs and we would get `true`.

d-torrance, line 126: If Tensor becomes immutable, then these constructor methods would need to be refactored, e.g.,:

```m2
new Tensor from (
    join(apply(indexset, i -> i => 0_R), {
    "format" => dims, 
     ... })
)
```

d-torrance: It would also be good if `Tensor` had a `net` method so we don't just get the default `MutableHashTable` one:

```m2
i2 : zeroTensor {2,2}

o2 = Tensor{...7...}
```

Even better, also add a `texMath` method for webapp support.

A slick way to do this is define an [`expression`](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/___Expression.html) method and then define:

```m2
net Tensor := net @@ expression
texMath Tensor := texMath @@ expression
mathML Tensor := mathML @@ expression
```

d-torrance, line 185: Here's another `append` in a for loop

d-torrance, line 200: Maybe there should be a more informative error message if the index isn't found.  Right now it's not clear what went wrong:

```m2

i11 :  R = QQ[p_{0,0}, p_{0,1}, p_{1,0}, p_{1,1}]

o11 = R

o11 : PolynomialRing

i12 : getVariableToIndexset(R, {3,3})
stdio:12:21:(3):[1]: error: no method for binary operator _ applied to objects:
            R (of class PolynomialRing)
      _     null (of class Nothing)
```

d-torrance, line 252: Here's another `append` inside a for loop

d-torrance, line 251: `poly` is already a reserved variable, so it would be good to use something else

(Nothing breaks since it's a local variable here, but it's good practice to avoid clashing names like this)

d-torrance, line 270: It would be good to do some checks here, e.g., if `X` is empty, then we'll get an error here when we try to find `X#0`.  So maybe something like:

```m2
if #X == 0 then error "expected a nonempty list"
```

We should also probably check that all the elements are tensors

d-torrance, line 272: One thing about using `getSymbol` is that it will overwrite existing variables, e.g.:

```m2
i15 : p = 5

o15 = 5

i16 : correlatedEquilibria randomGame{2,2,2}
warning: clearing value of symbol p to allow access to subscripted variables based on it
       : debug with expression   debug 1140   or with command line option   --debug 1140

o16 = Polyhedron{...1...}

o16 : Polyhedron

i17 : p

o17 = p

o17 : IndexedVariableTable
```

This isn't unheard of (`GF` does the exact same thing with the symbol `a`), but it would be good to document this because it can be surprising.

d-torrance, line 432: I don't think this method needs to be defined because `directProduct` is already a binary method, i.e., it will take 3 or more arguments and fold them, e.g., `directProduct(x,y,z) = `directProduct(directProduct(x,y), z)`.

d-torrance, line 470: Here's another `append` inside a for loop

d-torrance, line 500: Another `append` in a for loop

d-torrance, line 548: I see here is a place where we're using the mutability of tensors.  Maybe another constructor method would work here, if there were a switch to immutable tensors.

So maybe something like `tensor(Ring)`?

d-torrance, line 566: We don't need a function here - `sum vars` will work

d-torrance, line 636: Rather than using `fold` and `||` here, we can just use `matrix`, e.g.,:

```m2
i30 : matrix apply(4, i -> {random(ZZ^2, ZZ^2)})

o30 = | 2 2 |
      | 1 2 |
      | 7 3 |
      | 1 7 |
      | 9 5 |
      | 4 1 |
      | 3 1 |
      | 8 9 |
```

d-torrance, line 736: Did you mean to `return` here?

d-torrance, line 2069: The example for this documentation node takes a long time:

```m2
 -- capturing example results for "spohnCI"                                  -- 91.3713s elapsed
```

Is there a shorter example that could be used?

d-torrance (changes requested):
Did a first pass review with @klee669

We only got through the code.  Expect possibly some more comments related to documentation and/or tests

M2/Macaulay2/packages/GameTheory.m2
d-torrance, line 4: The version number and date should be updated

d-torrance (changes requested):

M2/Macaulay2/packages/GameTheory.m2
klee669: $d_i$ is the number of strategies for the $i$-th player, so $j\leq d_i-1$ is correct, not $d_j-1$.

klee669, line 1320: Another $d_j-1$.

klee669, line 1318: $d_i$ is the number of strategies for the $i$-th player, so $j\leq d_i-1$ is correct, not $d_j-1$.

klee669 (changes requested):

M2/Macaulay2/packages/GameTheory.m2
HannahTillmannMorris: No - we've separated out this case because when I + V = ideal(1) then we don't need to saturate it, whereas otherwise we do. In both cases we want to return I + V saturated, not terminate the function.

HannahTillmannMorris: We've found which example is taking so long: it's the first of the two examples for spohnCI with the vertices relabelled. If we change d from (2,3,2) to (2,2,2) in line 2128, this makes the example run much quicker.

klee669, line 854: The direction here seems reversed. The Segre embedding maps the space of independent mixed strategies into the space of joint probability distributions. Could this sentence be rephrased accordingly?

klee669, line 1195: Xi is the payoff tensor of player i, not a tensor of strategy probabilities;

klee669, line 1234: Another Xi issue.

klee669, line 1568: The description says that when given a list, this function first calls probabilityRing, but only the Ring method is implemented and documented. 

```M2
i5 : probabilitySumIdeal {2,3,4}
stdio:6:19:(3):[1]: error: no method found for applying probabilitySumIdeal to:
     argument   :  {2, 3, 4} (of class List)
```

probabilitySumIdeal {2,3,4} currently produces a “no method found” error.

klee669 (commented):
I've run the second round of the review, which has been mostly focused on documents. Still a bit more to come soon.

M2/Macaulay2/packages/GameTheory.m2
d-torrance: This item was resolved in #4726

PR #4348, opened 2026-05-21
Debugger finish

trevorkarn:
Add the debugger functionality developed at the 2026 Macaulay2 Atlanta workshop.

This:

- fixes the functionality of `step` to work with `breakpoint`
- adds the `finish` command to the debugger
- adds to the documentation 

Joint with @jkyang92 and @Rick3yHuang.


M2/Macaulay2/d/binding.d
```diff
@@ -326,7 +326,8 @@ bumpPrecedence();
      export breakS         := special("break",        nunaryop, precSpace, wide);
      export continueS      := special("continue",     nunaryop, precSpace, wide);
      export stepS          := special("step",         nunaryop, precSpace, wide);
-     special("new",   unarynew,   precSpace, narrow);
+     export finishS        := special("finish",       nunaryop, precSpace, wide);
+	 special("new",   unarynew,   precSpace, narrow);
```
d-torrance: Would you put the `new` indentation back to the way it was?

d-torrance (changes requested):

M2/Macaulay2/d/evaluate.d
```diff
@@ -1537,6 +1538,10 @@ augmentedAssignmentFun(x:augmentedAssignmentCode):Expr := (
 
 -----------------------------------------------------------------------------
 steppingFurther(c:Code):bool := steppingFlag && (
+	 if finishTargetDepth > 0 then (
+		if recursionDepth > finishTargetDepth then return true;
+		return false;
+		);
```
mahrud: This piece can be simplified:
```suggestion
    if finishTargetDepth > 0 then return recursionDepth > finishTargetDepth;
```
(Also the standard is 4-space indents and 8-space tabs.)

M2/Macaulay2/d/evaluate.d
```diff
@@ -1603,6 +1610,11 @@ handleError(c:Code,e:Expr):Expr := (
 					);
 				   eval(c))
 			      else if z.message == continueMessage then eval(c)
+				  else if z.message == finishMessage then (
+					finishTargetDepth = recursionDepth - 1;
+					setSteppingFlag();
+					eval(c)
+				  )
```
mahrud: Very minor, but it looks like the second `else if` is inadvertently indented:
```suggestion
			      else if z.message == finishMessage   then (
				  finishTargetDepth = recursionDepth - 1;
				  setSteppingFlag();
				  eval(c)
				  )
```

M2/Macaulay2/d/evaluate.d
```diff
@@ -1874,6 +1887,14 @@ breakFun(a:Code):Expr := (
      when e is Error do e else Expr(Error(dummyPosition,breakMessage,e,false,dummyFrame)));
 setupop(breakS,breakFun);
 
+export finishFun(a:Code):Expr := (
+     e := if a == dummyCode then nullE else eval(a);
+     when e is Error do e else (
+	  Expr(Error(dummyPosition,
+	       finishMessage,
+	       e,false,dummyFrame))));
```
mahrud: 1. Since `finishFun` is not used in any other files in the `d/` directory, it doesn't need to be exported.
2. Just for the sake of consistency with `breakFun` above, I would do this, but it's very minor change and it's fine if you don't agree:
```suggestion
finishFun(a:Code):Expr := (
     e := if a == dummyCode then nullE else eval(a);
     when e is Error do e else Expr(Error(dummyPosition,finishMessage,e,false,dummyFrame)));
```

M2/Macaulay2/d/interp.dd
```diff
@@ -489,6 +490,7 @@ setupfun("commandInterpreter",commandInterpreter);
 currentS := setupvar("current",nullE);
 debugger(f:Frame,c:Code):Expr := (
      -- stdIO << "-- recursionDepth = " << recursionDepth << endl;
+	 if finishTargetDepth != -1 then finishTargetDepth = -1;
```
mahrud: Why not just set it instead of checking and then setting?
```suggestion
     finishTargetDepth = -1;
```

M2/Macaulay2/d/evaluate.d
```diff
@@ -22,6 +22,7 @@ threadLocal export backtrace := true;
 threadLocal export profiling := false;
 threadLocal lastCode := dummyCode;
 threadLocal lastCodePosition := Position("",ushort(0),ushort(0),ushort(0),ushort(0),ushort(0),ushort(0),ushort(0));
+export threadLocal finishTargetDepth := -1;
```
mahrud, line 25: Just an observation: I wonder if `export threadLocal` and `threadLocal export` are treated differently by scc.

M2/Macaulay2/d/evaluate.d
```diff
@@ -1861,7 +1871,7 @@ continueFun(a:Code):Expr := (
 	       e,false,dummyFrame)));
 setupop(continueS,continueFun);
 
-stepFun(a:Code):Expr := (
+export stepFun(a:Code):Expr := (
```
mahrud, line 1874: I just noticed stepFun is also only used here and therefore doesn't need to be exported, but it's harmless, so only fix this if there's something else that really needs fixing.

mahrud (approved):

M2/Macaulay2/packages/Macaulay2Doc/ov_debugging.m2
```diff
@@ -602,6 +604,27 @@ document {
      beforehand.",
      }
 
+document {
+  Key => "breakpoint",
+  Headline => "set a breakpoint to stop execution at a particular point in the code and enter the debugger",
+  Inputs => { "s" => String => "a string containing Macaulay2 code" },
```
d-torrance, line 610: Note for a future PR:  I don't think this is accurate.  We can put `breakpoint` in front of any M2 code, not just a string

mahrud: Ah I missed this. It should probably be documented similar to `profile` or `time`, ie. not specifying the type of input.

PR #4349, opened 2026-05-22
Minor output routine updates

pzinn:
This is a minor update to `latex.m2`:
- a workaround for a bug in KaTeX is no longer needed since the bug was fixed
- added a couple of new `texMath` for certain keywords
- there was a bug in a regex for parsing variable names that I fixed
- I added `texMath CCi`

Also a small addendum to #4272 -- missed `net` of a `null` inside a `Hypertext`

### DIFF
--- a/M2/Macaulay2/m2/format.m2
+++ b/M2/Macaulay2/m2/format.m2
@@ -20,13 +20,16 @@ needs "nets.m2"
 -- Common utilities for formatting documentation nodes
 -----------------------------------------------------------------------------
 
+-- skip Options; TODO: define parser Option := null instead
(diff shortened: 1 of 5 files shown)


mahrud:
It would help with reviewing this PR if you could add a few tests that fail without it! There's of course the side benefit that it won't break by accident again :)

pzinn:
will do

pzinn:
added some tests for both this PR and #4272 -- as well as lots more for `texMath` since the `texmath.m2` test file for practically empty.

mahrud (approved):
Thank you! I think this looks good to me.

d-torrance (approved):
Thanks!

PR #4351, opened 2026-05-22
fixed position of nunaryops like 'return'

### LINKED ISSUE #4329: `code methods positions` is broken
What is going on here? Why is the position to `0:0`?
```m2
i1 : code methods positions

o1 = -- code for method: positions(VisibleList,Function)
     -- code for method: positions(MutableList,Function)
     ../linuxbrew/.linuxbrew/Cellar/macaulay2/1.26.05/share/Macaulay2/Core/lists.m2:86:36-0:0: --source code:
```

Here is the source code:
https://github.com/Macaulay2/M2/blob/c6aff10750a8b8a7e221f04c72d3d8b90c902ea9/M2/Macaulay2/m2/lists.m2#L84-L86

@pzinn @jkyang92 seems like a code position error?

mahrud:
This fixes #4329 (and deletes a piece of dead code and adds a couple of comments).

Before:
```m2
i1 : code (() -> return)

o1 = stdio:1:6-0:0: --source code:

i2 : code (() -> if true then 1 else continue)

o2 = stdio:2:6-0:0: --source code:

```
After:
```m2
i1 : code (() -> return)

o1 = stdio:1:6-1:19: --source code:
     code (() -> return)

i2 : code (() -> if true then 1 else continue)

o2 = stdio:2:6-2:41: --source code:
     code (() -> if true then 1 else continue)
```

Subtle side effect: entering this fails:
```m2
code (() -> return)

code (() -> if true then 1 else continue)

```
with error:
```m2
i1 : code (() -> return)

code (() -> if true then 1 else continue)

o1 = stdio:1:6-1:19: --source code:
     code (() -> return)
stdio:3:4:(3):[1]: error: line number 1 not found in file stdio
```
cc: @jkyang92 @pzinn

### DIFF
--- a/M2/Macaulay2/d/parser.d
+++ b/M2/Macaulay2/d/parser.d
@@ -215,10 +215,12 @@ export unaryop(token1:Token,file:TokenFile,prec:int,obeylines:bool):ParseTree :=
      if ret == errorTree then ret
      else accumulate(ParseTree(Unary(token1,ret)),file,prec,obeylines));
 export nunaryop(token1:Token,file:TokenFile,prec:int,obeylines:bool):ParseTree := (
+    -- e.g. return, continue, etc.
      ret := nparse(file,token1.word.parse.unaryStrength,obeylines);
      if ret == errorTree then ret
      else accumulate(ParseTree(Unary(token1,ret)),file,prec,obeylines));
 export nnunaryop(token1:Token,file:TokenFile,prec:int,obeylines:bool):ParseTree := (
+    -- e.g. comma
      if token1.word.parse.precedence <= prec
      then errorunary(token1,file,prec,obeylines)
      else (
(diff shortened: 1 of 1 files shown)


M2/Macaulay2/d/parser.d
```diff
@@ -556,7 +559,7 @@ export treePosition(e:ParseTree):Position := (
 	if t.newInitializer != dummyTree then t.newInitializer else
 	if t.newParent      != dummyTree then t.newParent      else t.newClass;
 	combinePositionL(t.newToken.position, treePosition(lastClass)))
-    is dummy do dummyPosition
+    is t:dummy do t.position
```
mahrud, line 562: The point is that in line 259 we return `ParseTree(dummy(token.position))` so even a dummy object may contain useful information.

pzinn:
where does that last error come from?

jkyang92:
Another possible fix for this is in the output of `treePosition` and/or `combinePositionL` and variations by not taking position values from dummy objects, since `convertr.d` calls `treePosition` to get the positions for `Code`. I don't think this resolves your weird error, which I think has more to do with how/where we are storing lines for stdio. In particular I don't know why it's always asking for line 1

mahrud:
@pzinn try copy and pasting this whole code block (not line by line):
```m2
code (() -> return)

code (() -> if true then 1 else continue)

```

@jkyang92 I prefer my one liner change :)

> I don't think this resolves your weird error, which I think has more to do with how/where we are storing lines for stdio. In particular I don't know why it's always asking for line 1

Yeah, I have no idea why it's always line 1!

If neither of you think this is super high priority to fix I can make this ready to merge.

d-torrance:
The line 1 thing is probably my fault... We use readline's history to get the code of stuff on stdio.  I'll take a look

d-torrance (approved):
Nice fix!  I'll open an issue for the weird error

PR #4352, opened 2026-05-22
Add .gitmodules to .gitignore

trevorkarn:
Add `.gitmodules` to `.gitignore` so that users can put submodules (for example https://github.com/Macaulay2/vscode-Macaulay2 or https://github.com/Macaulay2/Workshop-2026-Atlanta) inside of their `M2` directory.

Of course, this does mean that when a new git submodule really should be added one must manually add `.gitmodules` but this doesn't seem like a large burden or a frequent occurance.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@
 .DS_Store
 .#*
 .*.swp
+.gitmodules
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 /M2/Macaulay2/m2/TAGS.doc
 /M2/build-vscode
 /M2/BUILD/*/builds.tmp
+/M2/BUILD/build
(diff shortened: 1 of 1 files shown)


.gitignore
mahrud, line 4: While you're at this, could you also add `/M2/BUILD/build`?
```suggestion
/M2/BUILD/*/builds.tmp
/M2/BUILD/build
```

mahrud: I can also do this later tho so if this commit gets merged as a part of your other PR don't worry about it.

d-torrance: There are a few files under version control in `M2/BUILD/build`.  There's a `.gitignore` there, though, so I think that would be the place to put anything specific we want to ignore.

PR #4357, opened 2026-05-23
Reject zero substitutions in Laurent denominators

pzinn:
I forget if I ever submitted an issue:
```
   R=QQ[x,Inverses=>true,MonomialOrder=>Lex]
   f=matrix{{x^-1,1/2}}
   sub(f,x=>0)
```
will happily produce `matrix{{0,1/2}}` instead of throwing an error. Fixed with test.

# AI Disclosure

codex helped me diagnose this.

### DIFF
--- a/M2/Macaulay2/e/ringmap.cpp
+++ b/M2/Macaulay2/e/ringmap.cpp
@@ -5,6 +5,7 @@
 #include "matrices/matrix-con.hpp"
 #include "rings/polyring.hpp"
 #include "ring-elements/ring-element.hpp"
+#include "exceptions.hpp"
 
 #include <iostream>
 RingMap::RingMap(const Matrix *m) : R(m->get_ring())
@@ -143,12 +144,17 @@ ring_elem RingMap::eval_term(const Ring *sourceK,  // source coeff ring
                              int first_var,
                              int nvars_in_source) const
 {
+  bool result_is_zero = false;
   for (index_varpower i = vp; i.valid(); ++i)
     {
       int v = first_var + i.var();
       if (v >= nvars || _elem[v].is_zero)
-        return R->from_long(0);  // The result is zero.
+        {
+          if (i.exponent() < 0) throw exc::division_by_zero_error();
+          result_is_zero = true;
+        }
     }
+  if (result_is_zero) return R->from_long(0);
 
   // If K is a coeff ring of R, AND map is an identity on K,
   // then don't recurse: use this value directly.
--- a/M2/Macaulay2/tests/normal/subst.m2
+++ b/M2/Macaulay2/tests/normal/subst.m2
@@ -4,6 +4,17 @@ I = matrix{{x^2,y^2}}
 P=matrix({{2,3,4}})
 substitute(f,P)
 substitute(I,P)
+
+S = QQ[t,Inverses=>true,MonomialOrder=>Lex]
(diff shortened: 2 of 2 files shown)


pzinn:
Interestingly, this PR unearthed a bug in an example of `InvariantRing`:
https://www.unimelb-macaulay2.cloud.edu.au/usr/share/doc/Macaulay2/InvariantRing/html/_equivariant__Hilbert__Series.html
Note that last substitution at z_0, z_1 = 0.
@LuigiFerraro88 @galettof @fragandi @whuang259 @mnmastro @xlni

mahrud (changes requested):
This seems reasonable to me, but I fear the chance of conflicting with the two engine refactoring PRs is significant at the moment, so it would make sense to wait until those are merged first, then apply this small fix.

galettof:
Of course this makes sense mathematically. However, it was a convenient "feature" to able to use `sub` to extract the part of the polynomial that is constant in the z variables. I have some code that works but it is not particularly elegant.
```
R = QQ[x_1..x_3]
W = matrix{{-1,0,1},{0,-1,1}}
T = diagonalAction(W, R)
E = equivariantHilbertSeries(T, Order => 7)
(Z,phi) = flattenRing ring E;
(product toList coefficients(phi E,Monomials=>apply(7,i->Z_0^i)))_(0,0)
```
Surely, there must be a better, less ad-hoc way to achieve something like this. Any suggestions?

M2/Macaulay2/e/ringmap.cpp
```diff
@@ -5,6 +5,7 @@
 #include "matrix-con.hpp"
 #include "polyring.hpp"
 #include "relem.hpp"
+#include "exceptions.hpp"
```
d-torrance, line 8: There are some conflicts now (the paths of the headers in the `#include` statements immediately above this line have changed)

d-torrance (changes requested):

M2/Macaulay2/e/ringmap.cpp
```diff
@@ -147,7 +148,10 @@ ring_elem RingMap::eval_term(const Ring *sourceK,  // source coeff ring
     {
       int v = first_var + i.var();
       if (v >= nvars || _elem[v].is_zero)
-        return R->from_long(0);  // The result is zero.
+        {
+          if (i.exponent() < 0) throw exc::division_by_zero_error();
+          return R->from_long(0);  // The result is zero.
```
d-torrance: This can return too early.  For example:

```m2
i1 : R = QQ[x,y, Inverses => true, MonomialOrder => Lex]

o1 = R

o1 : PolynomialRing

i2 : sub(x^(-1)*y, {x => 0, y => 0})

o2 = 0

o2 : QQ
```

d-torrance (changes requested):
Reviewed with @local-ring.  One big comment

M2/Macaulay2/e/ringmap.cpp
pzinn: good catch

pzinn:
not sure what's going on with the failing test. works fine locally. maybe just restart it?

pzinn:
anything else I need to do? or can this be merged?

d-torrance (approved):
Looks good to me.  @local-ring - any comments?

Issue #4356, opened 2026-05-22
"line number not found in file stdio" when running "code"


d-torrance:
After #4351, we have:

```m2
i1 : code(() -> return)
code(() -> return)

o1 = stdio:1:5-1:18: --source code:
     code(() -> return)
stdio:2:4:(3):[1]: error: line number 1 not found in file stdio
```

Note that the two lines need to be entered together at the same prompt (e.g., copied and pasted) for this to fire.

pzinn:
on Macaulay2Web with 1.26.06 I'm regularly getting those cryptic
[stdio:111:4(3)](https://www.unimelb-macaulay2.cloud.edu.au/stdio#L111:C4): error: line number 1 not found in file stdio
somewhat randomly -- not sure what the pattern is.

d-torrance:
> **Written by Claude** (Claude Opus 5, via Claude Code), not by @d-torrance, whose account posted it -- please weigh it accordingly.

@pzinn — the pattern is **any input line after the first one in a pasted block**, which on
Macaulay2Web is most of them, since input arrives a block at a time. That is also why this needs the
two lines entered together.

It is not the positions. `locate` is right in every case:

```m2
i1 : peek locate(() -> return)

o1 = FilePosition{stdio, 1, 12, 1, 25, 1, 12}

i2 : peek locate(() -> return)

o2 = FilePosition{stdio, 2, 12, 2, 25, 2, 12}
```

It is the input history. When several lines arrive at one prompt, the whole block goes into history
entry 1 and the following entries are empty:

```m2
i3 : apply(1..4, i -> getHistory(i + historyOffset))

o3 = (debug Core                                     , , , )
      historyOffset
      apply(1..4, i -> getHistory(i + historyOffset))
      code(() -> return)
```

One entry holding four lines, then three empty strings. `code` reconstructs stdio source by stacking
`getHistory` over the line range in the `FilePosition` (`m2/code.m2`, the `filename === "stdio"`
branch), so for a function defined on line 2 it stacks `""`, and then

```m2
if #file < stop then error("line number ", toString stop, " not found in file ", filename);
```

fires with `0 < 1`.

So #4351 is the trigger rather than the cause: it gave `return` a position, so `code` now attempts a
lookup it used to skip. The history-indexing gap is older and independent of it, and anything else
that reaches this branch for a non-first line of a block will hit the same error.

[...]

PR #4358, opened 2026-05-23
fix code() failing with "line number not found in file stdio" on pasted input

d-torrance:
readline 8+ enables bracketed paste mode by default.  When the user pastes multiple lines at once, readline() can return a single string containing embedded newlines (e.g. "expr1\nexpr2") rather than one line per readline() call.

The old code called add_history() on the entire returned string, so the whole paste was stored as one history entry.  Meanwhile M2's parser incremented lineNumber for every '\n' it saw, creating a mismatch: after a two-line paste the session line counter was 2 but only one new history entry existed.

code.m2 maps M2 line numbers to history entries via
    getHistory(lineNumber + historyOffset)
so getHistory(2 + historyOffset) would call history_get() with an out-of-range index, return NULL, and ultimately produce:

    error: line number 1 not found in file stdio

Fix: instead of a single add_history(p), split p at each embedded '\n' and add each segment as its own history entry (save/restore the '\n' in-place).  For single-line input (the common case) this is identical to the previous behaviour.

Fixes: #4356 

## AI Disclosure

This was 100% Claude Code

### DIFF
--- a/M2/Macaulay2/d/M2lib.c
+++ b/M2/Macaulay2/d/M2lib.c
@@ -175,7 +175,24 @@ static int read_via_readline(char *buf,int len,char *prompt) {
     if (p == NULL) return 0;	/* EOF */
     i = 0;
     plen = strlen(p);
-    add_history(p);
+    /* readline 8+ with bracketed paste mode can return a multi-line string
+       when the user pastes text containing newlines.  add_history on such a
+       string would store one entry for the whole paste, but M2's line counter
+       increments for every '\n' it sees, breaking the 1:1 correspondence that
+       code.m2 relies on when fetching history by line number.  Split at each
+       embedded newline so every logical line gets its own history entry. */
+    {
+      char *s = p;
+      char *nl;
+      while ((nl = strchr(s, '\n')) != NULL) {
+	char saved = *nl;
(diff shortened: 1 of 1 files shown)


mahrud:
I don't think Claude identified the correct solution. We don't want to split user's input into several lines. It is a _feature_ that if I paste multiple lines, then want to redo the same thing, I can go up until I get to the **full** multiline input and enter it again.

I'm not even sure Claude identified the correct problem. e.g. this works fine:
```m2
i1 : g = () -> (
    L = {};
    for i to 4000 do (
        L = append(L, i)
        )
    )

o1 = g

o1 : FunctionClosure

i2 : code g

o2 = stdio:1:4-6:5: --source code:
     g = () -> (
         L = {};
         for i to 4000 do (
             L = append(L, i)
             )
         )
```
Which indicates that lines 1 through 6 were correctly retrieved from history.

pzinn:
I'm a little confused what this has to do with `readline` -- neither emacs nor WebApp mode use `readline`, and I've definitely encountered some version of this bug.

d-torrance:
They do use readline a bit -- in particular for grabbing the code of things defined in stdio

Issue #2232, opened 2021-09-08
degrees ring not being cached appropriately


mikestillman:
The following assertion fails.  These should be the same rings.

```
R = ZZ/101[a]
degreesRing R
degreesRing 1
assert(degreesRing R === degreesRing 1) -- fails
```

DanGrayson:
Yes, they should be the same.

PR #4359, opened 2026-05-23
Allow more Core tests to pass via "capture"

d-torrance:
When running `check Core`, each test is first attempted via `capture` (in-process) before falling back to an external M2 subprocess. This PR fixes several tests that were failing the capture step unnecessarily, due to two distinct root causes:

- `ext-total.m2` set a global debugging flag and never reset it, causing subsequent tests in the same process to fail spuriously.
- Several tests relied on `use ring` to install named ring generators as global variables, which does not work correctly in the `capture` context due to how it manages symbol dictionaries. These tests have been rewritten to reference ring generators directly.

After this, all `Core` tests should pass on the first try, i.e., no more "capture failed" messages.

## AI Disclosure

Claude Code did all the hard work.  I just copy-pasted error messages and it took it from there.


M2/Macaulay2/tests/normal/ext-total.m2
```diff
@@ -84,6 +84,8 @@ comp (degs,heft0,degs1,degs2,degDB)
 t = tally degrees target presentation E
 assert ( t === u )
 
+flagInhomogeneity = false
```
mahrud, line 87: `flagInhomogeneity` should probably be added to the no-capture regex list.

M2/Macaulay2/tests/normal/poincare2.m2
```diff
@@ -4,7 +4,7 @@ assert( degrees source vars R === degrees source vars r )
 assert( degrees R^{{2,3},{4,1},{1,0},{0,1}} === degrees r^{{2,3},{4,1},{1,0},{0,1}} )
 assert( degree x === degree X )
 assert( degree y === degree Y )
-f = map(ring poincare r, ring poincare R)
+f = map(ring poincare r, ring poincare R, generators ring poincare r)
```
mahrud, line 7: Why did this fail without specifying the generators?

mahrud:
I agree that replacing `use R` with `R_i` fixes the problems, but I think there's utility in making sure that 'use R' and 'x_i' and things like that continue to work and don't break unintentionally. I would consider adding `no-capture-flag` instead for some of these, but don't feel strongly about it.

M2/Macaulay2/tests/normal/ext-total.m2
d-torrance: It's not exported, so I can't imagine too many tests would use it

d-torrance:
> I agree that replacing `use R` with `R_i` fixes the problems, but I think there's utility in making sure that 'use R' and 'x_i' and things like that continue to work and don't break unintentionally. I would consider adding `no-capture-flag` instead for some of these, but don't feel strongly about it.

Good point.   After these changes, `use` still appears 110 times in the Core tests, so I think it's probably well tested!

mahrud:
Sorry, let me rephrase: it's worth thinking whether there's a fix in Core (involving use/global symbols/etc) we could make to fix this capture errors. (I mentioned no-capture-flag earlier because to me every no-capture-flag indicates a bug that should be fixed in the future)

added: these may be very very edge cases, but I'm sure some users come upon them and don't know what's going on.

M2/Macaulay2/tests/normal/poincare2.m2
d-torrance: Great question!  Here's the test failure:

```m2
 -- i1 : -- test source: ../tests/normal/poincare2.m2:1:1-50:7
 --      r = QQ[x,y,Degrees => {{2,1},{1,3}}, Heft => {1,1}]
 -- 
 -- o1 = r
 -- 
 -- o1 : PolynomialRing
 -- 
 -- i2 : R = QQ[X,Y,Degrees => {{2,1},{1,3}}]
 -- 
 -- o2 = R
 -- 
 -- o2 : PolynomialRing
 -- 
 -- i3 : assert( degrees source vars R === degrees source vars r )
 -- 
 -- i4 : assert( degrees R^{{2,3},{4,1},{1,0},{0,1}} === degrees r^{{2,3},{4,1},{1,0},{0,1}} )
 -- 
 -- i5 : assert( degree x === degree X )
 -- 
 -- i6 : assert( degree y === degree Y )
 -- 
 -- i7 : f = map(ring poincare r, ring poincare R)
 -- 
 -- o7 = map (ZZ[T ..T ], ZZ[T ..T ], {0, 0})
 --               0   1       0   1
 -- 
 -- o7 : RingMap ZZ[T ..T ] <-- ZZ[T ..T ]
 --                  0   1          0   1
 -- 
 -- i8 : assert( poincare r === f poincare R )
 -- 
 -- i9 : assert( poincare (r/(x)) === f poincare (R/(X)) )
 -- currentString:10:6:(3):[14]: error: assertion failed
```

I'm guessing that `map(Ring, Ring)` runs into some shenanigans with the symbol `T`.

d-torrance:
Good point.  I think the underlying bug for most of these might be https://github.com/Macaulay2/M2/issues/2232.

M2/Macaulay2/tests/normal/ext-total.m2
mahrud: Oh wait I just saw there's `debug Core` at the top of these file, shouldn't `debug` automatically disable `capture`?

mahrud: Hmm I guess the side effects of `debug` itself are usually reverted by other stuff that `capture` does, but also `debug Core` enables a whole bunch of shenanigans we don't look for ... maybe `debug Core` should specifically force not using `capture`? idk!

PR #4360, opened 2026-05-23
Fix misleading documentation in breakpoint

trevorkarn:
See the comments in #4348


mahrud (approved):
lgtm, but it's still a draft.

PR #4362, opened 2026-05-23
Use markdown for links when possible

d-torrance:
When an `ANCHOR` object contains only "href" and possibly "title" attributes, then it's possible to use pure markdown to render the link, e.g.:

* `<a href="foo">bar</a>` ⟶ `[bar](foo)`
* `<a href="foo" title="baz">bar</a>` ⟶ `[bar](foo "baz")`

We do this when possible in the (unexported) `markdown(ANCHOR)` method and update `markdown(TO)` and `markdown(TO2)` accordingly.  (We also add a helper function for converting `TO` ⟶ `TO2` since we do the same thing in several places.)

This significantly improves the appearance of the "signature help" feature in the latest draft of the language server when using Eglot in Emacs.  

### Before

```m2
i1 : debug Core

i2 : markdown TO rank

o2 = <a title="compute the rank" href="/usr/share/doc/Macaulay2/Macaulay2Doc/html/_rank.html">rank</a>
```

#### Using the language server in Emacs

image

### After

```m2
i1 : debug Core

i2 : markdown TO rank

o2 = [rank](/usr/share/doc/Macaulay2/Macaulay2Doc/html/_rank.html "compute the rank")
```

#### Using the language server in Emacs

image

## AI Disclosure

This one was pretty much all me.  I did have Claude Code review my code.  It suggested a few changes, but I disagreed with them and it was convinced after I stated my case.

### DIFF
--- a/M2/Macaulay2/m2/format.m2
+++ b/M2/Macaulay2/m2/format.m2
@@ -289,7 +289,7 @@ infoTagConvert DocumentTag := tag -> (
 
 -- TODO: can this be simplified?
 -- checking if doc is missing can be very slow if node is from another package
-info TO  := x -> info TO2{x#0, format x#0 | if x#?1 then x#1 else ""}
+info TO  := x -> info TO2 x
 info TO2 := x -> (
      tag := fixup x#0;
      if isMissingDoc tag or isUndocumented tag then concatenate(x#1, " (missing documentation)")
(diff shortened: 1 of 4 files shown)


d-torrance:
I added another commit which removes the newline after `SPAN` objects (there was already a comment suggesting maybe we should do this).  This fixes the awkward break we had in the "Inputs" section in the screenshots above:

image

mahrud:
> it was convinced after I stated my case.

XD

mahrud (approved):
Looks reasonable. I'd ask for some tests, but I didn't write any so can't expect you to start.

PR #4363, opened 2026-05-23
Add locate(MethodFunction) and friends

d-torrance:
Returns a numbered vertical list of the locations of all the methods:

```m2
i1 : locate rank

o1 = {0 => (../packages/Complexes/ChainComplex.m2:689:16-690:50)}
     {1 => (../packages/Varieties.m2:342:25-342:46)             }
     {2 => (matrix1.m2:673:15-677:5)                            }
     {3 => (methods.m2:695:25-695:88)                           }
     {4 => (mutablemat.m2:311:22-316:5)                         }

o1 : NumberedVerticalList
```

This will be useful for the language server.

## AI Disclosure

Just some old school, human-only Macaulay2 code!

### DIFF
--- a/M2/Macaulay2/m2/debugging.m2
+++ b/M2/Macaulay2/m2/debugging.m2
@@ -244,6 +244,11 @@ locate Error       :=
 locate Symbol      := FilePosition => locate'
 locate Command     := FilePosition => C -> locate'(C#0)
 locate List        := List     => x -> apply(x, locate)
+locate MethodFunction            :=
+locate MethodFunctionBinary      :=
+locate MethodFunctionSingle      :=
+locate MethodFunctionWithOptions := List => (
+    f -> unique nonnull apply(methods f, locate'))
 protect symbol locate
 
 sortByLocation = sortBy(toString @@ locate)
--- a/M2/Macaulay2/packages/Macaulay2Doc/functions/locate-doc.m2
+++ b/M2/Macaulay2/packages/Macaulay2Doc/functions/locate-doc.m2
@@ -17,12 +17,16 @@ Node
    (locate, List)
    (locate, ZZ)
    (locate, Error)
+   (locate, MethodFunction)
+   (locate, MethodFunctionBinary)
+   (locate, MethodFunctionSingle)
+   (locate, MethodFunctionWithOptions)
   Headline
     locate source code
   Usage
     locate f
   Inputs
-    x:{Function,Sequence,Symbol,List,ZZ}
+    f:{Function,Sequence,Symbol,List,ZZ}
   Outputs
     :{FilePosition,List,Nothing}
       {\tt {filename, start,startcol, stop,stopcol, pos,poscol}}, respectively
@@ -43,9 +47,15 @@ Node
     Example
       locate(jacobian, Matrix)
     Text
-      In particular, the output of @TO "methods"@ can be used as an input to @TT "locate"@.
+      If @TT "f"@ is a method function, then a list is returned containing the
+      locations of all the corresponding methods.
     Example
-      locate methods jacobian
+      locate jacobian
+    Text
+      If @TT "f"@ is an integer, then the location of the function at the given
+      position in the last call to @TO methods@, @TO hooks@, or @TO tests@ is
+      returned.
+    Example
       methods doc
       locate 0
       peek oo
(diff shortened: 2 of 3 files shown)


mahrud:
I think at some point I considered this, but I personally don't like it because you get different types of output:
```m2
i1 : locate headlines

o1 = {0 => (../../Macaulay2/m2/help.m2:412:18-415:63)}

i2 : locate synonym

o2 = ../../Macaulay2/m2/classes.m2:12:10-12:59

o2 : FilePosition
```
I think it would be lovely if `locate headlines` gave the location of `headlines = method()`, like this:
```m2
i1 : locate symbol headlines

o1 = ../linuxbrew/.linuxbrew/Cellar/macaulay2/1.26.05/share/Macaulay2/Core/help.m2:406:0-406:9
```

PS: of course, the trouble is that for annoying reasons, if the symbol appears in a file that is loaded before the method is defined, you get a random location smh ...

mahrud:
Oh also, pedagogically I think of the following as the equivalent of piping POSIX tools for M2:
```m2
locate methods rank
code methods rank
headlines methods rank
makeDocumentTag methods rank
locate makeDocumentTag methods rank
etc.
```

d-torrance:
Yeah, I'm not a huge fan of the differing output types either.  I wouldn't have proposed it if we didn't have `locate(Package)` already, which also returnes a numbered vertical list.

But it's really nice for language server support :).  The "definition" method can return either a location or a list of locations, so this works great.  I can roll my own method if you feel strongly

mahrud:
I don't feel super strongly, but I don't follow the connection with `locate(Package)`. Also, can you say more about why it's useful for the language server?

d-torrance:
`locate(Package)` returns all the package's files:

```m2
i1 : locate Polyhedra

o1 = {0 => (/usr/share/Macaulay2/Polyhedra.m2:0:0)                                                       }
     {1 => (/usr/share/Macaulay2/Polyhedra/exports.m2:0:0)                                               }
     {2 => (/usr/share/Macaulay2/Polyhedra/symbols.m2:0:0)                                               }
     {3 => (/usr/share/Macaulay2/Polyhedra/helpers.m2:0:0)                                               }
     {4 => (/usr/share/Macaulay2/Polyhedra/engine/properties.m2:0:0)                                     }
     {5 => (/usr/share/Macaulay2/Polyhedra/engine/alternatives.m2:0:0)                                   }
     {6 => (/usr/share/Macaulay2/Polyhedra/core/globalMethods.m2:0:0)                                    }
     {7 => (/usr/share/Macaulay2/Polyhedra/core/cone/constructors.m2:0:0)                                }
     {8 => (/usr/share/Macaulay2/Polyhedra/core/cone/properties.m2:0:0)                                  }
     {9 => (/usr/share/Macaulay2/Polyhedra/core/cone/methods.m2:0:0)                                     }
     {10 => (/usr/share/Macaulay2/Polyhedra/core/fan/constructors.m2:0:0)                                }
     {11 => (/usr/share/Macaulay2/Polyhedra/core/polyhedron/constructors.m2:0:0)                         }
     {12 => (/usr/share/Macaulay2/Polyhedra/core/fan/methods.m2:0:0)                                     }
     {13 => (/usr/share/Macaulay2/Polyhedra/core/fan/properties.m2:0:0)                                  }
...
```

For LSP, the [textDocument/definition](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_definition) method can either return a location or a list of locations.  So for method functions, we could get a list of all the various locations of the installed methods and jump to them nicely.  (This works with <kbd>M-.</kbd> out of the box in Emacs.)

mahrud:
What happens if you click on a type or instance of a type?

d-torrance:
Currently nothing since `locate(Type)` doesn't exist.

mahrud:
I'm not sure exactly what's the end goal from a Macaulay2 IDE perspective, but I imagine separating two functionalities of looking up where a method is defined and declared (e.g. `blah = method()` vs `blah ZZ := ...`) is going to be useful at some point. I would be hesitant to conflate these two, potentially making it harder to go back later.

At the very least, don't document this behavior for now because people should use `locate methods jacobian` not `locate jacobian`.

d-torrance:
Fair enough -- I can just use `locate methods foo` in my language server code.  I'll close this

PR #4364, opened 2026-05-24
Modernize Vim plugin for Macaulay2

d-torrance:
Restructure the vim/ directory as a standard Vim plugin layout (ftdetect/, ftplugin/, syntax/, dict/) so files can be installed by copying directly without renaming.

Replace the screen/AppleScript-based M2 integration with Vim's built-in terminal (term_start/term_sendkeys, requires Vim 8.1+), eliminating the dependencies on screen, Terminal.app, and VimM2.scpt.

Merge the two platform-specific READMEs into a single README.md and update the installation instructions.

Update CMakeLists.txt and make-M2-symbols.m2 to generate files at their new paths (vim/syntax/m2.vim, vim/dict/m2.vim.dict).

## AI Disclosure

This was all Claude.  I'm an Emacs guy and wouldn't know where to begin.  But I tested it out and it works!


mahrud:
Would it make sense to allow a "Vim person" to modernize the plugin so that they can take ownership and maintain it going forward? e.g. the people who worked on it at the workshop?

d-torrance:
This is a distinct plugin (vim v. neovim), but I'll mention something in Zulip.  I did add a link to the neovim plugin repo in the README

dimpase:
why have you checked in `M2/Macaulay2/editors/vim/dict/m2.vim.dict`? Isn't it generated from `M2/Macaulay2/editors/vim/dict/m2.vim.dict.in` ?

d-torrance:
It is, but this way users don't need to regenerate it themselves

mahrud:
For what it's worth, the reason I mentioned it's better to have a Vim person take ownership is that dropping a Claude generated plugin on someone to maintain is burdening them with technical debt that they didn't ask for, which I don't think is fair even if they themselves would use AI to generate a plugin. In the end I'm not going to be using it so it's up to the people who would to decide if they want this or not.

dimpase:
> It is, but this way users don't need to regenerate it themselves

only if you consider the github repo as a distribution source. But once you start adding build artefacts into the repo, there is way too much generated junk which starts polluting the repo with frequently updated largish files, starts creating annoying merge conflicts, etc.

dimpase:
There is also nothing on how all these files are generated. I don't see anything besides
```
# TODO: generate vim grammar
```
in `M2/Macaulay2/editors/CMakeLists.txt`  - which makes me think that the magic incantations to invoke a template fill-in process aren't public info.

d-torrance:
Nice catch!  That `TODO` comment is out of date.

The magic incantation is in `M2/Macaulay2/editors/make-M2-symbols.m2`, which calls [`generateGrammar`](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Style/html/_generate__Grammar.html).

I suppose we could remove the generated files from git and add the `generateGrammar` instructions to the README for users to generate their own when installing.

dimpase:
Yes, please, this would be great

M2/Macaulay2/editors/vim/README.md
```diff
@@ -15,6 +15,16 @@ Syntax highlighting and key mappings to send code to a running Macaulay2 session
...
 ```
 
+Then, in Macaulay2, generate the symbol files.  You may also do this after a
+new Macaulay2 release to update the files with any new symbols.
+
+```m2
+needsPackage "Style"
+changeDirectory "~/.vim"
+generateGrammar("dict/m2.vim.dict", demark_" ")
+generateGrammar("syntax/m2.vim", demark_" ")
```
dimpase, line 25: Can we, instead for copying templates to `~/.vim/`, keep them in the M2 sources dir?
Then we would only need to `cp -r  ftdetect ftplugin ~/.vim/`.

I'd do the latter with [copyDirectory](https://www.macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_copy__Directory_lp__String_cm__String_rp.html), and wrap the whole thing in a small M2 function `installVimPlugin` of something like this.

This can be done in another PR, though, if you prefer.

d-torrance: Yeah, that makes sense.  Another PR sounds good.  Should we merge this and you can start from there?

dimpase: ok. By the way, how does one navigate M2 source tree from M2? I presume there's some build-in for the root of it?

dimpase (approved):

M2/Macaulay2/editors/vim/README.md
d-torrance: If you're running from a just-built copy of M2, then `topSrcdir` tells you the root directory.  It's unexported, so you have to `debug Core` or `importFrom(Core, "topSrcdir")`.  It's just `null` when running in a distributed copy of M2.

PR #4366, opened 2026-05-24
Update pushForward routine in Core to handle towers better

joel-dodge:
Some issues arise when computing `graphIdeal f`:
* when `source f` and `target f` have different coefficient rings. 
* when `f` is the inclusion of a coefficient ring and `target f` includes relations on the generators of `source f`.

Here are some examples that have improved behavior after this patch:

```
i1 : R = kk[a];
S = R[b, Join => false] / {a^2, b^2};
T = kk[t];
f = map(S, T, {a});
graphIdeal f

o4 : RingMap S <-- T
stdio:5:10:(3):[1]: error: expected polynomial rings over the same ring
```

```
i1 : R = kk[a];
S = R[b]/{a^2, b^2};
f = map(S, R);
graphIdeal f

o3 : RingMap S <-- R
o4 = ideal ()
o4 : Ideal of S
```

Also switches variable names `(R, S)` <> `(S, R)` in code to match direction of comments. I changed variables instead of comment since I agree with the comment direction in wanting my ring maps to go from R --> S.

### DIFF
--- a/M2/Macaulay2/m2/pushforward.m2
+++ b/M2/Macaulay2/m2/pushforward.m2
@@ -118,26 +118,26 @@ pushNonLinear := (opts, f, M) -> (
     -- returns R-presentation matrix for the pushforward of M
     -- written by Mike Stillman and David Eisenbud
 
-    -- first flatten R, S so that this computation works for towers.
+    -- first flatten source and target of f so that this computation works for towers.
     -- note this forces us to unflatten the return value when we are done.
-    (S, phiS) := flattenRing target f;
-    (R, phiR) := flattenRing source f;
-    m := phiS presentation M;
-    f = phiS * f * phiR^-1;
+    (R, phiR) := flattenRing target f;
+    (S, phiS) := flattenRing source f;
+    m := phiR presentation M;
+    f = phiR * f * phiS^-1;
 
     -- set up some variables that are used throughout
-    deglenR := degreeLength R;
-    numgensR := numgens R;
+    deglenS := degreeLength S;
     numgensS := numgens S;
+    numgensR := numgens R;
 
     monorder := opts.MonomialOrder;
-    monorder  = if ordertab#?monorder then (ordertab#monorder)(numgensS, numgensR)
+    monorder  = if ordertab#?monorder then (ordertab#monorder)(numgensR, numgensS)
     else error("pushForward: MonomialOrder option expected one of ",
 	demark_", " \\ toString \ keys ordertab);
 
     J := graphIdeal(f, MonomialOrder => monorder, VariableBaseName => local X);
     G := ring J;
-    xvars := map(G, S, submatrix(vars G, toList(0..numgensS - 1)));
+    xvars := map(G, R, submatrix(vars G, toList(0..numgensR - 1)));
     m1 := presentation (cokernel xvars m  **  cokernel generators J);
 
     if opts.UseHilbertFunction and all({f, m}, isHomogeneous) then (
@@ -148,15 +148,15 @@ pushNonLinear := (opts, f, M) -> (
 	-- cache poincare
 	poincare cokernel m1 = hf);
 
-    mapbackdeg := d -> take(d, -deglenR);
+    mapbackdeg := d -> take(d, -deglenS);
     -- that choice of degree map was chosen to make the symmetricPower functor homogeneous, but it doesn't have much
     -- else to recommend it.
-    -- we should really be *lifting* the result to R along the natural map R ---> G
-    mapback := map(R, G, map(R^1, R^numgensS, 0) | vars R, DegreeMap => mapbackdeg );
+    -- we should really be *lifting* the result to S along the natural map S ---> G
+    mapback := map(S, G, map(S^1, S^numgensR, 0) | vars S, DegreeMap => mapbackdeg );
 
     -- let's at least check it splits f's degree map:
-    for i from 0 to deglenR-1 do (
-	e := for j from 0 to deglenR-1 list if i === j then 1 else 0;
+    for i from 0 to deglenS-1 do (
+	e := for j from 0 to deglenS-1 list if i === j then 1 else 0;
 	if mapbackdeg f.cache.DegreeMap e =!= e
 	then error "not implemented yet: unexpected degree map of ring map");
 
@@ -164,8 +164,8 @@ pushNonLinear := (opts, f, M) -> (
 	StopBeforeComputation => opts.StopBeforeComputation,
 	DegreeLimit           => opts.DegreeLimit,
 	PairLimit             => opts.PairLimit);
-    -- MES: check if the monomial order restricts to S.  If so, then do `` forceGB result ''
-    phiR^-1 mapback selectInSubring(if numgensS > 0 then 1 else 0, generators g))
+    -- MES: check if the monomial order restricts to R.  If so, then do `` forceGB result ''
+    phiS^-1 mapback selectInSubring(if numgensR > 0 then 1 else 0, generators g))
 
 -*
 pushLinear := opts -> (f,M) -> (
--- a/M2/Macaulay2/m2/pushforward.m2
+++ b/M2/Macaulay2/m2/pushforward.m2
@@ -117,19 +117,19 @@ pushNonLinear := (opts, f, M) -> (
     -- given f: R --> S, and M an S-module, finite over R,
     -- returns R-presentation matrix for the pushforward of M
     -- written by Mike Stillman and David Eisenbud
-    (R, S) := (target f, source f);
-    deglen := degreeLength S;
-    n1 := numgens R; -- TODO: what if R is a tower?
+    (S, R) := (target f, source f);
+    deglen := degreeLength R;
+    n1 := numgens S; -- TODO: what if S is a tower?
 
     monorder := opts.MonomialOrder;
-    monorder  = if ordertab#?monorder then (ordertab#monorder)(numgens R, numgens S)
+    monorder  = if ordertab#?monorder then (ordertab#monorder)(numgens S, numgens R)
     else error("pushForward: MonomialOrder option expected one of ",
 	demark_", " \\ toString \ keys ordertab);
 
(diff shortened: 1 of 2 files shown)


joel-dodge:
@mahrud related to this I have been working my way through understanding how the core pushforward works and something funny I noticed is that `pushNonLinear` is sensitive tot he presentation of the module `M` that it is passed even though it is described as returning the presentation of it's argument `M` as a ` source f` module, I think that it only does this when `gens M` contains a generating set for `M` as a `source f` module. In particular, in this example

```
kk = ZZ/101
R = kk[a]
S = kk[s, t]/{s^2, t^2}
f = map(S, R, {s})
M = S^1
M1 = subquotient(matrix basis M, relations M)
```

(this copies loosely the fussing that the default pushforward strategy does)

we have that `M == M1` but

```
i60 : prune cokernel pushNonLinear(options pushForward, f, M1)
o60 = cokernel {0} | a2 0  |
               {1} | 0  a2 |

i57 : pushNonLinear(options pushForward, f, M)
o57 = | a2 |
```

does this check out with your understanding of the functioning of the core pushforward?  I am working towards a soft goal of understanding if / how the core pushforward can be folded into pushFwd and removed.

M2/Macaulay2/m2/pushforward.m2
```diff
@@ -117,19 +117,29 @@ pushNonLinear := (opts, f, M) -> (
...
     -- returns R-presentation matrix for the pushforward of M
     -- written by Mike Stillman and David Eisenbud
-    (R, S) := (target f, source f);
-    deglen := degreeLength S;
-    n1 := numgens R; -- TODO: what if R is a tower?
+    (S, R) := (target f, source f);
+
+    -- nudge source, target to have same coefficient ring for graphIdeal
+    (S', phiS) := try(flattenRing S) else (S, id_S);
+    (R', phiR) := try(flattenRing R) else (R, id_R);
```
mahrud: Does flattenRing ever return an error?
```suggestion
    (S', phiS) := flattenRing S;
    (R', phiR) := flattenRing R;
```

M2/Macaulay2/m2/pushforward.m2
```diff
@@ -117,19 +117,29 @@ pushNonLinear := (opts, f, M) -> (
     -- given f: R --> S, and M an S-module, finite over R,
     -- returns R-presentation matrix for the pushforward of M
     -- written by Mike Stillman and David Eisenbud
-    (R, S) := (target f, source f);
-    deglen := degreeLength S;
-    n1 := numgens R; -- TODO: what if R is a tower?
+    (S, R) := (target f, source f);
```
mahrud: Let's keep this as `S --> R`, but you can fix the comment a few lines above.
The reason is that S is often a **S**ymmetric algebra, aka polynomial ring, and R a quotient of it.

This also reduces some of the changes below.

M2/Macaulay2/m2/pushforward.m2
```diff
@@ -117,19 +117,29 @@ pushNonLinear := (opts, f, M) -> (
...
+
+    -- nudge source, target to have same coefficient ring for graphIdeal
+    (S', phiS) := try(flattenRing S) else (S, id_S);
+    (R', phiR) := try(flattenRing R) else (R, id_R);
+    if (S' =!= S) or (R' =!= R) then (
+	f' := phiS * f * phiR^-1;
+	M' := cokernel phiS presentation M;
+	return phiR^-1 pushNonLinear(opts, f', M');
+    );
```
mahrud: I'm wondering if this should happen here or in the main body of the `pushForward` function above. Two things to be aware of in deciding:
1. calling `presentation M` has a high cost, in this case it's fine because we need to call presentation a few lines below anyway, but it might not be necessary for all pushforward algortihms.
2. it's best to minimize the recursion depth. Even if we keep this here, doing it this way would achieve this:
3. 
```suggestion
    if (S' =!= S) or (R' =!= R) then (
	f = phiS * f * phiR^-1;
	M = cokernel phiS presentation M;
    );
```
And then further down change to `mapback := phiR^-1 * map(...`.

mahrud (changes requested):
Looks good, but left a few suggestions.

Note to self for next review: the most important thing I want to make sure this PR doesn't break is the ability to take advantage of the poincare polynomial in the gb computation, since it's a massive slowdown if it breaks.

M2/Macaulay2/m2/pushforward.m2
joel-dodge: You are confident that S -> R is better? I'll change it back following your suggestion but including what underpins my choice for context:

I always think of ring maps as going from R(ing) to S(next in line alphabetically). In a tower situation then we clearly have S -> T coming next.

joel-dodge: fwiw the codebase also prefers R -> S generally :D

```
➜  ~/code/M2 $ rg 'R [-]+> S' | wc -l
      27
➜  ~/code/M2 $ rg 'S [-]+> R' | wc -l
       6
```

M2/Macaulay2/m2/pushforward.m2
joel-dodge: Good catch I think I was just being cowardly!

M2/Macaulay2/m2/pushforward.m2
joel-dodge: Excellent suggestion. I moved the use of presentation up so that it is crystal clear that we must use it and we do so just once.

joel-dodge:
Removed the conditional check on whether flattening did anything since it is unnecessary and just makes it harder to reason about. 

I left the rings named R -> S pending resolution of the comment thread here. Edit: thread resolved. This PR no longer changes the variable names although oops maybe I forgot to update the comment now

M2/Macaulay2/m2/pushforward.m2
mahrud: At least in this case, I'm confident that the comment was an error. In particular, if you look in the main pushForward code, it says:
```m2
    if target f =!= R then error "expected module over the target of the ring map";
```
Having the same convention in both places makes debugging easier. Also, think S for **S**ource :)

joel-dodge: that works for me! I'll update.

joel-dodge: done

joel-dodge:
> Note to self for next review: the most important thing I want to make sure this PR doesn't break is the ability to take advantage of the poincare polynomial in the gb computation, since it's a massive slowdown if it breaks.

When you do, would you mind giving visibility into what steps you take to check this?

mahrud:
Oh I meant to push this to my fork and make a pull request to yours, but forgot to specify and it got pushed here.

I wanted to add a safety valve to make sure flattening the ring doesn't lose homogeneity, because if that happens gb will be much slower and the UseHilbertFunction option gets ignored.

mahrud (approved):
Once/if the tests pass, this is good to merge.

joel-dodge:
Excellent makes sense. Np on pushing straight here just hank you for explaining!

joel-dodge:
I'm also going to take some inspiration from @Devlin-Mallory on the other PR and do some timing tests with this before merging.

M2/Macaulay2/m2/pushforward.m2
```diff
@@ -113,17 +113,21 @@ ordertab := new HashTable from {
...
-    (R, phiR) := flattenRing target f;
-    (S, phiS) := flattenRing source f;
+    (R, phiR) := flattenRing target f0;
+    (S, phiS) := flattenRing source f0;
     m := phiR presentation M;
-    f = phiR * f * phiS^-1;
+    f := phiR * f0 * phiS^-1;
+
+    -- making sure that homogeneity isn't broken
+    if isHomogeneous M  then assert isHomogeneous m;
```
joel-dodge: The two things that a user wants out of the use of this function are I think: 1. correctness and 2. speed

I have introduced some changes to improve correctness. This check introduces a guard to ensure that it is not costing speed but I am not sure if I understand what the user ought to do in a situation where this assert fails. They certainly may *want* to proceed with a slow computation or they may want to proceed without flattening and hope that they are in one of the situations where flattening is not required for correctness. 

What would you think about something like:

```
stillHgs =  (not isHomogeneous M  or isHomogeneous m) and
    (not isHomogeneous f0 or isHomogeneous f);
if not stillHgs then (
  m = presentation M;
  f = f0;
  phiR = id_R
  -- emit some warning to the user that flattening failed
)
```

this would: 
- let the computation proceed without sacrificing speed
- inform the user that this computation *can* be incorrect in this case and give a hint that they can flatten on their own if they want to mitigate this

joel-dodge: what if there was something like

```
bug = (msg) -> (
  print msg;
  print "please let us know by creating an issue at https://github.com/Macaulay2/M2/issues."
)
```

joel-dodge: @mahrud wdyt about this update

mahrud: An assertion failure should be considered a internal error and reported. We could add a message like you suggest, but that should be done globally and that's outside the scope of this PR. I'd rather leave this as is and not complicate the code.

mahrud:
Did a test or example fail? Please remove the fallback. If automatically flattening the ring in pushForward doesn't work, then the user should flatten their rings before calling pushForward, in which case this PR is moot.

joel-dodge:
No test failed I just added this to illustrate what I meant. I will revert the PR adding the fallback. done!

joel-dodge:
@mahrud I think this is ready to merge and afaict that is something you must do.

mahrud:
My understanding was that this is now Doug's purview, but sure.

d-torrance:
It's all good!  We'll hash out some sort of PR-merging policy after the release.  I would have just merged it after your approval anyway!

joel-dodge:
Got it ty @mahrud I was unclear on procedure but that makes sense.

Issue #4368, opened 2026-05-25
check(41, "NumericalAlgebraicGeometry") fails when Bertini is installed


d-torrance:
```m2
i1 : check(41, "NumericalAlgebraicGeometry")
 -- running   check(41, "NumericalAlgebraicGeometry")                                                                                                                                                      -- running failed
 -- error log:  /tmp/M2-17591-0/1.tmp:0:1:(3):[15]: /usr/share/Macaulay2/NumericalAlgebraicGeometry/Bertini/Bertini.test.m2:29:14:(3):[5]: error: no method found for applying bertiniRefineSols to:
 --      argument 1 :  302 (of class ZZ)
 --                      2    2
 --      argument 2 :  {x  + y  - 1, x*y} (of class List)
 --      argument 3 :  {{1, 1e-8}} (of class List)
 -- 
 -- input code: /usr/share/Macaulay2/NumericalAlgebraicGeometry.m2:533:5-536:3
*** error: M2 exited with status code 1                                        
```

The method for `bertiniRefineSols` that we're trying to call was removed in #3121.

PR #4353, opened 2026-05-22
EuclideanDistanceDegree v1

whuang259:
Attached is a pull request for the `EuclideanDistanceDegree` package, coauthored with @JoseMath

Issue #1955, opened 2021-02-22
Run tests and examples that use optional components on GitHub


DanGrayson:
External programs upon which cached example output depends should probably be installed as part of our github workflow, so we can test those examples regularly.  Those programs include phcpack, bertini, qepcad, bergman, and polymake.

_Originally posted by @DanGrayson in https://github.com/Macaulay2/M2/issues/1953#issuecomment-783471096_

d-torrance:
> External programs upon which cached example output depends should probably be installed as part of our github workflow, so we can test those examples regularly. Those programs include phcpack, bertini, qepcad, bergman, and polymake.

FYI, polymake is already in Ubuntu, so that would be easy to add.  phcpack, qepcad, and bergman are all on my to-do list.  (bertini has a nonfree license, so it's off the table)

mahrud:
Does everything on the PPA also need to be free? How about creating a binary package and downloading it from macaulay2.com?

d-torrance:
Bertini has binary tarballs on their website, so we could just grab them from there, I suppose

d-torrance:
Here's the current state of packages that don't have optional components present on the GitHub builds:

* *Bertini* (#4369)
* *CoincidentRootLoci* (needs `qepcad`, currently installed in the Ubuntu builds but not the macOS ones)
* *MapleInterface* and the packages that depend on it (*AdjointIdeal*, *ConvexInterface*, *Parametrization*)
* *MonomialIntegerPrograms* (needs `scip`, which just recently appeared in Debian and would need to be backported to Ubuntu 24.04.  It's packaged for Homebrew but w/o ZIMPL support so it doesn't work for us)
* *NCAlgebra* (needs `bergman`, which is packaged for Ubuntu but not in Homebrew)
* *OnlineLookup* (maybe a false positive -- we cache the example output for offline builds)
* *PHCpack* and the packages that depend on it (*DecomposableSparseSystems*, *NumericalSchubertCalculus* -- phcpack is installed in the Ubuntu builds but not the macOS ones)
* *StatePolytope* (needs `polymake`, installed in Ubuntu builds but not the macOS ones)

PR #4369, opened 2026-05-25
Install Bertini in GitHub workflows

d-torrance:
This way, we can properly test packages that use it (e.g., #4353).  The download should be cached so we don't overload the Notre Dame servers.

I also took the opportunity to simplify the `apt-get install` step a bit -- there were a few packages we aren't using or are already installed on the GitHub Ubuntu runners.

Partially addresses #1955

Draft for now, as the tests will fail until #4368 is fixed.

## AI Disclosure

Claude Code wrote the code to download and cache Bertini.

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -79,13 +79,13 @@ jobs:
         run: |
           sudo add-apt-repository -y -n ppa:macaulay2/macaulay2
           sudo apt-get update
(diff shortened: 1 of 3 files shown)


d-torrance:
I added a fix for #4368 (mostly written by Claude).

antonleykin (approved):
Yes, bertini is somewhat restrictive in this sense. The PR fixes the issue for NumericalAG wrapper.

@JoseMath : for `Bertini` package it'd be good to have specific (foolproofing) `error`s produced in situations like this.

PR #4370, opened 2026-05-25
Update build of .dmg files

d-torrance:
We remove the old `--enable-dmg` option to configure and instead add a new recipe in `M2/BUILD/dmg`.  Rather than building all of the dependencies, we copy them over from the homebrew packages (or the distributed binaries in the case of phcpack and bertini).

We also add a new GitHub workflow to build dmg files for both Intel and ARM macOS at each release.  Mike has been test-driving one of these for the last day or so with positive reports.

## AI Disclosure

This was pretty much all Claude Code.  It took many iterations of me copy/pasting failing log messages from GitHub to get it right, though!


.github/workflows/build-dmg.yml
```diff
@@ -0,0 +1,156 @@
...
+          brew install texinfo || true
+          # M2 library dependencies
+          brew install \
+              bdw-gc boost eigen fplll jansson libffi tbb \
+              macaulay2/tap/factory fflas-ffpack flint frobby givaro \
+              mpfi mpsolve ntl \
+              memtailor mathic mathicgb
+          brew link macaulay2/tap/factory --force
+          # External programs bundled into the DMG
+          brew install 4ti2 cohomcalg csdp gfan lrs msolve nauty normaliz topcom
```
mahrud, line 58: Why not this?
```suggestion
          brew install --only-dependencies macaulay2/tap/M2
```
Or alternatively to include build dependencies as well:
```suggestion
          brew install $(brew deps --1 --include-build macaulay2/tap/M2)
```

mahrud: Occasionally the homebrew build might be outdated and you may need to add a few dependencies manually, but at least reduces duplication of these lists.

mahrud:
Instead of an additional workflow, why not add a new line to the `autotools-macos` build, like this section in the `cmake-ubuntu` build for producing a deb package:
```
      - name: Install packages using Ninja
        if: matrix.build-system == 'cmake'
        run: |
          cmake --build . --target install-packages
          if [[ "${{ runner.os }}" == "Linux" ]]; then
              sudo apt-get install -y -q --no-install-recommends dpkg-dev
              echo "GIT_COMMIT=`git describe --dirty --always --match HEAD`" >> $GITHUB_ENV
              cpack -G DEB
          fi
```
Then people could also grab the latest dmg file from any build.

mahrud:
Also, I'll point out that the real test is whether this works on macos 10.15 and below, which homebrew doesn't officially support, not Mike's top of the line Macbook :) See https://docs.brew.sh/Support-Tiers

In fact, pretty soon homebrew will drop all intel builds, which is going to be annoying ...

d-torrance:
I don't expect that this will work on any versions earlier than macOS 14 (arm) or macOS 15 (intel), since that's the versions of the GitHub runners that were used to build them.

d-torrance:
> Then people could also grab the latest dmg file from any build.

Do you think there would be demand for this?  I was envisioning just running this at each release.

.github/workflows/build-dmg.yml
d-torrance: I actually used the `--only-dependencies` version in an earlier iteration of the workflow, but (probably irrationally) removed it for a perceived sense of  control.

mahrud:
You have the development builds for ubuntu, and something comparable is lacking for macos. Also, GitHub deletes artifacts after 24hr so it's not even necessarily a waste of resources, but I certainly wouldn't use it either way so up to you.

> I don't expect that this will work on any versions earlier than macOS 14 (arm) or macOS 15 (intel), since that's the versions of the GitHub runners that were used to build them.

Then who is this for? My impression was that this is for people for whom we can't bottle M2.

If it's a duplicate way to get brew bottles for people that can already use brew then I'm against this PR because people will be confused about which M2 they are running and the problem will _always_ be that they have some old dmg version installed from long ago and their new version (whether from brew or newer dmg) isn't being used. This used to be a pain that was basically eliminated since brew made it a single command to upgrade M2.

mahrud:
> We also add a new GitHub workflow to build dmg files for both Intel and ARM macOS at each release.

Also, distributing dynamic libraries and binaries of _other programs_ on our GitHub release seems super sketchy 👀 what would your debian friends thing? XD (Probably fine if it's just uploaded on Macaulay2.com though)

d-torrance:
I think the target audience is probably Intel macOS users for whom "brew install" currently starts compiling a bunch of things from source.  I'm not a fan of the format either, but there there's been some demand for it.

mahrud:
Do you know what macos version they had? I've only seen people on 10.15 or 11 with this problem recently.

d-torrance:
Discussed with Mike.  Still need to discuss how to distribute .dmg files

mahrud:
ohhh I see what happened, Github changed the name of their actions runners and as a result we literally only built arm64 bottled for M2 in this release (except x86 for linux which I use). I'm surprised nobody made an issue or report this.

See https://github.com/Macaulay2/homebrew-tap/issues/339

PR #4371, opened 2026-05-26
Return rhs in parallel assignment even when lhs is null

d-torrance:
Suggested by @mahrud during my "new features" presentation in Atlanta.

### Before

```m2
i1 : (x,,z) = (1,2,3)

o1 = (1, , 3)

o1 : Sequence
```

### After
```m2
i1 : (x,,z) = (1,2,3)

o1 = (1, 2, 3)

o1 : Sequence
```

## AI Disclosure

Just old school, human-written code.

### DIFF
--- a/M2/Macaulay2/d/evaluate.d
+++ b/M2/Macaulay2/d/evaluate.d
@@ -1282,7 +1282,7 @@ parallelAssignmentFun(x:parallelAssignmentCode):Expr := (
 			then AssignNewOfFromFun(CodeSequence(
 				y.newClause, y.ofClause, y.fromClause, c))
 			else ParallelAssignmentErrorAt(i + 1))
-		    is nullCode do nullE
+		    is nullCode do values.i
 		    else ParallelAssignmentErrorAt(i + 1);
 		    when r
 		    is Error do (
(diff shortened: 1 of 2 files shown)


d-torrance:
I didn't touch parallel augmented assignment.  I'm not sure what the correct behavior is there.  What should happen if we do the following?

```m2
i1 : x = 2

o1 = 2

i2 : (x,) += (3,4)
```

Currently, we return `(5,)`.  But should we return `(5,4)`?  Or maybe even raise an error because we can't add `null + 4`?

*Edit:*  FWIW, Claude recommended `(5,4)`.

mahrud:
Did we recently change something so `ht#k += 3` initializes `ht#k = 3` if it's not already defined? I can't remember if I just suggested in some issue or saw a pull request doing this. Returning `(5,4)` makes sense in that world.

d-torrance:
That was just a suggestion, I think.  Right now it's an error:

```m2
i1 : ht = new MutableHashTable

o1 = MutableHashTable{}

o1 : MutableHashTable

i2 : ht#k += 3
stdio:2:5:(3):[1]: error: key not found in hash table:
        k (of class Symbol)
```

mahrud:
If we allow this, then I think `c += 3` (where `c` isn't pre-defined) and `ht#k += 3` and `(a,) += (3,3)` should all work I think.

d-torrance:
Discussed with Mike -- we should raise an error (e.g., `null + 4`)

d-torrance:
I've updated things so we treat empty entries on the left hand side like `null` in augmented assignment.  So for example:

```m2
i1 : x = 2

o1 = 2

i2 : (x,) += (3,4)
stdio:2:8:(3):[1]: error: no method for binary operator + applied to objects:
            null (of class Nothing)
      +     4 (of class ZZ)

i3 : Nothing + ZZ := (x,y) -> y;

i4 : (x,) += (3,4)

o4 = (8, 4)

o4 : Sequence

i5 : (x,) ??= (5,6)

o5 = (8, 6)

o5 : Sequence
```

d-torrance:
Discussed with Mike

Issue #3533, opened 2024-10-22
Annotations for GitHub Actions


mahrud:
We should consider adding [annotations](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions) to the workflow to make it easier to identify any warnings/errors in compilation or which examples or tests failed.

This would look something like:
![image](https://github.com/user-attachments/assets/c867d2dd-710e-4840-90ce-72bb52e286ff)

mahrud:
Another useful and potentially straightforward Actions feature would be uploading the generated documentation somewhere accessible, if only temporarily, so reviewing documentation changes could be done easily.

PR #4375, opened 2026-05-27
Add problem matchers for annotations during GitHub builds

d-torrance:
The cpptools regex is from https://github.com/microsoft/vscode-cpptools

Closes: #3533 

## AI Disclosure

Some brainstorming with Claude Chat.  It wrote the first draft of some of the regexes, but honestly did a less than stellar job.

### DIFF
--- a/.github/problem-matchers.json
+++ b/.github/problem-matchers.json
@@ -1,0 +1,29 @@
+{
+  "problemMatcher": [
+    {
+      "owner": "cpptools",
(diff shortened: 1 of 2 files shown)


d-torrance:
Oh dear, all the "ambiguous reference" warnings are getting flagged as errors...

mahrud:
Wouldn't be bad to fix those.

PR #4378, opened 2026-05-27
Add lock(Function) method for thread-safe functions

d-torrance:
This is a simpler alternative to #4367 for providing a `lock(Function)` method without adding a new type of statement to the Macaulay2 language.

For example:

```m2
i1 : x = 0

o1 = 0

i2 : f = i -> x += 1;

i3 : parallelApply(1..1000, f);

i4 : x

o4 = 345

i5 : x = 0

o5 = 0

i6 : parallelApply(1..1000, lock f);

i7 : x

o7 = 1000
```

## AI Disclosure

No AI used!

### DIFF
--- a/M2/Macaulay2/d/pthread.d
+++ b/M2/Macaulay2/d/pthread.d
@@ -279,6 +279,32 @@ unlock(e:Expr):Expr := (
     else WrongArgMutex());
 setupfun("unlock0", unlock);
 
+-- env.0 = the mutex to lock/unlock
+-- env.1 = the function to call
+lockedFunction(e:Expr, env:Sequence):Expr := (
+    if length(env) == 2 then (
+	l := lock(env.0);
+	when l is Error do return l else nothing;
+	r := applyEE(env.1, e);
+	u := unlock(env.0);
+	when u is Error do u else r)
+    else buildErrorPacket("internal error: expected 2 values in environment"));
+
+lockFunction(e:Expr):Expr := (
+    when e
+    is a:Sequence do (
+	if length(a) == 2 then (
+	    when a.0
+	    is mutexCell do (
+		if isFunction(a.1)
+		then Expr(CompiledFunctionClosure(
+			lockedFunction, nextHash(), a))
+		else WrongArg(2, "a function"))
+	    else WrongArg(1, "a mutex"))
+	else WrongNumArgs(2))
+    else WrongNumArgs(2));
+setupfun("lockFunction", lockFunction);
+
 -- Local Variables:
 -- compile-command: "echo \"make: Entering directory \\`$M2BUILDDIR/Macaulay2/d'\" && make -C $M2BUILDDIR/Macaulay2/d pthread.o "
 -- End:
--- a/M2/Macaulay2/m2/engine.m2
+++ b/M2/Macaulay2/m2/engine.m2
@@ -187,7 +187,7 @@ processWeights = (nvars,weights) -> (
 	       then error("Weights: expected weight vector of length ",toString nvars," but got ",toString (#wt))));
      weights);
 
-makeMonomialOrdering = (monsize,inverses,nvars,degs,weights,ordering) -> (
+makeMonomialOrdering = lock((monsize,inverses,nvars,degs,weights,ordering) -> (
      -- 'monsize' is the old MonomialSize option, usually 8 or 16, or 'null' if unset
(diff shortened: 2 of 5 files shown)


mahrud:
I have something very similar locally, but I was trying to figure out how to make a syntax like one of the following work before making a PR:
```m2
lock blah ZZ := x -> (...)
blah ZZ := lock x -> (...)
blah ZZ := x -> lock (...)
```
I think `lock` would have to become a keyword for any of these, but the mutex logic for all of them is identical to yours.

More broadly, I wish M2 had macros, because I also would like one of these:
```m2
async blah ZZ := x -> (...)
blah ZZ := async x -> (...)
blah ZZ := x -> async (...)
```

in general, the fewer parentheses needed, the better :)

mahrud:
I should add that I already like this a lot, but here's a small caveat:
```m2
f = i -> x += 1;
g = n -> parallelApply(1..n, lock f)
parallelApply(1..10, g)
```
Would probably call `lock f` 10 _separate_ times, therefore producing 10 separate locked functions which may collide with each other.

jkyang92:
> I have something very similar locally, but I was trying to figure out how to make a syntax like one of the following work before making a PR:
> 
> ```m2
> lock blah ZZ := x -> (...)
> blah ZZ := lock x -> (...)
> blah ZZ := x -> lock (...)
> ```
> 
> I think `lock` would have to become a keyword for any of these, but the mutex logic for all of them is identical to yours.
> 

The first one would require either some globalAssignHook magic or changes in d, possibly to the parser, as I'm not certain what sortof code this assignment is supposed to produce. Although in typing this out, you might be able to create something like a "locked method" type which associates locks appropriately to calls of it. (Not necessarily a good idea but it's an idea)

For the middle one, it can be implemented at top level, whether you want to is a different question.

```m2
lock = f -> (
    m = new Mutex;
    x -> (lock m; r := f x; unlock m; r)
)
```
(You'd need some try/catch stuff  to be exception safe too)

The last one would have to be implemented in d but wouldn't require syntactic changes to the language

> I should add that I already like this a lot, but here's a small caveat:
> 
> ```m2
> f = i -> x += 1;
> g = n -> parallelApply(1..n, lock f)
> parallelApply(1..10, g)
> ```
> 
> Would probably call `lock f` 10 _separate_ times, therefore producing 10 separate locked functions which may collide with each other.

This is definitely what the reasonable semantics would do. Unfortunately it will defy many users expectations. You could implement it as a keyword using d, then since the code object doesn't change it would be possible to associate the lock to the code object `f`. This is weird to me though since it we can't actually tell if `f` is a function until  the code is actually run.

Also, it might be useful to have a version that takes a mutex and a function and associates that mutex to that function. Or a version that can say that some set of functions should use the same mutex.

d-torrance:
> Also, it might be useful to have a version that takes a mutex and a function and associates that mutex to that function. 

Ooh, I like that!  That should be easy to implement.  I'm guessing we'd want `lock(Mutex, Function)` so folks could do `lock_mutex f`

mahrud:
> You could implement it as a keyword using d, then since the code object doesn't change it would be possible to associate the lock to the code object `f`. This is weird to me though since it we can't actually tell if `f` is a function until the code is actually run.

This is what I was trying to figure out. I hope it is doable, but I'm actually not sure. I think this feature wouldn't be very useful with this caveat.

d-torrance:
I went ahead and added documentation, tests, and the fix for #3239 to this PR.

I'm personally a big fan of the `with` idea, but that's a bigger discussion.  If we do eventually go that way and if this PR is merged, then it would be easy to rewrite the code in this PR to use it.

On the other hand, adding a 2-argument method for `lock` as I'm proposing here would make a possible future switch to `lock` as a keyword a breaking change, as `lock_mutex f` wouldn't work anymore.   Currently, it wouldn't really matter since the only method that's defined at the moment is `lock(Mutex)`, and who cares if `lock` is a method function or keyword for that to work.

*Edit:*  I guess it would be a breaking change even with the existing code if someone were to do so something like `scan(listOfMutexes, lock)`.

mahrud:
I think given the issue in https://github.com/Macaulay2/M2/pull/4378#issuecomment-4558379554 it's fair to say this feature is experimental and will likely change in the future.

d-torrance:
Discussed with Mike

mahrud:
I would hold off on announcing it in the changelog until its less experimental though. e.g. I don't think at this stage packages should use this.

PR #4380, opened 2026-05-27
fix for PR #4357, TeX printing for all classes

galettof:
Small changes as per title.
I did not increase the package version, and I do not mind if this is not announced (but I can updated the changes file if desired).


d-torrance:
I usually only add a changelog entry for a package if there's been a version bump

d-torrance:
I played around with the new `texMath` methods in Jupyter and they look awesome!

image
image
image
image

d-torrance (approved):
Looks good -- will merge assuming all the builds finish successfully

galettof:
Thank you for trying Jupyter! I didn't think about it and only tested in Macaulay2Web and VSCode.

PR #4383, opened 2026-05-28
add ARingRRx2: double-double (~106-bit) approximate real field

andrew-tawfeek:
This is a work-in-progress of bringing over the techniques of [MultiFloats.jl](https://github.com/dzhang314/MultiFloats.jl) to Macaulay2 in an effort to increase the speed of floating-point arithmetic. This fits in a class between ARingRR and ARingRRR, which is called simply "ARingRRx2," for arithmetic with more than 53-bit but less than arbitrary precision.

It seems with this implementation (the conversion of the algorithms from the above-cited Juilia implementation to C++ done with Claude assistance) that NAG, dense linear algebra computations, as well as homotopy continuation are all strongly sped up -- see comment below the PR.

AI-assisted technical breakdown below.

-----

Adds a new approximate-real field to Macaulay2's engine — **`ARingRRx2`** — implementing **double-double (`Float64x2`, ~106-bit) arithmetic** in pure C++, filling the gap between `ARingRR` (53-bit hardware double) and `ARingRRR` (MPFR, arbitrary precision) at the fixed precision band where MPFR's per-op overhead dominates.

The technique (error-free transformations via TwoSum / TwoProd-FMA plus Hida-Li-Bailey QD-family algorithms) gives ~106-bit precision at a fraction of MPFR's per-op cost.

## Empirical speedup vs `ARingRRR(106)` at matched ~106-bit accuracy

**Dense LU-solve at varying N** (single-thread, both fields, residuals ~1e-31 both sides):

| N | ARingRRx2 | ARingRRR(106) | speedup |
|---:|---:|---:|---:|
| 64 | 1.1 ms | 9.5 ms | 8.3× |
| 128 | 8.8 ms | 84.0 ms | 9.6× |
| 256 | 64.4 ms | 588.1 ms | 9.1× |
| 512 | 503.8 ms | 4488.9 ms | 8.9× |

**Newton-step kernel** — the homotopy-tracker inner loop (K=5 refactor-and-solve on a perturbed dense Jacobian):

| N | ARingRRx2 | ARingRRR(106) | speedup |
|---:|---:|---:|---:|
| 64 | 5.2 ms | 46.8 ms | 9.0× |
| 128 | 41.3 ms | 359.6 ms | 8.7× |
| 256 | 301.2 ms | 2778.8 ms | 9.2× |

An N=256 homotopy step at ~30-digit precision drops from ~2.78 s on MPFR to ~301 ms on the new field — same residuals.

**Scalar throughput**, 20 M ops, single-thread:

| op | ARingRRx2 | ARingRRR(106) | speedup |
|---:|---:|---:|---:|
| add | 18.8 ns | 133 ns | 7.1× |
| mul | 9.2 ns | 147 ns | 16× |

## What's in this PR

`ElementType` is a 16-byte POD `{hi, lo}` stored directly by `dmat<ARingRRx2>` — no `ring_elem` boxing in the dense-LA hot path. `ring_elem` boxing is lossless via a 106-bit MPFR (mirrors `ARingRRR`'s pattern).

- `Macaulay2/e/aring.hpp` — adds `ring_RRx2` to the `RingID` enum (before `ring_old`; does not shift any existing enum value)
- `Macaulay2/e/aring-RRx2.hpp` — the field itself (`class ARingRRx2 : public SimpleARing<ARingRRx2>`)
- `Macaulay2/e/unit-tests/ARingRRx2Test.cpp` — 7 gtests modelled on `ARingRRTest.cpp`
- `Macaulay2/e/CMakeLists.txt` — register the new test

After this lands, `./M2-unit-tests --gtest_filter='ARingRRx2*'` passes **7/7** (create, negate, add, subtract, multDivide, axioms, power_and_invert).

## Build prerequisite

When `libflint-dev` < 3.0 is installed on the build host (e.g. Debian 12), a cmake misconfiguration causes factory to be compiled against the wrong FLINT — tracked + fixed in #4381 (orthogonal to this PR).

### DIFF
--- a/M2/Macaulay2/e/CMakeLists.txt
+++ b/M2/Macaulay2/e/CMakeLists.txt
@@ -459,6 +459,7 @@ if(BUILD_TESTING)
     unit-tests/ARingRRTest.cpp
     unit-tests/ARingCCTest.cpp
     unit-tests/ARingRRRTest.cpp
+    unit-tests/ARingRRx2Test.cpp
     unit-tests/ARingRRiTest.cpp
     unit-tests/ARingCCCTest.cpp
     unit-tests/NCGroebnerTest.cpp
--- a/M2/Macaulay2/e/CMakeLists.txt
+++ b/M2/Macaulay2/e/CMakeLists.txt
@@ -459,7 +459,7 @@ if(BUILD_TESTING)
     unit-tests/ARingRRTest.cpp
     unit-tests/ARingCCTest.cpp
     unit-tests/ARingRRRTest.cpp
-    unit-tests/ARingRRx2Test.cpp
+    unit-tests/ARingRRddTest.cpp
     unit-tests/ARingRRiTest.cpp
     unit-tests/ARingCCCTest.cpp
     unit-tests/NCGroebnerTest.cpp
--- a/M2/Macaulay2/e/Makefile.files.in
+++ b/M2/Macaulay2/e/Makefile.files.in
@@ -222,6 +222,7 @@ NAMES_H = \
 	hash \
 	f4/ntuple-monomial \
 	f4/varpower-monomial \
+	aring-RRx2 \
 	res-a0-pair
 
 C_FILES = error table exptable complex # interface/m2-types interface/m2-mem
--- a/M2/Macaulay2/e/Makefile.files.in
+++ b/M2/Macaulay2/e/Makefile.files.in
@@ -222,7 +222,7 @@ NAMES_H = \
 	hash \
 	f4/ntuple-monomial \
 	f4/varpower-monomial \
-	aring-RRx2 \
+	aring-RRdd \
 	res-a0-pair
 
 C_FILES = error table exptable complex # interface/m2-types interface/m2-mem
--- a/M2/Macaulay2/e/aring-RRdd.hpp
+++ b/M2/Macaulay2/e/aring-RRdd.hpp
@@ -4,16 +4,16 @@
 // ~/research/dd-proto/). Modeled on aring-RR.hpp (value-type SimpleARing) for the
 // interface, and aring-RRR.hpp for the ring_elem boxing.
 //
-// ElementType is a 16-byte POD (hi+lo); dmat<ARingRRx2> stores it directly (the fast
+// ElementType is a 16-byte POD (hi+lo); dmat<ARingRRdd> stores it directly (the fast
 // numerical-LA path where the speedup lands). It is too big for a ring_elem inline
 // slot, so to/from_ring_elem box it losslessly through a 106-bit mpfr (gmp_RR) — the
 // ring_elem path is not performance-critical.
 //
-// TODO(build-verify): add `ring_RRx2` to the RingID enum (aring.hpp) and a front-end
+// TODO(build-verify): add `ring_RRdd` to the RingID enum (aring.hpp) and a front-end
 // ring constructor; until then ringID uses a placeholder. Confirm moveTo_gmpRR/get_mpfr
 // signatures against ringelem.hpp during the first engine compile.
-#ifndef _aring_RRx2_hpp_
-#define _aring_RRx2_hpp_
+#ifndef _aring_RRdd_hpp_
+#define _aring_RRdd_hpp_
 
 #include <cmath>
 #include "interface/gmp-util.h"  // moveTo_gmpRR
@@ -115,9 +115,9 @@ static inline DoubleDouble dd_div_d(DoubleDouble a, double b)
 // applied lane-wise.  Inputs/outputs are structure-of-arrays — one __m256d
 // packs 4 hi-parts, another packs 4 lo-parts.
 //
-// Header-guarded so the rest of aring-RRx2.hpp keeps compiling on non-AVX2
+// Header-guarded so the rest of aring-RRdd.hpp keeps compiling on non-AVX2
 // targets (older Intel without AVX2, non-x86 archs).  Downstream kernels
-// (e.g., a future dmat<ARingRRx2> SIMD specialisation) can include this
+// (e.g., a future dmat<ARingRRdd> SIMD specialisation) can include this
 // section conditionally to opt in.
 //
 // Portability (per review): even when the compiler targets AVX2+FMA, the build
@@ -182,7 +182,7 @@ static inline void v4_dd_broadcast(DoubleDouble a, __m256d* hi, __m256d* lo) {
 // Computes y[i] -= f * x[i] for i in [0, N), the inner loop of LU pivot
 // updates, dense mat-vec, and the perturbed solves in homotopy continuation.
 // Vectors are in structure-of-arrays layout (separate hi[] / lo[] arrays);
-// SoA is the natural SIMD layout — dmat<ARingRRx2> kernels wanting maximum
+// SoA is the natural SIMD layout — dmat<ARingRRdd> kernels wanting maximum
 // throughput should adopt it.  AoS callers can pre-split or use the scalar
 // dd_* primitives instead.  Tail elements (when N % 4 != 0) handled by the
 // scalar v3.0 ops.
@@ -223,17 +223,17 @@ static inline DoubleDouble dd_from_mpfr(mpfr_srcptr x)
 /**
 \ingroup rings
 */
-class ARingRRx2 : public SimpleARing<ARingRRx2>
+class ARingRRdd : public SimpleARing<ARingRRdd>
 {
   // approximate real numbers, double-double (~106-bit), MultiFloats Float64x2.
  public:
-  static const RingID ringID = ring_RRx2;
+  static const RingID ringID = ring_RRdd;
   static const unsigned long PRECISION = 106;
 
   typedef DoubleDouble elem;
   typedef elem ElementType;
 
-  ARingRRx2() {}
+  ARingRRdd() {}
   size_t characteristic() const { return 0; }
   unsigned long get_precision() const { return PRECISION; }
   void text_out(buffer &o) const;
--- a/M2/Macaulay2/e/aring-RRx2.hpp
+++ b/M2/Macaulay2/e/aring-RRx2.hpp
@@ -119,7 +119,17 @@ static inline DoubleDouble dd_div_d(DoubleDouble a, double b)
 // targets (older Intel without AVX2, non-x86 archs).  Downstream kernels
 // (e.g., a future dmat<ARingRRx2> SIMD specialisation) can include this
 // section conditionally to opt in.
-#if defined(__AVX2__) && defined(__FMA__)
+//
+// Portability (per review): even when the compiler targets AVX2+FMA, the build
+// can force the scalar path by defining M2_NO_DD_SIMD (e.g. a configure-time
+// --disable-dd-simd). This matters because Macaulay2 is frequently compiled on
+// one machine and run on another: a binary built with -march=native on an AVX2
+// host would otherwise emit AVX2 instructions that fault (SIGILL) on a baseline
+// CPU. Distributions that build for a baseline ISA already won't define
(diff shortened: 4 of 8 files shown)


andrew-tawfeek:
Some research-level context — what this enables in M2 day-to-day.

M2's existing precision ladder leaves a gap:

```
  ARingRR   — 53-bit hardware double, ~16 decimal digits, fastest
  ( gap )
  ARingRRR  — MPFR, arbitrary precision the user picks, heavy per-op cost
```

Anyone needing ~30 decimal digits (a very common ask in numerical AG) currently has to jump straight to MPFR-backed `ARingRRR` and pay its per-op overhead — even though fixed ~106-bit double-double is plenty for the workload. `ARingRRx2` is exactly that missing tier.

All numbers below: vs `ARingRRR(106)` at matched ~106-bit accuracy (residuals ~1e-31 both sides), single-thread CPU. M2-syntax sketches use `RR_106x2` as a placeholder for whatever language binding name is eventually chosen.

## Homotopy continuation

A homotopy tracker spends most of its wall time in the Newton-refine inner loop: repeated dense linear solves on a slowly-perturbed Jacobian. Timing that kernel (K=5 perturbed refactor-and-solve):

| N | ARingRRx2 | ARingRRR(106) | speedup |
|---:|---:|---:|---:|
| 64 | 4.5 ms | 48.7 ms | 10.8× |
| 128 | 39.1 ms | 330.8 ms | 8.5× |
| 256 | 311.2 ms | 3036.5 ms | 9.8× |

A single step at N=256 unknowns and ~30-digit precision: **~3.04 s on MPFR → ~311 ms here**. A 100-step path → minutes vs ~30 seconds; multiply by however many paths you're tracking.

```m2
needsPackage "NumericalAlgebraicGeometry";
-- BEFORE: R = RR_106;    -- MPFR-backed, ~3.04 s per linear solve at N=256
-- AFTER:  R = RR_106x2;  -- the new field, ~311 ms (9.8× faster, same residual)
M = mutableMatrix(R, 256, 256);
b = mutableMatrix(R, 256, 1);
solve(M, b)
```

## Iterative refinement / certified-solution polishing

Standard numerical-AG pattern: take an approximate solution `x ≈ A⁻¹b`, then iterate `r = b − Ax`; `Az = r`; `x += z` at high precision until the residual hits the precision floor. Hot loop is mat-vec + back-sub. K=10 refinement steps:

| N | ARingRRx2 | ARingRRR(106) | speedup |
|---:|---:|---:|---:|
| 64 | 14.1 ms | 120.8 ms | 8.5× |
| 128 | 86.0 ms | 768.5 ms | 8.9× |
| 256 | 590.0 ms | 5957.2 ms | 10.1× |

For users polishing approximate solutions to certified high-precision accuracy this is the dominant cost: **~5.96 s → ~590 ms at N=256**.

```m2
R = RR_106x2;
solveAndRefine = (A, b) -> (
    x := solve(A, b);             -- initial LU + back-sub
    for k from 1 to 10 do (
        r := b - A * x;            -- high-precision residual
        x = x + solve(A, r);       -- correction step
    );
    x
)
```

## Power iteration for dominant eigenvalue

K=50 iterations of `y = A*x; λ = ||y||∞; x = y/λ`. Pure mat-vec + scalar normalize — a very different kernel mix from LU-heavy workloads, so a useful counter-example:

| N | ARingRRx2 | ARingRRR(106) | speedup |
|---:|---:|---:|---:|
| 64 | 3.8 ms | 18.5 ms | 4.9× |
| 128 | 14.2 ms | 71.7 ms | 5.0× |
| 256 | 59.6 ms | 327.6 ms | 5.5× |

The speedup is smaller here (~5× rather than ~9×) because the kernel is pure mul+add, exactly where MPFR's per-op overhead is least painful relative to its useful work. Still substantial — any spectral analysis at 30-digit precision (conditioning estimates, eigen-localization in endgame algorithms) sees this.

```m2
R = RR_106x2;
x = mutableMatrix(R, 256, 1, ...);      -- random starting vector
for k from 1 to 50 do (
    y := A * x;                          -- mat-vec
    lam := norm(y, infinity);            -- ||y||_∞
    x = (1/lam) * y;                     -- normalize
)
```

andrew-tawfeek:
## Second optimisation pass (4 commits on `add-ARingRRx2`)

Following up with the four commits pushed overnight, each as a separate commit for clean history. Headline numbers below are at matched ~106-bit accuracy versus the current MPFR / `RRR(106)` baseline.

### Commits

| | Hash | Description |
|---|---|---|
| C1 | `a0d13bbe` | MultiFloats.jl v3.0 normalisation-free dd algorithms (Zhang 2025) |
| C2 | `86acf153` | dd × double mixed helpers (`add_d` / `sub_d` / `mul_d` / `div_d`) |
| C3 | `d72df0bf` | AVX2 4-wide SIMD primitives (`v4_dd_add` / `v4_dd_mul`, etc.) |
| C4 | `b0a72558` | High-level SIMD `dd_axpy` kernel (LU / mat-vec inner loop) |

### Benchmarks vs MPFR @ 106 bits

**Dense LU-solve** — the heart of any numerical-AG linear-algebra call:

|   N | RRx2     | MPFR      | speedup    |
|----:|---------:|----------:|:-----------|
|  64 |   0.3 ms |    6.8 ms | **26.6×**  |
| 128 |   1.8 ms |   66.5 ms | **37.0×**  |
| 256 |  16.4 ms |   506  ms | **30.8×**  |
| 512 | 125   ms |  4168  ms | **33.2×**  |

**Homotopy-path inner loop** (K = 5 perturbed refactor-and-solve — the NAG / `HomotopyContinuation.jl` hot loop):

|   N | RRx2     | MPFR      | speedup   |
|----:|---------:|----------:|:----------|
|  64 |   1.7 ms |   41.8 ms | **25.2×** |
| 128 |  12.1 ms |  328.9 ms | **27.2×** |
| 256 | 111   ms | 2566   ms | **23.0×** |

**Iterative refinement** (1 LU + K = 10 residual-and-back-sub):

|   N | RRx2     | MPFR      | speedup   |
|----:|---------:|----------:|:----------|
|  64 |   4.4 ms |   83.1 ms | **18.7×** |
| 128 |  30   ms |  663   ms | **22.1×** |
| 256 | 207   ms | 5259   ms | **25.4×** |

**Power iteration** (K = 50 mat-vec + normalise) — dominated by mat-vec rather than divisions, so the MPFR overhead is smaller and the win is correspondingly modest:

|   N | RRx2    | MPFR     | speedup  |
|----:|--------:|---------:|:---------|
|  64 |  3.1 ms |  25.0 ms | **8.1×** |
| 128 | 12.6 ms |  69.2 ms | **5.5×** |
| 256 | 51.3 ms | 290.8 ms | **5.7×** |

### For a homotopy-continuation users:

A path with 200 Newton refactor steps at *N* = 256 that took ~9.2 min on `RRR(106)` now takes **~22 s** on RRx2 — a 100k-path tracking job that was a long lunch break becomes a quick coffee.

### Honest note on the SIMD layer (C3 / C4)

The AVX2 SIMD primitives are correct and shave a few percent off LU at *N* ≥ 128 (you can see SIMD-LU is consistently faster than scalar-dd LU above — e.g. 125 ms vs 178 ms at *N* = 512). On an isolated AXPY microbench, hand-vectorisation only ties GCC's `-O3 -march=native` auto-vectorisation on this Skylake-class box. The primitives still matter for engine integration (where `dmat<ARingRRx2>` passes `dd` as a struct and the compiler cannot always vectorise across that boundary), and they are the cleanest path for an AVX-512 specialisation later. Reasonable to keep them in this PR; happy to split C3 + C4 into a follow-up if reviewers would rather land C1 + C2 first.

### Correctness

200 000 random trials versus MPFR @ 200-bit:

| op    | min accurate bits |
|:------|:------------------|
| add   | 105.7             |
| mul   | 105.5             |
| div   | 104.0             |
| sqrt  | 104.8             |

Drops 0.3 – 2 bits below the theoretical 106 in worst-case operands — well inside RRx2's documented spec.

### Build / repro

Bench scaffold at `~/research/dd-proto/numag_kernels.cpp`:

```bash
g++ -O3 -march=native -mavx2 -mfma numag_kernels.cpp -o numag_kernels -lmpfr -lgmp
./numag_kernels
```

d-torrance:
> `R = RR_106x2`

I don't understand -- this syntax isn't supported.  You would need to make some changes to the interpeter to not get an error.  And if the syntax isn't supported, then where are the timing results coming from?

Although it probably make sense for `RR_106` to give us double-double out of the box like `RR_53` gives us double now.  (Plus `106x2` kind of implies we have 212-bit precision!)

A couple other thoughts:

* The autotools build should be supported.
* There should be an option to either turn off the SIMD support at configure time or make it detectable at runtime.  Very frequently, Macaulay2 is compiled on one machine and then run on another, so for example Debian builds all binaries for a baseline instruction set for each architecture.

andrew-tawfeek:
> > `R = RR_106x2`
> 
> I don't understand -- this syntax isn't supported. You would need to make some changes to the interpeter to not get an error. And if the syntax isn't supported, then where are the timing results coming from?

The frontend isn't connected here (though I began some work on that on a separate un-PR'd branch) -- so these timing results are coming from the unit tests! Namely the committed unit test `Macaulay2/e/unit-tests/ARingRRx2Test.cpp` is correctness-only — it checks each operation (add/sub/mul/div/sqrt/axioms/power) against MPFR to ~104–106 accurate bits over random trials, and does not measure timing. To distinguish it from my (human) comments, I've italicized Claude's commenting on some more technical aspects below.

_The speedup tables came from standalone field-level microbenchmarks compiled with `g++ -O2 -march=native … -lmpfr -lgmp`, single-thread, each comparing the exact double-double primitives in a`ring-RRx2.hpp` against `mpfr_t` at precision 106 (matched ~106-bit residual, ~1e-31 both sides): dense LU-solve with partial pivoting (N=64–512), a Newton-step homotopy-path kernel (K=5 perturbed refactor-and-solve), iterative refinement (1 LU + K=10 residual/back-sub), and power iteration (K=50 mat-vec + normalize). These exercise the field/engine layer directly — the same arithmetic `a dmat<ARingRRx2>` uses — which is why they're meaningful even though this branch has no interpreter binding yet (the `RR_106x2` in my snippets was a placeholder name). These were just now committed these as an in-tree benchmark so they're reproducible under the build._

Also, the autotools + SIMD-guard changes got pushed.


> Although it probably make sense for `RR_106` to give us double-double out of the box like `RR_53` gives us double now. (Plus `106x2` kind of implies we have 212-bit precision!)

Connecting RR_106 directly to `ARingRRdd` sounds like a good idea (also, currently working on renaming all engine instances of `ARingRRx2` to `ARingRRdd`). From what I understand that's frontend work and I've prototyped some of this on another branch (public, called "add-ARingRRx2-frontend") -- this requires changes in `d/interface.dd` + `m2/reals.m2` + `e/interface/ring.{cpp,h}` + `aring-translate`.

Also, one warning from Claude about this re-routing:
_One caveat worth documenting: double-double is two IEEE doubles, so it carries double's exponent range (~1e±308) and non-uniform ULP — not MPFR-106's uniform mantissa with arbitrary exponent. It matches MPFR-106 to ~104–106 bits across the normal range (verified against MPFR), but workloads relying on very large/small exponents or exact 106-bit ULP semantics would see a difference. So if RR_106 is repurposed to dd, that tradeoff should be documented, and we may want true MPFR-106 still reachable for the arbitrary-exponent case — happy to follow whatever convention you prefer._

Just pushed all the changes. Currently testing compilation (`./configure && make + ctest -R`) on a Google Cloud server, will edit this post with a confirmation that all went well/if there were any issues. (edit: compiled cleanly)

d-torrance:
If `RR_106` were to use this, then that's all engine.  The `d` directory doesn't know about anything but MPFR's.  Right now, if you type `1.0` in the REPL, it constructs a 53-bit precision MPFR object.  We don't actually get `double`'s until we start playing around with matrices and polynomials.  I think the dispatch happens in `interface/ring.cpp`.

Issue #3653, opened 2025-02-11
make -C check-configure check fails when using --enable-build-libraries


d-torrance:
From https://github.com/Macaulay2/M2/actions/runs/13263574577/job/37025478589?pr=3594:

```
configure: error:  *** - automatic building of libraries disabled, but some must be built: givaro 
```

We ran `configure --enable-build-libraries="fflas_ffpack givaro"`, so we shouldn't have gotten this error.

d-torrance:
Closed by #4385

PR #4384, opened 2026-05-28
Bump givaro to v4.2.2

d-torrance:
A new version of givaro was just released.

Draft for now to test the builds


d-torrance:
Builds mostly worked.  We ran into #3653, but I can't reproduce it locally and don't understand what's wrong.  Actually building M2 against the new givaro worked fine, which is what matters.

https://github.com/Macaulay2/M2/actions/runs/26581869542?pr=4384

Issue #556, opened 2016-12-29
disabling a library breaks a test


DanGrayson:
If we use one of the options `--disable-frobby  --disable-gfan`, then it doesn't get passed to the check in `check-configure` and the check fails, because it expects to find the library present.

d-torrance:
Closed by #4385

PR #4385, opened 2026-05-28
Remove check-configure

d-torrance:
I've never been quite sure what it's useful for.  Checking the `configure` script, I suppose, but it seems like successfully running the `configure` script is sufficient for that.  It's just something else to maintain alongside `configure.ac` (e.g., if you add/change a feature there, then you probably have to change something here).

Closes: #556, #3653 by making them non-issues

Partially addresses #1091


d-torrance:
Discussed with Mike

PR #4386, opened 2026-05-29
Fix F4 GB algorithm on 32-bit ARM; add MonomialInfo tests

d-torrance:
### Summary

Replace `long` with `int64_t` for `monomial_word` in the F4 Gröbner basis engine, fixing test failures on 32-bit ARM (armhf). Add C++ unit tests for `MonomialInfo` and extend the `EngineTests` suite with additional F4 GB tests over extension fields and wider variable/packing configurations.

### Background

The F4 Gröbner basis algorithm stores packed monomials as arrays of `monomial_word`, which was `typedef long monomial_word`. On 64-bit platforms `long` is 64 bits, but on 32-bit ARM (armhf) it is only 32 bits. The original code even carried a comment acknowledging the uncertainty: `// Is this OK?`

The test `EngineTests/GB.Test.LinearAlgebra.m2` (test 75), which computes a Gröbner basis over `GF(125)` in a ring with `MonomialSize=>8`, was failing on Ubuntu 18.04 armhf with:

```
-- Internal error: array out of order   (6 times)
-- i7 :   assert(gbC == gbB)
-- error: assertion failed
```

The "array out of order" message comes from a consistency check in `F4GB::reorder_columns` (`f4.cpp`): after sorting the F4 matrix columns by monomial order, it verifies that each row's component array is strictly increasing. When this fires, the subsequent Gaussian elimination uses the wrong pivot columns and produces an incorrect Gröbner basis.

The test was added in commit fa31ec2ba7 after the F4 algorithm was extended (in 37fda3a6b6) to support non-default monomial orders by storing a total-degree weight slot inside each packed monomial. Both commits were written and validated on 64-bit only.

### Changes

**`M2/Macaulay2/e/f4/monomial-word.hpp`** (new file) — extracts the `monomial_word` typedef into a standalone header so that both `moninfo.hpp` and `varpower-monomial.hpp` can include it without a circular dependency.

**`M2/Macaulay2/e/f4/moninfo.hpp` / `moninfo.cpp`** — `monomial_word` changed from `long` to `int64_t` via the new header; all function parameters and return types that hold monomial slot values updated accordingly; comparison temporaries changed from `varpower_word` to `monomial_word` to avoid narrowing; `compare_schreyer` tiebreaker parameters use `int` since they are component indices; `hash_value` return type changed to `uint64_t`, consistent with the Macaulay2 interpreter's convention for hash values; `printf` format specifiers updated to `PRId64`.

**`M2/Macaulay2/e/f4/varpower-monomial.hpp`** — `ExponentList<long, false>` → `ExponentList<monomial_word, false>`, so `varpower_word` stays in sync with `monomial_word` automatically.

**`M2/Macaulay2/e/f4/f4-monlookup.hpp` / `.cpp`**, **`f4-m2-interface.cpp`**, **`monhashtable.hpp`** / **`.cpp`**, **`schreyer-resolution/res-poly-ring.cpp`** — callers updated to use `monomial_word` (or `auto`) where they previously used `long` to store values originating from `monomial_word` slots.

**Behavior on 64-bit is completely unchanged.** On all mainstream 64-bit Unix platforms (Linux, macOS) `long` and `int64_t` are the same type, so the generated code is identical.

### Tests

**`M2/Macaulay2/e/unit-tests/MonomialInfoTest.cpp`** (new file, registered in `Makefile.files`) — direct C++ unit tests for `MonomialInfo`, exercising the properties that would be silently broken by a too-narrow `monomial_word`: round-trip correctness of `from_expvector`/`to_expvector`; exponent arithmetic under `unchecked_mult` and `unchecked_divide`; the additive hash property `hash(m·n) == hash(m) + hash(n)` (the invariant that breaks on 32-bit if the hash accumulator truncates); and comparison correctness for GRevLex, GRevLex4 (`MonomialSize=>8`), GRevLex2 (`MonomialSize=>16`), and Lex orderings with 4, 5, 7, and 10 variables.

**`M2/Macaulay2/packages/EngineTests/GB.Test.LinearAlgebra.m2`** — six new `TEST` blocks verifying that `Algorithm => LinearAlgebra` produces the same GB as the default algorithm, covering: GF(4), GF(9), GF(27) with MonomialSize=>8; GF(125) with MonomialSize=>16 (the GRevLex2 packing path); ZZ/101 with 10 variables and MonomialSize=>8; and GF(125) with 10 variables and MonomialSize=>8.

### Test plan

Run the new C++ unit tests with `./M2-unit-tests --gtest_filter="MonomialInfo*"`. Confirm the original armhf failure (test 75 over GF(125) with MonomialSize=>8) is resolved on a 32-bit ARM build. Verify no regressions on 64-bit.

## AI Disclosure

This was all Claude, baby!

*Draft for now pending the "great sort"*

### DIFF
--- a/M2/Macaulay2/e/buffer.cpp
+++ b/M2/Macaulay2/e/buffer.cpp
@@ -57,6 +57,14 @@ void buffer::put(long n)
   put(s, strlen(s));
 }
 
+void buffer::put(long long n)
+{
+  const int N = 100;
+  char s[N];
+  snprintf(s, N, "%lld", n);
+  put(s, strlen(s));
+}
+
 void buffer::put(double n)
 {
   const int N = 100;
--- a/M2/Macaulay2/e/buffer.hpp
+++ b/M2/Macaulay2/e/buffer.hpp
@@ -59,6 +59,7 @@ class buffer : public our_new_delete
   void put(int n);              // Format the integer, place into buffer
   void put(int n, int width);   // Format the integer, with given width field.
   void put(long n);             // Format the integer, place into buffer
+  void put(long long n);        // Format the integer, place into buffer
   void put(double n);             // Format the double, place into buffer
   void put(long n, int width);  // Format the integer, with given width field.
   void put(unsigned int n);     // Format the integer, place into buffer
@@ -100,6 +101,11 @@ class buffer : public our_new_delete
     return *this;
   }
   buffer &operator<<(long n)
+  {
+    put(n);
+    return *this;
+  }
+  buffer &operator<<(long long n)
   {
     put(n);
     return *this;
--- a/M2/Macaulay2/e/f4/f4-m2-interface.cpp
+++ b/M2/Macaulay2/e/f4/f4-m2-interface.cpp
@@ -93,7 +93,7 @@ vec F4toM2Interface::to_M2_vec(const VectorArithmetic* VA,
   const monomial_word *w = f.monoms;
   for (int i = 0; i < f.len; i++)
     {
-      long comp;
+      monomial_word comp;
       MI->to_expvector(w, lexp, comp);
       w = w + MI->monomial_size(w);
       for (int a = 0; a < M->n_vars(); a++) exp[a] = static_cast<int>(lexp[a]);
--- a/M2/Macaulay2/e/f4/f4-monlookup.cpp
+++ b/M2/Macaulay2/e/f4/f4-monlookup.cpp
@@ -261,7 +261,7 @@ void F4MonomialLookupTableT<Key>::reset_expvector(
 }
 
 template <typename Key>
-bool F4MonomialLookupTableT<Key>::find_one_divisor_vp(long comp,
+bool F4MonomialLookupTableT<Key>::find_one_divisor_vp(monomial_word comp,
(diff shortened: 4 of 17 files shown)


jkyang92:
I'll do a more thorough review later, but I'm curious what the speed implications for 32bit systems is. Historically, 64bit ops on 32bit were slow, but that might not be true for recent 32bit systems. Of course supporting two different widths is not exactly an easy thing.

Second, it's not so obvious to me where the actual bug is. In particular, simply using `long` would be fine as long as no one ever expects it to be anything other than a `long`. So if you know where that is, it would help clarify whether the bug has actually been fixed or simply moved to a different location.

d-torrance:
I'm not too concerned about speed - my main goal is just for the tests to pass on i386 and armhf when building the Debian package for the sake of M2 transitioning from unstable -> testing nicely at each release.

I've started noticing EngineTests 75 failing more and more on GitHub builds, so I'm starting to think that this diagnosis from Claude may just be slop and it's not even a 32-bit thing (but maybe just shows up more frequently there).

d-torrance:
I'm going to move this to draft for now and look at it some more rather than just blindly trusting Claude.  Unassigning reviewers

Issue #4231, opened 2026-04-29
select on mutable hash tables


mahrud:
I vaguely recall this being discussed before, but couldn't find the issue: why doesn't `select*` work on mutable hash tables? It's true that the result might change at a later time, but the alternative (e.g. `select(keys h, blah)` or `selectKeys(new HashTable from h, blah)`) both seem aweful.

d-torrance:
Here's a silly contrived example that shows maybe why we wouldn't want it:

```m2
i1 : peek(x = new MutableHashTable from pairs toList(0..10))

o1 = MutableHashTable{0 => 0  }
                      1 => 1
                      2 => 2
                      3 => 3
                      4 => 4
                      5 => 5
                      6 => 6
                      7 => 7
                      8 => 8
                      9 => 9
                      10 => 10

i2 : peek selectPairs(x, (k,v) -> (x#k = v + 1;  even v))

o2 = MutableHashTable{0 => 1  }
                      2 => 3
                      4 => 5
                      6 => 7
                      8 => 9
                      10 => 11
```

mahrud:
I don't see why this is undesirable. It did the job correctly, no?
I agree that handling mutable objects requires extra care, but I don't see why it shouldn't be allowed.

mahrud:
I don't know if this arises to the level of discussing it in an internal meeting or in Atlanta, but would be curious about other opinions, e.g. @antonleykin or @mikestillman or @pzinn.

d-torrance:
> **Written by Claude** (Claude Opus 5, via Claude Code), not by @d-torrance, whose account posted it -- please weigh it accordingly.

> I vaguely recall this being discussed before, but couldn't find the issue

This is very likely it. The discussion predates the move to GitHub, so it was never
an issue — it was a file in the old `bugs/` tree, and here it is in full:

> There is no way to get just one element of a mutable hash table whose keys are
> unknown, because `select(1,x,i->true)` requires x to be immutable. Fix by
> relaxing that requirement, or by doing something else. We probably thought that
> requirement was important for thread safety, but now we can use mutexes or
> something.

So the restriction was deliberate rather than an oversight, and the stated reason
is thread safety — along with Dan's own view that mutexes would make it
unnecessary. That may be worth weighing against the alternatives you list.

Still reproduces on 1.26.06: `select(1, x, i -> true)` on a `MutableHashTable`
errors out.

[...]

PR #4387, opened 2026-05-29
Allow selectPairs & friends to work on mutable hash tables

d-torrance:
Closes: #4231

### DIFF
--- a/M2/Macaulay2/d/actors3.d
+++ b/M2/Macaulay2/d/actors3.d
@@ -2198,9 +2198,7 @@ applyPairs(e:Expr):Expr := (
     is a:Sequence do (
 	if length(a) == 2 then (
 	    when a.0
-	    is o:HashTable do (
-		if o.Mutable then WrongArgImmutableHashTable(1)
-		else mappairs(a.1, o))
+	    is o:HashTable do mappairs(a.1, o)
 	    -- # typical value: applyPairs, BasicList, Function, List
 	    -- # typical value: applyPairs, Dictionary, Function, List
 	    -- # typical value: applyPairs, Thing, Function, Iterator
--- a/M2/Macaulay2/d/actors4.d
+++ b/M2/Macaulay2/d/actors4.d
@@ -250,10 +250,7 @@ selectPairs(e:Expr):Expr := (
     is a:Sequence do (
 	if length(a) == 2 then (
 	    when a.0
-	    is obj:HashTable do (
-		if obj.Mutable
-		then WrongArgImmutableHashTable(1)
-		else selectPairs(obj.numEntries, obj, a.1))
+	    is obj:HashTable do selectPairs(obj.numEntries, obj, a.1)
 	    -- # typical value: selectPairs, BasicList, Function, List
 	    else select(pairs(a.0), a.1))
 	else if length(a) == 3 then (
@@ -262,10 +259,7 @@ selectPairs(e:Expr):Expr := (
 		if !isInt(n) then WrongArgSmallInteger(1)
 		else (
 		    when a.1 is obj:HashTable
-		    do (
-			if obj.Mutable
-			then WrongArgImmutableHashTable(2)
-			else selectPairs(toInt(n), obj, a.2))
+		    do selectPairs(toInt(n), obj, a.2)
 		    -- # typical value: selectPairs, ZZ, BasicList, Function, List
 		    else select(a.0, pairs(a.1), a.2, nullE, nullE)))
 	    else WrongArgZZ(1))
--- a/M2/Macaulay2/d/evaluate.d
+++ b/M2/Macaulay2/d/evaluate.d
@@ -2011,11 +2011,8 @@ export mapkeysmerge(f:Expr,o:HashTable,g:Expr):Expr := (	-- o is not Mutable
 mapkeysfun(e:Expr):Expr := (
      when      e is a:Sequence do
      if        length(a) == 2 || length(a) == 3
-     then when a.0 is o:HashTable 
-     do        
-     if        o.Mutable
-     then      WrongArgImmutableHashTable()
-     else      if length(a) == 2 then mapkeys(a.1,o) else mapkeysmerge(a.1,o,a.2)
+     then when a.0 is o:HashTable
+     do        if length(a) == 2 then mapkeys(a.1,o) else mapkeysmerge(a.1,o,a.2)
      else      WrongArgHashTable(1)
      else      WrongNumArgs(2,3)
      else      WrongNumArgs(2,3));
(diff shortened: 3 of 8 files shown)


antonleykin:
Just discussed this with @d-torrance
As long as `select` is allowed for `MutableList`, methods like `selectPairs` should be allowed for `MutableHashTables` for reasons of consistency.

Mutex notes:
* One would assume `selectPairs` locks the old table.
* If one does, the example in #4231 shall fail!
* `MutableList` has no mutex... yet `select` is implemented.

General notes on select on hash tables:
*  `selectPairs(ZZ, HashTable, Function)` is crazy: which elements it returns seem to depend on hash values. We think it should be disallowed (and wonder if it is used). 
* out of the ensemble of "selects" only `selectValues` is used (once!) across the distribution.

mahrud:
re: mutexes, I think this feature should remain minimal, with a documented caveat that one should know what they're doing and use mutexes when necessary themselves, because depending on the situation the right solution will vary.

re: selectPairs(ZZ,...), I think you should treat this as selecting from an unordered set. In fact, the only times I've used it the input was a Set, in order to avoid the cost of calling `toList` or `keys` first, which also constructs a list depending on the order of hash values!

mahrud (approved):

d-torrance:
Hmm, after playing around with this a bit more, I'm inclined to agree with Mahrud and think we should keep it as is and not deal with mutexes since we have a good chance of deadlocking.  For example, after adding `if obj.Mutable then lockRead(obj.mutex)` (and then the appropriate `unlock(obj.mutex)`'s as neded), we get:

```m2
i1 : x = new MutableHashTable from pairs toList(0..10);

i2 : selectPairs(x, (k,v) -> (x#k = v + 1;  even v)) -- deadlocks
  C-c C-c  C-c C-c
Exit (y=yes/n=no/a=abort/b=backtrace)? a
returning to top level
i3 : x#0 = 5 -- deadlocks again -- we never unlocked the mutex!
```

@antonleykin - Thoughts?

d-torrance:
Also, if we add this, should we also add `applyValues` and friends for mutable hash tables as well?

mahrud:
If `applyValues` could accept mutable hash tables, that could significantly simplify @mikestillman's upcoming PR, specifically commit https://github.com/mikestillman/M2/commit/f59d2e0ec1b74e4af1d3c90ebd698efdc33a9b81.

Mike, all the applyValues that you removed in that PR were added because they provided very significant speed improvements.

Since ComplexMap is becoming a mutable hash table, with Doug's suggestion above `applyValues(C.dd.map, ...)` would presumably change it in place, but the simple change `applyValues(new HashTable from C.dd.map, ...)` would stay efficient I believe.

d-torrance:
The current proposal creates a new hash table, but doing things in place makes a ton of sense!

mahrud:
Looking at Mike's commit, I'm realizing that it's going to conflict very badly with a PR I need to open for Varieties. If applyValues created a new hash table, it would actually help a lot, but I would also like it to eventually change in place.

What do you think? Should it create a new one for now and later change to in place? At that point we can replace with `applyValues(new HashTable from X, ...)`.

d-torrance:
That seems like a reasonable plan.  It's certainly easier in the short term as there's less code to write!

It might also make sense to do one big "mutable objects do things in place" transition at the same time and make that change the same time as doing similar things (`append`, `prepend`, `join`, etc.) for mutable lists.

d-torrance:
Okay -- I've changed things to also allow `applyPairs`, `applyKeys` and `applyValues` to act on mutable hash tables (creating a new object for now), and have also added a caveat to all 6 functions to exercise caution in this case.

mahrud (approved):
I think this checks all my boxes.

One thought: I was thinking about adding `hashTable MutableHashTable` to create a immutable copy, then it occurred to me whether it would be useful to have a O(1) way of making a mutable list or hash table immutable (e.g. instead of making a copy with `toList MutableList`, maybe we could have a special way to flip the mutable bit and return a List object).

d-torrance:
That seems like a cool idea!  One thing to consider -- how to deal with hash codes?  Immutables lists and hash tables have hash codes computed based on their contents, but this isn't the case for the mutable versions since the contents can change.

On one hand, it seems like the hash code should be changed to be based on the contents when going from mutable -> immutable, but on the other hand, we probably shouldn't be changing hash codes of objects after they're created...

d-torrance:
Discussed with Mike -- we'll gather more opinions before making a decision.  @jkyang92 - any thoughts?

mahrud:
Is this still going in this release?

d-torrance:
Possibly.  Mike and I wanted to get a bit more feedback before merging

mahrud:
Okay, as long as the Complexes and Varieties PRs aren't merged before this.

jkyang92:
I said some of this to @d-torrance on Zulip a couple days ago, but for completeness let me put them here as well.

My primary concern would be whether the fact that it returns a whole new hash table will be unexpected to users. I am also curious what workflow wants to apply `selectPairs` to a mutable hash table while returning a new mutable hash table and whether there's a better way to do it. In particular, I'd almost want it to return an immutable hash table even for a mutable hash table, but that seems a bit weird. I will note the closest replacement (which I believe @mahrud mentioned) `selectPairs(new HashTable from table, ...)` does return an immutable hash table.

As part of this concern I would also advocate that we should also have in-place versions of this for mutable hash tables and encourage the use of them for most applications involving mutable hash tables.

Additionally in response to an earlier example:
> ```m2
> i1 : x = new MutableHashTable from pairs toList(0..10);
> 
> i2 : selectPairs(x, (k,v) -> (x#k = v + 1;  even v)) -- deadlocks
>   C-c C-c  C-c C-c
> Exit (y=yes/n=no/a=abort/b=backtrace)? a
> returning to top level
> i3 : x#0 = 5 -- deadlocks again -- we never unlocked the mutex!
> ```

This line 2 is certainly a bad idea, it's changing the table it's selecting from, that's not to say that someone won't do it. The second deadlock is not salvageable anyways, after you return from that level, who knows what state the interpreter is in.

mahrud:
Since progress is halted, could someone explain why this is such a big deal when mutable lists already can have the same undefined behavior if users aren't careful? It's even worse because for instance `scan` and `apply` don't behave the same and `apply` seems to have some sort of overflow:
```m2
i136 : peek(L = new MutableList from (0..10))

o136 = MutableList{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10}

i137 : scan(L, i -> (print i; if L#?(i+1) then print(i, remove(L, L#(i+1)))))
0
(0, 1)
2
(2, 5)
3
(3, 8)
4
(4, 10)
6
7
9

i138 : peek L

o138 = MutableList{0, 2, 3, 4, 6, 7, 9}

i139 : peek(L = new MutableList from (0..10))

o139 = MutableList{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10}

i140 : peek apply(L, i -> (print i; if L#?(i+1) then print(i, remove(L, L#(i+1)))))
0
(0, 1)
2
(2, 5)
3
(3, 8)
4
(4, 10)
6
7
9
0
(0, 3)
2
stdio:152:61:(3):[4]: error: array index 6 out of bounds 0 .. 5
```

Editing mutable things while looping over them requires care but that doesn't mean we should tie our hands to avoid mistakes.

PR #4396, opened 2026-06-02
Engine polyhedral/cone functions and Triangulations package

mikestillman:
This PR contains two things, that perhaps should be distinct, but I found it hard to tease apart.  The Triangulations package depends on at least rawConeInteriorPoint, and I worked on them simultaneously, in order to test correctness of the raw engine routines.  The StringTorics package uses all of these new engine functions (especially rawGVInvariants).  

- New functions in the engine: rawGVInvariants, rawLatticePoints (both single file code from people in cytools group), rawLatticePointsNormaliz (calls normaliz to get these), and rawConeInteriorPoint (uses GLPK to solve the corresponding linear program).
- Triangulations: all functions are now documented, and a number of new functions have been added.  This package should now work around deficiencies in topcom, and generating triangulations where we cannot construct all of them (e.g. too many of them) can now be done.  Triangulations work now for both point configurations and vector configurations (including the non-acyclic case, i.e. constructing simplicial fans, given a set of rays).

AI disclaimer: I used AI to help in a number of ways:

- it wrote the first versions of rawLatticePointsNormaliz, rawConeInteriorPoint, although I tested these quite a bit on reflexive polytopes arising from StringTorics, and helped me organize how to add all the raw functions included here.
- it wrote many of the doc nodes in the Triangulations package, although I have modifed many of them.
- it helped me in a number of ways to fix my M2 level code generating all triangulations, (which is of course much slower than `allTriangulations`, when the number of triangulations is not too large), and to add secondary cones (which have been used to provide a faster method for determining if a triangulation is regular, however, it has been strongly tested against topcom's routines (I did this part).

edit: after the "big sort" PR, I will have to rearrange where these functions live in the engine, and fix the included hpp files...

edit2: There is a secondaryFan function, but it has not been tested very much yet.  So it should be made private for this PR.


mikestillman:
It is true that we need to be more systematic about making libnormaliz (and other) functions available.  This is just a start on that, doing only a couple of these routines. 

I also forgot to say that I added doxygen comments for these functions.  The first versions were written using AI (claude).  The doc was heavy-handed, probably because of what I told it, but I modified the doc to hopefully be less so, and to be accurate.

mikestillman:
I will likely unexport some functions, and possibly remove a couple of unexported functions (involving finding regular triangulation weights, which are not as good as topcom's function).

mikestillman:
I unexported secondaryFan, as it is not well tested yet.  One can still get it by `debug Triangulations`.

If this passes all tests, then it is ready to go in, except for doing the "great sort" PR first...  There will be a few changes to be made after that PR.

mikestillman:
I just made some minor changes: remove debug statement, add a note in the documentation, add my latest BUILD/mike/Makefile.

I am taking this off of "draft".  It is ready to go, if all tests pass (and no one sees any further issues).

PR #4401, opened 2026-06-02
Fix deprecation warning on adjointPolytope

trevorkarn:
Addresses #4399


d-torrance (approved):
Nice catch -- thank you!!

PR #4405, opened 2026-06-03
profiler updates + profileDepth + guard keyword

mahrud:
This is a subset of changes to the profiler from the internals workshop, extracted from [jkyang92/profile-improvements](https://github.com/jkyang92/M2/tree/profile-improvements).

- **added 'guard' keyword which blocks debugger and profiler**
- **added 'profileDepth' for tuning the profiler**
- **@LukeOeding and @rosie-0525 added resetProfileTable**
- **store labels in ProfileTable**
- **measure adjacentCode when profiling**
- **@LukeOeding and @rosie-0525 updated profiling documentation**

What is not included here:
- options added to profiler (teasing apart the commits was a little difficult, left for a later PR)
- start/end timer (I'm worried about the added overhead in more serious profiling runs, but it can be added as a separate feature later)
- flamegraph generator (should probably be added as a package or similar)

cc: @jkyang92


mahrud:
@d-torrance I really wish the build system would list all the errors at once ... this is so cumbersome!

M2/Macaulay2/d/profiler.dd
```diff
@@ -40,7 +41,14 @@ setupfun("stacktrace", stacktrace).Protected = false; -- will be overloaded in m
 -- TODO: this should be thread local as well
 ProfileTable := newHashTable(Tally, nothingClass);
 ProfileTable.beingInitialized = false; -- mutex locks will occur
-setupconst("ProfileTable", Expr(ProfileTable));
+profileTableS := setupvar("ProfileTable", Expr(ProfileTable));
+
+resetProfileTable(e:Expr):Expr := (
```
d-torrance, line 46: Should this check to ensure that `e` is a sequence of length 0?  Or do we care if we call it with arbitrary arguments?

M2/Macaulay2/packages/Macaulay2Doc/ov_repl.m2
```diff
@@ -439,6 +439,12 @@ document {
      ignoring interrupts."
      }
 
+document {
+    Key => "guard",
+    Headline => "guard a piece of internal code from the debugger and profiler",
+    TT "guard x", " -- prevent the debugger and profiler from interacting with the code."
```
d-torrance, line 445: I think this should be clarified.  Currently, `guard` only prevents the debugger/profiler from interacting with the top-level code, but its children aren't guarded.

For example:

```m2
i1 : f = x -> x^2; g = y -> y^3;

i3 : profile(guard f g 3)

o3 = 729

i4 : profileSummary

o4 = #run  cost    code              position       
     1     17.85   g                 stdio:2:16-2:19
     1     .0001s  total time (cpu)                 
```

`f` was guarded, but `g` wasn't

d-torrance (changes requested):
This is great!  Just a couple comments

M2/Macaulay2/packages/Macaulay2Doc/ov_repl.m2
d-torrance: Also, maybe add an example

M2/Macaulay2/d/actors5.d
```diff
@@ -1867,6 +1869,10 @@ export setErrorDepth(b:ushort):void := (
      errorDepth = b;
      setGlobalVariable(errorDepthS,toExpr(b));
      );
+export setProfileDepth(b:ushort):void := (
+     profileDepth = b;
+     setGlobalVariable(profileDepthS,toExpr(b));
+     );
```
d-torrance, line 1875: This function is never called -- do we need it?

M2/Macaulay2/d/profiler.dd
```diff
@@ -65,6 +66,10 @@ increment(key:Expr, data:Sequence):Expr := (
 -- increments the stopwatch, pops the stack, and returns the value!
 -- Note: wallTimer calls evalraw, not eval
 measure(c:Code):Expr := stackpop(increment(stackpush(codePosition(c)), wallTimer(c)));
```
d-torrance, line 68: Do we still need the old `measure(Code)` function?  We get a unused function warning when compiling

d-torrance (changes requested):
Reviewed w/ @klee669 

In addition to the comments above, it would be good to add a couple tests to ensure that profiling works as expected, e.g., maybe check that `ProfileTable` has the right keys, maybe before and after changing `profileDepth`.

klee669 (commented):
Could we document the full semantics of `profileDepth`, perhaps similarly to `errorDepth`? The implementation accepts values from 0 to 255, but the current documentation only mentions the default and `profileDepth = 1`. In particular, it would be helpful to explain what 0, 1, 2, and 3 mean, and that increasing the value reduces the amount of code being profiled.

PR #4406, opened 2026-06-04
Varieties v0.6: complexes of sheaves + weighted projective stacks

mahrud:
This is a major overhaul of the package, including changes implemented at the AIM and Utah workshops and contributions by @btotaro for supporting subvarieties of weighted projective stacks.

I will add a few more commits to make the package stylistically consistent before merging.

@mikestillman I think it's easiest if I cherry-pick your commits fixing Complexes on top of this PR. Let me know what you think.

PR #4407, opened 2026-06-04
Varieties v0.5: complexes of sheaves

mahrud:
For some reason #4406 keeps warning that it can't be merged because of conflicts, and I found an online suggestion to split large PR's to have less than 100 commits.

### DIFF
--- a/M2/Macaulay2/packages/Varieties/SheafComplexes.m2
+++ b/M2/Macaulay2/packages/Varieties/SheafComplexes.m2
@@ -31,6 +31,7 @@ isMorphism SheafMap := isAbelianCategory CoherentSheaf := x -> true
 
 complex CoherentSheaf := Complex => lookup(complex, Module)
 
+
 -----------------------------------------------------------------------------
 -- Basic operations between sheaves, complexes, etc.
 -----------------------------------------------------------------------------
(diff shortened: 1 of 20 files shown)


M2/Macaulay2/packages/Varieties/SheafComplexes.m2
```diff
@@ -54,19 +52,21 @@ Complex ** CoherentSheaf := Complex => {} >> opts -> (C, F) -> tensor(C, F, opts
...
 
-sheaf Complex := Complex => C -> C.cache.sheaf ??= (
+sheaf          Complex  := Complex =>     C  -> sheaf(variety ring C, C)
+sheaf(Variety, Complex) := Complex => (X, C) -> C.cache.sheaf ??= (
     if isSheafComplex C then return C;
     (lo, hi) := C.concentration;
-    if lo === hi then return complex(sheaf C_lo, Base => lo);
-    D := complex applyValues(C.dd.map, sheaf);
+    if lo === hi then return complex(sheaf_X C_lo, Base => lo);
+    D := complex applyValues(C.dd.map, sheaf_X);
```
d-torrance, line 60: In #4395, the proposal is that `ComplexMap` objects will have a `Function` key for lazy evaluation instead of just the various integer indices.  So `applyValues` will cause this to fail since we'll be calling `sheaf_X` on some function.

mahrud: I'm aware. I don't think the proposed architecture is compatible with my goals from complexes and resolutions.

PR #4410, opened 2026-06-04
DGAlgebras: fix minimization in minimalSemifreeResolution

kellerlv:
## Summary

`minimalSemifreeResolution(A, M)` could leave residual scalar-unit entries
in the output differentials — acyclic, but not minimal over the DG algebra.
`isMinimalSemifreeResolution` only checked the augmentation-ideal property,
so non-acyclic complexes with aug-ideal entries spuriously returned ``true``.
This PR fixes both ends and adds a proper minimization post-processor.

## Changes to ``DGAlgebras.m2``

1. **``isMinimalSemifreeResolution(M)``** now also verifies acyclicity
   (``H_i = 0`` for ``1 ≤ i ≤ maxVdeg − 1``). The previous predicate was
   necessary but not sufficient.

2. **New method ``minimizeDGModule(M)``**: proper graded Gauss elimination
   on a semifree DGModule. For each ``(i, j, u)`` where ``d(e_j)`` has a
   scalar-unit ``u`` on row ``i``, modify same-degree generators by
   ``M.diff#l ← M.diff#l − (c_{l,i}/u)·M.diff#j`` and drop both ``e_i`` and
   ``e_j`` symmetrically. The ``d² = 0`` identity guarantees the ``j``-th entry
   of higher-degree generators is ``0`` in the new basis.

3. **``minimalSemifreeResolution`` gains ``Minimize => true`` option** (default).
   Post-processes with ``minimizeDGModule``. Pass ``Minimize => false`` to
   recover the pre-fix behavior exactly.

4. **``adjoinVariables`` uses ``degree z + tempDegree`` uniformly** instead of
   bifurcating on ``A.isHomogeneous``. The previous else-branch emitted
   length-1 degrees that mismatched length-2 ``A.Degrees`` from
   ``koszulComplexDGA``, causing ``freeDGAlgebra`` to error on ideals with
   mixed-degree generators (e.g., the defining ideal of ``k[t³, t⁴, t⁵]``
   in ``k[a,b,c]``).

## New tests in ``DGAlgebras/tests.m2``

13 new TEST blocks covering:

- CI Tor matching: ``k`` over ``k[a,b]/(a², b²)``.
- HMF Ω³ over ``(a⁴, b⁴, c⁴)``: ranks ``{6, 9, 13, 18}`` regression case.
- Strengthened predicate rejecting a deliberately broken non-acyclic DGModule.
- Mixed-degree acyclic closure on numerical semigroup ``k[t³, t⁴, t⁵]``.
- Weighted-graded numerical semigroup ``k[t⁴, …, t⁷]``.
- Algorithmic consistency: ``Minimize`` toggle agreement, idempotence,
  depth monotonicity.
- Deep HMF Ω⁹ over ``(a⁴, b⁴, c⁴)``.
- ADE A₂ cusp, generic determinantal 2×3 minors, Avramov-Buchweitz line
  support ``R/(a)`` over ``k[a,b]/(a²,b²)``.
- Twisted cubic in P³, Reiten trivial extension (Fibonacci Betti numbers),
  A₃-type matrix factorization module.

## Backward compatibility

- ``minimalSemifreeResolution`` defaults to ``Minimize => true``. Pass
  ``Minimize => false`` to recover the exact prior output.
- ``isMinimalSemifreeResolution`` is strictly more restrictive — anything
  that returned ``true`` and *was* acyclic still returns ``true``; anything
  that returned ``true`` despite being non-acyclic now correctly returns
  ``false`` (this was a bug being fixed).

## Test plan

- [x] All 43 existing DGAlgebras TEST blocks pass.
- [x] 13 new TEST blocks added — 56 total, all pass via ``check DGAlgebras``.
- [x] Verified on diverse literature cases (Tate / Avramov pure-power CIs,
      Eisenbud-Peeva HMFs, Golod, Burch, Stanley-Reisner, numerical
      semigroup, ADE, Avramov-Buchweitz support varieties, Lescot fiber
      products, edge ideals, char-2 sanity).
- [x] HomotopyLieAlgebra (downstream package using ``minimalSemifreeResolution``)
      passes its full test suite against the patched DGAlgebras.

### DIFF
--- a/M2/Macaulay2/packages/DGAlgebras/tests.m2
+++ b/M2/Macaulay2/packages/DGAlgebras/tests.m2
@@ -1191,3 +1191,347 @@ assert(all(0..3, n -> (
     source cmn == target cmn and cmn == id_(source cmn)
     )))
 ///
+
+-- ============================================================
+-- minimizeDGModule + isMinimalSemifreeResolution (strengthened):
+-- Locks in the fix for the bug where minimalSemifreeResolution could
+-- leave residual scalar-unit entries (non-minimal over the DG algebra)
+-- and where isMinimalSemifreeResolution previously accepted any
+-- non-acyclic complex that happened to have aug-ideal entries.
+-- ============================================================
+
+-- Test: on a complete intersection, minimal A-semifree resolution
+-- per-degree ranks match Tor^S(M_S, k) by uniqueness of minimal res.
+TEST ///
+kk = ZZ/101
+P = kk[a,b]
+Q = P/(ideal(a^2, b^2))
+A = koszulComplexDGA ideal Q
+M = coker vars P
+F = minimalSemifreeResolution(A, M, EndDegree => 3)
+-- Expected: ranks = Betti numbers of k over Q = {1, 2, 3, 4, 5}.
+ranks = apply(max(F.Degrees / first) + 1, d -> #select(F.Degrees, e -> first e == d))
+assert(ranks == {1, 2, 3, 4, 5})
+assert(isMinimalSemifreeResolution F)
+-- Idempotent post-processing.
+F2 = minimizeDGModule F
+ranks2 = apply(max(F2.Degrees / first) + 1, d -> #select(F2.Degrees, e -> first e == d))
+assert(ranks == ranks2)
+///
+
+-- Test: the HMF Omega^3 regression case. Original algorithm gave a
+-- non-minimal output at hom-deg >= 2 (rank 14 at deg 2 instead of 13);
+-- patched gives the canonical minimal Betti numbers from Tor^S(M3, k).
+-- Uses EndDegree => 1 to keep the check fast while still exhibiting
+-- the raw-vs-minimized discrepancy.
+TEST ///
+kk = ZZ/101
+P = kk[a, b, c]
+Q = P/ideal(a^4, b^4, c^4)
+A = koszulComplexDGA ideal Q
+phi = map(Q, P)
+NS = coker (phi matrix{{P_0, P_1, P_2}, {P_1, P_2, P_0}})
+Fres = freeResolution(NS, LengthLimit => 4)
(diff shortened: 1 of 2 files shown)


M2/Macaulay2/packages/DGAlgebras/tests.m2
```diff
@@ -1191,3 +1191,300 @@ assert(all(0..3, n -> (
...
+assert(ranks == take(expected, #ranks))
+assert(isMinimalSemifreeResolution F)
+///
+
+-- Test: algorithmic consistency invariants —
+--   I1: Minimize=>false then minimizeDGModule gives same result as default.
+--   I2: minimizeDGModule is idempotent.
+--   I3: ranks at lower hom-deg agree across different EndDegree settings.
+TEST ///
+needsPackage "DGAlgebras"
```
d-torrance: Calling `needsPackage` on the package you're testing isn't necessary -- it's already loaded by default.

Also, this test (and the next one) are too big.  On my system:

```m2
i5 : check({48, 49}, "DGAlgebras")
 -- capturing check(48, "DGAlgebras")            -- 50.5178s elapsed
 -- capturing check(49, "DGAlgebras")            -- 89.1936s elapsed
```

Are there smaller cases you can use for these?

d-torrance (changes requested):

M2/Macaulay2/packages/DGAlgebras/tests.m2
kellerlv: Whoops, yes. Those should have been commented out.

d-torrance:
One other comment -- are these changes worth bumping the version to 2.1?

kellerlv:
> One other comment -- are these changes worth bumping the version to 2.1?

Yup, will do one this last build finishes.

d-torrance (approved):
Thanks!

PR #4417, opened 2026-06-05
Improve the search bar in the html documentation

d-torrance:
We use [lunr.js](https://lunrjs.com/) to power the search bar in the html documentation, which currently just sends folks to Google.

We add a couple new JavaScript files to the *Style* package, one to power the search and one containing an index constructed by the calls to `installPackage` during the build.

In addition, there's a new workflow that I'll plan on running close to each release to update the search index file in the repository (and also the prism.js highlighting symbols while we're at it).,

Here it is in action:  https://d-torrance.github.io/M2

## AI Disclosure

The code was almost completely written by Claude with lots of human oversight.


joel-dodge:
I tested this a bit via the link in the description. The UX is very nice and e.g. searching for "modules" turned up all the nodes I know of that I wanted to see (within a page or two).

M2/Macaulay2/packages/Style/doc.css
```diff
@@ -104,6 +104,81 @@ div#buttons {display: flex;}
...
+.search-pkg {
+  font-size: 75%;
+  color: #888;
+  flex-shrink: 0;
+}
+
+.search-tagline {
+  font-size: 85%;
+  color: #444;
+  flex-basis: 100%;
```
seangrate, line 161: I believe `flex-basis` here is set to `100%` simply to push the tagline to the next line, and with the text being aligned to the right, this causes the tagline to visually appear a bit far from the search result and associated package. 

Purely an aesthetic preference, but I would be tempted to have the tagline right below the search result:
Image
But this is not a strong preference :)

seangrate (approved):
A welcome change for sure. Only a minor suggestion on how to visually display the search results.

M2/Macaulay2/packages/Style/doc.css
d-torrance: Great idea!  I had Claude fix it to match your image.  The example documentation linked above is re-building right now and should be ready in a couple hours with the new look.

klee669 (approved):
I tried it out using the link, and it works great! One question: is there still a way to send the query to Google, for example, when the local search doesn’t find what we’re looking for?

d-torrance:
No Google backup at the moment -- but that's a good idea!  I'll play around with it

PR #4400, opened 2026-06-02
Add recipes for building PALP

d-torrance:
[PALP](https://hep.itp.tuwien.ac.at/~kreuzer/CY/CYpalp.html) is a **P**ackage for **A**analyzing **L**attice **P**olytops that @mikestillman would like to call from Macaulay2 for the string torics package.

We add recipes to build PALP in both the autotools and cmake builds.  (It's already in Ubuntu and I hope to have it on our Homebrew tap soon.)

*Draft for now to test the builds*

## AI Disclosure

Claude Chat helped me debug some cmake issues.

PR #4418, opened 2026-06-05
A few palp updates

d-torrance:
This is a quick followup to #4400.

A few small changes:

* Don't build palp by default (autotools).  You can still build it by passing `--enable-build-libraries=palp` to `configure`, but there's no reason yet to require its presence.  (This was causing all the PPA builds to fail, as they were all trying to download the tarball but internet access isn't available during Debian package builds!)
* Use `https://` for all palp URL's. (I was using a few `http://`'s, so we were redirecting needlessly when downloading the tarball.)
* Go ahead and install palp in the GitHub builds (both in Ubuntu via `apt` and macOS via `brew` from the M2 tap) so it's there when we need it.

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -64,7 +64,7 @@ jobs:
           brew config
           brew update
           brew tap macaulay2/tap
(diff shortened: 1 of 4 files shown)


d-torrance:
Going to go ahead and merge this so the PPA builds stop failing

PR #4424, opened 2026-06-08
Fix -Wparentheses compiler warning

d-torrance:
Fixes the following:

```c++
../../../../Macaulay2/e/buffer.cpp: In member function ‘void buffer::put(cci_srcptr)’: ../../../../Macaulay2/e/buffer.cpp:175:28: warning: suggest parentheses around ‘&&’ within ‘||’ [-Wparentheses]
  175 |       mpfi_is_zero(&x->re) && mpfi_is_zero(&x->im)) {
      |       ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~
```

## AI Disclosure

No AI used

### DIFF
--- a/M2/Macaulay2/e/buffer.cpp
+++ b/M2/Macaulay2/e/buffer.cpp
@@ -135,7 +135,7 @@ void buffer::put(mpfi_srcptr x)
 
 void buffer::put(cc_doubles_srcptr x)
 {
-  if (x->re !=0 || (x->re == 0 && x->im == 0)) {
+  if (x->re !=0 || x->im == 0) {
     put(x->re);
     if (x->im > 0)
       put('+');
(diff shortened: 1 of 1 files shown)


M2/Macaulay2/e/buffer.cpp
```diff
@@ -172,7 +172,7 @@ void buffer::put(cc_srcptr x)
 void buffer::put(cci_srcptr x)
 {
   if (!mpfi_is_zero(&x->re) ||
-      mpfi_is_zero(&x->re) && mpfi_is_zero(&x->im)) {
+      (mpfi_is_zero(&x->re) && mpfi_is_zero(&x->im))) {
```
jkyang92: This check seems silly. It's `!a || (a && b)` which is equivalent to `!a || b`. The change itself is fine.

d-torrance: Nice catch!  I've updated the commit to use this simpler logic (and also the other two complex-number-printing routines in this file)

jkyang92 (approved):

klee669 (approved):
Looks good to me. Also, using mpfr_zero_p is appropriate for the zero checks. Thanks!

PR #4425, opened 2026-06-08
move tries block in isIsomorphism

Devlin-Mallory:
A small change in order of operations for `isIsomorphic`: before, when running `isIsomorphic(M, N)`, the (deterministic) check that the degrees of `M` could be shifted to the degrees of `N` was occurring after the line `if tries > 1 then return any(tries, i -> isIsomorphic(N, M, o, Tries => 1))`, so the same check was getting run `tries` many times; if this check failed, we'd scan a list of `Tries` many `false` entries. Now the degree check occurs first, so that we reject immediately. For a toy example where this speeds things up considerably, try:

```
R = (ZZ/5)[x,y]
elapsedTime isIsomorphic(R^{1}/x,R^{2}/y,Tries=>20000,Strict=>true)
```

### DIFF
--- a/M2/Macaulay2/packages/Isomorphism.m2
+++ b/M2/Macaulay2/packages/Isomorphism.m2
@@ -235,10 +235,6 @@ isIsomorphic(Module, Module) := Boolean => o -> (N, M) -> (
     if isFreeModule M and isFreeModule N then
     return isIsomorphismFree(N, M, o);
     --
-    tries := o.Tries ?? defaultNumTries char S;
-    if tries > 1 then return any(tries,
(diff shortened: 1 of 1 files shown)


d-torrance (approved):
This is great -- thanks!!

mahrud:
I appreciate the fix, but if an author review was requested, I would have asked for a test and removing redundant empty lines. Also in #4322, I requested a redundant addition to this package be removed, but it was merged anyway.

d-torrance:
D'oh -- sorry!  It will be good when we have some explicit PR policies rather than "Doug gets click-happy near release time"...

PR #4434, opened 2026-06-10
Fixed bug in A1BrouwerDegrees

wgabrielong:
A bug was identified and fixed by Claude Code in `A1BrouwerDegrees`.

Consider the quadratic form
`beta = makeDiagonalForm(QQ, (-1,-1,-1,-1,-1,1))`

Since `beta ≅ <-1,-1,-1,-1> + H`, its anisotropic part is `<-1,-1,-1,-1>`, of rank 4.
However, `getAnisotropicPart beta` silently returns the rank-2 form `<-1,1>` — the
hyperbolic plane, which is isotropic and violates the Witt decomposition identity:

```
i6 : a = getAnisotropicPart beta
o6 = | -1 0 |
     | 0  1 |
i8 : isAnisotropic a
o8 = false
i9 : isIsomorphicForm(beta, addGW(a, makeHyperbolicForm(QQ, 2*getWittIndex beta)))
o9 = false
```

**Cause.** `reduceAnisotropicPartQQDimension3` (`Code/Decomposition.m2`) must return
`<alpha>` such that `beta + <-alpha>` has anisotropic dimension 2. It chooses `alpha`
from congruence conditions at the finite primes only, ignoring the real place, where the
anisotropic dimension equals `|signature|`. Here the running form `<-1,-1,-1,-1,-1,1,1>`
has signature -3 and an empty relevant-prime list, so the solver returns `alpha = 1`;
the real anisotropic part is negative definite, so `alpha` must be negative. The
anisotropic dimension rises from 3 to 4 instead of dropping to 2, the `== 2`/`== 1`
branches of `getAnisotropicPartQQ` never fire, and a truncated form is returned with no
error.

**Fix (by Claude Code).** The congruences determine `alpha` only modulo the product `M`
of the moduli, so the real place is accommodated by shifting to a representative of the
correct sign:

```m2
M := product(L1 | L2);
if getSignature(beta) == -3 and alpha > 0 then alpha = (alpha % M) - M;
if getSignature(beta) == 3 and alpha < 0 then alpha = (alpha % M) + M;
```

`getAnisotropicPart beta` now returns `<-1,-1,-1,-1>`. Tests 48–49 add regression
coverage (the existing Test 47 never reached the reduction chain); the full package
suite passes under Macaulay2 1.25.11.

### DIFF
--- a/M2/Macaulay2/packages/A1BrouwerDegrees.m2
+++ b/M2/Macaulay2/packages/A1BrouwerDegrees.m2
@@ -1,8 +1,8 @@
 --A1BrouwerDegrees.m2
 newPackage (
     "A1BrouwerDegrees",
-    Version => "2.0",
-    Date => "October 13, 2025",
+    Version => "2.1",
+    Date => "July 2, 2026",
     Authors => {
         {Name=> "Stephanie Atherton",
     Email => "satherton@student.otis.edu"},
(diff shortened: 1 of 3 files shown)


timduff35 (approved):
I verified all tests pass and package builds with the latest development. The code changes are minimal and fit the PR's description.

seangrate (approved):
Looks good to me. The additional comments in the code are appreciated!

PR #4435, opened 2026-06-11
Allow Core pushforward to accept more ring maps

joel-dodge:
When computing the push-forward of a module `M` along a ring map `f`, the core pushforward.m2 includes a check enforcing the invariant that a certain degree map used in it's computations is a section of the degree map of `source f` along `f`. This ends up placing a restriction on which ring maps `f` can be used in this calculation that is I think more restrictive than it has to be. 

There is a comment in the code that the degree map used is somewhat arbitrary and chosen just to ensure a nice property of the symmetricPower functor. I was not able to identify how this pushforward functionality interacts with the symmetricPower method though. 

This PR introduces a more general construction of the degree map for the projection from the graph ring back to `source f` - one which specializes to the previous one. My understanding is that the degree map used ends up modifying the result just in terms of the shifts of the free modules in the resulting presentation. 

Previously the test case that is added here would fail on computing the pushforward module:
```
i1 : R = kk[a, b, DegreeGroup => ZZ^2];
S = kk[s, t, u, DegreeGroup => ZZ^3]/ideal {s^3, t^5, u^7};
f = map(S, R, {s^2, t^2*u^3}, DegreeMap => d -> d_0 * {2, 0, 0} + d_1 * {0, 2, 3});
P = pushForward(f, S^1);

o3 : RingMap S <-- R
stdio:4:15:(3):[1]: error: not implemented yet: unexpected degree map of ring map
```

@mikestillman I think you are the original author of the `pushNonLinear` method. Do you have any context you can provide on the necessity of the check that I am proposing to remove here? To my understanding this algorithm is quite happy to compute the correct result in more generality and the check is not needed but I would love to understand more what this was trying to guard against.


M2/Macaulay2/m2/pushforward.m2
```diff
@@ -129,47 +135,50 @@ pushNonLinear := (opts, f0, M) -> (
...
+    --  we should really be *lifting* the result to S along the natural map
+    --  S -> G but lifting along an arbitrary RingMap is not supported.
+    mapback := map(S, G, map(S^1, S^r, 0) | vars S);
+    result := phiS^-1 mapback selectInSubring(if r > 0 then 1 else 0, generators g);
+
     -- MES: check if the monomial order restricts to R.  If so, then do `` forceGB result ''
-    phiS^-1 mapback selectInSubring(if numgensR > 0 then 1 else 0, generators g))
+
+    -- repair homogeneity if we had it to begin with.
+    if isHgs then result = try map(target result, , result) else result;
```
mahrud: In my experience doing this fakes homogeneity rather than repairing it, possibly giving a false impression by covering up errors or mistakes.

joel-dodge: By "fakes homogeneity" do you mean that the result is not homogeneous even in cases where `isHomogeneous` may say that it is?

Alternatively maybe you mean that while the result may be homogeneous it is not graded in a way that allows for a homogenous bijection with `m`?

mahrud: `isHomogeneous` fails when either the source or target of the matrix are not correct. Here you're assuming that the target is correct and updating the source accordingly. But if there was a mistake that made the source incorrect, how do you know the target was correct? If the target was incorrect, now the source is set incorrectly based on that too, but in a way that everything looks homogeneous.

joel-dodge: Ack that makes perfect sense. In that case I think it makes sense to just revert the change calling `map` and return `result` as is. This leaves it up to the caller of this method to know what the right source / target are and avoids giving a potentially misleading impression that the map returned is the right graded presentation for the push forward module being computed.

Does this sound right to you?

edit: the clarification I am looking for here is that there is no way to know what the "right" source and target are so no attempt should be made to make the result homogeneous.

joel-dodge:
OK the build failure is from generating examples and this has put my finger on the dependence between symmetricPower and the degree manipulations in the pushforward package. 

IMO ideally the symmetricPower context would be able to set up the degrees on the result of pushforward and not require injecting this specific degreemap into the push-forward logic. I will see what that might look like.

M2/Macaulay2/m2/pushforward.m2
mahrud: Oh no there is definitely a right source and target that should be set by the computation. My point is that if we get one or both wrong, we should hope to find out that something went wrong because the result is not homogeneous. Fixing homogeneity this way covers up potential mistakes.

joel-dodge: Ok I understand. I'll think more about the right way to preserve homogeneity.

joel-dodge:
@mahrud following your nudging, I landed on an approach that follows the intention of the previous impl - setting a custom degree map on the projection `G -> S` - but that just does some more work to produce a map which is a splitting in more generality than the previous one. 

I ~still~ don't understand clearly what I think ought to be done in the case when `degreeGroup source f` does not inject into `degreeGroup target f` and I have placed a note in a todo to this effect. Note this is not a regression as certainly any case like this would not have passed the previous invariant check anyways.

M2/Macaulay2/m2/pushforward.m2
```diff
@@ -111,7 +111,36 @@ ordertab := new HashTable from {
...
+
+    -- if M is not injective then we cannot produce a section
+    if (kernel M != 0) then return null;
+
+    D := image M;
+    (d) -> (
+        -- d' is the element of degreeGroup target f corresponding to d
+        d' := matrix transpose {d};
+        -- get coefficients for d' against generators for G ~ D.
+        -- this produces mild nonsense if d' is not in D but that doesn't matter
```
joel-dodge, line 140: I considered also checking explicitly whether `d' % D == 0` and returning `{0, ..., 0}` in that case. It is not obvious how to interpret the return value of this function in cases when the argument is not in the image but we only ever apply the resulting `mapback` to elements in `image f` so I could not convince myself that it mattered.

joel-dodge:
@mahrud you mentioned that there was an example with negative grading that you found did not work well with these changes. Can you reproduce it and share the example and your expectations for it?

M2/Macaulay2/m2/pushforward.m2
```diff
@@ -150,26 +173,25 @@ pushNonLinear := (opts, f0, M) -> (
...
 	DegreeLimit           => opts.DegreeLimit,
-	PairLimit             => opts.PairLimit);
+	PairLimit             => opts.PairLimit
+    );
+
+    -- todo - what should happen here if the map on degree groups induced by f is not injective?
+    mapback := if (degmapback := makeSectionForDegreeMap(f)) =!= null then
+        map(S, G, map(S^1, S^r, 0) | vars S, DegreeMap => degmapback)
+    else
+        map(S, G, map(S^1, S^r, 0) | vars S);
```
joel-dodge, line 192: perhaps raising an error here as the original code did would be preferable

joel-dodge:
@mahrud pinging again in case this one slipped through the cracks.

joel-dodge:
As one data point in favor of this approach, I have confirmed that with with changes proposed here ALL of the "bug" test cases listed in the gutter of PushForward.m2 now work. I won't move them to actual test cases until the big PushForward PR is merged just to avoid merge hell.

joel-dodge:
@d-torrance it seems this PR will need a reviewer - is it possible to place this on the rails of the new PR review process?

d-torrance:
Yup -- that will be happening soon!

joel-dodge:
I just updated with two new commits:
* merged in a newer development branch
* the PushForward package that relies on the changes here had some examples of test cases that did not previously work but that now can run after the changes to the core pushforward functionality here. I have moved them to be honest unit test cases.

joel-dodge:
Reviewers I am happy to chat or provide any context on these changes if it would help make reviewing easier.

M2/Macaulay2/m2/pushforward.m2
ggsmith: Raising an error does seem like a good idea to me.

M2/Macaulay2/packages/PushForward/test.m2
```diff
@@ -688,12 +688,71 @@ N = pushFwd(f, M)
...
--- it would be useful if there were a standardized way to encode functionality
--- gaps in macaulay2.
--- NONTEST - this should work but doesn't
+-- test 36
+TEST ///
+s = symbol s; t = symbol t
+kk = ZZ/101
+A = frac(kk[s,t])
+L = A[symbol a.. symbol d]/(d-t, a-s, b*c-s*t, b^2-(s/t)*c^2)
+describe L
```
ggsmith: Omit the `describe L`.  Doesn't seem to be part of the test.

M2/Macaulay2/packages/PushForward/test.m2
```diff
@@ -688,12 +688,71 @@ N = pushFwd(f, M)
...
+ML = pushFwd(map(L,frac A), L^1) -- dim 4, free -- FAILS
```
ggsmith: Omit the `-- FAILS`, assuming that it now works.

M2/Macaulay2/packages/PushForward/test.m2
```diff
@@ -688,12 +688,71 @@ N = pushFwd(f, M)
...
+///
```
ggsmith, line 695: Add an `assert` statement?

M2/Macaulay2/packages/PushForward/test.m2
```diff
@@ -688,12 +688,71 @@ N = pushFwd(f, M)
...
+kk = ZZ/101
+A = frac(kk[s,t])
+L = A[symbol b, symbol c]/(b*c-s*t, b^2-(s/t)*c^2)
+basis L
+describe L
+inc = map(L, A)
+assert isInclusionOfCoefficientRing inc
+assert isModuleFinite L
+pushFwd inc
+ML = pushFwd(map(L,frac A), L^1)
```
ggsmith, line 708: Add an `assert` involving ML.

M2/Macaulay2/packages/PushForward/test.m2
```diff
@@ -688,12 +688,71 @@ N = pushFwd(f, M)
...
+
+-- test 37
+TEST ///
+debug needsPackage "PushForward"
+s = symbol s; t = symbol t
+kk = ZZ/101
+A = frac(kk[s,t])
+L = A[symbol b, symbol c]/(b*c-s*t, b^2-(s/t)*c^2)
+basis L
+describe L
```
ggsmith: I would omit `basis L` and `describe L` because there aren't really part of the test.

M2/Macaulay2/packages/PushForward/test.m2
```diff
@@ -688,12 +688,71 @@ N = pushFwd(f, M)
...
+TEST ///
+debug needsPackage "PushForward"
+s = symbol s; t = symbol t
+A = QQ
+L = A[symbol b, symbol c]/(b*c-13, b^3-c^2)
+inc = map(L, A)
+assert isInclusionOfCoefficientRing inc
+assert isModuleFinite L
+(LA, bas, pf) = pushFwd inc -- this works
+pf(b^2+c^2) -- maybe a better way?
```
ggsmith: Again need an `assert` involving some of `(LA, bas, pf)`.

M2/Macaulay2/packages/PushForward/test.m2
```diff
@@ -688,12 +688,71 @@ N = pushFwd(f, M)
...
+
+-- test 39
+TEST ///
+s = symbol s; t = symbol t
+kk = ZZ/101
+A = frac(kk[s,t])
+L = A[symbol b, symbol c]/(b^2-(s/t)*c^2 - c, c^3)
+inc = map(L, A)
+-- pushForward(inc, L^1) this fails due to issue computing presentation of fraction field ring
+pushFwd(inc, L^1)
```
ggsmith: Need an `assert` involving the final `pushFwd`.

M2/Macaulay2/packages/PushForward/test.m2
```diff
@@ -688,12 +688,71 @@ N = pushFwd(f, M)
...
+kk = ZZ/101
+A = kk[s,t]
+C = A[x,y,z]/(x^2, y^2, z^2)
+phi = map(C,A)
+f = map(C^1, A^4, phi, {{x,s*y,t*y, z}})
+ker f
+
 kk = ZZ/101
-M = first pushFwd kk
-assert(M == module kk)
```
ggsmith, line 698: Did one really eliminate the only `assert`?  This doesn't make sense to me.

M2/Macaulay2/packages/PushForward/test.m2
```diff
@@ -688,12 +688,71 @@ N = pushFwd(f, M)
...
+A = kk[s,t]
+B = frac A
+C = B[x,y,z]/(x^2, y^2, z^2)
+phi = map(C,B)
+f = map(C^1, B^3, phi, {{x,s*y,z}})
+ker f
```
ggsmith: What is `test 40` actually indented to test?

ggsmith (changes requested):
@trevorkarn and I didn't have an significant concerns related to the code. However, the tests could be improved: rather than just seeing that `pushforward` runs, it would be nice to verify the output is correct.

M2/Macaulay2/packages/PushForward/test.m2
joel-dodge: As you can see in the diff the tests that are added in this PR are all copied from the end of the PushForward.m2 file where they were previously placed as cases that did not work but did not come with any lines indicating the expected output. I had just copied the code verbatim to test cases but I can add some asserts.

M2/Macaulay2/packages/PushForward/test.m2
joel-dodge: the diff is confusing. This was part of a block of code that was not run as a test but which i had previously committed as an aspirational test case for functionality that should work. I changed my mind about how important it is to call this out with a non-running test-like block of code so I removed it in this PR. sorry for the confusion.

joel-dodge:
@ggsmith @trevorkarn all the comments make sense! I will make these fixes and ping you when this is ready for review again.

joel-dodge:
> Raising an error does seem like a good idea to me.

@ggsmith i reworked this a bit and the current state of the code errors only in case the input data is homogeneous which at least is the only case where you would ever care about having a coherent relationship between degrees in the source and target of the ring map. 

I am actually not sure if I think that it makes sense to raise an error at all though and would like to hear more of your thoughts. It is true that having a degree map here that is a section of the degree map of `f` is desirable for having the degrees set correctly on the module that is produced. On the other hand, in cases where there is no section - that is in cases where the map on degree groups is not injective - I am not sure what the caller of this function would even expect in terms of the state of the pushforward module with respect to the grading. It seems reasonable to me to compute the pushforward module as an abstract (non-graded) thing in this case instead of erroring. wdyt?

joel-dodge:
@ggsmith I am interpreting your 👍 as agreement that not erroring is reasonable and so have pushed another commit to make it so that no error is raised in case a section of the degree map of `f` is impossible to construct.

joel-dodge:
@ggsmith @trevorkarn any thoughts on when this can be re-reviewed? I believe I have addressed all of the (excellent) concerns raised.

The only point on which I am less confident we are aligned is the discussion on raising an error vs. not: I have made my take clear enough in the current state of the code but am happy to hear your thoughts.

trevorkarn:
I agree @joel-dodge that you made all of these changes as well as added some helpful comments. This gets a positive review from me. @ggsmith could you set this to positive review? Thank you @joel-dodge  for your contributions on this!

trevorkarn (approved):
Approved

joel-dodge:
@d-torrance I still see "requested change" since I think @ggsmith requested some changes. Not sure the protocols here but just to let you know this is now in at least partial approved state and maybe can be merged?

Issue #4442, opened 2026-06-13
realpath rejects `#`


pzinn:
This section of hypertext.m2 is problematic:
```
fileExists' = pth -> (
    if match("#",pth) then pth = substring(0,lastMatch#0#0,pth);
    fileExists pth
)

-- TODO: phase this one out eventually
toURL = method()
toURL String := pth -> (
     urlEncode if isAbsolutePath pth then concatenate(rootURI,
	  if fileExists' pth then realpath pth
```
Note that `fileExists` has been modified to allow extra `#` information after the file name, which is typically used by `FilePosition` and error messages.
There's a first basic issue which is, what if a filename contains a `#`?
In any case, a more fundamental problem is that `pth` is fed directly to `realpath` without removing the `#` part.
On my laptop, `realpath` rejects filenames with a `#` -- more precisely, it sends EONENT (i.e., file not found, as it should), though that signal is not correctly forwarded to M2, because stat(NULL) replaces ENOENT with EFAULT (“Bad address”), resulting in the cryptic error message
```
i18 : toURL "/home/pzinn/M2/M2/Macaulay2/m2/hypertext.m2#L451:C26-L453:C15"
stdio:18:5:(3):[1]: error: realpath failed: Bad address
```
I've only started encountering this problem recently (in fact, on the Macaulay2Web server, realpath seems happy with `#`), so not sure what exactly changed. but there's definitely an issue.

pzinn:
the fact that the code is marked `-- TODO: phase this one out eventually` suggests one shouldn't work too hard on this -- but at the moment it is still being called, so something needs to be done.

PR #4443, opened 2026-06-13
Fix realpath handling for URL fragments

pzinn:
This fixes two bugs:
- one is #4442 -- now `toURL` correctly splits the `#` part before feeding it to `realpath`
- the other one, also mentioned in #4442, is that the error code of `realpath` wasn't correctly sent back to M2.
- I've added a couple of tests as well -- I hope they're not too specific, if they fail I can relax/remove them.


pzinn:
this PR is ready

PR #4444, opened 2026-06-13
Record stdio source independently of readline history

### LINKED ISSUE #4356: "line number not found in file stdio" when running "code"

pzinn:
This is my solution to #4356: simply stop using readline and use our own, more reliable, source code for stdio.
Simpler, less fragile.

### DIFF
--- a/M2/Macaulay2/d/M2lib.c
+++ b/M2/Macaulay2/d/M2lib.c
@@ -136,14 +136,6 @@ int system_appendHistory(int n, char *filename)
   return append_history(n, filename);
 }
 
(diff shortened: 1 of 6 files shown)


pzinn:
same, PR ready, needs approving

PR #4450, opened 2026-06-17
Use numerator when calling "roots" on element of fraction field

d-torrance:
This used to segfault.  Also add some unit tests.

## Before

```m2
i1 : R = QQ[x];

i2 : roots((x - 1)/x)
-- SIGSEGV
-* stack trace, pid: 103461
 0# profiler_stacktrace(std::ostream&, int) at /usr/src/macaulay2-1.26.06+git202606141309-0ppa202606152342~ubuntu24.04.1/M2/Macaulay2/d/main.cpp:128
 1# segv_handler at /usr/src/macaulay2-1.26.06+git202606141309-0ppa202606152342~ubuntu24.04.1/M2/Macaulay2/d/main.cpp:244
 2# 0x000079654F245330 in /lib/x86_64-linux-gnu/libc.so.6
 3# rawRoots at interface/polyroots.cpp:30
 4# interface_rawRoots at /usr/src/macaulay2-1.26.06+git202606141309-0ppa202606152342~ubuntu24.04.1/M2/Macaulay2/d/interface.dd:1517
 5# evaluate_evalraw at /usr/src/macaulay2-1.26.06+git202606141309-0ppa202606152342~ubuntu24.04.1/M2/Macaulay2/d/evaluate.d:1662
 6# evaluate_evalSequence at /usr/src/macaulay2-1.26.06+git202606141309-0ppa202606152342~ubuntu24.04.1/M2/Macaulay2/d/evaluate.d:457
...
```

## After

```m2
i1 : R = QQ[x];

i2 : roots((x - 1)/x)

o2 = {1}

o2 : List
```

## AI Disclosure

Claude helped with the C++ code

### DIFF
--- a/M2/Macaulay2/e/interface/polyroots.cpp
+++ b/M2/Macaulay2/e/interface/polyroots.cpp
@@ -32,11 +32,11 @@ engine_RawRingElementArrayOrNull rawRoots(const RingElement *p,
     p = RingElement::make_raw(R, frac->numerator(p->get_value()));
   }
   const PolynomialRing *P = R->cast_to_PolynomialRing();
-  const Monoid *M = P->getMonoid();
   if (P == nullptr) {
     ERROR("expected a polynomial ring");
     return nullptr;
   }
+  const Monoid *M = P->getMonoid();
   if (P->n_vars() != 1) {
     ERROR("expected a univariate polynomial ring");
     return nullptr;
--- a/M2/Macaulay2/e/interface/polyroots.cpp
+++ b/M2/Macaulay2/e/interface/polyroots.cpp
@@ -12,6 +12,7 @@
 #include "basic-rings/aring.hpp"
 #include "error.h"
 #include "monoid.hpp"
+#include "rings/frac.hpp"
 #include "rings/polyring.hpp"
 #include "ring-elements/ring-element.hpp"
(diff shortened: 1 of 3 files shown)


andrew-tawfeek (commented):
Two comments on this, after investigating how it touches other files:

- In `polyroots.cpp`, could we move `P->getMonoid()` after the `P == nullptr` check? Non-polynomial inputs can still segfault before the error message.
- In `roots.m2`, maybe its worth making this a tolerance-based check instead of an exact equality (referring to `assert Equation(roots f_(frac R), Z)`) between two numerical root lists to avoid possible ordering/rounding errors.

d-torrance:
Thanks for the review!  I believe I've addressed both comments:

* The check that `P` is nonull has been added in a new commit since it's not directly related to the other commit.
* I amended the original commit updating the test file with the recommended changes.

andrew-tawfeek (approved):

jkyang92 (commented):
If we are going to support rational functions, the documentation should probably say so. No issues with the code itself

d-torrance:
> If we are going to support rational functions, the documentation should probably say so. No issues with the code itself

Good point!  Just pushed a new commits updating the `roots` docs

jkyang92 (approved):

PR #4458, opened 2026-06-24
Added package CotangentCohomology for computing (higher) cotangent cohomology

nilten:
I added a new package "CotangentCohomology" which does two things:
- it computes cotangent cohomology of a ring S with coefficients in an S-module M by considering the derivation complex coming from a Koszul-Tate resolution of S
- it creates so-called arborescent resolutions, a special kind of Koszul-Tate resolution

### DIFF
--- a/M2/Macaulay2/packages/CotangentCohomology.m2
+++ b/M2/Macaulay2/packages/CotangentCohomology.m2
@@ -1,0 +1,459 @@
+--*- coding: utf-8 -*-
+---------------------------------------------------------------------------
+-- PURPOSE: Calculating arborescent resolutions and higher cotangent cohomology
+-- PROGRAMMER : Nathan Ilten
+-- UPDATE HISTORY : August 2025
+---------------------------------------------------------------------------
+newPackage("CotangentCohomology",
+    Headline => "arborescent resolutions and cotangent cohomology",
+    Version => "1.0",
+    Date => "June 18, 2026",
+    Authors => {
+        {Name => "Nathan Ilten",
+	  HomePage => "http://www.sfu.ca/~nilten/",
+	  Email => "nilten@sfu.ca"}},
+    HomePage => "https://github.com/Macaulay2/M2",
+    PackageExports => {"DGAlgebras"},
+    Keywords => {"Deformation Theory","Homological Algebra"},
+    )
+
+---------------------------------------------------------------------------
+-- COPYRIGHT NOTICE:
+--
+-- Copyright 2026 Nathan Owen Ilten
+--
+--
+-- This program is free software: you can redistribute it and/or modify
+-- it under the terms of the GNU General Public License as published by
+-- the Free Software Foundation, either version 3 of the License, or
+-- (at your option) any later version.
+--
+-- This program is distributed in the hope that it will be useful,
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
+-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+-- GNU General Public License for more details.
+--
+-- You should have received a copy of the GNU General Public License
+-- along with this program.  If not, see <http://www.gnu.org/licenses/>.
+--
+---------------------------------------------------------------------------
+
+export {"ArborescentResolution",
+    "arborescentResolution",
+    "liftArborescentResolution",
+    "dualCotangent",
+    "cotangentCohomology",
+    "TT",
+    "Arborescent"
+    }
+
+
+
+---------------------------------------------------------
+-- Code  for creating Arborescent resolutions
+---------------------------------------------------------
+ArborescentResolution= new Type of DGAlgebra
+
+--Create first 3 steps of arborescent resolution from free resolution
+setupArborescentResolution:=F->(
+    K1:=setDiff(freeDGAlgebra(ring F,apply(degrees F_1,i->{1}|i)),flatten entries F.dd_1,InitializeComplex=>false,InitializeDegreeZeroHomology=>false);
+    if F.dd_2==0 then (
+	K1.cache#"resolution"=F;
+	return new ArborescentResolution from K1);
+    K2:=adjoinVariables(K1,flatten entries(getBasis(1,K1)*F.dd_2));
+    rho:=((toComplex(K1,2)).dd_2)//(F.dd_2);
+    K3:=new ArborescentResolution from adjoinVariables(K2,
+	(flatten entries (-((vars K2.natural)_(positions(K2.Degrees,i->i_0==2))*rho)+sub(getBasis(2,K1),K2.natural)))|(
+	    flatten entries ((vars K2.natural)_(toList (numgens F_1..numgens F_1+numgens F_2-1))*F.dd_3)));
+    K3.cache#"resolution"=F;
+    K3.cache#"rho"=rho;
+    K3.cache#"P''"=getBasis(2,K1);
+    K3.cache#"P'"=getBasis(3,K2);
+    K3)
+
+
+--Extend arborescent resolution one step further
+liftArborescentResolution=method(TypicalValue=>ArborescentResolution)
+liftArborescentResolution ArborescentResolution:=K->(
+    F:=K.cache#"resolution";
+    if F.dd_2==0 then return K;
+    i:=1+max apply(K.Degrees,i->i_0); -- 1+ highest cohom degree of generators; we'll add gens in this degree
+    P:=getBasis(i-1,K); -- basis for P_(i-1)
+    P':=K.cache#"P'"; -- basis for P_(i-1)^old=P'_i
+    P'':=K.cache#"P''"; -- basis for P_(i-2)^old=P'_(i-1)
+    p:=numgens source P;
+    p':=numgens source P';
+    p'':=numgens source P'';
+    r':=numgens F_(i-2);
+    r:=numgens F_(i-1);
+    --let's figure out gamma_(i-1) and alpha_(i-1)
+    d:=(toComplex(K,i-1)).dd_(i-1);
+    n:=numgens target d;
+    gamma:=(d_(toList (0..p'-1))^(toList (0..p''-1))); --gamma_(i-1) as matrix
+    alpha:=(d_(toList (0..p'-1))^(toList (n-r'..n-1))); --alpha_(i-1) as matrix
+    oldrho:=K.cache#"rho"; --rho_(i-2) as matrix
+    newrho:=(oldrho*gamma+alpha)//(F.dd_(i-1)); --rho_(i-1) as matrix
+    --change matrices to lists
+   gammaL:=sub(P''*gamma,K.natural);
+   --alphaL:=(matrix{((gens(K.natural)))_(toList(-r..-1))})*alpha;
+   newdiff:=flatten entries ((P*(id_(ZZ^(p'))||(-gamma)||-newrho))|(P_(toList (p-r..p-1))*(F.dd_i)));
+   newK:=new ArborescentResolution from adjoinVariables(K,newdiff);
+   newK.cache#"resolution"=F;
+   newK.cache#"rho"=newrho;
+   newK.cache#"P''"=P';
+   newK.cache#"P'"=getBasis(i,K);
+   if K.cache.?Ideal then newK.cache.Ideal=K.cache.Ideal; 
+   newK)
+
+
+--create arborescent resolution of length n from free resolution F
+arborescentResolution=method(TypicalValue=>ArborescentResolution,Options=>{Verbose=>0})
+
+arborescentResolution (ChainComplex,ZZ) :=
+arborescentResolution (Complex,ZZ):=opts->(F,n)->(
+    K:=setupArborescentResolution F;
+    i:=3;
+    while (i<n) do (
+	if opts#Verbose>0 then print ("Lifting to degree "|toString(i+1));
+	K=liftArborescentResolution K;
+	i=i+1);
+    K)
+
+arborescentResolution (Ideal,ZZ):=opts->(I,n)->(
+    local Q;
+    if I.cache.?ArborescentResolution then Q=I.cache.ArborescentResolution else Q=res I;
+    K:=arborescentResolution(Q,n,opts);
+    I.cache.ArborescentResolution=K;
+    K)
+
+arborescentResolution (ArborescentResolution,ZZ):=opts->(K,n)->(
+    i:=max apply(K.Degrees,i->i_0);
+    while (i<n) do (
+	if opts#Verbose>0 then print ("Lifting to degree "|toString(i+1));
+	K=liftArborescentResolution K;
+	i=i+1);
+    K)
+
+------------------------------------------------------------------
+-- Code for cotangent cohomology
+------------------------------------------------------------------
+
+--replaces any polynomial with the terms of degree at most 1
+degreeOnePart:=x->(
+    C:=coefficients x;
+    pL:=positions(flatten entries (coefficients x)_0,i->member(i,{1_(ring x)}|gens ring x));
+    (((C_0)_pL)*(C_1)^pL)_(0,0))
+
+
+--creates complex from which cotangent cohomology is computed out to degree n
+dualCotangent=method(TypicalValue=>Complex,Options=>{Verbose=>0,Arborescent=>true})
+dualCotangent (DGAlgebra,ZZ):=opts->(K,n)->(
+    dL:=K.Degrees;
+    if n>max apply(dL,i->i_0) then (
+    	if opts#Verbose>0 then print "Warning: degree bound exceeds largest homological degree of generators.";
+	n=max apply(dL,i->i_0));
+    dL2:=-1*apply(dL,i->drop(i,1));
+    vL:=gens K.natural; 
+    gL:={0}|apply(n,i->#positions(dL,j->j_0<=i+1)); 
+    R:=K.ring;
+    modL:=apply(n,i->R^(dL2_(toList(gL_i..gL_(i+1)-1))));
+    if opts#Verbose>0 then print "Reducing differential";
+    d:=apply((flatten entries matrix K.diff)_(toList(0..gL_n-1)),x->degreeOnePart x);
+    local iLa;
+    local iLb;
+    if opts#Verbose>0 then print "Computing differential in terms of basis";
+    mapL:=apply(n-1,i->(
+	    iLa=toList(gL_i..gL_(i+1)-1);
+	    iLb=toList(gL_(i+1)..gL_(i+2)-1);
+	    matrix {d_iLb}//matrix {vL_iLa}));
+    jac:=jacobian sub(matrix {d_(toList(gL_0..gL_1-1))},R);
+    firstMap:=dual map(dual modL_0,,dual jac);
+    dual (chainComplex ({firstMap}|(apply(n-1,i->map(modL_i,modL_(i+1),sub(mapL_i,R))))))
+    )
+
+dualCotangent (Ideal,ZZ):=opts->(I,n)->(
+    local K;
+    if opts#Arborescent then (
+	if opts#Verbose>0 then print "Calculating arborescent resolution";
+	K=arborescentResolution(I,n,Verbose=>opts#Verbose));
+    if not opts#Arborescent then (
+	if opts#Verbose>0 then print "Calculating acyclic closure";
+	K=acyclicClosure(koszulComplexDGA I,EndDegree=>(n-1)));
+    dualCotangent(K,n,opts)
+    )
+
+--computes cotangent cohomology modules
+cotangentCohomology=method(Options=>{Verbose=>0,Arborescent=>true})
+cotangentCohomology (DGAlgebra,Module,ZZ) :=
+cotangentCohomology (Ideal,Module,ZZ):=opts->(I,M,i)->(
+    HH^i(dualCotangent(I,i+1,opts)**M))
+
+cotangentCohomology (DGAlgebra,ZZ):=opts->(K,i)->(cotangentCohomology(K,comodule ideal zerothHomology K,i,opts))
+cotangentCohomology (Ideal,ZZ):=opts->(I,i)->(cotangentCohomology(I,comodule I,i,opts))
+
+    
+
+TT = new ScriptedFunctor from {
+     superscript => (
+	  j -> new ScriptedFunctor from {
+	       argument => (
+		    X -> cotangentCohomology((sequence X)|(sequence j))
+		    )
+	       }
+	  ),
+   argument => (
+	  X -> cotangentCohomology(sequence X)
+	  )
+     }
+
+
+beginDocumentation()
+document {
+     Key => CotangentCohomology,
+     Headline => "arborescent resolutions and cotangent cohomology",
+     PARA{
+     "This package is used to construct arborescent resolutions and compute cotangent cohomology groups.
+      An arborescent resolution of an algebra R is a special kind of free differential graded algebra
+      resolving R. See"},
+      UL {{"Hancharuk, Laurent-Gengoux, and Strobl, Koszul-Tate resolutions and decorated trees, 
+      2024, arXiv:2406.03955"}},
+     PARA{"For mathematical background on how arborescent resolutions are computed here, see"},
+     UL {{"Ilten, Meazzini, and Petracci, Higher cotangent cohomology for Stanley-Reisner Rings, 2026, arXiv:2606.16829"}},
+     PARA{"Cotangent cohomology for an algebra R is computed (in characteristic zero) via a complex obtained from a free DG-algebra resolution, either
+     an arborescent resolution or one obtained via acyclic closure."},
+     PARA{"The author thanks Francesco Meazzini, Frank Moore, and Andrea Petracci for helpful conversations and suggestions."}
+ }
+
+
+ 
+
+
+
+doc ///
+  Key
+    arborescentResolution
+    (arborescentResolution,ChainComplex,ZZ)
+    (arborescentResolution,Complex,ZZ)
+    (arborescentResolution,Ideal,ZZ)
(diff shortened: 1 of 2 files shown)


M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+A=arborescentResolution(I,6)
+assert (numgens source basis TT^1(I)==1 and TT^17(I)==0)
+///
+
+TEST ///
+R=QQ[x,y,z,Degrees=>entries id_(ZZ^3)]
+I=ideal {x*y*z}
+assert (2==numgens source basis({1,1,0},TT^0(I)))
+///
+
```
ggsmith, line 490: There are currently no direct tests for `liftArborescentResolution`, and `dualCotangent`. To ensure future functionality, this should be changed.

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+I=ideal {x*y}
+A=arborescentResolution(I,6)
+assert (numgens source basis TT^1(I)==1 and TT^17(I)==0)
+///
+
+TEST ///
+R=QQ[x,y,z,Degrees=>entries id_(ZZ^3)]
+I=ideal {x*y*z}
+assert (2==numgens source basis({1,1,0},TT^0(I)))
+///
```
ggsmith, line 489: All the tests involve only monomial ideals. If this reflects the scope of the package, then the documentation needs to be updated. Otherwise, it seems that at least one test should involve an ideal not generated by monomials.

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+
+
+TEST ///
+R=QQ[x,y,z]
+I=ideal {x*y,x*z,y*z}
+assert ((3,0,1,3,6)==apply(1..5,i->numgens source basis TT^i(I)))
+assert ((3,0,1,3,6)==apply(1..5,i->numgens source basis TT^i(I,Arborescent=>false)))
+///
+
```
ggsmith, line 465: The `TT` method does not seem to handle the zero ideal correctly.  For example, we have

```
i3 : S = QQ[x,y,z];

i4 : TT^1 ideal 0_S
stdio:4:4:(3):[1]: error: no method for binary operator | applied to objects:
            {0} (of class List)
      |     <<a list>> (of class InfiniteNumber)
```

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+    K1:=setDiff(freeDGAlgebra(ring F,apply(degrees F_1,i->{1}|i)),flatten entries F.dd_1,InitializeComplex=>false,InitializeDegreeZeroHomology=>false);
+    if F.dd_2==0 then (
+	K1.cache#"resolution"=F;
+	return new ArborescentResolution from K1);
+    K2:=adjoinVariables(K1,flatten entries(getBasis(1,K1)*F.dd_2));
+    rho:=((toComplex(K1,2)).dd_2)//(F.dd_2);
+    K3:=new ArborescentResolution from adjoinVariables(K2,
+	(flatten entries (-((vars K2.natural)_(positions(K2.Degrees,i->i_0==2))*rho)+sub(getBasis(2,K1),K2.natural)))|(
+	    flatten entries ((vars K2.natural)_(toList (numgens F_1..numgens F_1+numgens F_2-1))*F.dd_3)));
+    K3.cache#"resolution"=F;
```
ggsmith: Wouldn't it be better to use a `Symbol' rather than a `String' as the keys in the cache table?

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+    K3.cache#"rho"=rho;
+    K3.cache#"P''"=getBasis(2,K1);
+    K3.cache#"P'"=getBasis(3,K2);
+    K3)
```
ggsmith, line 77: I fear that `P''` and `P'` are not informative enough to be used as keys in the cache table. I recommend using a protected `Symbol'.

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+    }
+
+
+
+---------------------------------------------------------
+-- Code  for creating Arborescent resolutions
+---------------------------------------------------------
+ArborescentResolution= new Type of DGAlgebra
+
+--Create first 3 steps of arborescent resolution from free resolution
```
ggsmith, line 61: Even if this local function is not exported, I would be tempted to make it a `method`. This syntax force one to specify the input and output types—easier for humans to read the code?

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+setupArborescentResolution:=F->(
+    K1:=setDiff(freeDGAlgebra(ring F,apply(degrees F_1,i->{1}|i)),flatten entries F.dd_1,InitializeComplex=>false,InitializeDegreeZeroHomology=>false);
```
ggsmith: I would like to see a space before and after the operator `:=`.  Makes the code easier to read.

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+    K3.cache#"resolution"=F;
+    K3.cache#"rho"=rho;
+    K3.cache#"P''"=getBasis(2,K1);
+    K3.cache#"P'"=getBasis(3,K2);
+    K3)
+
+
+--Extend arborescent resolution one step further
+liftArborescentResolution=method(TypicalValue=>ArborescentResolution)
+liftArborescentResolution ArborescentResolution:=K->(
```
ggsmith: Longer names for local variables would make it easier to read the code.

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+   newK.cache#"P''"=P';
+   newK.cache#"P'"=getBasis(i,K);
+   if K.cache.?Ideal then newK.cache.Ideal=K.cache.Ideal; 
+   newK)
+
+
+--create arborescent resolution of length n from free resolution F
+arborescentResolution=method(TypicalValue=>ArborescentResolution,Options=>{Verbose=>0})
+
+arborescentResolution (ChainComplex,ZZ) :=
```
ggsmith: In the very near future, the `ChainComplex` type will disappear.

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+arborescentResolution (Complex,ZZ):=opts->(F,n)->(
+    K:=setupArborescentResolution F;
+    i:=3;
+    while (i<n) do (
+	if opts#Verbose>0 then print ("Lifting to degree "|toString(i+1));
```
ggsmith: Should it be `opts.Verbose`.

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+	K=liftArborescentResolution K;
+	i=i+1);
+    K)
```
ggsmith, line 118: The white space/indentation seems unconventional, particularly in the while loop.

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+
+arborescentResolution (Ideal,ZZ):=opts->(I,n)->(
```
ggsmith: Better to use `??=` to handle storing the result in the cache table.

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+    P:=getBasis(i-1,K); -- basis for P_(i-1)
+    P':=K.cache#"P'"; -- basis for P_(i-1)^old=P'_i
+    P'':=K.cache#"P''"; -- basis for P_(i-2)^old=P'_(i-1)
+    p:=numgens source P;
+    p':=numgens source P';
+    p'':=numgens source P'';
+    r':=numgens F_(i-2);
+    r:=numgens F_(i-1);
+    --let's figure out gamma_(i-1) and alpha_(i-1)
+    d:=(toComplex(K,i-1)).dd_(i-1);
```
ggsmith: Would using `polyDifferential` be more efficient?  In particular, this methods appear to return just the relevant map rather than the entire complex.

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+    K)
+
+------------------------------------------------------------------
+-- Code for cotangent cohomology
+------------------------------------------------------------------
+
+--replaces any polynomial with the terms of degree at most 1
+degreeOnePart:=x->(
+    C:=coefficients x;
+    pL:=positions(flatten entries (coefficients x)_0,i->member(i,{1_(ring x)}|gens ring x));
```
ggsmith: Replace `(coefficients x)_0` with `C_0`.

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+	if opts#Verbose>0 then print ("Lifting to degree "|toString(i+1));
+	K=liftArborescentResolution K;
+	i=i+1);
+    K)
+
+------------------------------------------------------------------
+-- Code for cotangent cohomology
+------------------------------------------------------------------
+
+--replaces any polynomial with the terms of degree at most 1
```
ggsmith, line 140: Is `x` an odd name for a polynomial?  How about 'f' or 'p'?

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+    local iLa;
+    local iLb;
+    if opts#Verbose>0 then print "Computing differential in terms of basis";
+    mapL:=apply(n-1,i->(
+	    iLa=toList(gL_i..gL_(i+1)-1);
+	    iLb=toList(gL_(i+1)..gL_(i+2)-1);
+	    matrix {d_iLb}//matrix {vL_iLa}));
+    jac:=jacobian sub(matrix {d_(toList(gL_0..gL_1-1))},R);
+    firstMap:=dual map(dual modL_0,,dual jac);
+    dual (chainComplex ({firstMap}|(apply(n-1,i->map(modL_i,modL_(i+1),sub(mapL_i,R))))))
```
ggsmith: Should `chainComplex` be replaced by `complex'?

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+	    flatten entries ((vars K2.natural)_(toList (numgens F_1..numgens F_1+numgens F_2-1))*F.dd_3)));
+    K3.cache#"resolution"=F;
+    K3.cache#"rho"=rho;
+    K3.cache#"P''"=getBasis(2,K1);
+    K3.cache#"P'"=getBasis(3,K2);
+    K3)
+
+
+--Extend arborescent resolution one step further
+liftArborescentResolution=method(TypicalValue=>ArborescentResolution)
```
ggsmith: I recommend not using `TypicalValue` because it sidesteps some useful error checking. For each assignment, I suggest that one indicate the actual output type.

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+degreeOnePart:=x->(
+    C:=coefficients x;
+    pL:=positions(flatten entries (coefficients x)_0,i->member(i,{1_(ring x)}|gens ring x));
+    (((C_0)_pL)*(C_1)^pL)_(0,0))
+
+
+--creates complex from which cotangent cohomology is computed out to degree n
+dualCotangent=method(TypicalValue=>Complex,Options=>{Verbose=>0,Arborescent=>true})
+dualCotangent (DGAlgebra,ZZ):=opts->(K,n)->(
+    dL:=K.Degrees;
```
ggsmith: What happens when `K.Degrees` is the empty list? I think this arising when dealing with the zero ideal.

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+    I:Ideal
+    F:Complex 
+    F:ChainComplex
+    n:ZZ
+    Verbose=>ZZ
+  Outputs
+    A:ArborescentResolution
+  Description
+    Text
+	Inductively constructs a truncated arborescent resolution up to homological degree @ TT "n" @. 	For @ TT "I" @ an ideal in a
```
ggsmith: Strange amount of white space between "." and "For".

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+	polynomial ring @ TT "R"@, @ TT "arborescentResolution(I,n)"@ constructs an arborescent resolution of @ TT "R/I" @. The resulting
+	arborescent resolution is cached in @ TT "I.cache.ArborescentResolution" @.
+	For more control, one may specify the underlying free resolution @ TT "F" @ of @ TT "R/I"@ by using @ TT "arborescentResolution(F,n)"@.
```
ggsmith, line 257: Traditionally, one caches the output of function`F` in `foo.cache.F`.  In particular, I would replace `ArborescentResolution` with `arborescentResolution`.

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+    "Arborescent"
+    }
+
+
+
+---------------------------------------------------------
+-- Code  for creating Arborescent resolutions
+---------------------------------------------------------
+ArborescentResolution= new Type of DGAlgebra
+
```
ggsmith, line 60: Does this new type need associated `net` function? The output inherited from DGAlgebras isn't particularly nice.

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+   Text
+	Given an existing truncated arborescent resolution @ TT "B" @, @ TT "arborescentResolution(B,n)" @ extends it to homological
+	degree @ TT "n" @.
+   Text
+       The optional input @ TT "Verbose" @ controls the level of verbosity during the computation.
+
+   Example
+      R=QQ[x_1..x_5]
+      I=ideal {x_1*x_2,x_2*x_3,x_3*x_4,x_4*x_5,x_5*x_1}
+      A=arborescentResolution(I,3)
```
ggsmith, line 265: Add `SeeAlso`?

ggsmith, line 265: Can one include an example that illustrates the `Verbose` option?

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+  Headline
+    lifts a truncated arborescent resolution to one degree higher
+  Usage
+    A = liftArborescentResolution(B)
+  Inputs
+    B:ArborescentResolution 
+  Outputs
+    A:ArborescentResolution
+  Description
+   Example
```
ggsmith, line 294: Needs a description. Should probably emphasize that this method is unlikely to be used directly by a user.

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+      R=QQ[x_1..x_5]
+      I=ideal {x_1*x_2,x_2*x_3,x_3*x_4,x_4*x_5,x_5*x_1}
+      B=arborescentResolution(I,3)
+      liftArborescentResolution(B)
```
ggsmith, line 298: SeeAlso should definitely include `arborescentResolution'.

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+
+doc ///
+  Key
+    ArborescentResolution
+  Headline
+    The class of all arborescent resolutions
+  Description
+    Text
+      An @ TT "ArborescentResolution" @ @ TT "A"@  is a special class of @ TO DGAlgebra @ encoding a truncation of an arborescent resolution.
+      The hash table @ TT "A.cache" @ contains extra information that allows one to lift the truncation to higher degree via @ TO liftArborescentResolution @.
```
ggsmith, line 313: Somewhere in the package one should actually define an arborescent resolution. This might be the appropriate place.

M2/Macaulay2/packages/CotangentCohomology.m2
ggsmith, line 60: I recommend adding/overloading `isWellDefined` for this type. In my experience, this is the best way to document all of the features/properties of an object having this type.

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+      prune HH_1(C)
+      prune HH_2(C)
+///
+
+
+
+doc ///
+  Key
+    Arborescent
+  Headline
```
ggsmith, line 334: I would be tempted to eliminate this separate documentation page. Instead, I would just include the relevant information on the documentation pages for `dualCotangent` and `cotangentCohomology`.  In particular, I would add the key `Arborescent` to one of these.

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+    Arborescent=>Boolean
+  Outputs
+    C:ChainComplex
+  Description
+    Text
+        Given an ideal @ TT "I"@ of a polynomial ring @ TT "R"@, @ TT "dualCotangent(I,n)"@ constructs a sequence @ TT "C"@ of free @TT "R"@ modules
+	of length @ TT "n" @ such that after tensoring with an @ TT "R/I"@-module @ TT "M"@, the cohomology of the resulting complex up to degree @ TT "n-1" @ is
+	the cotangent cohomology of @ TT "R/I" @ with coefficients in @ TT "M" @. If @ TT "Arborescent" @ is true, as is the case by
+	default, this sequence is obtained from an arborescent resolution; if @ TT "Arborescent" @ is false, then an acyclic closure
+	is used instead.
```
ggsmith: What is an "acyclic closure"? Ideally, one would include the relevant information. At a minimum, one should provide a citation.

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+    Text
+	If @ TT "A" @ is a @ TO DGAlgebra @ resolving @ TT "R/I" @, then @ TT "dualCotangent(A,n)" @ returns a similar sequence,
+	making use of the already computed DG-algebra resolution @TT "A"@.
+    Text
+       The optional input @ TT "Verbose" @ controls the level of verbosity during the computation.
+   Example
+      R=QQ[x_1..x_5]
```
ggsmith: Replace `R=QQ[x_1..x_5]` by `R = QQ[x_1..x_5];`.

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+    Text
+        Given an ideal @ TT "I"@ of a polynomial ring @ TT "R"@, @ TT "dualCotangent(I,n)"@ constructs a sequence @ TT "C"@ of free @TT "R"@ modules
+	of length @ TT "n" @ such that after tensoring with an @ TT "R/I"@-module @ TT "M"@, the cohomology of the resulting complex up to degree @ TT "n-1" @ is
+	the cotangent cohomology of @ TT "R/I" @ with coefficients in @ TT "M" @. If @ TT "Arborescent" @ is true, as is the case by
+	default, this sequence is obtained from an arborescent resolution; if @ TT "Arborescent" @ is false, then an acyclic closure
+	is used instead.
+    Text
+	If @ TT "A" @ is a @ TO DGAlgebra @ resolving @ TT "R/I" @, then @ TT "dualCotangent(A,n)" @ returns a similar sequence,
+	making use of the already computed DG-algebra resolution @TT "A"@.
+    Text
```
ggsmith, line 387: I would move the paragraph about `Verbose` to after the first example.

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+       The optional input @ TT "Verbose" @ controls the level of verbosity during the computation.
+   Example
+      R=QQ[x_1..x_5]
+      I=ideal {x_1*x_2,x_2*x_3,x_3*x_4,x_4*x_5,x_5*x_1}
+      prune TT^3(I)
+      prune TT^3(I,comodule ideal {x_1,x_2*x_3,x_3*x_4,x_4*x_5})
+ Caveat
+       This method is only guaranteed to give correct results if @ TT "R"@ has characteristic zero.
+///
```
ggsmith, line 446: Could be fun to include an example over a finite field which returns an incorrect result?

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+I=ideal {x*y,x*z,y*z}
+assert ((3,0,1,3,6)==apply(1..5,i->numgens source basis TT^i(I)))
+assert ((3,0,1,3,6)==apply(1..5,i->numgens source basis TT^i(I,Arborescent=>false)))
+///
+
+TEST ///
+R=QQ[x,y,z]
+I=ideal {x*y,x*z,y*z}
+assert (2==numgens source basis TT^2(I,comodule ideal {x,y,z}))
+///
```
ggsmith, line 470: Why not combine this Test with the previous one?

M2/Macaulay2/packages/CotangentCohomology.m2
```diff
@@ -0,0 +1,459 @@
...
+
+TEST ///
+R=QQ[x,y,z,Degrees=>entries id_(ZZ^3)]
```
ggsmith: Isn't `R = QQ[x, y, z, DegreeRank => 3];` the same ring?

ggsmith (changes requested):
@aalmousa and I completed this review. Most of our comments relate to small issues in the code, documents, and tests.

M2/Macaulay2/packages/CotangentCohomology.m2
joel-dodge: Just a passerby observation:

If you do add a non-working example, then be wary that there are many cases in docs where the text describes behavior that the code no longer exhibits. I don't know if there is a standardized way to maintain invariants like this but in cases where I want to make sure that the result of executing code actually provides an example of something that the text is describing I have been putting `assert` statements in the example so that if the behavior drifts from what is expected (even if it drifts from incorrectness to correctness) the example generation will fail and either text / example must be updated.

M2/Macaulay2/packages/CotangentCohomology.m2
nilten: Is it? It seems to me that I would need to do:

Q=res I;
Q??=I.cache.ArborescentResolution;

but that seems less efficient.

M2/Macaulay2/packages/CotangentCohomology.m2
nilten: This could be fun, but I don't know any off the top of my head. The underlying issue is that in characteristic p, one needs to use a DG-Algebra resolution using divided powers instead of just adjoining standard variables. If you use standard variables, your algebra agrees with a divided powers one up to at least degree 2p-1, so the cotangent cohomology will be correctly computed up to degree 2p-2.

In lieu of an example, I'll expand this Caveat as described above.

M2/Macaulay2/packages/CotangentCohomology.m2
nilten: I agree that the output isn't particularly nice, but doesn't this seem more like a problem for the DGAlgebras package?

M2/Macaulay2/packages/CotangentCohomology.m2
kellerlv: > This could be fun, but I don't know any off the top of my head. The underlying issue is that in characteristic p, one needs to use a DG-Algebra resolution using divided powers instead of just adjoining standard variables. If you use standard variables, your algebra agrees with a divided powers one up to at least degree 2p-1, so the cotangent cohomology will be correctly computed up to degree 2p-2.
> 
> In lieu of an example, I'll expand this Caveat as described above.

I think there's a more fundamental issue here, no? A choice of minimal model will still compute the dg cotangent complex in whatever characteristic you'd like, but the more fundamental issue is that the dg cotangent complex does not compute Andre--Quillen (co)homology in positive characteristic since simplicial algebras and dg algebras are not equivalent in positive characteristic.

nilten: I agree that simplicial algebras and dg algebras are not equivalent in positive characteristic. I had thought that you could nonetheless still use a Tate resolution (with divided powers) to compute Andre--Quillen cohomology. But I can't find a reference for this, so probably I was mistaken. @kellerlv do you have a concrete example in characteristic p where a Tate resolution doesn't compute the Andre--Quillen cohomology?

nilten:
Thanks for the review. I just committed changes addressing the comments.

Issue #58, opened 2013-10-28
quotient of a Weyl algebra by a left ideal


DanGrayson:
Reported by:    anton (on trac)

Forming a QuotientRing as in the example below should produce an error (for noncommutative algebras)... as this makes little sense. 

```
i1 : W = QQ[x,dx,WeylAlgebra=>{x=>dx}]

o1 = W

o1 : PolynomialRing

i2 : A := W/ideal dx

  W
o2 = --                                                                         
 dx

o2 : QuotientRing

i3 : dx_A*x_A == (dx*x)_A

o3 = false
```

An alternative behaviour would be to return the module W^1/I.

PR #4459, opened 2026-06-25
Compute Hom and Ext over SkewCommutative rings

joel-dodge:
NOT READY FOR REVIEW AT ALL

This is just a draft for now as it also includes changes from https://github.com/Macaulay2/M2/pull/4435 and https://github.com/Macaulay2/M2/pull/4176 that will hopefully be reviewed and merged soon.


joel-dodge:
Fascinating there is a failure building examples from the BernsteinSato package

```
	     W = QQ[x, D, WeylAlgebra=>{x=>D}]
	     M = W^1/ideal(D-1)
	     N = W^1/ideal((D-1)^2)
	     DHom(M,N)
```

That boils down to an issue computing Hom over a non-commutative ring - a check that is added in this PR. I added this based on my understanding that the methods for computing R-module Homs without the use of pushFwd all produce an R-module as the output and that this cannot be correct as these things are not in fact R-modules. I need to dig into Weyl algebras a bit to understand better what *should* be happening and may relax the invariant in this case if appropriate.

joel-dodge:
Fascinating. The DHom algorithm indeed relies on computing Hom for modules over the Weyl algebra W but IIUC only does so in the case of computing a dual of a free resoltuion in which case the source of the Hom is free. ~Of course Hom(M, N) IS a left W-module when N is a bimodule so this works fine.~ Hom(M, N) has a left R-module structure exctly when M is also a right R-module so in these cases the result actually is a bimodule and I think the existing Hom impl gives the right result!

Unfortunately, It seems that querying whether a given module is a bimodule or not seems to be not well supported in the Core M2 offerings - maybe NCAlgebras does this better... but Weyl algebras / skew symmetric algebras are not part of that package. 

I found this old issue that is tangentially related: https://github.com/Macaulay2/M2/issues/58

While this is still WIP I am going to change the invariant check that forbids using bare Hom for non-commutative rings with one that instead checks that the target is free in the non-commutative case. I will expect discussion on this point.

Issue #3973, opened 2025-09-06
Internal error inverting element of tower ring constructed with toField


d-torrance:
Reported by Nick in the Google group (https://groups.google.com/g/macaulay2/c/8VY3X-gvkVE/m/-bhwUZ9wAAAJ):

```m2
i1 : F = toField(QQ[i]/(i^2 + 1));

i2 : K = toField(F[r]/(r^4 - 2));

i3 : r^-1
--internal error: ring element gcd computation failed
Aborted (core dumped)
```

d-torrance:
> **Written by Claude** (Claude Opus 5, via Claude Code), not by @d-torrance, whose account posted it -- please weigh it accordingly.

The same entry point fails in the zero ring, and there the element being refused really is invertible.

```m2
i1 : R = QQ[x]/1;

i2 : 1_R == 0_R

o2 = true

i3 : isUnit x

o3 = true

i4 : x * (1//x) == 1

o4 = true

i5 : x^0

o5 = 0

i6 : x^-1
stdio:6:2:(3): error: either element not invertible, or no method available to compute its inverse

i7 : 1_R^-1
stdio:7:5:(3): error: either element not invertible, or no method available to compute its inverse
```

The zero ring is correct to call `x` a unit: `1 = 0` there, so `0 * 0 = 0 = 1`. The inverse being
withheld is `0`, which `1//x` returns without complaint. Note `i7` — the ring cannot invert its own
identity element.

Four parts of M2 already agree, including the matrix path:

```m2
i8 : inverse matrix {{x}}

o8 = map(R^1, R^1, 0)
```

So `inverse` on a 1×1 matrix over `R` gives the right answer while `x^-1` on its only entry does not.

### Why it is the same defect as this issue

`Ring::power`, on a negative exponent, calls `invert` and then uses `is_zero` on the result as its
signal that `invert` failed:

https://github.com/Macaulay2/M2/blob/development/M2/Macaulay2/e/rings/ring.cpp#L127-L139

In the zero ring a correct inverse *is* zero, so the sentinel cannot tell success from failure. This
issue's `r^-1` reaches the same negative-exponent path and fails a little earlier, inside `invert`
itself; the shared route is `Ring::power`, which is why this is being added here rather than filed
separately.

That test was introduced by
[`d97f303f2d`](https://github.com/Macaulay2/M2/commit/d97f303f2d) (2006-09-23, "fixed: inverses,
[...]

PR #4462, opened 2026-07-02
"fraction" fixes

### LINKED ISSUE #4461: error multiplying matrices with entries in fraction field
am I the only one encountering such issues?...
```
i1 : F=frac(QQ[y])

o1 = F

o1 : FractionField

i2 : a=matrix{{1/2}}*matrix{{y}}

o2 = | 1/2y |

             1      1
o2 : Matrix F  <-- F

i3 : b=matrix{{y/2}}

o3 = | y/2 |

             1      1
o3 : Matrix F  <-- F

i4 : a==b

o4 = false

i5 : a_(0,0)===b_(0,0)

o5 = false

i6 : ring a===ring b

o6 = true

i7 : source a===source b

o7 = true

i8 : target a===target b

o8 = true

i9 : a-b==0

o9 = true
```

pzinn:
This fixes two somewhat related bugs in the handling of fractions:
- the first one is #4461: I fixed it the obvious way, by making sure there is a canonical form numerator/denominator.
- the second one is this:
```
R=QQ[a,b]/(a*b-1)
1/a -- before: fails to realise it's b, after: gives b
lift(oo,R) -- before: fails to lift, after: lifts
```
basically, M2 did not correctly simplify the denominator when it was a unit, leading to lots of oddities (this is only the simplest example). now it does a much cleaner job.
## AI disclosure
I formulated the fixes, but codex helped with the c++.


M2/Macaulay2/e/rings/frac.cpp
```diff
@@ -111,12 +111,13 @@ void FractionField::simplify(frac_elem *f) const
   if (use_gcd_simplify)
     {
       y = f->denom;
-      if (R_->is_equal(y, R_->one())) return;
```
d-torrance, line 114: What happened to this return?  What if the denominator is 1?

M2/Macaulay2/e/rings/frac.cpp
```diff
@@ -99,6 +100,30 @@ ring_elem FractionField::set_non_unit_frac(ring_elem top) const
   return zero();
 }
 
+bool FractionField::simplify_unit_denominator(frac_elem *f) const
+{
+  if (R_->is_equal(f->denom, R_->one())) return true;
+
+  ring_elem denom_inverse;
+  if (dynamic_cast<const PolyRingQuotient *>(R_) != nullptr)
```
d-torrance: What's happening with this check?  Are we testing if `R` is a quotient or something?  And why the two different cases?

M2/Macaulay2/tests/normal/frac.m2
```diff
@@ -80,6 +80,13 @@ assert try (a/b; false) else true
 getNonUnit B
 isField B
 -----------------------------------------------------------------------------
+R = QQ[a,b]/(a*b-1)
+F = frac R
+assert(numerator(1/b) == a)
+assert(denominator(1/b) == 1_R)
+assert(lift(1/b,R) == a)
+assert(lift((b+1)/b,R) == a+1)
+-----------------------------------------------------------------------------
```
d-torrance, line 89: Since we're also updating this for local rings, it would be good to add some tests for them too.

d-torrance (changes requested):
Reviewed with @local-ring.  Just a couple questions and comments

M2/Macaulay2/e/rings/frac.cpp
pzinn: it got moved to `FractionField::simplify_unit_denominator`, which is called earlier.

M2/Macaulay2/e/rings/frac.cpp
pzinn: yeah this is a bit cryptic. the point is, for quotient rings `is_unit` calls `invert`. so we just call `invert` directly. But for other rings we can't do that because `invert` might throw an error, so we call `is_unit` first.
ideally, one should rewrite some of the code related to inverting ring elements for more consistent behaviour, but it seemed outside the scope of this PR.

M2/Macaulay2/tests/normal/frac.m2
pzinn: test added

M2/Macaulay2/e/rings/frac.cpp
d-torrance: Would you add a comment mentioning this?

M2/Macaulay2/packages/LocalRings/tests.m2
```diff
@@ -957,6 +957,15 @@ end--
   gens gb Iloc
 ///
 
+TEST /// -- normalize ambient-unit denominators
```
d-torrance: This test never gets run because of the `end` on line 875.  Would you move it up?

(While you're at it, would you mind moving up the `promote`/`lift` test just underneath it?  I fell into the same trap in #4204 lol.)

M2/Macaulay2/e/rings/frac.cpp
```diff
@@ -99,6 +100,31 @@ ring_elem FractionField::set_non_unit_frac(ring_elem top) const
...
 
+bool FractionField::simplify_unit_denominator(frac_elem *f) const
+{
+  if (R_->is_equal(f->denom, R_->one())) return true;
+
+  ring_elem denom_inverse;
+  // TODO uniformise behaviour of invert for noninvertible elements
+  if (dynamic_cast<const PolyRingQuotient *>(R_) != nullptr)
+    {
+      denom_inverse = R_->invert(f->denom); // for quotient rings, don't call is_unit since it calls invert internally
```
d-torrance, line 110: With this, we can run into #3973, where we didn't before:

```m2
i1 : K = toField(QQ[a]/(a^2 - 2))

o1 = K

o1 : PolynomialRing

i2 : B = K[t]/(t^2 - a)

o2 = B

o2 : QuotientRing

i3 : 1/(t + 1)
--internal error: ring element gcd computation failed
Aborted (core dumped)
```

In 1.26.06, we get:

```m2
i3 : 1/(t + 1)

       1
o3 = -----
     t + 1

o3 : frac B
```

M2/Macaulay2/e/rings/frac.cpp
```diff
@@ -108,15 +134,17 @@ ring_elem FractionField::fraction(const ring_elem top,
...
   if (use_gcd_simplify)
     {
       y = f->denom;
-      if (R_->is_equal(y, R_->one())) return;
-      x = f->numer;
-      const RingElement *a = RingElement::make_raw(R_, x);
-      const RingElement *b = RingElement::make_raw(R_, y);
-      const RingElement *c = rawGCDRingElement(a, b, nullptr, false);
-      if (!c) return;
+      if (!R_->is_equal(y, R_->one()))
```
d-torrance: If $y = 1$, then we would have already returned a few lines ago (since `simplify_unit_denominator(f)` would have returned `true`, so this `if` statement isn't necessary.  (And the same comment for the local ring version below.)

M2/Macaulay2/e/rings/frac.cpp
```diff
@@ -99,6 +100,31 @@ ring_elem FractionField::set_non_unit_frac(ring_elem top) const
   return zero();
 }
 
+bool FractionField::simplify_unit_denominator(frac_elem *f) const
+{
+  if (R_->is_equal(f->denom, R_->one())) return true;
+
+  ring_elem denom_inverse;
+  // TODO uniformise behaviour of invert for noninvertible elements
+  if (dynamic_cast<const PolyRingQuotient *>(R_) != nullptr)
```
d-torrance: I think `R_->is_quotient_ring()` would be clearer

M2/Macaulay2/e/rings/frac.cpp
```diff
@@ -99,6 +100,31 @@ ring_elem FractionField::set_non_unit_frac(ring_elem top) const
...
+    {
+      denom_inverse = R_->invert(f->denom); // for quotient rings, don't call is_unit since it calls invert internally
+      if (R_->is_zero(denom_inverse)) return false; // for non invertible elements, returns zero denominator
```
d-torrance: What does "returns zero denominator" mean in this comment?

d-torrance (changes requested):
Thanks for making the changes!  I have a few other suggestions.  (Thanks to Claude :robot: for suggesting a few things, too.)

M2/Macaulay2/e/rings/frac.cpp
pzinn: oh yeah the comment was not quite right, I just meant, returns zero. fixing now.

pzinn:
all suggestions implemented. these are minimal fixes, e.g., for the example of #3973, M2 will output an error message instead of crashing; I have not tried to make it produce the actual inverse.

d-torrance (approved):
Looks good to me.  @local-ring - do you have any other comments?

PR #4470, opened 2026-07-15
Bump GitHub macOS runners to Tahoe

d-torrance:
GitHub is currently migrating the `macos-latest` runner to `macos-26` (see https://github.com/actions/runner-images/issues/14167), so we follow suit.

Draft for now to make sure the builds pass.

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -39,17 +39,17 @@ jobs:
           - cmake
         os:
           - ubuntu-24.04
-          - macos-15
+          - macos-26
         compiler:
           - default
         include:
           # This build tests Clang rather than AppleClang (keep)
           - build-system: cmake
-            os: macos-15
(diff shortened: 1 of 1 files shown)


d-torrance:
Builds passed!

dimpase (approved):
lgtm

dimpase (changes requested):
one needs to fix [.github/workflows/build-dmg.yml](https://github.com/Macaulay2/M2/blob/1a37f6fe95badb1e5f99707a485503d213fd9d49/.github/workflows/build-dmg.yml#L31) in the same way, too

dimpase:
perhaps migrate to `macos-latest` is a better idea than pinning the version explicitly

d-torrance:
For the build-dmg workflow, we want to lowest possible macos version since we're trying to support older machines with the dmg files

PR #4471, opened 2026-07-15
Fix discriminant error with polynomials with identically vanishing derivatives (issue #4449)

### LINKED ISSUE #4449: discriminant throwing an error when it probably shouldn't
typical example:
```
i1 : R=ZZ/2[x]

o1 = R

o1 : PolynomialRing

i2 : F=x^24+x^22+x^20+x^14+x^12+x^8+x^4

      24    22    20    14    12    8    4
o2 = x   + x   + x   + x   + x   + x  + x

o2 : R

i3 : discriminant(F,x)
stdio:3:12(3): error: expected nonzero polynomials
```
you can replace `ZZ/2` with any other `ZZ/p` and exponents divisible by `p`.

tbrazel:
Fixes https://github.com/Macaulay2/M2/issues/4449

Given a univariate polynomial whose derivative is identically zero (e.g. over a finite field), its discriminant should return the zero polynomial, rather than an error.

### DIFF
--- a/M2/Macaulay2/packages/Elimination.m2
+++ b/M2/Macaulay2/packages/Elimination.m2
@@ -332,6 +332,13 @@ assert(isHomogeneous J)
 
 assert(eliminate(x,eliminate(y,I)) == eliminate(y,eliminate(x,I)))
 ///
+
+-- Resultant of a polynomial whose derivative is identically zero should return zero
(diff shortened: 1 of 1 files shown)


tbrazel:
Oops sorry should target dev lemme do that again 😄

PR #4472, opened 2026-07-15
Fix discriminant error with polynomials with identically vanishing derivatives (issue #4449)

tbrazel:
Fixes https://github.com/Macaulay2/M2/issues/4449

Given a univariate polynomial whose derivative is identically zero (e.g. over a finite field), its discriminant should return the zero polynomial, rather than an error.

### DIFF
--- a/M2/Macaulay2/packages/Elimination.m2
+++ b/M2/Macaulay2/packages/Elimination.m2
@@ -332,6 +332,13 @@ assert(isHomogeneous J)
 
 assert(eliminate(x,eliminate(y,I)) == eliminate(y,eliminate(x,I)))
 ///
+
+-- Resultant of a polynomial whose derivative is identically zero should return zero
+TEST ///
+R=ZZ/2[x]
+F=x^24+x^22+x^20+x^14+x^12+x^8+x^4
+assert(discriminant(F,x) == 0)
+///
 end
 loadPackage "Elimination"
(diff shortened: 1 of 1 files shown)


M2/Macaulay2/packages/Elimination.m2
```diff
@@ -125,8 +125,10 @@ resultant(RingElement, RingElement, RingElement) := o -> (f,g,x) ->
      det sylvesterMatrix(f,g,x)
 
 discriminant = method(Options => { Algorithm => null })
-discriminant(RingElement, RingElement) := RingElement => o -> (f,x) -> resultant(f, diff(x,f), x, o)
-
+discriminant(RingElement, RingElement) := RingElement => o -> (f,x) -> (
+     if diff(f,x) == 0 then return 0;
+     resultant(f, diff(x,f), x, o)
+)
```
mikestillman, line 131: The return value here is 0 in the ring ZZ.  

Perhaps instead we should use return substitute(0_(ring f), x => 0).  This will put the answer into either the ring R of f (if more variables than x), or in the coefficient ring of R.  Bit of a hack, to be sure.  This is basically what sylvesterMatrix does though.

What do you think?

mikestillman (approved):
I would be happy with your 0, or with the suggested on in the review comments. Please choose which you think is better.

PR #4373, opened 2026-05-26
Add workflow to prevent PR's against stable

d-torrance:
To hopefully keep us (well, me) from doing things like merging #4287 into the wrong branch again....

Opening this against `stable` for now to test it out.  (*Edit:* It [worked](https://github.com/Macaulay2/M2/actions/runs/26475742560/job/77960721576?pr=4373)!)

## AI Disclosure

Claude wrote pretty much everything

PR #4473, opened 2026-07-16
Test "no PR's to stable" workflow

d-torrance:
The workflow introduced in #4373 was supposed to prevent PR's *to* `stable`, but it fired in #4449, which correctly targetted `development` but is coming *from* `stable`. 

The name of a branch on a user's fork doesn't matter, so it shouldn't have fired.

This is just a silly PR to see if it fires again with a PR from `stable` -> `development`.


d-torrance:
Huh, it didn't fire.  Weird...

d-torrance:
Oh I think I see what happened -- https://github.com/Macaulay2/M2/pull/4471 targetted `stable` but was closed.  #4472 correctly targetted `development` but maybe it just carries the same test failure info since it's based on the same commit as the first PR

PR #4478, opened 2026-07-18
Check that we're working over the same ring in "coefficient"

### LINKED ISSUE #4452: InvariantRing + coefficient causes crash
Here's my broken example:

```
needsPackage"InvariantRing"
S = QQ[x_(0,0)..x_(2,2)]
E = id_(ZZ^3)
W =transpose matrix  flatten flatten apply(3, i-> apply(3,j-> entries ( (E^{i})|(E^{j}))))
T = diagonalAction(W, S)
equivariantHilbertSeries T
hs4 = equivariantHilbertSeries(T, Order => 4);
coefficient(ζ_0*ζ_1*ζ_2*ζ_3*ζ_4*ζ_5,hs4)
```
This pauses M2 with the output:
```-- SIGSEGV
-* stack trace, pid: 74763
```
Replacing the ```coefficient``` call with ```diff(ζ_0*ζ_1*ζ_2*ζ_3*ζ_4*ζ_5,hs4)``` gives the expected output. It seems that  ```coefficient``` should work if ```diff``` works in this situation.

d-torrance:
This is already what the [docs](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_coefficient.html) say we should do, and it can lead to strange behavior (e.g., over tower rings a la #4452) if we don't check.  Also add some unit tests.

Draft for now to test the builds -- this is a slight breaking change.

## AI Disclosure

Claude Code helped me diagnose the problem, but I wrote the code myself.

### DIFF
--- a/M2/Macaulay2/packages/Schubert2.m2
+++ b/M2/Macaulay2/packages/Schubert2.m2
@@ -1282,7 +1282,7 @@ extensionAlgebra(RingMap, RingElement) := opts -> (f, c) -> (
      -- 1) it's not a ring map
      -- 2) in general promote(promote(d,B),D) != d.
      promote(D,B) := (d,B) -> (
-	  dcoeffs := apply(E_0..E_(n-1), m -> f(coefficient(D1_(m), DtowertoD^-1 d)));
+	  dcoeffs := apply(E_0..E_(n-1), m -> f(coefficient(D1_(m), (DtowertoD^-1 d)^D1)));
 	  dmatrix := matrix{apply(dcoeffs, coeff -> {coeff})};
 	  first flatten entries (Bbasis * dmatrix)
 	  );
--- a/M2/Macaulay2/m2/monoids.m2
+++ b/M2/Macaulay2/m2/monoids.m2
@@ -162,7 +162,7 @@ coefficient(RingElement, RingElement) := (m, f) -> (
     R := ring f;
     k := coefficientRing R;
     -- TODO: audit this code and how it is used in towers and inexact fields
-    if size m === 1 and leadCoefficient m == 1
+    if instance(m, R) and size m === 1 and leadCoefficient m == 1
     then promote(rawCoefficient(raw k, raw f, rawLeadMonomialR m), k)
     else error "coefficient: expected a monomial from the same ring")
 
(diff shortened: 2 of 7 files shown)


d-torrance:
Alright, finally whack-a-moled all the packages that were using `coefficient` with arguments from different rings.  I'm not sure if I fixed each of them in the most efficient way, but now everything passes.

pzinn:
can you give an example of something that used to work and no longer does with this PR?

d-torrance:
> can you give an example of something that used to work and no longer does with this PR?

Sure:

```m2
i1 : R = QQ[x,y,z,w];

i2 : f = random(2, R)

     9 2   1      3 2   1      7        2   4      2      8      7 2
o2 = -x  + -x*y + -y  + -x*z + -y*z + 7z  + -x*w + -y*w + -z*w + -w
     2     9      2     5      8            3      9      3      6

o2 : R

i3 : S = ZZ[x,y,z,w];

i4 : coefficient(x^2, f) -- x^2 is in S

     9
o4 = -
     2

o4 : QQ
```

pzinn:
weird. definitely shouldn't work!

d-torrance:
Tagging authors of packages where we updated `coefficient` to avoid the new error to make sure that everything looks okay:

* A1BrouwerDegrees @tbrazel 
* InvariantRing @galettof 
* PieriMaps @kellerlv 
* Schubert2 @mikestillman
* SparseResultants @giovannistagliano

M2/Macaulay2/packages/Schubert2.m2
```diff
@@ -1282,7 +1282,7 @@ extensionAlgebra(RingMap, RingElement) := opts -> (f, c) -> (
      -- 1) it's not a ring map
      -- 2) in general promote(promote(d,B),D) != d.
      promote(D,B) := (d,B) -> (
-	  dcoeffs := apply(E_0..E_(n-1), m -> f(coefficient(D1_(m), DtowertoD^-1 d)));
+	  dcoeffs := apply(E_0..E_(n-1), m -> f(coefficient(D1_(m), (DtowertoD^-1 d)^D1)));
```
MichaelABurr, line 1285: What is this ^D1 doing here?  This seems to change the ring.

MichaelABurr (approved):
This mostly looks good to me.  There's one change that appears to change the ring, but it seems like this is what the ring should have been originally.

M2/Macaulay2/packages/Schubert2.m2
d-torrance: Yeah it's lifting the 2nd argument to `D1` to match the 1st argument

galettof:
Thank you for fixing this in InvariantRing.

tbrazel:
This looks good for me, `A1BrouwerDegrees` still runs as expected

giovannistagliano:
I ran a few quick tests on `SparseResultants` and everything seems to work as expected. Thanks.

local-ring (approved):

PR #4479, opened 2026-07-19
Add M2-language-server manpage

### LINKED ISSUE #4446: M2-language-server needs a manual page
We're getting a [no-manual-page](https://udd.debian.org/lintian-tag/no-manual-page) Lintian warning on the Debian package.

d-torrance:
We add a manpage for the `M2-language-server` script, fixing #4446.  The manpage itself is generated using `help2man` from the new `--help` and `--version` options that have been added to `M2-language-server`.

For now, this only installs the manpage using the autotools build (which the Debian package uses, so this will fix the Lintian warning mentioned in #4446).  The language server itself isn't installed yet in the cmake build (see #4453), so we hold off on the manpage there.

## AI Disclosure

I used Claude Code to write the `usage` function and the while loop for checking command line options in `M2-language-server`.


M2/Macaulay2/packages/LanguageServer/M2-language-server
```diff
@@ -1,5 +1,51 @@
...
+	    ;;
+	-v|--version)
+	    M2 --silent                                         \
+	       -e 'needsPackage "LanguageServer"'               \
+	       -e '<< "M2-language-server "'                    \
+	       -e '<< (options LanguageServer).Version << endl' \
+	       -e 'exit 0'
+	    exit 0
+	    ;;
+	--)
```
MichaelABurr, line 33: An empty "--" is treated as a non-command?  This seems fine, but this could return an error.

M2/Macaulay2/packages/LanguageServer/M2-language-server
```diff
@@ -1,5 +1,51 @@
...
+	    shift
+	    break
+	    ;;
+	-*)
+	    echo "$progname: unrecognized option '$1'" >&2
+	    usage >&2
+	    exit 1
+	    ;;
+	*)
```
MichaelABurr, line 42: Any non-options are ignored?  This is ok, but also could return an error.

M2/Macaulay2/packages/LanguageServer/Makefile
```diff
@@ -0,0 +1,11 @@
+# run "make clean && make" when updating the LanguageServer package version
+# and/or the M2-language-server --help output, and then commit
+# M2-language-server.1 to git
+
+M2-language-server.1: M2-language-server
```
MichaelABurr, line 7: Personally, I like to put "default: M2-language-server.1" as the first item in a Makefile, so that multiple targets can be default, if desired.  But, this is more a style thing than anything else

MichaelABurr (approved):
This looks good to me.  I put a few comments in, but they are more stylistic and can be ignored.

M2/Macaulay2/packages/LanguageServer/M2-language-server
d-torrance: As soon as we encounter a non-option, it breaks the loop.  Same with `--` above (see https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html Guideline 10)

M2/Macaulay2/packages/LanguageServer/Makefile
d-torrance: Good point -- I'll add that

PR #4480, opened 2026-07-19
Raise an error when attempting to close stdio or stderr

### LINKED ISSUE #4392: closing stdio breaks M2
This happened to me by accident. I feel like it either it should give an error, or stdio should reopen automatically:
```m2
$ M2
Macaulay2, version 1.26.05
Type "help" to see useful commands

i1 : << close
stdio:1:0:(3):[1]: error: at print: expected an output file
stdio:1:0:(3):[1]: error: after print: expected an output file
stdio:2:0:(3):[1]: error: at print: expected an output file
```

d-torrance:
`stdio` supports both input and output, so we check in both `closeIn` and `closeOut`

`stderr` only supports output, so we only check in `closeOut` (`stderr << closeIn` already raises an error since it's not open for input)

Closes: #4392 

No AI used!

### Before

```m2
i1 : stdio << close
stdio:1:6:(3):[1]: error: at print: expected an output file
stdio:1:6:(3):[1]: error: after print: expected an output file
```

```m2
i1 : stderr << close

o1 = stderr

o1 : File

i2 : 1/0

i3 : -- where's the error message?
```

### After

```m2
i1 : stdio << close
stdio:1:6:(3):[1]: error: can't close stdio

i2 : stderr << close
stdio:2:7:(3):[1]: error: can't close stderr
```

### DIFF
--- a/M2/Macaulay2/d/stdio.d
+++ b/M2/Macaulay2/d/stdio.d
@@ -518,6 +518,7 @@ export closeListener(o:file):(errmsg or null) := (
      null());
 
 export closeIn(o:file):(errmsg or null) := (
+     if o == stdIO then return errmsg("can't close stdio");
      stat := 0;
      if o.infd == NOFD then return errmsg("close: file not open");
      if o == stdIO then return null();			    -- silently refuse to close stdIO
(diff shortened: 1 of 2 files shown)


MichaelABurr (approved):
Simple, and looks OK to me.  I don't expect this code to be encountered frequently, so I don't expect a performance hit.

PR #4482, opened 2026-07-21
Add texMath and mathML methods for PadicFieldFamily

d-torrance:
Also bump package version to 0.3

As I was preparing my talk on this package for ICMS, I realized that $p$-adic fields look awful in webapp mode!

### Before

image

### After
image

## AI Disclosure

No AI used

### DIFF
--- a/M2/Macaulay2/packages/Padic.m2
+++ b/M2/Macaulay2/packages/Padic.m2
@@ -56,7 +56,10 @@ endpkg = msg -> (
 if not ForeignFunctions#"private dictionary"#?"foreignFunction"
 then endpkg "foreign function interface is not available"
 
-flint = try openSharedLibrary "flint" else endpkg "flint is not available"
+flint = (
+    try openSharedLibrary "flint"
+    else try foreignFunction("fmpz_init", void, voidstar) then null
+    else endpkg "flint is not available")
 
 export {
     -- methods
--- a/M2/Macaulay2/packages/Padic.m2
+++ b/M2/Macaulay2/packages/Padic.m2
@@ -16,8 +16,8 @@
 
 newPackage("Padic",
     Headline => "p-adic numbers",
-    Version => "0.2",
-    Date => "June 5, 2026",
+    Version => "0.3",
+    Date => "July 19, 2026",
     Authors => {{
 	    Name => "Doug Torrance",
 	    Email => "dtorrance9@gatech.edu",
@@ -32,6 +32,9 @@ newPackage("Padic",
 ---------------
 
 -*
+
+0.3 (2026-07-19, M2 1.26.11)
+* add texMath and mathML methods
 
 0.2 (2026-06-05, M2 1.26.06)
 * update my contact info
@@ -216,6 +219,8 @@ PadicFieldFamily.synonym = "p-adic field family"
 expression PadicFieldFamily := kk -> Subscript(QQ, prime kk)
 net PadicFieldFamily := net @@ expression
 toString PadicFieldFamily := toString @@ expression
+texMath PadicFieldFamily := texMath @@ expression
+mathML PadicFieldFamily := mathML @@ expression
 
 PadicNumber = new Type of Number
 PadicNumber.synonym = "p-adic number"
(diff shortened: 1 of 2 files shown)


d-torrance:
I also added corresponding methods for `PadicNumber` objects.

### Before
image

### After
image

M2/Macaulay2/packages/Padic.m2
```diff
@@ -242,6 +247,19 @@ toString PadicNumber := x -> (
     -- from src/padic/get_str.c
     n := (N - v) * (2 * numdigits p + numdigits max(abs v, abs N) + 5) + 1;
     value padicGetStr(concatenate(n:"\0"), x.number, x.context))
+expression PadicNumber := x -> (
```
MichaelABurr, line 253: Perhaps there should be a global variable indicating the number of terms displayed?  This would be different from the precision.

M2/Macaulay2/packages/Padic.m2
```diff
@@ -242,6 +247,19 @@ toString PadicNumber := x -> (
...
+    s := toString x;
+    Sum apply(separate(" \\+ ", s), term -> (
```
MichaelABurr, line 255: Are p-adics implemented so that the coefficients are never negative?  It looks like that is the case, but I didn't check.  Either way, it wouldn't be a problem to have +- appearing.

M2/Macaulay2/packages/Padic.m2
d-torrance: That's the precision of the p-adic, which is built into the object itself. By default, it's 20, but you can set it with, e.g., `QQ_7(30, -1)`

M2/Macaulay2/packages/Padic.m2
MichaelABurr, line 255: What is the desired behavior when a term is 0?

MichaelABurr (approved):
These changes are OK, but I've provided some feedback for potential improvement.

M2/Macaulay2/packages/Padic.m2
d-torrance: That's right - they're always between 0 and p - 1

d-torrance: It just isn't printed, e.g.,:

```m2
i2 : QQ_7 50

o2 = 1 + 1*7^2

o2 : QQ  (of precision 20)
       7
```

d-torrance:
I added a couple commits (also to the ForeignFunctions package) that should help with loading the package on systems (such as macos) where flint is hard to find

M2/Macaulay2/packages/Padic.m2
```diff
@@ -242,6 +250,19 @@ toString PadicNumber := x -> (
...
+            factors := separate("\\*", term);
+            if #factors == 1 then factors#0
```
andrew-tawfeek: Could `factors#0` be converted with `value` in both branches? Right now coefficients remain strings, so `texMath` renders them as `\texttt{5}` and `mathML` uses `<mtext>` rather than numeric markup.

d-torrance: Good catch!  Updated

M2/Macaulay2/packages/Padic.m2
d-torrance: Ah, just noticed the post-edit note about this being distinct from the precision.  So something like a p-adic `printingPrecision`? 

I think that would make more sense if M2 was in charge of converting these things into strings, but we're letting FLINT deal with that.  In this PR, we're just re-formatting the strings emitted by FLINT for better webapp/TeXmacs support.

andrew-tawfeek (approved):
This looks good. I exercised the new Nothing overloads for single-argument, variadic, and foreignSymbol calls, and confirmed that the Padic fallback binds FLINT through RTLD_DEFAULT when openSharedLibrary cannot find the unversioned library. The expression conversion also produced numeric TeX and MathML for zero, coefficients, and negative exponents.

PR #4489, opened 2026-08-01
Add support for generating coverage reports w/ gcov

d-torrance:
This replaces https://github.com/MichaelABurr/M2/pull/65

As discussed in the C++ refactoring group Zoom meeting yesterday:  If we build M2 using `./configure --enable-gcov`, then we can run `make coverage-report` to get a [gcovr](https://gcovr.com/)-generated coverage report of the engine and interpreter.  In particular, everything that was covered since we built M2 (or ran `make coverage-reset`).

This will be useful to figuring out which features still need unit testing (e.g., by running `make coverage-reset && make -C Macaulay2/e check && make coverage-report`, we can see what all the unit tests hit.)  There's also CMake support that I haven't tested.

## AI Disclosure

:robot: This was Claude through and through :robot:

### DIFF
--- a/M2/cmake/gcov.cmake
+++ b/M2/cmake/gcov.cmake
@@ -1,8 +1,12 @@
-# gcc/gcov code coverage for the engine (configure with -DGCOV=ON).
+# gcc/gcov code coverage (configure with -DGCOV=ON).
 # Instrumentation flags are added in configure.cmake; this module only provides
 # convenience targets to clear and report on the accumulated .gcda data.  The data
 # itself is generated automatically whenever a binary built in coverage mode exits
 # (the unit tests, ctest, or the M2 binary run on your own code).
+#
+# By default the report covers the whole source tree; pass extra gcovr options
+# via GCOVR_OPTIONS, e.g. -DGCOVR_OPTIONS="--filter Macaulay2/e/" to scope it to
+# the engine, or "--exclude .*/unit-tests/.*" to drop the test sources.
 #
 # Usage:
 #   cmake -S . -B BUILD/cov -GNinja -DGCOV=ON -DCMAKE_BUILD_TYPE=Debug -DBUILD_TESTING=ON
@@ -13,9 +17,11 @@
 
 if(GCOV)
   find_program(GCOVR NAMES gcovr)
-  set(_engine_objdir ${CMAKE_BINARY_DIR}/Macaulay2/e/CMakeFiles/M2-engine.dir)
   set(_coverage_dir  ${CMAKE_BINARY_DIR}/coverage)
   set(_coverage_index ${_coverage_dir}/index.html)
+  set(GCOVR_OPTIONS "" CACHE STRING
+    "Extra options passed to gcovr for the coverage-report target")
+  separate_arguments(_gcovr_options UNIX_COMMAND "${GCOVR_OPTIONS}")
 
   # coverage-reset clears all accumulated data; the M2 binary may have written
   # .gcda anywhere in the tree, so this is deliberately not engine-scoped.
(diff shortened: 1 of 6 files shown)


mikestillman:
I suppose we need to install gcovr first?  Where should we put that info?  I don't want (or can't) install using pip globally...

mikestillman:
Here is one set of instructions (from chatgpt).  Sound reasonable?  (I've never used pipx before).

Do the following once (using homebrew, on mac)
```
brew install pipx
pipx ensurepath
pipx install gcovr
```

Restart a terminal, do

```
gcovr --version
```

to upgrade or remove:
```
pipx upgrade gcovr
pipx uninstall gcovr
```

d-torrance:
It's available from most package managers (`apt install gcovr`, `brew install gcovr`, etc.)  Maybe we should add a wiki page or something?  Since it isn't a build dependency and we only need it for generating the reports, I think it might be overkill to check for it

d-torrance:
Yeah `pipx` is cool for scripts written in Python.  But `gcovr` itself is in Homebrew, so I don't think that's necessary -- just `brew install gcovr`.

mikestillman:
The info about pipx suggested by chatGPT is too complicated and not needed on a mac: I believe just doing `brew install gcovr` does the trick.

M2/cmake/gcov.cmake
```diff
@@ -0,0 +1,56 @@
...
+# Usage:
+#   cmake -S . -B BUILD/cov -GNinja -DGCOV=ON -DCMAKE_BUILD_TYPE=Debug -DBUILD_TESTING=ON
+#   cmake --build BUILD/cov --target M2-engine M2-unit-tests
+#   cmake --build BUILD/cov --target coverage-reset   # start from a clean slate
+#   ctest --test-dir BUILD/cov -R unit-tests          # (or run M2 however you like)
+#   cmake --build BUILD/cov --target coverage-report  # writes coverage/index.html
+
+if(GCOV)
+  find_program(GCOVR NAMES gcovr)
+  set(_coverage_dir  ${CMAKE_BINARY_DIR}/coverage)
```
andrew-tawfeek: Could we add a small CI smoke test with coverage enabled, especially for the CMake path? The current checks all use the default `GCOV=OFF`, so they don't exercise the instrumentation flags or either coverage target; the PR also notes that the CMake support hasn't been tested. Even a minimal configure/build/run/`coverage-report` job would catch toolchain and `gcovr` invocation problems.

d-torrance: That sounds like a great idea, but I think it's out of scope for this PR.  In particular, I don't think we want a distinct coverage-enabled build running as part of our usual test_build.yml workflow that runs during every PR.

But maybe like a weekly build that runs the report for the engine unit tests and uploads it to GitHub pages or something.

d-torrance:
Finally got around to testing the cmake version.  It was broken lol.  Claude fixed it up, cleaned up the code, and squashed everything down to 3 logical commits.

For review, this is how it works.  First install [gcovr](https://gcovr.com/en/stable/) (`brew install gcovr`, `apt install gcovr`, etc.).  Then:

## autotools
```shell
./configure --enable-gcov # + whatever other options
make -jN # N = number of cores
# C-c once you have a working binary (we really need "make M2-core" lol)
make coverage-reset
# run whatever M2 code you want to look at
make coverage-report
```

## cmake
```shell
cmake -DGCOV=ON # + whatever other options
ninja M2-core
ninja coverage-reset
# run whatever M2 code you want to look at
ninja coverage-report
```

andrew-tawfeek (approved):
This looks good. I verified both generated Autotools and CMake coverage paths with GCC and gcovr: instrumentation produced counter data, coverage-report wrote a nonempty HTML report, and coverage-reset removed the counters. The Autotools configuration also applied --coverage -O0 after Fortran and BLAS detection and disabled stripping as intended.

mikestillman:
I suggest we merge this now, and fix it if there are any issues.  As it only seems to affect people who are using it, accepting it should not cause real problems for anyone?

M2/cmake/gcov.cmake
```diff
@@ -0,0 +1,44 @@
...
+  set(_coverage_link "coverage report: \
+${_esc}]8;;file://${_host}${_coverage_index}${_esc}\\${_coverage_index}${_esc}]8;;${_esc}\\")
+
+  # not engine-scoped: the M2 binary may write .gcda anywhere in the tree
+  add_custom_target(coverage-reset
+    COMMENT "Deleting accumulated .gcda coverage counters"
+    COMMAND find ${CMAKE_BINARY_DIR} -name "*.gcda" -delete
+    VERBATIM)
+
+  if(NOT GCOVR)
```
MichaelABurr, line 28: If gcovr is not found, is -O0 still used?  Perhaps this would be desired if gcovr could be added later and the code wouldn't need to be recompiled.

MichaelABurr (commented):
Post merge review, but it looks good.

PR #4519, opened 2026-08-04
Mark M2_flint_abort as FLINT_NORETURN

d-torrance:
Otherwise, compilation can fail w/:

    ../../../../Macaulay2/d/main.cpp:92:3: fatal error: no matching function for call to 'flint_set_abort'
       92 |   flint_set_abort(M2_flint_abort);
          |   ^~~~~~~~~~~~~~~
    /usr/include/flint/flint.h:222:6: note: candidate function not viable: no known conversion from 'void ()' to 'void (*)() __attribute__((noreturn))' for 1st argument
      222 | void flint_set_abort(FLINT_NORETURN void (*func)(void));
          |      ^               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

I ran into this problem compiling M2 with clang 18.1.3 in Ubuntu 24.04.

### DIFF
--- a/M2/Macaulay2/bin/main.cpp
+++ b/M2/Macaulay2/bin/main.cpp
@@ -52,7 +52,7 @@ bool interrupts_interruptShield;
 void* interpFunc(ArgCell* vargs);
 void* profFunc(ArgCell* p);
 void* testFunc(ArgCell* p);
-void  M2_flint_abort(void);
+FLINT_NORETURN void  M2_flint_abort(void);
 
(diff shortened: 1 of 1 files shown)


andrew-tawfeek (approved):
This looks good. `FLINT_NORETURN` matches the callback type expected by `flint_set_abort`, and applying it to both the declaration and definition is consistent with the fact that the function always calls `abort()`. I also checked that the macro exists throughout M2's supported FLINT range back to 2.6.3.

PR #4609, opened 2026-08-10
Bundle Factory's gftables in the macOS DMG and test the DMG on a clean runner

### LINKED ISSUE #4596: macOS 1.26.06 DMG contains Homebrew-specific Factory gftables path
# macOS 1.26.06 DMG contains Homebrew-specific Factory gftables path

## Environment

- Macaulay2 1.26.06
- official macOS DMG
- macOS 14.8.8
- Apple Silicon (M3)
- MacPorts installed
- Homebrew not installed

## Problem

Starting Macaulay2 from the official 1.26.06 DMG fails during initialization with

```text
Macaulay2, version 1.26.06
.../share/Macaulay2/Core/factor.m2:175:10:(1):[11]: error:
sample Factory finite field addition table file missing,
needed for factorization:
/opt/homebrew/Cellar/factory/4.4.1_5/share/factory/
```

The DMG contains

```text
lib/Macaulay2/lib/libfactory-4.4.1.dylib
```

but does not contain the Factory `gftables`.

The installed runtime layout contains the build-time Homebrew path

```text
/opt/homebrew/Cellar/factory/4.4.1_5/share/factory/
```

in

```text
share/Macaulay2/Core/startup.m2
```

and the same path is also embedded in `M2-binary`.

Editing the installed `startup.m2` does not affect runtime behavior, so the embedded layout appears to be used during startup.

## Workaround

MacPorts installs the required Factory tables with Singular under

```text
/opt/local/share/factory/gftables/
```

including

```text
/opt/local/share/factory/gftables/961
```

Adding the following line to `share/Macaulay2/Core/factor.m2`, immediately before the search for `gftables/961`,

```m2
gfdirs = prepend("/opt/local/share/factory/", gfdirs)
```

allows Macaulay2 to start normally.

Finite-field factorization then works, e.g.

```m2
R = GF(31^2)[x]
factor(x^4 + x + 1)
```

as does ordinary factorization.

TeXmacs integration with Macaulay2 1.26.06 also works correctly after this workaround.

## Expected behavior

The macOS DMG should either

1. bundle the Factory finite-field tables required at runtime, or
2. determine their location without relying on an absolute Homebrew build-machine path.

The current DMG therefore appears to contain a packaging/build-layout issue rather than a Macaulay2 core or TeXmacs problem.

d-torrance:
Fixes #4596.

## The bug

The 1.26.06 macOS DMG bakes in a Homebrew-specific path to Factory's finite
field addition tables, so M2 fails during initialization on any Mac that does
not happen to have that exact Homebrew cellar:

```text
sample Factory finite field addition table file missing, needed for factorization:
/opt/homebrew/Cellar/factory/4.4.1_5/share/factory/
```

## What this does

**Bundle the tables** (`60065dfcb3`, `3525e382c9`) — stage Factory's `gftables`
into `share/Macaulay2/Core/factory/` and make the configured path relative to
the prefix instead of absolute. `PRE_GFTABLESDIR` and `m4/ax_recursive_eval.m4`
become unused and are removed.

**Test the DMG on a runner that never installed the build dependencies**
(`f44f66a296`, `c009e02631`) — the build job installs the Homebrew formulas M2
is linked against, so it structurally cannot notice anything the DMG failed to
bundle. Testing on a separate, clean runner reproduces a user's machine. That
job is what found every remaining problem in this PR.

It also runs even when one architecture fails to build, so an x86_64 build
failure no longer silently discards the arm64 test results
([example of the old behavior](https://github.com/d-torrance/M2/actions/runs/31271280354),
where the whole test job was skipped because the Intel build failed).

**Make bundled external programs reachable** (`2230f6b69a`, `5305187d14`) —
`PHCpack.m2` and `Bertini.m2` locate their programs with a shell `type`, not
`findProgram`, so they need them on `PATH`. The DMG staged both in
`libexec/Macaulay2/bin`, which only `findProgram` searches, so both packages
reported their program as absent and skipped their tests even though the
binaries were bundled and working. They now go in `bin/`, the directory the DMG
instructs users to add to `PATH`, and the test targets run M2 with that
directory on `PATH` the way a user would.

**Alias unversioned dylib names** (`854d8e60e8`) — `bin/M2` already puts
`lib/Macaulay2/lib` on `DYLD_LIBRARY_PATH`, but dylibbundler copies only
versioned files (`libmpfi.0.dylib`), while `openSharedLibrary "mpfi"` opens
`libmpfi.dylib`. Symlink the unversioned names for mpfi (ForeignFunctions'
test), flint (Padic), and mpfr (used in the ForeignFunctions documentation).

**Tolerate concurrent directory creation in the engine makefile**
(`43611ac33d`) — not DMG-specific, and the one commit here that affects normal
builds. On macOS `MKDIR_P` is `/usr/bin/install -c -d`, since autoconf does not
consider BSD `mkdir -p` thread-safe; unlike `mkdir -p`, `install -d` fails when
the directory already exists. A parallel build in which two targets under the
same subdirectory start together then dies:

```text
MKDEP schreyer-resolutions/res-dep-graph.cpp
MKDEP schreyer-resolutions/res-f4.cpp
install: mkdir schreyer-resolutions: File exists
make[3]: *** [Makefile.common:69: .../res-f4.dep] Error 71
```

## Testing

Built and tested on `macos-14` (arm64) and `macos-15-intel` in
[this run](https://github.com/d-torrance/M2/actions/runs/31327717726). Both
DMGs build; both install and pass `--check 1` and `--check 2` on a clean
runner; arm64 passes `--check 3` apart from the known issue below. Package
tests remain arm64-only, since the Intel runners time out on `--check 3`.

With bertini now discoverable, `check(41, "NumericalAlgebraicGeometry")` runs
for the first time and hits #4368:

```text
error: no method found for applying bertiniRefineSols to: ...
```

`NumericalAlgebraicGeometry/Bertini/Bertini.interface.m2:68` calls
`bertiniRefineSols(ZZ, List, List)`, but `Bertini.m2` defines only
`(String, ZZ, List, String)`, `(ZZ, List, String)`, and `(ZZ, List)`. This is
pre-existing API drift between the two packages, unrelated to packaging, and is
not addressed here — it was simply invisible in this workflow while bertini
could never be found.

## Attribution

The code and this description were written by Claude Code under my review. I
verified the behavior against the CI runs and am responsible for the content.


dimpase:
I don't know what these DMGs are, and why a macOS-specific format is used to package data (?). Would be nice to know more about it.

I guess DMG contains a full macOS installer of M2, which bundles parts of Homebrew?

d-torrance:
Yeah a .dmg file is a disk image file used to distribute applications in macOS. It was the main way we used to distribute M2 before Homebrew.

A few people have expressed interest in having them again, so I had Claude create this Github workflow to do it for us at the last release. Rather than building all the dependencies ourselves, we grab them from Homebrew, but we don't want the user to have to have any Homebrew packages installed. That's the bug this PR is fixing - it was still pointing at the Homebrew version of factory's finite field tables.

dimpase:
So this is an independent of Homebrew macOS installer, which should be putting everything into `/Applications/`, but it didn't?

Or it might mess with the existing installation of Homebrew?

dimpase:
Normally speaking such installers adjust installation paths at installation time, as far as I know. Thus the whole PR looks suspicious to me.

d-torrance:
> So this is an independent of Homebrew macOS installer, which should be putting everything into `/Applications/`, but it didn't?

Exactly -- it had everything it needed in `/Applications` *except* for the factory gftables.  M2 looked for them in Homebrew's installation location (since that's where they were when everything was built in the GitHub workflow), but of course it probably couldn't find them unless the user had also used Homebrew to install factory at some point.

dimpase (approved):
OK, good.

PS. I'd like to know more about the technology of packaging Homebrew bits into a DMG, it doesn't seem to be documented anywhere...

d-torrance:
It's all in `M2/BUILD/dmg/Makefile`, in the `bundle` target.  It looks like using [dylibbundler](https://github.com/auriamg/macdylibbundler) is the key.  (Full disclosure - this was all written by Claude...)

andrew-tawfeek (approved):
This looks good. I verified the DMG configure path locally: the Factory tables are staged with gftables/961 present, and startup.m2 records the ${prefix}-relative path rather than the Homebrew cellar. I also checked the separate clean-runner workflow results at the tested tree; both architectures built and passed install/basic/core, with only the documented pre-existing NumericalAlgebraicGeometry package failure on arm64.

PR #4635, opened 2026-08-12
Add some convenience targets for autotools builds

d-torrance:
`M2-engine`, `M2-binary`, and `M2-core` to match the cmake build

We also copy (using `cp`, not `installPackage`) the *Style* package so we don't get warnings when we fire up M2.

No AI used!


MichaelABurr (approved):
Looks good to me.

PR #4653, opened 2026-08-16
Improve pruning of SkewCommutative rings

joel-dodge:
Better support for SkewCommutative variables when pruning a ring. Remap skew indices to match with new variables array. 
When configuring a LocalRing, try pruning the residue ring before computing it's field of fractions.


joel-dodge:
Pruning improvements were moved to https://github.com/Macaulay2/M2/pull/4658

The local ring stuff here is incomplete and requires more investigation into how the LocalRings package works.

PR #4655, opened 2026-08-17
Add HomotopyContinuationInterface package

glitch1729:
This PR adds the HomotopyContinuationInterface package, an interface 
from Macaulay2 to HomotopyContinuation.jl for numerical algebraic 
geometry computations.

This package accompanies my submission to the Journal of Software 
for Algebra and Geometry (JSAG).


M2/Macaulay2/packages/HomotopyContinuationInterface.m2
```diff
@@ -0,0 +1,681 @@
...
+
+newPackage(
+    "HomotopyContinuationInterface",
+    Version => "0.2",
+    Date => "August 2026",
+    Authors => {{Name => "Aditya Tyagi", Email => "", HomePage => ""}},
+    Headline => "an interface to HomotopyContinuation.jl",
+    Keywords => {"Numerical Algebraic Geometry"},
+    PackageExports => {"NAGtypes"},
+    DebuggingMode => true
```
d-torrance:
```suggestion
```

`DebuggingMode` should be false for distributed packages.  This is the default, so just removing this line is an easy fix.

d-torrance (changes requested):
Thanks for the contribution!  I can't count the number of times I've heard folks talk about HomotopyContinuation.jl recently, so this will be a valuable addition to Macaulay2, I think!

In addition to the `DebuggingMode` comment, a couple other notes:

* PR's should almost always target the `development` branch, not `stable`.  I've already fixed this for you.
* JSAG PR's should be in *draft* status until the technical review is complete.  I've also taken care of this for you.

M2/Macaulay2/packages/HomotopyContinuationInterface.m2
d-torrance: There should be an `OptionalComponentsPresent` key here to check that we have both Julia and HomotopyContinuation.jl available when running the package, and then also a `CacheExampleOutput => true` key/value pair so that the examples are cached.

Then commit those cached example files and include them in this PR.  That way, builds won't fail for folks who don't have Julia and/or HomotopyContinuation.jl available.  See some of the other interface packages like *Bertini* and *PHCpack* for examples.

It would be good to have them installed, though, on our GitHub runners, though, so we can actually check that the tests and examples work.  I'm working on that and should have a PR to your branch soon

d-torrance (changes requested):

M2/Macaulay2/packages/HomotopyContinuationInterface.m2
d-torrance: Okay -- once you merge https://github.com/glitch1729/M2/pull/1, this PR should be updated automatically, and then we should be able to run your examples and tests on the GitHub runners.

glitch1729:
Looked into the failing checks:

macOS builds: unrelated to this package both fail with an Abort trap: 6 crash during example capture in the SubalgebraBases package ("Translation and rotation sub-actions of the adjoint action of SE(3)"), with no reference to HomotopyContinuationInterface in the log.

Ubuntu build: this one does look related install fails with:

`ERROR: Load Error: Argument error: Package HomotopyContinuation not found in current path.
`
while capturing example results for HomotopyContinuationInterface. Looks like the Julia HomotopyContinuation.jl install from the CI workflow may not be getting picked up in this environment/path. @d-torrance since you set up that install step . could you take a look, or point me at what might be missing?

d-torrance:
Those failing builds were from before my PR installing Julia & HomotopyContinuation.jl was merged.  They're running again now! :crossed_fingers:

glitch1729:
Hi Doug Torrance,
All 5 checks have completed successfully. So What would be the next step from here?

Thanks for your time and your efforts.
Regards,
Aditya tyagi

d-torrance:
There will be a technical review.  That may take weeks/months, so be patient :)

M2/Macaulay2/packages/HomotopyContinuationInterface.m2
```diff
@@ -0,0 +1,680 @@
...
+-- examples are very welcome.
+
+newPackage(
+    "HomotopyContinuationInterface",
+    Version => "0.2",
+    Date => "August 2026",
+    Authors => {{Name => "Aditya Tyagi", Email => "", HomePage => ""}},
+    Headline => "an interface to HomotopyContinuation.jl",
+    Keywords => {"Numerical Algebraic Geometry"},
+    PackageExports => {"NAGtypes"},
```
d-torrance, line 32: As I mentioned above, even though it's working in the GitHub builds now, it would still be good to have `OptionalComponentsPresent => ...` and `CacheExampleOutput => true` (and the example files commited to git) here so folks can still build Macaulay2 w/o having Julia installed.

I'm not sure what the best thing to do is for `OptionalComponentsPresent`.  Certainly, running `command -v julia` will give us half the story, but we also want to check if HomotopyContinuation.jl is available.

d-torrance (changes requested):

M2/Macaulay2/packages/HomotopyContinuationInterface.m2
glitch1729: Thanks for the pointers! I went with checking for HC.jl itself rather than just julia ; the script tries using HomotopyContinuation in a try/catch and checks the exit code, so it should catch the case where Julia's there but the package isn't.

One thing I'm not 100% sure about: this means spinning up a Julia process just to check at load time, which isn't free. If you know of a better pattern other interface packages use for this, I'm all ears.

Still need to actually generate the cached example output on a machine with Julia + HC.jl ; will follow up with that once it's done.

d-torrance: Yeah, firing up a Julia process probably isn't great.  Maybe something like the following?

```m2
run "command -v julia > /dev/null" == 0 and isDirectory "~/.julia/packages/HomotopyContinuation"
```

glitch1729: That's much cleaner ; pushed, thanks! Also went ahead and generated + committed the cached example output while I had Julia set up locally, so both pieces from your original comment should be covered now.

M2/Macaulay2/packages/HomotopyContinuationInterface.m2
```diff
@@ -0,0 +1,687 @@
...
+-- only for auxiliary output (extracting the numeric linear slice of a
+-- witness set) are wrapped in a Julia try/catch, so a version mismatch
+-- there degrades gracefully (you still get points/dimension/degree) instead
+-- of crashing the whole computation. Contributions and testing on real
+-- examples are very welcome.
+
+-- Check whether `julia` is on PATH *and* the HomotopyContinuation.jl
+-- package is installed and loadable. We do this by asking Julia
+-- itself, rather than just checking for the executable, since a
+-- bare Julia install with no HC.jl would otherwise report as present.
```
d-torrance, line 22: This comment is out of date now

d-torrance (changes requested):

M2/Macaulay2/packages/HomotopyContinuationInterface.m2
```diff
@@ -0,0 +1,687 @@
...
+newPackage(
+    "HomotopyContinuationInterface",
+    Version => "0.2",
+    Date => "August 2026",
+    Authors => {{Name => "Aditya Tyagi", Email => "", HomePage => ""}},
+    Headline => "an interface to HomotopyContinuation.jl",
+    Keywords => {"Numerical Algebraic Geometry"},
+    PackageExports => {"NAGtypes"},
+    OptionalComponentsPresent => hcPresent,
+    CacheExampleOutput => true,
```
d-torrance, line 34:
```suggestion
    CacheExampleOutput => true,
    AuxiliaryFiles => true,
```

Now that we have the generated example files, we need `AuxiliaryFiles` here to tell us that there's more than just the .m2 file to this package.

d-torrance (changes requested):

glitch1729:
Applied, thanks!

glitch1729:
Just checking in on this. Let me know if there's anything else needed from my side. Happy to make any further changes. :)

glitch1729:
Resolved the merge conflict in test_build.yml — kept the updated cache config from development and preserved the Julia detection/caching steps. Would appreciate a look to make sure I merged that correctly, and the workflow run will need approval to kick off.

.github/workflows/test_build.yml
```diff
@@ -131,10 +131,27 @@ jobs:
       - name: Set up ccache
         uses: hendrikmuhs/ccache-action@v1.2.23
         with:
+          path: |
+            ~/.ccache
+            ~/work/M2/M2/BUILD/build/usr-host
```
d-torrance:
```suggestion
```

It looks like these lines got added back in with the merge.  They were removed in #4664

d-torrance (changes requested):

PR #4658, opened 2026-08-20
Improving behavior of `prune` for skew commutative rings

joel-dodge:
Update behavior of pruning with respect to SkewCommutative variable settings on a polynomial / quotient ring. Previously, `SkewCommutative` option was passed through unchanged producing non-sensical and confusing misconfigured rings when pruning away skew commutative variables. 

General improvement to handling of `SkewCommutative` option when creating a monoid. In particular, validate user input and normalize it so that it is stored as a sorted deduped list of non-negative integers.

Previous `prune` behavior:
```
i1 : kk = ZZ/17; R = kk[a..d, SkewCommutative => true]
o2 = R
o2 : PolynomialRing, 4 skew commutative variable(s)

i3 : I = ideal {a, b, c, d}
o3 = ideal (a, b, c, d)
o3 : Ideal of R

i4 : R' = prune(R/I)
o4 = R'
o4 : PolynomialRing, 4 skew commutative variable(s)

i5 : R'.SkewCommutative
o5 = {0, 1, 2, 3}
o5 : List

i6 : vars R'
o6 = 0
              1
o6 : Matrix R'  <-- 0
```

See improved behavior in test cases included with this PR.


M2/Macaulay2/m2/polyrings.m2
```diff
@@ -154,6 +154,7 @@ Ring Monoid := PolynomialRing => (R, M) -> (
     MWeyl := if MOpts.?WeylAlgebra     then MOpts.WeylAlgebra else {};
     RSkew := if ROpts.?SkewCommutative then ROpts.SkewCommutative else {};
     MSkew := if MOpts.?SkewCommutative then MOpts.SkewCommutative else {};
+    if any(MSkew, i -> i > nvars) then error "SkewCommutative indices must be a subset of variable indices";
```
d-torrance: Should this move deeper into the constructor of the monoid `M`?  Also, I think `i == nvars` should probably also error, as well as `i < -nvars` (w/ negative values of `i` being interpreted as counting backwards from the end).

d-torrance (changes requested):

M2/Macaulay2/m2/polyrings.m2
joel-dodge: I felt ok about putting here alongside the validation of the skew and weyl options on the next line but yes I think it seems like it would be reasonable to place this somewhere deeper where `R` and `M` are constructed (I see now that here I have only validated `MSkew` but i think it is natural to similarly validate `RSkew` as well).

That negative indices are allowed is interesting and actually suggests that this interface is a bit of a hot mess. In addition to currently allowing passing any integers for indices of skew variables, one can *over*specify skewness by setting the same variable as skew with a positive and a negative index. eg. 

```
R = kk[a, SkewCommutative => {-1, 0}]
```

Macaulay2 thinks this is a `PolynomialRing, 2 skew commutative variable(s)` 

In light of all this some things I am not sure of:
1. Is it appropriate to manipulate user input to normalize it to a list of non-duplicated integers between 0 and nvars - 1? is there some design principle that compels us to preserve the config as it was passed to us?
2. What is the right place to do this validation / normalizing? I would have to sniff around to see where this `R` and `M` are constructed with their various options but maybe you know the right place already?

joel-dodge: oh and also i want to be very explicit that you absolutely caught me in an off by one error haha thank you!

d-torrance: 1.  I think normalizing makes a lot of sense.  I don't see a compelling reason to hang on to a bunch of duplicate/negative entries.
2. It looks like `processSkew` in `monoids.m2` would be the natural place to do this.

joel-dodge: I found the right place in monoids.m2 where skew variables are processed and added some (now correct) validation. I also added normalization of indices to always maintain them as a sorted list of non-negative integers with no duplicates. If this level of aggression in normalizing user input seems wrong to you I look forward to hearing about it!

joel-dodge: Oh haha i should have checked for a reply before picking this up this morning. I wrote my message above without seeing your reply and you would have pointed me to the place I found faster :D

M2/Macaulay2/m2/monoids.m2
```diff
@@ -494,11 +494,18 @@ processVars(Thing, Thing) := (x, xx) -> findSymbols {x}
 processVars(ZZ,    Thing) := (n, xx) -> makeVars(n, xx)
 processVars ZZ := x -> {x}
 
-processSkew := (n, skewvars) -> toList(
+processIndex = (n, i) -> (
+    if i < -n or i >= n then error "invalid variable index";
+    if i < 0 then n + i else i
+)
+
+processSkew := (n, skewvars) -> sort toList(
```
d-torrance: This `sort` is causing some issues when we mix variables and indices (which is currently legal, although weird lol):

```m2
i1 : QQ[a,b,c,d, SkewCommutative => {a, 1}]
stdio:1:2:(3):[1]: error: no method for binary operator ? applied to objects:
            1 (of class ZZ)
      ?     a (of class Symbol)
```

d-torrance (changes requested):

M2/Macaulay2/m2/monoids.m2
joel-dodge: Fascinating. Did this surface in a docs example or test or do you just know to test this quirky allowed configuration?

What would you think about normalizing to a list of nonnegative integers in every case regardless of what type of input is given?

d-torrance: Claude found it lol :robot:

Yeah, I think normalizing to nonnegative integers for everything makes a lot of sense

M2/Macaulay2/m2/monoids.m2
```diff
@@ -494,11 +494,22 @@ processVars(Thing, Thing) := (x, xx) -> findSymbols {x}
 processVars(ZZ,    Thing) := (n, xx) -> makeVars(n, xx)
 processVars ZZ := x -> {x}
 
-processSkew := (n, skewvars) -> toList(
+processItem = (vars, v) -> (
+    i := if instance(v, Symbol) then position(vars, s -> s == v ) else v;
```
d-torrance: We also have the `IndexedVariable` case to worry about:

```m2
i1 : errorDepth = 1;

i2 : R = QQ[x_0..x_4, SkewCommutative => {x_0, x_1}]
M2/Macaulay2/m2/monoids.m2:499:37:(1):[11]: error: SkewCommutative: invalid value `x_0`
M2/Macaulay2/m2/monoids.m2:499:37:(1): entering debugger (enter 'help' to see commands)
M2/Macaulay2/m2/monoids.m2:499:32-499:92: --source code:
    if not instance(i, ZZ) then error("SkewCommutative: invalid value `" | toString v | "`");

ii3 : i

oo3 = x
       0

oo3 : IndexedVariable
```

M2/Macaulay2/packages/Macaulay2Doc/functions/monoid-doc.m2
```diff
@@ -451,39 +451,45 @@ Node
...
+            c*b*a*d
+            ZZ/101[x,y,vars(0..4), SkewCommutative => {2..6}]
+            c*b*a*d
+        Example
+            R = ZZ[x,y,z, SkewCommutative => {x,y}]
+            x*y
+            y*x
+            x*z-z*x
+        Text
+            Integer and symbols can be mixed. You can specify positive or
```
d-torrance:
```suggestion
            Integers and symbols can be mixed. You can specify positive or
```

M2/Macaulay2/tests/normal/prune4.m2
```diff
@@ -0,0 +1,20 @@
...
+J = ideal {a^2 + 1, b, c, d}
+S' = prune(S/J)
+assert(isCommutative S')
+
+-- mix of skew and commuting variables reindexed correctly
+T = kk[a..d, SkewCommutative => {1, 2, 3}]
+K = ideal {c, d}
+T' = prune(T/K)
+assert(isSkewCommutative T')
+assert(T'.SkewCommutative == {1})
```
d-torrance:
```suggestion
assert(T'.SkewCommutative == {1})

```
Gotta have those trailing newlines!

d-torrance (changes requested):

M2/Macaulay2/m2/monoids.m2
joel-dodge: of course. fixing this now and adding more complete tests for monoid creation wrt the SkewCommutative configuration. ty for all your help!

d-torrance (approved):
Looks good to me!  Tentative approval pending discussing with @seangrate

seangrate (approved):
Looks good to me :+1:

Issue #3372, opened 2024-07-21
Using Intel's Math Kernel Library with M2


mahrud:
Here are the problems I ran into:

0. First, you need to set the compiler variables, e.g for MKL 2024.1:
```
. /opt/local/oneapi/2024.1/setvars.sh --force
 
:: initializing oneAPI environment ...
   -bash: BASH_VERSION = 5.2.15(1)-release
   args: Using "$@" for setvars.sh arguments: --force
:: advisor -- latest
:: compiler -- latest
:: debugger -- latest
:: dev-utilities -- latest
:: dpcpp-ct -- latest
:: dpl -- latest
:: ipp -- latest
:: ippcp -- latest
:: mkl -- latest
:: mpi -- latest
:: tbb -- latest
:: oneAPI environment initialized ::
```
Though finding this file was a bit tricky. Then `ninja build-givaro build-fflas-ffpack` detected MKL fine as well, and this was reported by cmake:
```
## Library information
     Linear Algebra    = /opt/local/oneapi/2024.1/mkl/2024.1/lib/libmkl_intel_lp64.so;/opt/local/oneapi/2024.1/mkl/2024.1/lib/libmkl_intel_thread.so;/opt/local/oneapi/2024.1/mkl/2024.1/lib/libmkl_core.so;/opt/local/oneapi/2024.1/compiler/2024.1/lib/libiomp5.so;-lm;-ldl;-lm;-ldl
```

1. Conflicting declaration of C functions from lapack.hpp, e.g.:
```cpp
/usr/people/sayrafi/M2/M2/Macaulay2/e/lapack.hpp:18:5: error: conflicting declaration of C function ‘int dgesv_(int*, int*, double*, int*, int*, double*, int*, int*)’
   18 | int dgesv_(int *n,      // number of rows in A
      |     ^~~~~~
compilation terminated due to -Wfatal-errors.
```
I got around this by adding `#ifndef _MKL_LAPACK_H_` around the `extern "C"` section of our `lapack.hpp`. It seems to me like we should be including this section from somewhere, why is it here?

2. Incompatible complex types:
```cpp
/usr/people/sayrafi/M2/M2/Macaulay2/e/dmat-lu-inplace.hpp: In member function ‘void DMatLUinPlace<RT>::computeLU() [with RT = M2::ARingCC]’:
[...]

PR #4659, opened 2026-08-21
Don't redefine cblas/lapack routines from MKL

d-torrance:
Also define `MKL_Complex16` as `double` before including `ffpack.h`.  If it was built with MKL support, then compilation will fail with a bunch of `cannot convert 'double*' to 'MKL_Complex16*'` errors.

After these changes, I was able to build M2 and FFLAS-FFPACK using MKL as their BLAS/LAPACK implementation.

Here are the steps to reproduce the issue.  In particular, we need FFLAS-FFPACK to pull in the MKL headers.  This is on Ubuntu 24.04 with `intel-mkl` installed:

```sh
sudo update-alternatives --config libblas.so-x86_64-linux-gnu
# select /usr/lib/x86_64-linux-gnu/libmkl_rt.so
cd M2/M2/BUILD/build
../../autogen.sh
../../configure CPPFLAGS="$(pkg-config --cflags blas)"
# need fflas-ffpack built w/ MKL support, and it only detects it using CBLAS_LIBS
make -C libraries/fflas_ffpack CBLAS_LIBS="$(pkg-config --libs blas)"
make
```

Closes: #3372

I did a bit of brainstorming with Claude, but the code is all mine.

### DIFF
--- a/M2/Macaulay2/e/basic-mutable-matrices/lapack.hpp
+++ b/M2/Macaulay2/e/basic-mutable-matrices/lapack.hpp
@@ -18,6 +18,7 @@ typedef DMat<M2::ARingCC> DMatCC;
 /* MES, On my mac, 10.12.4, lapack include file is at
   /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/Headers/clapack.h
 */
+#ifndef _MKL_LAPACK_H_
 extern "C" {
 int dgesv_(int *n,      // number of rows in A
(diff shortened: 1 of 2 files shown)


andrew-tawfeek (approved):
This looks good. I compiled the affected FFLAS and mutable-matrix translation units both normally and with FFLAS forced through the MKL path against Intel MKL headers; the header guards suppress the conflicting declarations and the complex type matches the existing double-based storage.

antonleykin (approved):

PR #4661, opened 2026-08-21
Create HilbertSchemeStrata.m2

foschreyer:
a package computing Hilbert scheme strata


d-torrance (changes requested):
Thanks for the contribution, Frank!

This needs an accompanying update to `=distributed-packages`.  You did this in #4662, but it should be pushed the same branch (`patch-4`) to update this PR.

To do this in the browser interface, select the branch in the upper-left hand corner before you edit:

Image

M2/Macaulay2/packages/HilbertSchemeStrata.m2
```diff
@@ -0,0 +1,760 @@
...
+	   not member(t,support tailTerm_(0,i))));
+    assert(#pos==r);
+    subst:=flatten apply(pos,i->{lTm_(0,i) => -tailTerm_(0,i)});
+    assert all(r,i-> sub(gJ*m1*m2*D_{i},subst_i)==0);
+    graph:=flatten apply(r,i->apply(
+	   select(r,j-> member(lTm_(0,i),support tailTerm_{j})),j->(i,j)));
+    isDirected :=isDirectedGraph graph;
+--graph
+--isDirected = true;
+    if isDirected then (
```
d-torrance, line 230: If `isDirected` is false then we just return null, e.g.:

```m2
i20 : R = QQ[x,y,z];

i21 : removeVariables ideal(x - y*z, y - x*z)

i22 : 
```

Should there be an error instead?

M2/Macaulay2/packages/HilbertSchemeStrata.m2
```diff
@@ -0,0 +1,760 @@
...
+    subst1 := if maxPathLength==0 then subst else (
+    	--tailTerm1:=tailTerm;
+        scan(maxPathLength,i->tailTerm=sub(tailTerm,subst));
+        flatten apply(r,i->{lTm_(0,i) => -tailTerm_(0,i)}));
+    J1 :=trim sub(J,subst1);
+    kk :=coefficientRing R;R1:=null;
+    h:=sub(vars R,subst1);
+    if numgens J1 == 0 then (R1=kk[support h]; return (ideal 0_R1,sub(h,R1)));
+    R1 =kk[support h];
+    return (sub(J1,R1),sub(h,R1)));
```
d-torrance, line 247: We don't need `return` in the last line of a function (if we remove `;`)

M2/Macaulay2/packages/HilbertSchemeStrata.m2
```diff
@@ -0,0 +1,760 @@
...
+    (ideal I0,unf1)    
+    )
+
+
+
+flatteningRelations=method()
+flatteningRelations(Ideal,Matrix,Ring) := (I0,unf,R) -> (
+    assert(
+	 rank source gens gb I0== rank source gens I0
+	 );
```
d-torrance: It would be good to raise informatiive error messages using `error` rather than just getting a cryptic `assertion failed` from `assert`.

```suggestion
    if rank source gens gb I0 != rank source gens I0
    then error "expected ideal to be generated by a minimal Gröbner basis";
```

Same comment for the several other places with `assert`.

M2/Macaulay2/packages/HilbertSchemeStrata.m2
```diff
@@ -0,0 +1,760 @@
...
+///
+
+doc ///
+  Key 
+   extraTerms
+   (extraTerms,Ideal,ZZ)
+   (extraTerms,Ideal,ZZ,Ideal)
+   (extraTerms,Ideal,ZZ,RingElement)
+  Headline 
+   computes the unfolding of the GB induced by an unfolding of the generators
```
d-torrance: I think this might be a copy/paste bug - it's the same as `passToGB`.

M2/Macaulay2/packages/HilbertSchemeStrata.m2
```diff
@@ -0,0 +1,760 @@
...
+///
+
+doc ///
+  Key 
+   substituteFamily
+   (substituteFamily,Matrix,Ring,Matrix)
+  Headline 
+   substitute a family of ideal using h
+  Usage
+    fam1 = substitute(fam,S,h)
```
d-torrance:
```suggestion
    fam1 = substituteFamily(fam,S,h)
```

M2/Macaulay2/packages/HilbertSchemeStrata.m2
```diff
@@ -0,0 +1,760 @@
...
+doc ///
+  Key 
+   unfolding
+   (unfolding,Ideal)
+   (unfolding,Ideal,Ring)
+   (unfolding,Ideal,Ideal)
+  Headline 
+   computes the unfolding of an ideal
+  Usage
+   (unf,R) = unfolding(I)
```
d-torrance:
```suggestion
   (unf,R) = unfolding(I0)
```

... to match the `Inputs` section below

M2/Macaulay2/packages/HilbertSchemeStrata.m2
```diff
@@ -0,0 +1,760 @@
...
+   (flatteningRelations,Ideal, Matrix, Ring)
+  Headline 
+    Compute the flattening relation of an unfolding in the coefficient ring R
+  Usage
+    base = flatteningRelations(I, unf, R)
+  Inputs
+    I: Ideal
+        generated by a GB of a subscheme of PP^n
+    unf: Matrix
+          of generators of an unfolding of I with coefficients in R
```
d-torrance, line 550: It would be good to include what `R` is here too

M2/Macaulay2/packages/HilbertSchemeStrata.m2
```diff
@@ -0,0 +1,760 @@
...
+    L2)		     
+
+    
+
+beginDocumentation()
+
+document {
+  Key => HilbertSchemeStrata,
+  Headline => "Computation of strata of an Hilbert scheme",
+  "This package contains the implementation of an Groebner basis approach
```
d-torrance, line 282:
```suggestion
  "This package contains the implementation of a Groebner basis approach 
```

M2/Macaulay2/packages/HilbertSchemeStrata.m2
```diff
@@ -0,0 +1,760 @@
...
+
+document {
+  Key => HilbertSchemeStrata,
+  Headline => "Computation of strata of an Hilbert scheme",
+  "This package contains the implementation of an Groebner basis approach 
+  to the Hilbert scheme. The basic set-up is to compute an unfolding of the 
+  generators of an ideal corresponding to a point in the Hilbert scheme, and then to compute the 
+  flattening relations via Buchberger tests. The main purpose of this package 
+  is to provide code for the computations of Section 13.5 of [S25]. 
+  A more complete package is profide by Nathan Ilten [N].",
```
d-torrance:
```suggestion
  A more complete package is provided by Nathan Ilten [N].", 
```

M2/Macaulay2/packages/HilbertSchemeStrata.m2
```diff
@@ -0,0 +1,760 @@
...
+  Inputs
+   L: List
+       a list of edges of a oriented graph
+  Outputs
+   b: Boolean
+      true if the graph is directed
+  Description
+    Text
+      The list L is list of edges of an oriented graph.
+      The functions checks whether the graph is dericted, i.e., contains no loops.
```
d-torrance:
```suggestion
      The function checks whether the graph is directed, i.e., contains no loops.
```

M2/Macaulay2/packages/HilbertSchemeStrata.m2
```diff
@@ -0,0 +1,760 @@
...
+   (substituteFamily,Matrix,Ring,Matrix)
+  Headline 
+   substitute a family of ideal using h
+  Usage
+    fam1 = substitute(fam,S,h)
+  Inputs
+    fam: Matrix
+        of generators of ideal depending on some parameter in R
+    S: Ring
+       the embient ring of of the special fiber
```
d-torrance:
```suggestion
       the ambient ring of of the special fiber
```

M2/Macaulay2/packages/HilbertSchemeStrata.m2
```diff
@@ -0,0 +1,760 @@
...
+    h: Matrix 
+       describing a substitution R to ring h   	 
+  Outputs
+    fam1: Matrix
+        describing the generators of the pullback family
+  Description
+    Text
+      Given a family fam over Spec R, the function returns the 
+      pullback family under the map R to ring h.
+      We illustate this with the Piene-Schlessinger example.
```
d-torrance:
```suggestion
      We illustrate this with the Piene-Schlessinger example.     
```

M2/Macaulay2/packages/HilbertSchemeStrata.m2
```diff
@@ -0,0 +1,760 @@
...
+
+extraTerms(Ideal,ZZ,Ideal) := (I,d,qperp) ->(
+    q1:=ideal image basis(d,module qperp);
+    mingens trim ideal (gens q1%I))
+
+
+unfolding=method()
+unfolding(Ideal,Ring) := (I,R) -> (
+    kk:= coefficientRing R;
+    lTs:=apply(numgens I,i-> leadTerm I_i);
```
d-torrance, line 73: According to the docs, we're supposed to use the monomial order from `R`, but here we're using the one from `ring I`.

Same issue with the call to `extraTerms` on the next line.

Which should be fixed -- the code or the docs?

d-torrance (changes requested):
Thanks, Frank!

I have a few comments

M2/Macaulay2/packages/HilbertSchemeStrata.m2
foschreyer: I prefer null because an error interrupts the computation. It is possible that isDirected is false but one would like to continue otherwise.

M2/Macaulay2/packages/HilbertSchemeStrata.m2
foschreyer: I sometimes like to use return for emphasis and syntax high lightening

M2/Macaulay2/packages/HilbertSchemeStrata.m2
foschreyer: correct is
Headline
  compute the extra terms to be used in building the unfolding

M2/Macaulay2/packages/HilbertSchemeStrata.m2
foschreyer: the code should be fixed., which means that the ideal has to be moved to R.
It is possible that implicitly I assumed R== ring I

foschreyer:
Should I do the changes in unfolding?

d-torrance:
> Should I do the changes in unfolding?

Yes, that would be great!

andrew-tawfeek (changes requested):
In `unfolding(Ideal, Ring)`, the leading terms are currently computed in `ring I` and then substituted into `R`:

```m2
lTs := apply(numgens I, i -> leadTerm I_i);
lTs1 := apply(lTs, t -> sub(t, R));
```

If the two rings have different monomial orders, a generator can have a different leading term in each ring. For example, with variables ordered `x > y > z`, the polynomial `x*z + y^2` has leading term `y^2` under graded reverse lexicographic order, but `x*z` under lexicographic order. Substituting just `y^2` into the second ring cannot recover `x*z`, because the other term has already been discarded.

Since the ring argument specifies the monomial order for the unfolding, could you substitute each full generator into `R` **before** computing its leading term, and use those leading terms throughout the construction?

PR #4663, opened 2026-08-22
A few autotools updates for the packages directory

d-torrance:
Some related changes:

* Scrap the redundant `DEVEL` variable, which was used for `EngineTests` when it wasn't distributed.  But now it is.
* Use `:` instead of `::` for the `check` target.  This enables us to run the package tests in parallel, which should speed up the `autotools-ubuntu` build considerably.
* Only run tests for packages in the `PACKAGES` variable, e.g., `make check PACKAGES=Foo` now only runs the checks for the *Foo* package, where it used to run *all* the tests.  This brings it in line with `make PACKAGES=Foo`, which already only built the documentation for *Foo*.

I wrote all the code, but the recommendations were all from Claude after it did a review of our GitHub workflow.

### DIFF
--- a/M2/Macaulay2/packages/Makefile.in
+++ b/M2/Macaulay2/packages/Makefile.in
@@ -46,7 +46,7 @@ m2-need-template = "needsPackage(\"$(1)\",LoadDocumentation=>true,DebuggingMode=
 m2-check-template = "check($(1),UserMode=>false,Verbose=>$(Verbose)); exit 0"
 $(foreach i,\
 	$(sort $(ALL_PACKAGES)),\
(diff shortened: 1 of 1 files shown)


d-torrance:
The results are in!  

The `autotools-ubuntu-24.04-default` workflow finished in [143m](https://github.com/Macaulay2/M2/actions/runs/32600376381/job/97097635137?pr=4663), while the same workflow from #4661 finished in [190m](https://github.com/Macaulay2/M2/actions/runs/32483009201/job/96773397693?pr=4661).  Not bad!

dimpase (approved):
lgtm

PR #4664, opened 2026-08-23
Update to use of ccache in GitHub workflow

d-torrance:
Previously, we tried to cache our [ccache](https://ccache.dev/) files, but we were caching the wrong directory!

We simplify things by using an [existing action](https://github.com/hendrikmuhs/ccache-action) to handle caching for us. We now use it for both the autotools and cmake builds (previously just cmake).

We now trigger a new set of builds when a PR is merged into `development` so we can save the cache. (If the PR build saved it, then other PR's wouldn't be able to access it.)  There's no need to build documentation or run the tests when doing this, so we set a `SEED_CACHE_ONLY` variable when we do this and use it to skip a bunch of steps

We also stop caching the `usr-host` directory, which was mostly pointless as we primarily use system packages for our dependencies.

Here's a test run:

* Builds that seeded the cache: [9m 23s](https://github.com/d-torrance/M2/actions/runs/32660322645)
* Builds that restored and used the cache: [5m 2s](https://github.com/d-torrance/M2/actions/runs/32660996677)

Claude wrote the code, but I went over every single line and we ended up making some substantial edits.  I also dropped all its extraneous comments lol.  I also didn't let it commit and wrote the commit message myself.

### DIFF
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -14,6 +14,9 @@ on:
     branches:
       - stable
       - development
+  push:
+    branches:
+      - development # seed cache after merging a PR
   schedule:
     - cron: '0 6 * * SUN' # runs tests on the main branch every Sunday at 06:00 UTC
 
@@ -25,6 +28,10 @@ concurrency:
 defaults:
   run:
     working-directory: M2/BUILD/build
+
+env:
+  # skip documentation & tests when seeding cache
+  SEED_CACHE_ONLY: ${{ github.event_name == 'push' }}
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


dimpase:
The idea is to speed up building across different CI runs, right?
Where is the cache being saved? 

One way or another, it's assuming that you get allocated the same hardware across different jobs (otherwise you might get weird errors  about a wrong CPU command etc). How is this guaranteed?

d-torrance:
> The idea is to speed up building across different CI runs, right? Where is the cache being saved?

Yes, exactly.

You can see the caches from a repo's worfklow under the Actions tab, on the left under "Management".

Here's ours: https://github.com/Macaulay2/M2/actions/caches

Here's the cache on my fork from testing this PR: https://github.com/d-torrance/M2/actions/caches

> One way or another, it's assuming that you get allocated the same hardware across different jobs (otherwise you might get weird errors about a wrong CPU command etc). How is this guaranteed?

I suppose, yeah.  If that were to happen though, it looks we'd get a cache miss from ccache (see https://github.com/ccache/ccache/discussions/725), and just compile as usual.

[ccache-action](https://github.com/hendrikmuhs/ccache-action) is used by more than 10k repos, so I'm guessing this hasn't been an issue.

dimpase:
> I suppose, yeah. If that were to happen though, it looks we'd get a cache miss from ccache (see https://github.com/ccache/ccache/discussions/725), and just compile as usual.

that's a different story - what I meant is the scenario when you deploy the same compiler on slightly different CPUs: the compiler knows what commands to emit for each CPU, and potentially a cached result satisfies the constraints to be used, but then you get a runtime error as your resulting executable has an invalid for the CPU command...

Or perhaps it's not happening as what's cached is higher level stuff...

dimpase (approved):
OK

andrew-tawfeek (approved):
This looks good. The cache key separates build system, OS, and compiler; PR jobs restore without saving, while development pushes seed fresh caches and skip only package installation, tests, validation, and artifact upload after the core build. The action inputs match v1.2.23 and all current workflow checks pass.

PR #4666, opened 2026-08-25
Some  mutableMatrix updates

d-torrance:
A few semi-related things that all grew out of fixing the first thing:

## Rank errors

According to `mutable-matrix.h`:

```c++
/**
   returns the rank of the matrix M.  If 'rank' is not defined on this type of
   matrix, then returns -1 (and an error message is given).
 */
long rawLinAlgRank(MutableMatrix *M);
```

However, no such error message is given, and we just get -1!

```m2
i1 : rank mutableMatrix(ZZ/101, {{1}}, Dense => false)

o1 = -1
```

So I added a little check in the interpreter -- if we get -1, return the last engine error instead, so now we have:

```m2
i1 : rank mutableMatrix(ZZ/101, {{1}}, Dense => false)
stdio:1:4:(3):[1]: error: 'rank' not implemented for this kind of matrix over this ring
```

## Constructor methods

While testing the above, I was sad that I had to type `{{` and `}}`' when constructing a $1\times 1$ mutable matrix and wanted the `Number`/`RingElement` constructors we have for immutable matrices.  So I'm proposing 6 new methods:

```m2
i1 : mutableMatrix 2 -- mutableMatrix(Number)

o1 = | 2 |

o1 : MutableMatrix

i2 : mutableMatrix(QQ, 2) -- mutableMatrix(Ring, Number)

o2 = | 2 |

o2 : MutableMatrix

i3 : mutableMatrix(RR, 2) -- mutableMatrix(RingFamily, Number)

o3 = | 2 |

o3 : MutableMatrix

i4 : R = QQ[x]; mutableMatrix x -- mutableMatrix(RingElement)

o5 = | x |

o5 : MutableMatrix

i6 : mutableMatrix(R[y], x) -- mutableMatrix(Ring, RingElement)

o6 = | x |

o6 : MutableMatrix
```

There's also `mutableMatrix(RingFamily, RingElement)`, which won't do anything now except raise errors about promoting.  But it might work in the future (e.g., after adding $p$-adics to the engine).

## Documentation updates

When I went to document the new constructor methods, I saw that a bunch of the mutable matrix docs were in the old format and in the poorly named `doc3.m2` file.  So I moved them and had Claude update them to *SimpleDoc* format.

## `AfterPrint` method

While writing this pull request description, I realized that mutable matrices don't tell us the size or what ring we're working over like we get for immutable matrices.  That's a great use for `AfterPrint` methods so we just use the `Matrix` one:

```m2
i1 : mutableMatrix(QQ[x,y,z,w], 6, 7)

o1 = 0

                                6                  7
o1 : MutableMatrix (QQ[x..z, w])  <-- (QQ[x..z, w])
```

I wrote the actual d and m2 code, but got lazy and had Claude do the boring stuff (documentation and tests) lol.

### DIFF
--- a/M2/Macaulay2/tests/normal/mutmat.m2
+++ b/M2/Macaulay2/tests/normal/mutmat.m2
@@ -83,3 +83,11 @@ assert Equation(M^{-1}, mutableMatrix {{4, 5, 6}})
 A = mutableMatrix {{10}}
 B = submatrix(A, {0}, {0})
 assert(hash A != hash B)
+
+-- 1 by 1 constructors
+S = QQ[t]
+assert Equation(mutableMatrix 5, mutableMatrix {{5}})
+assert Equation(mutableMatrix t, mutableMatrix {{t}})
+assert Equation(mutableMatrix(QQ, 5), mutableMatrix {{5/1}})
+assert Equation(mutableMatrix(S, t), mutableMatrix {{t}})
+assert Equation(mutableMatrix(RR, 5), mutableMatrix {{5.}})
--- a/M2/Macaulay2/d/interface.dd
+++ b/M2/Macaulay2/d/interface.dd
@@ -2927,7 +2927,10 @@ setupfun("rawMutableMatrixFillRandom",rawMutableMatrixFillRandom);
 export rawLinAlgRank(e:Expr):Expr := (
      when e
      is m:RawMutableMatrixCell
-     do toExpr(Ccode(long, "rawLinAlgRank(", m.p, ")" ))
+     do (
+         r := Ccode(long, "rawLinAlgRank(", m.p, ")" );
+         if r == long(-1) then engineErrorMessage()
+         else toExpr(r))
      else WrongArgMutableMatrix());
 setupfun("rawLinAlgRank",rawLinAlgRank);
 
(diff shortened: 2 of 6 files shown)


M2/Macaulay2/tests/normal/mutmat.m2
```diff
@@ -83,3 +83,11 @@ assert Equation(M^{-1}, mutableMatrix {{4, 5, 6}})
 A = mutableMatrix {{10}}
 B = submatrix(A, {0}, {0})
 assert(hash A != hash B)
+
+-- 1 by 1 constructors
+S = QQ[t]
+assert Equation(mutableMatrix 5, mutableMatrix {{5}})
```
seangrate, line 89: While the use of `Equation` might be valid, I feel like `==` or `===` is probably more standard.

seangrate (approved):
Overall looks good from what I can tell. One small comment on the format of tests.

M2/Macaulay2/tests/normal/mutmat.m2
d-torrance: I use `assert(Expression)` almost exclusively.  It's nice for figuring out what happened when tests fail.  Compare the following:

```m2
i1 : assert(3 == 4)
stdio:1:6:(3):[1]: error: assertion failed

i2 : assert Equation(3, 4)
stdio:2:6:(3):[1]: error: assertion failed:
3 == 4 is false
```

seangrate: Nifty, I may have to start using that now.

andrew-tawfeek (approved):
This looks good. The -1 sentinel from rawLinAlgRank is now translated through the existing engine-error path, the new 1-by-1 constructors follow the immutable-matrix promotion pattern, and the constructor/rank cases are covered by tests. The current Linux/macOS autotools and CMake checks pass.

PR #4667, opened 2026-08-25
Remove some unused code from Core

d-torrance:
We remove a few unused method functions that aren't exported and never had any methods installed, as well as some unused key/value pairs in ring types that are created in `initializeEngineLinearAlgebra`.

Claude pointed these out while I was looking at a slightly related issue, but I did all the removing myself.


klee669 (approved):
Looks good to me! Thanks for the cleanup!

seangrate (approved):

PR #4670, opened 2026-08-27
Fix #732

### LINKED ISSUE #732: Odd behavior of gb / poincare

pzinn:
The recent Claude comments prompted me to revisit #732. The fix is actually quite easy, it's one line in the c++ code that makes sure the GB computation remains in the correct state.
That being said, adding a test in `gb-hilbert.m2` and trying to run it led to a massive headache that only codex could help me out of: first `gb-hilbert.m2` started with `restart` which means `ctest` never actually ran it; second, after removing `restart`, a test failed because it was based on some given value of a random seed (and when I run it with `ctest` I get 36 rather than 37...) I replaced it with a more reasonable test. Finally, a test had the comment `internal error` which is a no no for `ctest`, so that needed rewriting as well. Phew.
### AI DISCLOSURE
see above

### DIFF
--- a/M2/Macaulay2/e/groebner-computations/gb-default.cpp
+++ b/M2/Macaulay2/e/groebner-computations/gb-default.cpp
@@ -2671,8 +2671,6 @@ Computation /* or null */ *gbA::set_hilbert_function(const RingElement *hf)
 {
   // TODO Problems here:
   //  -- check that the ring is correct
-  //  -- if the computation has already been started, this will fail
-  //     So probably an error should be given, and 0 returned in this case.
 
   // We may only use the Hilbert function if syzygies are not being collected
   // since otherwise we will miss syzygies
@@ -2690,7 +2688,10 @@ Computation /* or null */ *gbA::set_hilbert_function(const RingElement *hf)
       hf_diff = RingElement::make_raw(hf->get_ring(), ZERO_RINGELEM);
       use_hilb = true;
       hilb_new_elems = true;
-      state = STATE_HILB;
+      // A computation stopped mid-degree must first finish auto-reduction and
+      // create pairs for its new elements.  Those phases naturally lead to
+      // STATE_HILB; only a computation at a degree boundary must be redirected.
+      if (state == STATE_NEWDEGREE) state = STATE_HILB;
     }
 
   return this;
--- a/M2/Macaulay2/tests/normal/gb-hilbert.m2
+++ b/M2/Macaulay2/tests/normal/gb-hilbert.m2
@@ -48,3 +48,11 @@ f = x^4 + y^4 + z^4;
 R = S / ideal(f);
 I = ideal(x^3, x^2+y^2);
 integralClosure(I) -- crashed, now ok.
+
+-- git issue #732: resume a min-gens computation with a Hilbert hint
+R = QQ[a..k]
+I = trim ideal(a*b+c*d, a*e+b*f+c*g+d*h)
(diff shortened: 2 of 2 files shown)


d-torrance (approved):
Looks good to me -- thanks!

M2/Macaulay2/e/groebner-computations/gb-default.cpp
```diff
@@ -2690,7 +2688,10 @@ Computation /* or null */ *gbA::set_hilbert_function(const RingElement *hf)
       hf_diff = RingElement::make_raw(hf->get_ring(), ZERO_RINGELEM);
       use_hilb = true;
       hilb_new_elems = true;
-      state = STATE_HILB;
+      // A computation stopped mid-degree must first finish auto-reduction and
+      // create pairs for its new elements.  Those phases naturally lead to
+      // STATE_HILB; only a computation at a degree boundary must be redirected.
```
mikestillman, line 2693: I don't understand this comment.  Can you rephrase it?

pzinn: yeah sorry I had codex format the PR and it added an incomprehensible comment. What's going on is pretty simple: it's that `StopWithMinimalGenerators` may stop half-way through processing a given degree. This means some S-pairs may be missing, and if we go straight to STATE_HILB it will conclude (possibly incorrectly) that the GB is finished. Instead we only go straight to STATE_HILB if we're done with the given degree. I'll try to rephrase.

PR #4674, opened 2026-08-31
Integrating Google Benchmark into Macaulay2

andrew-tawfeek:
This is a 🤖 attempt to bring in [Google Benchmark ](https://github.com/google/benchmark) into Macaulay2. I worked alongside Codex and primarily tried to mimic the installation of GoogleTest.

Nothing else is purposely apart of the PR -- comments welcome.

Further notes:
- I've added documentation on the Benchmark suite in BENCHMARK.md at the top level. I know this is a bad spot to have this, so suggested movement on these (AI-gen) guidelines is welcome. Happy to rewrite after people explore.
- Google Benchmark version here is v1.9.5
- Automake and CMake builds passed in Docker
- Default is benchmarking disabled unless a flag is specified -- should be usual M2 without flag
- Generated artifacts from benchmarking are ignored in git

Comments welcome. 

(This benchmarking setup is what will be later used to test the garbage collector's impact on Macaulay2.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,8 @@
 /M2/.idea
 /M2/tmp
 /builds
+/build/
+/benchmark-results/
 *~
 .DS_Store
 .#*
--- a/.gitmodules
+++ b/.gitmodules
@@ -28,3 +28,6 @@
 [submodule "M2/Macaulay2/editors/emacs"]
 	path = M2/Macaulay2/editors/emacs
 	url = https://github.com/Macaulay2/M2-emacs.git
+[submodule "M2/submodules/benchmark"]
+	path = M2/submodules/benchmark
+	url = https://github.com/google/benchmark.git
--- a/BENCHMARK.md
+++ b/BENCHMARK.md
@@ -1,0 +1,169 @@
+# Benchmarking Macaulay2
+
+Macaulay2 uses [Google Benchmark](https://github.com/google/benchmark) for
+repeatable C++ microbenchmarks of engine code. The integration intentionally
+parallels the existing GoogleTest setup: a pinned submodule supplies a fallback
+dependency, a small wrapper initializes the M2 engine, and benchmark sources
+live beside the engine unit tests.
+
+Benchmarking is opt-in. It is not part of `make check` or `ctest`, because
+performance measurements are sensitive to build flags, hardware, and machine
+load and should not determine whether correctness tests pass.
+
+## Quick start with CMake
+
+Use a release build for meaningful timings. From the repository root:
+
+```sh
+cmake -S M2 -B build -GNinja \
+  -DCMAKE_BUILD_TYPE=Release \
+  -DBUILD_BENCHMARKS=ON
+cmake --build build --target build-libraries
+cmake --build build --target M2-benchmarks
+./build/Macaulay2/e/M2-benchmarks
+```
+
+The `build-libraries` step installs any missing M2 dependencies, including the
+pinned Google Benchmark fallback, into `build/usr-host` and reruns CMake. If all
+dependencies are already available on the system, that step is harmless.
+
+Useful runner options include:
+
+```sh
+# List benchmark names without running them.
+./build/Macaulay2/e/M2-benchmarks --benchmark_list_tests
+
+# Run only subset benchmarks.
+./build/Macaulay2/e/M2-benchmarks '--benchmark_filter=BM_Subset.*'
+
+# Repeat runs and report aggregate statistics.
+./build/Macaulay2/e/M2-benchmarks \
+  --benchmark_repetitions=10 \
+  --benchmark_report_aggregates_only=true
+
+# Save machine-readable results.
+mkdir -p benchmark-results
+./build/Macaulay2/e/M2-benchmarks \
+  --benchmark_out=benchmark-results/results.json \
+  --benchmark_out_format=json
+```
+
+See the upstream [Google Benchmark user
+guide](https://google.github.io/benchmark/user_guide.html) for all runner
+options, fixtures, counters, and timing controls.
+
+## Quick start with Autotools
+
+Add `--enable-benchmarks` when configuring an out-of-tree M2 build. For
+example, from an empty build directory:
+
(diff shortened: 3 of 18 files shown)


d-torrance:
Is there a strong reason to include support for building benchmark from source?  It's available via both `apt` and `brew`, and it's not something that most folks would need anyway.  Someone wanting to build M2 w/ benchmarking support on a system that does't have a readily-available benchmark package certainly could build it themselves.

If the answer to the above question is "yes", then is there a strong reason to use a submodule rather than a tarball for getting the benchmark source?

andrew-tawfeek:
> Is there a strong reason to include support for building benchmark from source? It's available via both `apt` and `brew`, and it's not something that most folks would need anyway. Someone wanting to build M2 w/ benchmarking support on a system that does't have a readily-available benchmark package certainly could build it themselves.

Most of my motivation came from seeing its similarity to Google Test and how the Benchmark suite could be installed in the same way -- making them easy to maintain together. I think it's nice that this also fixes the version everyone uses (though "everyone" here would just mean people working on the engine -- who all can communicate the version to eachother in any case). Since the flag being omitted doesn't download or build anything, it feels like little overhead.

Note, similar to unit tests, there's a `/e/benchmarks/` directory now where explicit benchmark cases would go -- but this doesn't change that it's easy to `apt`/`brew` install then do `find_package(benchmark 1.9.5)` it looks like.

> If the answer to the above question is "yes", then is there a strong reason to use a submodule rather than a tarball for getting the benchmark source?

I've only used submodules in the past for similar things so I had to do some reading around: so if the proposal is to replace having this as a git submodule and instead just point directly to v1.9.5.tar.gz in google/benchmark, then I agree completely that that's easier to deal with and could make that change.

All-in-all, I'm happy to defer an opinion on this: my main motivation of including it was to match Google Tests implementation.

d-torrance:
I'm generally anti-vendoring and when we must vendor, anti-submodules. I'm sure my opinion is skewed by my Debian work, where we de-vendor everything and use tarballs for everything lol

andrew-tawfeek:
I didn't think about this before: but we could alternatively build our own benchmarking toolkit?

andrew-tawfeek:
Removing the vendoring here and just going to keep the benchmark directory and toss in a README on usage instructions

andrew-tawfeek:
Updated this PR to remove Google Benchmark vendoring while keeping the engine benchmark suite and its opt-in build targets.

- Removed the Google Benchmark submodule and the CMake/Autotools fallback builds. Enabling benchmarks now requires an external installation; configuration fails if it is missing. CMake no longer requires version 1.9.5 specifically, so packaged versions can be used.
- Moved the guide from `BENCHMARK.md` to `M2/Macaulay2/e/benchmarks/README.md` and updated the root README link. The guide covers installation with apt/Homebrew or upstream instructions, custom installation prefixes, CMake and Autotools builds, runner options, and adding benchmarks.
- Benchmarking remains disabled by default.

Validation: CMake Release build and benchmark smoke runs with an external Google Benchmark installation; full Autotools configuration; focused dependency checks for both build systems with benchmarks disabled, the dependency installed, and the dependency missing. A full Autotools build was not completed: the fresh build tree needs its normal engine dependencies (mathicgb headers) installed first.

M2/Macaulay2/e/benchmarks/README.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Quick start with CMake
+
+Use a release build for meaningful timings. From the repository root:
+
+```sh
+cmake -S M2 -B build -GNinja \
+  -DCMAKE_BUILD_TYPE=Release \
+  -DBUILD_BENCHMARKS=ON
+cmake --build build --target build-libraries
```
dimpase: what are these `--build build` here? Do you assume people normally invoke `cmake` in `M2/BUILD/` ?

The instructions ought to be consistent with other `cmake` instructions in other READMEs

dimpase:
ought to be tagged with 🤖

dimpase (changes requested):
it would be good to know how to read/interpret these
```
...
BM_ExponentVectorMultiply/4_mean            2.24 ns         2.23 ns           10 items_per_second=1.79035G/s
BM_ExponentVectorMultiply/4_median          2.23 ns         2.22 ns           10 items_per_second=1.79889G/s
BM_ExponentVectorMultiply/4_stddev         0.027 ns        0.027 ns           10 items_per_second=21.2448M/s
BM_ExponentVectorMultiply/4_cv              1.21 %          1.21 %            10 items_per_second=1.19%
...
```

M2/Macaulay2/e/benchmarks/README.md
andrew-tawfeek: Updated the benchmark README locally to follow the existing CMake guide: run from `M2/BUILD/build`, use `-S ../.. -B .` and `cmake --build .`, and adjust the executable paths accordingly. Pushed!

andrew-tawfeek:
> it would be good to know how to read/interpret these
> 
> ```
> ...
> BM_ExponentVectorMultiply/4_mean            2.24 ns         2.23 ns           10 items_per_second=1.79035G/s
> BM_ExponentVectorMultiply/4_median          2.23 ns         2.22 ns           10 items_per_second=1.79889G/s
> BM_ExponentVectorMultiply/4_stddev         0.027 ns        0.027 ns           10 items_per_second=21.2448M/s
> BM_ExponentVectorMultiply/4_cv              1.21 %          1.21 %            10 items_per_second=1.19%
> ...
> ```

Can you be a bit more specific on what I should try and specify here?

dimpase:
Benchmarking is about comparing different things, e.g. one can compare ZZ implementations (flint.vs non/flint, for instance). I just don't understand what being compared to what there.

dimpase:
> Benchmarking is about comparing different things, e.g. one can compare ZZ implementations (flint.vs non/flint, for instance). I just don't understand what being compared to what there.

I still want this addressed in docs or wiki

dimpase (changes requested):
see above

PR #4675, opened 2026-08-31
Fix MonomialSize docs

d-torrance:
We mentioned in two places that the monomial size can be 64.  It can't -- 32 is the largest.

```m2
i1 : R = QQ[x, MonomialSize => 2^100]

o1 = R

o1 : PolynomialRing

i2 : degreesMonoid R

o2 = monoid[T, Degrees => {1}, MonomialOrder => {MonomialSize => 32}, Inverses => true, Global => false]
                                                {Weights => {-1}   }
                                                {GroupLex => 1     }
                                                {Position => Up    }

o2 : GeneralOrderedMonoid
```

Skipping the builds since this is a doc-only change.


klee669 (approved):
Looks good to me! Thanks for the catch!

seangrate (approved):

PR #4679, opened 2026-09-03
Adding the new package KoszulFiltrations

emily-berghofer:
KoszulFiltrations is a package for working with Koszul filtrations. 
      The package aids in finding them, confirming that a set of ideals forms one, and finding linear flags. 
      The package is largely based on algorithms from: "Berghofer, Nicklasson, Thompson and Westerbäck, Constructing Koszul filtrations: existence and non-existence for G-quadratic algebras, arXiv:2602.06490."

### DIFF
--- a/M2/Macaulay2/packages/KoszulFiltrations.m2
+++ b/M2/Macaulay2/packages/KoszulFiltrations.m2
@@ -1,0 +1,728 @@
+newPackage(
+    "KoszulFiltrations",
+    Version => "1.0",  
+    Date => "August, 2026",
+    Authors => {
+	{Name => "Emily Berghofer", Email => "emily.berghofer@mdu.se",
+	    HomePage=>"https://emily-berghofer.github.io"},
+	{Name => "Lisa Nicklasson", Email => "lisa.nicklasson@mdu.se",
+	    HomePage=>"https://www.mdu.se/en/malardalen-university/staff?id=lisa.nicklasson"},
+	{Name => "Peder Thompson", Email => "peder.thompson@mdu.se",
+	    HomePage=>"https://sites.google.com/view/pederthompson"},
+	{Name => "Thomas Westerbäck", Email => "thomas.westerback@mdu.se",
+	    HomePage=>"https://www.mdu.se/en/malardalen-university/staff?id=thomas.westerback"}},
+    Headline => "Computational tools for Koszul filtrations",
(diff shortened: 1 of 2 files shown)


M2/Macaulay2/packages/KoszulFiltrations.m2
```diff
@@ -0,0 +1,728 @@
...
+	{Name => "Emily Berghofer", Email => "emily.berghofer@mdu.se",
+	    HomePage=>"https://emily-berghofer.github.io"},
+	{Name => "Lisa Nicklasson", Email => "lisa.nicklasson@mdu.se",
+	    HomePage=>"https://www.mdu.se/en/malardalen-university/staff?id=lisa.nicklasson"},
+	{Name => "Peder Thompson", Email => "peder.thompson@mdu.se",
+	    HomePage=>"https://sites.google.com/view/pederthompson"},
+	{Name => "Thomas Westerbäck", Email => "thomas.westerback@mdu.se",
+	    HomePage=>"https://www.mdu.se/en/malardalen-university/staff?id=thomas.westerback"}},
+    Headline => "Computational tools for Koszul filtrations",
+    Keywords => {"Koszul algebra","Koszul filtration","Linear filtration","Linear flag"},
```
d-torrance: The `Keywords` key here is poorly named, and it really should be something more like `Subject`.  It's used for generating the [packages provided with Macaulay2](https://www.macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_packages_spprovided_spwith_sp__Macaulay2.html) documentation page, and each package appears under the heading (or headings) specified by its `Keywords` field.

Packages should usually only have one or two keywords, and they almost always should use an existing one unless there's really no existing keyword that fits.

d-torrance (changes requested):
Thanks for the contribution!

Note that pull requests should target the `development` branch instead of `stable`.  I've already fixed it for you, but just a note for the future.

I made a quick comment about the `Keywords` field.  A more thorough review will follow later.

M2/Macaulay2/packages/KoszulFiltrations.m2
emily-berghofer: Oh sorry then it should be:
```suggestion
    Keywords => {"Commutative Algebra"},
```

PR #4682, opened 2026-09-09
Fix check for GC_start_performance_measurement

d-torrance:
It's a function, not a macro, so this block was never firing, and so `time` was always reporting "0s (gc)"

This fixes the issue discussed at this afternoon's C++ refactoring meeting

## Before
```m2
i1 : scan(10000, i -> {})

i2 : time collectGarbage()
 -- used 0.105673s (cpu); 0.0149833s (thread); 0s (gc)
```

## After
```m2
i1 : scan(10000, i -> {})

i2 : time collectGarbage()
 -- used 0.111014s (cpu); 0.0133115s (thread); 0.021s (gc)
```

### DIFF
--- a/M2/Macaulay2/bin/main.cpp
+++ b/M2/Macaulay2/bin/main.cpp
@@ -55,7 +55,7 @@ void* testFunc(ArgCell* p);
 void  M2_flint_abort(void);
 
 static void * GC_start_performance_measurement_0(void *) {
-#ifdef GC_start_performance_measurement /* added in bdwgc 8 */
+#if GC_VERSION_MAJOR >= 8 /* GC_start_performance_measurement added in bdwgc 8 */
   GC_start_performance_measurement();
 #endif
   return NULL;


MichaelABurr (approved):
This seems reasonable, but do we want to add a flag to decide whether this information is collected?

d-torrance:
> This seems reasonable, but do we want to add a flag to decide whether this information is collected?

What do you mean?

andrew-tawfeek (approved):

MichaelABurr:
> > This seems reasonable, but do we want to add a flag to decide whether this information is collected?
> 
> What do you mean?

Coming back to this, I think that it doesn't matter so much.  I was thinking that it might not be interesting to have gc info for most users when using elapsedTime.  I was also worried that there might be a performance hit when recording the gc info.  However, I doubt that there's any real performance hit as elapsedTime should only be used for diagnostic purposes and incrementing an integer shouldn't be so expensive.

PR #4685, opened 2026-09-10
Increase virtual memory limit for test failure in ComputationsBook::v…

shankerwangmiao:
When compiled on a host with many number of processors, the test for `ComputationsBook::varieties` may fail with:

```
testing: ComputationsBook::varieties
make[6]: *** [../Makefile.chapter:46: test.out] Error 134
make[5]: *** [Makefile:10: check] Error 2
```

and the content in the corresponding `test.out.tmp` is:

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  pthread_create has failed: Resource temporarily unavailable
Aborted
```

Increasing the limit of the virtual memory would solve the problem.

### DIFF
--- a/M2/Macaulay2/tests/ComputationsBook/Makefile.chapter.in
+++ b/M2/Macaulay2/tests/ComputationsBook/Makefile.chapter.in
@@ -18,7 +18,7 @@ ifeq (@ULIMIT_T@,yes)
 LIMIT += ulimit -t 1200 ;
 endif
 ifeq (@ULIMIT_V@,yes)
-LIMIT += ulimit -v 1600000 ;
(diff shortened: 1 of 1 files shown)


d-torrance (approved):
Thanks for the fix!

As I dug into this, I found a number of related issues that we should probably fix (not asking you to! Just making a note that at some point someone -- probably me lol -- should):

* We don't handle ulimit overrides in the ComputationsBook test Makefiles like we do everywhere else
* We aren't catching TBB errors like this. They're `runtime_error`'s, not `engine_errors`'s, so M2 crashes rather than just raising an error at top level.

dimpase (approved):
lgtm

PR #4689, opened 2026-09-11
Make ulimit overrides consistent in the autotools build

### LINKED ISSUE #4687: Handle ulimit overrides consistently in Makefiles
We don't handle ulimit overrides in the ComputationsBook test Makefiles like we do everywhere else

_Originally posted by @d-torrance in https://github.com/Macaulay2/M2/pull/4685#pullrequestreview-5171622084_

d-torrance:
We set various `ulimit`'s when running tests in the autotools build.  We weren't consistent in how we did this (#4687), so we factor everything out into `M2/include/config.Makefile.in` so that we can do any of the following:

* Passing `ULIMIT_T=no` to `configure` turns it off completely (or `ULIMIT_V`/`ULIMIT_S`)
* Values can be overriden by passing `TLIMIT` to `make` (or `VLIMIT`/`SLIMIT`), or defining new values per-directory in their respective Makefile

Also:
* Remove the essentially unused `ulimit -m` (unsupported in Linux and not enforced in macOS)
* Increase `VLIMIT` based on the number of processors (rather that playing whack-a-mole a la #4685 the next time someone runs the tests on a beefy machine)
* Factor setting `GC_MAXIMUM_HEAP_SIZE` into an overrideable variable.
* Restore the `basictests` Makefile

Claude did everything, with *lots* of hand-holding.  In retrospect, I probably should have done it myself lol.


dimpase:
lgtm

dimpase (approved):

MichaelABurr (approved):
This is a good idea to implement.

PR #4690, opened 2026-09-13
Add cohomology zero loci package

K3Ale:
The package provides methods for computing invariants of zero loci of homogeneous vector bundles inside rational homogeneous varieties, including:

- cohomology of homogeneous vector bundles restricted to the zero locus; 
- Hodge numbers and Hochschild numbers of the zero locus;
- Chern classes and intersection theory.

It depends on Schubert2, WeylGroups, and Python (used internally for some combinatorial computations, e.g. tensor/exterior/symmetric power decompositions).


K3Ale:
Hi @d-torrance,

I am trying to add my first M2 package, I apologize for any rookie mistake I made. In particular, since my package uses python scripts I wrote, I am having troubles with the usage of the python package:

While debugging CI failures for this PR I ran into a few issues, two of which I've fixed locally (not pushed yet, want to confirm the approach first), and one that I think is worth discussing since it doesn't seem specific to my package.

(1. Typos flagged by codespell: already fixed and pushed.)

2. numpy not found on Linux even though pipInstall succeeds. On autotools-ubuntu-24.04 and cmake-ubuntu-24.04, the log shows:

Successfully installed numpy-2.5.3
CohomologyZeroLociInHomogeneousVarieties: numpy was not found; attempting automatic installation with pipInstall...
error: ... requires the Python package numpy, which could not be found or installed automatically.

So pip install numpy succeeds, but the immediately following import "numpy" still fails. My guess is that the user site-packages directory isn't on the already-running interpreter's sys.path. I added a workaround that explicitly appends site.getusersitepackages() to sys.path before retrying the import, and locally this seems like the right fix; but I wanted to check whether this is a known limitation of pipInstall, or if there's a more idiomatic way to handle it within the Python package itself.

3. pipInstall fails outright on macOS (Homebrew). On autotools-macos-15 and cmake-macos-15-brew-clang, pipInstall "numpy" fails before even reaching the import step:

error: externally-managed-environment
× This environment is externally managed
...
hint: See PEP 668 for the detailed specification.

This is Homebrew's Python refusing system-wide pip install per PEP 668. The fix in point 2 above doesn't help here, since the install itself never completes. I don't have a good fix in mind for this one.

Do you have any suggestions for point 3, or overall comments?

Thank you very much for your time!

Alessandro Frassineti

d-torrance:
I wouldn't use `pipInstall` inside a package -- that method is intended for users to run themselves after using `setupVirtualEnvironment` and restarting the package with the "executable" option set.  See https://www.macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Python/html/___Python_sptutorial_co_spcreating_spa_spvirtual_spenvironment_spand_spinstalling_sp__Num__Py.html

For the purposes of the PR builds, I'd just install numpy system-wide:

```diff
diff --git a/.github/workflows/test_build.yml b/.github/workflows/test_build.yml
index a6cbafb70c..caa626dff6 100644
--- a/.github/workflows/test_build.yml
+++ b/.github/workflows/test_build.yml
@@ -65,7 +65,7 @@ jobs:
           brew update
           brew trust macaulay2/tap
           brew tap macaulay2/tap
-          brew install automake bison boost libtool tbb ccache ctags llvm make yasm libffi msolve googletest fplll eigen jansson r palp
+          brew install automake bison boost libtool tbb ccache ctags llvm make yasm libffi msolve googletest fplll eigen jansson r palp numpy
           brew install texinfo || true # sometimes post-install step fails
           brew install --only-dependencies macaulay2/tap/M2
           brew link factory --force
@@ -86,7 +86,7 @@ jobs:
                   libgivaro-dev libboost-regex-dev fflas-ffpack libflint-dev libmps-dev libfrobby-dev \
                   libsingular-dev singular-data libcdd-dev cohomcalg topcom 4ti2 libnormaliz-dev normaliz coinor-csdp \
                   libnauty-dev nauty lrslib polymake pipx phcpack w3c-markup-validator libtbb-dev qepcad libomp-16-dev \
-                  msolve libfplll-dev libjansson-dev r-base palp
+                  msolve libfplll-dev libjansson-dev r-base palp python3-numpy
 
 # ----------------------
 #   Steps common to all build variants
```

K3Ale:
Thank you very much, I tried to follow your instructions. Hoping to haven't mess up anything!
Best,
Alessandro Frassineti

K3Ale:
Hi @d-torrance, 

I am sorry to bother again but I have one more thing worth flagging: with numpy now installed on the macOS ARM runner (cmake-macos-15-brew-clang), check(12, "Python") fails with:

AttributeError: module 'numpy' has no attribute 'float128'. Did you mean: 'float16'?

at Python.m2:950-1010. 

This looks like a pre-existing issue in the Python package's own test suite, and this test was probably never exercised before on this runner since numpy wasn't installed there until this PR. 

Am I correct? Is there something I should do to go on and make my pull request successful?

d-torrance:
Don't worry about the Python failures -- I can take care of those in the next Python package update.  The numpy support isn't well-tested outside of Ubuntu 24.04.  I'm glad your PR uncovered them!

K3Ale:
Thank you for your reply.
Is it still possible to pull my package in the meantime? Or is there something else I can do?

d-torrance:
There will likely be more comments forthcoming after we do a more thorough review

PR #4693, opened 2026-09-15
Remove vestigial --enable-xcode configure option

d-torrance:
`--enable-xcode` has had no effect since 2010.

It gated a rule in `Macaulay2/d/Makefile.in` that built `interpret.a` for Mike to link against from Xcode:

```make
interpret.a: $(M2_OBJECTS); $(AR) rcs $@ $^
ifeq "@XCODE@" "yes"
# Mike uses Xcode, an IDE, under MacOS 10 to link against
all: interpret.a
endif
```

That rule was removed in d6b1b11bae (Apr 2010), replaced by `all: $(M2_LIBDEPS) M2inits.o`. Because it tested `@XCODE@` directly at substitution time rather than the make variable exported via `include/configuration`, the `AC_SUBST` lost its only consumer then. The commented-out CMake mirror (`option(XCODE "build Macaulay2/d/interpret.a")`) was dropped in c2cd12e465 (2020).

So the only remaining effect was advertising an option in `./configure --help` that promises a build artifact we no longer produce.

This removes both remaining references:
- `M2/configure.ac` — the `AC_SUBST`/`AC_ARG_ENABLE` pair
- `M2/include/configuration.in` — `XCODE=@XCODE@`

Unrelated Xcode mentions are left alone: the CMake `-GXcode` generator target in `M2/BUILD/frank/Makefile`, `xcode-select` in `M2/BUILD/mahrud/Makefile`, Doxygen docset settings, and an NCAlgebra note about the Xcode command line tools.

Verified that `autoconf` regenerates `configure` cleanly, the result has no `XCODE` references, and the neighboring `--enable-strip`/`--enable-memdebug` options still appear correctly in `--help`.


dimpase (approved):
lgtm

MichaelABurr (approved):

PR #4718, opened 2026-09-16
Document using Ninja as the default CMake generator

dimpase:
Document how to make Ninja the default CMake generator by setting `export CMAKE_GENERATOR=Ninja` in the shell or its startup file. This lets developers omit `-G Ninja` when configuring new build directories.

Update the source and build READMEs and the CMakeLists quick instructions. Explain that the environment setting applies across projects, explicit `-G` overrides it, and existing build directories retain their generator. Also replace the stale `INSTALL-CMake.md` reference with the build README and wiki: that file was deleted in commit `af1a5237e5` in favor of the [CMake build wiki page](https://github.com/Macaulay2/M2/wiki/Building-M2-from-source-using-CMake), but the CMakeLists comment still referred to it.

This PR changes documentation only; CI continues to select Ninja explicitly with `-GNinja`.

Validation: `git diff --check` passes; the final diff against development contains only the two READMEs and CMakeLists comments.


dimpase:
@d-torrance - trivial review, doc-only PR (also fixes doc errors)

d-torrance (changes requested):
The `M2/CMakeLists.txt` changes are good, but I think the other two changes should go in the wiki instead

dimpase:
OK, as requested. The wiki additions will be done after this is merged.

M2/CMakeLists.txt
```diff
@@ -2,16 +2,17 @@
...
-## See INSTALL-CMake.md for a comprehensive guide to building M2 using CMake.
+## See BUILD/README.md and the Macaulay2 wiki for CMake build instructions.
 ## Take a look at cmake/configure.cmake for a list of options
 ## and read cmake/check-libraries.cmake for a list of requirements.
 ##
-## Short instructions:
-##  1. cmake -GNinja -S . -B BUILD/cmake
+## Short instructions (from this directory; requires Ninja):
+##     export CMAKE_GENERATOR=Ninja
+##  1. cmake -S . -B BUILD/cmake
```
d-torrance: I think maybe keeping both options here would be good, i.e. users can either use `-GNinja` or export the shell variable. For a quick build where you're not planning on running cmake multiple times, the former is slightly fewer keystrokes

d-torrance (changes requested):

PR #4721, opened 2026-09-16
Reuse development build containers in Linux CMake CI

dimpase:
Linux CMake CI currently rebuilds from scratch. This change publishes a tested development build to GHCR and reuses its sources, objects, external dependencies and package installations for subsequent development commits and PR merge revisions. Package-only changes reinstall the affected packages and their dependents; a core change rebuilds affected objects, relinks M2 and reinstalls all packages. Package checks run every time.

Depends on #4704 and #4722, which are still open and included in this branch. macOS and Autotools jobs retain their existing behavior.

### Build and publication

- Build external fallback dependencies, including patched Factory, through M2's existing targets in the environment image. Preserve their libraries and completion stamps along with the full M2 build.
- Keep source/build paths fixed and preserve nanosecond timestamps in a PAX archive. Replace sources by checksum so unchanged inputs retain their timestamps.
- Select images by revision and environment key. Compiler/library settings, dependency patches and submodule revisions affect that key; ordinary package and build-orchestration edits do not. Missing images, incompatible environments, deleted files and file-type changes fall back to clean builds. An explicit cache-version file supports refreshes.
- Split build/installation and testing into separate 180-minute jobs with a two-day intermediate artifact. Run package checks, Core checks, C++ unit tests, ComputationsBook checks and Debian packaging before publication.
- Retain a bounded 512 MiB ccache. Limit build concurrency by CPU count and approximately 2 GiB RAM per job; restrict nested numerical-library threads to one.
- Use `WITH_MAPLE=OFF` and honor each package's cached-example policy. All packages remain selected. Keep generated example caches out of the source tree with `CacheExampleOutput=false`, preventing unnecessary reinstalls when restoring the original source snapshot.
- PR jobs cannot publish. Build containers receive neither registry credentials nor a Docker socket. Snapshots omit Git metadata and credentials. Docker login uses its Node 24 action version.

The registry namespace follows the repository owner: fork testing uses `ghcr.io/dimpase/m2-ci-cmake`; upstream will use `ghcr.io/macaulay2/m2-ci-cmake`. The first GHCR publication must be made public for anonymous pulls from fork PRs. Required-check names may need updating for the reusable Linux CMake workflow. The guide at `M2/BUILD/docker/incremental/README.md` documents setup, image selection, invalidation, refreshes and local reproduction.

### Validation

[...]

PR #4704, opened 2026-09-16
Track package dependencies for incremental CMake installation

dimpase:
Depends on #4705 (package dependency audit and CI); merge that PR first. This branch includes the prerequisite commits so the combined changes can be tested while both PRs are open.

Restarting `install-packages` after a failure currently reinstalls packages that already succeeded, including Style. Use file-based installation rules with a completion stamp written only after package installation and Info compression succeed. Unchanged completed packages are skipped; failed installs retry.

`all-packages` and `all-<package>` reuse these incremental installation rules, then run checks on every invocation. A failed installation prevents that package’s checks; a check failure can be retried without repeating a successful installation.

Generate and check in a CMake dependency manifest using an external M2 and the `readPackage` audit from #4705. CI regenerates the manifest in check mode, so configuring a fresh source build does not require an installed M2. Changes to imports require regenerating the manifest alongside the source changes.

Declared `PackageImports` and `PackageExports` become installation prerequisites. Selecting a package also selects its declared imports. Preserve `M2-core` ordering and track the binary and Core sources. Order the documentation bootstrap as Style, FirstPackage, then Macaulay2Doc; all remaining packages, including development-only packages, depend on Macaulay2Doc. Reject cycles introduced by combining declared imports with this bootstrap order.

Literal `needsPackage`, `loadPackage`, and `importFrom` calls in bodies, auxiliary files, documentation and tests provide transitive source dependencies. They do not impose installation ordering: lazy imports such as Polyhedra/FourTiTwo can legitimately be cyclic. Each install rule has its own CMake subdirectory to prevent repeated expansion of shared prerequisite rules in the full dependency graph.

Validation:
- Generator tests cover literal imports, comments, example/test strings, stale manifests, declared cycles and bootstrap cycles.
- Ninja and Unix Makefiles fixtures cover failed/restarted installs, unchanged rebuilds, edits/additions/removals of sources, Core and binary changes, options, compression failure, missing outputs, selected-package import closure, and cyclic source dependencies.
- Combined-target regression tests reproduce the previous unconditional reinstall and verify repeated checks, source edits, installation failures and check-failure retries under both Ninja and Makefiles.
- All 299 real distributed packages configure with both generators using a stub runtime; the generated Ninja combined installation/check graph contains one installation and one check command per package and is acyclic.
- The manifest contains 541 declared-import edges and 702 source-dependency edges. Its freshness check and `git diff --check` pass.

[...]

PR #4722, opened 2026-09-17
Detect optional Maple and convex installations in CMake

dimpase:
CMake currently has a TODO for Maple and the convex library, with no usability check. Add optional detection that requires Maple to execute an arithmetic probe successfully, then independently checks whether `with(convex)` loads. Missing executables, license failures, unexpected output and timeouts leave the dependency unavailable without failing configuration.

`WITH_MAPLE` defaults to `ON` (use if available); `OFF` skips execution and clears availability results even after an enabled configuration. Configure output distinguishes enabled, unavailable and explicitly disabled states. This switch controls build-host detection, not runtime package loading.

`MAPLE_EXECUTABLE` and `MAPLE_CONVEX_DIR` support custom installations; `MAPLE_PROBE_TIMEOUT` defaults to ten seconds. Results are `Maple_FOUND` and `Maple_Convex_FOUND`, with diagnostic logs under `CMakeFiles/Maple`. Probes rerun during configuration to avoid caching expired license availability. Cross-compilation does not execute the probes.

This PR targets development independently of #4704. CMake now preserves cached examples for MapleInterface, AdjointIdeal and Parametrization when Maple is unavailable or WITH_MAPLE=OFF; ConvexInterface also preserves them when convex is unavailable. Their installation commands receive RerunExamples=false, even if reruns are globally requested. Other packages retain the requested setting. Packages and independent checks remain enabled; missing or stale caches still report errors. Runtime MapleCommand/ConvexPath configuration remains separate.

#4721 depends on this integration and uses WITH_MAPLE=OFF instead of excluding MapleInterface and its dependents.

Validation: eleven tests using simulated executables cover missing Maple, zero-exit license failure, echoed input, nonzero exit, timeout, working Maple without convex, both available with paths containing spaces, changed availability, cross-compilation, disabled probing and enabled-to-disabled transitions. A generated-command test additionally verifies all five combinations of availability and rerun policy against the real package CMake rules. All pass locally at low priority. A lightweight Actions workflow runs them without a Maple license; actionlint and git diff --check pass. No licensed Maple installation was available for an end-to-end positive test.

Probe syntax follows Maplesoft's [command-line reference](https://www.maplesoft.com/support/help/errors/view.aspx?path=maple) and the existing ConvexInterface library-loading code.

Documentation is in `M2/BUILD/README.md`, linked from `M2/README.md`, the CMake entry point and the configuration option listing. It covers defaults, ON/OFF commands, status output, paths, timeouts, logs, cross-compilation and the boundary with runtime package behavior. The stale `INSTALL-CMake.md` reference in `M2/CMakeLists.txt` is replaced with the current wiki entry point.


dimpase (commented):
should be reviewed by someone with Maple installed

dimpase:
@d-torrance

dimpase:
@mikestillman - review?

PR #4485, opened 2026-07-26
Center of skew rings and kernel of RingMap to skew rings

joel-dodge:
Add `centerRing` method to compute the center of a ring. Currently supports skew polynomial rings (and commutative rings 😂 ). 

Then adds a new strategy for computing the kernel of a RingMap whose target is SkewCommutative by reducing to a map to the center of the target. 

This requires some awkward packaging since `kernel` is in the core and the strategy for computing the center of a skew commutative ring depends on the PushForward package. We rely here on the "hack" of calling `needsPackage` inside the hook only when needed although @d-torrance says "I'm not sure if this is a very elegant solution" and I am inclined to agree.

_Note while editing ringmap.m2 I took the opportunity to clean up ws formatting in the block declaring different kernel strategies for RingMap. Consider toggling the "hide whitespace" option while reviewing this PR._

PR #4724, opened 2026-09-17
Add "center" method to Core

d-torrance:
Previously defined in Graphs and GradedLieAlgebras

As discussed in #4485 

Cc: @joel-dodge


joel-dodge:
Nice!

M2/Macaulay2/packages/Macaulay2Doc/shared.m2
```diff
@@ -17,6 +17,7 @@ document { Key => extend, Headline => "extend an object",    methodstr,
     },
 }
 
+document { Key => center,       methodstr, SeeAlso => { "Graphs::center(Graph)", "GradedLieAlgebras::center(LieAlgebra)" } }
```
joel-dodge, line 20: I don't have a good model on where methods like this should be anchored as SubNodes in the documentation tree.

In the abtract, it would seem reasonable to me to have no node at all for bare `center` (not sure if this is even allowed) since all it should probably ever do is point to the various different override docs.

d-torrance: Yeah, there's not a great place.  I went ahead and just put it under [exterior algebras](https://macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Macaulay2Doc/html/_exterior_spalgebras.html) like you have it in #4485.

In #4485, you might consider moving it out of `shared.m2` (both the *Core* and *Macaulay2Doc* versions) anyway since it will actually be a *Core* method.  And I could see the `center` and `center(Ring)` nodes being merged, with the other packages' versions in `SeeAlso`.

joel-dodge:
This looks almost identical to a branch I had worked up for https://github.com/Macaulay2/M2/pull/4485. Context has rotted since July so I don't recall exactly why I walked it back and switched to `centerRing`- perhaps it was the question of where to place `center` as a doc subnode?

Regardless looks good to me and were I a reviewer I would approve.

PR #4725, opened 2026-09-18
Avoid checking out unused CMake dependency submodules

dimpase:
CMake currently initializes every Git submodule during configuration, even when it finds the corresponding system libraries. For example, a system FLINT installation avoids compiling FLINT but still causes its repository to be downloaded. This adds unnecessary work on slow connections and in fresh CI checkouts.

Select submodules after library detection and `BUILD_LIBRARIES` overrides. Always retain the bundled engine sources and Emacs package; initialize BDWGC, FLINT, Frobby, Givaro, FFLAS-FFPACK, and GoogleTest only when their fallback is needed. Declare those fallback build rules only when selected, so empty unused source directories no longer prevent configuration. GoogleTest is unnecessary when testing is disabled.

Previously installed fallback components retain their sources and build/test targets across reconfiguration, using their installation stamps. `GIT_SUBMODULE=OFF` and source archives remain supported when required sources are present; missing required sources now produce an actionable error. Remove the old `.nogit` placeholder workaround. Documentation on submodule selection, `BUILD_LIBRARIES`, and offline builds is staged in the [draft wiki pages on dimpase/M2](https://github.com/dimpase/M2/tree/wiki-cmake-submodules). These update the CMake build guide and related FAQ; they will be transferred to the upstream wiki after this PR is merged, incorporating any further review changes. The added README documentation has been removed from this PR.

Validation:
- Seven regression tests using local Git repositories pass, covering selective checkout, pinned revisions, missing dependencies, disabled testing, retained installations, offline/source-archive operation, and failed updates. These run in the CMake CI jobs without network access.
- Full local M2 configuration succeeds with all six optional submodule directories empty and the corresponding system libraries present. Their fallback targets are absent.
- Full configuration from a Git worktree initializes only the four required source submodules (using local repository mirrors for validation).
- `BUILD_LIBRARIES=FLINT` creates the fallback targets; a subsequent configuration with a simulated installation stamp retains its test target.
- Workflow lint and diff checks pass. Local checks ran at low priority. No full M2 rebuild was performed.

The development CMake workflow already starts with a non-recursive checkout and can benefit directly. The recursive checkout/source-snapshot handling in #4721 needs a separate follow-up to benefit from this selection as well. No network-speed improvement has been benchmarked.


dimpase (commented):
lgtm, @d-torrance

M2/BUILD/README.md
```diff
@@ -36,3 +36,21 @@ such as `x86_64-Linux-Ubuntu-8.10`.  (Actually, this is being changed...)
...
+it before configuring, for example `cmake -DBUILD_LIBRARIES=FLINT ../..`, then run
+`cmake --build . --target build-libraries`. Multiple library names use a quoted
+semicolon-separated list. Unneeded submodule-backed `build-<library>` targets
+are not created until that fallback is selected.
+
+For offline builds, initialize the required submodules first and configure with
+`-DGIT_SUBMODULE=OFF`; configuration reports any missing required sources.
+This option disables Git updates, not the source requirements. Source archives
+must likewise contain the required submodule contents. CI checkouts should avoid
+initializing all submodules recursively in advance to benefit from this selection.
```
d-torrance: I think this documentation all belongs in the wiki instead

M2/cmake/configure.cmake
```diff
@@ -21,7 +21,7 @@ option(MEMDEBUG		"Enable memory allocation debugging"	OFF)
 option(PROFILING	"Enable profiling build flags"		OFF)
 option(COVERAGE		"Enable Clang code coverage test"	OFF)
 option(GCOV		"Enable gcc/gcov code coverage"		OFF)
-option(GIT_SUBMODULE	"Update submodules during build"	ON)
+option(GIT_SUBMODULE	"Update required submodules during configuration"	ON)
```
d-torrance, line 24: I'd drop this change.  It is more accurate, but it messes up the spacing

M2/cmake/submodules.cmake
```diff
@@ -0,0 +1,39 @@
+# check-libraries.cmake has already applied BUILD_LIBRARIES overrides.
+# The engine and editor sources are always needed. ExternalProject sources are
+# needed only for missing libraries, or to keep testing a previous local build.
+set(_m2_required_submodules submodules/memtailor submodules/mathic submodules/mathicgb)
+set(GOOGLETEST_FOUND ${GTEST_FOUND})
```
d-torrance: Why did this line get moved from `build-libraries.cmake`?  That seems like a more obvious place for it

M2/cmake/tests/test_submodules.py
```diff
@@ -0,0 +1,130 @@
+"""Exercise selective CMake checkout with local Git repositories (no network)."""
```
d-torrance, line 1: What's the purpose of this python script?

d-torrance (changes requested):
Thanks for the contribution!  I have a few comments and questions

M2/BUILD/README.md
dimpase: OK, let's put all the documentation written here into a PR for the wiki repo.

d-torrance: The wiki doesn't support PR's, I think -- you can just edit it.

dimpase: anyhow, the part edited out by https://github.com/Macaulay2/M2/pull/4725/changes/fe2f7bf38ddada37eba52fc34081eead192bdf64 is now on my fork in a [branch](https://github.com/dimpase/M2/commit/46c0a08d4d42e9785bd8367e215f6cc5787cc265), and will be added to the wiki once this PR is merged.

M2/cmake/configure.cmake
dimpase: I'm not sure what spacing it's messing - can you explain?

M2/cmake/tests/test_submodules.py
dimpase: it's to be run in CI. 🤖 says:
```
The Python script tests the new submodule-selection logic using temporary local Git repositories,
without network access or compiling M2. 
It checks that required submodules are fetched at their pinned revisions, 
unused dependencies are skipped, and offline builds and failed updates are handled correctly. 
It runs in CI to catch regressions.
```

M2/cmake/configure.cmake
d-torrance: I don't know lol -- I think I thought this was going to be printed by `cmake --help` or something, but it's not.  Ignore my comment -- this is a good change

M2/cmake/submodules.cmake
dimpase: fixed by https://github.com/Macaulay2/M2/pull/4725/commits/c030b9e97ef0da882b8b496376f586f95806125f

M2/cmake/tests/test_submodules.py
d-torrance: This seems like overkill to check something as simple as "did we check out the right submodules"? Also the `cmake/tests` directory is new, and we don't have any similar Python scripts anywhere.

I'd recommend swapping this out for a simple shell command in the workflow. Claude suggested the following:

```yaml
- name: Check unused submodules were not checked out
  if: matrix.build-system == 'cmake'
  run: |
    for m in bdwgc flint frobby givaro fflas_ffpack googletest; do
      if git submodule status ../../submodules/$m | grep -qv '^-'; then
        echo "::error::submodule $m was checked out"; exit 1
      fi
    done
```

d-torrance: In fact, we could drop the "if" since the autotools build already has this behavior

PR #4727, opened 2026-09-19
Update existing workflows to Node 24 action versions

dimpase:
Existing workflows still select actions targeting Node 20, which GitHub runners now force onto Node 24 with a deprecation warning. Update the testbot workflow to `docker/login-action@v4` and `docker/build-push-action@v7`, and the style-assets workflow to `actions/checkout@v6`. These versions declare Node 24 support.

Validation: actionlint and `git diff --check` pass. The affected workflows are manually dispatched; publishing the testbot image and updating style assets have not been triggered for this change. Docker action metadata confirms the existing registry/login and build/push inputs remain supported.

The new container workflows in #4721 and its fork testing PR receive their corresponding login-action updates separately.


dimpase (commented):
👍🏽 @d-torrance - an easy GH Actions bump

d-torrance (approved):

PR #4728, opened 2026-09-19
Use external memtailor, mathic and mathicgb in CMake builds

dimpase:
CMake always builds the bundled memtailor, mathic and mathicgb sources, even when suitable packages are installed. Select installed libraries using their CMake configs or pkg-config metadata, with `MEMTAILOR_PROVIDER`, `MATHIC_PROVIDER` and `MATHICGB_PROVIDER` set to `AUTO` (default), `SYSTEM`, or `BUNDLED`.

Configuration reports each selection and checks the installed libraries' debug ABI flags through compile/link probes, including transitive dependency flags and mathicgb's TBB setting. System mathic requires system memtailor; system mathicgb requires both. `AUTO` propagates necessary fallbacks and incompatible explicit `SYSTEM` requests fail with a reason. Minimum versions are memtailor 1.4, mathic 1.5 and mathicgb 1.4.

Remove the engine's unconditional Debug ABI definitions and inherit the selected libraries' actual flags. A Debug M2 can therefore use a complete set of Release system libraries. Pinned bundled Debug consumers still require Debug dependency layouts, which selection enforces. Only bundled libraries are installed/exported by M2.

Keep the three library unit-test targets with either provider. For external libraries, copy the pinned tests and their support headers into the build directory so tests use installed library headers rather than accidentally mixing in bundled ones. Add 15 isolated provider regression tests and a lightweight CI job. Document configuration, dependency constraints, ABI behavior and prefix syntax in `M2/BUILD/README.md`.

Local validation on Gentoo, at low priority with two build jobs:
- Full Debug M2 core and test executables built against installed memtailor-1.4-r1, mathic-1.5-r1 and mathicgb-1.4-r2; the executable resolves all three shared libraries under `/usr/lib64`.
- All 548 C++ tests passed. Explicit MGB and F4 Groebner-basis computations in M2 passed.
- All 273 library tests passed using installed Release libraries; all 273 also passed with bundled Debug libraries and TBB disabled.
- Forced pkg-config discovery selected all three real installed packages with their Release ABI flags.
- All 15 provider regression tests, workflow lint and diff checks passed. Hosted CI is pending. A full installation/check of every M2 package was not run locally.

Related metadata improvements: Macaulay2/memtailor#23, Macaulay2/mathic#29 and Macaulay2/mathicgb#84. Config packages from those PRs are supported; existing pkg-config installations are supported when their metadata accurately describes the installed ABI. This PR incorporates the previously local memtailor integration; Autotools behavior and submodule acquisition are unchanged.


M2/BUILD/README.md
```diff
@@ -36,3 +36,66 @@ such as `x86_64-Linux-Ubuntu-8.10`.  (Actually, this is being changed...)
 
 Alternatively, see the [Docker guide](docker/README.md) for instructions on
 using Docker to build Linux container images running Macaulay2.
+
```
d-torrance: Any build docs should go on the wiki.

You brought up a good point on Zoom yesterday, by the way, about the wiki not being reviewed. So this is definitely something to consider. But for the time being, let's stick with current practice.

M2/cmake/tests/test_mathic_providers.py
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env python3
```
d-torrance: Why are we adding this Python script?

d-torrance (changes requested):
I like the basic idea of this PR - the cmake build should absolutely support system-installed mathic and friends, which the autotools build has done since the beginning (or at least, for a very long time).

A couple quick initial comments for now from my phone. I'll dig into the actual code later

M2/cmake/tests/test_mathic_providers.py
dimpase: it's called in CI, as a sanity check on pre-reqs.

Can potentially be used interactively, to debug an installation of external packages

d-torrance: It looks like it's writing a little fake library, though, and not using the system packages. I don't see the point of it.

I think it would be great if eventually we have some CI workflows that test the M2 builds both with system dependencies and with building our own dependencies, but I think that should probably come system-wide and not just for one little family of dependencies.

dimpase: the point is that we don't expect every distro out there to have a mathic package; it's inevitable that people might want to still have mathic installed separately, somewhere - and in such a case having a checker like this might help debugging the setup like this.

PR #4730, opened 2026-09-22
Unbundle Emacs support after #4692

dimpase:
Follow-up to #4692: install Emacs support independently from [Macaulay2/M2-emacs](https://github.com/Macaulay2/M2-emacs), instead of downloading, generating, and installing it as part of M2.

**Merge prerequisites:** #4692, #4725, #4453, and Macaulay2/M2-emacs#103. This draft includes the current commits from #4692, #4725, and #4453 so the submodule removal and selective-checkout adjustments can be reviewed together. It targets `development`; the prerequisite changes will drop out of the PR diff as they merge. Keep this draft until the standalone installation and migration path is accepted.

Remove the Emacs gitlink and `.gitmodules` entry, CMake/Autotools generation and installation rules, source-archive handling, explicit CI/Docker build targets, RPM bundled-file entries, and installation-layout settings. Keep Prism, Vim, and Pygments generation. Update #4725's selection logic and regression fixtures to work without an Emacs submodule.

`setup()` continues to configure command shells but no longer edits Emacs initialization. Retain `setupEmacs()` as a deprecated compatibility notice pointing to the independent package; it does not write files. Preserve existing Emacs documentation names as links to the independent interface, whose repository now owns the user guides and migration instructions. Runtime communication with Emacs is retained.

Validation (local builds ran at low CPU/I/O priority, with at most eight build jobs):
- Full CMake configuration from prepared source archives without the Emacs submodule, and a fresh `M2-core` build: passed.
- The freshly built M2 starts, evaluates arithmetic, and runs `setupEmacs()` and `setup()` with a temporary home: existing `.emacs` is unchanged, no `.emacs-Macaulay2` is created, and shell setup still works.
- Seven selective-submodule regression tests pass with no Emacs submodule in their source fixture.
- CMake `M2-editors` generates the remaining four grammar files without Emacs sources or a `M2-emacs` target.
- Autoreconf and full Autotools configuration pass; the default editor make target has nothing to install.
- Revised documentation parses with M2, and diff checks pass.
- The companion Emacs PR passes six installed-package tests, including a real M2 session.

A complete Autotools compilation, full package documentation build, distribution package builds, and interactive GUI testing have not been performed. Existing user configuration is not automatically rewritten; the companion README explains removing obsolete bundled-interface paths. MELPA submission remains a separate follow-up; archive and direct-Git installation do not depend on MELPA acceptance.

LSP validation: integrated #4453's launcher installation and sibling-M2 selection unchanged. A new CI smoke test initializes the real server, opens a document, requests completions, and shuts down, with a deliberately incorrect `M2` first on PATH. It passes against the newly installed launcher and fails against the old launcher. The independent Emacs package also passes real Eglot and lsp-mode initialization/completion tests against the rebuilt M2 server. Removed the remaining generic CMake install rule for staged Emacs files, which could otherwise copy an obsolete interface from a reused build tree.


dimpase:
I  did a manual check of this PR, together with its companion https://github.com/Macaulay2/M2-emacs/pull/103,
on a Gentoo Linux machine running systemwide emacs 31.1. Looked and worked OK.

dimpase:
from zulip: @mikestillman @d-torrance 
```
Michael Stillman: I think I'm good removing emacs support from M2.  Just need to make sure it is easy to install...
Michael Stillman: Not from M2, from the M2 git repo...!
Doug Torrance: I'm thinking that we still have setupEmacs work so most folks will never notice
Michael Stillman: yes, that would be good.  But it might need to download the mode...?
Dima Pasechnik: my PRs make setupEmacs into a noop, printing a warning on how to proceed
Doug Torrance: It can edit .emacs so that Emacs will download it when neededMichael Stillman: That sounds like a good way to do that!
```
OK, let's amend this here.

dimpase:
I don't like [e3cf252](https://github.com/Macaulay2/M2/pull/4730/commits/e3cf2526645e3c89030681b898e27caacfb201fe) - it messes up with git instead of using an Emacs mechanism to fetch plugin updates

dimpase:
🤖 wrote:
```
 I corrected M2 PR 4730: setupEmacs() now writes package.el commands to
  ~/.emacs. It installs M2 on first startup and checks for updates after Emacs
  becomes idle. M2-emacs PR 103 now builds the Macaulay2 package archive. A
  local archive test passed for both installation and upgrade.

  The public archive is not live yet. PR 103 must merge first; its publishing
  workflow then creates the archive that setupEmacs() uses.
```

dimpase:
force-push to add missing co-signed commit headers

PR #4731, opened 2026-09-22
Remove the generated Style assets workflow

d-torrance:
The seed-cache build that ran after merging into `development` failed:

https://github.com/Macaulay2/M2/actions/runs/35736246700

The `Gather Style assets` step died with

```
cp: cannot stat 'usr-dist/common/share/Macaulay2/Style/search-data.js': No such file or directory
```

Those builds set `SEED_CACHE_ONLY`, which skips `Install packages using Make`, so `installSearchData` never runs and the search index is never generated. The step itself was gated only on `autotools && Linux && success()`, so it ran anyway.

Rather than just guarding the step, this removes the whole mechanism — the `style-assets` artifact in `test_build.yml` and the `update-style-assets.yml` workflow it fed:

* `search-data.js` doesn't belong in the repository. `installPackage` regenerates it from scratch on every documentation install (clearing it when *Style* installs first), and its paths are relative to the install layout, so a committed copy would be overwritten before it was ever read. Distro packaging builds the documentation anyway, so it gets a correct index for free.
* `prism.js` is the one asset that genuinely needs a build-time toolchain, and it has been updated at each release since 2023 via `make -C Macaulay2/editors update-syntax-highlighting`, which already builds, copies, stages, and commits it. The workflow was a more elaborate reimplementation of that target which additionally required an artifact upload on every autotools-Linux run and `contents: write` on the repository.

Nothing else references the artifact, and the workflow had never actually run — it was only ever on `development`, and `search-data.js` was never committed to the source tree.

One known consequence: `defaultHEAD` in `Macaulay2/m2/html.m2` unconditionally emits a `<script>` tag for `search-data.js`, so that request 404s in a tree where the documentation was built without a full install. This only affects people building from source and is left as-is.

## AI Disclosure

Claude found the cause, made the change, and wrote this pull request message, with human oversight.


dimpase:
So you think this test is just not needed?

d-torrance:
> So you think this test is just not needed?

Exactly. I don't remember why I thought it was a good idea. The search data is generated during build and shouldn't be committed to git

dimpase (approved):
OK, fine.

PR #4732, opened 2026-09-22
Fix parallel make race installing Core/*.m2

d-torrance:
A [Launchpad PPA build on focal/arm64](https://launchpadlibrarian.net/877739526/buildlog_ubuntu-focal-arm64.macaulay2_1.26.06+git202609211330-0ppa202609181047~ubuntu20.04.1_BUILDING.txt.gz) failed with:

```
/usr/bin/install -c -m 644 Core.m2 .../share/Macaulay2/Core/Core.m2
/usr/bin/install: cannot create regular file '.../share/Macaulay2/Core/Core.m2': No such file or directory
make[4]: *** [Makefile:89: .../share/Macaulay2/Core/Core.m2] Error 1
```

This is a race in `Macaulay2/m2/Makefile.in`, exposed by `make -j8`:

```make
@pre_packagesdir@       :; @INSTALL@ -d $@
@pre_packagesdir@/Core  : ; @INSTALL@ -d $@
@pre_packagesdir@/Core/% : % ; @INSTALL_DATA@ $^ $@
all: @pre_packagesdir@ @pre_packagesdir@/Core $(addprefix @pre_packagesdir@/Core/, $(FILES_TO_COPY))
```

The pattern rule copying the files has no prerequisite on the rule creating the directory, and the `Core` rule has none on its parent. Listing all three groups as prerequisites of `all` expresses no ordering, so under `-j` make can start the copies concurrently with `install -d`.

It usually goes unnoticed because the directory survives from an earlier build, or because the `tvalues.m2` rule — which *does* list `@pre_packagesdir@/Core` as a prerequisite — happens to get scheduled first. Nothing guarantees either.

The fix uses order-only prerequisites, the idiom already used in `Macaulay2/tests/normal/Makefile.in` for the parallel `Core/tests` copies. It also switches `$^` to `$<` in the copy recipe: `$^` was already the wrong automatic variable there, and would expand to include the directory once it became a prerequisite.

### Testing

- Reproduced the race in an isolated harness mirroring the rule structure, with `install -d` slowed to widen the window: **5/5** runs failed with the identical error before the change, **0/5** after.
- Ran the real regenerated rules against an empty prefix with `-j16`, five times: no failures, all files copied.
- Re-ran `make -j16 all` on a populated build tree: no-op, and `tvalues.m2` is not needlessly regenerated by the order-only change on its rule.

The commit message and this description were written by Claude Opus 5.


dimpase (approved):
lgtm

